PR #3, opened 2020-01-31
feat(decorator): add a space around decorator

deorukhkarchaitanya:


src/decorators/SpaceAround/SpaceAround.native.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import styled from 'styled-components';
```
kamleshchandnani: I would suggest let's move it to `storybook/native/index.js` itself since it's not going to be used by any other component. No point of abstracting it

deorukhkarchaitanya: Yeah makes more sense

PR #5, opened 2020-01-31
build(android): fix android entry point which was breaking apk builds

deorukhkarchaitanya:


android/app/build.gradle
```diff
@@ -76,7 +76,7 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    entryFile: "storybook/native",
+    entryFile: "storybook/native/index.js",
```
kamleshchandnani: Mind renaming it to `main.js` now? So it's similar to what storybook on web expects

deorukhkarchaitanya: Yeah cool

PR #7, opened 2020-02-03
feat: tokens for design system

shubhnik:
Adds tokens for colors and typography


src/tokens/fonts.js
```diff
@@ -0,0 +1,33 @@
+const fonts = {
+  family: {
+    lato: 'Lato',
+  },
+  sizes: {
```
deorukhkarchaitanya: Shouldn't this be ```size```?

src/tokens/colors.js
```diff
@@ -0,0 +1,169 @@
+const colors = {
+  azzure: {
+    100: '',
```
deorukhkarchaitanya: Let's not keep it empty since it will cause a crash

src/tokens/fonts.js
```diff
@@ -0,0 +1,33 @@
...
+    xxxl: '24px',
+    xxxxl: '32px',
+  },
+  weight: {
+    light: '300',
+    regular: '400',
+    bold: '700',
+  },
+  lineHeight: {
+    s: 1.2,
```
deorukhkarchaitanya: We should use ```px``` since this will be consumed in styled-components
```suggestion
    s: '1.2px',
```

src/tokens/fonts.js
```diff
@@ -0,0 +1,33 @@
...
+    m: 1.25,
+    l: 1.5,
+  },
+  letterSpaceing: {
+    s: 0,
+    m: 0.5,
```
deorukhkarchaitanya: Same with this

```suggestion
    m: '0.5px',
```

src/tokens/index.js
```diff
@@ -0,0 +1,4 @@
+import colors from './colors';
+import fonts from './fonts';
+
+export { colors, fonts };
```
deorukhkarchaitanya: I think this should not be plural. When we consume it, it would make more sense if it wasn't plural

eg. 
```color.green.300```
```font.size.m```

File: src/tokens/colors.js (hunk 2 above)
Comment (kamleshchandnani) [reply src/tokens/colors.js]: Agreed. Right now we just wanted to see the big picture. We can remove it

File: src/tokens/fonts.js (hunk 1 above)
Comment (kamleshchandnani) [reply src/tokens/fonts.js]: This should be size

File: src/tokens/fonts.js (hunk 3 above)
Comment (kamleshchandnani) [reply src/tokens/fonts.js]: line height never have unit associated with them

File: src/tokens/index.js (hunk 5 above)
Comment (kamleshchandnani) [reply src/tokens/index.js]: It's a collection of properties stored in object literal. So it should be plural. Because we are saying in the collection of colors choose green of 300

File: src/tokens/fonts.js (hunk 3 above)
Comment (deorukhkarchaitanya) [reply src/tokens/fonts.js]: RN throws a warning that it needs a unit. I just tried it out

File: src/tokens/index.js (hunk 5 above)
Comment (deorukhkarchaitanya) [reply src/tokens/index.js]: Android follows a similar pattern https://developer.android.com/reference/android/R.color

So does treebo :P https://github.com/treebohotels/leaf-ui/blob/master/src/theme/theme.web.js

Comment (kamleshchandnani) [reply src/tokens/index.js]: Lol we made a mistake at Treebo hence recommended plural 😄

File: src/tokens/spacings.js [hunk 6]
```diff
@@ -0,0 +1,14 @@
+const unit = 8;
```
kamleshchandnani: `export const unit = 8;`

src/tokens/spacings.js
```diff
@@ -0,0 +1,14 @@
...
+
+const spacings = {
+  xxs: `${unit / 4}px`,
+  xs: `${unit / 2}px`,
+  s: `${unit}px`,
+  m: `${(3 * unit) / 2}px`,
+  l: `${unit * 2}px`,
+  xl: `${(5 * unit) / 2}px`,
+  xxl: `${unit * 3}px`,
+  xxxl: `${unit * 4}px`,
```
kamleshchandnani: Let's keep these as absolute values. I remember I had only asked you to do it based on computations but it's very hard to read and becomes complicated. Let's go for simplicity

PR #13, opened 2020-02-10
feat(text): Add Text component

deorukhkarchaitanya:


src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,43 @@
...
+  align-self: ${props.align};
+`,
+);
+
+Text.propTypes = {
+  children: PropTypes.node.isRequired,
+  weight: PropTypes.oneOf(Object.keys(theme.fonts.weight)),
+  size: PropTypes.oneOf(Object.keys(theme.fonts.size)),
+  letterSpacing: PropTypes.oneOf(Object.keys(theme.fonts.letterSpacing)),
+  color: PropTypes.oneOf(flattenArray(Object.values(theme.colors).map(Object.values))),
```
deorukhkarchaitanya: @kamleshchandnani @shridhar52war @shubhnik  I need some feedback/suggestions on this

PropTypes need to understand what are the acceptable values. Color must be a string value that is defined in our design system colors.

https://jsfiddle.net/nz05xutp/8/

Check the fiddle above to understand how it works and what the output is(check the console)

LMK if this is okay or we can do something better

deorukhkarchaitanya:
Let's keep line-height on hold and get this component merged so that it can be consumed at other places

src/atoms/Text/Text.native.js
shridhar52war: LGTM

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,43 @@
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
+import theme from '../../tokens/theme';
+import { Text as NativeText } from 'react-native';
+import { flattenArray } from '../../utils/index';
+
+const fontFamilies = {
```
shridhar52war: Should we move this to a different file `fontFamilies`?

deorukhkarchaitanya: We can but it will not be used anywhere else

src/atoms/Text/Text.stories.js
```diff
@@ -0,0 +1,36 @@
...
+};
+
+storiesOf('Text', module)
+  .addParameters({
+    component: Text,
+  })
+  .addDecorator(withKnobs)
+  .add('default', () => (
+    <Text
+      size={select('Size', Object.keys(theme.fonts.size), 'l')}
```
kamleshchandnani: I'm a bit afraid about keeping the text component too flexible. The reason is if you refer our Figma designs they have like 5 variations of `Text` component. How about if we just have the api something like below
```jsx
<Text
type="xxs/xs/s/m/l/xxl"
color="our colors"
truncate=true/false
align="not sure if this makes sense for react native"
as="p/div/span - this is for web mostly but api should remain consistent"
</Text>
```
All the font weights, sizes let them be the implementation details.

Thoughts?

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,43 @@
...
+  align: PropTypes.oneOf(['flex-start', 'center', 'flex-end']),
+};
+
+Text.defaultProps = {
+  weight: 'regular',
+  size: 'm',
+  letterSpacing: 's',
+  underline: false,
+  align: 'flex-start',
+  testID: 'ds-text',
```
kamleshchandnani: for `testIds` we can basically create a file in `_helpers/automation-attributes` and export a function from there and use that in all the components.
[Something like this](https://github.com/auth0/cosmos/blob/776187d1fca7e61412903500125a5e9dc84ea7a2/core/components/_helpers/automation-attribute.ts) and see the usage [here](https://github.com/auth0/cosmos/blob/776187d1fc/core/components/atoms/text/text.tsx#L36)

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,43 @@
...
+  regular: `Lato-Regular`,
+  bold: `Lato-Bold`,
+};
+
+const Text = styled(NativeText)(
+  (props) => `
+  font-family: ${fontFamilies[props.weight]};
+  font-size: ${props.theme.fonts.size[props.size]};
+  letter-spacing: ${props.theme.fonts.letterSpacing[props.letterSpacing]};
+  color: ${props.color || props.theme.colors.shade[800]};
```
kamleshchandnani: As per the figma file the color is `tone` and not `shade`? 🤔

src/atoms/Text/Text.native.js
kamleshchandnani: We'll discuss this in person

src/atoms/Text/Text.native.js
deorukhkarchaitanya: I'm not sure what would be the advantage of doing that?

@shubhnik test expert, opinion please!

src/atoms/Text/Text.stories.js
deorukhkarchaitanya: `xxs/xs/s/m/l/xxl` are variants for font size. Weights are still independent of this. There is a `H6`  & `T2` with font size as `m` but font-weight as `bold` & `regular`. How do we interpret weight based on solely the size?

deorukhkarchaitanya: It depends on what `align' does on web. On react-native `align-self` is something that we can probably use often which is why I exposed that. We could probably have a single `align` that behaves platform specific

src/atoms/Text/Text.native.js
deorukhkarchaitanya: I spoke to Nishant regarding this. He will be removing the color column from there. Color for the text will be defined by the component consuming it. I've kept shade only as a fallback.

src/atoms/Text/Text.native.js
shubhnik: Lol no one becomes expert in writing 2 test cases 😆 
I had a concern which I raised in this PR here https://github.com/razorpay/blade/pull/17#discussion_r379844338

deorukhkarchaitanya:
@kamleshchandnani @shridhar52war @shubhnik I've made all the changes as per our discussions.

The only pending change is for color support. We still need to make a decision on dot notation and absolute colors.

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
+import theme from '../../tokens/theme';
+import { Text as NativeText } from 'react-native';
+import { getAllColorNames, resolveColorFromString } from '../../utils/colors';
+import automation from '../../_helpers/automation-attributes';
+
+const fontFamilies = {
+  light: `Lato-Regular`,
```
shubhnik: Shouldn't it be `light: Lato-Light`?

deorukhkarchaitanya: DS only has Regular & Bold

src/atoms/Button/Button.native.js
```diff
@@ -1,31 +1,26 @@
...
+  const theme = useContext(ThemeContext);
 
-const Button = ({ onClick, children, ...rest }) => (
-  <StyledButton onPress={onClick} {...rest}>
-    <StyledText {...rest}>{children}</StyledText>
-  </StyledButton>
-);
+  return (
+    <StyledButton onPress={onClick} testID="ds-button">
+      <Text color={theme.colors.tone[900]} align="center">
```
kamleshchandnani: not `tone.900`?

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
+import theme from '../../tokens/theme';
+import { Text as NativeText } from 'react-native';
+import { getAllColorNames, resolveColorFromString } from '../../utils/colors';
+import automation from '../../_helpers/automation-attributes';
+
+const fontFamilies = {
```
kamleshchandnani: put this in `fonts.js`?

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,57 @@
...
+  );
+};
+
+Text.propTypes = {
+  children: PropTypes.string,
+  weight: PropTypes.oneOf(Object.keys(theme.fonts.weight)),
+  size: PropTypes.oneOf(['xxsmall', 'xsmall', 'small', 'medium', 'large']),
+  color: PropTypes.oneOf(getAllColorNames()),
+  underline: PropTypes.bool,
+  align: PropTypes.oneOf(['flex-start', 'center', 'flex-end']),
```
kamleshchandnani: align logically is one of `'left', 'center', 'right', 'justify'`. Is it that react-native has different align props?

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,57 @@
...
+    </StyledText>
+  );
+};
+
+Text.propTypes = {
+  children: PropTypes.string,
+  weight: PropTypes.oneOf(Object.keys(theme.fonts.weight)),
+  size: PropTypes.oneOf(['xxsmall', 'xsmall', 'small', 'medium', 'large']),
+  color: PropTypes.oneOf(getAllColorNames()),
+  underline: PropTypes.bool,
```
kamleshchandnani: `is*` for booleans? so this becomes `isUnderlined`.


Also is it because link uses this that's why we expose this prop?

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,57 @@
...
+  font-family: ${fontFamilies[props.weight]};
+  font-size: ${props.theme.fonts.size[props.size]};
+  color: ${resolveColorFromString(props.theme, props.color) || props.theme.colors.shade[800]};
+  text-decoration-line: ${props.underline ? 'underline' : 'none'};
+  align-self: ${props.align};
+`,
+);
+
+const Text = ({ weight, size, underline, align, testID, color, children }) => {
+  return (
```
kamleshchandnani: shouldn't we expose just `size` and `color`? Is it because button has some weird styles that's why we are exposing everything?

src/tokens/fonts.js
```diff
@@ -3,30 +3,30 @@ const fonts = {
...
-    xxs: '10px',
-    xs: '12px',
-    s: '13px',
-    m: '14px',
-    l: '16px',
-    xl: '18px',
-    xxl: '20px',
-    xxxl: '24px',
-    xxxxl: '32px',
+    xxsmall: '10px',
```
kamleshchandnani: looks beautiful 🌟

src/tokens/theme.web.js
```diff
@@ -57,6 +57,7 @@ const lightTheme = {
       600: colors.mustard[600],
       900: colors.mustard[900],
     },
+    ...colors,
```
kamleshchandnani: this should come on the top.

src/utils/colors.js
```diff
@@ -0,0 +1,20 @@
...
+
+const colorRange = [100, 200, 300, 400, 500, 600, 700, 800, 900];
+
+const getAllColorNames = () => {
+  return Object.keys(theme.colors)
+    .map((color) => colorRange.map((range) => `${color}.${range}`))
+    .reduce((acc, val) => [...acc, ...val]);
+};
+
+const resolveColorFromString = (currentTheme, colorString) => {
```
kamleshchandnani: rename it to `getColor = (theme, color)`

src/utils/colors.js
```diff
@@ -0,0 +1,20 @@
...
+  const colorEntities = colorString.split('.');
```
kamleshchandnani: `const [base,shade] = colorString.split('.');`

src/utils/colors.js
```diff
@@ -0,0 +1,20 @@
+import theme from '../tokens/theme';
+
+const colorRange = [100, 200, 300, 400, 500, 600, 700, 800, 900];
+
+const getAllColorNames = () => {
```
kamleshchandnani: `getColorKeys`

src/atoms/Text/Text.web.js
```diff
@@ -0,0 +1,3 @@
+const Text = () => null;
+
+export default Text;
```
kamleshchandnani: Thank You 🙏

src/atoms/Text/Text.native.js
shubhnik: `<Button />` will most probably need `size`, `color` and `weight`.

src/atoms/Text/Text.native.js
shubhnik: why are we keeping it here? any reasons?

deorukhkarchaitanya: Because it will not be used anywhere else but here

src/atoms/Button/Button.native.js
deorukhkarchaitanya: It's the dummy button component, I'll just reset it to what's on master so that @shubhnik 's PR gets merged without conflict

src/atoms/Text/Text.native.js
deorukhkarchaitanya: Yeah. These are the prop values for RN

src/atoms/Text/Text.native.js
deorukhkarchaitanya: I'll update it to isUnderlined. 

and yes, its because Link needs it

src/atoms/Text/Text.native.js
deorukhkarchaitanya: Yes

src/tokens/theme.web.js
deorukhkarchaitanya: Sure

src/atoms/Text/Text.native.js
deorukhkarchaitanya: This will only be used here. Do we still need to put it there?

src/atoms/Text/Text.native.js
kamleshchandnani: But what if we expose `left, right, center` and then handle it internally? Because on web this doesn't work. It's the same conversation as `onPress` vs `onClick`

src/atoms/Text/Text.native.js
kamleshchandnani: Now the more I think about it the more I get inclined towards private and public props. The pattern we can follow is
`_privatepropName`
`publicPropName`
anything that starts with underscore is private and if someone uses it(consumer apps) then it will be easier to catch in PR reviews. With that we can make our consumer API much more leaner and also have flexobility.

<Text
   size="small"
   color="azzure.100"
   _weight="bold"
  _isUnderlined={true}
/>

So the consumers can get the max benefit without writing size weight and what not everytime.

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,70 @@
...
+
+const StyledText = styled(NativeText)`
+  font-family: ${styles.fontFamily};
+  font-size: ${styles.fontSize};
+  color: ${styles.color};
+  text-decoration-line: ${styles.textDecorationLine};
+  align-self: ${styles.alignSelf};
+`;
+
+const Text = ({ _weight, size, _isUnderlined, align, testID, color, children }) => {
```
kamleshchandnani: put all the `_` props to the end. Easy to read as well

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <StyledText
+      _weight={_weight}
```
kamleshchandnani: put all the `_` props to the end. Easy to read as well

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,70 @@
...
+      color={color}
+      {...automation(testID)}
+    >
+      {children}
+    </StyledText>
+  );
+};
+
+Text.propTypes = {
+  _weight: PropTypes.oneOf(Object.keys(baseTheme.fonts.weight)),
```
kamleshchandnani: put all the `_` props to the end. Easy to read as well

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,70 @@
...
+  fontFamily({ theme, _weight }) {
+    return theme.fonts.family[_weight];
+  },
+  fontSize({ theme, size }) {
+    return theme.fonts.size[size];
+  },
+  color({ theme, color }) {
+    if (color) {
+      return getColor(theme, color);
+    } else return theme.colors.shade[800];
```
kamleshchandnani:
```
else {
}
```
put it in braces?

src/atoms/Text/Text.native.js
kamleshchandnani: Lol don't resolve conversations which are discussions. I keep forgetting what was the concern I had and then open all the conversations 😄

deorukhkarchaitanya: Hahaha sorry didn't see your comment here

deorukhkarchaitanya: Yeah, we can have it as left, right and center

src/tokens/fonts.js
```diff
@@ -1,32 +1,33 @@
 const fonts = {
   family: {
-    lato: 'Lato',
+    regular: `Lato-Regular`,
```
kamleshchandnani: this puts a constraint for us to add a new family. The ideal structure should be
```js
family: {
lato: {
regular:'Lato-Regular',
bold:'Lato-Bold',
}
}
```

so you can add n number of families

deorukhkarchaitanya: Oh okay

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,74 @@
...
+      return 'none';
+    }
+  },
+  alignSelf({ align }) {
+    return align;
+  },
+};
+
+const StyledText = styled(NativeText)`
+  font-family: ${styles.fontFamily};
```
kamleshchandnani: I couldn't see `_lineHeight` prop?

deorukhkarchaitanya:
I've added _lineHeight & _letterSpacing props as private props which we'll override for button & heading

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,107 @@
...
+  testID: PropTypes.string,
+  _weight: PropTypes.oneOf(Object.keys(baseTheme.fonts.weight)),
+  _isUnderlined: PropTypes.bool,
+  _letterSpacing: PropTypes.oneOf(Object.keys(baseTheme.fonts.letterSpacing)),
+  _lineHeight: PropTypes.oneOf(Object.keys(baseTheme.fonts.lineHeight)),
+};
+
+Text.defaultProps = {
+  size: 'large',
+  align: 'left',
```
dhruvdutt: Add default `color`.

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,107 @@
...
+  testID: 'ds-text',
+  _weight: 'regular',
+  _isUnderlined: false,
+  _letterSpacing: 'small',
+  _lineHeight: 'large',
+};
+export default Text;
```
dhruvdutt: Newline, please. 😬

src/atoms/Text/Text.stories.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import Text from './Text';
```
dhruvdutt: Move relative imports after package imports. You can also add an eslint rule for that.

src/atoms/Text/Text.web.js
dhruvdutt: TODO. 😅

src/tokens/fonts.js
```diff
@@ -1,32 +1,35 @@
 const fonts = {
   family: {
-    lato: 'Lato',
+    lato: {
+      regular: `Lato-Regular`,
```
dhruvdutt: No backticks for static literals. Add a linting rule for this.

deorukhkarchaitanya: We should add these rules in universe configs @kamleshchandnani

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,87 @@
...
+      _letterSpacing={_letterSpacing}
+      {...automation(testID)}
+    >
+      {children}
+    </StyledText>
+  );
+};
+
+Text.propTypes = {
+  children: PropTypes.string,
```
dhruvdutt:
```suggestion
  children: PropTypes.string.isRequired,
```

deorukhkarchaitanya: Shouldn't be required. RN Text component doesn't impose this restriction anyway

src/_helpers/testing.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { render } from '@testing-library/react-native';
+import { ThemeProvider } from 'styled-components/native';
```
dhruvdutt:
```suggestion
import { ThemeProvider } from 'styled-components';
```

src/atoms/Text/Text.native.js
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
```
dhruvdutt:
```suggestion
import styled from 'styled-components';
```

src/_helpers/testing.js
deorukhkarchaitanya: We need /native

src/atoms/Text/Text.native.js
deorukhkarchaitanya: We need /native

kamleshchandnani: This doesn't work in react native. There's an alternative to make it work [here](https://styled-components.com/docs/basics#simpler-usage-with-the-metro-bundler). But that's again an issue with storybook and the consumers will also need to make their metro work like this.

PR #15, opened 2020-02-10
feat: theme for design system [ME-4243]

shubhnik:
Inbuilt and custom themes can be injected to the Design System.

PR #17, opened 2020-02-11
feat(link): add Link component

deorukhkarchaitanya:
This PR is chained to PR https://github.com/razorpay/blade/pull/15

### DIFF
--- a/src/atoms/Flex/Flex.native.js
+++ b/src/atoms/Flex/Flex.native.js
@@ -1,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+const Flex = styled(
+  ({
+    alignContent,
+    alignItems,
+    alignSelf,
+    display,
+    flex,
+    flexBasis,
+    flexDirection,
+    flexGrow,
+    flexShrink,
+    flexWrap,
+    justifyContent,
+    order,
+    children,
+    ...props
+  }) => React.cloneElement(children, props),
+)`
+  ${(props) => (props.alignContent ? `align-content: ${props.alignContent};` : '')};
+  ${(props) => (props.alignItems ? `align-items: ${props.alignItems};` : '')};
+  ${(props) => (props.alignSelf ? `align-self: ${props.alignSelf};` : '')};
+  ${(props) => (props.flex ? `flex: ${props.flex};` : '')};
+  ${(props) => (props.flexBasis ? `flex-basis: ${props.flexBasis};` : '')};
+  ${(props) => (props.flexDirection ? `flex-direction: ${props.flexDirection};` : '')};
+  ${(props) => (props.flexGrow ? `flex-grow: ${props.flexGrow};` : '')};
+  ${(props) => (props.flexShrink ? `flex-shrink: ${props.flexShrink};` : '')};
+  ${(props) => (props.flexWrap ? `flex-wrap: ${props.flexWrap};` : '')};
+  ${(props) => (props.justifyContent ? `justify-content: ${props.justifyContent};` : '')};
+`;
+
+Flex.propTypes = {
+  alignContent: PropTypes.oneOf([
+    'center',
+    'flex-end',
+    'flex-start',
+    'space-around',
+    'space-between',
+    'stretch',
+  ]),
+  alignItems: PropTypes.oneOf(['baseline', 'center', 'flex-end', 'flex-start', 'stretch']),
+  alignSelf: PropTypes.oneOf(['baseline', 'center', 'flex-end', 'flex-start', 'stretch']),
+  flex: PropTypes.number,
+  flexBasis: PropTypes.string,
+  flexDirection: PropTypes.oneOf(['column-reverse', 'column', 'row-reverse', 'row']),
+  flexGrow: PropTypes.number,
+  flexShrink: PropTypes.number,
+  flexWrap: PropTypes.oneOf(['nowrap', 'wrap-reverse', 'wrap']),
+  justifyContent: PropTypes.oneOf([
+    'center',
+    'flex-end',
+    'flex-start',
+    'space-around',
+    'space-between',
+    'space-evenly',
+  ]),
+  children: PropTypes.element.isRequired,
+};
+
+Flex.defaultProps = {
+  display: 'flex',
+  flexDirection: 'column',
+};
+
+export default Flex;
--- a/src/atoms/Link/Link.native.js
+++ b/src/atoms/Link/Link.native.js
@@ -33,7 +33,7 @@ const Link = ({ children, onClick, size, testID }) => {
         {...automation(testID)}
       >
         <Text
-          isUnderlined={active}
+          _isUnderlined={active}
           color={active ? activeTextColor : inactiveTextColor}
           size={linkToTextSize[size]}
         >
--- a/src/atoms/Link/Link.native.js
+++ b/src/atoms/Link/Link.native.js
@@ -5,12 +5,6 @@ import automation from '../../_helpers/automation-attributes';
 import Flex from '../Flex/Flex.native';
 import { TouchableOpacity } from 'react-native';
 
-const linkToTextSize = {
-  small: 'xsmall',
-  medium: 'medium',
-  large: 'large',
-};
-
 const Link = ({ children, onClick, size, testID }) => {
   const [active, setActive] = useState(false);
   const inactiveTextColor = 'primary.800';
@@ -35,7 +29,7 @@ const Link = ({ children, onClick, size, testID }) => {
         <Text
           _isUnderlined={active}
           color={active ? activeTextColor : inactiveTextColor}
-          size={linkToTextSize[size]}
+          size={size}
         >
           {children}
         </Text>
@@ -47,7 +41,7 @@ const Link = ({ children, onClick, size, testID }) => {
 Link.propTypes = {
   children: PropTypes.string.isRequired,
   onClick: PropTypes.func,
-  size: PropTypes.oneOf(['small', 'medium', 'large']),
+  size: PropTypes.oneOf(['xsmall', 'small', 'medium', 'large']),
   testID: PropTypes.string,
 };
 
--- a/src/atoms/Link/Link.native.js
+++ b/src/atoms/Link/Link.native.js
@@ -1,0 +1,60 @@
+import React, { useState, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import Text from '../Text';
+import automation from '../../_helpers/automation-attributes';
+import Flex from '../Flex/Flex.native';
+import { TouchableOpacity } from 'react-native';
+
+const linkToTextSize = {
+  small: 'xsmall',
+  medium: 'medium',
+  large: 'large',
+};
+
+const Link = ({ children, onClick, size, testID }) => {
+  const [active, setActive] = useState(false);
+  const inactiveTextColor = 'primary.800';
+  const activeTextColor = 'primary.600';
+
+  const onPressIn = useCallback(() => {
+    setActive(true);
+  }, [setActive]);
+  const onPressOut = useCallback(() => {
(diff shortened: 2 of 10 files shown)


src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,71 @@
...
+const Link = ({ children, visited, onClick, size }) => {
+  const theme = useContext(ThemeContext);
+  const [active, setActive] = useState(false);
+
+  const onPressIn = useCallback(() => {
+    setActive(true);
+  }, [setActive]);
+  const onPressOut = useCallback(() => {
+    setActive(false);
+  }, [setActive]);
```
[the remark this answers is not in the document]
deorukhkarchaitanya: We don't need any underlayColor. We need a touchable that does not do anything but just gives us pressIn/Out events.

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,71 @@
...
+
+const StyledTouchable = styled.TouchableOpacity`
+  align-self: flex-start;
+`;
+
+const resolveSizeToTextProps = (linkSize) => {
+  //TODO: Add lineHeight prop for all cases
+  switch (linkSize) {
+    case 's':
+      return { size: 'xs', weight: 'regular' };
```
shridhar52war: For `case 's'`, it should return `{ size: 's', weight: 'regular' }`?

deorukhkarchaitanya: No. I had to declare a mapping because in the designs there are three sizes each mapping to text size of xs,m and l. Nothing for s

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,71 @@
+import React, { useContext, useState, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import Text from '../Text';
+import styled, { ThemeContext } from 'styled-components/native';
+
+const StyledTouchable = styled.TouchableOpacity`
+  align-self: flex-start;
```
kamleshchandnani: Let's create `flex` component as we had discussed today

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,71 @@
+import React, { useContext, useState, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import Text from '../Text';
+import styled, { ThemeContext } from 'styled-components/native';
```
kamleshchandnani: You can probably do [this](https://styled-components.com/docs/basics#simpler-usage-with-the-metro-bundler)

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,71 @@
...
+      </Text>
+    </StyledTouchable>
+  );
+};
+
+Link.propTypes = {
+  visited: PropTypes.bool,
+  children: PropTypes.node.isRequired,
+  onClick: PropTypes.func,
+  size: PropTypes.oneOf(['s', 'm', 'l']),
```
kamleshchandnani: we need to now discuss whether we should call all these `s/m/l` variants as `size` prop or `type` prop. Since it's already different in Text and now here Link 😆

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,71 @@
+import React, { useContext, useState, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import Text from '../Text';
+import styled, { ThemeContext } from 'styled-components/native';
+
+const StyledTouchable = styled.TouchableOpacity`
+  align-self: flex-start;
+`;
+
+const resolveSizeToTextProps = (linkSize) => {
```
kamleshchandnani: create an object `styles` and then keep adding functions there so you have the context tied to it.

```js
const styles = {
	size: (size) => {
		switch()
		...
	}
}
```

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,71 @@
...
+
+  const onPressIn = useCallback(() => {
+    setActive(true);
+  }, [setActive]);
+  const onPressOut = useCallback(() => {
+    setActive(false);
+  }, [setActive]);
+
+  const inactiveTextColor = visited ? theme.colors.secondary[900] : theme.colors.primary[800];
+  const activeTextColor = visited ? theme.colors.secondary[600] : theme.colors.primary[600];
```
kamleshchandnani: move them before funcitons.

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <StyledTouchable
+      activeOpacity={1}
+      onPress={onClick}
+      onPressIn={onPressIn}
+      onPressOut={onPressOut}
+      testID="ds-link-button"
+    >
```
shubhnik: We will be passing `testId` from the app. That `testId` will be used in the e2e tests. We need to pass that `testId` to the root native element.

src/atoms/Link/__tests__/Link.native.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+  it('snapshot testing, renders large link', () => {
+    const displayText = 'Displaying some link';
+    const { container } = renderWithTheme(<Link size="l">{displayText}</Link>);
+    expect(container).toMatchSnapshot();
+  });
+
+  it('onPressIn/onPressOut link styles look as expected', () => {
+    const { getByTestId, container } = renderWithTheme(<Link>{'Click Me'}</Link>);
+    const button = getByTestId('ds-link-button');
```
shubhnik: We can try this instead of `testId`
```
const button = getByText(/click me/i).parentNode;
fireEvent.press(button);
```

src/atoms/Link/Link.native.js
deorukhkarchaitanya: https://github.com/styled-components/styled-components/issues/2536

Running into this after trying that. The workaround mentioned doesn't seem clean


Would it be okay if we did this?

**.babelrc.js:**
```
plugins: [
        [
          'module-resolver',
          {
            alias: {
              '@storybook/react': '@storybook/react-native',
              'styled-components': 'styled-components/native',
            },
          },
        ],
      ],
```

src/atoms/Link/Link.native.js
deorukhkarchaitanya: I will have to do something like this
```
      <Text
        underline={active}
        color={active ? activeTextColor : inactiveTextColor}
        size={styles.size({ size })}
      >
        {children}
      </Text>
```

```
const styles = {
  size: ({ size }) => {
    switch (size) {
      case 'small':
      default:
        return 'xsmall';
      case 'medium':
        return 'medium';
      case 'large':
        return 'large';
    }
  },
};
```

since we're resolving the size and then passing it down to another Text component



Is that okay?

deorukhkarchaitanya: Another way is this

```
const linkToTextSize = {
  small: 'xsmall',
  medium: 'medium',
  large: 'large',
};
```

```
size={linkToTextSize[size]}
```

src/atoms/Link/__tests__/Link.native.test.js
deorukhkarchaitanya: Makes sense. Done.

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,64 @@
...
+  small: 'xsmall',
+  medium: 'medium',
+  large: 'large',
+};
+
+const Link = ({ children, visited, onClick, size, testID }) => {
+  const theme = useContext(ThemeContext);
+  const [active, setActive] = useState(false);
+  const inactiveTextColor = visited ? theme.colors.primary[800] : theme.colors.primary[800];
+  const activeTextColor = visited ? theme.colors.primary[600] : theme.colors.primary[600];
```
kamleshchandnani: I didn't understand this? Irrespective of the `visited` prop we are rendering `theme.colors.primary[600]` 🤔

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,64 @@
+import React, { useContext, useState, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import Text from '../Text';
+import { ThemeContext } from 'styled-components/native';
+import automation from '../../_helpers/automation-attributes';
+import Flex from '../Flex/Flex.native';
+import { TouchableOpacity } from 'react-native';
+
+const linkToTextSize = {
+  small: 'xsmall',
```
kamleshchandnani: let's maintain our standards and semantics of small, xsmall etc. Screw design. It's too confusing and annoying to write mappings in each and every file

src/atoms/Link/Link.native.js
deorukhkarchaitanya: Because the design was changed to not have a different color for visited links but as per our discussion in standup, we wanted to leave the code untouched so that it can be accomodated later

src/atoms/Link/Link.native.js
deorukhkarchaitanya: The link design has font sizes of 12,14 & 16 which is xsmall, medium & large. They don't have small in link

src/atoms/Link/Link.native.js
deorukhkarchaitanya: @kamleshchandnani need feedback on this

src/atoms/Link/Link.native.js
deorukhkarchaitanya: Removed visited now

src/atoms/Link/Link.native.js
kamleshchandnani: looks good to me. But might break on consumers if they don't add the same alias. Let it be explicit

deorukhkarchaitanya: Hmm yeah

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,54 @@
+import React, { useState, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import Text from '../Text';
+import automation from '../../_helpers/automation-attributes';
+import Flex from '../Flex/Flex.native';
+import { TouchableOpacity } from 'react-native';
```
dhruvdutt: Move above relative imports.

src/atoms/Link/Link.stories.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import Link from './Link';
+import { text, select } from '@storybook/addon-knobs';
+import { withKnobs } from '@storybook/addon-ondevice-knobs';
```
dhruvdutt: Move above relative imports.

src/atoms/Link/__tests__/Link.native.test.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import Link from '../index';
+import { renderWithTheme } from '../../../utils/testing';
+import { fireEvent } from '@testing-library/react-native';
```
dhruvdutt: Move above relative imports.

src/atoms/Link/Link.native.js
```diff
@@ -0,0 +1,54 @@
...
+
+const Link = ({ children, onClick, size, testID }) => {
+  const [active, setActive] = useState(false);
```
dhruvdutt: Isn't there any hover state for `Link`?

src/atoms/Flex/Flex.native.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+const Flex = styled(
```
dhruvdutt: Include `Flex` in the PR title.

src/atoms/Link/Link.native.js
kamleshchandnani: in mobile there's no hover 😛

PR #18, opened 2020-02-14
feat(icon): add Icon component [ME-4248]

deorukhkarchaitanya:


src/atoms/Icon/Icon.native.js
```diff
@@ -0,0 +1,22 @@
...
+
+const Icon = ({ size, name, fill }) => {
+  const IconComponent = icons[name];
+  return <IconComponent fill={fill} height={iconSize[size]} width={iconSize[size]} />;
+};
+
+Icon.propTypes = {
+  size: PropTypes.oneOf(['s', 'm', 'l']),
+  name: PropTypes.oneOf(Object.keys(icons)),
+  fill: PropTypes.string,
```
deorukhkarchaitanya: I'll change ```PropTypes.string``` to whatever way we decide in the Text PR

src/icons/index.js
```diff
@@ -0,0 +1,7 @@
+import Info from './Info';
+
+const icons = {
+  info: Info,
```
deorukhkarchaitanya: We can rather do this ```  info: require('./Info').default```

deorukhkarchaitanya: LMK which is the better approach

src/atoms/Icon/Icon.native.js
```diff
@@ -0,0 +1,22 @@
...
+  l: 20,
+};
+
+const Icon = ({ size, name, fill }) => {
+  const IconComponent = icons[name];
+  return <IconComponent fill={fill} height={iconSize[size]} width={iconSize[size]} />;
+};
+
+Icon.propTypes = {
+  size: PropTypes.oneOf(['s', 'm', 'l']),
```
shubhnik: Shouldn't size be `small`, `medium` and `large` for Component props?

deorukhkarchaitanya: We're following ```s, m, l``` pattern for font sizes. Wouldn't this be consistent with that?

shubhnik: For internal usage, i.e tokens `s, m, l` is fine. I am talking about the API being exposed, a developer writing `<Icon size='small' />` seems more obvious and understandable than `<Icon size='s' />`. What do you say?

deorukhkarchaitanya: Text is not only internal usage. We're using s,m,l there too. 

I agree that small, medium and large seems better. But then we have to keep it consistent across the DS.

shubhnik: Yeah for `Text` also. Across all the react components of our DS.

src/atoms/Icon/Icon.native.js
```diff
@@ -0,0 +1,26 @@
...
+  l: 20,
+};
+
+const Icon = ({ size, name, fill }) => {
+  const IconComponent = icons[name];
+  return <IconComponent fill={fill} height={iconSize[size]} width={iconSize[size]} />;
+};
+
+Icon.defaultProps = {
+  size: 'l',
```
shubhnik: Should we settle for a common default prop across components? For e.g, the default size is large here in Icon but in Link it is medium.

deorukhkarchaitanya: Sure, makes sense.

src/atoms/Icon/Icon.native.js
```diff
@@ -0,0 +1,26 @@
...
+
+const iconSize = {
+  s: 10,
+  m: 16,
+  l: 20,
+};
+
+const Icon = ({ size, name, fill }) => {
+  const IconComponent = icons[name];
+  return <IconComponent fill={fill} height={iconSize[size]} width={iconSize[size]} />;
```
shridhar52war: Should we expose other SVG style properties like `strokeWidth`, `fillOpacity?

src/atoms/Icon/Icon.native.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import icons from '../../icons';
+
+const iconSize = {
```
shridhar52war: Replace this with 
```
const iconSize = {
  l: 24,
  m: 20,
  s: 16,
  xs: 12
};
```

src/atoms/Icon/Icon.native.js
deorukhkarchaitanya: I'm thinking we should do `{...props}` so that we don't need to manually add support for all svg props

src/icons/index.js
kamleshchandnani: let's stick with this. Bundlers understand and are able to treeshake this better

src/atoms/Icon/Icon.native.js
kamleshchandnani: Let's rename the fill as `color`? It remains consistent with all our namings for color as well.

src/icons/index.js
deorukhkarchaitanya: https://facebook.github.io/react-native/docs/performance#ram-bundles--inline-requires

Check this out. If the list of icon grows, importing this icon file could be expensive since it imports all the icon components at the top level. If we do inline requires, it won't import everything.

src/atoms/Icon/Icon.native.js
deorukhkarchaitanya: There might be more props here that we may discover later like fillOpacity, stroke, etc. These are all SVG props and we should ideally keep them the way they are going to be interpreted in the svg

storybook/native/main.js
```diff
@@ -14,6 +14,7 @@ const theme = {
 
 // import stories
 configure(() => {
+  require('../../src/atoms/Icon/Icon.stories');
```
kamleshchandnani: Can't you add a regex here? for all `**.stories.js`

src/atoms/Icon/Icon.native.js
kamleshchandnani: that's again flexibility over constraints 😆. Cool let's come back to this once we have the entire icon component

src/icons/index.js
kamleshchandnani: This is totally opposite of how other bundlers work 🙈 sure then let's do inline

storybook/native/main.js
deorukhkarchaitanya: No :( 

We explored some other options like plugins to do this but nothing was good enough. We need to manually add each story until storybook adds proper configuration support like web

PR #25, opened 2020-02-20
feat(text): add support for line height

deorukhkarchaitanya:


src/_helpers/fonts.js
```diff
@@ -1,5 +1,14 @@
...
-  return parseFloat(theme.fonts.lineHeight[lineHeight]) * parseFloat(theme.fonts.size[size]);
+const getDownwardEvenNumber = (val) => {
+  if (val % 2 === 0) {
+    return `${val}px`;
+  } else {
+    return getDownwardEvenNumber(val - 1);
+  }
+};
+
+const getLineHeight = (fontSize, lineHeightMultiplier) => {
```
kamleshchandnani: So we have a 1.4 multiplier. we can get rid of all this 😆

deorukhkarchaitanya: I can't see those changes on figma. 1.4 multiplier for everything?

src/_helpers/fonts.js
```diff
@@ -0,0 +1,17 @@
+const getLineHeightFromTextSize = (textSize, theme) => {
```
kamleshchandnani: rename to `getLineHeight`
Also rearrange the arguments in colors we have theme first and then the color let's follow the same here

deorukhkarchaitanya: done

PR #28, opened 2020-02-20
feat(text): add numberOfLines prop

deorukhkarchaitanya:


kamleshchandnani:
Why do we need `numberOfLines` prop?

deorukhkarchaitanya:
> @deorukhkarchaitanya Why do we need `numberOfLines` prop?

For immediate usage, we're using numberOfLines in the TextInput Component for HelperText. I thought of making this a private prop but we might need to trim our Text in a lot of places on the app. This is one of the most frequently used props of Text so I made it public

kamleshchandnani:
we might need to come up with an API. Since there's nothing like `numberOfLines` in web. we might also need `ellipsizeMode` from react-native. So the idea should be that we come up with a common API. Also in helper text did we ask designers what is the desired behavior when we have multiline helper text in text input? This also applies to `checkbox` and `radio`

deorukhkarchaitanya:
> we might need to come up with an API. Since there's nothing like `numberOfLines` in web. we might also need `ellipsizeMode` from react-native. So the idea should be that we come up with a common API. 

`ellipsizeMode` needs `numberOfLines` check [here](https://reactnative.dev/docs/text#ellipsizemode)
`ellipsizeMode` only defines how trimming should look like
Eg) **Hell...orld** or **Hello Wo...** or .**..World**
Its's `numberOfLines` that dictates when to trim


> Also in helper text did we ask designers what is the desired behavior when we have multiline helper text in text input? This also applies to `checkbox` and `radio`

Yes, helper is max 2 lines

kamleshchandnani:
Yea I had read that that's how landed on `ellipsizeMode` 😄 . So the proposal I'm trying to make is that we can just expose one prop i.e `isTruncated` and do all the stuff as part of implementation details. Not sure yet need to think of usecases

PR #29, opened 2020-02-20
build(version): bump app version to 1.0.2

deorukhkarchaitanya:


ios/blade/Info.plist
```diff
@@ -17,11 +17,11 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.0</string>
+	<string>1.0.3</string>
```
dhruvdutt: Mismatch between gradle version and this? 🤔

deorukhkarchaitanya: Should be 1.0.2

ios/blade/Info.plist
```diff
@@ -40,7 +40,13 @@
...
 	</dict>
 	<key>NSLocationWhenInUseUsageDescription</key>
-	<string/>
+	<string></string>
+	<key>UIAppFonts</key>
+	<array>
+		<string>Lato-Bold.ttf</string>
+		<string>Lato-Light.ttf</string>
+		<string>Lato-Regular.ttf</string>
+	</array>
```
kamleshchandnani: Do the consumers need to add this as well when they use our DS or when they `link` it will be setup automatically?

deorukhkarchaitanya: This will be set up automatically. This diff is also XCode rearranging it.

PR #33, opened 2020-02-25
refactor(colors): new color palletes and aliases for theme

shubhnik:
BREAKING CHANGE: recheck and fix if new color scheme breaks your code


shubhnik:
Holding merge, the design team is fixing background color aliases for light/dark theme.

PR #35, opened 2020-02-26
refactor(colors): new color palletes and aliases for theme

shubhnik:
BREAKING CHANGE: recheck and fix if new color scheme breaks your code


src/tokens/theme.web.js
```diff
@@ -97,25 +118,44 @@ const darkTheme = {
...
+      960: colors.green[960],
     },
     neutral: {
-      300: colors.honey[300],
-      600: colors.honey[600],
       900: colors.honey[900],
+      930: colors.honey[930],
+      960: colors.honey[960],
+    },
+    exceptions: {
```
kamleshchandnani: It will break. We don't have this structure. All you need to do is have this as top level objects don't nest it under `exception`

PR #41, opened 2020-02-27
feat(Tab): Add tab component

shridhar52war:


src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,133 @@
...
+
+  const [tabBackground, setTabBackground] = useState('transparent');
+  const [labelColor, setLabelColor] = useState(initialLabelColor);
+
+  const theme = useContext(ThemeContext);
+
+  const onPressIn = () => {
+    if (isActive) {
+      setLabelColor('primary.900');
+      setTabBackground(getColor(theme, 'primary.200'));
```
deorukhkarchaitanya: Can't we set colors through the component using only it's name like 
```suggestion
      setTabBackground('primary.200');
```

so that you don't need to get ```theme``` from context. You can resolve this color in styles above

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,133 @@
...
+      return;
```
deorukhkarchaitanya: Why not ```if...else```? Isn't that more straight forward?

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,133 @@
...
+
+const Tab = ({ label, isActive, icon, onPress, disabled, testID }) => {
+  const initialLabelColor = getLabelColor({ isActive, disabled });
+
+  const [tabBackground, setTabBackground] = useState('transparent');
+  const [labelColor, setLabelColor] = useState(initialLabelColor);
+
+  const theme = useContext(ThemeContext);
+
+  const onPressIn = () => {
```
deorukhkarchaitanya: Wrap this around a ```useCallback()```

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,133 @@
...
+    if (isActive) {
+      setLabelColor('primary.900');
+      setTabBackground(getColor(theme, 'primary.200'));
+      return;
+    }
+    setLabelColor('shade.970');
+    setTabBackground(getColor(theme, 'tone.940'));
+  };
+
+  const onPressOut = () => {
```
deorukhkarchaitanya: Wrap this around a ```useCallback()```

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,133 @@
...
+    const color = getLabelColor({ isActive, disabled });
+    setLabelColor(color);
+    setTabBackground('transparent');
+  };
+
+  const isValidIcon = icons[icon];
+  return (
```
deorukhkarchaitanya: New line before return please?

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,133 @@
...
+    setTabBackground(getColor(theme, 'tone.940'));
+  };
+
+  const onPressOut = () => {
+    const color = getLabelColor({ isActive, disabled });
+    setLabelColor(color);
+    setTabBackground('transparent');
+  };
+
+  const isValidIcon = icons[icon];
```
deorukhkarchaitanya: We probably don't need this check. If it's not a valid icon, let it crash instead of hiding the icon

src/atoms/Tab/Tab.stories.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withKnobs } from '@storybook/addon-ondevice-knobs';
```
deorukhkarchaitanya: Don't use this. We have declared this in ```native/main.js```

src/atoms/Tab/Tab.stories.js
```diff
@@ -0,0 +1,36 @@
...
+import Tab from './Tab';
+import Flex from '../Flex';
+import View from '../View';
+
+storiesOf('Tabs', module)
+  .addDecorator(withKnobs)
```
deorukhkarchaitanya: Don't use this. We have declared this in ```native/main.js```

src/atoms/Tab/Tab.stories.js
```diff
@@ -0,0 +1,36 @@
...
+  .addParameters({
+    component: Tab,
+  })
+  .add('show tabs', () => (
```
deorukhkarchaitanya: Let's keep app sub categories of stories as default
```suggestion
  .add('default', () => (
```

src/atoms/Tab/Tab.native.js
shridhar52war: Yeah, wanted to have a discussion on this! Is is not good to render nothing for invalid icon rather than forcing a crash?

src/atoms/Tab/Tab.native.js
shridhar52war: Both are same no?

src/atoms/Tab/Tab.native.js
shridhar52war: I have to change the background color on `pressIn` events, how do I connect the styles object with the events that are fired by the component?

src/atoms/Tab/Tab.native.js
deorukhkarchaitanya: Yeah but ```if...else``` is a little more verbose and clear to understand. Keeps it open to accommodate more ```if...else if...else``` cases

src/atoms/Tab/Tab.native.js
deorukhkarchaitanya: It's better to crash and let the implementor know something is wrong rather than hiding it and letting the implementor debug what went wrong

src/atoms/Tab/Tab.native.js
deorukhkarchaitanya: You can have a ```isPressed``` state that is used to derive colors of Label & Tab

src/atoms/Tab/Tab.native.js
shridhar52war: This was wrapped around `useCallback`, I don't why the hell did I remove this?! Shoot me!

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,134 @@
...
+
+  const onPressIn = useCallback(() => {
+    setTabPressed(true);
+  }, []);
+
+  const onPressOut = useCallback(() => {
+    setTabPressed(false);
+  }, []);
+
+  const isValidIcon = icons[icon];
```
deorukhkarchaitanya: Can we remove this check? We should let it crash, right?

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,134 @@
...
+    </Flex>
+  );
+};
+
+Tab.propTypes = {
+  isActive: PropTypes.bool,
+  onPress: PropTypes.func,
+  disabled: PropTypes.bool,
+  label: PropTypes.string.isRequired,
+  icon: PropTypes.oneOf(Object.keys(icons)),
```
deorukhkarchaitanya: Should we have a helper function for this? Sorry, I should've created it with Icon.

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,132 @@
...
+
+const StyledTabButton = styled(TouchableOpacity)`
+  min-width: 48px;
+  background-color: ${styles.tab.backgroundColor};
+  border-top-left-radius: 2px;
+  border-top-right-radius: 2px;
+`;
+
+const Label = styled(View)`
+  padding: 8px 0px;
```
deorukhkarchaitanya: Use ```Space```

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,132 @@
...
+  disabled: PropTypes.bool,
+  label: PropTypes.string.isRequired,
+  icon: PropTypes.oneOf(getIconKeys()),
+  testID: PropTypes.string,
+};
+
+Tab.defaultProps = {
+  isActive: false,
+  onPress: () => {},
+  disabled: false,
```
deorukhkarchaitanya: default prop for icon?

src/atoms/Tab/__tests__/Tab.native.test.js
```diff
@@ -0,0 +1,188 @@
...
+
+    test('should match snapshot when disabled, with icon present', () => {
+      const { container } = renderWithTheme(
+        <Tab label="Active state" icon="info" disabled={true} isActive={true} />,
+      );
+      expect(container).toMatchSnapshot();
+    });
+
+    test('should throw error when invalid icon name is passed', () => {
+      // const { container } = renderWithTheme(
```
deorukhkarchaitanya: Remove comments

src/atoms/Tab/Tab.native.js
shridhar52war: added

src/atoms/Tab/__tests__/Tab.native.test.js
shridhar52war: Removed

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,137 @@
...
+        }
+        return getColor(theme, 'tone.940');
+      }
+      return 'transparent';
+    },
+  },
+};
+
+const StyledTabButton = styled(TouchableOpacity)`
+  min-width: 48px;
```
dhruvdutt:
```suggestion
  makePxValue(6);
```
Use pixel helper function.

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,137 @@
...
+      }
+      return 'transparent';
+    },
+  },
+};
+
+const StyledTabButton = styled(TouchableOpacity)`
+  min-width: ${makePxValue(6)};
+  background-color: ${styles.tab.backgroundColor};
+  border-top-left-radius: 2px;
```
shubhnik: `props.theme.spacings.xxsmall` just for consistency across components.

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,137 @@
...
+import Icon from '../Icon';
+import Space from '../Space';
+import automation from '../../_helpers/automation-attributes';
+import isEmpty from '../../_helpers/isEmpty';
+
+const styles = {
+  label: {
+    borderBottomWidth({ isActive }) {
+      if (isActive) {
+        return '1px';
```
shubhnik: can we use `makePxValue` helper here?

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,137 @@
...
+  border-top-right-radius: 2px;
+`;
+
+const Label = styled(View)`
+  border-bottom-width: ${styles.label.borderBottomWidth};
+  border-bottom-color: ${styles.label.borderBottomColor};
+`;
+
+const Tab = ({ label, isActive, icon, onPress, disabled, testID }) => {
+  const [isPressed, setTabPressed] = useState(false);
```
shubhnik: `const [pressed, setPressed] = useState(false)`, since it is used in context of Tab component, we can avoid explicitly telling **tab**Pressed? 

And keeping consistency across states that define visuals of component e.g `pressed`, `disabled`, `active`.

src/atoms/Tab/Tab.native.js
```diff
@@ -0,0 +1,137 @@
...
+  border-top-left-radius: 2px;
+  border-top-right-radius: 2px;
+`;
+
+const Label = styled(View)`
+  border-bottom-width: ${styles.label.borderBottomWidth};
+  border-bottom-color: ${styles.label.borderBottomColor};
+`;
+
+const Tab = ({ label, isActive, icon, onPress, disabled, testID }) => {
```
shubhnik: `active` instead of `isActive`? Just like it's `disabled` and not `isDisabled`.

src/atoms/Tab/Tab.native.js
shridhar52war: Not sure. We can have discussion on this.

chaitanyadeorukhkar:
Closing in favour of https://github.com/razorpay/blade/pull/114

PR #42, opened 2020-03-02
style(main.js): give parent container of storybook 100% height & width

deorukhkarchaitanya:


storybook/native/main.js
```diff
@@ -22,7 +22,9 @@ configure(() => {
 
 // add decorators
 const SpaceAround = styled.View`
-  margin: 20px;
+  padding: 20px;
```
dhruvdutt: Can use `Space` component for this.

storybook/native/main.js
```diff
@@ -22,7 +22,9 @@ configure(() => {
...
+  height: 100%;
+  width: 100%;
```
dhruvdutt: We need to implement `Size` component for these use-cases. @kamleshchandnani

storybook/native/main.js
deorukhkarchaitanya: Avoiding the usage of design system components here. This should be agnostic of what design system follows since it's just the story.

deorukhkarchaitanya: My story container should not break if DS components change. Stories don't have test that can protect them from DS changes

storybook/native/main.js
kamleshchandnani: Yes already on roadmap sir for next sprint

PR #46, opened 2020-03-03
feat(View): add view component web

susheelachoudry:


dhruvdutt:
Can we stop doing Bitrise checks based on a file pattern or something? // @kamleshchandnani

src/atoms/View/View.web.js
```diff
@@ -1,3 +1,20 @@
...
+  ${''}
+`;
+
+View.propTypes = {
+  component: PropTypes.string.isRequired,
+  children: PropTypes.node,
+};
+
+View.defaultProps = {
+  component: 'div',
```
dhruvdutt:
```suggestion
  as: 'div',
```

Let's call this `as`

src/atoms/View/View.stories.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import View from './View';
+
+storiesOf('View', module)
+  .addParameters({
+    component: View,
+  })
+  .add('with text', () => <View>Simple Div</View>);
```
deorukhkarchaitanya:
```suggestion
  .add('with text', () => <View><Text>Simple View</Text></View>);
```
All text must be inside ```Text``` component. We need to be careful when we write stories since they are platform agnostic. This would break on native since ```View``` cannot contain text

src/atoms/View/__tests__/View.web.test.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from '@testing-library/react';
```
deorukhkarchaitanya: You should go ahead and create a helper function called ```renderWithTheme``` you can refer what we've done for native since you'll need all your stories to have the ```ThemeContext``` from ```styled-components```

src/atoms/View/View.stories.js
susheelachoudry: text web component is yet to be created.

src/atoms/View/__tests__/View.web.test.js
susheelachoudry: done

src/atoms/View/View.stories.js
kamleshchandnani: @susheelachoudry But the stories are common for react and react-native and hence they'll break. So create the text component as well before we merge this or raise a new PR to this branch for Text

src/_helpers/testing-web.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
kamleshchandnani: the filename should be `testing.web.js`

src/atoms/View/__tests__/View.web.test.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { renderWithTheme } from '../../../_helpers/testing-web';
+
+import View from '../index';
+
+describe('Renders <View /> correctly', () => {
```
kamleshchandnani: `Renders <View />` no need of adding correctly since this is a test suite and not a test.

src/atoms/View/__tests__/View.web.test.js
```diff
@@ -0,0 +1,11 @@
...
+  it('snapshot testing', () => {
```
kamleshchandnani: `it('renders with text', () => {`

src/_helpers/testing-web.js
susheelachoudry: ok

susheelachoudry:
I will be updating this PR once text component gets merged to master.

src/atoms/View/View.web.js
```diff
@@ -1,3 +1,18 @@
-const View = () => null;
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+const View = styled(({ as, children, ...props }) => React.createElement(as, props, children))`
+  ${''}
+`;
```
deorukhkarchaitanya:
```suggestion
const View = styled.div``;
```

src/atoms/View/View.web.js
```diff
@@ -1,3 +1,18 @@
...
+
+View.propTypes = {
+  as: PropTypes.string.isRequired,
```
deorukhkarchaitanya:
```suggestion
  as: PropTypes.oneOf(['div', 'p', 'span']).isRequired,
```

add more values to it if you want but accept only specific values that you want to allow

src/atoms/View/View.web.js
susheelachoudry: thought of, but what about children ?

src/atoms/View/View.web.js
```diff
@@ -1,3 +1,18 @@
...
+  children: PropTypes.node,
+};
+
+View.defaultProps = {
+  as: 'div',
```
deorukhkarchaitanya: We should specify defaultProps for all all non isRequired props. In this case ```children: undefined```

src/atoms/View/View.web.js
kamleshchandnani: @deorukhkarchaitanya let it be string. We'll restrict if we see the issue because on web it has lot of nuances and then the lead time to get the PR merged is hugeee.

@susheelachoudry Remove the `isRequired` though

src/atoms/View/__tests__/View.web.test.js
```diff
@@ -0,0 +1,11 @@
...
+import { renderWithTheme } from '../../../_helpers/testing';
+
+import View from '../index';
+
+describe('Renders with text ', () => {
+  it('snapshot testing', () => {
+    const { container } = renderWithTheme(<View>{'simple div'}</View>);
+    expect(container).toMatchSnapshot();
+  });
+});
```
deorukhkarchaitanya:
```suggestion
describe('<View />', () => {
  describe('as', () => {
    it('renders View as div', () => {
      const { container } = renderWithTheme(
        <View as="div">
          <Text>{'simple div'}</Text>
        </View>,
      );
      expect(container).toMatchSnapshot();
    });
    it('renders View as span', () => {
      const { container } = renderWithTheme(
        <View as="span">
          <Text>{'simple div'}</Text>
        </View>,
      );
      expect(container).toMatchSnapshot();
    });
    it('renders View as p', () => {
      const { container } = renderWithTheme(
        <View as="p">
          <Text>{'simple div'}</Text>
        </View>,
      );
      expect(container).toMatchSnapshot();
    });
  });
});
```

This is how we try to write out tests. I'll send more details on slack. For now you can just use this suggestion for the test.

src/atoms/View/__tests__/View.web.test.js
kamleshchandnani: change this text as well

src/atoms/View/View.web.js
deorukhkarchaitanya: It'll behave like normal component

PR #51, opened 2020-03-04
feat(size): create Size component

deorukhkarchaitanya:


src/atoms/Size/Size.native.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components/native';
+
+const Size = styled(
+  ({ minWidth, width, maxWidth, minHeight, height, maxHeight, children, ...props }) =>
+    React.cloneElement(children, props),
```
shridhar52war: Add a check to accept only single child.
```
if (React.Children.toArray(children).length !== 1) {
    throw new Error('Expected a single child for Size component');
 }
return React.cloneElement(children, props);
```

src/atoms/Size/__tests__/Size.native.test.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { render } from '@testing-library/react-native';
+import { View } from 'react-native';
+import Size from '../index';
+
+describe('Renders <Size /> correctly', () => {
+  it('snapshot testing', () => {
+    const { container } = render(
```
shridhar52war: Test case to accept more than one child

PR #54, opened 2020-03-04
feat(text): implementation of text component

susheelachoudry:


src/atoms/Text/__tests__/Text.web.test.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { renderWithTheme } from '../../../_helpers/testing';
+
+import Text from '../index';
+
+describe('Renders Text', () => {
+  it('snapshot testing', () => {
```
kamleshchandnani: where are different variants test? Refer native tests and update it

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,92 @@
...
+  line-height: ${styles.lineHeight};
+`;
+
+Text.propTypes = {
+  as: PropTypes.string,
+  size: PropTypes.oneOf(['xxsmall', 'xsmall', 'small', 'medium', 'large']),
+  color: PropTypes.oneOf(getColorKeys()),
+  align: PropTypes.oneOf(['left', 'center', 'right']),
+  testID: PropTypes.string,
+  numberOfLines: PropTypes.number,
```
kamleshchandnani: numberOfLines doesn't exist in web. We need to implement a prop called truncated which will implement ellipsis. Discuss with @deorukhkarchaitanya for more clarity on API

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,92 @@
...
+  color({ theme, color }) {
+    return getColor(theme, color);
+  },
+  textDecorationLine({ _isUnderlined }) {
+    if (_isUnderlined) return 'underline';
+    else {
+      return 'none';
+    }
+  },
+  alignSelf({ align }) {
```
kamleshchandnani: rename to `align`

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,92 @@
...
+  testID: PropTypes.string,
+  numberOfLines: PropTypes.number,
+  _weight: PropTypes.oneOf(Object.keys(baseTheme.fonts.weight)),
+  _isUnderlined: PropTypes.bool,
+  _letterSpacing: PropTypes.oneOf(Object.keys(baseTheme.fonts.letterSpacing)),
+  _lineHeight: PropTypes.oneOf(Object.keys(baseTheme.fonts.lineHeight)),
+};
+
+Text.defaultProps = {
+  as: 'div',
```
dhruvdutt: Add `children` here.

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,92 @@
...
+  font-size: ${styles.fontSize};
+  color: ${styles.color};
+  text-decoration-line: ${styles.textDecorationLine};
+  align-self: ${styles.alignSelf};
+  letter-spacing: ${styles.letterSpacing};
+  line-height: ${styles.lineHeight};
+`;
+
+Text.propTypes = {
+  as: PropTypes.string,
```
dhruvdutt: Add `children` here.

dhruvdutt:
```suggestion
  as: PropTypes.node,
```

src/atoms/Text/Text.web.js
dhruvdutt: Implementation for `numberOfLines` is pending.

dhruvdutt: [Refer here for implementation](https://github.com/treebohotels/leaf-ui/blob/master/src/Text/Text.web.js#L25-L27).

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,92 @@
...
+    return theme.fonts.family.lato[_weight];
+  },
+  fontSize({ theme, size }) {
+    return theme.fonts.size[size];
+  },
+  color({ theme, color }) {
+    return getColor(theme, color);
+  },
+  textDecorationLine({ _isUnderlined }) {
+    if (_isUnderlined) return 'underline';
```
dhruvdutt:
```js
if (_isUnderlined) {
    return 'underline';
}
return 'none';
```

src/atoms/Text/Text.web.js
susheelachoudry: why ? children

kamleshchandnani: it would be string only
`<Text as="div"/>` and not <Text as={<div></div>}/>`

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,92 @@
...
+    as,
+    fontFamily,
+    fontSize,
+    children,
+    textDecorationLine,
+    alignSelf,
+    letterSpacing,
+    lineHeight,
+    ...props
+  }) => React.createElement(as, props, children),
```
deorukhkarchaitanya:
```suggestion
    style,
    ...props
  }) => React.createElement(as, props, children),
```

Just so that ```style``` is extracted out and no one can override the style?

src/atoms/Text/__tests__/Text.web.test.js
deorukhkarchaitanya: We'll have a discussion today on testing

storybook/web/preview.js
```diff
@@ -2,8 +2,7 @@ import React from 'react';
 import { ThemeProvider } from 'styled-components';
 import { addDecorator } from '@storybook/react';
 import { withKnobs } from '@storybook/addon-knobs/react';
-
-const theme = {};
+import theme from '../../src/tokens/theme';
```
deorukhkarchaitanya: This will restrict us to only light theme. We need ThemeSwitcher. Maybe not in this PR but we should take it up quickly

src/atoms/Text/__tests__/Text.web.test.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
dhruvdutt: Not needed. 😅

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,117 @@
...
+  as: PropTypes.string,
+  size: PropTypes.oneOf(['xxsmall', 'xsmall', 'small', 'medium', 'large']),
+  color: PropTypes.oneOf(getColorKeys()),
+  align: PropTypes.oneOf(['left', 'center', 'right']),
+  'data-testid': PropTypes.string,
+  _weight: PropTypes.oneOf(Object.keys(baseTheme.fonts.weight)),
+  _isUnderlined: PropTypes.bool,
+  _letterSpacing: PropTypes.oneOf(Object.keys(baseTheme.fonts.letterSpacing)),
+  _lineHeight: PropTypes.oneOf(Object.keys(baseTheme.fonts.lineHeight)),
+  truncate: PropTypes.bool,
```
dhruvdutt: Put this above internal props.

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,117 @@
...
+};
+
+Text.defaultProps = {
+  as: 'div',
+  size: 'large',
+  align: 'left',
+  color: 'shade.980',
+  'data-testid': 'ds-text',
```
dhruvdutt: Add default value for `truncate`.

src/atoms/Text/Text.web.js
deorukhkarchaitanya: What is this prop & why are we deviating from a common API for web & mobile? :(

dhruvdutt: Oh, yes. This is `numberOfLines` in native. 😀

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
deorukhkarchaitanya: @susheelachoudry can we use a common API for web & mobile here?

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,122 @@
...
+    fontSize,
+    children,
+    textDecorationLine,
+    align,
+    letterSpacing,
+    lineHeight,
+    style,
+    ...props
+  }) =>
+    React.createElement(as, props, children, {
```
deorukhkarchaitanya: ```as``` is a styled-components prop which when passed will use it and your code will never reach this createElement line.

https://styled-components.com/docs/api#as-polymorphic-prop

You can just do 
```
const Text = styled.div`
```

and everything will work

src/atoms/Text/Text.web.js
susheelachoudry: I am not sure that we numberOfLines in web. @deorukhkarchaitanya

src/atoms/Text/Text.web.js
susheelachoudry: wont it restrict to be only div ?

deorukhkarchaitanya: It wont. It will be ```div``` if no ```as``` is passed. If you pass ```as```  then it will take that as the component

src/atoms/Text/Text.web.js
deorukhkarchaitanya: And native doesn't have truncate. We need to decide on a common API that maps to truncate on web & numberOfLines on native

deorukhkarchaitanya: ```numberOfLines``` will restrict the text to the mentioned number of lines. Can we replicate that behaviour on web?

susheelachoudry: we cant replicate numberOfLines, as far as i know. but we can achieve via giving height. @deorukhkarchaitanya

src/atoms/Text/Text.web.js
susheelachoudry: Done

src/atoms/Text/Text.web.js
deorukhkarchaitanya: Seems tricky. Continuing this conversation on slack.

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,103 @@
...
+
+const Text = styled.div`
+  font-family: ${styles.fontFamily};
+  font-size: ${styles.fontSize};
+  color: ${styles.color};
+  text-decoration-line: ${styles.textDecorationLine};
+  align-self: ${styles.align};
+  letter-spacing: ${styles.letterSpacing};
+  line-height: ${styles.lineHeight};
+  overflow: ${styles.overflow};
```
kamleshchandnani: For single ternary conditions let's stop avoiding use of styles. it seems like overdoing 
```suggestion
  overflow: ${(props) => props.truncate ? 'hidden' : ''};
```

storybook/web/preview.js
susheelachoudry: Sure

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,97 @@
...
+Text.propTypes = {
+  size: PropTypes.oneOf(['xxsmall', 'xsmall', 'small', 'medium', 'large']),
+  color: PropTypes.oneOf(getColorKeys()),
+  align: PropTypes.oneOf(['left', 'center', 'right']),
+  'data-testid': PropTypes.string,
+  _weight: PropTypes.oneOf(Object.keys(baseTheme.fonts.weight)),
+  _isUnderlined: PropTypes.bool,
+  _letterSpacing: PropTypes.oneOf(Object.keys(baseTheme.fonts.letterSpacing)),
+  _lineHeight: PropTypes.oneOf(Object.keys(baseTheme.fonts.lineHeight)),
+  truncate: PropTypes.bool,
```
kamleshchandnani: We don't need `truncate` now. That's the whole point of bringing in `maxLines`

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,97 @@
...
+  lineHeight({ theme, size, _lineHeight }) {
+    if (_lineHeight) {
+      return theme.fonts.lineHeight[_lineHeight];
+    } else {
+      return getLineHeight(theme, size);
+    }
+  },
+  maxHeight({ theme, truncate, _lineHeight, maxLines }) {
+    if (truncate) {
+      const lineHeight = `${theme.fonts.lineHeight[_lineHeight]}`.replace('px', '');
```
kamleshchandnani: should be `parseFloat(theme.fonts.lineHeight[_lineHeight])`

src/atoms/Text/Text.web.js
susheelachoudry: what should be default value for maxLines then?  i was using truncate as switch to enable disable maxlines @kamleshchandnani

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
kamleshchandnani: default value for maxLines should be 1

susheelachoudry: > I was using truncate as switch to enable disable maxlines
@kamleshchandnani 

For example 
 overflow: ${(props) => (props.truncate ? 'hidden' : '')};
I this scenario i would need truncate. i cant depend on maxlines.

src/atoms/Text/Text.stories.js
```diff
@@ -19,4 +19,25 @@ storiesOf('Text', module)
...
       {text('Display Text', 'The quick brown fox jumps over the lazy dog')}
     </Text>
+  ))
+  .add('as span', () => (
+    <Text size={select('Size', sizeOptions, 'large')} as="span">
+      {text('Display Text', 'The quick brown fox jumps over the lazy dog')}
+    </Text>
+  ))
+  .add('show content in only 1 line and hide the rest ', () => (
+    <Text size={select('Size', sizeOptions, 'large')} truncate maxLines={1} _lineHeight="medium">
```
deorukhkarchaitanya: all ```_``` props like ```_lineHeight``` are private props are we should not use it unless it's an exception from the design end. Can we not use it here?

src/atoms/Text/Text.stories.js
```diff
@@ -19,4 +19,25 @@ storiesOf('Text', module)
     <Text size={select('Size', sizeOptions, 'large')}>
       {text('Display Text', 'The quick brown fox jumps over the lazy dog')}
     </Text>
+  ))
+  .add('as span', () => (
+    <Text size={select('Size', sizeOptions, 'large')} as="span">
+      {text('Display Text', 'The quick brown fox jumps over the lazy dog')}
+    </Text>
+  ))
+  .add('show content in only 1 line and hide the rest ', () => (
```
deorukhkarchaitanya: This should be in the knobs as just a prop you can play with

src/atoms/Text/Text.stories.js
```diff
@@ -19,4 +19,25 @@ storiesOf('Text', module)
...
+    <Text size={select('Size', sizeOptions, 'large')} truncate maxLines={1} _lineHeight="medium">
+      {text(
+        'Display Text',
+        'The quick brown fox jumps over the lazy dog he quick brown fox jumps over the lazy dog he quick brown fox jumps over the lazy dog he quick brown fox jumps over the lazy dog he quick brown fox jumps over the lazy dog he quick brown fox jumps over the lazy dog ',
+      )}
+    </Text>
+  ))
+  .add('show content in only 3 line and hide the rest ', () => (
+    <Text size={select('Size', sizeOptions, 'large')} truncate maxLines={3} _lineHeight="medium">
```
deorukhkarchaitanya: Can we remove ```truncate``` now that we have ```maxLines```?

src/atoms/Text/Text.stories.js
susheelachoudry: like rename the props ? _truncate or _maxLines ?

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,109 @@
...
+  },
+  lineHeight({ theme, size, _lineHeight }) {
+    if (_lineHeight) {
+      return theme.fonts.lineHeight[_lineHeight];
+    } else {
+      return getLineHeight(theme, size);
+    }
+  },
+  maxHeight({ theme, truncate, _lineHeight, maxLines }) {
+    if (truncate) {
```
deorukhkarchaitanya: Let's remove the ```truncate``` prop and do this instead
```suggestion
    if (isDefined(maxLines)) {
```

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,109 @@
...
+      const lineHeight = `${parseFloat(theme.fonts.lineHeight[_lineHeight])}`.replace('px', '');
+      const height = `${lineHeight * maxLines}px`;
```
deorukhkarchaitanya: Can we create a helper function for this?

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,109 @@
...
+      return 'initial';
+    }
+  },
+  whiteSpace({ truncate, maxLines }) {
+    if (truncate && maxLines > 1) {
+      return 'wrap';
+    } else if (truncate) {
+      return 'nowrap';
+    } else {
+      return '';
```
deorukhkarchaitanya: I'm not sure if this is an acceptable value for ```white-space```

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,109 @@
...
+    }
+  },
+};
+
+const Text = styled.div`
+  font-family: ${styles.fontFamily};
+  font-size: ${styles.fontSize};
+  color: ${styles.color};
+  text-decoration-line: ${(props) => (props._isUnderlined ? 'underline' : 'none')};
```
deorukhkarchaitanya: Create a style for this
```suggestion
  text-decoration-line:  ${styles.textDecorationLine};
```

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,109 @@
...
+  align-self: ${styles.align};
+  letter-spacing: ${styles.letterSpacing};
+  line-height: ${styles.lineHeight};
+  overflow: ${(props) => (props.truncate ? 'hidden' : '')};
```
deorukhkarchaitanya: Let's perform all styles related login in the styles object

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,109 @@
...
+  white-space: ${styles.whiteSpace};
+  text-overflow: ${(props) => (props.truncate ? 'ellipsis' : '')};
+  max-height: ${styles.maxHeight};
+  position: relative;
+  ${({ truncate, maxLines }) =>
```
deorukhkarchaitanya: Can we do this in styles as well?

src/atoms/Text/Text.web.js
deorukhkarchaitanya: We can remove the truncate prop now

deorukhkarchaitanya: I think default value for ```maxLines``` should be undefined and let it occupy all the space unless specified otherwise

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,109 @@
...
+Text.defaultProps = {
+  size: 'large',
+  align: 'left',
+  color: 'shade.980',
+  'data-testid': 'ds-text',
+  _weight: 'regular',
+  _isUnderlined: false,
+  _letterSpacing: 'small',
+  truncate: false,
+  maxLines: 1,
```
deorukhkarchaitanya: move this above private props. Let's keep all private props at the end

src/atoms/Text/__tests__/Text.web.test.js
```diff
@@ -0,0 +1,30 @@
...
+    expect(container).toMatchSnapshot();
+  });
+
+  it('snapshot testing with underlined text', () => {
+    const displayText = 'Displaying some text';
+    const { container } = renderWithTheme(<Text _isUnderlined>{displayText}</Text>);
+    expect(container).toMatchSnapshot();
+  });
+
+  it('snapshot testing with truncate text', () => {
```
deorukhkarchaitanya: Can you change this to ```maxLines```?

src/atoms/Text/Text.web.js
susheelachoudry: @kamleshchandnani suggested this change

src/atoms/Text/Text.stories.js
deorukhkarchaitanya: No, I'm saying don't use _lineHeight. Line height is not something we should manually define here. It is only exposed so that we can use it in internal components like Button where we have no choice but to change the line-height.

src/atoms/Text/Text.web.js
susheelachoudry: maxLines is always defined, since we have it in the default values, lmk if i am wrong

src/atoms/Text/__tests__/Text.web.test.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.stories.js
susheelachoudry: done

src/atoms/Text/Text.stories.js
susheelachoudry: done

src/atoms/Text/Text.stories.js
```diff
@@ -19,4 +19,22 @@ storiesOf('Text', module)
     <Text size={select('Size', sizeOptions, 'large')}>
       {text('Display Text', 'The quick brown fox jumps over the lazy dog')}
     </Text>
+  ))
+  .add('as span', () => (
+    <Text size={select('Size', sizeOptions, 'large')} as="span">
+      {text('Display Text', 'The quick brown fox jumps over the lazy dog')}
+    </Text>
+  ))
+  .add('show content in only 3 line and hide the rest ', () => (
```
deorukhkarchaitanya: Add maxLines as a knob

src/atoms/Text/Text.stories.js
```diff
@@ -19,4 +19,22 @@ storiesOf('Text', module)
...
+    <Text
+      size={select('Size', sizeOptions, 'large')}
+      truncate
+      maxLines={select('maxLines', 'number of lines', 3)}
+      _lineHeight="medium"
```
deorukhkarchaitanya: Don't use _lineHeight. line height needs to be derived from size.

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,140 @@
...
+import { getColorKeys, getColor, getLineHeight } from '../../_helpers/theme';
+
+/*
+ * Sets height for the div
+ * @param {Object} theme
+ * @param {String} _lineHeight
+ * @param {number} maxLines
+ */
+const calculateHeight = ({ theme, _lineHeight, maxLines }) => {
+  const lineHeight = `${parseFloat(theme.fonts.lineHeight[_lineHeight])}`;
```
deorukhkarchaitanya: lineHeight needs to be derived from size. Refer to the figma file. Line height is different for different font sizes

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,140 @@
...
+  lineHeight({ theme, size, _lineHeight }) {
+    if (_lineHeight) {
+      return theme.fonts.lineHeight[_lineHeight];
+    } else {
+      return getLineHeight(theme, size);
+    }
+  },
+  maxHeight({ theme, _lineHeight, maxLines }) {
+    if (maxLines) {
+      const maxHeight = calculateHeight({ theme, _lineHeight, maxLines });
```
deorukhkarchaitanya: This lineHeight is not correct. use ```getLineHeight(theme, size) ``` because lineHeight is based on the size. The prop _lineHeight is used only in exceptional cases and shouldn't be used anywhere at all otherwise

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,140 @@
...
+  },
+  overflow({ maxLines }) {
+    if (maxLines) {
+      return 'hidden';
+    } else {
+      return 'initial';
+    }
+  },
+  textOverflow({ maxLines }) {
+    if (maxLines) {
```
deorukhkarchaitanya:
```suggestion
    if (isDefined(maxLines)) {
```

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,140 @@
...
+  },
+  textDecoration({ _isUnderlined }) {
+    if (_isUnderlined) {
+      return 'undefined';
+    } else {
+      return 'none';
+    }
+  },
+  ellipses({ maxLines }) {
+    if (maxLines) {
```
deorukhkarchaitanya:
```suggestion
    if (isDefined(maxLines)) {
```

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,140 @@
...
+
+Text.defaultProps = {
+  size: 'large',
+  align: 'left',
+  color: 'shade.980',
+  'data-testid': 'ds-text',
+  _weight: 'regular',
+  _isUnderlined: false,
+  _letterSpacing: 'small',
+  maxLines: undefined,
```
deorukhkarchaitanya: We should keep all ```_``` props at the end. Can we move maxLines above al ```_``` props

src/atoms/Text/Text.stories.js
susheelachoudry: i have done that  in the separate story. above story shows  how to use 'as' prop

src/atoms/Text/Text.web.js
susheelachoudry: is it a helper function to check if it is defined or not ?
Also if ( maxLines ) should work for the undefined values.

src/atoms/Text/Text.stories.js
deorukhkarchaitanya: "show content in only 3 line and hide the rest " looks odd on mobile :(. The text is too long. We can skip this story and let the user play with the knobs if they need to.

src/atoms/Text/Text.stories.js
```diff
@@ -19,4 +19,21 @@ storiesOf('Text', module)
...
+  .add('as span', () => (
+    <Text size={select('Size', sizeOptions, 'large')} as="span">
+      {text('Display Text', 'The quick brown fox jumps over the lazy dog')}
+    </Text>
+  ))
+  .add('show content in only 3 line and hide the rest ', () => (
+    <Text
+      size={select('Size', sizeOptions, 'large')}
+      truncate
+      maxLines={select('maxLines', 'number of lines', 3)}
```
deorukhkarchaitanya: Weren't we going to remove ```truncate``` as a prop?

src/atoms/Text/Text.web.js
deorukhkarchaitanya: ```maxLines``` is a number so ```if(0)``` ```if(1)``` will act as boolean checks

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,140 @@
...
+Text.propTypes = {
+  size: PropTypes.oneOf(['xxsmall', 'xsmall', 'small', 'medium', 'large']),
+  color: PropTypes.oneOf(getColorKeys()),
+  align: PropTypes.oneOf(['left', 'center', 'right']),
+  'data-testid': PropTypes.string,
+  _weight: PropTypes.oneOf(Object.keys(baseTheme.fonts.weight)),
+  _isUnderlined: PropTypes.bool,
+  _letterSpacing: PropTypes.oneOf(Object.keys(baseTheme.fonts.letterSpacing)),
+  _lineHeight: PropTypes.oneOf(Object.keys(baseTheme.fonts.lineHeight)),
+  maxLines: PropTypes.number,
```
deorukhkarchaitanya:
```suggestion
  maxLines: PropTypes.number,
  _weight: PropTypes.oneOf(Object.keys(baseTheme.fonts.weight)),
  _isUnderlined: PropTypes.bool,
  _letterSpacing: PropTypes.oneOf(Object.keys(baseTheme.fonts.letterSpacing)),
  _lineHeight: PropTypes.oneOf(Object.keys(baseTheme.fonts.lineHeight)),
```

src/atoms/Text/__tests__/Text.web.test.js
```diff
@@ -0,0 +1,34 @@
...
+  it('snapshot testing with underlined text', () => {
+    const displayText = 'Displaying some text';
+    const { container } = renderWithTheme(<Text _isUnderlined>{displayText}</Text>);
+    expect(container).toMatchSnapshot();
+  });
+
+  it('snapshot testing with maxLines in text', () => {
+    const displayText = 'Displaying some text';
+    const { container } = renderWithTheme(
+      <Text truncate maxLines={2}>
```
deorukhkarchaitanya: We can remove ```truncate``` prop, right?

src/atoms/Text/Text.web.js
susheelachoudry: if (0) isn't false ? other numbers true ?

src/atoms/Text/Text.stories.js
susheelachoudry: done

src/atoms/Text/Text.stories.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.stories.js
susheelachoudry: my bad. removed it now

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/__tests__/Text.web.test.js
susheelachoudry: removed

src/atoms/Text/Text.stories.js
deorukhkarchaitanya: @susheelachoudry I think you missed this one

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,140 @@
...
+import { getColorKeys, getColor, getLineHeight } from '../../_helpers/theme';
+
+/*
+ * Sets height for the div
+ * @param {Object} theme
+ * @param {String} size
+ * @param {number} maxLines
+ */
+const calculateHeight = ({ theme, size, maxLines }) => {
+  // const lineHeight = `${parseFloat(theme.fonts.lineHeight[_lineHeight])}`;
```
deorukhkarchaitanya: Can we remove the comment now?

src/atoms/Text/Text.web.js
deorukhkarchaitanya: ```if(0)``` is false. The purpose of this check is not to treat the value of maxLines as a boolean check but to see if the prop is passed by the consumer and take action accordingly

susheelachoudry: @deorukhkarchaitanya 
if (typeof maxLines !== 'undefined')
You are suggesting something like this ?

deorukhkarchaitanya: No, I'm suggesting:

```
    if (isDefined(maxLines)) {
```

src/atoms/Text/Text.stories.js
susheelachoudry: my bad,  done

src/atoms/Text/Text.web.js
susheelachoudry: sorry i didn't get you,  what is isDefined  ?

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,139 @@
-const Text = () => null;
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import baseTheme from '../../tokens/theme';
+import isDefined from '../../_helpers/isDefined';
+import { getColorKeys, getColor, getLineHeight } from '../../_helpers/theme';
+
+/*
+ * Sets height for the div
```
kamleshchandnani: Remove this comments. We haven't figured out a way of adding doc style comments and this might conflict later on with some other tools. So we'll add it later

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,139 @@
...
+  },
+  fontSize({ theme, size }) {
+    return theme.fonts.size[size];
+  },
+  color({ theme, color }) {
+    return getColor(theme, color);
+  },
+  align({ align }) {
+    switch (align) {
+      case 'left':
```
kamleshchandnani: keep it sepate. Short circuits operations are hard to read

```js
switch (align) {
case left:
//styles
case right:
//styles
case center:
//styles
case default:
//styles
}
```

src/atoms/Text/Text.web.js
kamleshchandnani: Why did we create a new function here? The rule of thumb is create functions only if you write the same logic more than 2 times and over here we are just using it at once place that too in the same component. This affects readability.

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,139 @@
...
+      return maxHeight;
+    } else {
+      return 'initial';
+    }
+  },
+  whiteSpace({ maxLines }) {
+    if (isDefined(maxLines)) {
+      return 'initial';
+    } else {
+      return 'nowrap';
```
kamleshchandnani: Do we really need this whiteSpace styles? The wrap should be default browser and we only make a choice whether to show or hide the `...` by truncating the text

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,139 @@
...
+  },
+  overflow({ maxLines }) {
+    if (isDefined(maxLines)) {
+      return 'hidden';
+    } else {
+      return 'initial';
+    }
+  },
+  textOverflow({ maxLines }) {
+    if (isDefined(maxLines)) {
```
kamleshchandnani: you can shorten all your styles like this

```js
if (isDefined(maxLines)) {
  return "ellipsis";
}

return "initial";
```

When you have return statements in `if` and you just have `if..else` pair then you can skip the `else` portion. Make this change across this file

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,139 @@
...
+ */
+const calculateHeight = ({ theme, size, maxLines }) => {
+  const lineHeight = getLineHeight(theme, size);
+  const containerHeight = `${parseFloat(lineHeight) * maxLines}px`;
+  return containerHeight;
+};
+
+const styles = {
+  fontFamily({ theme, _weight }) {
+    return theme.fonts.family.lato[_weight];
```
kamleshchandnani: when we don't have any condition then put this directly in the component styles itself below.

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,139 @@
...
+  },
+  fontSize({ theme, size }) {
+    return theme.fonts.size[size];
```
kamleshchandnani: when we don't have any condition then put this directly in the component styles itself below.

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,139 @@
...
+  },
+  color({ theme, color }) {
+    return getColor(theme, color);
```
kamleshchandnani: when we don't have any condition then put this directly in the component styles itself below.

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,139 @@
...
+      default:
+        return 'flex-start';
+      case 'center':
+        return 'center';
+      case 'right':
+        return 'flex-end';
+    }
+  },
+  letterSpacing({ theme, _letterSpacing }) {
+    return `${theme.fonts.letterSpacing[_letterSpacing]}`;
```
kamleshchandnani: when we don't have any condition then put this directly in the component styles itself below.

src/atoms/Text/Text.web.js
kamleshchandnani: typo `ellipses` -> `ellipsis`

src/atoms/Text/__tests__/Text.web.test.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it('snapshot testing with underlined text', () => {
+    const displayText = 'Displaying some text';
+    const { container } = renderWithTheme(<Text _isUnderlined>{displayText}</Text>);
+    expect(container).toMatchSnapshot();
+  });
+
+  it('snapshot testing with maxLines in text', () => {
+    const displayText = 'Displaying some text';
```
kamleshchandnani: To test the `maxLines` scenario atleast add text which has content for 2 lines?

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,139 @@
...
+  textOverflow({ maxLines }) {
+    if (isDefined(maxLines)) {
+      return 'ellipsis';
+    } else {
+      return 'initial';
+    }
+  },
+  textDecoration({ _isUnderlined }) {
+    if (_isUnderlined) {
+      return 'undefined';
```
kamleshchandnani: There's no such value called `undefined` maybe it should have been `underlined`? This should have been caught in tests.

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,139 @@
...
+      return '';
+    }
+  },
+};
+
+const Text = styled.div`
+  font-family: ${styles.fontFamily};
+  font-size: ${styles.fontSize};
+  color: ${styles.color};
+  text-decoration-line: ${styles.textDecoration};
```
kamleshchandnani: why `text-decoration-line` and not `text-decoration`? `text-decoration-line` is not supported on IE

kamleshchandnani:
Screenshot 2020-03-23 at 1 33 01 PM
@susheelachoudry Look at the text in the storybook. The `...` appears on the `dog` itself. Also the knob for `maxLines` doesn't work

src/atoms/Text/Text.stories.js
```diff
@@ -19,4 +19,16 @@ storiesOf('Text', module)
     <Text size={select('Size', sizeOptions, 'large')}>
       {text('Display Text', 'The quick brown fox jumps over the lazy dog')}
     </Text>
+  ))
+  .add('as span', () => (
+    <Text
+      size={select('Size', sizeOptions, 'large')}
+      as="span"
+      maxLines={select('maxLines', 'number of lines', 3)}
```
kamleshchandnani: This doesn't work. you need to provide the list of options
`select('maxLines', maxLinesOptions, 3)`
The as prop should also become the knob with few most possible values used `div, span, p,`

src/atoms/Text/__tests__/Text.web.test.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { renderWithTheme } from '../../../_helpers/testing';
+
```
kamleshchandnani: Where's the test for testing `as` prop?

src/atoms/Text/__tests__/Text.web.test.js
susheelachoudry: have added the knob to check maxlines.

src/atoms/Text/Text.web.js
susheelachoudry: sorry i didnt get this, separate as in ?

kamleshchandnani:
Screenshot 2020-03-23 at 3 12 25 PM
Issues when `maxLines` set to 2 👆 

Screenshot 2020-03-23 at 3 12 46 PM
Issues when `maxLines` set to 5 👆. Basically if lines have less text then the available width in the maxLines the issue occurs.

src/atoms/Text/Text.web.js
kamleshchandnani: don't do this
```js
switch (align) {
case left:
case default:
//styles
```
Basically don't club 2 cases

src/atoms/Text/Text.web.js
kamleshchandnani: This is not resolved?

src/atoms/Text/__tests__/Text.web.test.js
kamleshchandnani: Knobs are in storybook. These are tests

src/atoms/Text/__tests__/Text.web.test.js
susheelachoudry: done

src/atoms/Text/Text.stories.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.stories.js
```diff
@@ -11,12 +11,28 @@ const sizeOptions = {
...
     component: Text,
   })
   .add('default', () => (
-    <Text size={select('Size', sizeOptions, 'large')}>
-      {text('Display Text', 'The quick brown fox jumps over the lazy dog')}
+    <Text size={select('Size', sizeOptions, 'large')} as="span">
+      {text('Display Text', 'The quick brown fox jumps over the lazy dog ')}
+    </Text>
+  ))
+  .add('text with as, maxLines', () => (
```
kamleshchandnani: `with maxLines`

susheelachoudry: as prop is also there

src/atoms/Text/Text.web.js
susheelachoudry: based on our conversation , keeping it here since it is extracted via theme.

kamleshchandnani: theme is available inside styled props

susheelachoudry: I have already pushed these changes, @kamleshchandnani

src/atoms/Text/Text.web.js
susheelachoudry: done

src/atoms/Text/Text.web.js
```diff
@@ -1,3 +1,108 @@
...
+
+const styles = {
+  color({ theme, color }) {
+    return getColor(theme, color);
+  },
+  align({ align }) {
+    switch (align) {
+      case 'left':
+        return 'flex-start';
+      default:
```
kamleshchandnani: default comes at the last 🙈

PR #55, opened 2020-03-05
fix(text input): fix Label & alignment of TextInput

deorukhkarchaitanya:


src/atoms/TextInput/TextInput.native.js
```diff
@@ -14,18 +14,19 @@ import automation from '../../_helpers/automation-attributes';
...
 const IS_ANDROID = Platform.OS === 'android';
 
 const styles = {
   textInput: {
     padding({ variant, hasLeftIcon, hasPrefix, hasText }) {
-      const paddingTop = IS_ANDROID ? '8px' : '0px';
-      const paddingRight = '0px';
-      const paddingBottom = IS_ANDROID ? '0px' : hasText ? '4px' : '2px';
-      const paddingLeft = variant === 'outlined' || hasLeftIcon || hasPrefix ? '0px' : '8px';
+      const paddingTop = IS_ANDROID ? (variant === 'filled' ? 0 : 1) : 0;
```
shridhar52war: Just avoiding nested ternary operations.
```
let [paddingTop, paddingRight,paddingBottom, paddingLeft ] = [0,0,0,0];

if(IS_ANDROID){
  paddingTop = variant === 'filled' ? 0 : 1;
}else{
  paddingBottom = hasText ? 0.5 : 0.25;
}

paddingLeft = variant === 'outlined' || hasLeftIcon || hasPrefix ? 0 : 1;

return [paddingTop, paddingRight,paddingBottom, paddingLeft];
```

deorukhkarchaitanya:
@shubhnik @shridhar52war @dhruvdutt @kamleshchandnani review, please.

src/atoms/TextInput/Label.native.js
```diff
@@ -257,11 +267,15 @@ RegularLabel.propTypes = {
...
     y: PropTypes.number,
-  }).isRequired,
+  }),
+  labelPosition: PropTypes.oneOf(['left', 'top']).isRequired,
+  disabled: PropTypes.bool,
 };
 
 RegularLabel.defaultProps = {
   children: 'Label',
+  inputLayoutDimensions: undefined,
```
dhruvdutt: Won't this be `undefined` by default since `props` is an object?

dhruvdutt: You can also keep this `{}`.

src/atoms/TextInput/TextInput.native.js
dhruvdutt: Isn't this rule enabled on eslint? [`no-nested-ternary`](https://eslint.org/docs/rules/no-nested-ternary)

deorukhkarchaitanya: I don't think so. It's coming from universe

src/atoms/TextInput/Label.native.js
deorukhkarchaitanya: We can, but I don't see the benefit of making it an object when it is actually undefined on initial render

dhruvdutt: Even if we don't explicitly mention in `defaultProps`, wouldn't it be `undefined` anyway?

src/atoms/TextInput/TextInput.native.js
dhruvdutt: [It's off currently.](https://github.com/kentcdodds/eslint-config-kentcdodds/blob/8e5720c101c895856f615b7de7b549114b1b02fc/stylistic.js#L49)
@kamleshchandnani

src/atoms/TextInput/Label.native.js
deorukhkarchaitanya: We decided to explicitly make it undefined for all non-```isRequired``` props. We're doing this everywhere

dhruvdutt: Cool, makes sense. 👍

src/atoms/TextInput/TextInput.native.js
```diff
@@ -182,8 +199,13 @@ const TextInput = ({
   return (
     <Flex justifyContent="flex-end">
       <View>
+        {!hasAnimatedLabel && labelPosition === 'top' ? (
```
kamleshchandnani: what if we just write `labelPosition==='top'`. Any reason you explicitly need to check `!hasAnimated`?

src/atoms/TextInput/TextInput.native.js
```diff
@@ -200,8 +222,14 @@ const TextInput = ({
         <Flex flexDirection="row" alignItems="flex-start">
           <View>
             {/* Fixed Left Label */}
-            {hasLeftLabel && layoutDimensions ? (
-              <Label.Regular inputLayoutDimensions={layoutDimensions}>{label}</Label.Regular>
+            {!hasAnimatedLabel && layoutDimensions && labelPosition === 'left' ? (
```
kamleshchandnani: just `labelPosition === 'left'` would work?

src/atoms/TextInput/TextInput.native.js
```diff
@@ -140,6 +157,11 @@ const TextInput = ({
     iconRight,
   });
 
+  if (labelPosition === 'left' && variant === 'outlined') {
+    // Outlined Text Input has only a top label
+    throw Error('Cannot have a left label on an outlined Text Input');
```
kamleshchandnani: refer the error message format in `Space` `Flex` etc.

src/atoms/TextInput/Label.native.js
```diff
@@ -257,11 +267,15 @@ RegularLabel.propTypes = {
     width: PropTypes.number,
     x: PropTypes.number,
     y: PropTypes.number,
-  }).isRequired,
+  }),
+  labelPosition: PropTypes.oneOf(['left', 'top']).isRequired,
```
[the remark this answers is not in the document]
deorukhkarchaitanya: Can't use ```position``` it is a keyword here for ```position: 'absolute'``` & ```position: 'relative'```

src/atoms/TextInput/TextInput.native.js
deorukhkarchaitanya: label position can be 'top' and still have an animated label for outlined

src/atoms/TextInput/TextInput.native.js
deorukhkarchaitanya: We need a check for ```layoutDimensions``` and ```!hasAnimatedLabel``` is needed to ensure that we don't render this label for position left & outlined variant

PR #57, opened 2020-03-06
feat(badge): add badge for react-native

shubhnik:
Adds Badges(pills/rectangular badges) for DS


src/atoms/Badge/Badge.web.js
```diff
@@ -0,0 +1,3 @@
+const View = () => null;
```
kamleshchandnani:
```suggestion
const View = () => <div>WIP</div>;
```

PR #61, opened 2020-03-16
fix(TextArea/native): modify label and accessory position

deorukhkarchaitanya:


src/atoms/TextInput/AccessoryIcon.native.js
```diff
@@ -42,12 +46,14 @@ AccessoryIcon.propTypes = {
   disabled: PropTypes.bool,
   hasError: PropTypes.bool,
   variant: PropTypes.oneOf(['filled', 'outlined']).isRequired,
+  _isMultiline: PropTypes.bool,
```
kamleshchandnani: This is basically mixing concerns. AccessoryIcon, AccessoryText should not be aware whether they are part of multiline or not.

kamleshchandnani: Everything is not flowing here and there. Maybe we can rethink this. Let me give it more thougts

deorukhkarchaitanya: TextInput & TextInput with multiline prop has very different visual properties on Android & iOS. Right now these accessories were not aligning with TextInput multiline but are aligning with TextInput.

deorukhkarchaitanya: The UI is broken without this. I suggest we merge this for now & revisit it later. No point keeping it half baked

PR #64, opened 2020-03-17
fix(Space): replace native renderer with react renderer for web

krishnaacc123:
fix(space.web.test.js): Replace native render to react react render

PR #65, opened 2020-03-17
feat(Amount): create Amount component for react-native

deorukhkarchaitanya:


kamleshchandnani:
Can you check why did build fail for web? Just run this command on local
`yarn test:web && yarn:build:web`

deorukhkarchaitanya:
> Can you check why did build fail for web? Just run this command on local
> `yarn test:web && yarn web:storybook:build`

Failing because this is not merged https://github.com/razorpay/blade/pull/64

kamleshchandnani:
This should go to molecules

deorukhkarchaitanya:
> This should go to molecules

Oh yeah!

src/atoms/Amount/Amount.native.js
```diff
@@ -0,0 +1,162 @@
...
+        case 'center':
+          return 'center';
+        case 'right':
+          return 'flex-end';
+      }
+    },
+  },
+};
+
+const BaseText = styled(NativeText)`
```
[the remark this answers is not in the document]
deorukhkarchaitanya: The ```BaseText``` here is a little more complex than what our DS ```Text``` is. DS ```Text``` only accepts a string and renders it. Here I need to pass different components of Text like Heading & DS Text. 

DS Text only accepts a string as children and we're passing components here.

src/molecules/Amount/Amount.native.js
```diff
@@ -0,0 +1,169 @@
...
+        >{`.${fractionDisplay}`}</Text>
+      </BaseText>
+    </Flex>
+  );
+};
+
+Amount.propTypes = {
+  children: PropTypes.string.isRequired,
+  size: PropTypes.oneOf(['medium', 'large', 'xlarge', 'xxlarge', 'xxxlarge']),
+  align: PropTypes.oneOf(['left', 'center', 'right']),
```
kamleshchandnani: remove align and make it left aligned as we discussed.

src/molecules/Amount/Amount.native.js
```diff
@@ -0,0 +1,169 @@
...
+import styled from 'styled-components/native';
+import automation from '../../_helpers/automation-attributes';
+import Flex from '../../atoms/Flex';
+import Text from '../../atoms/Text';
+import Heading from '../../atoms/Heading';
+
+const styles = {
+  text: {
+    color({ size }) {
+      switch (size) {
```
kamleshchandnani: Why do you need a switch case here? It's just

```js
if (size === "xxxlarge") {
  return "shade.950";
}

return "shade.940";
```

src/molecules/Amount/Amount.native.js
```diff
@@ -0,0 +1,169 @@
...
+};
+
+const Amount = ({ size, align, testID, children }) => {
+  if (isNaN(children)) {
+    throw new Error(`Expected children to be number \n(Eg. "1234", "12.34")`);
+  }
+
+  const currency = '₹';
+
+  const [integerPart, fractionPart] = children.split('.');
```
kamleshchandnani: Wouldn't this fail like the `fractionPart` will be `undefined`? The case where the user will enter `123`.
You can do it this way
```js
const [integerPart, fractionPart] = String(parseFloat(amount).toFixed(2)).split('.')
```

src/molecules/Amount/Amount.native.js
```diff
@@ -0,0 +1,169 @@
...
+
+  const fractionDisplay = getFractionalDisplay(fractionPart);
```
kamleshchandnani: once you do the above I guess you don't need this function at all.

src/molecules/Amount/Amount.native.js
```diff
@@ -0,0 +1,169 @@
...
+};
+
+const getLocaleString = (numberString) => {
+  if (Platform.OS === 'android') {
+    // only android needs this polyfill
+    require('intl');
+    require('intl/locale-data/jsonp/en-IN');
+    return new Intl.NumberFormat('en-IN').format(numberString);
+  } else {
+    return parseInt(numberString, 10).toLocaleString('en-IN');
```
kamleshchandnani:
```js
const formatAmount = ({ amount, currency }) => {
  //polyfill here

  //normal toLocale method
  const formattedAmount = (+amount).toLocaleString("en-IN", {
    style: "currency",
    currency,
    minimumFractionDigits: 2,
    maximumFractionDigits: 2
  });
  return formattedAmount;
};
```

src/molecules/Amount/Amount.native.js
```diff
@@ -0,0 +1,169 @@
+import React from 'react';
```
kamleshchandnani: How above this API
```jsx
<Amount size="medium" currency="INR">
  1234
</Amount>;
```

The currency can be validated for the this https://www.currency-iso.org/dam/downloads/lists/list_one.xml -- for now we can just use INR but this opens up door for future support

The default value of currency can be set to INR as well

src/molecules/Amount/Amount.native.js
kamleshchandnani: You can then do splits and other things to apply stylings to individual pieces below

src/molecules/Amount/Amount.native.js
deorukhkarchaitanya: It won't fail because ```const getFractionalDisplay = (fractionalPart = '') => {```

But yeah, I can try the ```toFixed``` thing

src/molecules/Amount/Amount.native.js
deorukhkarchaitanya: I thought about it but supporting currency right now will add a lot more things like currency symbols based on what's selected, currency formatting based on selection. 
And since we're using a polyfill, we also need to ensure all the formatting is consistent across both the functions.

Not sure if it's worth the efforts as of now.

kamleshchandnani: I guess the intl is just a polyfill it should just work fine in the polyfill with the same options. See my below comment for the formatAmount function. It's just the matter of passing the currency to it nothing else

src/molecules/Amount/Amount.native.js
deorukhkarchaitanya: I'll raise a separate PR for removing ```align``` from Text & other components. Okay?

src/_helpers/currency.js
```diff
@@ -0,0 +1,185 @@
+const geISOCurrencyList = () => {
```
kamleshchandnani: move this to `molecules/Amount` directory only. Since right now no one else will use it. we'll move it here when we have more use cases

src/molecules/Amount/Amount.native.js
```diff
@@ -0,0 +1,136 @@
...
+  return formattedAmount;
+};
+
+const Amount = ({ size, testID, children, currency }) => {
+  if (isNaN(children)) {
+    throw new Error(`Expected children to be number \n(Eg. "1234", "12.34")`);
+  }
+
+  const formattedAmount = formatAmount({ amount: children, currency });
+  const [currencySymbol, amount] = formattedAmount.split(/\s/);
```
kamleshchandnani: `currency, amount`

deorukhkarchaitanya: ```currency``` is a prop

src/_helpers/currency.js
deorukhkarchaitanya: Okay

src/molecules/Amount/Amount.native.js
kamleshchandnani: fair enough 😄

PR #68, opened 2020-03-23
chore: remove align prop from Text & Heading for react-native

deorukhkarchaitanya:


deorukhkarchaitanya:
@shubhnik @shridhar52war please checkout to this PR and see if your components are affected. I had to update snapshots for all your components. I've done a visual check and things seem okay. Please approve this once you've gone through the components.

dhruvdutt:
Why are we removing this prop?

kamleshchandnani:
> Why are we removing this prop?
Alignment should be handled by parent and not the component itself. Rendering the text component and doing text alignment won't make any difference. We have enhancers for this purpose

src/atoms/Text/Text.native.js
```diff
@@ -50,14 +39,12 @@ const StyledText = styled(NativeText)`
   font-size: ${styles.fontSize};
   color: ${styles.color};
   text-decoration-line: ${styles.textDecorationLine};
```
kamleshchandnani: you didn't had to mention align left as a defualt style as you were mentioning the other day? It takes it by default?

deorukhkarchaitanya: The parent will take care

PR #70, opened 2020-03-24
fix(Amount/native): add more variants, weights and sizes

deorukhkarchaitanya:


src/molecules/Amount/Amount.native.js
```diff
@@ -122,15 +155,21 @@ const Amount = ({ size, testID, children, currency }) => {
 
 Amount.propTypes = {
   children: PropTypes.string.isRequired,
-  size: PropTypes.oneOf(['medium', 'large', 'xlarge', 'xxlarge', 'xxxlarge']),
+  size: PropTypes.oneOf(['xsmall', 'medium', 'large', 'xlarge', 'xxlarge', 'xxxlarge']),
   testID: PropTypes.string,
   currency: PropTypes.oneOf(geISOCurrencyList()),
+  variant: PropTypes.oneOf(['camel', 'normal', 'camel-subtle', 'normal-subtle']),
```
kamleshchandnani: Wait what?? We don't do composition in props
why `subtle`?

kamleshchandnani: If we keep allowing this then we'll end up scenarios like `camel-bold` `camel-bold-subtle` and so on and then everything goes for a toss

deorukhkarchaitanya: Camel & normal have a subtle version too. Our initial API was supposed to have variants as subtle & normal but subtle somehow got missed during implementation.

src/molecules/Amount/Amount.native.js
```diff
@@ -87,7 +99,7 @@ const formatAmount = ({ amount, currency }) => {
   return formattedAmount;
 };
 
-const Amount = ({ size, testID, children, currency }) => {
+const Amount = ({ size, testID, children, currency, weight, variantColor, variant, subtle }) => {
```
kamleshchandnani:
```suggestion
const Amount = ({ size, testID, children, currency, weight, variant, variantColor, subtle }) => {
```
sorry not sorry 🙈

src/molecules/Amount/Amount.native.js
```diff
@@ -96,24 +108,46 @@ const Amount = ({ size, testID, children, currency }) => {
   const [currencySymbol, amount] = formattedAmount.split(/\s/);
   const [integerPart, fractionPart] = amount.split('.');
 
+  const IntegerComponent =
+    size === 'xlarge' || size === 'xxlarge' || size === 'xxxlarge' ? Heading : Text;
```
kamleshchandnani:
```suggestion
  const Component =
    (size === 'xlarge' || size === 'xxlarge' || size === 'xxxlarge') ? Heading : Text;
```
Integer is not proper name because for normal it has the whole amount `$123.00` while in camel it's just the `123`
Couldn't come up with a better name(Even AmountComponent doesn't fit). Open for anything else

src/molecules/Amount/Amount.native.js
```diff
@@ -122,15 +156,23 @@ const Amount = ({ size, testID, children, currency }) => {
...
 };
 
 Amount.defaultProps = {
   size: 'medium',
   testID: 'ds-amount',
   currency: 'INR',
+  variant: 'camel',
+  weight: 'bold',
+  variantColor: 'shade',
+  subtle: false,
```
kamleshchandnani:
```suggestion
  weight: 'bold',
  variant: 'camel',
  variantColor: 'shade',
  subtle: false,
```

just better code readability when someone directly looks at the API. same applies to propTypes

src/molecules/Amount/__tests__/Amount.native.test.js
```diff
@@ -6,31 +6,211 @@ beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
...
       expect(container).toMatchSnapshot();
     });
-    it('renders amount with size xlarge', () => {
-      const displayText = '1234.00';
-      const { container } = renderWithTheme(<Amount size="xlarge">{displayText}</Amount>);
-      expect(container).toMatchSnapshot();
+  });
+
+  describe('size', () => {
+    describe(' with camel variant', () => {
```
kamleshchandnani: extra white space
```suggestion
    describe('with camel variant', () => {
```

src/molecules/Amount/__tests__/Amount.native.test.js
```diff
@@ -6,31 +6,211 @@ beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
...
+        );
+        expect(container).toMatchSnapshot();
+      });
     });
-    it('renders amount with size xxxlarge', () => {
-      const displayText = '1234.00';
-      const { container } = renderWithTheme(<Amount size="xxxlarge">{displayText}</Amount>);
-      expect(container).toMatchSnapshot();
+
+    describe(' with normal variant', () => {
```
kamleshchandnani:
```suggestion
    describe('with normal variant', () => {
```

src/molecules/Amount/__tests__/Amount.native.test.js
```diff
@@ -6,31 +6,211 @@ beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
 afterAll(() => jest.restoreAllMocks());
 
 describe('<Amount />', () => {
```
kamleshchandnani: These tests look so beautiful 💟

src/molecules/Amount/Amount.native.js
deorukhkarchaitanya: ```PrimaryText```&  ```SecondaryText```

src/molecules/Amount/Amount.native.js
deorukhkarchaitanya: Hahaha thanks. Now I can't unsee it 😂

PR #71, opened 2020-03-24
Add testing and stories in Space Web

krishnaacc123:


src/atoms/Space/Space.stories.js
```diff
@@ -0,0 +1,25 @@
...
+  large: 'large',
+};
+
+storiesOf('Space', module)
+  .addParameters({
+    component: Space,
+  })
+  .add('default', () => (
+    <Space margin={[1, 1]} padding={[2, 2]}>
+      <Text _weight="bold" _isUnderlined size={select('Size', sizeOptions, 'medium')} as="span">
```
deorukhkarchaitanya: Don't use any underscore ```_``` props unless necessary(from design)

krishnaacc123: But it's a part of Text component

src/atoms/Space/Space.web.js
```diff
@@ -4,11 +4,14 @@ import styled from 'styled-components';
...
 
-const Space = styled(({ margin, padding, children, ...props }) => {
+const Space = styled(({ margin, padding, children, style, ...props }) => {
   if (React.Children.toArray(children).length !== 1) {
     throw new Error('Expected a single child for Space component');
   }
-  return React.cloneElement(children, props);
+  return React.cloneElement(children, {
+    ...props,
+    style: [style, children.props.style].filter(Boolean),
```
deorukhkarchaitanya: React.js does not accept Array as a valid type for ```style``` prop. That is only supported on React Native. I tested this out and it crashes. You can reproduce this by trying to pass inline style to Text in your story.

Fix:

```suggestion
    style: { ...style, ...children.props.style },
```

src/atoms/Space/Space.stories.js
deorukhkarchaitanya: All ```_``` props are called "private props" which are meant to be used only as an exception from design

dhruvdutt: Private props usage shouldn't be part of stories.

src/atoms/Space/Space.stories.js
```diff
@@ -0,0 +1,25 @@
...
+  large: 'large',
+};
+
+storiesOf('Space', module)
+  .addParameters({
+    component: Space,
+  })
+  .add('default', () => (
+    <Space margin={[1, 1]} padding={[2, 2]}>
+      <Text size={select('Size', sizeOptions, 'medium')} as="span">
```
dhruvdutt: Don't think size knob for Text component is useful for Space stories.
Better have knobs around Space component itself.

src/atoms/Space/Space.web.js
krishnaacc123: added

krishnaacc123:
Closing this PR. Had some problem with. New PR: https://github.com/razorpay/blade/pull/73

PR #76, opened 2020-03-25
chore(eslint): add import/order rule

kamleshchandnani:


src/atoms/Switch/Switch.native.js
```diff
@@ -219,7 +219,7 @@ const Switch = ({ disabled, on, defaultOn, onChange, size, testID }) => {
       animatedRightValue.setValue(animationConfig.on.rightSpace);
       setToggle(true);
     }
-  }, [on, defaultOn]);
+  }, [on, defaultOn, toggle, animatedLeftValue, animatedRightValue, animationConfig]);
```
dhruvdutt: Why aren't these things caught earlier as warning?

dhruvdutt: We have this [`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks), right?

kamleshchandnani: Yes! 2 possible reasons. Either the lint was bypassed or the plugin was added later 🤦‍♂

PR #81, opened 2020-03-29
feat(Size/react): add component

susheelachoudry:


src/atoms/Size/Size.stories.js
```diff
@@ -0,0 +1,23 @@
...
+    component: Size,
+  })
+  .add('default', () => (
+    <Size
+      height={text('height', '100%')}
+      width={text('width', '10px')}
+      maxHeight={text('maxHeight', '20px')}
+      maxWidth={text('maxWidth', '20px')}
+      minHeight={text('minHeight', '5px')}
+      minWidth={text('minWidth', '5px')}
```
dhruvdutt: This story is not helpful visually. Can you try to add a background color or border so we'll know the dimensions when playing around knobs?

susheelachoudry: I have provided a text component for that

src/atoms/Size/Size.web.js
```diff
@@ -1,3 +1,34 @@
...
+
+const Size = styled(
+  ({ minWidth, width, maxWidth, minHeight, height, maxHeight, children, style, ...props }) => {
+    if (React.Children.toArray(children).length !== 1) {
+      throw new Error('Expected a single child for Size component');
+    }
+    return React.cloneElement(children, {
+      ...props,
+      style: [style, children.props.style].filter(Boolean),
+      'data-testid': 'size-view',
```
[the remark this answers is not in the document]
susheelachoudry: @shubhnik thanks, I missed out on this.

src/atoms/Size/Size.stories.js
dhruvdutt: Yes but it's not possible to visually identify if the container height / width has been changed or not.

dhruvdutt: No matter what height / width I set, the Text is rendered same.

src/atoms/Size/__tests__/Size.web.test.js
```diff
@@ -0,0 +1,49 @@
...
+          maxWidth="20px"
+          minHeight="5px"
+          minWidth="5px"
+        >
+          <View />
+        </Size>,
+      );
+      expect(container).toMatchSnapshot();
+    });
+    it('should render child component with no sizings applied', () => {
```
kamleshchandnani:
```suggestion
    it('renders a child component without any sizing properties being applied', () => {
```

src/atoms/Size/Size.stories.js
```diff
@@ -0,0 +1,23 @@
...
+    >
+      <Text>simple div</Text>
```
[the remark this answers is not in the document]
susheelachoudry: Is it always necessary to wrap inside view? asking out of curiosity   @kamleshchandnani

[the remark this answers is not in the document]
kamleshchandnani: On native it might break things and our stories are common hence we need to come on some common ground

src/atoms/Size/Size.stories.js
susheelachoudry: should we add  overflow: hidden; ideally child component should occupy the parent space not more than that  @dhruvdutt

kamleshchandnani:
Please follow the naming convention for PR titles and PR messages. I've updated the PR title for you in this PR

src/atoms/Size/Size.web.js
```diff
@@ -1,3 +1,31 @@
...
+const Size = styled(
+  ({ minWidth, width, maxWidth, minHeight, height, maxHeight, children, ...props }) => {
+    if (React.Children.toArray(children).length !== 1) {
+      throw new Error('Expected a single child for Size component');
+    }
+
+    return React.cloneElement(children, props);
+  },
+)`
+  ${(props) => (props.width ? `width: ${props.width}` : '')};
```
kamleshchandnani: all the props here should be 
```js
${(props) => (props.width ? `width: ${makePxValue(props.width)}` : '')};
```
refer space.web.js for reference

I don't know how this got missed on native 🙊 
cc @shubhnik

shubhnik: @kamleshchandnani I would fix it and raise a PR. I remember I just changed the `cloneElement` thing in `<Size />` but should have caught this mistake earlier 🙈

susheelachoudry: this is done

susheelachoudry:
> Please follow the naming convention for PR titles and PR messages. I've updated the PR title for you in this PR

My bad, I will follow this from now on

src/atoms/Size/__tests__/Size.web.test.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { renderWithTheme } from '../../../_helpers/testing';
+import View from '../../View';
+import Size from '../index';
+
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
+afterAll(() => jest.restoreAllMocks());
+
```
kamleshchandnani: the test that actually takes scales as input is missing for example
`height={1}` will go through `makePxValue` and become `8px`.
So you need to write a test that actually takes these numeric values apart from `'10px'`

src/atoms/Size/__tests__/Size.web.test.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { renderWithTheme } from '../../../_helpers/testing';
+import View from '../../View';
+import Size from '../index';
+
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
```
kamleshchandnani: Why do we need this?

shubhnik: @kamleshchandnani we need this to suppress errors logged by React component into console while testing for "error" cases:

Screenshot 2020-04-02 at 6 51 34 PM

src/atoms/Size/Size.web.js
```diff
@@ -1,3 +1,33 @@
...
+  ${(props) => (props.width ? `width: ${makePxValue(props.width)}` : '')};
+  ${(props) => (props.height ? `height: ${makePxValue(props.height)}` : '')};
+  ${(props) => (props.minWidth ? `min-width: ${makePxValue(props.minWidth)}` : '')};
+  ${(props) => (props.maxWidth ? `max-width: ${makePxValue(props.maxWidth)}` : '')};
+  ${(props) => (props.minHeight ? `min-height: ${makePxValue(props.minHeight)}` : '')};
+  ${(props) => (props.maxHeight ? `max-height: ${makePxValue(props.maxHeight)}` : '')};
+`;
+
+Size.propTypes = {
+  width: PropTypes.string,
```
shubhnik, line 25: All the props should accept a string or number. String for absolute value and number for defining scale.
`width: PropTypes.oneOfType([PropTypes.number, PropTypes.string])`

susheelachoudry: what about px or % ? @shubhnik

shubhnik: That would just be strings.

Here the three possibilities:
`<Size height="20px"/>` (typeof height is string)
`<Size height="100%"/>` (typeof height is string)
`<Size height={2}/>` (typeof height is number)

src/atoms/Size/__tests__/Size.web.test.js
kamleshchandnani: that happens for tests that are testing error cases. So this suppress should be part of that particular test and not `beforeAll` which runs before each test run.

src/atoms/Size/Size.web.js
susheelachoudry: if 100 is passed as width, will it be 100px or 100%?

kamleshchandnani: if you just pass 
```
<Size height={100}/>
```
it will pass it through makePxValue function which then multiplies it with 8px grid and make it 100*8=800px

Check the native implementation and also the makePxValue function

src/atoms/Size/__tests__/Size.web.test.js
```diff
@@ -0,0 +1,54 @@
...
+        minWidth="5px"
+      >
+        <View />
+      </Size>,
+    );
+    expect(container).toMatchSnapshot();
+  });
+  it('renders a child by applying height, width, maxHeight, maxWidth, minWidth, minHeight in numeric values', () => {
+    const { container } = renderWithTheme(
+      <Size height="100" width="10" maxHeight="20" maxWidth="20" minHeight="5" minWidth="5">
```
kamleshchandnani, line 24:
```suggestion
      <Size height={100} width={10} maxHeight={20} maxWidth={20} minHeight={5} minWidth={5}>
```
When you want to use scale you need to pass numbers and not strings

susheelachoudry: oops, how did I miss this

PR #82, opened 2020-03-31
fix(Flex/react): add jest spy in tests

krishnaacc123:


src/atoms/Flex/__tests__/Flex.web.test.js
```diff
@@ -7,6 +7,9 @@ import Text from '../../Text';
 import Space from '../../Space';
 import Flex from '../index';
 
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
```
kamleshchandnani: why do you need this?

shubhnik: https://github.com/razorpay/blade/pull/81#discussion_r402307800

krishnaacc123: To suppress the error message that comes in test case where we pass more than on child and try to throw error.

krishnaacc123: Will move it to that test case instead of beforeAll

kamleshchandnani: Yes please move it to that particular test case

PR #84, opened 2020-03-31
fix(Amount/native): amount crash on different environments

deorukhkarchaitanya:
Use the local intl library since different node environments
have different behaviors for intl functions

### DIFF
--- a/src/molecules/Amount/Amount.native.js
+++ b/src/molecules/Amount/Amount.native.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Text as NativeText, Platform } from 'react-native';
+import { Text as NativeText } from 'react-native';
+import Intl from 'intl';
 import automation from '../../_helpers/automation-attributes';
 import Flex from '../../atoms/Flex';
 import AtomText from '../../atoms/Text';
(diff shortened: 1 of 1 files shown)


src/molecules/Amount/Amount.native.js
```diff
@@ -1,15 +1,15 @@
...
 import automation from '../../_helpers/automation-attributes';
 import Flex from '../../atoms/Flex';
 import AtomText from '../../atoms/Text';
 import Heading from '../../atoms/Heading';
 import baseTheme from '../../tokens/theme';
 import { getVariantColorKeys } from '../../_helpers/theme';
 import geISOCurrencyList from './geISOCurrencyList';
-
-const IS_ANDROID = Platform.OS === 'android';
+import 'intl/locale-data/jsonp/en-IN';
```
kamleshchandnani: when we say en-IN there's an issue if you play with knobs. the currency is not formatted properly. for ex: 5,00,000 is INR, while 500,000 in EUR but it always shows 5,00,000. Maybe the fix is not required immediately but we'll need it

deorukhkarchaitanya: Yeah, I thought about that too. The formatting will be in en-In format but I've kept it that way for now since other formatting are way more complex. Some of them have currency symbols at the end. 

We'll modify this based on product requirement (the currencies we want to officially support)

PR #85, opened 2020-03-31
fix(TextInput/native): change icon size from 12 to 16

deorukhkarchaitanya:


kamleshchandnani:
It’s not chore. It’s a fix. Chore is something that doesn’t modifies any code source files.

PR #86, opened 2020-04-02
feat(TextInput/native): add more props as per native TextInput

deorukhkarchaitanya:
This will make the behavior same as text input on native & fix our OTP problem


src/atoms/TextInput/TextInput.native.js
```diff
@@ -249,6 +249,10 @@ const TextInput = ({
     [layoutDimensions, setLayoutDimensions],
   );
 
+  useEffect(() => {
+    if (!isEmpty(children)) setInput(children);
```
dhruvdutt, line 253: What is the use-case?

deorukhkarchaitanya: This makes the component controlled. So when you pass children to text input, it will use that as the value to render but also give you the flexibility to edit it and send updates via ```onChangeText```

Our use case here is to render OTP text field with empty state and update it after retrieving from sms

dhruvdutt: Got it.

dhruvdutt, line 253:
```suggestion
    setInput(children);
```

dhruvdutt: Why are you ignoring the empty case? If you want to clear the text field then?

deorukhkarchaitanya: You're right. Let me rectify this.

deorukhkarchaitanya:
This PR blew up _a little_ but we do need to support all these use-cases on the app and need these props

src/atoms/TextInput/TextInput.native.js
```diff
@@ -249,6 +253,12 @@ const TextInput = ({
     [layoutDimensions, setLayoutDimensions],
   );
 
+  useEffect(() => {
+    if (isDefined(value)) {
+      setInput(value);
```
shubhnik, line 258: `setInput(value||'')`

It should be similar to `useState(value || '')`. Empty string for undefined, just like what RN TextInput does. For undefined values it is empty.

deorukhkarchaitanya: That's for input. This only happens when a value is defined or changed. Refer to the declaration of ```input```. That's where we assign default values

shubhnik: What I was suggesting is, whatever the new value be just set it(for undefined set it to '') and even if you set `undefined`, RN `TextInput` will be able to handle this. What say?

deorukhkarchaitanya: The idea is, that if someone is using it as a controlled TextInput then they _must_ provide value to the ```value``` prop. Providing value as undefined is as good as making it uncontrolled. If you want to clear the text on a controlled TextInput pass the value as ```value=""```

deorukhkarchaitanya: If it's made undefined then the value prop holds no value and the component should be assumed to be uncontrolled

src/atoms/TextInput/TextInput.stories.js
```diff
@@ -13,6 +13,29 @@ const sizeOptions = {
   auto: 'auto',
 };
 
+const keyboardTypeOptions = {
```
shridhar52war, line 16: Shoud we have some props specifics to native and web? I don't think web has this. Also the `returnKey` types

deorukhkarchaitanya: Yeah. Which is why it's grouped separately on the knobs under "Native". This is the best we can do for now to make it clear that these are Native only props.

shridhar52war: Oh Sooper!!

src/atoms/TextInput/__tests__/TextInput.native.test.js
```diff
@@ -227,4 +225,84 @@ describe('Renders <TextInput /> filled variant correctly', () => {
...
+      expect(container).toMatchSnapshot();
+    });
+    it('send', () => {
+      const { container } = renderWithTheme(<TextInput returnKeyType="send" />);
+      expect(container).toMatchSnapshot();
+    });
+  });
+
+  describe('autoCapitalize', () => {
+    it('none', () => {
```
shubhnik, line 291: I wonder so many tests are really helpful for these cases 🤔

src/atoms/TextInput/TextInput.native.js
```diff
@@ -448,6 +473,10 @@ TextInput.defaultProps = {
   testID: 'ds-text-input',
   labelPosition: 'top',
   width: 'medium',
+  secureTextEntry: false,
```
kamleshchandnani, line 476: can we accept a prop called `type` and then map internally to secureTextEntry? which then remains close to web as well?
so it'll be `type="text"` or `type="password"`

src/atoms/TextInput/TextInput.native.js
```diff
@@ -420,14 +435,24 @@ TextInput.propTypes = {
...
+  value: PropTypes.string,
   iconLeft: PropTypes.string,
   iconRight: PropTypes.string,
   maxLength: PropTypes.number,
   label: PropTypes.string,
   testID: PropTypes.string,
   labelPosition: PropTypes.oneOf(['top', 'left']),
   width: PropTypes.oneOf(['small', 'medium', 'auto']),
+  type: PropTypes.oneOf(['text', 'password']),
+  keyboardType: PropTypes.oneOf([
```
dhruvdutt, line 447: Can we derive this based on `type`? Web won't have any such prop.

src/atoms/TextInput/TextInput.native.js
```diff
@@ -420,14 +435,24 @@ TextInput.propTypes = {
...
-  children: PropTypes.string,
+  value: PropTypes.string,
   iconLeft: PropTypes.string,
   iconRight: PropTypes.string,
   maxLength: PropTypes.number,
   label: PropTypes.string,
   testID: PropTypes.string,
   labelPosition: PropTypes.oneOf(['top', 'left']),
   width: PropTypes.oneOf(['small', 'medium', 'auto']),
+  type: PropTypes.oneOf(['text', 'password']),
```
dhruvdutt, line 446: Can also include `email`, `number`.

PR #90, opened 2020-04-07
fix(Text/native): adds text-align prop

shridhar52war:


src/atoms/Text/Text.native.js
```diff
@@ -32,6 +32,12 @@ const styles = {
       return getLineHeight(theme, size);
     }
   },
+  textAlign({ align }) {
+    if (align == 'inherit' || align == 'initial') {
```
deorukhkarchaitanya, line 36: Shouldn't we use ```===```?
```suggestion
    if (align === 'inherit' || align === 'initial') {
```

dhruvdutt: Don't we have this check as eslint rule? 🤦‍♂

src/atoms/Text/Text.native.js
```diff
@@ -32,6 +32,12 @@ const styles = {
...
+      return 'auto';
```
dhruvdutt, line 37: Is this because React Native doesn't have `inherit` or `initial`?

shridhar52war: RN supports `'auto', 'left', 'right', 'center', 'justify' `

src/atoms/Text/Text.native.js
kamleshchandnani: We do. Something wrong with Shridhar's machine. I guess I need to now run linters in CI as well

PR #93, opened 2020-04-07
feat(Icon/react): add component

krishnaacc123:
Add Icon component


src/atoms/Icon/Icon.web.js
```diff
@@ -1,3 +1,41 @@
...
+
+const Icon = ({ size, name, fill, testID, ...rest }) => {
+  const IconComponent = icons[name];
+  const theme = useContext(ThemeContext);
+  return (
+    <IconComponent
+      fill={getColor(theme, fill)}
+      height={iconSize[size]}
+      width={iconSize[size]}
+      test-id={testID}
```
deorukhkarchaitanya, line 22: use ```automation-attributes```

src/atoms/Icon/__tests__/Icon.web.test.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { renderWithTheme } from '../../../_helpers/testing';
+import Icon from '../index';
+
+describe('<Icon />', () => {
+  describe('size', () => {
```
deorukhkarchaitanya, line 6: Add one for ```fill```?

krishnaacc123: added

src/atoms/Icon/Icon.web.js
```diff
@@ -1,3 +1,42 @@
...
+import { getColorKeys, getColor } from '../../_helpers/theme';
+
+const iconSize = {
+  xsmall: 12,
+  small: 16,
+  medium: 20,
+  large: 24,
+};
+
+const Icon = ({ size, name, fill, testID, ...rest }) => {
```
kamleshchandnani, line 15: Shouldn’t this be data-testid? testID is for native

krishnaacc123: @kamleshchandnani 
the attribute returned from automation method will be `data-testid` but the param name is still testID.

kamleshchandnani: Sure. so in that case we don't need to expose this prop. This is required for react-native

for web we can just use `{...automation('ds-icon')}` and you can remove this from prop-types and everywhere

src/atoms/Icon/Icon.web.js
krishnaacc123: added

PR #97, opened 2020-04-13
fix(Icon/react): update info icon component

krishnaacc123:


src/icons/Info/Info.web.js
```diff
@@ -2,25 +2,8 @@ import React from 'react';
 
 function Info(props) {
```
dhruvdutt, line 3: `Info` is just an internal component. Won't be part of commit message header.
~~`fix(Info/react)`~~ `fix(Icon/react)`

krishnaacc123: Sure

PR #100, opened 2020-04-14
chore(ci): create github workflow

dhruvdutt:


deorukhkarchaitanya:
Why are we doing this?

dhruvdutt:
@deorukhkarchaitanya Just for running tests for now. We'll add eslint, stylelint, commitlint.

deorukhkarchaitanya:
>@deorukhkarchaitanya Just for running tests for now. We'll add eslint, stylelint, commitlint.

Aren't we doing tests on ZEIT anyway?

dhruvdutt:
In Zeit, there is no visible in terms of steps (like tests, build, lint) of what all things are happening.  There's only one giant log which mostly has all build output. It's not possible to figure out what all ran. The config of what npm scripts are running is not in the repo. Again no visibility of what is happening.

kamleshchandnani:
Valid points. In that case we can switch to [Zeit GitHub action](https://github.com/marketplace/actions/zeit-now-deployment)
Because the test and build needs to be in sync/same platform else if tests fail how will zeit know. To make it work we need to run the tests in both zeit and Actions. So might as well switch to Zeit's official GitHub action

kamleshchandnani:
Closing this. We'll move to zeit GH action.

PR #102, opened 2020-04-15
chore(package.json): move react-native-svg to peer and dev dependencies

shridhar52war:


deorukhkarchaitanya:
Why did the snapshots that include svg changed?

shridhar52war:
> Why did the snapshots that include svg changed?

Version got upgraded, so the props got changed.

shridhar52war:
> Why did the snapshots that include svg changed?

https://github.com/react-native-community/react-native-svg/commit/363c1b491ff035b5f4c2666f7cacd6665029d2a8, 
Here in this commit, the properties inherited by the child from children are applied directly to the Group element. May be because of this the snaphots got affected.

PR #103, opened 2020-04-16
feat(Position/native): add component

shridhar52war:


src/atoms/Position/Position.native.js
```diff
@@ -0,0 +1,41 @@
...
+  top: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  right: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  bottom: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  left: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  zIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+};
+
+Position.defaultProps = {
+  position: 'relative',
+  top: undefined,
```
kamleshchandnani, line 34: We actually don’t need undefined in defaultProps I have realised it just yesterday. Because if you don’t put the props in default it’s already undefined. We’ll clear other components as well. But in this you can remove

src/atoms/Position/Position.native.js
```diff
@@ -0,0 +1,36 @@
...
+  ${(props) => (props.position ? `position: ${props.position};` : '')};
+  ${(props) => (props.top ? `top: ${makePxValue(props.top)};` : '')};
+  ${(props) => (props.right ? `right: ${makePxValue(props.right)};` : '')};
+  ${(props) => (props.bottom ? `bottom: ${makePxValue(props.bottom)};` : '')};
+  ${(props) => (props.left ? `left: ${makePxValue(props.left)};` : '')};
+  ${(props) => (props.zIndex ? `z-index: ${props.zIndex};` : '')};
+`;
+
+Position.propTypes = {
+  position: PropTypes.oneOf(['relative', 'absolute']),
```
kamleshchandnani, line 24: One of - static | relative | absolute | sticky | fixed

shridhar52war: For native it should be only `relative` and `absolute`

PR #107, opened 2020-04-18
fix(storybook): add fonts to storybook head

kamleshchandnani:


storybook/web/webpack.config.js
```diff
@@ -1,5 +1,27 @@
...
+          name: 'fonts/[name].[ext]',
+        },
+      },
+    ],
+  });
+  config.plugins.push(
+    new CopyWebpackPlugin([
+      {
+        from: 'public/fonts',
+        to: 'fonts',
```
dhruvdutt, line 21: Why do we need this? Can't we directly do `url('public/fonts/Lato-Regular.ttf')`

src/atoms/Flex/Flex.stories.js
```diff
@@ -1,87 +1,113 @@
...
+const alignItemsOptions = {
+  baseline: 'baseline',
+  center: 'center',
+  'flex-end': 'flex-end',
+  'flex-start': 'flex-start',
+  stretch: 'stretch',
+};
+
+const StyledView = styled(View)`
+  background-color: ${(props) => props.theme.colors.shade[940]};
```
dhruvdutt, line 44: Stories look great now. 💯 🙌

storybook/web/webpack.config.js
kamleshchandnani: Because the folder that’s published is `docs` hence it won’t resolve the url. Or I have to publish public as well which is incorrect. And that’s were copywebpack plugin comes in and meant for such use cases.

PR #108, opened 2020-04-23
feat(Icon/react): add Check Icon

krishnaacc123:


src/icons/Check/Check.web.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { getThemeColors } from '../../_helpers/theme';
+import colors from '../../tokens/colors';
+
+function Check({ width, height, fill }) {
+  return (
+    <svg
```
dhruvdutt, line 8: Add `Check.native.js` as well.
Look other Icons. You would just have to use `Svg` and `Path` components.

src/icons/index.js
```diff
@@ -11,6 +11,7 @@ const icons = {
   arrowRight: require('./ArrowRight').default,
   arrowLeft: require('./ArrowLeft').default,
   close: require('./Close').default,
+  check: require('./Check').default,
```
dhruvdutt, line 14: Can you please sort all properties?

src/icons/Check/Check.web.js
krishnaacc123: added

PR #112, opened 2020-04-27
fix(Switch/native): moves defaultOn to a separate useEffect

shubhnik:


src/atoms/Switch/Switch.native.js
```diff
@@ -189,6 +189,20 @@ const Switch = ({ disabled, on, defaultOn, onChange, size, testID }) => {
     new Animated.Value(animationConfig.off.rightSpace),
   );
 
+  useEffect(() => {
+    if (defaultOn) {
```
kamleshchandnani, line 193: why was this required?

shubhnik: This was for `defaultOn`. You did the eslint formating thing, which introduced the bug as I hadn't added all the dependencies in the useEffect.

PR #130, opened 2020-05-11
feat(Icon): adds copy icon

shubhnik:


src/icons/Copy/Copy.native.js
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import Svg, { Path } from 'react-native-svg';
+import { getThemeColors } from '../../_helpers/theme';
+import colors from '../../tokens/colors';
+
+function Copy({ width, height, fill }) {
+  return (
+    <Svg width={width} height={height} viewBox="0 0 22 22" fill="none">
```
chaitanyadeorukhkar, line 9: viewbox should be "0 0 24 24"

You might have picked a wrong svg

chaitanyadeorukhkar: Don't change it from 22 to 24 here. Make sure you pick up the correct svg that has 24x24 because the path will also change based on that

shubhnik: I picked it from that icon sheet in Figma. All are of varying sizes 🤔

kamleshchandnani: Viewbox should be 24x24 and then the icon could be of any size i.e defined by path

src/icons/Copy/Copy.web.js
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import { getThemeColors } from '../../_helpers/theme';
+import colors from '../../tokens/colors';
+
+function Copy({ width, height, fill }) {
+  return (
+    <svg width={width} height={height} viewBox="0 0 22 22" fill="none">
```
kamleshchandnani, line 8: change it here as well

PR #134, opened 2020-05-12
fix(native): fix snapshots for Tabs

chaitanyadeorukhkar:
onIndexChanged returns a function instead of a number while running it in jest. This commit ensures that tab is changed only if the callback index is a number.

### DIFF
--- a/src/molecules/Tabs/Tabs.native.js
+++ b/src/molecules/Tabs/Tabs.native.js
@@ -76,10 +76,12 @@ const Tabs = ({ children, defaultValue, value, onChange }) => {
 
   const onIndexChange = useCallback(
     (newIndex) => {
-      if (!isDefined(value)) {
-        setIndex(newIndex);
+      if (typeof newIndex === 'number') {
+        if (!isDefined(value)) {
+          setIndex(newIndex);
+        }
+        if (onChange) onChange(getRouteValueFromIndex({ routes, index: newIndex }));
       }
-      if (onChange) onChange(getRouteValueFromIndex({ routes, index: newIndex }));
     },
     [onChange, value, routes],
   );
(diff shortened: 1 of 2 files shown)


src/molecules/Tabs/Tabs.native.js
```diff
@@ -76,10 +76,12 @@ const Tabs = ({ children, defaultValue, value, onChange }) => {
 
   const onIndexChange = useCallback(
     (newIndex) => {
-      if (!isDefined(value)) {
-        setIndex(newIndex);
+      if (typeof newIndex === 'number') {
+        if (!isDefined(value)) {
```
dhruvdutt, line 80: Use `isEmpty`.

chaitanyadeorukhkar: Won't it be better if I do a check for number since I _want_ a number?

dhruvdutt: You can keep that, I'm asking to use `isEmpty` instead of `isDefined`.

kamleshchandnani: IsDefined and isEmpty are different

src/molecules/Tabs/Tabs.native.js
```diff
@@ -76,10 +76,12 @@ const Tabs = ({ children, defaultValue, value, onChange }) => {
 
   const onIndexChange = useCallback(
     (newIndex) => {
-      if (!isDefined(value)) {
-        setIndex(newIndex);
+      if (typeof newIndex === 'number') {
```
kamleshchandnani, line 79: why typeOf check?

chaitanyadeorukhkar: During jest tests it returns an object with a function in the callback instead of an index. Couldn't identify what the problem is with the library but this adds a safeguard so that even during tests the components work as intended

PR #135, opened 2020-05-12
fix(Link/react): pass props

krishnaacc123:


chaitanyadeorukhkar:
We shouldn't do this. This opens up the possibility of passing any props including style props that will break all the design restrictions

krishnaacc123:
@chaitanyadeorukhkar But then how will we make custom components out of this with particular css at the consumer side.

src/atoms/Link/Link.web.js
```diff
@@ -50,7 +50,7 @@ const StyledText = styled(Text)`
   }
 `;
 
-const Link = ({ size, children, href, target, testID, rel, disabled }) => {
+const Link = ({ size, children, href, target, testID, rel, disabled, ...props }) => {
```
kamleshchandnani, line 53: Give some reasoning behind this?

krishnaacc123: We are using custom components out of this with particular css at the consumer side.

kamleshchandnani:
As discussed closing this

PR #140, opened 2020-05-19
feat(Icon): add new icons

chaitanyadeorukhkar:


src/icons/LifeBuoy/LifeBuoy.native.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
kamleshchandnani, line 1: LifeBuoy?? 🤣

chaitanyadeorukhkar: Picked it from the figma file. Didn't know what else to call it 😅

src/icons/Settings/Settings.native.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
kamleshchandnani, line 1: Shouldn't this be called `Gear` Icon ?

chaitanyadeorukhkar: It's `settings` on figma. Even I was confused. Should I change it to gear?

PR #142, opened 2020-05-21
refactor(theme): rename theme.web to theme

dhruvdutt:


kamleshchandnani:
@dhruvdutt tests are failing. you need to update snapshots

src/tokens/theme.native.js
```diff
@@ -1,5 +1,5 @@
 import { fonts } from '../tokens';
-import { lightTheme as baseLightTheme, darkTheme as baseDarkTheme } from './theme.web';
+import { lightTheme as baseLightTheme, darkTheme as baseDarkTheme } from './theme';
```
chaitanyadeorukhkar, line 2: React native will pick up theme.native.js :P

This is why bitrise is failing

chaitanyadeorukhkar: For react native's bundler an import from `./theme` is as good as `./theme.native`

dhruvdutt: But it should fallback to pick `theme.js` if `theme.native.js` is not present. 🤔

chaitanyadeorukhkar: We're changing `theme.web.js` to `theme.js` but there is still a file in the same location called `theme.native.js`. It's picking that up instead of `theme.js`

kamleshchandnani: Oh damn yes. That was the reason we had kept it `theme.web.js` and `theme.native.js` now I remember @dhruvdutt . This is causing recursion

kamleshchandnani: The other option we had evaluated was to have a file called `baseTheme.js` and then let native and web import from them and override anything specific. But again right now we didn't had any overrides as such so thought not to add one more unnecessary file.

kamleshchandnani:
Closing this based on this discussion
https://github.com/razorpay/blade/pull/142#discussion_r432812830

PR #144, opened 2020-05-22
feat(native): add Snackbar component

chaitanyadeorukhkar:


chaitanyadeorukhkar:
@kamleshchandnani can you give me access to Vercel. Idk why this is failing

src/molecules/Snackbar/__tests__/useSnackbar.native.test.js
```diff
@@ -0,0 +1,55 @@
...
+  </ThemeProvider>
+);
+
+Wrapper.propTypes = {
+  children: PropTypes.node,
+};
+
+describe('useSnackbar hook', () => {
+  it('show the snackbar', () => {
+    const { result } = renderHook(() => useSnackbar(), { wrapper: Wrapper });
```
kamleshchandnani, line 21: beautiful ❤️

src/molecules/Snackbar/SnackbarProvider.native.js
```diff
@@ -0,0 +1,113 @@
...
+          autoDismiss,
+          iconName,
+          position,
+        });
+      }
+    },
+    [invokeNewSnackBar, dismiss, isVisible],
+  );
+
+  const snackbarActions = React.useMemo(() => ({ show, dismiss, isVisible }), [
```
kamleshchandnani, line 95: You don't need memo here since you're anyways using functions which are memoised using `useCallback` so this can just be a simple object

src/molecules/Snackbar/SnackbarProvider.native.js
```diff
@@ -0,0 +1,113 @@
...
+  const [isVisible, setIsVisible] = useState(false);
+  const timerRef = useRef(null);
+  const dismissAfterDuration = useCallback((duration = DEFAULT_DISMISS_DURATION) => {
+    if (timerRef.current) {
+      clearTimeout(timerRef.current); // clear existing timer
+    }
+    timerRef.current = setTimeout(() => setIsVisible(false), duration);
+  }, []);
+
+  const dismiss = useCallback(() => {
```
kamleshchandnani, line 28: `close`

src/molecules/Snackbar/SnackbarProvider.native.js
```diff
@@ -0,0 +1,113 @@
...
+
+  const snackbarActions = React.useMemo(() => ({ show, dismiss, isVisible }), [
+    show,
+    dismiss,
+    isVisible,
+  ]);
+
+  return (
+    <SnackbarContext.Provider value={snackbarActions}>
+      {children}
```
kamleshchandnani, line 103: why these are adjacent? Children and Snackbar?

src/molecules/Snackbar/Snackbar.native.js
```diff
@@ -0,0 +1,205 @@
...
+  }, []);
+
+  useEffect(() => {
+    if (isVisible) {
+      Animated.timing(visibility, {
+        toValue: animationConfig.animationValue.final,
+        useNativeDriver: true,
+        duration: animationConfig.duration,
+      }).start();
+    } else if (!isVisible) {
```
kamleshchandnani, line 84: why `else if`? just `else` won't work?

src/molecules/Snackbar/Snackbar.native.js
```diff
@@ -0,0 +1,205 @@
...
+    top: PropTypes.number,
+    right: PropTypes.number,
+    bottom: PropTypes.number,
+    left: PropTypes.number,
+  }),
+};
+
+Snackbar.defaultProps = {
+  variant: 'positive',
+  position: {
```
kamleshchandnani, line 197: just don't put default for `position` it'll be undefined

src/molecules/Snackbar/SnackbarProvider.native.js
```diff
@@ -0,0 +1,113 @@
...
+      clearTimeout(timerRef.current); // clear existing timer
+    }
+    if (isVisible) {
+      setIsVisible(false);
+    }
+  }, [isVisible]);
+
+  const invokeNewSnackBar = useCallback(
+    (config) => {
+      setSnackbarProps(config);
```
kamleshchandnani, line 39: Why do you need another setState for snackbar? the props you need to pass to snackbar cab just be plain object and when you toggle isVisible anyways it'll toggle the behavior

src/molecules/Snackbar/SnackbarProvider.native.js
```diff
@@ -0,0 +1,113 @@
...
+    actionText: undefined,
+    onAction: undefined,
+    showDismissButton: undefined,
+    onDismiss: undefined,
+    maxLines: undefined,
+    position: { top: undefined, bottom: 0, left: 0, right: 0 },
+  });
+  const [isVisible, setIsVisible] = useState(false);
+  const timerRef = useRef(null);
+  const dismissAfterDuration = useCallback((duration = DEFAULT_DISMISS_DURATION) => {
```
kamleshchandnani, line 21: `hideAfterMs`

src/molecules/Snackbar/Snackbar.stories.js
```diff
@@ -0,0 +1,93 @@
...
+    <Flex flex={1}>
+      <View>
+        <SnackWrapper
+          variant={select('Variant', variantOptions, undefined)}
+          text={text('Text', 'Snackbar text here')}
+          maxLines={number('maxLines', undefined)}
+          actionText={text('Action Text', 'Retry')}
+          onAction={() => {}}
+          onDismiss={() => {}}
+          showDismissButton={boolean('showDismissButton', true)}
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: We might not necessarily need an onClose, the user might just want to give a close button to close the snackbar but not take any action once it's closed right?

src/molecules/Snackbar/SnackbarProvider.native.js
chaitanyadeorukhkar: Because we'll be wrapping the whole app in the snackbar provider. The app will be children and it'll have Snackbar as a globally available component

src/molecules/Snackbar/SnackbarProvider.native.js
chaitanyadeorukhkar: Snackbar is already rendered at the top of the app, it's only hidden using state. Now we need to change the state so that we can make Snackbar visible & update it's props

src/molecules/Snackbar/Snackbar.stories.js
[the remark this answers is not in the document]
kamleshchandnani: Not sure why would you do that? Even in modals we show the close button when the user have provided `onClose` function. Can you think of a use case?

src/molecules/Snackbar/SnackbarProvider.native.js
kamleshchandnani: You mean you'll always render the snackbar in the DOM? If yes then it's incorrect. if the snackbar is hidden then it should not be rendered at all the provider should return `null` basically in that case

src/molecules/Snackbar/Snackbar.stories.js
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah, most places in our app the snackbar should autohide & also have a close button which means the user can close it whenever they want before the hide timeout. Here we don't want to do anything when the user hides the snackbar but we just need to allow them the ability to dismiss it before timeout

src/molecules/Snackbar/Snackbar.stories.js
```diff
@@ -0,0 +1,86 @@
...
+  warning: 'warning',
+  neutral: 'neutral',
+};
+
+const iconOptions = getIconNames().reduce(
+  (options, option) => ({ ...options, [option]: option }),
+  {},
+);
+
+const SnackDemo = (props) => {
```
kamleshchandnani, line 24:
```suggestion
const SnackbarDemo = (props) => {
```

src/molecules/Snackbar/SnackbarProvider.native.js
```diff
@@ -0,0 +1,109 @@
+import React, { useState, useCallback, useRef } from 'react';
+import PropTypes from 'prop-types';
+import Snackbar from './Snackbar';
+import SnackbarContext from './SnackbarContext';
+
+const DEFAULT_CLOSE_DURATION = 3000;
```
kamleshchandnani, line 6:
```suggestion
const DEFAULT_HIDE_DURATION = 3000;
```

src/molecules/Snackbar/SnackbarProvider.native.js
chaitanyadeorukhkar: If I return null when the user dismisses it, then it won't have that hide animation. 

Does this mean that I have to maintain another boolean state similar to `isVisible` that will return the snackbar or null and will be turned to false when exit animation is complete?

src/molecules/Snackbar/Snackbar.stories.js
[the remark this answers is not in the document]
kamleshchandnani: So let's keep it to `onClose` for now and we'll add this prop once we see a use case?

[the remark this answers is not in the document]
chaitanyadeorukhkar: Cool cool

PR #148, opened 2020-05-29
feat(native/Modal): add component

shubhnik:


src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,246 @@
...
+  showDragBar: PropTypes.bool,
+};
+
+Modal.defaultProps = {
+  type: 'centered',
+  visible: false,
+  showDragBar: false,
+};
+
+ModalHeader.propTypes = {
```
kamleshchandnani, line 217: Should be a separate file

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,246 @@
...
+  type: PropTypes.oneOf(['bottomsheet', 'centered', 'fullscreen']),
+  onClose: PropTypes.func,
+};
+
+ModalHeader.defaultProps = {
+  type: 'centered',
+  onClose: () => {},
+};
+
+ModalContent.propTypes = {
```
kamleshchandnani, line 228: Should be a separate file

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,246 @@
...
+  children: PropTypes.node.isRequired,
+  type: PropTypes.oneOf(['bottomsheet', 'centered', 'fullscreen']),
+};
+
+ModalContent.defaultProps = {
+  type: 'centered',
+};
+
+ModalFooter.propTypes = {
```
kamleshchandnani, line 237: should be a separate file

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,246 @@
...
+
+Modal.propTypes = {
+  children: PropTypes.node.isRequired,
+  type: PropTypes.oneOf(['bottomsheet', 'centered', 'fullscreen']),
+  visible: PropTypes.bool,
+  onClose: PropTypes.func,
+  onBackdropClick: PropTypes.func,
+  onSwipeComplete: PropTypes.func,
+  onBackButtonPress: PropTypes.func,
+  showDragBar: PropTypes.bool,
```
kamleshchandnani, line 208: This shouldn't be a separate prop. How will you control 
```
showDragbar={true}
type="fullscreen"
```
dragbar is derived based on the type of Modal

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,246 @@
...
+Modal.Content = ModalContent;
+Modal.Footer = ModalFooter;
+
+Modal.propTypes = {
+  children: PropTypes.node.isRequired,
+  type: PropTypes.oneOf(['bottomsheet', 'centered', 'fullscreen']),
+  visible: PropTypes.bool,
+  onClose: PropTypes.func,
+  onBackdropClick: PropTypes.func,
+  onSwipeComplete: PropTypes.func,
```
kamleshchandnani, line 206: why do you need `onSwipeComplete` and  `onBackButtonPress` prop? Why can't it be same as `onClose`?

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,246 @@
...
+        onBackdropPress={onBackdropClick}
+        onBackButtonPress={onBackButtonPress}
+      >
+        <CenteredModalBody>
+          {onClose ? (
+            <Flex alignSelf="flex-end">
+              <Space padding={[0.5, 1, 0, 0]}>
+                <View>
+                  <Button
+                    testID="centered-cross"
```
kamleshchandnani, line 134:
```suggestion
                    testID="close-button"
```

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,246 @@
...
+const CenteredModalBody = styled(View)`
+  background-color: ${(props) => props.theme.colors.background[100]};
+  border-radius: ${(props) => props.theme.spacings.xsmall};
+`;
+
+const FullscreenModalBody = styled(View)`
+  background-color: ${(props) => props.theme.colors.background[400]};
+`;
+
+const BottomsheetModalBody = styled(View)`
```
kamleshchandnani, line 20: why can't this be a ModalContentContainer and then get the styles basis the type? The way it's done here you'll end up creating 3 views in the memory but use just one at a time.

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,246 @@
...
+`;
+
+const Divider = styled(View)`
+  background-color: ${(props) => props.theme.colors.shade[920]};
+  height: 1px;
+  width: 100%;
+`;
+
+const ModalHeader = ({ children, type, onClose }) => {
+  return type === 'centered' ? (
```
kamleshchandnani, line 52: please avoid using nested tertiary

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,246 @@
...
+    <Space padding={[1.5, 2, 0.5, 2]}>
+      <View>{children}</View>
+    </Space>
+  ) : type === 'fullscreen' ? (
+    <Flex flexDirection="row" justifyContent="space-between">
+      <Space padding={[1, 2]}>
+        <FullScreenModalHeader>
+          {children}
+          <Button
+            testID="fullscreen-cross"
```
kamleshchandnani, line 62:
```suggestion
            testID="close-button"
```
Why do you need different id for different types of close? at a time only one variant will be open right?

src/molecules/Modal/Modal.native.js
shubhnik: In the above case, the dragbar won't be shown. 

DragBar will only be shown in the Bottomsheet Modal. In Figma, they have Bottomsheet without dragBar.

Maybe, I should rename it to `showBottomsheetDragbar` to avoid confusion?

src/molecules/Modal/Modal.native.js
shubhnik: `onSwipeComplete` and `onBackButtonPress` are the props exposed by `react-native-modal`. Do you mean to say that `onSwipeComplete` and `onSwipeComplete` would always close the modal?

shubhnik: `onSwipeComplete` and `onBackButtonPress` are the props exposed by `react-native-modal`. Did you mean that `onSwipeComplete` and `onSwipeComplete` would always close the modal?

src/molecules/Modal/Modal.native.js
kamleshchandnani: Whutt. We'll ask them to have the dragbar in all the bottomsheet.

src/molecules/Modal/Modal.native.js
kamleshchandnani: I'm saying pass the same function you get in `onClose` to `onSwipeComplete`, `onBackButtonPress`. Just because we are using third party libs doesn't mean that we should expose their props to our consumers. Also the use case you're mentioning I don't see a probability of someone doing that behavior.

storybook/native/main.js
```diff
@@ -32,6 +32,7 @@ configure(() => {
   require('../../src/molecules/Amount/Amount.stories');
   require('../../src/molecules/SegmentControl/SegmentControl.stories');
   require('../../src/molecules/Tabs/Tabs.stories');
+  require('../../src/molecules/Modal/Modal.stories');
```
kamleshchandnani, line 35: sort this

src/molecules/Modal/ModalHeader.native.js
```diff
@@ -0,0 +1,67 @@
...
+            icon="close"
+            size="medium"
+            onClick={onClose}
+          />
+        </FullScreenModalHeader>
+      </Space>
+    </Flex>
+  ) : type === 'bottomsheet' ? (
+    <React.Fragment>
+      <Space>
```
kamleshchandnani, line 46: Space without any props is equivalent of not having it at all

src/molecules/Modal/ModalHeader.native.js
```diff
@@ -0,0 +1,67 @@
...
+        <View>{children}</View>
+      </Space>
+      <Size height="1px" width="100%">
```
kamleshchandnani, line 49: you can't make this assumption. This will fail in our app use case only. The status bottomsheet doesn't has a divider

src/molecules/Modal/ModalHeader.native.js
```diff
@@ -0,0 +1,67 @@
...
+import styled from 'styled-components/native';
+import PropTypes from 'prop-types';
+import Space from '../../atoms/Space';
+import Size from '../../atoms/Size';
+import Flex from '../../atoms/Flex';
+import View from '../../atoms/View';
+import Button from '../../atoms/Button';
+
+const FullScreenModalHeader = styled(View)`
+  background-color: ${(props) => props.theme.colors.background[200]};
```
kamleshchandnani, line 11: why can't you use our card and override just the border-radius?

src/molecules/Modal/ModalHeader.native.js
```diff
@@ -0,0 +1,67 @@
...
+  shadow-color: ${(props) => props.theme.colors.primary[930]};
+  shadow-offset: 0px 4px;
+  shadow-opacity: 1;
+  shadow-radius: 15px;
+  elevation: 2;
+`;
+
+const Divider = styled(View)`
```
kamleshchandnani, line 19: Create Divider as a separate component in `atoms` same way we did in app.

src/molecules/Modal/Modal.native.js
shubhnik: I will remove it, my only concern was the use case like those.

src/molecules/Modal/ModalFooter.native.js
```diff
@@ -0,0 +1,38 @@
...
+
+const ModalFooter = ({ children, type }) => {
+  return type === 'centered' ? (
+    <Space padding={[2, 3]}>
+      <View>{children}</View>
+    </Space>
+  ) : type === 'fullscreen' ? null : type === 'bottomsheet' ? (
+    <React.Fragment>
+      <Size height="1px" width="100%">
+        <Divider />
```
kamleshchandnani, line 20: Again same assumption. Confirm with designers what do they really want

src/molecules/Modal/ModalHeader.native.js
```diff
@@ -0,0 +1,67 @@
...
+  background-color: ${(props) => props.theme.colors.shade[920]};
+`;
+
+const ModalHeader = ({ children, type, onClose }) => {
+  return type === 'centered' ? (
```
kamleshchandnani, line 24: why can't you put conditions here and return from that instead of so many tertiary operators?

src/molecules/Modal/ModalFooter.native.js
```diff
@@ -0,0 +1,38 @@
...
+import Space from '../../atoms/Space';
+import Size from '../../atoms/Size';
+import View from '../../atoms/View';
+
+const Divider = styled(View)`
+  background-color: ${(props) => props.theme.colors.shade[920]};
+`;
+
+const ModalFooter = ({ children, type }) => {
+  return type === 'centered' ? (
```
kamleshchandnani, line 13: remove nested tertiary and move to plain `if(conditon) { return }`

src/molecules/Modal/ModalContent.native.js
```diff
@@ -0,0 +1,27 @@
...
+import View from '../../atoms/View';
+
+const ModalContent = ({ children, type }) => {
+  return type === 'centered' ? (
+    <Space padding={[2]}>
+      <View>{children}</View>
+    </Space>
+  ) : type === 'fullscreen' ? (
+    children
+  ) : type === 'bottomsheet' ? (
```
kamleshchandnani, line 13: confirm with Nishant once if this the intended behavior. As per my understanding the centerd and bottomsheet variant content should have fixed surrounding padding

src/molecules/Modal/ModalHeader.native.js
shridhar52war, line 46: `Space` is not required here?

src/molecules/Modal/ModalHeader.native.js
shubhnik: Can you point me to the design where the footer doesn't have the divider, I can't find one? And if that's the case then it should be DS compliant?

src/molecules/Modal/ModalHeader.native.js
shubhnik: I honestly feel that the card is not the correct component to use here. Further, the shadow constructs are not the same and nor can I override the card border.

src/atoms/Divider/Divider.native.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import styled from 'styled-components/native';
+import PropTypes from 'prop-types';
+import View from '../View';
+import { getColor, getColorKeys } from '../../_helpers/theme';
+
+const StyledView = styled(View)`
```
kamleshchandnani, line 7:
```suggestion
const Divider = styled(View)`
```

src/atoms/Divider/Divider.native.js
```diff
@@ -0,0 +1,25 @@
...
+  height: ${(props) => (props.dividerDirection === 'horizontal' ? '1px' : '100%')};
+  width: ${(props) => (props.dividerDirection === 'horizontal' ? '100%' : '1px')};
+  background-color: ${(props) => getColor(props.theme, props.color)};
+`;
+
+const Divider = ({ direction, color }) => <StyledView dividerDirection={direction} color={color} />;
```
kamleshchandnani, line 13: you don't need this. A styled component becomes a React compinent in the end. So it'll get all the props

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,157 @@
...
+  return renderModal();
+};
+
+Modal.Header = ModalHeader;
+Modal.Content = ModalContent;
+Modal.Footer = ModalFooter;
+
+Modal.propTypes = {
+  children: PropTypes.node.isRequired,
+  type: PropTypes.oneOf(['bottomsheet', 'centered', 'fullscreen']),
```
kamleshchandnani, line 146: I missed this somehow. It should be `variant` basis the convention we follow across
```suggestion
  variant: PropTypes.oneOf(['bottomsheet', 'centered', 'fullscreen']),
```

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,157 @@
...
+      'shadow-offset': '0px -4px',
+      'shadow-opacity': '1',
+      'shadow-radius': '15px',
+      elevation: '2',
+    };
+  }
+  return null;
+};
+
+const ModalContentContainer = styled(View)(({ type, theme }) => styles({ type, theme }));
```
kamleshchandnani, line 42:
```suggestion
const ModalContainer = styled(View)(({ type, theme }) => styles({ type, theme }));
```

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,157 @@
...
+};
+
+Modal.Header = ModalHeader;
+Modal.Content = ModalContent;
+Modal.Footer = ModalFooter;
+
+Modal.propTypes = {
+  children: PropTypes.node.isRequired,
+  type: PropTypes.oneOf(['bottomsheet', 'centered', 'fullscreen']),
+  visible: PropTypes.bool,
```
kamleshchandnani, line 147:
```suggestion
  isVisible: PropTypes.bool,
```

src/molecules/Modal/Modal.native.js
```diff
@@ -0,0 +1,157 @@
...
+              </ModalContentContainer>
+            </RNModal>
+          </Flex>
+        </Space>
+      );
+    }
+    return null;
+  };
+
+  return renderModal();
```
kamleshchandnani, line 137: why do you need this function explicitly? directly `if` conditions would work

src/molecules/Modal/Modal.stories.js
```diff
@@ -0,0 +1,185 @@
...
+import Flex from '../../atoms/Flex';
+import Text from '../../atoms/Text';
+import Heading from '../../atoms/Heading';
+import Space from '../../atoms/Space';
+import Size from '../../atoms/Size';
+import View from '../../atoms/View';
+import Button from '../../atoms/Button';
+import Modal from './Modal';
+
+const HeaderIllustration = styled(View)`
```
kamleshchandnani, line 13:
```suggestion
const HeaderPlaceholder = styled(View)`
```

src/molecules/Modal/ModalFooter.native.js
```diff
@@ -0,0 +1,40 @@
...
+      </Space>
+    );
+  }
+  if (type === 'fullscreen') {
+    return null;
+  }
+  if (type === 'bottomsheet') {
+    return (
+      <React.Fragment>
+        <Divider color="shade.920" />
```
kamleshchandnani, line 21: Confirm once with Nishant about this. Not sure if this divider should exist in all the bottomsheet

src/molecules/Modal/ModalFooter.native.js
```diff
@@ -0,0 +1,40 @@
...
+
+const ModalFooter = ({ children, type }) => {
+  if (type === 'centered') {
+    return (
+      <Space padding={[2, 3]}>
+        <View>{children}</View>
+      </Space>
+    );
+  }
+  if (type === 'fullscreen') {
```
kamleshchandnani, line 15: Can you confirm the behavior with Nishant on this as well? All the modals follow a construct which means the footer in this case should also have somehting. It can't be null the API becomes unpredictable since I can write `Modal.Footer` in fullscreen as well and it'll keep returning null

src/molecules/Modal/ModalHeader.native.js
```diff
@@ -0,0 +1,68 @@
...
+          </FullScreenModalHeader>
+        </Space>
+      </Flex>
+    );
+  }
+  if (type === 'bottomsheet') {
+    return (
+      <React.Fragment>
+        <View>{children}</View>
+        <Divider color="shade.920" />
```
kamleshchandnani, line 50: Again confirm with Nishant. The bottomsheet doesn't shows divider always. It's upto the consumer. Look at the instances of bottomsheet in our app. All those should be moved to this component. So you should decide if this suffice all the use cases in the app as well if not raise it to designers

src/molecules/Modal/Modal.native.js
shubhnik: But we are following the current implementation only? `checked` and not `isChecked` in `Checkbox`, `on` and not `isOn` in `Switch`.

PR #154, opened 2020-06-13
feat(icon): adds wifi icons

shubhnik:


src/icons/Wifi/Wifi.native.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Svg, { Path, G, Defs, ClipPath } from 'react-native-svg';
+import { getThemeColors } from '../../_helpers/theme';
+import colors from '../../tokens/colors';
```
kamleshchandnani, line 5: rename this as `WiFiOn`?

src/icons/WifiOff/WifiOff.native.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
kamleshchandnani, line 1: Wi and Fi are different words so rename it as
`WiFiOff`

src/icons/Wifi/Wifi.native.js
shubhnik: Let's keep it WiFi only?, unless the icon explicitly expresses an action, like in WiFiOff?

PR #162, opened 2020-07-27
feat(native/TextArea): add forward ref, onSubmitEditing

shubhnik:


src/atoms/TextArea/TextArea.native.js
```diff
@@ -75,6 +83,7 @@ TextArea.propTypes = {
   testID: PropTypes.string,
   labelPosition: PropTypes.oneOf(['top', 'left']),
   width: PropTypes.oneOf(['small', 'medium', 'auto']),
+  onSubmitEditing: PropTypes.func,
```
chaitanyadeorukhkar, line 86: Should we call this `onSubmit` might be better aligned with other `onXYZ` of our APIs?

shubhnik: Basically this is called when you are done with "editing" and press the done button in keyboard and keyboard is dismissed, taken from react-native's own `onSubmitEditing` prop. `onSubmit` might sound vague here?

PR #163, opened 2020-07-29
feat(native/textArea): add signifier to TextArea

shubhnik:


shubhnik:
Screenshot 2020-07-29 at 1 56 36 PM
This PR adds a signifier to TextArea in the bottom right 👆

chaitanyadeorukhkar:
Without the signifier:
Screenshot 2020-07-31 at 2 43 14 PM

With the signifier:
Screenshot 2020-07-31 at 2 43 22 PM

Figma:
Screenshot 2020-07-31 at 2 44 41 PM


@shubhnik This doesn't look right. Maybe we need to make it position absolute but that will overlap with the text.  Or we shouldn't give it that much of a height. Need to find another way to implement this without creating that much gap

shubhnik:
Yeah in Figma they have that 8px View in which the signifier resides.

chaitanyadeorukhkar:
If you see the height of the TextInput & TextArea on figma, they're the same. Here we're adding additional 8px to the bottom. We shouldn't do that

src/atoms/TextInput/TextInput.native.js
```diff
@@ -408,9 +415,26 @@ const TextInput = React.forwardRef(
                         </InputContainer>
                       </Size>
                     </Flex>
+                    {_isMultiline ? (
+                      <Position position="absolute" bottom="1px">
+                        <Size height="8px" width="8px">
```
chaitanyadeorukhkar, line 420: Why not ?
```suggestion
                        <Size height={2} width={2}>
```

chaitanyadeorukhkar:
Closing this until we find a better approach for it @shubhnik

PR #165, opened 2020-07-30
feat(native/textInput): add autoFocus prop

shubhnik:


src/atoms/TextArea/TextArea.native.js
```diff
@@ -79,6 +82,7 @@ TextArea.propTypes = {
 
 TextArea.defaultProps = {
   placeholder: '',
+  autoFocus: false,
```
chaitanyadeorukhkar, line 85: Can we keep it as 
```suggestion
  autoFocus: undefined,
```

This way the existing snapshots won't change either. It's internally false if undefined anyway

PR #166, opened 2020-07-31
fix(tokens): line height for xsmall

dhruvdutt:


src/tokens/fonts.js
```diff
@@ -23,7 +23,7 @@ const fonts = {
     bold: '700',
   },
   lineHeight: {
-    xsmall: '14px',
+    xsmall: '18px',
```
kamleshchandnani, line 26: jaha use kr re hai waha `small` krna bhai why to break the scale?

PR #177, opened 2020-08-14
fix(native/bottomsheet): update content height based on footer

shridhar52war:
1. Calculates content height based on footer height when `adjustToContentHeight` prop is true. 
2. Disables the gradient effect when `adjustToContentHeight` is true.
3. Removes FooterComponent prop from test cases(my bad, wonder why did I leave it completely unnoticed!)

### DIFF
--- a/src/molecules/BottomSheet/BottomSheet.native.js
+++ b/src/molecules/BottomSheet/BottomSheet.native.js
@@ -80,6 +80,7 @@ const BottomSheet = ({
   const bottomSheetVisibility = useRef(visible);
   const [headerHeight, setHeaderHeight] = useState(0);
   const [contentHeight, setContentHeight] = useState(0);
+  const [footerHeight, setFooterHeight] = useState(0);
   const bottomsheetChildrenGroupByDisplayName = reactChildrenGroupByDisplayName(children);
 
   const headerComponent = bottomsheetChildrenGroupByDisplayName.BottomSheetHeader;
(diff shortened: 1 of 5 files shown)


src/molecules/BottomSheet/BottomSheet.native.js
```diff
@@ -119,11 +120,18 @@ const BottomSheet = ({
     setContentHeight(e.nativeEvent.layout.height);
   }, []);
 
+  const handleFooterLayoutChange = useCallback((e) => {
```
chaitanyadeorukhkar, line 123:
```suggestion
  const handleFooterLayoutChange = useCallback(({ nativeEvent }) => {
```

chaitanyadeorukhkar: @shridhar52war just this one change :P

src/molecules/BottomSheet/BottomSheetFooter.native.js
```diff
@@ -13,7 +13,11 @@ const BottomSheetFooter = ({ children }) => {
   return children ? (
     <FooterContainer>
       <Divider color="shade.920" horizontal />
-      {children ? <Space padding={[1, 2, 1, 2]}>{children}</Space> : null}
+      {children ? (
```
shubhnik, line 16: why 2 checks for `children` on lines 13 and 16? 🤔

PR #214, opened 2020-12-11
feat(tools): add figma version history plugin

saurav12:
Added version history plugin to tools folder.


packages/tools/ds-version-history/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
kamleshchandnani, line 1: rename the folder to `figma-plugin-version-history` since we'll be prefixing all the plugins with this so it'll be organised in a nice manner in the tree hierarchy in the directory structure

packages/tools/ds-version-history/package.json
```diff
@@ -0,0 +1,34 @@
+{
+	"name": "vue-plugin-template",
```
kamleshchandnani, line 2:
```suggestion
	"name": "@razorpay-blade-tools/version-history",
```

packages/tools/ds-version-history/LICENSE
saurav12: Done @kamleshchandnani!

packages/tools/ds-version-history/yarn.lock
```diff
@@ -0,0 +1,3950 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
kamleshchandnani, line 1: you don't need this file. aren't things automatically installing once you run yarn?

kamleshchandnani (changes requested):

packages/tools/ds-version-history/yarn.lock
saurav12: This was auto-generated when I ran yarn. Also, if we don't need this file, we can add it to `gitignore`.

PR #220, opened 2020-12-24
feat(theme): add dark theme colors

PujanShah22:


src/tokens/theme.web.js
```diff
@@ -164,11 +166,11 @@ const darkTheme = {
       980: colors.white[980],
     },
     dark: {
-      730: colors.slate[930],
-      740: colors.slate[940],
+      930: colors.slate[930],
```
kamleshchandnani, line 169: this seems like breaking change. Just re-validate if this doesn't breaks anything in components. Because light theme has 730 and 740. At any point your themes can't be out of sync

src/tokens/theme.web.js
```diff
@@ -93,67 +93,69 @@ const darkTheme = {
...
     },
     neutral: {
-      900: colors.honey[900],
-      930: colors.honey[930],
-      960: colors.honey[960],
+      900: colors.xHoney[900],
+      930: colors.xHoney[930],
+      960: colors.xHoney[960],
     },
     light: {
```
kamleshchandnani, line 158: the 800 again seems out of sync with the theme.

kamleshchandnani (changes requested):
Just cross check that both themes have same colors else things will randomly break

src/tokens/theme.web.js
PujanShah22: > this seems like breaking change. Just re-validate if this doesn't breaks anything in components. Because light theme has 730 and 740. At any point your themes can't be out of sync

Same way, Shade 500, 600 & 700 are added in dark theme while it's not there in light theme. 
I checked that there is no direct use of dark 730 and 740 in blade components.

Note: Whenever we will add dark & light themes in same project, we will need same aliases in both theme.

src/tokens/theme.web.js
PujanShah22: updated

kamleshchandnani (approved):

PR #221, opened 2020-12-24
feat(tokens): add colors

kamleshchandnani:
This PR adds the new color system tokens as part of our re-architecture exercise.


packages/blade/tokens/opacities.ts
```diff
@@ -0,0 +1,15 @@
+type opacity = Readonly<{
+  1: number;
```
chaitanyadeorukhkar, line 2: Why 1,2,3,9? Scale seems very weird to me looking at it without context.

packages/blade/tokens/colors.ts
```diff
@@ -0,0 +1,243 @@
+import opacities from './opacities';
```
chaitanyadeorukhkar, line 1:
```suggestion
import opacity from './opacity';
```

So that we can use `opacity[9]`


Colors made sense to me because colors have a range of colors and then a scale. Opacity directly has a scale and not a range of opacities. Otherwise it's like having colors.**neutrals**.**warmGrays**[100]

chaitanyadeorukhkar (changes requested):

packages/blade/tokens/colors.ts
kamleshchandnani: Even I was confused a lot while naming. So I was looking up patterns across. The reason why it's `opacities` is that it's a scale of opacity(a.k.a range) now if it's a collection or not doesn't matter. Similarly `fonts`, `fontSizes`, `aliases` etc.

packages/blade/tokens/opacities.ts
kamleshchandnani: The scale will be filled up. @saurav12 is working on it.

packages/blade/tokens/opacity.ts
```diff
@@ -0,0 +1,25 @@
...
+const opacities: opacity = {
+  1: 0.09,
+  2: 0.18,
+  3: 0.32,
+  4: 1.0,
+  5: 1.0,
+  6: 1.0,
+  7: 1.0,
+  8: 1.0,
+  9: 1.0,
```
chaitanyadeorukhkar, line 22: Hahahaha 😆

chaitanyadeorukhkar (approved):

PR #227, opened 2021-01-05
docs(rfc): add token naming convention

kamleshchandnani:
**[View Formatted RFC](https://github.com/razorpay/blade/blob/add/token-naming-rfc/rfcs/2021-01-04-tokens-naming-convention.md)**

If you any suggestions maybe start putting here in the comments by quoting the sections of the RFC


rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,406 @@
...
+Button.color.background 
+Button.color.border 
+Button.font.weight
+Button.font.size
+```
+
+#### Behavior
+Category-property pairs are exceedingly general and not purposefully useful when we want to create few generalised global tokens. 
+
+For example if we need to create a global token which applies to all the actionable elements i.e Button, Link etc. then we can't something general by just using category:property pairing(`color.text`) or if we want to have a token which applies to the `body` or maybe just elements which give feedbacks `snackbar`, `alerts` etc. Hence, we need another level in between which is `behavior`.
```
saurav12, line 190: Might be a spelling error:

> For example if we need to create a global token which applies to all the actionable elements i.e Button, Link etc. then `we can't something general by` just using category:property

Instead:

> For example if we need to create a global token which applies to all the actionable elements i.e Button, Link etc. then `we can't generalise something by` just using category:property

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,406 @@
...
+> Assume I've a Button component and I want to set the text color of a button when it's in hovered state. Now I don't know whether `Button.color.text.hover` is the right way or `Button.hover.color.text` is the right way
+
+As you can see with the current system there are a lot of issues with the naming structure since there's no guideline or a framework because of which we can't establish a standard way of naming things and the consumers of this design system can't understand what they are accessing at what level and what value they'll get at a particular level. Because of this our token naming and accessing them becomes unpredictable.
+
+### What is the expected outcome?
+- A well defined naming framework that we can use to name and access tokens at any level in our design system.
+- A structured system which makes accessing our tokens predictable.
+
+# Detailed Design
+Tokens are the core foundation of our design system since tokens are something
```
saurav12, line 110: This looks like an incomplete line towards the end.
`something...`

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,406 @@
...
+* `Button.font.size.primary.disabled.m` - says that this token points to the font size medium of a primary button when in disabled state.
+
+I had tried other ways to form a reasonably simpler structure but all of them broke as the complexity increased. Hence we needed a verbose framework which covers current as well as future scenarios.
+# Adoption strategy
+* Immediately, this won't impact any existing consumers of the design system.
+* This is being built in isolation(though merged into master regularly as we follow Trunk based development) and hence nothing will break on the projects using blade design system directly from git commit.
+* Since this is being built as an activity of re-architecture of our design system the migration of the old system to the new will be a breaking change which is out of the scope of this RFC and will be handled separately.
+
+# How do we educate people?
+* The best way to educate people is to put more and more visual representation of this staruture with contextual examples.
```
saurav12, line 397: We can add that visual image (updated) we made on Figma here.

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,406 @@
...
+
+But now if we want to refer a component within a component for example we need to refer to the `color` of an `icon` component inside a `Button` component then how shall we do that? An object can also split further and can have sub-component nesting to it as well. Eg: `button.leftIcon.color.background` 
+
+Similarly, if we want to refer to a token value within a theme like `font.family` we can't do it alone since it doesn't has any context attached to it and we don't even know if what are we referring to? a theme token or a component token?
+
+But as soon as we attach the context our problem will be solved. For eg: `theme.font.family` tells us that we are trying to refer to font family token inside our theme.
+
+#### TL;DR
+**Object** refers to
+  1. **component** - `button`, `theme` etc.
```
saurav12, line 141: Why are we not considering `namespace` for `themes`?
Like, differentiating `objects` & `namespace`.

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,406 @@
...
+#### Scale
+Scale serves as a mofifer to create more granular token name in scenarios where we want to create a token which work differently on different scales
+
+* Linear Scale: `1`, `2`, `3`, `4` and so on
+* T-shirt Size Scale: `s`, `m`, `l`, `xl` and so on
+
+For example: `theme.font.size.s`, `Button.space.paddingLeft.primary.hover.m`
+
+#### Mode
+Mode is usually a modifier to help us create tokens to distinguish values across two or more surface/background settings on which elements appear
```
saurav12, line 268: Where can we use this? Like if we are already using `theme` as an `object`? Or this is something inside of `theme`?

saurav12 (commented):
Really good explanation!

saurav12 (changes requested):
@kamleshchandnani Here are the images to be updated.

**Nomenclature diagram**
image

**Nomenclature Example**
image

rfcs/0001-tokens-naming-convention.md
kamleshchandnani: inside theme. Check `Principles` section

saurav12 (dismissed):

divyanshu013:
I think it would give better context to people if we can add some example usage of a component with the proposed tokens system. For example there is a mention of:

```jsx
<Icon
  name={icon}
  size={styles.iconSize({ size, children })}
  fill={styles.fontColor({ variant, variantColor, disabled })}
  testID="Button-left-icon"
/>
```

How would this look / improve with the new system in place.

divyanshu013:
This is a very detailed and well covered naming convention 💯 
Have you explored any ways to enforce these (maybe linting rules or can we do this with Typescript)? Otherwise on PRs we'll have to manually verify that the naming conventions are being followed.

kamleshchandnani:
> This is a very detailed and well covered naming convention 💯
> Have you explored any ways to enforce these (maybe linting rules or can we do this with Typescript)? Otherwise on PRs we'll have to manually verify that the naming conventions are being followed.

The tooling for enforcing this still needs to be figured out. Thanks for pointing this out. I've added it to list of open questions

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,495 @@
...
+But now if we want to refer a component within a component for example we need to refer to the `color` of an `icon` component inside a `Button` component then how shall we do that? An object can also split further and can have sub-component nesting to it as well. Eg: `Button.leftIcon.color.background` 
+
+Similarly, if we want to refer to a token value within a theme like `font.family` we can't do it alone since it doesn't has any context attached to it and we don't even know if what are we referring to? a theme token or a component token?
+
+But as soon as we attach the context our problem will be solved. For eg: `theme.font.family` tells us that we are trying to refer to font family token inside our theme.
+
+#### TL;DR
+**Object** refers to
+  1. **component** - `Button`, `theme` etc.
+  2. **sub-component** - one or more component within a component - `left-icon`, `right-icon` etc.
```
dhruvdutt, line 151: Are we calling this element or sub-component?

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,495 @@
...
+```jsonc
+// Object.Base<category.property>
+Button.color.text
+Button.color.background 
+Button.color.border 
+Button.font.weight
+Button.font.size
+```
+
+#### Behavior
```
dhruvdutt, line 196: I understand the need for this segmentation but `"behaviour"` with values `"heading"` `"body"` is slightly confusing.

rfcs/0001-tokens-naming-convention.md
kamleshchandnani: Element is a generic term. A component is an element and a sub-component is also an element. But you never will use these keywords it's just to set the meaning. Hence it says `one or more component within a component` along with the example

rfcs/0001-tokens-naming-convention.md
kamleshchandnani: Yes and hence it's in the open question :)

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,495 @@
...
+// and then we use it like this
+<Icon
+  name={icon}
+  size={styles.iconSize({ size, children })}
+  fill={styles.fontColor({ variant, variantColor, disabled })}
+  testID="Button-left-icon"
+/>
+```
+
+>P.S Some work still needs to be done. I need to figure out if we can completely get rid of styles.fontColor function
```
dhruvdutt, line 414: I think there still might be certain edge if/else conditions based on prop values and their derivations. Let's keep this function for now. After building some components, if we feel this is getting redundant we can remove it.

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,495 @@
...
+
+*behavior*, *category*, *variant* and other levels can overlap and compete. For example, an “error red” can be both generic(*category.behavior.variant*) i.e. `color.feedback.error` and specific(*object.category.propery.variant*) i.e. `Notification.color.text.error`. This forces us to decide:
+
+> At what level(s) do I store this purposeful decision? Is it ok to store the same decision in two different locations? If purpose of two different choices is nearly identical, should it be 1 or 2 tokens?
+
+Both `color.feedback` and `Notification.color.text` concepts have other variants (`warning`, `success`, `info` and `label`, `value`, and `helper-text`, respectively) for which `error` completes a set. Even if the actual red value is the same, we would consider aliasing the specific one to the generic one
+
+```js
+// specific = generic
+Notification.color.text.error = color.feedback.error
```
dhruvdutt, line 425: 💯

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,495 @@
...
+## Principles Used
+There are certain principles that were kept in mind while creating this naming framework
+### Specificity over Flexibility
+Tokens like `theme.color.success` combine *category* (`color`) and *variant* (`success`) but leaves the interpretation to the user to apply `theme.color.success` to any of `background`, `border` or `text`.
+
+Flexibility comes at the expense of specificity. A `success` color may only be intended for `text` or `background` but not both. Even more, an object reflecting `success` may require distinct colors for `text` versus `background` versus `border`. In this case, including a *property* level in a token results in a more specific yet less flexiblity eg: `theme.color.background.success` or `theme.color.text.success`.
+
+### Start within, then promote across
+
+Start with making the tokens `local` to the components, then follow the usage pattern, if used more than twice then promote it to `global` token. This is a healthy way to add tokens gradually without adding things blindly to local components or globally.
```
dhruvdutt, line 438: Let's use same terminology across:
~`local`~ `component`
~`global`~ `theme`

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,495 @@
...
+### Theme ≠ Mode
+
+A theme may eventually require `light`, `dark` modes. PG might require a `light` and `dark` mode but their theme is different from Razorpay X. Similarly, Razorpay X may require `dark` and `light` mode but their dark and light could be different from any other product's theme
```
dhruvdutt, line 441: Is there a 1:1 mapping between "theme" and product?

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,495 @@
...
+## How/Where will we store these tokens?
+### Storing theme tokens
+The theme tokens shall we created under `packages/blade/themes`
```
dhruvdutt, line 444:
```suggestion
The theme tokens shall we created under (`packages/blade/themes`)[https://github.com/razorpay/blade/tree/master/packages/blade/themes]
```

rfcs/0001-tokens-naming-convention.md
```diff
@@ -0,0 +1,495 @@
...
+### Storing component tokens
+The component tokens can be created in 2 ways:
+1. Co-located inside the component file.
+2. Create a token file under the component's directory i.e `packages/blade/components/Button/tokens.ts`.
+3. Create component tokens under `packages/blade/tokens`.
+
+Until now I believe that if we store the component level tokens under the components directory that'll be better because:
+1. We can refer them from a central token file in all the platform level implementation of a particular component(React/React Native/Svelte etc.)
```
dhruvdutt, line 452: "tokens under the components" is not "central".
Central is the 3rd approach.

dhruvdutt: Another difference with 2nd and 3rd approach is tokens would be stored in separate packages. And version for these packages might be different as we had discussed earlier.
Would we be enforcing people to upgrade to latest theme every time they want to upgrade even a single component package? How would we maintain backward compatibility?

dhruvdutt: Oh, this is also there in open questions. Saw it now. 🙈

rfcs/0001-tokens-naming-convention.md
dhruvdutt: Yes, saw it later. 🙈

dhruvdutt (commented):
Very well thought through and detailed proposal. Loved the idea around tokenization. 💯 
The RFC title could be "Theme Tokenization". It's not just a naming convention but it's a way in which we're storing and deriving all theme tokens.

PR #228, opened 2021-01-06
docs(rfc): add typography for multiple devices

saurav12:
### [View Formatted RFC](https://github.com/razorpay/blade/blob/add/rfc-typography-scale/rfcs/2021-01-05-typography-scaling.md)


rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+| `<h2>` | 25px   | 32px             |
+| `<h1>` | 32px   | 40px             |
+
+That means, for devices like, **desktops/tablets** if `<h6> = 16px` than, for **mobiles** `<h6> = (16 / 1.125) = 14px`.
+
+>📝 Note: Here we're rounding-off all the numbers for the sake of simplicity and to align the content according to the grid-system.
+
+# Drawbacks/Constraints
+Few things to consider here:
+- As we are moving towards adaptive-layout so we need to consider - should we also support responsive layout? For both cases, I don't see any challenges as the typography system remains the same.
```
divyanshu013, line 105: By responsive layout here do you mean that typography would also scale responsively? For example `h1` varying between 32px and 40px progressively based on screen width?

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+- How will this be managed on the tech-side? This might create more work as we'll have to maintain four scales,
+  - Landing/Marketing page (mobile & desktop/tablets)
+  - Product/Dashboard (mobile & desktop/tablets)
```
divyanshu013, line 108: I think this can be done with theme configuration (different design tokens for different usecases). A theme context can be provided at the root and the child components would be able to adjust accordingly.

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+>**Similar vertical & horizontal ratio across multiple breakpoints (for headings)**
+
+Our product is majorly used on mobiles/tablets/desktops. Therefore, We'll  consider only two breakpoints for the typography scale, one for mobiles & one for both desktops & tablets. Which is for devices having width `>=991px` (approx.)
+
+![Scaling: Similar vertical & horizontal ratio](https://miro.medium.com/max/2000/1*LpBKPTjjtW3I3xKEcae--Q.png)
+
+In this approach, we use the same ratio of `1.125` down each page and across each breakpoint. Pretty straightforward, we multiply everything down and across by the same ratio.
+
+### **Exception**
+This only applies to heading styles as we don't want to change our base font size (`14px`) which remains consistent across all devices/breakpoints.
```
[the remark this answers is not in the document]
saurav12: Yes! @divyanshu013. I've done an internal audit of all our dashboards. And it comes out that we're mostly using `13px` & `14px` as a base font-size on the body text. For our internal products, it is better to go with `14px` as a base font-size as dashboard content is pretty dense and will help users to consume more. Although, we can still play with spacing so that it doesn't feel complex.

rfcs/0002-typography-scaling.md
saurav12: Typography will scale responsively, given that it will only scale in segments. Which means if the `h1` at `1440px` is `40px` than at `991px` it will become `32px` directly rather than going from `**40px** -> 38px -> 36px -> 34px -> **32px**`. I hope you get my point @divyanshu013.

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+- **On Tablets** (user can capture a payment while on-the-go to their home back from the office)
+- **On Desktops** (user can request for access to the international payment via the dashboard)
+
+Or in-general a user can visit the company's marketing/landing page and consume information about us and our offerings.
+
+### **Expected Outcome**
+A typography system that is more readable/consumable across different devices (mobiles/tablets/desktops).
+
+# Detailed Design
+We are following a [**Major Second**](https://type-scale.com/?size=14&scale=1.125&text=A%20Visual%20Type%20Scale&font=Lato&fontweight=400&bodyfont=Poppins&bodyfontweight=400&lineheight=1.65&backgroundcolor=white&fontcolor=%23333&preview=false) modular scale (1.125) for our typography system but and that currently has two major limitations,
```
kamleshchandnani, line 69:
```suggestion
We are following a [**Major Second**](https://type-scale.com/?size=14&scale=1.125&text=A%20Visual%20Type%20Scale&font=Lato&fontweight=400&bodyfont=Poppins&bodyfontweight=400&lineheight=1.65&backgroundcolor=white&fontcolor=%23333&preview=false) modular scale (1.125) for our typography system but that currently has two major limitations,
```

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+- **Does not cater for landing/marketing pages.** Which are generally much larger compared to the size-ratio on the dashboard (Maybe a [Major Third](https://type-scale.com/?size=14&scale=1.125&text=A%20Visual%20Type%20Scale&font=Lato&fontweight=400&bodyfont=Poppins&bodyfontweight=400&lineheight=1.65&backgroundcolor=white&fontcolor=%23333&preview=false) scale will work better). This is currently being looked by the comm. design team based on the same principles.
```
kamleshchandnani, line 70: These limitations shall go under `Drawbacks/Constraints` section

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+- **Same type scale across all devices.** After the type scale is defined, we also need to define the scale across multiple device sizes. There are few approaches that we've came up with and after research from other DS/products like *GitHub*, *Eightshapes*, *Paystack*, *Stripe*, and a few more we've observed that for dashboards generally tweaking the **heading** sizes works best and the base font-size will remain same for all other devices/breakpoints.
+
+We'll look into the responsive type scale and see what approach fits best for our use case.
+
+### **Our Approach**
+>**Similar vertical & horizontal ratio across multiple breakpoints (for headings)**
+
+Our product is majorly used on mobiles/tablets/desktops. Therefore, We'll  consider only two breakpoints for the typography scale, one for mobiles & one for both desktops & tablets. Which is for devices having width `>=991px` (approx.)
+
+![Scaling: Similar vertical & horizontal ratio](https://miro.medium.com/max/2000/1*LpBKPTjjtW3I3xKEcae--Q.png)
```
kamleshchandnani, line 80: Shouldn't this image come before this section? Because right now if I read the below paragraph and then look at the image it feels disconnected.
>Our product is majorly used on mobiles/tablets/desktops. Therefore, We'll  consider only two breakpoints for the typography scale, one for mobiles & one for both desktops & tablets. Which is for devices having width `>=991px` (approx.)

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+
+In this approach, we use the same ratio of `1.125` down each page and across each breakpoint. Pretty straightforward, we multiply everything down and across by the same ratio.
```
kamleshchandnani, line 82: Can you add some examples or elaborate this point? Not able to understand or relate to it. Maybe I think you can remove the above image and add it in context to this point

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+
+### **Exception**
```
kamleshchandnani, line 84: Before diving into exception can we have a section which defines more details about how our scale will look like with some examples?

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+
+### **Custom Scaling**
+As the name suggests this type of scaling is preferred based on visual assumptions and is very custom in nature. The decisions around the type scale here doesn't hold any principles and are best suited for fancy marketing/landing pages.
+
+> [Read here](https://eightshapes.com/articles/typography-in-design-systems/) for more info on responsive typography.
+
+# Adoption strategy
+We need to implement this type-scale as a core part of DS. This will automatically drive adoption.
+- This won't impact the existing DS.
+- It is a part of the new DS and whenever the new DS is implemented the whole typography across all devices and pages will be affected.
```
kamleshchandnani, line 145: Add one more point
* This will be implemented as tokens in the new version of DS

kamleshchandnani, line 145:
```suggestion
- It is a part of the new version of DS and whenever the new version of DS is implemented the whole typography across all devices and pages will be affected.
```

kamleshchandnani (changes requested):

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+- [References](#references)
+
+# Summary
+Typography is an integral part of web design. The user who is consuming information from our dashboard should not face any challenges while reading/analyzing data. Therefore we need to make sure that the type element should scale well across multiple devices (desktops/mobile/tablets/etc.).
+
+# Basic Example
+This is how the typography will look after the implementation of type scales.
+
+#### **On Desktop**
+<img width="600" src="./images/website-example-desktop.png" />
```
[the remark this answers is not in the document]
saurav12: There is no annotation on the first image, as it is not defined yet! 😬

kamleshchandnani:
Putting my thoughts about the Open Questions
>How are we going to implement this type-scale? Using styled-components or CSS?

We'll be using styled-components which is an abstraction on the raw css so I don't see any challenges for this

>Marketing/Landing pages will need their own responsive type-scale as the scale ratio for such sites is more compared to the product. This is already in progress from the Comm. design team. Should we include that as well in the current system?

We should bake our design language for almost every need of the Razorpay. The components for these might not exist but atleast the design decisions should be derived from blade itself. Would like to know more why is this blocking for Marketing/Landing pages

rfcs/0002-typography-scaling.md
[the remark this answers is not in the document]
kamleshchandnani: @saurav12 Can you add this explanation in the doc as an explanation of why we are choosing 14px as base?

rfcs/0002-typography-scaling.md
saurav12: Yes, I added it there only at first, but this doesn't make sense as the user will need this context before diving into the whole typography scaling thing.

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+Simply put, heading styles for mobile devices will be always in decrements of `1.125` times that of the current desktop/tablet heading style.
+
+### **Example:**
+
+| Style  | Mobile | Desktops/Tablets |
+|--------|--------|------------------|
+| `<h6>` | 14px   | 16px             |
+| `<h5>` | 16px   | 18px             |
+| `<h4>` | 18px   | 20px             |
+| `<h3>` | 20px   | 25px             |
```
dhruvdutt, line 95: > heading styles for mobile devices will be always in decrements of `1.125` times that of the current desktop/tablet heading style.

This looks being follow for h4, h5, h6 but in h1, h2, h3 it seems there's a bigger difference.

rfcs/0002-typography-scaling.md
```diff
@@ -0,0 +1,162 @@
...
+#### **On Desktop**
+<img width="600" src="./images/website-example-desktop.png" />
+<img width="600" src="./images/dashboard-example-desktop.png" />
+
+---
+
+#### **On Mobile**
+<div style="display:flex; align-items: flex-start">
+<img width="300" src="./images/website-example-mobile.png" />
+<img width="300" src="./images/dashboard-example-mobile.png" />
```
dhruvdutt, line 48: The "Recent Activity" section on desktop is `20px` and on mobile its `18px`. It's scaling down?

rfcs/0002-typography-scaling.md
dhruvdutt: Also, why are you showing all breakpoints instead of only those that we are following?

rfcs/0002-typography-scaling.md
saurav12: > This looks being follow for h4, h5, h6 but in h1, h2, h3 it seems there's a bigger difference in rounding.

Yes! Because as we go higher we skip a few sizes. If we follow the same pattern the size for `h1` it would look much smaller. This is considered more from the visual aspect rather than just incrementing of numbers. Although, the numbers are always in multiples of `1.125`.

rfcs/0002-typography-scaling.md
saurav12: Yes! Because it is proportionate for smaller devices (like mobile).

divyanshu013:
Regarding line height:

> we generally have a simple formula that derives the line-height and we consider it to be 150% of the font-size.

I don't think this is a one size fit all. Line height should be set based on the font size. Larger the font size, smaller the line height. In larger text (such as hero section or headings) you don't need as much line height for readability (a line height of `1` or `1.1` works), whereas on smaller text like paragraphs a good line height is crucial for readability (for example `1.4 - 1.7`).

There are also certain nuances based on line length. Larger the line length, more you need line height. For example for a text content of about 80 words per line more line height is needed than for 120 words per line.

divyanshu013:
Also came across this project [capsize](https://seek-oss.github.io/capsize/). They're tackling typography in a bit different and interesting manner.

rfcs/0002-typography-scaling.md
saurav12: We will only have a few breakpoints, not as much as shown in the image. This is just to give an idea of how the scaling works. Breakpoint discussion is still pending from our end.

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+Our product is majorly used on mobiles/tablets/desktops. Therefore, We'll  consider only two breakpoints for the typography scale, one for mobiles & one for both desktops & tablets. Which is for devices having width `>=991px` (approx.)
+
+![Scaling: Similar vertical & horizontal ratio](./images/typography/typography-same-vertical-horizontal-ratio.png)
+
+In this approach, we use the same ratio of `1.125` down across each breakpoint. Pretty straightforward, we multiply everything down and across by the same ratio.
+
+### **Example:**
+| Style  | Mobile | Desktops/Tablets |
+|--------|--------|------------------|
+| `<h6>` | 14px   | 16px             |
```
kamleshchandnani, line 122: Let's just change the column title to `Heading` and then make the content as `Level 1`, `Level 2` as `<h>` as a component doesn't exist on mobile apps as a platform so let's use the generic terms here

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+2. **Same type scaling across all devices/breakpoints.** After the type scale is defined, we also need to define the scale across multiple device sizes. There are few approaches that we've came up with and after research from other DS/products like *GitHub*, *Eightshapes*, *Paystack*, *Stripe*, and a few more we've observed that for dashboards generally tweaking the **heading** sizes works best and the base font-size will remain same for all other devices/breakpoints.
+
+We'll look into the responsive type scale and see what approach fits best for our use case.
+
+>📝 Note: **Responsive Typography** will scale in segments. Which means if `h1` at `1440px` is `40px`, then at `991px` it will become `32px` directly rather than going from `40px -> 38px -> 36px -> 34px -> 32px...`.
+
+### **Our Approach**
```
kamleshchandnani, line 110: Can we first talk about our general approach and examples for body, paragraphs and everything and then go to headings? Because right now it implies that we are talking about only headings which we aren't actually.  

Something similar to [this](https://spectrum.adobe.com/page/typography/#Font-sizes)

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+>**Similar vertical & horizontal ratio across multiple breakpoints (for headings)**
+
+Our product is majorly used on mobiles/tablets/desktops. Therefore, We'll  consider only two breakpoints for the typography scale, one for mobiles & one for both desktops & tablets. Which is for devices having width `>=991px` (approx.)
+
+![Scaling: Similar vertical & horizontal ratio](./images/typography/typography-same-vertical-horizontal-ratio.png)
```
kamleshchandnani, line 115: Can we just have an image which dictates our use cases? This image looks overwhelming. Also you already have this in references so if someone is curious they can go ahead and read it.

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+
+### **Base font size**
+After auditing the whole product it came out that we are mostly using `13px` & `14px` as a base font-size on the body text. For the internal products, it is better to go with **`14px`** as a base font-size. Because for dashboards the content is pretty dense and we need to maintain a small font-size ratio compared to marketing/landing pages. This will help the users to consume more content into a readable format. Although, we can still play with spacings so that it doesn't feel much complex.
+
+### **Line Height**
+Line heights also plays a significant role in the whole typography spectrum. If the font-size is smaller the text will be cramped and difficult to read, instead if it is larger the eyes can get lost. So for optimal readability & accessibility aim for about `140% - 180%` line-height.
+
+Considering blade we generally have a simple formula that derives the line-height and we consider it to be 150% of the font-size. i.e,
```
kamleshchandnani, line 95: Capture few things similar to [this](https://spectrum.adobe.com/page/typography/#Line-height) for line heights? Easier to understand?

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+>📝 Note: **Responsive Typography** will scale in segments. Which means if `h1` at `1440px` is `40px`, then at `991px` it will become `32px` directly rather than going from `40px -> 38px -> 36px -> 34px -> 32px...`.
+
+### **Our Approach**
+>**Similar vertical & horizontal ratio across multiple breakpoints (for headings)**
+
+Our product is majorly used on mobiles/tablets/desktops. Therefore, We'll  consider only two breakpoints for the typography scale, one for mobiles & one for both desktops & tablets. Which is for devices having width `>=991px` (approx.)
+
+![Scaling: Similar vertical & horizontal ratio](./images/typography/typography-same-vertical-horizontal-ratio.png)
+
+In this approach, we use the same ratio of `1.125` down across each breakpoint. Pretty straightforward, we multiply everything down and across by the same ratio.
```
kamleshchandnani, line 117: Can we simplify this terminology? `down across each breakpoint` sounds confusing.

Maybe something simple as
> We'll use the ratio of `1.125`(Major Second) across all the breakpoints. This means that each size is multiplied or divided by 1.125 from the previous size, starting with the base size, and rounded.

And then in the table below maybe you can highlight the above thing?

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+| `<h4>` | 18px   | 20px             |
+| `<h3>` | 20px   | 25px             |
+| `<h2>` | 25px   | 32px             |
+| `<h1>` | 32px   | 40px             |
+
+That means, for devices like, **desktops/tablets** if font-size for `<h6> = 16px` than, for **mobiles** font-size for `<h6> = (16 / 1.125) = 14px`.
+
+>📝 Note: Here we're rounding-off all the numbers for the sake of simplicity and to align the content according to the grid-system. Also, various browsers treat fractional font sizes differently.
+
+This is how it will look like across desktop & mobile.
```
kamleshchandnani, line 133: maybe add annotations to show the font sizes? This screenshot just shows the headings

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+![Scaling: Desktop & Mobile](./images/typography/type-scale-blade.png)
+
+### **Exception**
+This only applies to heading styles as we don't want to change our base font size (`14px`) which remains consistent across all devices/breakpoints.
+Simply put, heading styles for mobile devices will be always in decrements of `1.125` times that of the current desktop/tablet heading style.
```
kamleshchandnani, line 138: Can you elaborate this or explain this in simpler terms? Couldn't relate to what is the exception compared to the normal texts and what do you actually mean by decrements in `1.125` in this context.

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+- [References](#references)
+
+# Summary
+Typography is an integral part of web design. The user who is consuming information from our dashboard should not face any challenges while reading/analyzing data. Therefore we need to make sure that the type element should scale well across multiple devices (desktops/mobile/tablets/etc.).
+
+# Basic Example
+This is how the typography will look after the implementation of type scales.
+
+#### **On Desktop**
+<img width="600" src="./images/typography/website-example-desktop.png" />
```
kamleshchandnani, line 45: The razorpay.com landing page is not adding any value when placed side by side with the dashboard image. Also it doesn't have any annotations so the intent is not clear what changed when you say "this is how the page will look like after the new implementation"

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+<img width="600" src="./images/typography/dashboard-example-desktop.png" />
+
+---
+
+#### **On Mobile**
+<div style="display:flex; align-items: flex-start">
+<img width="300" src="./images/typography/website-example-mobile.png" />
```
kamleshchandnani, line 52: The razorpay.com landing page is not adding any value when placed side by side with the dashboard image. Also it doesn't have any annotations so the intent is not clear what changed when you say "this is how the page will look like after the new implementation"

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+- Dashboard Invoices (Mobile) - Font Ratios:
+![Dashboard Invoices (Mobile): Font Ratios](./images/typography/dashboard-invoices-mobile-problems.png)
+
+### **Expected Outcome**
+A typography system that is more readable/consumable across different devices (mobiles/tablets/desktops).
+
+# Detailed Design
+
+### **Introduction**
+We are following a [**Major Second**](https://type-scale.com/?size=14&scale=1.125&text=A%20Visual%20Type%20Scale&font=Lato&fontweight=400&bodyfont=Poppins&bodyfontweight=400&lineheight=1.65&backgroundcolor=white&fontcolor=%23333&preview=false) modular scale (`1.125`) for our typography system. Simply put, the font-sizes will be derived in multiples of 1.125. Also, we will be using [**Lato**](https://fonts.google.com/specimen/Lato?sidebar.open=true&selection.family=Lato:wght@400;700) as our primary font with **Regular(400)** & **Bold(700)** weights *(as of now)*.
```
kamleshchandnani, line 87: We shall also add the details like what is the fallback font we are using. This is required when we are downloading a font at that time we need to show a system fallback font. 

Read more about [fallback fonts](https://css-tricks.com/css-basics-fallback-font-stacks-robust-web-typography/) and we can use [this tool](https://meowni.ca/font-style-matcher/) to generate a fallback font for us

kamleshchandnani (changes requested):

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,200 @@
...
+
+Or in-general a user can visit the company's marketing/landing page and consume information about us and our offerings.
+
+### **Examples from the current system**
+Here are a few examples in the current system.
+- Dashboard Route - Desktop vs. Mobile:
+![Dashboard Route: Desktop vs. Mobile](./images/typography/dashboard-route-mobile-desktop-comparison.png)
+
+- Dashboard Invoices - Desktop vs. Mobile:
+![Dashboard Invoices: Desktop vs. Mobile](./images/typography/dashboard-invoices-mobile-desktop-comparison.png)
```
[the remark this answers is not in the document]
saurav12: @kamleshchandnani This is just to show how our current system looks like when we scale across devices.

kamleshchandnani (changes requested):

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,201 @@
...
+“Lato”, -apple-system, BlinkMacSystemFont, San Francisco, Segoe UI, Roboto, Helvetica Neue, sans-serif
+```
+
+### **Base font size**
+After auditing the whole product it came out that we are mostly using `13px` & `14px` as a base font-size on the body text. For the internal products, it is better to go with **`14px`** as a base font-size. Because for dashboards the content is pretty dense and we need to maintain a small font-size ratio compared to marketing/landing pages. This will help the users to consume more content into a readable format. Although, we can still play with spacings so that it doesn't feel much complex.
+
+### **Line Height**
+Line heights also plays a significant role in the whole typography spectrum. If the font-size is smaller the text will be cramped and difficult to read, instead if it is larger the eyes can get lost. So for optimal readability & accessibility aim for about `110% - 150%` line-height. This also depends on the font and font-size you are using. Generally for Headings it is around `110% - 125%` of the font-size and for other body text elements it is around `125% - 150%` of the font-size.
+
+>**`line-height = font-size * height-factor`**
```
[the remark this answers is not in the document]
saurav12: I don't think we should store the `line-height` separately. As the implementation doesn't have any logic to it. It is too much dependent on the type of font or the `font-family` we are using. So I'd suggest that we should store everything inside the typography scale itself (`line-height`, `letter-spacing`, `case`).

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,201 @@
...
+```
+“Lato”, -apple-system, BlinkMacSystemFont, San Francisco, Segoe UI, Roboto, Helvetica Neue, sans-serif
+```
+
+### **Base font size**
+After auditing the whole product it came out that we are mostly using `13px` & `14px` as a base font-size on the body text. For the internal products, it is better to go with **`14px`** as a base font-size. Because for dashboards the content is pretty dense and we need to maintain a small font-size ratio compared to marketing/landing pages. This will help the users to consume more content into a readable format. Although, we can still play with spacings so that it doesn't feel much complex.
+
+### **Line Height**
+Line heights also plays a significant role in the whole typography spectrum. If the font-size is smaller the text will be cramped and difficult to read, instead if it is larger the eyes can get lost. So for optimal readability & accessibility aim for about `110% - 150%` line-height. This also depends on the font and font-size you are using. Generally for Headings it is around `110% - 125%` of the font-size and for other body text elements it is around `125% - 150%` of the font-size.
```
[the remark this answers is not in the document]
saurav12: According to WCAG, the `line-height` should be between **150% - 200%** ([Read it here](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html)). But there are articles where people say it is also dependent on the font & the platform we are using.
**For example:**
```
.headingXXL-Desktop {
    line-height: 1.25;
    font-size: 40px;
}

.headingXXL-Mobile {
    line-height: 1.5;
    font-size: 28px;
}

``` 
So, more `line-height` on mobile might make sense compared to desktop as the `font-size` is also changing for depending on the platform it is getting rendered.

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,201 @@
+---
+Start Date: 05-01-2021
+RFC PR: NA
+Blade Issue: NA
+---
+
```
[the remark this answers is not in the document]
saurav12: These have been added now. Here are the links to the above two.
- [Storage of units](https://github.com/razorpay/blade/blob/add/rfc-typography-scale/rfcs/2021-01-05-typography-scaling.md#storage-of-units)
- [Sample Scale](https://github.com/razorpay/blade/blob/add/rfc-typography-scale/rfcs/2021-01-05-typography-scaling.md#example)

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,223 @@
...
+In the above examples, you can see there is no scaling across different devices/screens. Which is not considered as a good experience as we are compromising on the readability of things on different devices.
+### **Expected Outcome**
+A typography system that is more readable/consumable across different devices (mobiles/tablets/desktops).
+
+# Detailed Design
+
+### **Introduction**
+We are following a [**Major Second**](https://type-scale.com/?size=14&scale=1.125&text=A%20Visual%20Type%20Scale&font=Lato&fontweight=400&bodyfont=Poppins&bodyfontweight=400&lineheight=1.65&backgroundcolor=white&fontcolor=%23333&preview=false) modular scale (`1.125`) for our typography system on larger screens (desktops/tablets). And [**Custom**](https://type-scale.com/?size=14&scale=1.125&text=A%20Visual%20Type%20Scale&font=Lato&fontweight=400&bodyfont=Poppins&bodyfontweight=400&lineheight=1.65&backgroundcolor=white&fontcolor=%23333&preview=false) modular scale (`1.1`) for our typography system on smaller screens (mobiles). Simply put, the font-sizes will be derived in multiples of 1.125 & 1.1 for large screens and small screens respectively.
```
kamleshchandnani, line 85: why custom scale for mobile? aren't the other standard scale working for us?

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,223 @@
...
+
+### **Typefaces**
+**Base:** We will be using [**Lato**](https://fonts.google.com/specimen/Lato?sidebar.open=true&selection.family=Lato:wght@400;700) as our base font family having two weights, **Regular(400)** & **Bold(700)**.
+
+**Code:** For code modules/snippets we'll use [**Menlo**](https://en.wikipedia.org/wiki/Menlo_(typeface)) font family throughout the system having two weights, **Regular** & **Bold**.
+
+**Fallbacks:** We'll use default system fonts (according to platform) as a fallback,
```
kamleshchandnani, line 92: how did you come up with these fallbacks? I mean did you use some tool to figure out what is the closest system fallback?

rfcs/2021-01-05-typography-scaling.md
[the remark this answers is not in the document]
kamleshchandnani: How do we calculate `height-factor`?

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,223 @@
...
+}
+```
+
+General Formula,
+
+>**`line-height = font-size * height-factor`**
+
+And will round-off the resulting value.
+
+### **Storage of Units**
```
kamleshchandnani, line 122: Let's discuss this in person

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,223 @@
...
+
+>📝 Note: Here we're rounding-off all the numbers for the sake of simplicity and to align the content according to the grid-system. Also, various browsers treat fractional font sizes differently.
+
+### **Example:**
+This is how it will look like across desktop & mobile.
+
+- **Type Unit Scale**
+![Type Unit Scale (Typography)](./images/typography/type-unit-scale.png)
+
+- **Typography Scale**
```
kamleshchandnani, line 154: Let's discuss this in person

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,223 @@
...
+
+- **Another visual indicator**
+![Scaling: Desktop & Mobile](./images/typography/type-scale-blade.png)
+
+
+# Drawbacks/Constraints
+Few things to consider here:
+- As we are moving towards adaptive-layout so we need to consider - should we also support responsive layout? For both cases, I don't see any challenges as the typography system remains the same.
+- How will this be managed on the tech-side? This might create more work as we'll have to maintain more scales,
+  - Landing/Marketing page (mobile & desktop/tablets)
```
kamleshchandnani, line 165: not sure if we also now need another scale for marketing

kamleshchandnani (changes requested):

rfcs/2021-01-05-typography-scaling.md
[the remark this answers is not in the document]
saurav12: According to what I've read, everyone follows their own hierarchy of `height factor`. We cannot calculate the `height factor` as this is something that is dependent on the visual look and feel. For now, we'll not keep any logical formula here.

rfcs/2021-01-05-typography-scaling.md
saurav12: No, the other scales were not working, the gap between them was too high. `1.1` was working well for our use case.
_Disclaimer If we change the font in the future, this might change._

rfcs/2021-01-05-typography-scaling.md
saurav12: This is industry standard. I've read few blogs and put them in a particular sequence. I've already tested these on different devices.

kamleshchandnani: the problem is the users see a flash of unstyled content(a.k.a FUC) until the font is loaded. So the recommended practice is to use a font family which is close to our actual fonts to reduce this flash. You can use [this tool](https://font-match.netlify.app/)

saurav12: Yeah! I've already used this tool and system fonts are closest to Lato. Which we are also using.

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,223 @@
...
+
+### **Line Height**
+Line heights also plays a significant role in the whole typography spectrum. If the font-size is smaller the text will be cramped and difficult to read, instead if it is larger the eyes can get lost. According to [WCAG](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html), the line-height should be between `150% - 200%` of the font-size. But instead of blindly copying these values, we should also consider the font-face and platform on which this is getting rendered. Therefore, for optimal readability & accessibility aim for about `125% - 160%` line-height.
+
+For example,
+```
+.headingXXL-Desktop {
+  line-height: 1.25; /* looks good on large screens */
```
kamleshchandnani, line 106: make it absolute `px` value since we are not going with storing unitless line-height

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,223 @@
...
+
+>**`line-height = font-size * height-factor`**
+
+And will round-off the resulting value.
+
+### **Storage of Units**
+
+Units such as `line-height`, `letter-spacing`, `case` should not be stored separately. As the implementation doesn't have any logic to it. It is too much dependent on the type of font or the font-family we are using. So I'd suggest that we should store everything inside the typography scale itself.
+
+Also, the scale should be stored at theme level rather than at a global level. So that, in case if there is any new product in the future, it will be flexible to interchange things that way. For now all the products will follow the same font structure as mentioned above.
```
kamleshchandnani, line 126: Based on the discussion let's make it global level

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,223 @@
...
+
+General Formula,
+
+>**`line-height = font-size * height-factor`**
+
+And will round-off the resulting value.
+
+### **Storage of Units**
+
+Units such as `line-height`, `letter-spacing`, `case` should not be stored separately. As the implementation doesn't have any logic to it. It is too much dependent on the type of font or the font-family we are using. So I'd suggest that we should store everything inside the typography scale itself.
```
kamleshchandnani, line 124: let's store the lineheight as a global token since we are advocating to use a common font across DS

rfcs/2021-01-05-typography-scaling.md
kamleshchandnani: Maybe we can rephrase it:
If any product wants to bring in their own typography scale then they need to override the current typography scale provided by blade. This also means that you cannot add/remove levels from the base scale

rfcs/2021-01-05-typography-scaling.md
kamleshchandnani: Can we now accommodate and make things clear about what are global tokens and what are theme and what are component level tokens as we discussed.

kamleshchandnani (changes requested):

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,221 @@
...
+>📝 Note: **Responsive Typography** will scale in segments. Which means if `h1` at `1440px` is `40px`, then at `991px` it will become `32px` directly rather than going from `40px -> 38px -> 36px -> 34px -> 32px...`.
+
+### **Our Approach**
+>**Multiple type-scaling across breakpoints**
+
+Our product is majorly used on mobiles/tablets/desktops. Therefore, We'll  consider only two breakpoints for the typography scale, one for mobiles & one for both desktops & tablets. Which is for devices having width `>=991px` (approx.)
+
+![Scaling: Multiple type scales ratio](./images/typography/typography-multi-type-scaling.png)
+
+In this type of scaling the modular scale changes on every breakpoint. For example: **mobiles** will use `Minor Third (1.2)` and **desktops/tablets** will use `Custom Scale (1.1)`.
```
kamleshchandnani, line 147:
```suggestion
In this type of scaling the modular scale changes on every breakpoint. For example: **desktops/tablets** will use `Minor Third (1.2)` and **mobiles** will use `Custom Scale (1.1)`.
```
typo 😄

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,221 @@
...
+
+>📝 Note: Here we're rounding-off all the numbers for the sake of simplicity and to align the content according to the grid-system. Also, various browsers treat fractional font sizes differently.
+
+### **Example:**
+This is how it will look like across desktop & mobile.
+
+- **Type Unit Scale:** These are stored as global tokens which can be utilised by the alias-level or component-level tokens to make sense of things.
```
kamleshchandnani, line 154:
```suggestion
- **Type Unit Scale:** These are stored as global tokens which can be utilised by the theme-level or component-level tokens to make sense of things.
```

do you mean theme level here?

rfcs/2021-01-05-typography-scaling.md
```diff
@@ -0,0 +1,221 @@
...
+![Type Unit Scale (Typography)](./images/typography/type-unit-scale.png)
+
+- **Typography Scale:** These are stored as component-level tokens which can be utilised by the components in the system to make meaning of things.
```
kamleshchandnani, line 157: in the image can we rename the columns as
Category -> Component Token Name
Size -> Alias

And in alias actually show `font-100` and for readability if you want you can add the actual size below `font-100`

kamleshchandnani: Also for line-height can we just show the px values? the multiplier doesn't adds any information here rather is creating more confusion.

kamleshchandnani (changes requested):

kamleshchandnani:
@saurav12 We didn't talk about capsize in this PR? Do you think we shall add that?

rfcs/2021-01-05-typography-scaling.md
saurav12: And instead of `Minor Second (1.2)` it will be `Major Second (1.125)` 😅

rfcs/2021-01-05-typography-scaling.md
saurav12: > in the image can we rename the columns as
> Category -> Component Token Name
> Size -> Alias
> 
> And in alias actually show font-100 and for readability if you want you can add the actual size below font-100

Didn't got this actually! Can you explain this once again? @kamleshchandnani

divyanshu013:
capsize is a really cool approach and worth evaluating. Posted about the same [long back](https://github.com/razorpay/blade/pull/228#issuecomment-759255592) 😛

rfcs/2021-01-05-typography-scaling.md
divyanshu013: How is this determined? Major / minor, didn't find any context in the doc, or maybe I missed it.

saurav12: This has been picked up visually, there is a little context we've mentioned. As for the tools like the dashboard, it is better to use `Major Second` as we get **more control and flexibility of choosing font sizes** with this approach.

kamleshchandnani (approved):

PR #235, opened 2021-01-28
docs(rfc): spatial system

saurav12:
### [View Formatted RFC](https://github.com/razorpay/blade/blob/add/rfc-spatial-system/rfcs/2021-01-22-spatial-system-rfc.md)


rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,159 @@
...
+
+# Adoption strategy
+Spatial scale needs to be implemented as a core part of the DS in terms of the tokens which will automatically derive the adoption.
+- Implement the [spatial scale](#spatial-scale) (as tokens).
+- Implement the 3 types of [column-grid](#column-grid) structure using the same spatial scale.
+- Make sure all the components are using the same spatial scale tokens while building them.
+
+# How do we educate people?
+Both **designers** & **developers** need to be educated.
+- Designers should always use the spatial (multiple of 4pt) scale while working on any layout.
```
divyanshu013, line 147: Is there a way to enforce this on Figma?

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,159 @@
...
+- Similarly, developers should always use the spatial tokens while defining the space between the adjacent elements inside the UI which should not be a problem as it is coming from a designer.
+
+# Open Questions
+- We can name the tokens in a better way, just like how we have did for colors.
+- We already have static tokens (i.e. which works for both desktops & mobiles) in place, should be need dynamic tokens that changes with the device?
```
divyanshu013, line 152: By dynamic tokens does this mean a `space-4` token that could be say 16px on desktop and 12px on mobile?

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,159 @@
...
+As the name suggests, this type of grid will help you organise content into evenly spaced vertical columns. The spacer which divide this grid is called as `gutter`. We'll use a **12 column grid** as it helps divide the area into *halfs*, *thirds*, *fourths*, *sixths*.
+
+Here we'll have 3 types of column grids, these will be used depending on the use case.
+1. **Default:** This has a gutter of `32px` & will be used on the internal dashboard.
+![Default Column-Grid](./images/spatial/col-grid-default.png)
+
+2. **Condensed:** This has a gutter of `16px` & will be used on mobile layouts.
+![Condensed Column-Grid](./images/spatial/col-grid-condensed.png)
+
+3. **Spacious:** This has a gutter of `48px` & will be used for landing/marketing pages.
```
divyanshu013, line 103: Would these pages use the condensed layout on mobile as well?

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,159 @@
+---
+Start Date: 22-05-2020
+RFC PR: 235
```
dhruvdutt, line 3: You can remove this. This is for the implementation PR.

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,159 @@
+---
+Start Date: 22-05-2020
```
dhruvdutt, line 2: 2020 💀

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,159 @@
...
+| $spacer-8  | 40px  |
+| $spacer-9  | 48px  |
+| $spacer-10 | 56px  |
+
+>These are mixed values, it contains multiple of 4 as well as 8.
+
+# Drawbacks/Constraints
+There are no major drawbacks here,
+- We need to make sure that all the spacing between the elements should not be hardcoded. They should use the spatial tokens whenever required.
+- We might find few screens where it is difficult to organise the spacing. For example, iPhone 6 (`375x667`). For such scenarios we don't need to break the pixels, instead we need to keep the margins & paddings consistent. It's ok to have oodly-sized layout which keeps the grid consistent. Although, no user will likely see the actual measurements here.
```
dhruvdutt, line 132:
```suggestion
- We might find few screens where it is difficult to organise the spacing. For example, iPhone 6 (`375x667`). For such scenarios we don't need to break the pixels, instead we need to keep the margins & paddings consistent. It's ok to have oddly sized layout which keeps the grid consistent. Although, no user will likely see the actual measurements here.
```

rfcs/2021-01-22-spatial-system-rfc.md
saurav12: Hahaha! Can't enforce, but yes we can guide them through in a way where they should use the spacer grids before the implementation of the design.

rfcs/2021-01-22-spatial-system-rfc.md
saurav12: @divyanshu013 Yes! Somewhat similar to this - https://spectrum.adobe.com/page/design-tokens/#Global-tokens

rfcs/2021-01-22-spatial-system-rfc.md
saurav12: Yes, as we've less space on the mobile, condensed layout (`16px` gutter) is good enough. Although, we might use a different spacing inside the elements that has nothing to do with the layout grid.

rfcs/2021-01-22-spatial-system-rfc.md
saurav12: Fixed!

rfcs/2021-01-22-spatial-system-rfc.md
divyanshu013: Thanks for clarifying. I think we can add few usecases for dynamic tokens to help decide on this.

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,159 @@
...
+### **Why do we need a spatial system?**
+Without space none of your elements get seen, instead they become noise. There is a great relationship that exsist between space and elements. By controlling and shaping space in our designs, we create rhythm, direction, & motion.
+
+Uniformity in spacing & sizing doesn't only allows the product to look more consistent but it also helps designers and developers to communicate better & reduce number of decisions a designer has to make on daily-basis.
+
+There are so many inconsistencies which we see in the current UI, here are a few examples:
+![Inconsistent Spacing & Alignment](./images/spatial/uneven-spacing.png)
+
+### **Expected Outcome**
+Due to these inconsistencies the users using our product can get a feeling of untrustworthy, cheap design, & inconsistency. When everything looks aligned it creates a visual rhythm which in turns look pleasing and something our users expect from a brand like us especially in the fintech space.
```
kamleshchandnani, line 45: let's sanitise `cheap design` 🙈

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,159 @@
...
+>**📝Note:** This system will used across mobiles, tablets & desktops.
+
+### **But why 8pt grid?**
+There is no right or wrong answer here, but there are few reasons for choosing 8pt over all others (`5pt`, `6pt`, `7pt`...)
+- It scales perfectly in all type of screen displays, be it Android or iOS (`@0.75` or `@1.5`).
+- Apple & Google advice to use the same, without stating too many reasons though.
+- It's simple for us to do the math. The `4` & `8` are simpler to calculate and they are flexible enough to setup a scale.
+
+### **Base Unit**
+We'll start with the base unit, this will allow us to create a scale of all the sizes in your spatial system. For us this also means using multiples of `x8` to define dimensions, padding, and margin of elements.
```
kamleshchandnani, line 72: Maybe complete the sentence on the lines of `but x8 won't work for us let's see in the below section` because when I just read this point I got carried away with the thought that `x8` will just be too much in multiple terms and then I saw the section `What system will work for Razorpay`

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,159 @@
...
+![Content-First Spacing](./images/spatial/content-first.png)
+
+### **Grids**
+The grid system helps align page elements based on sequenced columns and rows. Basically it helps you arrange your structure into structured manner. Here we'll look into 2D organisation of information.
+
+### **Column Grid**
+As the name suggests, this type of grid will help you organise content into evenly spaced vertical columns. The spacer which divide this grid is called as `gutter`. We'll use a **12 column grid** as it helps divide the area into *halfs*, *thirds*, *fourths*, *sixths*.
+
+Here we'll have 3 types of column grids, these will be used depending on the use case.
+1. **Default:** This has a gutter of `32px` & will be used on the internal dashboard.
```
kamleshchandnani, line 97: in the image can you mark and highlight one of the gutter space with `32px`?

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,159 @@
...
+| $spacer-8  | 40px  |
+| $spacer-9  | 48px  |
+| $spacer-10 | 56px  |
+
+>These are mixed values, it contains multiple of 4 as well as 8.
+
+# Drawbacks/Constraints
+There are no major drawbacks here,
+- We need to make sure that all the spacing between the elements should not be hardcoded. They should use the spatial tokens whenever required.
+- We might find few screens where it is difficult to organise the spacing. For example, iPhone 6 (`375x667`). For such scenarios we don't need to break the pixels, instead we need to keep the margins & paddings consistent. It's ok to have oddly sized layout which keeps the grid consistent. Although, no user will likely see the actual measurements here.
```
kamleshchandnani, line 132: can you post the image as well of what is right way and wrong way for explanation of this point?

rfcs/2021-01-22-spatial-system-rfc.md
kamleshchandnani: Can you also mark the screen widths on which you have showed these grids? It'll be better to even see the pink bar widths in the image

kamleshchandnani (changes requested):

rfcs/2021-01-22-spatial-system-rfc.md
kamleshchandnani: Can we add this to the RFC as well. Like Spacious on mobile for marketing pages

rfcs/2021-01-22-spatial-system-rfc.md
kamleshchandnani: This depends on the usability testing only then we'll be able to identify if we need different scale on mobile

kamleshchandnani:
Few things we shall consider to add
1. add some explanation about hard vs soft grid. Visual difference and 2 liner shall be more than enough
2. How does this affects our typography and icons I'm not able to understand
3. Under the adoption section you shall add how will you create these grid components on Figma which can act as marker for designers. People should not everytime create these layouts on their artboards and rather should use a component before designing layouts

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,204 @@
...
+We should consider few cases that will make the system flexible for us. Few things to consider here are the **user needs**.
+
+As a merchant (user) do they want the interface to be spacious with large fonts and limited number of actions? Absolutely not, we need something that doesn't look *too-dense* or *too-spacious*. Therefore we'll consider **`4pt base unit`** which will give us the flexibility to scale the UI in multiples of `x4`.
+
+We also need to focus on the number of variables we'll have in our scale. These are the variables/tokens developers will use to while defining a grid layout.
+
+### **How to apply a spatial system?**
+Applying a system can come in the form of `padding`, `margins`, `width` & `height`. So there are two things we can control here *main sizing* & *internal element sizing*.
+
+- **Element-first (strict element sizing):** In this type the size takes the priority. For examples *buttons*, *form inputs*, *etc*.
```
kamleshchandnani, line 101: If we have fixed sizing then how will we scale it across devices?
* User's device font size will increase font size of then element and hence because we have fixed sizing the text will get truncated or get overflowed of the element?
* Zooming/magnifying the page on web. how will that affect this?

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,204 @@
...
+The grid system helps align page elements based on sequenced columns and rows. Basically it helps you arrange your structure into structured manner. Here we'll look into 2D organisation of information.
+
+### **Column Grid**
+As the name suggests, this type of grid will help you organise content into evenly spaced vertical columns. The spacer which divide this grid is called as `gutter`. We'll use a **12 column grid** as it helps divide the area into *halfs*, *thirds*, *fourths*, *sixths*.
+
+Here we'll have 3 types of column grids, these will be used depending on the use case. All of them are following the same maximum width of `1280px`.
+1. **Default:** This has a gutter of `32px` & will be used on the internal dashboard.
+![Default Column-Grid](./images/spatial/col-grid-default.png)
+
+2. **Condensed:** This has a gutter of `16px` & will be used on mobile layouts *(only for dashboards)*.
```
kamleshchandnani, line 138: we are saying mobile layout but we are showing the max-width as `1280px`? 🤔

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,204 @@
...
+### **Columns, Gutters and Margins**
+A place where the content is placed inside a layout. These vary depending on the screen size or device you are designing on. Here are few examples,
+
+- **Mobile Layout**
+![Mobile Layout](./images/spatial/mobile-layout.png)
+
+- **Tablet Layout**
+![Tablet Layout](./images/spatial/tablet-layout.png)
+
+- **Desktop Layout**
```
kamleshchandnani, line 157: Why are we using condensed layout grid for desktops? In the above section we said that this will only be used on mobile right?

kamleshchandnani (changes requested):

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,204 @@
...
+
+### **Column Grid**
+As the name suggests, this type of grid will help you organise content into evenly spaced vertical columns. The spacer which divide this grid is called as `gutter`. We'll use a **12 column grid** as it helps divide the area into *halfs*, *thirds*, *fourths*, *sixths*.
+
+Here we'll have 3 types of column grids, these will be used depending on the use case. All of them are following the same maximum width of `1280px`.
+1. **Default:** This has a gutter of `32px` & will be used on the internal dashboard.
+![Default Column-Grid](./images/spatial/col-grid-default.png)
+
+2. **Condensed:** This has a gutter of `16px` & will be used on mobile layouts *(only for dashboards)*.
+![Condensed Column-Grid](./images/spatial/col-grid-condensed.png)
```
kamleshchandnani, line 139: Shouldn't the width over should show `360px` which is the normal layout? because `375` is odd and you have shown what to do in that case below. Also the grid won't be equally divided on `375px` but the diagram shows equal distribution which is kind of misleading

rfcs/2021-01-22-spatial-system-rfc.md
```diff
@@ -0,0 +1,204 @@
...
+# How do we educate people?
+Both **designers** & **developers** need to be educated.
+- Designers should always use the spatial (multiple of 4pt) scale while working on any layout.
+- Similarly, developers should always use the spatial tokens while defining the space between the adjacent elements inside the UI which should not be a problem as it is coming from a designer.
+
+# Open Questions
+- We can name the tokens in a better way, just like how we have did for colors.
+- We already have static tokens (i.e. which works for both desktops & mobiles) in place, should be need dynamic tokens that changes with the device?
+
+# References
```
kamleshchandnani, line 199: can we add [this video](https://www.youtube.com/watch?v=aPwwv-6pRWY) in references?

rfcs/2021-01-22-spatial-system-rfc.md
saurav12: @kamleshchandnani Actually, we'll use mic of both condensed and default. Only one system cannot fulfil all the needs.
For example, we have the main layout which has a sidebar and a dashboard content window. Now the gutter between these can follow `default` system but when we the segregation happens inside the sidebar/content view we will use the `condensed` system.
> How should I convey this better?

rfcs/2021-01-22-spatial-system-rfc.md
saurav12: There can be cases other than mobile as well here. Will mention a few other cases here as well.

rfcs/2021-01-22-spatial-system-rfc.md
kamleshchandnani: I saw the notes you added for this. It conveys things better now 👍

kamleshchandnani (approved):

divyanshu013 (commented):
I think the open questions are answered now, should we add the resolution for them as well?

dhruvdutt (approved):

kamleshchandnani:
> I think the open questions are answered now, should we add the resolution for them as well?

Oops I merged it. I've asked @saurav12 to raise a PR to accommodate the resolutions to the open questions

PR #240, opened 2021-02-09
fix(Amount/native): amount text should be vertically center

### LINKED ISSUE #239: [Amount/native]: Amount text is not vertically center for Android
Amount component is behaving differently for Android and iOS.
In Android, it's starting from Top and for iOS it's starting from Center.

`<Amount currency={currency} weight="bold" size="xxlarge">`

Screenshot 2021-02-09 at 5 58 28 PM

PujanShah22:
In android, Amount text was started from top but it should be vertically center just like iOS. Found that It is
working fine if we add _lineHeight to the fractionPart text. So, I have applied changes in it.

fix #239

Before:
Screenshot 2021-02-09 at 5 58 28 PM

After(With This change):
Screenshot 2021-02-09 at 7 43 45 PM

### DIFF
--- a/src/molecules/Amount/Amount.native.js
+++ b/src/molecules/Amount/Amount.native.js
@@ -139,6 +139,7 @@ const Amount = ({ size, testID, children, currency, weight, variant, variantColo
           color={styles.text.color({ variant, variantColor, subtle })}
           size={styles.text.size({ size })}
           weight={weight}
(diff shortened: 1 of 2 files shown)


PujanShah22:
@chaitanyadeorukhkar May be this change will affect the current ui of Android version in commander app. Can you please check that for once?

shridhar52war (approved):

PR #243, opened 2021-02-15
feat(tokens): add spacing scale

kamleshchandnani:


saurav12 (approved):
I hope we can add/subtract more value into it in the future. Everything else is good to go.

packages/blade/tokens/global/spacing.ts
```diff
@@ -0,0 +1,29 @@
+type spacing = Readonly<{
```
saurav12, line 1: Can we use `spacer` instead of `spacing` here?

kamleshchandnani: Umm it might become misleading since it's a scale. The actual component can be named as `spacer` since it seems like an acting verb.

haquezameer, line 1:
```suggestion
type Spacing = Readonly<{
```

Good to have type name in PascalCase?

```javascript
const spacing: Spacing
```

haquezameer (dismissed):

PR #251, opened 2021-03-09
fix(textinput): update the value of input field when it is empty string.

susheelachoudry:
The component is not updating the empty string as a value when it is used as a controlled component to reset the input value.

### DIFF
--- a/src/atoms/TextInput/TextInput.web.js
+++ b/src/atoms/TextInput/TextInput.web.js
@@ -326,7 +326,7 @@ const TextInput = ({
   );
 
   useEffect(() => {
-    if (value) {
+    if (value !== null && typeof value !== 'undefined') {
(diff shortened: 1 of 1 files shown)


haquezameer (approved):
LGTM.

src/atoms/TextInput/TextInput.web.js
```diff
@@ -326,7 +326,7 @@ const TextInput = ({
   );
 
   useEffect(() => {
-    if (value) {
+    if (value !== null && typeof value !== 'undefined') {
```
kamleshchandnani, line 329: I think having negation checks are hard to understand. Can we just do this?
`if(value=== '' || value)`

kamleshchandnani (approved):

PR #259, opened 2021-03-30
feat(icons): add logout icon

PujanShah22:
Screenshot 2021-03-30 at 7 30 51 PM
Add logout icon

### DIFF
--- a/src/icons/index.js
+++ b/src/icons/index.js
@@ -35,7 +35,7 @@ const icons = {
   lifeBuoy: require('./LifeBuoy').default,
   link: require('./Link').default,
   lock: require('./Lock').default,
-  logOut: require('./LogOut').default,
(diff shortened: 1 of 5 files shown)


src/icons/index.js
```diff
@@ -35,6 +35,7 @@ const icons = {
   lifeBuoy: require('./LifeBuoy').default,
   link: require('./Link').default,
   lock: require('./Lock').default,
+  logOut: require('./LogOut').default,
```
dhruvdutt, line 38:
```suggestion
  logout: require('./Logout').default,
```

dhruvdutt: `logout` is a valid word. No need to camel case it.

dhruvdutt (approved):

PR #263, opened 2021-04-23
fix(Snackbar): remove content on hiding snackbar

PujanShah22:
On Android even if snackbar is not visible, Its content still remains on foreground. This is creating issue for the content which is at bottom on the screen. Users are unable to click whatever is behind the snackbar(even if snackbar is hidden).

<div style={{padding:"100px",alignItems:"center",display:"flex"}} >

</div>

### DIFF
--- a/src/molecules/Snackbar/Snackbar.native.js
+++ b/src/molecules/Snackbar/Snackbar.native.js
@@ -111,59 +111,61 @@ const Snackbar = ({ variant, title, action, onClose, maxLines, icon, position })
         }}
         onLayout={handleLayout}
       >
-        <Size width={`${SNACKBAR_WIDTH}px`}>
-          <Space padding={[1.5]}>
-            <Flex flexDirection="row" alignItems="center">
-              <SnackbarContainer variant={variant}>
-                {icon ? (
-                  <Space padding={[0, 1, 0, 0]}>
-                    <View>
-                      <Icon name={icon} size="medium" fill="light.900" testID="snackbar-icon" />
-                    </View>
+        {isVisible ? (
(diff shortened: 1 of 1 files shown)


sumit-gupta91 (dismissed):

src/molecules/Snackbar/Snackbar.native.js
```diff
@@ -111,59 +115,61 @@ const Snackbar = ({ variant, title, action, onClose, maxLines, icon, position })
...
-                        variant="secondary"
-                        size="xsmall"
-                        variantColor="light"
-                        onClick={handleAction}
-                        testID="ds-snackbar-action-button"
-                      >
-                        {action.label}
-                      </Button>
-                    </View>
+        {isVisible || isContentVisible ? (
```
shubhnik, line 118: `isContentVisible` alone should be enough? because here also `isVisible` will hide the snackbar immediately?

PujanShah22: - Only `isContentVisible` is not working in one case when app launched & need to show snackbar(for example if server is not connected)
- `isVisible || isContentVisible` will only hide content once both will be false.

I will add a video here for reference 👍

shubhnik (approved):

pritishvaidya (approved):

src/molecules/Snackbar/Snackbar.native.js
PujanShah22: https://user-images.githubusercontent.com/30696104/116347619-2d6a1b80-a80a-11eb-8ebd-e130add5737f.mp4

PR #266, opened 2021-04-26
test(theme): add theme utils tests

kamleshchandnani:
This PR adds tests for following modules

* `ThemeProvider` in `packages/blade/src/components`
* `getColorScheme` in `packages/blade/src/utils`
* `useColorScheme` in `packages/blade/src/utils`

Checklist

- [x] configure babel in `packages/blade`
- [x] configure jest in `packages/blade`
- [x] configure react-testing library in `packages/blade`
- [x] make tests pass in the new setup


.eslintrc.js
```diff
@@ -79,10 +79,11 @@ module.exports = {
       },
       plugins: ['@typescript-eslint/eslint-plugin'],
       rules: {
+        'react/jsx-uses-react': 'off',
+        'react/react-in-jsx-scope': 'off',
```
kamleshchandnani, line 83: turned this off since the [new jsx transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) released in react doesnt't needs react to be in scope

.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
kamleshchandnani, line 1: added this since the version 1.22 is erroring out 80% of the time with `expected xyz to exist in packages` when adding any package

packages/blade/.babelrc.js
```diff
@@ -0,0 +1,93 @@
...
+        [
+          'module-resolver',
+          {
+            alias: {
+              '@storybook/react': '@storybook/react-native',
+            },
+          },
+        ],
+        [
+          '@babel/plugin-transform-react-jsx',
```
kamleshchandnani, line 15: added this to support the new transform from react and hance we can get rid of `import React from 'react'` statements.

This is explicitly required for react-native projects.For react projects `@babel/preset-react` already adds this

packages/blade/jest.native.config.js
```diff
@@ -0,0 +1,24 @@
...
+  coverageThreshold: {
+    global: {
+      branches: 75,
+      functions: 75,
+      lines: 75,
+      statements: 75,
+    },
+  },
+  moduleFileExtensions: ['native.ts', 'native.tsx', 'ts', 'tsx', 'js', 'json', 'node'],
+  testMatch: ['**/__tests__/**/*.native.test.{ts,tsx}', '**/__tests__/**/*.test.{ts,tsx}'],
```
kamleshchandnani, line 16: testing all the files that matches
`.native.test.ts|js|tsx|jsx` and platform independednt files `.test.ts|js|tsx|jsx`

packages/blade/jest.web.config.js
```diff
@@ -0,0 +1,21 @@
...
+  coverageThreshold: {
+    global: {
+      branches: 75,
+      functions: 75,
+      lines: 75,
+      statements: 75,
+    },
+  },
+  moduleFileExtensions: ['web.ts', 'web.tsx', 'ts', 'tsx', 'js', 'json', 'node'],
+  testMatch: ['**/__tests__/**/*.web.test.{ts,tsx}', '**/__tests__/**/*.test.{ts,tsx}'],
```
kamleshchandnani, line 15: testing all the files that matches
`.web.test.ts|js|tsx|jsx` and platform independednt files `.test.ts|js|tsx|jsx`

packages/blade/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+  it('should render with provided theme and colorscheme', () => {
+    const { result } = renderHook(() => useTheme(), {
+      wrapper,
+      initialProps: {
+        theme: paymentsTheme,
+        colorScheme: 'light',
+        children: null,
+      },
+    });
+
```
kamleshchandnani, line 24: you might wonder what is this wrapper and initialprops so this is how we render context providers using `react-hooks` package. [Reference documentation](https://react-hooks-testing-library.com/usage/advanced-hooks)

.yarnrc
chaitanyadeorukhkar: `yarn-path ".yarn/releases/yarn-1.19.0.cjs"`

I know this is autogenerated but won't this ever cause any conflict if someone has a different yarn-path?

packages/blade/.babelrc.js
chaitanyadeorukhkar: I thought react 17 supports this without any extra config

packages/blade/src/utils/getColorScheme/__tests__/getColorScheme.native.test.ts
```diff
@@ -0,0 +1,37 @@
+import getColorScheme from '../../getColorScheme';
+
+jest.mock('react-native/Libraries/Utilities/Appearance', () => ({
```
chaitanyadeorukhkar, line 3: We mock such things in a common place for platoon so that we don't need to do this over and over again

https://github.com/razorpay/frontend-commander/blob/master/commander-platoon/jest-setup.js

packages/blade/src/utils/getColorScheme/getColorScheme.native.ts
```diff
@@ -1,7 +1,7 @@
 import { Appearance } from 'react-native';
 import { ColorSchemeNames, ColorSchemeNamesInput } from '../../tokens/theme';
 
-const getColorScheme = (colorScheme: ColorSchemeNamesInput): ColorSchemeNames => {
+const getColorScheme = (colorScheme: ColorSchemeNamesInput = 'light'): ColorSchemeNames => {
   // @TODO: convert this to hook as Appearance API also adds an eventListener which subsribes to the colorscheme changes on the device
```
chaitanyadeorukhkar, line 5:
```suggestion
  // @TODO: convert this to hook as Appearance API also adds an eventListener which subscribes to the colorscheme changes on the device
```

.yarnrc
kamleshchandnani: didn't try that 😞 
ideally is someone messes up with yarn path then logically this won't work and wouldn't fail as well as it'll pick yarn version from the globally set path

packages/blade/.babelrc.js
kamleshchandnani: I assumed the same but then was bit by errors when the writing tests 😄

packages/blade/src/utils/getColorScheme/__tests__/getColorScheme.native.test.ts
kamleshchandnani: Yes but for now I just have this mock being used at this place itself hence didn't find a use case to abstract it out

.yarnrc
dhruvdutt: Let's just use the engines key, otherwise, we'll end up doing this in all projects using yarn.
https://classic.yarnpkg.com/en/docs/package-json/#toc-engines

dhruvdutt (changes requested):

packages/blade/src/components/ThemeProvider/__tests__/ThemeProvider.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+    );
+  });
+
+  it(`should throw error when colorscheme is not one of [${colorSchemeNamesInput.toString()}]`, () => {
+    const initialColorScheme = 'random';
+    const { result } = renderHook(() => useTheme(), {
+      wrapper,
+      initialProps: {
+        theme: paymentsTheme,
+        // @ts-expect-error testing the error case when colorscheme is not one of [light, dark, system]
```
dhruvdutt, line 79: Remove this line?

chaitanyadeorukhkar (approved):

.yarnrc
kamleshchandnani: There's a difference. The change here in `.yarnrc` will switch to this version by default when running this project vs the engines key will just make sure you have a particular version of yarn. We need people to switch to this exact yarn version while running this project

PR #267, opened 2021-04-27
chore(restructure): move old blade to a workspace package

kamleshchandnani:
This PR moves the old blade inside `packages/blade-old`. Here are the reasons:

1. We want to support both the versions(old as well as new) until the new version is 100% complete and old consumers are migrated fully
2. Since the old blade was not part of `packages` it was getting cumbersome to make it work in monorepo structure as the dependencies for root package were getting installed at `blade/node_modules` and now if we want to work on `packages/blade` i.e the new version in isolation the dependencies started hoisting up which caused conflicts.
3. This will give us flexibility to work on both the versions in parallel without deviating and making things work in silos. Trunk based development - yay!
4. we can use central things like linters, prettier etc but still have few things contextual like different versions of dependencies for both the versions(because old blade is running on older dependencies, we can update it as needed but doesn't makes sense to do it immediately hence isolating in a different packages unblocks that)

What will change for consumers?
1. We'll be publishing this package on GH registry
2. The consumers need to have `.npmrc` that has the razorpay's github registry set
3. Run `yarn remove @razorpay/blade` 
4. Run `yarn add @razorpay/blade-old`
5. Replace all instances of `@razorpay/blade` with `@razorpay/blade-old`

Since it's a big PR here's what you can just skim - root level package.json, eslint, .npmrc rest everything is just moved inside


kamleshchandnani:
All checks passed, happy tears 😅

divyanshu013 (commented):
PR seems too big to check thoroughly. What should be reviewed?

saurabhdaware:
So from what I can figure out, 
We moved the root-level configurations and all the files to `package/old-blade/`

In a project that is using blade, If there are any patch fixes or updates in old-blade which we want to pull, will we have to change our imports? since they are currently dependent on directory structure?

kamleshchandnani:
> PR seems too big to check thoroughly. What should be reviewed?

Updated the PR description to answer this question. Thanks for raising this it was a valid concern

kamleshchandnani:
> So from what I can figure out,
> We moved the root-level configurations and all the files to `package/old-blade/`
> 
> In a project that is using blade, If there are any patch fixes or updates in old-blade which we want to pull, will we have to change our imports? since they are currently dependent on directory structure?

Thanks for asking this, updated the PR description and added section for what will change for consumer

.gitignore
```diff
@@ -1,5 +1,5 @@
 # dependencies
-node_modules
+**node_modules**
```
divyanshu013, line 2: What's the benefit of this?

package.json
```diff
@@ -1,125 +1,50 @@
 {
-  "name": "blade",
   "private": "true",
   "author": "kamleshchandnani",
   "version": "0.0.1",
   "license": "MIT",
   "engines": {
-    "node": ">=12.19.0"
+    "node": ">=14.15.0"
```
divyanshu013, line 7: Should we do `14.16.0`? LTS release

divyanshu013 (commented):
Have there been prettier changes? Lots of formatting changes are introduced.

.gitignore
kamleshchandnani: It's not picking up node_modules inside my packages 😞

package.json
kamleshchandnani: vercel doesn't supports 14.16 yet hence had to roll it back to this version else the build fails on vercel 😿

divyanshu013: Oh weird, I remember checking this [announcement](https://vercel.com/changelog/node-js-14-lts-is-now-available) sometime back.

kamleshchandnani: yes that's why I added 14.16 but it failed saying expected 14.16 but found 14.15 😞

.gitignore
divyanshu013: I think `node_modules/` should work for all `node_modules` directories, couldn't find any info on `**node_modules**` so wanted to confirm.

divyanshu013 (approved):
Looks good to me on a cursory glance. Added a couple of optional suggestions above.

.gitignore
kamleshchandnani: Okay I take the blame. My editor screwed up and was not respecting the `.gitignore` properly. I've reverted this. Thanks for pointing out

divyanshu013: I once wasted a couple of hours trying to figure out why a filename change wasn't being tracked by git (because of case insensitive nature of macOS) 😄

divyanshu013 (dismissed):
Looks good to me overall.

PR #268, opened 2021-04-28
ci(workflow): release old-blade

kamleshchandnani:
This PR adds a release workflow for blade-old package to publish it to GH npm registry

What will change for consumers?

1. We'll be publishing this package on GH registry
2. The consumers need to have `.npmrc` that has the razorpay's github registry set
3. Run yarn remove `@razorpay/blade`
4. Run yarn add `@razorpay/blade-old`
5. Replace all instances of `@razorpay/blade` with `@razorpay/blade-old`


packages/blade/src/utils/getColorScheme/getColorScheme.native.ts
```diff
@@ -1,14 +1,14 @@
...
-const getColorScheme = (colorScheme: ColorSchemeNames): Exclude<ColorSchemeNames, 'system'> => {
+const getColorScheme = (colorScheme: ColorSchemeNamesInput): ColorSchemeNames => {
   // @TODO: convert this to hook as Appearance API also adds an eventListener which subsribes to the colorscheme changes on the device
   if (colorScheme === 'light' || colorScheme === 'dark') {
     return colorScheme;
   }
 
   if (colorScheme === 'system') {
-    return Appearance.getColorScheme() || 'light';
+    return Appearance.getColorScheme() ?? 'light';
```
saurabhdaware, line 11: is this change supposed to be here?

.github/workflows/blade-old-release.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Codebase
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
dhruvdutt, line 18: Default is `1` which should be fine. Why `0`? Is it even a valid value? 🤔

packages/blade/src/utils/getColorScheme/getColorScheme.native.ts
kamleshchandnani: didn't get you?

saurabhdaware: oh I mean it doesn't look like related to release so is this an intentional change?

.github/workflows/blade-old-release.yml
kamleshchandnani: Yes it's a valid option. [Source](https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches)

changeset needs this. Right from the documentation of changeset
> This makes Actions fetch all Git history so that Changesets can generate changelogs with the correct commits

packages/blade/src/utils/getColorScheme/getColorScheme.native.ts
kamleshchandnani: If you mean if this change is contextual to this PR then yes since I updated few packages and ts-eslint started yelling at me and also I've added this to github check now else it'll fail the moment I merge this to master

dhruvdutt (approved):

PR #271, opened 2021-05-06
feat(theme): add override theme functionality

kamleshchandnani:
This PR adds a functionality to `override` existing theme and create a new theme out of it for very specific use cases where you want to override some values of base theme. The theme can then be passed to `<ThemeProvider>`. You can only override values of properties from existing theme, you can't remove the properties from the base theme as that might break the whole system and the components.


packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable no-console */
+import isObject from 'lodash/isObject';
+
+export type DeepPartial<T> = {
```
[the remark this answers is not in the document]
divyanshu013: :godmode:

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
```diff
@@ -0,0 +1,75 @@
...
+  [P in keyof T]?: T[P] extends Record<number | string, unknown> ? DeepPartial<T[P]> : T[P];
+};
+
+const isPartialMatchObjectKeys = <ActualObject>({
```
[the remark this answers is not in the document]
divyanshu013: Subjective: for functions with generic types I find regular `function` better for readability.

```ts
function isPartialMatchObjectKeys<ActualObject>(...) {
 // ...
}
```

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
```diff
@@ -0,0 +1,75 @@
...
+    for (const [key, valueToMatch] of Object.entries(innerObjectToMatch)) {
+      const valueToInspect = innerObjectToInspect[key as keyof DeepPartial<ActualObject>];
+
+      if (innerObjectToInspect.hasOwnProperty(key)) {
+        if (
+          valueToMatch === null ||
+          valueToMatch === undefined ||
+          valueToMatch === '' ||
+          Array.isArray(valueToMatch) ||
+          (!(valueToMatch instanceof Object) && typeof valueToMatch !== typeof valueToInspect)
```
divyanshu013, line 33: Can we simplify the conditional?

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
```diff
@@ -0,0 +1,75 @@
...
+        ) {
+          // this is an ivalid case, so we log error
```
divyanshu013, line 35: Minor typo:

```suggestion
          // this is an invalid case, so we log error
```

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
```diff
@@ -0,0 +1,75 @@
...
+        );
+        matchResponses.push(false);
+      }
+    }
+  };
+
+  matchObjectKeys({ innerObjectToMatch: objectToMatch, innerObjectToInspect: objectToInspect });
+
+  return matchResponses.every(Boolean);
+};
```
divyanshu013, line 73: For this module should we only log errors in `development` mode?

divyanshu013: Benefit: could be stripped out for production bundle with something like babel dev expression plugin?

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.test.ts
```diff
@@ -0,0 +1,121 @@
+import isPartialMatchObjectKeys from './isPartialMatchObjectKeys';
+
+const baseTheme = {
+  colors: {
+    brand: {
+      primary: {
+        300: {
+          onLight: 'somecolor',
```
divyanshu013, line 8: I probably missed this in some RFC but interested in knowing why are these `onLight`, `onDark` instead of `light`, `dark`?

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.test.ts
```diff
@@ -0,0 +1,121 @@
...
+                onLight: 'someothercolor',
+              },
+            },
+          },
+        },
+      },
+    };
+
+    const isPartialMatchObjectKeysResult = isPartialMatchObjectKeys({
+      // @ts-expect-error checking the error case
```
divyanshu013, line 86: Nice 👌

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      objectToMatch: overrides,
+      objectToInspect: baseTheme,
+    });
+    expect(isPartialMatchObjectKeysResult).toBe(false);
+  });
+
+  it('should return "false" when objectToMatch has falsy values for keys in objecttoInspect', () => {
```
divyanshu013, line 93: Minor typo:

```suggestion
  it('should return "false" when objectToMatch has falsy values for keys in objectToInspect', () => {
```

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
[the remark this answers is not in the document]
divyanshu013: Maybe add a conciser version of this info as jsdoc comment to function signature? Then it would also show up in VSCode when we hover over it.

packages/blade/src/tokens/theme/theme.d.ts
```diff
@@ -1,7 +1,7 @@
 import { Spacing } from '../global/spacing';
 import { Border } from '../global/border';
 
-export type ColorSchemeNames = 'light' | 'dark';
+export type ColorSchemeNames = 'dark' | 'light';
```
divyanshu013, line 4: Interesting reordering, darkness before light 👀

divyanshu013: Oh looks like you did this alphabetically, do we have to do this manually?

packages/blade/src/tokens/theme/overrideTheme.ts
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * [x] check if the theme is of type base theme else throw error for ts users
+   * [x] check if the overrides are valid else throw error for ts users
+   * [x] Make autocomplete work for TS users in properties
+   * [x] Nested pick type for partial object keys
+   * [x] write tests
+   * [x] check if the overrides are valid else throw error for non-ts users
+   * [x] check if the theme is of type base theme else throw error for non-ts users
+   * [x] deep merge theme and overrides and return the new theme
+   */
```
divyanshu013, line 20: Can this be removed now?

packages/blade/src/tokens/theme/overrideTheme.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (!isEqual(baseTheme, paymentsTheme)) {
+    throw new Error('[Blade:overrideTheme]: The base theme provided is not a valid Blade theme');
+  }
+
+  if (!isPartialMatchObjectKeys<Theme>({ objectToMatch: overrides, objectToInspect: baseTheme })) {
+    throw new Error('[Blade:overrideTheme]: The overrides object is not valid');
+  }
+
+  return merge(cloneDeep(baseTheme), overrides);
```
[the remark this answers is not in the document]
divyanshu013: Might be good to add this as a comment in code?

packages/blade/src/tokens/theme/__tests__/overrideTheme.test.ts
```diff
@@ -0,0 +1,131 @@
...
+              highContrast: {
+                onLight: 'someothercolor',
+              },
+            },
+          },
+        },
+      },
+    };
+
+    const overridenTheme: Theme = {
```
divyanshu013, line 35: Subjective / optional: lot of nested spreading in this test object, can we create a deep copy of the default theme and then selectively modify only the `colors.brand.primary[300]` and `feedback.background.positive.highContrast` keys?

```ts
const overridenTheme = deepCopy(paymentsTheme);

overridenTheme.colors.brand.primary[300] = {
  ...paymentsTheme.colors.brand.primary[300],
  ...overrides.colors.brand.primary[300],
}
```

Does this convey the intent of test?

packages/blade/src/tokens/theme/__tests__/overrideTheme.test.ts
```diff
@@ -0,0 +1,131 @@
+import overrideTheme from '../overrideTheme';
+import paymentsTheme from '../paymentsTheme';
+import { Theme } from '../theme.d';
+
+const invalidOverridesObjectError = '[Blade:overrideTheme]: The overrides object is not valid';
+const invalidBaseThemeError =
+  '[Blade:overrideTheme]: The base theme provided is not a valid Blade theme';
```
divyanshu013, line 7: Since we're checking for these errors explicitly, they can grow out of sync from the code. Maybe we can create a mapping of error types to error messages and then use the same in test and the module.

divyanshu013:
Also please add PR description 😛

packages/blade/jest.native.config.js
```diff
@@ -13,7 +13,12 @@ module.exports = {
     },
   },
   moduleFileExtensions: ['native.ts', 'native.tsx', 'ts', 'tsx', 'js', 'json', 'node'],
-  testMatch: ['**/__tests__/**/*.native.test.{ts,tsx}', '**/__tests__/**/*.test.{ts,tsx}'],
+  testMatch: [
+    '**/__tests__/**/*.native.test.{ts,tsx}',
+    '**/__tests__/**/*.test.{ts,tsx}',
+    '**/*.native.test.{ts,tsx}',
```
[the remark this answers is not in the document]
kamleshchandnani: we need to test for `.native` and also for `ts` so `.native.ts`, `.native.tsx` and so on

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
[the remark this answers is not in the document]
kamleshchandnani: >Subjective / optional: for functions with generic types I find regular function better for readability.

Yes I agree to this. But maybe we need to find one pattern to stick to? Else it's just multiple ways spread across the codebase. Either all `function` or all `arrow` functions

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable no-console */
```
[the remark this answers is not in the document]
kamleshchandnani: not every file has no-console hence keeping this upto the developer wherever they feel comfortable genuinely because for most cases we won't have console logs

packages/blade/src/tokens/theme/theme.d.ts
kamleshchandnani: yes I want to turn off this rule. It's annoying 😞

packages/blade/src/tokens/theme/__tests__/overrideTheme.test.ts
kamleshchandnani: Yes that's the ideal plan. Maybe I shall create one helper in the succeeding PR.

packages/blade/jest.native.config.js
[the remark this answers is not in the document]
divyanshu013: But the rule `*.test.{ts,tsx}` wouldn't it also run the tests matching `*.web.test.{ts,tsx}`?

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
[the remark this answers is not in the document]
divyanshu013: I think we can keep it as it's for now.

packages/blade/jest.native.config.js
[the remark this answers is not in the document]
kamleshchandnani: oh yes, now I understood. Good catch. Will update 👍

packages/blade/src/utils/merge/merge.test.ts
```diff
@@ -0,0 +1,24 @@
+import merge from './merge';
+
+describe('merge', () => {
```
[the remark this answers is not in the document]
kamleshchandnani: Ah okay, so if you see we are not using the merge directly in our codebase we do a re-export and the reason for that is that we get a flexibility to change the underline library or logic whenever we want at a single place and all our consumers would still work. Now if sometime tomorrow we change the underlying implementation lets say we move from lodash to something else then tests will ensure that it is not breaking any use case. So it's a double safeguard

packages/blade/src/utils/isEqual/isEqual.test.ts
```diff
@@ -0,0 +1,27 @@
+import isEqual from './isEqual';
+
+describe('isEqual', () => {
```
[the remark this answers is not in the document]
kamleshchandnani: https://github.com/razorpay/blade/pull/271#discussion_r628706891

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
kamleshchandnani: any suggestions? The reason is all the other conditions I can test without negation but not getting any simpler way for this one. The idea here is that I want to check if it's not instance of an object then the typeof **match** should be equal to type of **inspect** and if it's not then it's a failure case.

Maybe I'll add a comment to make it easier to read?

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
kamleshchandnani: Good one. Again a generic solution that needs to be figure. Created an issue https://github.com/razorpay/blade/issues/272 to track this

packages/blade/src/utils/merge/merge.test.ts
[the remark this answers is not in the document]
divyanshu013: I think we can add tests if / when we move to a different implementation.

divyanshu013:
@kamleshchandnani can you please push the new changes? I can take a look again after that from my side.

kamleshchandnani:
> @kamleshchandnani can you please push the new changes? I can take a look again after that from my side.

I did push and left it in the background and just saw it failed lol 🙈 . You can review now
Screenshot 2021-05-08 at 2 31 03 PM

packages/blade/jest.native.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
     },
   },
   moduleFileExtensions: ['native.ts', 'native.tsx', 'ts', 'tsx', 'js', 'json', 'node'],
-  testMatch: ['**/__tests__/**/*.native.test.{ts,tsx}', '**/__tests__/**/*.test.{ts,tsx}'],
+  testMatch: ['**/*.native.test.{ts,tsx}', '**/*.test.{ts,tsx}'],
```
divyanshu013, line 16: The second glob pattern `**/*.test.{ts,tsx}` this would still match a test file for `web.ts`? For example `User.web.ts` wouldn't it? So it would end up running tests for `.web` as well.

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
divyanshu013: Okay, comment should be fine for now I think.

packages/blade/src/utils/merge/merge.test.ts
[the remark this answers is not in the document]
kamleshchandnani: Later never comes have experienced that multiple times 😞.

packages/blade/jest.native.config.js
kamleshchandnani: Nupe. I've explicitly ignored that. Here's [the magic](https://github.com/razorpay/blade/pull/271/files#diff-c44cf9f865bf8178b6ee129ff776089b04b4e9738e6c121eee474f329aaef0abR5)

divyanshu013 (commented):
Looks good to me.

divyanshu013 (approved):
👍

packages/blade/src/tokens/theme/overrideTheme.ts
```diff
@@ -0,0 +1,23 @@
...
+import { paymentsTheme } from '../theme';
+import type { Theme } from './theme.d';
+
+type OverrideTheme = {
+  baseTheme: Theme;
+  overrides: DeepPartial<Theme>;
+};
+
+const overrideTheme = ({ baseTheme, overrides }: OverrideTheme): Theme => {
+  if (!isEqual(baseTheme, paymentsTheme)) {
```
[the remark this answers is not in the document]
kamleshchandnani: it's just a base theme, take any theme it adheres to just one structure.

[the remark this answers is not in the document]
kamleshchandnani: okay wait my bad. I just want to check the structure. Damn how did I miss this 🤔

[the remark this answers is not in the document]
kamleshchandnani: Now I'm just thinking is it even worth it to have this check? Because I can't magically identify which theme to match it against. might as well remove this check?

packages/blade/src/tokens/theme/overrideTheme.ts
```diff
@@ -0,0 +1,23 @@
+import { cloneDeep, isEqual, merge, isPartialMatchObjectKeys, DeepPartial } from '../../utils';
+import { paymentsTheme } from '../theme';
+import type { Theme } from './theme.d';
```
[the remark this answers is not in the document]
divyanshu013: Can't we import types directly without `type` keyword?

packages/blade/src/utils/getColorScheme/getColorScheme.web.test.ts
```diff
@@ -1,5 +1,5 @@
 import MatchMediaMock from 'jest-matchmedia-mock';
-import getColorScheme from '../../getColorScheme';
+import getColorScheme from './';
```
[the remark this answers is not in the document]
divyanshu013: Yes, this looks clearer for understanding. Is there any other way to pick web and native files? For example how metro can pick `.android` or `.ios` file even if we import the file by it's name directly.

packages/blade/src/tokens/theme/overrideTheme.ts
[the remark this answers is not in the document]
kamleshchandnani: Yes we can. It's just better to be explicit and this helps bundler to eliminate types when there's ambiguity for example when the component and types name are same

packages/blade/src/tokens/theme/bankingTheme.ts
```diff
@@ -1,7 +1,7 @@
 import border from '../global/border';
 import spacing from '../global/spacing';
 import globalColors from '../global/colors';
-import { Theme, Colors, Shadows } from './theme.d';
+import type { Theme, Colors, Shadows } from '../theme';
```
dhruvdutt, line 4: "import type" ❤️

packages/blade/src/utils/getColorScheme/getColorScheme.web.test.ts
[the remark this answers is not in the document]
kamleshchandnani: yes `import getColorScheme from '../getColorScheme';` this will work and will pickup the name correctly but what I want to test here is that the re-exports are working correctly from `index` else there's no clear way for me to test if my re-exports work perfectly

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
```diff
@@ -0,0 +1,76 @@
...
+            `[isPartialMatchObjectKeys]: Unexpected value: ${JSON.stringify(
+              valueToMatch,
+            )} of type ${typeof valueToMatch} for key: ${key}`,
+          );
+          matchResponses.push(false);
+        }
+
+        if (typeof valueToMatch === 'string') {
+          // we have reached leaf node of the objectToMatch(i.e sub-object)
+          matchResponses.push(true);
```
dhruvdutt, line 47: No need to match or equate `valueToMatch` and `valueToInspect`?

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
```diff
@@ -0,0 +1,76 @@
...
+          // the condition checks if the "valueToMatch" is not of type object then "valueToMatch" type should be same as type of "valueToInspect"
+          (!(valueToMatch instanceof Object) && typeof valueToMatch !== typeof valueToInspect)
+        ) {
+          // this is an invalid case, so we log error
+          console.error(
+            `[isPartialMatchObjectKeys]: Unexpected value: ${JSON.stringify(
+              valueToMatch,
+            )} of type ${typeof valueToMatch} for key: ${key}`,
+          );
+          matchResponses.push(false);
```
dhruvdutt, line 42: Why not just break the recursion early here itself and return `false`?

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
```diff
@@ -0,0 +1,76 @@
...
+      const valueToInspect = innerObjectToInspect[key as keyof DeepPartial<ActualObject>];
+
+      if (innerObjectToInspect.hasOwnProperty(key)) {
+        if (
+          valueToMatch === null ||
+          valueToMatch === undefined ||
+          valueToMatch === '' ||
+          Array.isArray(valueToMatch) ||
+          // the condition checks if the "valueToMatch" is not of type object then "valueToMatch" type should be same as type of "valueToInspect"
+          (!(valueToMatch instanceof Object) && typeof valueToMatch !== typeof valueToInspect)
```
dhruvdutt, line 34: The `typeof` match check can be done regardless, no?

packages/blade/src/tokens/theme/overrideTheme.ts
```diff
@@ -0,0 +1,24 @@
+import { cloneDeep, isEqual, merge, isPartialMatchObjectKeys, DeepPartial } from '../../utils';
+import { paymentTheme, bankingTheme } from '../theme';
+import type { Theme } from '../theme';
+
+type OverrideTheme = {
+  baseTheme: Theme;
+  overrides: DeepPartial<Theme>;
```
dhruvdutt, line 7: How about calling this `themeOverrides`?

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
kamleshchandnani: elaborate?

packages/blade/src/tokens/theme/overrideTheme.ts
kamleshchandnani: seems redundant overrideTheme is the func name then again themeOverrides

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
kamleshchandnani: All the checks have been done above and hence it's reaching here so doesn't makes sense to do it again

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
kamleshchandnani: it's a recursion inside an iteration. So it doesn't checks nested things. This is the first thing I had implemented and then learnt it the hard way 😄

packages/blade/src/utils/isPartialMatchObjectKeys/isPartialMatchObjectKeys.ts
kamleshchandnani: because typeof is not straightforward as our human brains think. typeof null, undefined all leads to object so `valueToMatch: null === valueToInspect: {}` hence there's a place where I explicitly check for `isObject(valueToMatch) && isObject(valueToInspect)`

chaitanyadeorukhkar (approved):

dhruvdutt (approved):

PR #276, opened 2021-05-19
feat(theme): add typography scale

kamleshchandnani:
This PR adds typography tokens which accounts for font family, font size, font weight. Lineheight and Letterspacing tokens are yet to be worked upon by design so will add later in successive PR

### DIFF
--- a/packages/blade/src/tokens/global/fontFamily.native.ts
+++ b/packages/blade/src/tokens/global/fontFamily.native.ts
@@ -1,6 +1,8 @@
+import { Platform } from 'react-native';
+
 const fontFamily = {
   text: 'Lato',
-  code: 'monospace',
+  code: Platform.OS === 'ios' ? 'Courier' : 'monospace',
 };
 
 export default fontFamily;
--- a/packages/blade/src/tokens/global/fontFamily.native.ts
+++ b/packages/blade/src/tokens/global/fontFamily.native.ts
@@ -1,0 +1,6 @@
+const fontFamily = {
+  text: 'Lato',
+  code: 'monospace',
+};
+
+export default fontFamily;
--- a/packages/blade/src/tokens/global/typography.ts
+++ b/packages/blade/src/tokens/global/typography.ts
@@ -1,0 +1,101 @@
+import type { FontFamily } from './fontFamily';
+import fontFamily from './fontFamily';
+
+type FontWeight = {
+  regular: 400;
+  bold: 700;
+};
+
+type FontSize = {
+  10: number;
+  25: number;
+  50: number;
+  75: number;
+  100: number;
+  200: number;
+  300: number;
+  400: number;
+  500: number;
+  600: number;
+  700: number;
+  800: number;
+  900: number;
+  1000: number;
+};
+
+const fontWeight: FontWeight = {
+  regular: 400,
+  bold: 700,
+};
+
+export type Typography = Record<
+  'desktop' | 'mobile',
+  {
+    fonts: {
+      family: FontFamily;
+      size: FontSize;
+      weight: FontWeight;
+    };
+  }
+>;
+
+const typography: Typography = {
+  mobile: {
+    fonts: {
+      family: {
+        ...fontFamily,
+      },
+      size: {
+        10: 10,
+        25: 11,
+        50: 12,
+        75: 14,
+        100: 15,
+        200: 17,
+        300: 18,
+        400: 20,
+        500: 22,
+        600: 24,
+        700: 27,
+        800: 29,
+        900: 32,
+        1000: 35,
+      },
+      weight: {
+        ...fontWeight,
+      },
+    },
+    // lineHeights: {},
+    // letterSpacings: {},
+  },
+  desktop: {
+    fonts: {
+      family: {
+        ...fontFamily,
+      },
+      size: {
+        10: 9,
+        25: 10,
+        50: 11,
+        75: 12,
+        100: 13,
+        200: 16,
+        300: 18,
+        400: 20,
+        500: 22,
+        600: 25,
+        700: 28,
+        800: 32,
+        900: 36,
+        1000: 40,
+      },
+      weight: {
+        ...fontWeight,
+      },
+    },
+    // lineHeights: {},
+    // letterSpacings: {},
+  },
+};
+
+export default typography;
--- a/packages/blade/src/tokens/global/fontFamily.d.ts
+++ b/packages/blade/src/tokens/global/fontFamily.d.ts
@@ -1,0 +1,6 @@
+export type FontFamily = {
+  text: string;
+  code: string;
+};
+
+export { default } from './fontFamily.web';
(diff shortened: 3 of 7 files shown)


packages/blade/src/tokens/global/typography.ts
```diff
@@ -0,0 +1,101 @@
+import type { FontFamily } from './fontFamily';
+import fontFamily from './fontFamily';
+
+type FontWeight = {
+  regular: 400;
+  bold: 700;
+};
+
+type FontSize = {
+  10: number;
```
chaitanyadeorukhkar, line 10: Why such a scale for fontSize? RFC propsed something else https://github.com/razorpay/blade/blob/master/rfcs/2021-19-02-units-for-typography-and-layout.md#how-will-we-store-and-use-it

kamleshchandnani: such a scale in the sense? RFC proposed the same thing na `font-size-10` and so on?

chaitanyadeorukhkar: RFC says 1,2,3,4,... but why are we going ahead with 10, 25, 50, 75, 100?
Screenshot 2021-05-20 at 12 10 44 PM

dhruvdutt: fontSizes till `1000` 🤔

packages/blade/src/tokens/global/typography.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const typography: Typography = {
+  mobile: {
+    fonts: {
+      family: {
+        ...fontFamily,
+      },
+      size: {
+        10: 10,
+        25: 11,
```
dhruvdutt, line 50: Why are we having a separate key to value mapping? Why not same value in key and value?
```
12: 12,
14: 14,
```

packages/blade/src/tokens/global/typography.ts
```diff
@@ -0,0 +1,101 @@
...
+    // letterSpacings: {},
+  },
+  desktop: {
+    fonts: {
+      family: {
+        ...fontFamily,
+      },
+      size: {
+        10: 9,
+        25: 10,
```
dhruvdutt, line 78: Same as above.

packages/blade/src/tokens/global/typography.ts
```diff
@@ -0,0 +1,101 @@
...
+>;
+
+const typography: Typography = {
+  mobile: {
+    fonts: {
+      family: {
+        ...fontFamily,
+      },
+      size: {
+        10: 10,
```
dhruvdutt, line 49: Shouldn't it start from 14?

packages/blade/src/tokens/global/typography.ts
dhruvdutt, line 78: Looks inconsistent. For mobile 25: 11

packages/blade/src/tokens/global/typography.ts
dhruvdutt: +1

kamleshchandnani: RFC was just taking an example. Refer [this RFC](https://github.com/razorpay/blade/blob/master/rfcs/2021-01-05-typography-scaling.md)

kamleshchandnani: We are doing 10, 25 etc because that's a general convention when talking about font scales and color scales we follow 100-900 and then increment/decrement based on the requirement. For more context [read this](https://graphicdesign.stackexchange.com/a/126061)

packages/blade/src/tokens/global/fontFamily.native.ts
```diff
@@ -0,0 +1,6 @@
+const fontFamily = {
+  text: 'Lato',
+  code: 'monospace',
```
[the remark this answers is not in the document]
kamleshchandnani: Need to test. Will check and respond

packages/blade/src/tokens/global/typography.ts
kamleshchandnani: if both the scales would have been same why would I have splitted them into desktop and mobile? Read this section in [RFC](https://github.com/razorpay/blade/blob/master/rfcs/2021-01-05-typography-scaling.md#introduction) 🙈

packages/blade/src/tokens/global/typography.ts
```diff
@@ -0,0 +1,101 @@
...
+  900: number;
+  1000: number;
+};
+
+const fontWeight: FontWeight = {
+  regular: 400,
+  bold: 700,
+};
+
+export type Typography = Record<
```
saurav12, line 31: Haven't we incorporated the main scale yet? I mean - [Desktop/Mobile Scale](https://www.figma.com/file/LV9kjgzhkfsktvbRgTuGJ3/Blade-DSL?node-id=5%3A0)

saurav12 (commented):
Looks good!

dhruvdutt (approved):
LGTM. Verify once about `monospace` for native.

packages/blade/src/tokens/global/typography.ts
dhruvdutt: Got it. 🤦  Been a while since I had read that. 🙈

packages/blade/src/tokens/global/fontFamily.native.ts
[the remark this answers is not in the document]
chaitanyadeorukhkar: Any update?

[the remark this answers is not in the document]
chaitanyadeorukhkar: https://github.com/react-native-training/react-native-fonts


Checked a couple of places. Seems like iOS doesn't have `monospace` :(

packages/blade/src/tokens/global/typography.ts
kamleshchandnani: We have, you're looking at a TypeScript type 😄 
See [here](https://github.com/razorpay/blade/pull/276/files#diff-efc5afda90b97389459637e4fbe63fcfb21a7590bd46f798e5bcc9d9911a334fR43-R48)

packages/blade/src/tokens/global/fontFamily.native.ts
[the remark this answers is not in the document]
kamleshchandnani: Screenshot_1621838278

Tested on android. iOS my simulator is not in a mood to run today and failing randomly. Since this is a rare use case of using `monospace` it's acting as a placeholder. Can we come back to this once we have a case for this?

chaitanyadeorukhkar (approved):
Let's resolve `monospace` for ios when a use-cases pops up

packages/blade/src/tokens/global/fontFamily.native.ts
```diff
@@ -0,0 +1,8 @@
+import { Platform } from 'react-native';
+
+const fontFamily = {
+  text: 'Lato',
+  code: Platform.OS === 'ios' ? 'Courier' : 'monospace',
```
kamleshchandnani, line 5: @chaitanyadeorukhkar happy? 🙈

chaitanyadeorukhkar: Hahaha better 🚀

chaitanyadeorukhkar (approved):

PR #277, opened 2021-05-22
ci(workflows): use self-hosted runners

kamleshchandnani:
This PR migrates all the workflows to use `self-hosted` github runners


kamleshchandnani:
Can't run self-hosted on public repositories because of [security issues](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories)

PR #304, opened 2021-06-04
feat(blade-old): add wallet icon

parichay28:


packages/blade/src/tokens/theme/theme.d.ts
```diff
@@ -29,13 +29,35 @@ export type Shadows = {
   };
 };
 
-export type Feedback = 'information' | 'negative' | 'notice' | 'positive';
+export type Feedback = 'information' | 'negative' | 'neutral' | 'notice' | 'positive';
```
dhruvdutt, line 32: Seems like some merge issue.
Can you sync with latest master?

kamleshchandnani: no merge issue. I've added new value and disabled the sorting of keys eslint rule hence you see this weird addition

kamleshchandnani (approved):

PR #314, opened 2021-07-12
feat(blade): add type generation scripts

kamleshchandnani:
This PR generates types for `components`, `tokens`, `utils` that will be be published and can be used by consumers.

Read the detailed process and steps in the [RFC section](https://github.com/razorpay/blade/blob/generate-types/rfcs/2021-06-15-shipping-blade.md#generating-ts-types-for-consumers)

A glimpse of the entire process
generating-types

Checklist:
- [x] Add scripts to generate `types`
- [x] Update RFC with the type generation approach and process


packages/blade/rollup.config.js
```diff
@@ -1,8 +1,8 @@
-import fs from 'fs';
-import { babel } from '@rollup/plugin-babel';
-import peerDepsExternal from 'rollup-plugin-peer-deps-external';
-import resolve from '@rollup/plugin-node-resolve';
-import commonjs from '@rollup/plugin-commonjs';
+import { babel as pluginBabel } from '@rollup/plugin-babel';
+import pluginPeerDepsExternal from 'rollup-plugin-peer-deps-external';
+import pluginResolve from '@rollup/plugin-node-resolve';
+import pluginCommonjs from '@rollup/plugin-commonjs';
```
kamleshchandnani, line 4: ignore this, it's all formatting

packages/blade/rollup.config.js
```diff
@@ -1,8 +1,8 @@
...
+import pluginDeclarations from 'rollup-plugin-dts';
```
kamleshchandnani, line 5: added this plugin

packages/blade/rollup.config.js
```diff
@@ -88,11 +88,23 @@ const getNativeConfig = ({ exportCategory }) => ({
...
+  ],
+  plugins: [pluginDeclarations()],
+});
 
 const config = exportCategories
-  .map((exportCategory) => [getWebConfig({ exportCategory }), getNativeConfig({ exportCategory })])
+  .map((exportCategory) => [
+    getWebConfig({ exportCategory }),
+    getNativeConfig({ exportCategory }),
+    getDeclarationsConfig({ exportCategory }),
```
kamleshchandnani, line 106: we run `getDeclarationsConfig` after all source code is bundled.

packages/blade/tsconfig-generate-types.json
```diff
@@ -0,0 +1,11 @@
+{
```
kamleshchandnani, line 1: `tsconfig` which just focusses on generating types. It extends the base config itself. Kept it separate instead of passing the command line arguments for enabling/disabling flags for generating types

rfcs/2021-06-15-shipping-blade.md
```diff
@@ -460,6 +466,178 @@ One common question you might have: If exports is not supported by all the tools
...
+
+### Phase 4
+We need to bundle our declarations so that it maps to `components`, `tokens` and `utils`. We'll use rollup to bundle our declarations with [`rollup-plugin-dts`](https://github.com/Swatinem/rollup-plugin-dts).
+
+Here's how our rollup config will look like:
+```js
+const exportCategories = ['components', 'tokens', 'utils'];
+
+const getDeclarationsConfig = ({ exportCategory }) => ({
+  input: `${outputRootDirectory}/types/${exportCategory}/index.d.ts`,
```
chaitanyadeorukhkar, line 603: This section is not as clear. Seems like there's some magic going on in this phase 😛 

For instance, how does this step handle generating config for a nested directory like `/types/components/ThemeProvider/index.d.ts`

`exportCategory` will map it till `/types/components`. How does it figure out that there's another directory inside that it needs to go to and pick up that `index.d.ts` as well?

kamleshchandnani: it won't. We are just using top level categories' re-exported file. Look at the directory structure above. we have

```
src/
	components/
		index.d.ts --- this has all the re-exported things and points to nested things and stuff
	tokens/
		index.d.ts --- this has all the re-exported things and points to nested things and stuff
	utils/
		index.d.ts --- this has all the re-exported things and points to nested things and stuff
```

it's same as giving entry point while bundling. We just define one entry point in any of the bundling tool and then it goes and generates dependency graph. Not sure if it makes sense to add this much detail in RFC?

chaitanyadeorukhkar: Got it. The plugin is internally doing this but unless a reader knows this, it seems like there's a bug here. Maybe mention `the plugin will pick up all the nested declaration files as well`?

packages/blade/package.json
```diff
@@ -28,7 +28,12 @@
   },
   "scripts": {
     "prepublish": "yarn build",
-    "build": "rollup -c && node ./scripts/generateRootImports.js",
+    "build": "run-s build:*",
```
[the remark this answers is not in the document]
kamleshchandnani: yes the way you've defined it will follow the same sequence.

packages/blade/src/tokens/global/fontFamily.native.ts
```diff
@@ -1,6 +1,7 @@
 import { Platform } from 'react-native';
+import type { FontFamily } from './fontFamily.d';
```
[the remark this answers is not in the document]
kamleshchandnani: our tools only substitutes `.js`, `.ts`, `.native.ts` so now if I add `.d.ts` to resolution map in our tools it'll get confused and might pick the first match which will again be `fontFamily.native.ts` or `fontFamily.web.ts` so it might conflict. So added it explicitly to mean definitions

packages/blade/package.json
```diff
@@ -28,7 +28,12 @@
...
+    "build:clean": "rm -rf build",
+    "build:generate-types": "tsc -p ./tsconfig-generate-types.json",
+    "build:copy-declarations": "copyfiles -u 1 \"src/**/*.d.ts\" build/types",
+    "build:source": "rollup -c && node ./scripts/generateRootImports.js",
```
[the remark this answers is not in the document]
kamleshchandnani: yes you can run them individually nothing would break, it just won't give you desired output. If I add it as a single script it'll go out of my monitor(kidding lol) but it'll just create more confusion on single line and make understanding things complicated. And anyways no one really touches these scripts esp on the DS maintainers side since it's just tooling things

packages/blade/rollup.config.js
kamleshchandnani: comment in the sense?

divyanshu013: I meant your original comment

> we run getDeclarationsConfig after all source code is bundled.

Can be added in code for reference

packages/blade/package.json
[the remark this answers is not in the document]
divyanshu013: Interesting, I wonder how it maintains order of scripts because objects don't enforce order of keys

divyanshu013 (commented):
Looks good to me.

packages/blade/package.json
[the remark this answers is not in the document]
kamleshchandnani: Okay it's doing `Object.keys`
https://github.com/mysticatea/npm-run-all/blob/master/lib/read-package-json.js#L28

packages/blade/rollup.config.js
kamleshchandnani: Updated the comment. Does this makes sense? kept it simple. the previous one will be tied to RFC stuff

packages/blade/tsconfig-generate-types.json
dhruvdutt: Can we name this file `tsconfig-generate-custom-types`? Because it's only responsible for custom types & not all types.

packages/blade/tsconfig-generate-types.json
```diff
@@ -0,0 +1,11 @@
...
+  "extends": "./tsconfig.json",
+  "exclude": ["**/**.test.ts", "**/**.test.tsx", "__tests__"],
+  "include": ["src", "src/tokens/theme/theme.d.ts"],
```
dhruvdutt, line 4: Why do we need to include `"src/tokens/theme/theme.d.ts"` specifically? Does it not pick things from `src` or `src/**`?

packages/blade/tsconfig-generate-types.json
```diff
@@ -0,0 +1,11 @@
...
+  "compilerOptions": {
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "noEmit": false,
+    "declarationDir": "build/types"
```
dhruvdutt, line 9: Can add a brief comment around values for these options. They way it's mentioned in RFC Phase 2.

rfcs/2021-06-15-shipping-blade.md
```diff
@@ -460,6 +466,177 @@ One common question you might have: If exports is not supported by all the tools
...
+blade/
+└── src/
+    ├── components/
+    │   ├── ThemeProvider/
+    │   │   └── index.ts
+    │   └── index.ts
+    ├── tokens/
+    │   ├── theme/
+    │   │   ├── paymentTheme.ts
+    │   │   └── theme.d.ts
```
dhruvdutt, line 489: Are we thinking of these `d.ts` files incorrectly or something different from a TS design principle? Can we think of alternate ways to achieve this?

Let's discuss on Slack.

packages/blade/tsconfig-generate-types.json
dhruvdutt, line 4: How would we make sure we're not bundling any unnecessary d.ts files that we added for external npm packages (that don't have types)?

dhruvdutt: Also, how would we make sure we're not bundling any unnecessary d.ts files that we added for external npm packages (that don't have types)?

kamleshchandnani: that's a typo, removed it. It's not required. It was left over from my experimentation

packages/blade/tsconfig-generate-types.json
kamleshchandnani: It's ts stuff so you get documentation out of the box once you hover over it 😄 . Anyways added the comments

packages/blade/package.json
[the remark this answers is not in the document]
chaitanyadeorukhkar: > As a result, it's best not to rely on property order.

We sure it won't mess up in some edge case? Better to write all the steps one after the other? 😅

packages/blade/tsconfig-generate-types.json
chaitanyadeorukhkar: This is actually responsible for generating all the typescript types. This is the config that actually emits types for all of our TS files

chaitanyadeorukhkar (approved):

dhruvdutt (approved):

PR #321, opened 2021-07-23
build(blade): generate root `.d.ts`

kamleshchandnani:
This PR generates root `.d.ts` files which re-exports types from each category inside `blade/build`. This is required for TS to find type definitions when importing the package in TS projects.

```diff
blade/
 ├── components.ts
+├── components.d.ts
 ├── tokens.ts
+├── tokens.d.ts
 ├── utils.ts
+└── utils.d.ts
```


packages/blade/package.json
```diff
@@ -22,7 +22,7 @@
       "deafult": "./build/tokens/index.web.js"
     },
     "./utils": {
-      "react-native": "./build/tokens/index.native.js",
```
chaitanyadeorukhkar, line 25: Damn 🙈

PR #330, opened 2021-07-29
fix(blade): add description in token types

kamleshchandnani:
Since we are in the initial phases of DS and tokens adoption is our first rollout strategy, it becomes difficult for people to memorize the scales for typography, border, spacing, etc. and it'll take some time. But this hampers their workflow since they'll be copying token keys from Figma and Figma also has a limitation on what all things can be published as tokens. So in order to fix this, I've added a description to our TS types so while typing at least the consumer will get some IntelliSense help of what values map to what keys of tokens. 

Here's how things look after adding the description to the types

Kapture 2021-07-30 at 10 14 32


saurav12 (approved):
This is amazing! 🤩
Let's do this for all the things going forward.

PR #333, opened 2021-08-02
Allow Badge's variant prop to take non-semantic colors as well

parichay28:
### Description 

- PG and X to have separate color mappings for `tertiary` (cyan and xCyan respectively)
- Badge now accepts non-semantic variant `cyan`


packages/blade-old/src/atoms/Badge/Badge.native.js
```diff
@@ -38,9 +38,9 @@ const styles = {
         return getColor(theme, 'primary.900');
       default:
         if (fill === 'subtle') {
-          return getColor(theme, 'positive.930');
+          return variant ? getColor(theme, variant) : getColor(theme, 'positive.930');
```
sumit-gupta91, line 41: Once we start passing colors as variant, how will we make sure people are not passing colors which are not in theme, also does getColor function has a default case color ?

sumit-gupta91 (changes requested):

packages/blade-old/src/atoms/Badge/Badge.native.js
parichay28: getColor throws error on wrong color name

packages/blade-old/src/atoms/Badge/Badge.native.js
```diff
@@ -62,6 +67,8 @@ const styles = {
           return 'neutral.900';
         case 'neutral':
           return 'primary.900';
+        case 'cyan':
+          return 'tertiary.900';
```
chaitanyadeorukhkar, line 71: If variant is `cyan` and `fill` is `solid`, you'll end up having background as well as foreground as `tertiary.900`, right?

parichay28: In that case, it'll be `primary.900`
image

chaitanyadeorukhkar (approved):

chaitanyadeorukhkar (changes requested):
Can you update the tests as well?

https://github.com/razorpay/blade/blob/add%2Fbadge-non-semantic-colors/packages/blade-old/src/atoms/Badge/__tests__/Badge.native.test.js

chaitanyadeorukhkar (approved):

sumit-gupta91 (approved):

PR #349, opened 2021-09-13
Allow textinput right icon to be touchable

parichay28:
Enables use-case for adding password visibility toggler

### DIFF
--- a/.changeset/clean-lies-watch.md
+++ b/.changeset/clean-lies-watch.md
@@ -1,5 +1,5 @@
 ---
-"@razorpay/blade-old": patch
+"@razorpay/blade-old": major
 ---
 
 Allow textinput right icon to be touchable
--- a/packages/blade-old/src/atoms/TextInput/TextInput.native.js
+++ b/packages/blade-old/src/atoms/TextInput/TextInput.native.js
@@ -182,7 +182,7 @@ const TextInput = React.forwardRef(
       value,
       iconLeft,
       iconRight,
-      onIconRightClick,
+      onRightIconClick,
       maxLength,
       label,
       testID,
@@ -422,7 +422,7 @@ const TextInput = React.forwardRef(
                           ) : null}
                           {hasRightIcon ? (
                             <TouchableOpacity
-                              onPress={onIconRightClick}
+                              onPress={onRightIconClick}
                               activeOpacity={1}
                               {...automation(`${testID}-right-icon`)}
                             >
@@ -478,7 +478,7 @@ TextInput.propTypes = {
   value: PropTypes.string,
   iconLeft: PropTypes.string,
   iconRight: PropTypes.string,
-  onIconRightClick: PropTypes.func,
+  onRightIconClick: PropTypes.func,
   maxLength: PropTypes.number,
   label: PropTypes.string,
   testID: PropTypes.string,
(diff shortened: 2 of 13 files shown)


packages/blade-old/src/atoms/TextInput/TextInput.web.js
```diff
@@ -534,15 +535,20 @@ const TextInput = ({
...
-                                <AccessoryIcon
-                                  variant={variant}
-                                  name={iconRight}
-                                  disabled={disabled}
-                                  hasError={hasError}
-                                  //@ts-expect-error
-                                  isFocused={isFocused}
-                                  position="right"
-                                />
+                                <View
```
kamleshchandnani, line 538: why do you need to wrap it in a view?

packages/blade-old/src/atoms/TextInput/TextInput.web.js
```diff
@@ -221,6 +221,7 @@ const TextInput = ({
   value,
   iconLeft,
   iconRight,
+  onIconRightClick,
```
kamleshchandnani, line 224:
```suggestion
  onRightIconClick,
```

kamleshchandnani (changes requested):
Checks are failing

packages/blade-old/src/atoms/TextInput/TextInput.native.js
```diff
@@ -471,6 +478,7 @@ TextInput.propTypes = {
   value: PropTypes.string,
   iconLeft: PropTypes.string,
   iconRight: PropTypes.string,
+  onIconRightClick: PropTypes.func,
```
shridhar52war, line 481:
```suggestion
  onRightIconClick: PropTypes.func,
```
And elsewhere

packages/blade-old/src/icons/index.js
```diff
@@ -70,6 +70,8 @@ const icons = {
   wallet: require('./Wallet').default,
   wiFiOff: require('./WiFiOff').default,
   wiFiOn: require('./WiFiOn').default,
+  Eye: require('./Eye').default,
+  EyeOff: require('./EyeOff').default,
```
shridhar52war, line 74: Check casing. Also sort lines acending 🙈

shridhar52war (changes requested):

shridhar52war (dismissed):

packages/blade-old/src/atoms/TextInput/TextInput.web.js
parichay28: We want `AccessoryIcon` to be clickable hence wrapping in a `View`. Do we have a better approach here? @kamleshchandnani

kamleshchandnani: Not a RN expert but can't we have clickable without a view?

parichay28: We can, but then we'd need to make changes to `AccessoryIcon` component. Simplest way here would be to wrap our `AccessoryIcon` with `TouchableOpacity` for RN and div for web.

chaitanyadeorukhkar: It's better to have a TouchableOpacity around it instead of making the icon clickable since TouchableOpacity will give you a onPress feedback by reducing the opacity when pressed. Making the SVG clickable will not give you a touch feedback

shridhar52war: This change should be fine for `TextInput.web.js` IMO, changes for `TextInput.native.js` is already done right @parichay28

shridhar52war (approved):

.changeset/clean-lies-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade-old": patch
```
chaitanyadeorukhkar, line 2: This is a breaking change

```suggestion
"@razorpay/blade-old": major
```

chaitanyadeorukhkar (approved):

kamleshchandnani (approved):

PR #350, opened 2021-09-15
feat(semgrep): add semgrep

haquezameer:


.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,13 @@
+name: Semgrep
+on: [pull_request, push]
```
dhruvdutt, line 2: `push` would get triggered for any commit push: all PRs commits + master merge commit.

`pull_request` would get triggered only for all PR commits.

Doesn't make sense to have both. In this case, I think `pull_request` should suffice.

chaitanyadeorukhkar (approved):

PR #357, opened 2021-10-14
feat(blade): theme token restructure

kamleshchandnani:
### ⚠️ **Breaking Change** ⚠️ 
This PR introduces a major breaking change on how we access tokens.

### Why did we want to change the way we access tokens?
So, previously if you had to start consuming tokens from the new version of Blade you start with importing the theme provider:

```jsx
// App entry point
import { ThemeProvider } from '@razorpay/blade/components';
import { paymentTheme } from '@razorpay/blade/tokens';

function App(): JSX.Element {
  return (
    <React.Fragment>
      <GlobalStyle />
      <ThemeProvider theme={paymentTheme}>
        <Card />
      </ThemeProvider>
    </React.Fragment>
  );
}

export default App;
```
And then in one of our components, we'll use the `useTheme()` hook to get the theme and the mode like following 👇 

```jsx

const StyledCard = styled.div(
  ({ theme }: { theme: Theme }) => `
  width: 368px;
  background-color: ${theme.colors.surface.background.level2.lowContrast.onLight};
  border-radius: ${theme.border.radius.medium}px;
  box-shadow: ${theme.shadows.offsetX.level[1]}px ${theme.shadows.offsetY.level[1]}px ${theme.shadows.blurRadius.level[1]}px ${theme.shadows.color.level[1].onLight}, ${theme.shadows.offsetX.level[1]}px ${theme.shadows.offsetY.level[1]}px ${theme.shadows.blurRadius.level[1]}px ${theme.shadows.color.level[1].onLight};
  padding: ${theme.spacing[5]}px;
  display: flex;
  flex-direction: column;
`,
);

const Card = (): React.ReactElement => {
  const { theme } = useTheme();
  return (
    <React.Fragment>
      <DisplayLarge theme={theme}>Cash Advance </DisplayLarge>
      <StyledCard theme={theme}>
        <AlertInformation theme={theme}>
          The interest charged will be deposited back into your bank account within a day of
          repayment.
        </AlertInformation>
        <Divider theme={theme} />
        <CaptionRegular theme={theme}>
          This amount will be deducted in 3 installments from your settlement balance between Feb
          18-20 on a daily basis.
        </CaptionRegular>
      </StyledCard>
    </React.Fragment>
  );
};
```

#### Problem with the existing implementation:

So we pass the raw theme tokens which have everything light mode colors, dark mode colors. Different typography scales for desktop, mobile, etc. But as a consumer look at how do we access the tokens from the above file

```jsx

const { theme } = useTheme();

background-color: ${theme.colors.surface.background.level2.lowContrast.onLight};
font-size: ${theme.typography.desktop.fonts.size[200]}px;
```

* Isn't it weird to explicitly write `onLight`/`onDark` by hand while accessing color tokens? 
* Isn't it weird to explicitly write `desktop` to access the typography scale for desktop? 
* How would you as a developer change things let's say if the user toggles the color mode?
* How would you as a developer change the typography scale if the user switches from desktop to mobile or vice-versa?

You can't! Because we have **hardcoded** the object properties and which means we lost the power of dynamically changing these things based on the user's behavior which is incorrect.

#### What is the root cause of this issue?
The root cause is the mental model of how we store tokens and how do we consume them. Typically our tokens are nothing but our design decisions. So this means we need to store every decision in our token file, for eg: light mode colors, dark mode colors, typography scale for desktop, typography scale for mobile but when we consume them we want the system to take care of these things and give us single value for the modes and the platform. 

So we want something like this 👇 

```diff

const { theme } = useTheme();

-background-color: ${theme.colors.surface.background.level2.lowContrast.onLight};
+background-color: ${theme.colors.surface.background.level2.lowContrast};
-font-size: ${theme.typography.desktop.fonts.size[200]}px;
+font-size: ${theme.typography.fonts.size[200]}px;
```

Notice the removal of **`onLight`** and **`desktop`** keys from the theme object. 

So we want our system to behave in such a manner that:
* We input the raw theme(which has color modes and platform types)
* It will output the flat theme which will have color mode and platform type selected, so we don't have to hardcode `onLight`/`onDark` or `desktop`/`mobile`.

### What is the solution?
The solution is in this PR 😝.

Jokes aside, this system is nothing but our `BladeProvider`(previously known as `ThemeProvider`). It'll accept the raw theme as a prop and then based on the device type and color mode pick those values from `themeTokens` and set it in the context as `theme`. We can then use `useTheme()` hook to get the theme from the context which will be flattened.

This is how things will look after this change 👇 

```diff
// App entry point
-import { ThemeProvider } from '@razorpay/blade/components';
+import { BladeProvider } from '@razorpay/blade/components';
import { paymentTheme } from '@razorpay/blade/tokens';

function App(): JSX.Element {
  return (
    <React.Fragment>
      <GlobalStyle />
-      <ThemeProvider theme={paymentTheme}>
+      <BladeProvider themeTokens={paymentTheme}>
        <Card />
-      </ThemeProvider>
+      </BladeProvider>
    </React.Fragment>
  );
}

export default App;

// somewhere in the app
const { theme } = useTheme();

-background-color: ${theme.colors.surface.background.level2.lowContrast.onLight};
+background-color: ${theme.colors.surface.background.level2.lowContrast};
-font-size: ${theme.typography.desktop.fonts.size[200]}px;
+font-size: ${theme.typography.fonts.size[200]}px;
```

### Migration guide for apps using the older version
1. Rename **ThemeProvider** to **BladeProvider**
```diff
-import { ThemeProvider } from '@razorpay/blade/components';
+import { BladeProvider } from '@razorpay/blade/components';
```
2. Rename `theme` prop on provider to `themeTokens`
```diff
-<BladeProvider theme={paymentTheme}>
+<BladeProvider themeTokens={paymentTheme}>
```
3. import `Theme` type to be imported from `@razorpay/blade/components` instead of `@razorpay/blade/tokens`
```diff
-import type { Theme } from '@razorpay/blade/tokens';
+import type { Theme } from '@razorpay/blade/components';
```
4. Remove all the `onLight`/`onDark` keywords from the theme colors object
```diff
-background-color: ${theme.colors.surface.background.level2.lowContrast.onLight};
+background-color: ${theme.colors.surface.background.level2.lowContrast};
```
5. Remove all the `desktop`/`mobile` keywords from the theme typography object
```diff
-background-color: ${theme.colors.surface.background.level2.lowContrast.onLight};
+background-color: ${theme.colors.surface.background.level2.lowContrast};
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 </p>
 
 <p align="center">
-   A Design System that powers Razorpay
+   The Design System that powers Razorpay
 <p align="center">
 
 - [📦 Packages](#-packages)


packages/blade/src/tokens/theme/theme.d.ts
```diff
@@ -32,82 +33,65 @@ export type Shadows = {
...
   tertiary: ActionStates;
   link: ActionStates;
 };
 
-type ActionProperties = {
-  background: ActionVariants;
-  border: ActionVariants;
-  text: ActionVariants;
-  icon: ActionVariants;
+// export type ActionProperties = {
```
kamleshchandnani, line 55: intentionally commented this and kept it here

packages/blade/src/components/BladeProvider/useTheme.ts
```diff
@@ -0,0 +1,38 @@
+import { useContext, createContext } from 'react';
+import { Colors, Shadows, ShadowLevels } from '../../tokens/theme/theme.d';
+import type { Typography } from '../../tokens/global/typography';
+import type { Border } from '../../tokens/global/border';
+import type { Spacing } from '../../tokens/global/spacing';
+import { UseColorScheme } from '../../utils/useColorScheme';
```
dhruvdutt, line 6:
```suggestion
import type { UseColorScheme } from '../../utils/useColorScheme';
```

packages/blade/src/components/BladeProvider/useTheme.ts
```diff
@@ -0,0 +1,38 @@
...
+export const ThemeContext = createContext<ThemeContext>({
+  theme: undefined,
+  colorScheme: 'light',
+  setColorScheme: () => null,
+});
+
+const useTheme = (): ThemeContext => {
+  const themeContext = useContext<ThemeContext>(ThemeContext);
+  if (themeContext === undefined) {
+    throw new Error(`[@razorpay/blade]: useTheme must be used within Blade's Provider`);
```
dhruvdutt, line 33: "[@razorpay/blade]"
Let's follow the same convention for logs, errors at other places.

packages/blade/src/components/BladeProvider/BladeProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+    typography: themeTokens.typography[deviceType],
+  };
+
+  const themeContextValue = {
+    theme,
+    colorScheme,
+    setColorScheme,
+  };
+
+  return <ThemeContext.Provider value={themeContextValue}>{children}</ThemeContext.Provider>;
```
dhruvdutt, line 52: `ThemeContext` and `BladeProvider`, it's a little confusing. Can we just have `BladeContext.Provider`?

packages/blade/src/tokens/theme/theme.d.ts
```diff
@@ -126,10 +109,12 @@ export type Colors = {
...
-  colors: Colors;
+export type ColorsWithModes = Record<ColorSchemeModes, Colors>;
+
+export type ThemeTokens = {
+  colors: ColorsWithModes;
   border: Border;
   spacing: Spacing;
   shadows: Shadows;
-  typography: Typography;
+  typography: TypographyWithPlatforms;
```
dhruvdutt, line 119: Good naming `WithPlatforms`, `WithMode`.

packages/blade/src/utils/getDeviceType/getDevicetype.ts
```diff
@@ -0,0 +1,23 @@
+import isMobile from 'ismobilejs';
+
+const getDeviceType = (): 'desktop' | 'mobile' => {
+  // TODO: add event listener for orientation change
+  if (typeof navigator !== 'undefined' && navigator.product === 'ReactNative') {
```
dhruvdutt, line 5:
```suggestion
  if (navigator?.product === 'ReactNative') {
```

dhruvdutt: Would this suffice?

packages/blade/src/utils/toTitleCase/toTitleCase.ts
```diff
@@ -0,0 +1,8 @@
+const toTitleCase = (inputString: string): string =>
+  inputString
+    .toLowerCase()
+    .split(' ')
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(' ');
```
dhruvdutt, line 6: Why not just use lodash? We might miss certain cases.

Issue #363, opened 2021-10-22
[blade-old] Improved Icon Components


anuraghazra:
### The Problem 

Currently how blade-old's `Icon` component work is that we are [importing all](https://github.com/razorpay/blade/blob/master/packages/blade-old/src/icons/index.js) the icons and accessing them in runtime via [`icons[name]`](https://github.com/razorpay/blade/blob/2dde44c00342aa84a39c275116bf0927242fa171/packages/blade-old/src/atoms/Icon/Icon.web.js#L17)
This method lets us reference the icons via their name so we can do `<Icon name="download" />`

But the downside to this approach is that, this prevents treeshaking. 
Causing all the icons to be present on the final bundle even if we are not using them.
https://github.com/razorpay/blade/blob/master/packages/blade-old/src/atoms/Icon/Icon.web.js#L17

Total size of all the icons gzipped is around 12kb which is a lot: 

<img src="https://user-images.githubusercontent.com/35374649/138405278-7a8c95e6-e3de-4b10-b177-64bce30bccb7.png" width="500px" />

### Solution 

After discussing with @kamleshchandnani we decided that we can move onto a more flexible approach to directly using the icon component in the codebase. This way users would only import what they need. 

Users would do something like this: 

```js
import { DownloadIcon, LeftArrow, ChevronLeft } from 'blade-old/icons'; 

<DownloadIcon></DownloadIcon>
<LeftArrow></LeftArrow>
<ChevronLeft></ChevronLeft>
```

### Approach 

We will first create a BaseIcon component which will hold the common properties like width, height, viewbox, handle a11y and render a svg component. 

The implementation would look something like this: 

```js
const Icon = ({ label, hidden, viewbox, size, children }) => {
  const iconSize = {
    xsmall: 12,
    small: 16,
  };
  const theme = useTheme();

  return (
    <svg
      aria-label={label}
      aria-hidden={hidden}
      fill={getColor(theme, fill)}
      height={iconSize[size]}
      width={iconSize[size]}
      viewBox={viewbox}
    >
      {children}
    </svg>
  );
};
```

Then we will wrap all the icons with BaseIcon and reexport each of them separately. 

```js
// DownloadIcon.js

export function DownloadIcon({ size }) {
  return (
    <Icon size={size} viewBox="0 0 24" fill="none">
      <path
        fillRule="evenodd"
        clipRule="evenodd"
        d="..."
        fill={fill}
      />
    </Icon>
  );
}
```

Questions: 

[...]

PR #367, opened 2021-10-29
feat(blade-old): improved icon API

anuraghazra:
## Migration guide for new blade-old icons

As discussed in #363 we had decided to move to a more flexible and open API for the icons,
Instead of using a single Icon component and passing the name="" prop we will be directly using the icon by importing it.

## Migrating icons to new API

```diff
- import Icon from "@razorpay/blade-old";
+ import { AlertCircle } from "@razorpay/blade-old/icons";

- <Icon name="alert-circle" fill="shade.950" size="medium" />
+ <AlertCircle fill="shade.950" size="medium" />
```

To make this process easier I would suggest you use vscode search panel to search all the instances of `name="` which will show you all the places you used icons, and replace them with their respective icon.

## Changes in \<Icon />

Icon component is now a generic wrapper to create new icons. 
We are also exporting this \<Icon /> component so that users can build their own icons too.

```jsx
// User defined custom icons
export const MyCustomIcon = (props) => {
  return (
    <Icon viewBox="0 0 24 24" {...props}>
      <path d="" />
    </Icon>
  );
};

// usage
<Button icon={MyCustomIcon}>
```

## Migrating existing blade-old components

Few of the internal components were using the old \<Icon /> component, thus we also had to change the APIs for them.

> Basically anywhere you see `icon="name-of-the-icon"` replace it with an imported icon.

Modified components are:

- Button
- TextInput
- Snackbar
- Tabs.Tab
- SegmentControl.Option

### Button

```diff
- <Button icon="info" />
+ <Button icon={Info} />
```

### TextInput

```diff

<TextInput
  label="Label"
-  iconRight="info"
-  iconLeft="info"
+  iconRight={Info}
+  iconLeft={Info}
/>
```

### Snackbar

```diff
const snackbar = useSnackbar();

snackbar.show({ 
-  icon: "info",
+  icon: Info
})
```

### Tabs.Tab

```diff
  <Tabs.Tab
    value="payments"
    title="Payments"
    disabled={boolean('Disable Tab 1', false)}
-   icon="info"
+   icon={Info}
  >
```

### SegmentControl.Option

```diff
  <SegmentControl.Option
    value="1"
-   icon="info"
+   icon={Info}
    disabled={boolean('Disabled', false)}
    subText={text('Sub Text', undefined)}
  >
    Option 1
  </SegmentControl.Option>
```


kamleshchandnani:
> Migrating icons to new API
In this section in the diff remove the imports of the old icon

kamleshchandnani:
>We are also exporting this <Icon /> component so that users can build their own icons too.

I would suggest to not export this until there's a need. Because with the newer set of icons coming in people will start randomly use the icons directly in their respective projects and never add the icons back to Blade. So let's not export this publicly until we really need to?

kamleshchandnani (changes requested):

kamleshchandnani:
seems like there are 2 snapshots which are not updated in React Native. Can you run `yarn run test:native -u` once more? Bitrise is failing because of this

Screenshot 2021-11-02 at 12 55 11 PM

packages/blade-old/src/atoms/Checkbox/Checkbox.native.js
```diff
@@ -168,19 +181,11 @@ const Checkbox = ({
...
-                name={externalChecked ?? isChecked ? 'checkboxFilled' : 'checkboxOutlined'}
-                fill={styles.icon.fill({
-                  isChecked,
-                  disabled,
-                  variantColor,
-                  externalChecked,
-                })}
-                testID="checkbox-icon"
-              />
+              {externalChecked ?? isChecked ? (
```
divyanshu013, line 184: Suggestion: we can simplify the nested ternary.

packages/blade-old/src/atoms/TextArea/TextArea.native.js
```diff
@@ -83,8 +83,8 @@ TextArea.propTypes = {
   suffix: PropTypes.string,
   disabled: PropTypes.bool,
   value: PropTypes.string,
-  iconLeft: PropTypes.string,
-  iconRight: PropTypes.string,
+  iconLeft: PropTypes.elementType,
```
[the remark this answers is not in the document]
divyanshu013: I think it would be `Component` vs `<Component />`

divyanshu013 (commented):
What were the considerations for passing component type instead of JSX / component for icon props to other components like `Button`? For example - `icon={Info}` over `icon={<Info />}`

Also for future ref we should split larger PRs into smaller ones for ease of review, for example - stacked PRs:
- PR1: core changes in Icon component
- PR2 - n: Update few icons in each PR
- PRn: all automated (snapshot) changes, etc.
- Final integration PR (with all the above PRs)

packages/blade-old/src/atoms/Button/Button.native.js
```diff
@@ -290,9 +289,8 @@ const Button = ({
             {...automation(testID)}
           >
             <React.Fragment>
-              {icon && iconAlign === 'left' && (
+              {Icon && iconAlign === 'left' && (
```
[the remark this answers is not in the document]
anuraghazra: Yup, right will do.

packages/blade-old/src/atoms/TextInput/AccessoryIcon.web.js
```diff
@@ -64,7 +63,7 @@ const AccessoryIcon = ({ name, disabled, hasError, variant, isFocused, position
 };
 
 AccessoryIcon.propTypes = {
-  name: PropTypes.string,
+  icon: PropTypes.elementType,
```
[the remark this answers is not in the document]
anuraghazra: We can do that, but then again it will make the whole refactor pointless since we have to import all the icons to check (on the runtime) to see if they are the icons we provide. bundlesize issue will remain in that scenario.

anuraghazra:
> What were the considerations for passing component type instead of JSX / component for icon props to other components like Button? For example - icon={Info} over icon={<Info />}

So both will be almost similar semantically, the only difference is that using `icon={<Info />}` users can override the props of the icon, they can override `fill`/`size` etc 

IMO it's better to pass element type than the whole JSX component because 
1. parity with previous API, in previous API users also didn't had any override options. they just had to put the name of the icon. 
2. user's shouldn't have to worry about "how to render" the icon, that job is taken care by the button itself where it calculates [iconSize](https://github.com/razorpay/blade/blob/c2ee9454a75ee9f9f756261d186c9fbd970beb6a/packages/blade-old/src/atoms/Button/Button.web.js#L216-L244), [fontColor](https://github.com/razorpay/blade/blob/c2ee9454a75ee9f9f756261d186c9fbd970beb6a/packages/blade-old/src/atoms/Button/Button.web.js#L17-L40) etc etc styles depending on the passed props to button component. If we let users override these it can cause inconsistencies in the styles/variants.

packages/blade-old/src/icons/ArrowRight/ArrowRight.native.js
```diff
@@ -1,30 +1,20 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-import Svg, { Path } from 'react-native-svg';
-import { getThemeColors } from '../../_helpers/theme';
-import colors from '../../tokens/colors';
+import { Path } from 'react-native-svg';
+import Icon, { IconPropTypes } from '../../atoms/Icon';
 
-function ArrowRight({ width, height, fill }) {
+function ArrowRight(props) {
```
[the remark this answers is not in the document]
anuraghazra: done.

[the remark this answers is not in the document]
anuraghazra: @kamleshchandnani found some issues with not passing the props via spreading it on the svg. 

components like Flex, Space are passing down important props to the childrens for styling, but since icons doesn't take those it's messing up the styles. 

This bug is also present in current master, see in AlertCircle we are not spreading the props, but in [Info icon](https://github.com/razorpay/blade/blob/57b1bb99962b8ac8b1b19cb47c4b401cb901483d/packages/blade-old/src/icons/Info/Info.web.js#L5) (which is being used for snapshot testing) does have spreading props. 

Notice how there is no spacing between icons and input text. 

Screenshot 2021-12-01 at 2 07 26 PM


After spreading the props in icons. fixed the alignment and spacing bug.
Screenshot 2021-12-01 at 2 09 19 PM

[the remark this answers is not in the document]
kamleshchandnani: oh damn! Need to check because we actually clone the element in flex/space etc and attach props to it. Can you check further if this exactly is the issue?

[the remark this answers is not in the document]
anuraghazra: Will check, I'll try passing down className only see if that works. If not we have no choice but to spread all the props it seems 🤔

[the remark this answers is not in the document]
anuraghazra: Update: We decided to go with spreading props considering the tradeoffs.

packages/blade-old/src/_helpers/isValidElement.js
```diff
@@ -0,0 +1,8 @@
+import { isValidElementType } from 'react-is';
+
+const isValidElement = (element) => {
```
kamleshchandnani, line 3: why not name this as `isValidElementType`?

packages/blade-old/src/atoms/Button/Button.web.js
```diff
@@ -423,7 +423,7 @@ const Button = ({
             {...automation(testID)}
           >
             <React.Fragment>
-              {icon && iconAlign === 'left' ? (
+              {isValidElement(Icon) && iconAlign === 'left' ? (
```
kamleshchandnani, line 426: ❤️

kamleshchandnani (dismissed):
LGTM ❤️ 

Let's add the migration script as a chained PR to this and then we'll merge and release

packages/blade-old/src/_helpers/isValidElement.js
anuraghazra: isValidElement is smaller and doesn't clash with react-is's function name. 
naming is hard :p let's keep it this way?

divyanshu013:
Can we can do the codemod in a followup PR if it's a blocker?

anuraghazra:
> Can we can do the codemod in a followup PR if it's a blocker?

It's already merged, will fix the conflicts now.

packages/blade-old/MIGRATION.md
```diff
@@ -0,0 +1,195 @@
+- [From version 1.4.x to 2.x.x](#from-version-14x-to-2xx)
+
+
+
+# From version 1.4.x to 2.x.x
+
+## Migration guide for new blade-old icons
+
+### Migrating icons to new API
```
kamleshchandnani, line 9: remoev this section entirely

packages/blade-old/MIGRATION.md
```diff
@@ -0,0 +1,195 @@
...
++ import { AlertCircle } from "@razorpay/blade-old/icons";
+
+- <Icon name="alert-circle" fill="shade.950" size="medium" />
++ <AlertCircle fill="shade.950" size="medium" />
+```
+
+To make this process easier I would suggest you use vscode search panel to search all the instances of `name="` which will show you all the places you used icons, and replace them with their respective icon.
+
+
+## Changes in \<Icon />
```
kamleshchandnani, line 23: remove this section

packages/blade-old/MIGRATION.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+## Migrating existing blade-old components
+
+Few of the internal components were using the old \<Icon /> component, thus we also had to change the APIs for them.
+
+```sh
+# migrate-icons [directory_path_glob] [...jscodeshift_options]
+migrate-icons src/App/** parser=jsx
+```
```
kamleshchandnani, line 49: we probably don't need this since we are only focussing on migration here

packages/blade-old/MIGRATION.md
```diff
@@ -0,0 +1,195 @@
...
+
+```jsx
+import { Download } from "@razorpay/blade-old";
+
+<Download size="md" {...{}}>
+    <works />
+</Download>
+```
+
+### Limitations & Edge cases
```
kamleshchandnani, line 81:
```suggestion
### Manual Steps
```

Rename this to Manual steps and then start this section with the same content as it is now and then add the manual steps at the bottom of this section itself so you'll end up with just 2 sections

* Codemod usage
* Manual Steps

kamleshchandnani (changes requested):

packages/blade-old/MIGRATION.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+### Automigration
+
+We've written a codemod to automate the process of replacing the `icon` prop with imported icons. 
+
+Simply install the blade-old package and run
+
+```sh
+// migrate-icons [directory_path_glob] [...jscodeshift_options]
```
kamleshchandnani, line 15:
```suggestion
// migrate-icons [directory_path_glob] [...jscodeshift_options]
cd packages/blade-old
```

anuraghazra: Didn't get this 🤔 There won't be any `cd packages/blade-old`

The usage flow will be like this lets say:

1. User scaffolds a new universe react project
2. They used the older API for the Icons
3. They will upgrade blade-old
4. Then they will run the bin command directly `migrate-icons [path]` 
5. Icons will be migrated

packages/blade-old/MIGRATION.md
anuraghazra: Limitations are part of the codemod section, renamed it to `Limitations & Edge cases of codemod`

We now have
Codemod to migrate to new API
Limitations & Edge cases of codemod
Manual Migration

packages/blade-old/MIGRATION.md
```diff
@@ -0,0 +1,166 @@
...
+```jsx
+import { Download } from "@razorpay/blade-old";
+
+<Download size="md" {...{}}>
+    <works />
+</Download>
+```
+
+### Limitations & Edge cases of codemod
+
```
kamleshchandnani, line 40: add
There are some edge cases where its codemod won't work and hence you need to do those changes *manually*

packages/blade-old/MIGRATION.md
```diff
@@ -0,0 +1,166 @@
...
+Conditional rendering isn't supported
```
kamleshchandnani, line 41:
```suggestion
1. Conditional rendering isn't supported 
```

packages/blade-old/MIGRATION.md
```diff
@@ -0,0 +1,166 @@
...
+
+```ts
+<Icon
+  name={true ? "emptyCircle" : "alertCircle"}
+  fill="bg.950"
+  size="large"
+/>
+```
+
+Icon props on other components aren't handled.
```
kamleshchandnani, line 52:
```suggestion
2. Icon props on other components aren't handled. 
```

packages/blade-old/MIGRATION.md
```diff
@@ -0,0 +1,166 @@
...
+
+<Download size="md" {...{}}>
+    <works />
+</Download>
+```
+
+### Limitations & Edge cases of codemod
+
+Conditional rendering isn't supported 
+This code will not be transformed.
```
kamleshchandnani, line 42:
```suggestion

```

kamleshchandnani (approved):

PR #377, opened 2021-12-01
feat(icons): Added New direction Icon

kadarsh11:
## Description
This PR adds new direction icon

## Screenshot

Screenshot 2021-12-01 at 8 55 39 PM

### DIFF
--- a/packages/blade-old/src/icons/Direction/Direction.native.js
+++ b/packages/blade-old/src/icons/Direction/Direction.native.js
@@ -1,0 +1,30 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import Svg, { Path } from 'react-native-svg';
+import { getThemeColors } from '../../_helpers/theme';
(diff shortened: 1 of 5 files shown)


packages/blade-old/src/icons/Direction/Direction.native.js
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import Svg, { Path } from 'react-native-svg';
+import { getThemeColors } from '../../_helpers/theme';
+import colors from '../../tokens/colors';
+
+export default function Direction({ width, height, fill }) {
```
kamleshchandnani, line 7: since we are not spreading props when you apply flex and space on this Icon it'll break. This was raised by @anuraghazra here https://github.com/razorpay/blade/pull/367#discussion_r759962874

kadarsh11:
Need to change icon & icon name. Will create a new PR for this.

PR #381, opened 2021-12-03
Feat/add/payout link icon

kadarsh11:
## Description
-  Added PayoutLink icon.

## Screenshots
Screenshot 2021-12-03 at 4 07 11 PM

## Figma Link

https://www.figma.com/file/DfauMaJ2v1BKZZeswqInPg/%5BRazorpayX-Mobile-app%5D-%5Bv2%5D-Design-System?node-id=5464%3A21136

### DIFF
--- a/packages/blade-old/src/icons/PayoutLink/PayoutLink.native.js
+++ b/packages/blade-old/src/icons/PayoutLink/PayoutLink.native.js
@@ -1,0 +1,24 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import Svg, { Path } from 'react-native-svg';
+import { getThemeColors } from '../../_helpers/theme';
+import colors from '../../tokens/colors';
+
+export default function PayoutLink({ width = 24, height = 24, fill = colors.sapphire[800] }) {
+  return (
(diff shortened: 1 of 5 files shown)


packages/blade-old/src/icons/PayoutLink/PayoutLink.native.js
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import Svg, { Path } from 'react-native-svg';
+import { getThemeColors } from '../../_helpers/theme';
+import colors from '../../tokens/colors';
+
+export default function PayoutLink({ width = 24, height = 24, fill = colors.sapphire[800] }) {
```
kamleshchandnani, line 7: why do we need to add default color?

kadarsh11: It is just a fallback case, fill color is not provided it will be filled by this shapphire[800].

All the existing icon is also using this as a default color.

pritishvaidya (approved):

shridhar52war (approved):

PR #383, opened 2021-12-03
feat(icons): add codemod to migrate to new icon structure

anuraghazra:
- Added codemod for icons migration

This codemod automatically changes 

This
```
      <Icon size="md" name="download" {...{}}>
        <works />
      </Icon>
```

To this:

```jsx
      import { Download } from "@razorpay/blade-old";

      <Download size="md" {...{}}>
        <works />
      </Download>
```

### Limitations & Edge cases

Conditional rendering isn't supported 
This code will not be transformed.

```ts
      <Icon
        name={true ? "emptyCircle" : "alertCircle"}
        fill="bg.950"
        size="large"
      />
```

Icon props on other components aren't handled. 

```tsx
<Button  icon="chevronLeft">

<TextInput
  label="Label"
  iconRight="info"
  iconLeft="info"
/>

<Tabs.Tab
    value="payments"
    title="Payments"
    disabled={boolean('Disable Tab 1', false)}
    icon="info"
  >
<SegmentControl.Option
   value="1"
   icon="info"
   disabled={boolean('Disabled', false)}
   subText={text('Sub Text', undefined)}
>
  Option 1
</SegmentControl.Option>

snackbar.show({ 
  icon: "info",
})
```


kamleshchandnani:
@anuraghazra can you also add what is the limitation of this codemod? What are the edge cases that it doesn't cover and needs to be done manually?

.eslintignore
```diff
@@ -13,6 +13,7 @@ packages/blade/tokens.js
 packages/blade/tokens.d.ts
 packages/blade/utils.js
 packages/blade/utils.d.ts
+packages/blade-old/scripts
```
kamleshchandnani, line 16: Ignoring from eslint will also remove actual errors which are valid and also prettier will stop 😭

packages/blade-old/package.json
```diff
@@ -25,7 +28,8 @@
     "test:native": "BABEL_ENV=development jest -c ./jest.native.config.js",
     "test:native:coverage": "yarn run test:native --coverage",
     "test:web": "BABEL_ENV=web-test jest -c ./jest.web.config.js",
-    "test:web:coverage": "yarn run test:web --coverage"
+    "test:web:coverage": "yarn run test:web --coverage",
+    "postinstall": "cd ./scripts/icon-migrate-codemod && npm install"
```
kamleshchandnani, line 32:
```suggestion
    "postinstall": "cd ./scripts/icon-migrate-codemod && yarn"
```

also remove package-lock

packages/blade-old/scripts/icon-migrate-codemod/cli.js
```diff
@@ -0,0 +1,34 @@
...
+function run() {
+  const transformerPath = path.join(transformerDirectory, `blade-old-icon.ts`);
+
+  const parsedArgs = argParser(process.argv);
+  const input = process.argv.slice(2)[0];
+  const args = [input, '-t', transformerPath, ...process.argv.slice(3)];
+  const parser = parsedArgs.parser || 'jsx';
+
+  if (parser === 'tsx') {
+    args.push('--extensions=tsx,ts,jsx,js');
```
kamleshchandnani, line 17: why do we need tsx? for us default parse is jsx only right since everything in old-blade is jsx?

packages/blade-old/scripts/icon-migrate-codemod/transforms/blade-old-icon.ts
```diff
@@ -0,0 +1,84 @@
+import iconMap from "../iconMap";
+import {
```
kamleshchandnani, line 2: why create this as a new file? We are not going to use it elsewhere so why not put the content of this file under `cli.js` itself?

kamleshchandnani (changes requested):

.eslintignore
anuraghazra: Will scope it to only the ts files.

packages/blade-old/scripts/icon-migrate-codemod/cli.js
anuraghazra: It's for the applications which will be using this codemod, if they are on typescript they need to pass the parser arg in the cli `parser=jsx` so that jscodeshift can handle it.

packages/blade-old/scripts/icon-migrate-codemod/transforms/blade-old-icon.ts
anuraghazra: cli.js independent from the codemod, cli.js is the entrypoint to execute the `jscodeshift` 

jscodeshift works by passing a transformer file like so: 

```
jscodeshift -t ./transforms/blade-old-icon.ts
```

kamleshchandnani (approved):

PR #394, opened 2021-12-21
ci(blade): integrate chromatic

kamleshchandnani:


shubhnik (approved):

.github/workflows/blade-chromatic.yml
```diff
@@ -0,0 +1,35 @@
+# Workflow name
```
dhruvdutt, line 1: Remove this redundant comments

.github/workflows/blade-chromatic.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          install-command: yarn --frozen-lockfile
+      - name: Publish to Chromatic
+        uses: chromaui/action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          workingDir: packages/blade
+          buildScriptName: react:storybook:build
+          exitOnceUploaded: true
```
dhruvdutt, line 35: Wondering why this is `false` by default
https://github.com/chromaui/action/blob/main/action.yml#L84-L86

.github/workflows/blade-chromatic.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          node-version: 14
+      - name: Setup Cache & Install Dependencies
+        uses: bahmutov/npm-install@v1
+        with:
+          install-command: yarn --frozen-lockfile
+      - name: Publish to Chromatic
+        uses: chromaui/action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
chaitanyadeorukhkar, line 31: Why not use the `CI_BOT_TOKEN` here as well?

.github/workflows/blade-chromatic.yml
kamleshchandnani: Otherwise the deployment fails and CI check fails as well when there's a change.

.github/workflows/blade-chromatic.yml
chaitanyadeorukhkar: @kamleshchandnani bumping this. Do we need to use separate tokens in different steps?

kamleshchandnani: I'm not using separate token now. I've re-assigned `CI_BOT_TOKEN` to `GITHUB_ACCESS_TOKEN` [here](https://github.com/razorpay/blade/pull/394/files#diff-68c76d748a76d111004464c9f326c495d0be07056640f850ae3a37b3d7661344R5)

chaitanyadeorukhkar (approved):

PR #416, opened 2022-02-11
rfc: adaptive responsive layout approach

kamleshchandnani:
## [Formatted RFC](https://github.com/razorpay/blade/blob/rfc-adaptive-responsive-layout/rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md)

RFC on how to approach responsive and adaptive design


rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,391 @@
...
+
+<img alt="Responsive Modal" src="./images/responsive-adaptive-layout/adaptive-search-field.png">
+
+# Motivation
+
+### Why are we doing this?
+
+As of today, we don't have any strategy or approach in place for designing layouts for different screen sizes. Ensuring that elements fit within a page is not enough. There are different experiences that works for different screen sizes.
+
+Just going Responsive(Fluid) Design blindly is not correct. Imagine for example, you have different experiences(styles, javascript) for mobile vs desktop but if you just go blindly with responsive we might unnecessary download assets for all the screen sizes our application would run on. For example it doesn't makes sense to implement [this Search Field](#adaptive-search-field) as a responsive layout since it'll result in unnecessary download of the assets(modal for mobile) and hence hampering the experience and performance for the consumers.
```
saurav12, line 73: Improper capitalisation. Instead, it should be `Just going Responsive(fluid) design`

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,391 @@
...
+### What is the expected outcome?
+
+We'll basically define the approach and strategy we'll be taking in Blade to build components that will address the concerns of rendering based on different screen sizes.
+
+# Detailed Design
+
+## Responsive Design
+
+- Responsive is fluid and adapts to the size of the screen no matter what the target device is.
+- Responsive uses [CSS media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries) to change styles based on the target device such as display type, width, height, etc., and only one of these is necessary for the site to adapt to different screens.
```
saurav12, line 94: We can alt-code these in the statement:
display `type`, `width`, `height`, etc.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,391 @@
...
+- Ensuring that elements fit within a page is not enough. For a responsive design to be successful, the design must also be usable at all screen resolutions and sizes.
```
saurav12, line 95: Can we also attach some good/bad visual examples here?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,391 @@
...
+|                               | Tons of `@media` queries often leads to a maintenance overhead in code                                                                                                                                                                                                                       |
+|                               | Images and other high bandwidth features are simply scaled down rather than resized or eliminated outright(unless added some lines of JS which again degrades the performance and increases the mess in code), elements which may not be vital to user experience can cause poor performance |
+
+## Adaptive Design
+
+- While responsive design relies on changing the design pattern to fit the real estate available to it, **adaptive design** has multiple fixed layout sizes. When the site _detects_ the available space, it _selects_ the layout most appropriate for the screen. So, when you open a browser on the desktop, the site chooses the best layout for that desktop screen, resizing the browser has no impact on the design.
```
saurav12, line 108: When we say:
> resizing the browser has no impact on the design
Does that also mean that the resizing in-between desktop size (let's say from `1380px` to `1440px`) will have no impact? Or does it work in a responsive manner in-between similar desktop sizes?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,391 @@
...
+- Adaptive works to detect the screen size and load the appropriate layout for it – generally you would design an adaptive site for six common screen widths(confirm with Saurav once on exact sizes we are doing and update):
+  - 320
```
saurav12, line 110: We might replace it with `350px` instead as <1% of devices today are on the `320px` screen size.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,391 @@
...
+  - 480
+  - 760
+  - 960
```
saurav12, line 113: We might club `760px` & `960px` together. Need to dig deeper while researching on tablet-sized devices.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,391 @@
...
+- Most of the components that don't deal with layouts shall handle responsiveness internally. For eg: Button, Modal, Avatar, Select, etc.
+
+- The properties that affects responsiveness are:
+
+  - `width`
+  - `height`
+  - `font-size`
+  - `padding`
+  - `margin`
+
```
saurav12, line 155: I guess `position` property might also change, right?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,391 @@
...
+      </Stack>
+      <Hidden hide={{ s: false, l: true }}>
+        <BottomNavBar />
+      </Hidden>
+    </Stack>
+  );
+  ```
+
+  So here we have introduced a new utility component called `Hidden` which accepts responsive props and enables us to hide/show component based on different screen sizes
+
```
saurav12, line 346: The Blade code-side is amazing! Very well explained 👏

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,391 @@
...
+   - `margin`
+
+2. If the component is adaptive, then design it for all the screens and explicitly mention about the multiple variations of a component for different screen sizes.
+
+### On Consumer Design Side
+
+When designing layouts and consuming components from Design System Library on Figma make sure to do following things:
+
+1. Design the mockups for all the different screen sizes.
+2. Pick the right version of the component for respective screen sizes.
```
saurav12, line 366: By using internal component `variants`.

kamleshchandnani: didn't understand this part? 🤔

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: We can decide that. Ideally it will resize based on the breakpoints we give instead of resizing on every pixel when we resize

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: Yes I want input on this piece from you. we'll catchup and decide what screen sizes we need to support

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
sidag95: @kamleshchandnani @saurav12 [Find the list of breakpoints we have used in the docs-website ](https://github.com/razorpay/docs/blob/master/docs-website/src/components/utils/breakpoints.ts#L1) (please ignore the breakpoints names 🙈)

These have worked well for us

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,394 @@
...
+- The properties that affects responsiveness are:
+
+  - `width`
+  - `height`
+  - `font-size`
+  - `padding`
+  - `margin`
+  - `position`
+
+- While a lot of [well known design systems gives flexibility](#approaches-taken-by-well-known-design-systems) on every component for above properties to accept responsive values, I can't think of a use case where we'll have to provide similar functionality. Majority of the cases that I could think of wouldn't be any different even if the consumers define the values for them explicitly. So seems like we'll be adding unnecessary learning curve to the component's API.
```
divyanshu013, line 157: This would need strict design alignment which I'm doubtful of. For example on website there are usecases for different font sizes based on breakpoints. Similarly for other properties too.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
sidag95: Even positional values like top,left,right,bottom might be different

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,394 @@
...
+
+### On Blade Code Side
+
+- To begin with, as a first step we will first need to setup different screen sizes that we'll support and store it in tokens. This will handle our layout changes
+
+  ```
+  // confirm with Saurav once
+  s: 320
+  m: 480
+  l: 760
```
divyanshu013, line 197: Why not a more traditional `sm, md, lg`?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,394 @@
...
+  xl: 960
+  2xl: 1200
+  3xl: 1600
+  ```
+
+- We would then need to add a listener in the `BladeProvider` which will listen to screen size changes and trigger a layout change.
+
+- For handling typography for different screen sizes we already have 2 scales one for smaller screens which are less than `960px` and the other for larger screens which are greater than `960px`. Now, once the screen size changes, the above listener would trigger a state change and then we can listen to it and swap the scales between small and large screens.
```
divyanshu013, line 205: This sounds like it may "stutter / freeze" the UI if someone switches between different resolutions (for example decrease browser width) (compared to responsive layout with media queries since now transitions would be more resource intensive) 🤔

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,394 @@
...
+
+  - For eg: all the props(ones that are affected by responsive) for a component will then accept either single value or array/object of values for different screen sizes which is confusing and might lead to a lot of confusion at the API surface.
+
+    ```jsx
+    <Text fontSize="100" />
+
+    <Text fontSize={{ small: 100, medium: 100, large: 100, xl: 400 }} />
+    ```
+
+- Hence the idea here is that all these components except the Layout components will handle responsiveness based on predefined breakpoints internally without the consumers explicitly defining anything for them.
```
sidag95, line 167: @kamleshchandnani The components handling responsiveness on there own seem analogous to having design tokens, i.e. these promote consistency among all the consumer projects. If that is the case, then to adopt the Blade components, the respective designs will also need to be updated. How are we exposing these components along with the defined presents in figma?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
sidag95: Won't media queries be used for standard CSS changes for responsive components? 

I feel the listener would be useful for adaptive components. 

That is how we have solved this in docs. We use [getBreakpointStyles](https://github.com/razorpay/docs/blob/master/docs-website/src/components/utils/breakpoints.ts#L45) function to add media queries to each responsive components, like [here for setting responsive padding](https://github.com/razorpay/docs/blob/master/docs-website/src/home/Payments/Payments.tsx#L25)

Whereas we use [useBreakpoint](https://github.com/razorpay/docs/blob/master/docs-website/src/components/utils/useBreakpoint.ts) for adaptive components

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: naming is still not decided, this is just to set the context 😄

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: There's a slight difference WRT tokens vs components here. so the tokens will define the scale of spacing/margin/sizing for these components on different resolutions. Coming to design side they actually don't have to do much here because for them they'll always have to create multiple mockups because AFAIK we don't have that responsiveness in design tools yet like we do in browsers(except the auto layout).

On Figma, these components will have a property of screen size as a variant which the designers will toggle when placing the components in different screen frames

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: yes so the idea here is that for typography and a few other components we'll handle it based on the defined screen sizes(breakpoints) and internally toggle based on the screen size that the app will be running on. So the consumers don't have to worry about these things everytime. If you read below I'm proposing that for few components we'll handle internally but for few like layout things we'll give the flexibility to the consumers

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: > This sounds like it may "stutter / freeze" the UI if someone switches between different resolutions (for example decrease browser width) (compared to responsive layout with media queries since now transitions would be more resource intensive) 🤔

Why will it stutter? It's same as responsive only diff is we will switch things based on the defined breakpoints and not on every resize events

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
sidag95: Got it. So let's say they have different designs for different screens, how will the component dimensions look to the designers when they set a particular variant based on the screen size?

My question is, will setting the dimensions for these components in the design, be a manual process?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
sidag95: What is the advantage of doing this using JS rather than using media queries for responsive components?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: okay, so 2 parts over here:

1. The design on Figma for Design System side
2. The designs for the consumer side on Figma

So for the first case, assume if the components have a width that varies then the designer on the design system side will publish components such that it has variable width(auto-layout) and then when the consumer side designer uses it on any screen it won't matter

Another scenario where the components are different on different screen sizes for eg: modal with max-width property which takes full width on mobile and 80% on a desktop the designer on the design system side will design 2 different elements(even though it's a responsive modal) since we don't have matured design tools yet :D and then the designer on the consuming side need to use the version of the Modal based on the screen size they pick. Unfortunately, it's a manual process because again the limitation on how the current design tools are built.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
saurav12: I meant to say instead of writing:
> Pick the right version of the component for respective screen sizes.

We can end the statement by saying:
> Pick the right version of the component for respective screen sizes using component variants in Figma.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
saurav12: @sidag95 Yes, it's a manual process but designers just need to change the property in Figma's right panel to change the design from large to small screen.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
saurav12: @kamleshchandnani Here are a few **breakpoints** we should have in our system:
- `320px — 480px`: Mobile devices
- `481px — 768px`: Small tablets
- `769px — 1024px`: Small screens, Large Tablets
- `1025px — 1200px`: Desktops
- `1200px and above` —  Extra large screens, TVs

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+|                               | Images and other high bandwidth features are simply scaled down rather than resized or eliminated outright(unless added some lines of JS which again degrades the performance and increases the mess in code), elements which may not be vital to user experience can cause poor performance |
+
+## Adaptive Design
+
+- While responsive design relies on changing the design pattern to fit the real estate available to it, **adaptive design** has multiple fixed layout sizes. When the site _detects_ the available space, it _selects_ the layout most appropriate for the screen. So, when you open a browser on the desktop, the site chooses the best layout for that screen, resizing the browser has no impact on the design(until it hit the breakpoint if we have mentioned it).
+- Adaptive works to detect the screen size and load the appropriate layout for it – generally you would design an adaptive site for six common screen widths(confirm with Saurav once on exact sizes we are doing and update):
```
chaitanyadeorukhkar, line 111: Have you confirmed with @saurav12 ? Lets remove this before merging

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
sidag95: Understood. Thanks for the detailed explanation.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+
+- Responsive is fluid and adapts to the size of the screen no matter what the target device is.
+- Responsive uses [CSS media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries) to change styles based on the target device such as display `type`, `width`, `height`, etc., and only one of these is necessary for the site to adapt to different screens.
+- Ensuring that elements fit within a page is not enough. For a responsive design to be successful, the design must also be usable at all screen resolutions and sizes.
+  - An example of fluid search field which shrinks/grows based on screen width but isn't usable enough.
+    <img alt="Responsive Modal" src="./images/responsive-adaptive-layout/responsive-search-field.png">
+
+### Pros and Cons of Responsive Design
+
+| Pros                          | Cons                                                                                                                                                                                                                                                                                         |
```
RazorNish, line 101: Why have we written pros and cons in a table? They are not comparative, they should be a list.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
RazorNish: Can we add these details to the RFC, look like arbitrary break points without that

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+  - `font-size`
+  - `padding`
+  - `margin`
+  - `position`
+  - `top`
+  - `left`
+  - `right`
+  - `bottom`
+
+- While a lot of [well known design systems gives flexibility](#approaches-taken-by-well-known-design-systems) on every component for above properties to accept responsive values, I can't think of a use case where we'll have to provide similar functionality. Majority of the cases that I could think of wouldn't be any different even if the consumers define the values for them explicitly. So seems like we'll be adding unnecessary learning curve to the component's API.
```
RazorNish, line 163: Could you establish why the other design systems leave this flexibility and how our use case is different.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+To render content on different device screen resolutions there are different strategies. Modern web design gives us three options to use: Responsive, Adaptive, and Standalone Design, although standalone has fallen into disuse because of the overhead it comes with to maintain multiple codebase and their development lifecycle hence, we would just be focussing on Adaptive and Responsive.
+
+1. Responsive - Make the content fluid so it'll render based on the available space
+2. Adaptive - Make content targeted for specific screen sizes so it'll render exactly like that, regardless of the available space.
+
+This RFC will focus on what approach and strategies we shall take in our Design System.
+
+# Basic Example
+
+### Responsive Modal
```
chaitanyadeorukhkar, line 50: Instead of specifying "Modal" let's just call it an example for "Responsive Design". Looking at "Responsive Modal" I was expecting an "Adaptive Modal" after this.
```suggestion
### Responsive Design
```

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+
+👇🏻 An example of a responsive design where we have modal which is responsive to the screen size and has a `maxWidth` defined for it to render on larger screen sizes, where on smaller screen sizes(mobile) it'll take `100%` of the width of the screen.
+
+<img alt="Responsive Modal" src="./images/responsive-adaptive-layout/responsive-modal.png">
+
+<br/>
+
+### Adaptive Search Field
```
chaitanyadeorukhkar, line 58:
```suggestion
### Adaptive Design
```

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,394 @@
...
+
+<img alt="Responsive Modal" src="./images/responsive-adaptive-layout/adaptive-search-field.png">
+
+# Motivation
+
+### Why are we doing this?
+
+As of today, we don't have any strategy or approach in place for designing layouts for different screen sizes. Ensuring that elements fit within a page is not enough. There are different experiences that works for different screen sizes.
+
+Just going Responsive(fluid) design blindly is not correct. Imagine for example, you have different experiences(styles, javascript) for mobile vs desktop but if you just go blindly with responsive we might unnecessary download assets for all the screen sizes our application would run on. For example it doesn't makes sense to implement [this Search Field](#adaptive-search-field) as a responsive layout since it'll result in unnecessary download of the assets(modal for mobile) and hence hampering the experience and performance for the consumers.
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: +1

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+
+Similarly, just going adaptive blindly is not correct too. We might unnecessary be creating redundant variant of the same component. For example, imagine that we are creating 2 variants of a modal - 1 for desktop(which has a width of `600px`) and 1 for mobile(which has `100%` width). Now we could have just made it responsive by setting the `maxWidth` property of the modal so it could have been `600px` in width on desktop and `100%` in width on mobile. For eg, it doesn't makes sense to create multiple variant of [this Modal](#responsive-modal) while we can achieve this through responsive layout by setting `maxWidth` for Modal's Dialog component.
+
+But because we don't have these strategies in place we don't know how to approach building layouts and components for different screen sizes.
+
+### What use cases does it support?
+
+- Handling responsiveness internally in the components based on different screen sizes.
```
chaitanyadeorukhkar, line 81: Let's stick to similar language for our 3 cases everywhere?
```suggestion
- Components that can handle Responsiveness Internally
```

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+- Layout components that will help you define different layouts based on screen sizes.
```
chaitanyadeorukhkar, line 82:
```suggestion
- Components that can give some flexibility to handle Responsiveness to consumers
- ```

File: rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md [hunk 22]
```diff
@@ -0,0 +1,403 @@
...
+- Rendering different variants of a component based on screen sizes, optimising for performance without bloating the bundlesize.
```
chaitanyadeorukhkar, line 83: Had to rephrase this a little since "Components for Adaptive layout(rendering components conditionally)" doesn't fit well here
```suggestion
- Components that render different variants conditionally
```

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+| ----------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Abundance of templates to use | Elements can move around fluidly                                                                                                                                                                                                                                                             |
+| Often easier to implement     | Longer mobile download times.                                                                                                                                                                                                                                                                |
```
dhruvdutt, line 104: Why only longer download times on mobile? It'll impact both platforms since the bundle would be larger.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+|                               | Tons of `@media` queries often leads to a maintenance overhead in code                                                                                                                                                                                                                       |
+|                               | Images and other high bandwidth features are simply scaled down rather than resized or eliminated outright(unless added some lines of JS which again degrades the performance and increases the mess in code), elements which may not be vital to user experience can cause poor performance |
```
dhruvdutt, line 106: This statement is a little complicated to understand. Can you simplify this?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+
+- While responsive design relies on changing the design pattern to fit the real estate available to it, **adaptive design** has multiple fixed layout sizes. When the site _detects_ the available space, it _selects_ the layout most appropriate for the screen. So, when you open a browser on the desktop, the site chooses the best layout for that screen, resizing the browser has no impact on the design(until it hit the breakpoint if we have mentioned it).
+- Adaptive works to detect the screen size and load the appropriate layout for it – generally you would design an adaptive site for six common screen widths(confirm with Saurav once on exact sizes we are doing and update):
+  - 320
+  - 480
+  - 760
+  - 960
+  - 1200
+  - 1600
+- Because adaptive allows for a different experience between the mobile and desktop versions, products that want to deliver a different experience for their mobile customers can do so without sacrificing quality in either version. It can also be tacked on to an existing site, whereas responsive layout requires rewriting all the original HTML code in the back end.
```
dhruvdutt, line 118: Can you remove "in the back end" or change to something else?
It's confusing what back end means in this context.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: but essentially on desktop it barely hampers the performance to the end users. can rephrase it

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+|                               | Tons of `@media` queries often leads to a maintenance overhead in code                                                                                                                                                                                                                       |
+|                               | Images and other high bandwidth features are simply scaled down rather than resized or eliminated outright(unless added some lines of JS which again degrades the performance and increases the mess in code), elements which may not be vital to user experience can cause poor performance |
+
+## Adaptive Design
+
+- While responsive design relies on changing the design pattern to fit the real estate available to it, **adaptive design** has multiple fixed layout sizes. When the site _detects_ the available space, it _selects_ the layout most appropriate for the screen. So, when you open a browser on the desktop, the site chooses the best layout for that screen, resizing the browser has no impact on the design(until it hit the breakpoint if we have mentioned it).
```
dhruvdutt, line 110: > resizing the browser has no impact on the design(until it hit the breakpoint if we have mentioned it)

Didn't get this part. So, with this approach, we would download the appropriate bundle/styles when the system/server gets the request based on the initial screen height/width, right?
How would things change while resizing then even after a new breakpoint is hit?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,403 @@
...
+
+## Approaches taken by well-known Design Systems
+
+| Name                                                                                                    | Responsiveness handled internally/flexibility to consumers?                                                                          |
+| ------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| [Paste - Twilio](https://paste.twilio.design/primitives/text/#responsive-text)                          | Flexible with `[]` props accepting values for breakpoints on various components. Uses styled system under the hood                   |
+| [Primer - GitHub](https://primer.style/react/core-concepts#responsive-props)                            | Flexible with `[]` props accepting props accepting values for breakpoints on various components. Uses styled system under the hood   |
```
dhruvdutt, line 136:
```suggestion
| [Primer - GitHub](https://primer.style/react/core-concepts#responsive-props)                            | Flexible with `[]` props accepting values for breakpoints on various components. Uses styled system under the hood   |
```

chaitanyadeorukhkar:
As far as I understand our approach is - we allow to do it all, responsive and adaptive.

Our strategy consists of
1. **_Automatic_ Responsive**: Components will change pre-defined properties based on pre-defined screen sizes automatically
2. **_Configurable_ Responsive**: Components will change pre-defined properties based on pre-defined screen sizes as per a configuration passed by the consumer. Eg) `direction={{ s: 'column', l: 'row' }}`
3. **_Configurable_ Adaptive?**: Components will change entirely based on the consumer's logic for screen sizes. Eg) `{isMobile ? <SmallSearchInput /> : <LargeSearchInput />}`

#### I have 2 broad questions,

#### Why is there no _Automatic_ Adaptive?
As per the example mentioned for `LargeSearchInput` & `SmallSearchInput`, we expect our consumers to import both of these(in a bundle optimized manner) and apply their own logic to render either of the two (this is why I am calling it "Configurable" since the control is on the consumer end)

Similarly, can there not be components that are **_Automatically_ Adaptive**? For example, a SearchInput that internally renders `LargeSearchInput` or `SmallSearchInput` based on the screen size. 

A valid use-case here could be a `Dropdown` component where we _know_(*assuming it here) that we always want the dropdown to be almost fullscreen on mobile but a relatively small pop-over menu on desktop. Consumers can just import `Dropdown` and blade can internally render the `MobileDropdown` or `DesktopDropdown` based on the current screen size and take away the complexity of adding any logical checks for the consumers.

#### How do we communicate to consumers that a component they are using is Responsive or Adaptive?
As a consumer of blade, how would I know if a component **ComponentA** is automatically responsive or I need to pass it some props for it to become responsive or is it adaptive and I need to import another **SmallComponentA** and add `isMobile` check to be able to render it properly on a mobile device?

Supporting multiple strategies can introduce a huge learning curve for the consumer. Something as simple as trying to use a prop like `direction={{ s: 'column', l: 'row' }}` on an Automatic Responsive component. Eg) Typography is automatically responsive but what if a consumer tries to pass `fontSize={{ small: 100, medium: 100, large: 100, xl: 400 }}` and expects it to work? 

We can throw warnings or show TS errors but in the end, our components behave differently from one another. Where a `Text` component automatically takes care of screen sizes on its own, we are expecting our consumers to still worry about screen sizes because the `SearchInput` component is not going to look different on different screen sizes until the consumer adds conditional rendering to it. 

My concern is not the fact that we're allowing all strategies but the added overhead for consumers to think and understand each component's behavior with screen sizes deeply before using them anywhere.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: @saurav12 while defining do you think we'll have to mention for 1201 and above? If yes we need a deterministic breakpoint. This is what I have for now

```
s:  480px // Mobile devices
m:  768px // Large mobile devices
l: 1024px // Small screens, Tablets
xl: 1200px // Desktops
```

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: my bad. rephrasing this to

> a reload would be required to see the change

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: because it's a choice between flexibility vs constraints. I'll add this

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,414 @@
...
+- To begin with, as a first step we will first need to setup different screen sizes that we'll support and store it in tokens. This will handle our layout changes
+
+  ```
+  s:  480px // Mobile devices
+  m:  768px // Large mobile devices
+  l: 1024px // Small screens, Tablets
+  xl: 1200px // Desktops
+  ```
+
+- We would then need to add a listener in the `BladeProvider` which will listen to screen size changes and trigger a layout change.
```
aditya-razorpay, line 220: question: how will adaptive layout work with SSR? We would want to send only mobile bundle for mobile users but media query info is not available on server.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,414 @@
...
+
+- While most of the components can handle responsiveness internally there are use cases for components where we would need to give flexibility to accept responsive values for certain props. For eg: Layout Components that handles the overall layout of the page might change the layout on different screen sizes based on the available space.
+  <img alt="Adaptive Layout with flex direction" src="./images/responsive-adaptive-layout/adaptive-layout-with-flex-direction.png">
+- The idea here is to have certain components that deal with Layout of the page expose some flexibility to consumers to tweak if they want the layout to look different on different screen sizes
+
+### 3. Components for Adaptive layout(rendering components conditionally)
+
+- After handling responsiveness internally and also giving flexibility for few scenarios there are still few complex cases that we need to handle:
+
+  - There are cases where we might need to render a completely different variation of a component based on the screen size and user experience. For eg: A Search Field component might be different for larger screen sizes and smaller screen sizes.
```
anuraghazra, line 199: > A Search Field component might be different for larger screen sizes and smaller screen sizes.

I think the consideration here should not be just about the "size" but also the device and how users interact with the component while optimising for the best user experience. 

We usually see a dropdown on desktop becoming a modal on mobile device to improve the overall UX of the component since mobile devices has a very small touch area thus trying to select a floating dropdown can be cumbersome. 

Also on another note:
Some components can be hybrid means they can be "Responsively Adaptive" 

Take an example of a Select component where the Select's input can be fluid, means it will stretch based on screensize but will also have the functionality to change the collapsible dropdown to modal on mobile devices.

Ideally we should have a list of all the component which needs adaptive strategy, and ones which needs both strategies.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,414 @@
...
+  <img alt="Responsive Modal" src="./images/responsive-adaptive-layout/adaptive-search-field.png">
+
+  <br/>
+
+  Here's how the consumer will create a layout like this in code(_Note: the actual API might differ_).
+
+  ```jsx
+  import { LargeSearchInput, SmallSearchInput } from '@razorpay/blade/components';
+
+  const SearchScreen = () => (isMobile ? <SmallSearchInput /> : <LargeSearchInput />);
```
anuraghazra, line 278: **Point 1**

This looks fine but I'm more concerned that, A naive user of a component library shouldn't care about codesplitting the components with this much granularity. 

They should just be able to say `<SearchInput />` and the input should handle the isMobile logic internally. 


**Point 2**

Second point is we didn't to mention the internal implementation, which in this case i believe is very important. Let's replace the `<SearchInput />` with something more complex component say `<Select />`

Open ended questions: 

- How will the Select's core business logic be shared between both SmallSelectInput & LargeSelectInput? We certainly don't want to repeat the same code in both places. We need some sort of `BaseSelect` component or even [hooks based renderless approach](https://survivejs.com/blog/renderlesskit-interview/) for this, where all the internal logic should be in `useSelectComponent` hook so that building both variants of the component just becomes a work of adding few different styling in most cases thus adding different rendering logic becomes easy. 

- Are there any significant changes on the behaviour of the component? (Ideally it shouldn't)

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,414 @@
...
+  ```jsx
+  // With React.Lazy
+  import { Suspense } from 'react';
+
+  const LargeSearchInput = React.lazy(() => import('@razorpay/blade/components'));
+  const SmallSearchInput = React.lazy(() => import('@razorpay/blade/components'));
+
+  const SearchScreen = () => (
+    <Suspense fallback={<div>Loading...</div>}>
+      {isMobile ? <SmallSearchInput /> : <LargeSearchInput />}
```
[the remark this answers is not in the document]
kamleshchandnani: This is a similar question to what Chaitanya has posted above. While it can do that but there are a couple of downsides to it:
1. Every client might have a different way of lazy loading. for csr people might use `React.Lazy` but for SSR they might have to use `loadable` components. Now the moment we generalize this people will have concerns and differences of opinions plus unnecessary dependencies because I can just use `React.lazy` for csr why shall someone install loadable. 
2. Also there are high chances that the props will be completely different because the LargeSearchInput might have auto suggestions as drop down but the small search input might have recent searches which might come from a backend API so if we handle this internally we'll have to create another wrapper which will accept the props for both the variant of the component.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,414 @@
...
+
+  const DesktopNavBar = loadable(() => import('@razorpay/blade/components'));
+  const MobileNavBar = loadable(() => import('@razorpay/blade/components'));
+  const BottomNavBar = loadable(() => import('@razorpay/blade/components'));
+
+  const Dashboard = () => (
+    <Stack direction={{ s: 'column', l: 'row' }}>
+      {isMobile ? <MobileNavBar /> : <DesktopNavbar />}
+      <Stack direction="column">
+        <Hidden hide={{ s: true, l: false }}>
```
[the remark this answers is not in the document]
kamleshchandnani: yes this is just an example 🙂 . The idea is similar to what you just mentioned.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: @anuraghazra that's the summary of the entire RFC, not sure if you meant to add something on this line?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,414 @@
...
+    - [On Blade Design Side](#on-blade-design-side)
+    - [On Consumer Design Side](#on-consumer-design-side)
+- [Adoption strategy](#adoption-strategy)
+- [How do we educate people?](#how-do-we-educate-people)
+- [Open Questions](#open-questions)
+- [References](#references)
+
+# Summary
+
+There are multiple strategies to render content on devices with different screen resolutions. Modern web design gives us three options: Responsive, Adaptive, and Standalone Design. Of these, standalone is rarely used because of the overhead of maintaining multiple codebase and their development lifecycle. Hence, we would just be focussing on Adaptive and Responsive.
```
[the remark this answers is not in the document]
kamleshchandnani: that would be an implementation detail of how we implement the responsive components

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: Point 1 is similar to what @chaitanyadeorukhkar asked [here](https://github.com/razorpay/blade/pull/416#issuecomment-1049802978)

Point 2 is out of scope of this RFC. That will be covered in code sharing RFC which will be coming soon. That is not just limited to share logic for similar components but also across platforms(react web and react-native)

kamleshchandnani:
> #### Why is there no _Automatic_ Adaptive?
> As per the example mentioned for `LargeSearchInput` & `SmallSearchInput`, we expect our consumers to import both of these(in a bundle optimized manner) and apply their own logic to render either of the two (this is why I am calling it "Configurable" since the control is on the consumer end)
> 
> Similarly, can there not be components that are **_Automatically_ Adaptive**? For example, a SearchInput that internally renders `LargeSearchInput` or `SmallSearchInput` based on the screen size.
> 
> A valid use-case here could be a `Dropdown` component where we _know_(*assuming it here) that we always want the dropdown to be almost fullscreen on mobile but a relatively small pop-over menu on desktop. Consumers can just import `Dropdown` and blade can internally render the `MobileDropdown` or `DesktopDropdown` based on the current screen size and take away the complexity of adding any logical checks for the consumers.

We can do this internally it's not a big deal but that's exactly a reason why it's adaptive and not responsive. if it was just scaling or little things to change we could have made it responsive but adaptive itself means the implementation(UX, interactions, etc) could be different for the mobile variant vs a desktop variant.

for eg: I have a search field on for desktop which shows auto suggestions as a dropdown but its mobile variant is a full-screen modal that has filters in it, recent searches section(whose results can be fetched via some API) now I can't handle this use case internally else I've to create another wrapper which will expect all the things from consumers for supporting both the variants but it seems like messing our component API.

Surely we can consider this approach where the APIs will be the same for both the variants but then it's just the matter of why can't it be then a responsive component if only the layout changes. 

Even if we go ahead with this approach we might have to consider the developer(on design system side) overhead where they'll be confused of when to handle `isMobile` internally vs leaving it to the end consumers because we can't avoid different APIs for different variants.

But we can surely keep this approach into consideration and while this RFC is mostly about an overview of how we'll handle different strategies for supporting these layouts the implementation details are still left out and we'll take it on case by case basis.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: finally, [answered here](https://github.com/razorpay/blade/pull/416#issuecomment-1056790157)

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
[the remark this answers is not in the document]
kamleshchandnani: finally, [answered here](https://github.com/razorpay/blade/pull/416#issuecomment-1056790157) in detail

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
anuraghazra: Yup got it, Was reviewing async, didn't read the entire thing while commenting this. 

But second paragraph about "Responsively Adaptive" 
We are going to be adopting hybrid strategy if needed right?

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
anuraghazra: > Also our existing tokens use s, m, l
Oops! naming is hard indeed.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
kamleshchandnani: yes that's what we have [mentioned here](https://github.com/razorpay/blade/pull/416/files#diff-f4d103560de689c11de7f41e8dbd9106c41e45f89247d0d2508f9150040c463dR85)

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,487 @@
...
+  ```jsx
+  // With React.Lazy
+  import { Suspense } from 'react';
+
+  const LargeSearchInput = React.lazy(() => import('@razorpay/blade/components'));
+  const SmallSearchInput = React.lazy(() => import('@razorpay/blade/components'));
+
+  const SearchScreen = () => (
+    <Suspense fallback={<div>Loading...</div>}>
+      {isMobile ? <SmallSearchInput /> : <LargeSearchInput />}
```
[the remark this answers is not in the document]
saurabhdaware: Also since `isMobile` in your example is not a hook. Are we saying that this logic won't really listen to the screen change events that provider is making?

saurabhdaware (commented):
Sorry for the late review people 🙈 The RFC looks great!!!!

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,487 @@
...
+
+  const DesktopNavBar = loadable(() => import('@razorpay/blade/components'));
+  const MobileNavBar = loadable(() => import('@razorpay/blade/components'));
+  const BottomNavBar = loadable(() => import('@razorpay/blade/components'));
+
+  const Dashboard = () => (
+    <Stack direction={{ s: 'column', l: 'row' }}>
+      {isMobile ? <MobileNavBar /> : <DesktopNavbar />}
+      <Stack direction="column">
+        <Hidden hide={{ s: true, l: false }}>
```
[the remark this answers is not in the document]
saurabhdaware: Oh I just saw comment that said this is an overview and implementation details will be taken separately. So I'm guessing this comment is out of scope.

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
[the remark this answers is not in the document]
saurabhdaware: Oh saw the comment above, guessing this is out of scope for this RFC.

chaitanyadeorukhkar (approved):
LGTM 🚀 

My understanding: 

- Why no automatic adaptive? 
We cannot always internally handle adaptive components due to the high possibility of differences in APIs for a mobile & desktop version
- Won't consumers get confused due to different strategies? 
Yes. The pros outweigh the cons here. One easy way to understand different strategies would be that it's always Automatic Responsive unless it's a component thats a Wrapper which enables you to create layouts like Spacing, Flex, etc. 
For adaptive components, we will use naming conventions like `DesktopDropdown`/`MobileDropDown` or `LargeDropDown`/`SmallDropDown` to make it verbose enough that the component being used is intended for a specific type of device.

dhruvdutt (dismissed):

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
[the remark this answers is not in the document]
kamleshchandnani: The idea is to provide `isMobile` from within Blade itself which will take care of event listeners and stuff. This is jus the pseudo code to conceptualise Adaptive vs Responsive

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,487 @@
...
+<br/>
+
+Here's how the consumer will create a layout like this in code(_Note: the actual API might differ_)
+
+```jsx
+import { Stack, Card, Title } from '@razorpay/blade/components';
+
+const Dashboard = () => (
+  <Stack direction={{ s: 'column', l: 'row' }}>
+    <Card>
```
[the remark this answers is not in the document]
kamleshchandnani: That will the responsibility of the `Stack`/any similar layout component, we are just discussing adaptive and responsive strategy 🙂

rfcs/2022-02-11-responsive-and-adaptive-layout-strategy.md
```diff
@@ -0,0 +1,487 @@
...
+#### Pros
+
+- Allows designers to build the best UX for the appropriate device.
+- The system can evaluate the consumers screen resolution and send the appropriate content for them hence, preventing downloading unnecessary assets required for that screen size.
+- It offers an alternative to the “one-size-stretches-to-all” approach.
+
+#### Cons
+
+- Takes a lot of effort to implement since now we essentially are implementing multiple versions of the same component for different screen sizes.
+- Can leave out users “in the middle” if the screen sizes are not properly defined and the content is not available for a particular screen size
```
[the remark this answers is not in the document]
kamleshchandnani: yes

PR #424, opened 2022-03-07
feat(blade): add listener for toggling breakpoints

kamleshchandnani:
This PR adds the following:

* Add `breakpoints` token to the themes.
* Out of the box responsiveness support for typography tokens.
* Publish `useBreakpoint` hook.
* Following breakpoints are supported as of today.
  ```
  /** max width: 320px  */
  xs: 320;
  /** min width: 321px and max width: 480px */
  s: 480;
  /** min width: 481px and max width: 768px */
  m: 768;
  /** min width: 769 and max width: 1024px */
  l: 1024;
  /** min width: 1025 and max width: 1200px */
  xl: 1200;
  /** min width: 1201px  */
  max: 1201;
  ```
* For web the typography scale will toggle between mobile and desktop
  * `xs, s, m` are considered as mobile
  * `l, xl, xl` are considered as desktop
* For react native we always default to mobile typography scale


packages/blade/src/components/BladeProvider/BladeProvider.tsx
```diff
@@ -29,9 +34,12 @@ const BladeProvider = ({
   }
 
   const { colorScheme, setColorScheme } = useColorScheme(initialColorScheme);
+  const { matchedDeviceType } = useBreakpoint({
+    breakpoints: themeTokens.breakpoints,
+  });
```
kamleshchandnani, line 39: This registers an event listener for media query changes. So anytime the media query changes it'll re-render the entire children tree wrapped under `BladeProvider`

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,144 @@
...
+              } else {
+                mediaQuery = `screen and (max-width: ${breakpointSize}px)`;
+              }
+
+              return [breakpointTokenName, breakpointSize, mediaQuery];
+            },
+          )
+        : []) as [keyof Breakpoints, number, string][],
+    [breakpoints, supportsMatchMedia],
+  );
```
kamleshchandnani, line 38: the output of this function will look something like following if the device supports `matchMedia`

```
[
  ['xs', '320', `screen and (max-width: 320px)`]
  ['s', '480', `screen and (min-width:321px) and (max-width: 480px`]
  .
  .
  .
  ['max', '1201', `screen and (min-width: 1201px)`]
]
```
if the matchMedia isn't supported which will be the case for `ssr` and react native then the function will return `[]`

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  const getMatchedDeviceType = useCallback((matchedBreakpoint: Breakpoint): DeviceType => {
+    let matchedDeviceType: DeviceType = 'mobile';
+    if (typeof navigator !== 'undefined' && navigator.product === 'ReactNative') {
+      // react-native
+      matchedDeviceType = 'mobile';
+    } else if (typeof document !== 'undefined') {
```
kamleshchandnani, line 45: this checks will succeed for for csr(browser)

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+    } else if (typeof process !== 'undefined') {
+      // node
+      //@TODO: Check for useragent for node
+      matchedDeviceType = 'desktop';
+    }
+    return matchedDeviceType;
+  }, []);
+
+  const getMatchedBreakpoint = useCallback(
```
kamleshchandnani, line 61: this function gets called twice:
1. when the initial state is getting initialized
2. when the media query `change` event gets triggered when there's a screen size change

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,144 @@
...
+    (event?: MediaQueryListEvent): Breakpoint => {
+      const matchedBreakpoint =
+        breakpointsTokenAndQueryCollection.find(([_, __, mediaQuery]) => {
+          if (event?.media === mediaQuery) {
+            return true;
+          }
```
kamleshchandnani, line 67: this will fail when we the state is initialised for the first time when the hook is getting initialised and hence the event object will be empty so we'll fallback to browser's window object

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,144 @@
...
+          if (window.matchMedia(mediaQuery).matches) {
```
kamleshchandnani, line 68: evaluates to true when the even undefined and the media query matches the current browser sizes

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,144 @@
...
+    return {
+      matchedBreakpoint,
+      matchedDeviceType,
+    };
+  });
+
+  useEffect(() => {
+    if (!supportsMatchMedia) {
+      return undefined;
+    }
```
kamleshchandnani, line 91: we basically return early here when the supports matchmedia doesn't exist. this will be true for ssr and react-native apps

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,144 @@
...
+        const matchedBreakpoint = getMatchedBreakpoint(event);
+        const matchedDeviceType = getMatchedDeviceType(matchedBreakpoint);
+        if (
+          prevBreakpointAndDevice.matchedBreakpoint !== matchedBreakpoint ||
+          prevBreakpointAndDevice.matchedDeviceType !== matchedDeviceType
+        ) {
+          return { matchedBreakpoint, matchedDeviceType };
+        }
+        return prevBreakpointAndDevice;
+      });
```
kamleshchandnani, line 104: this is the handler that gets triggered when the media query's `change` event is triggered. so we check if the `matchedBreakpoint` or `matchedDeviceType` has changed and update the state accordingly

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,144 @@
...
+    };
+
+    const mediaQueryInstances = breakpointsTokenAndQueryCollection.map(([_, __, mediaQuery]) =>
+      window.matchMedia(mediaQuery),
+    );
```
kamleshchandnani, line 109: the mediaquery event listener is available on mediaQuery instances and not `window` so we need to iterate over all the breakpoints we have and register instance for each and store them as `mediaQueryInstances` so we can later unregister all of them.

More details about [MediaQuery vs window  `addEventListeners`](https://webdevetc.com/blog/matchmedia-events-for-window-resizes/#windowmatchmedia-to-the-rescue)

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    mediaQueryInstances.forEach((mediaQueryInstance) => {
```
kamleshchandnani, line 111: now if you see we basically iterate on all the instances we got above in `mediaQueryInstances` and attach a `change` event and a handler to it.

packages/blade/src/utils/useBreakpoint/useBreakpoint.web.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    const { result } = renderHook(() => useBreakpoint({ breakpoints }));
+
+    expect(result.current.matchedBreakpoint).toBe('xs');
+    expect(result.current.matchedDeviceType).toBe('mobile');
+  });
+
+  it('should match a specific media query and return the device type accordingly', () => {
+    const addEventListenerMock = jest
+      .fn()
+      .mockImplementation((_, callback: (arg: { media: string }) => void) => {
```
kamleshchandnani, line 21: the reason why we do this is because we don't have a browser window here so all we do is mock the event listener and call it with a mediaQuery that we might receive in actual browser i.e `media: 'screen and (min-width: 769px) and (max-width: 1024px)'` and then do our assertions. 

Then after our assertions, we again call this callback explicitly [here](https://github.com/razorpay/blade/pull/424/files#diff-b9d80003bb947c85eab50a87209022f7751656022bef098c6655d561d488a5edR36) to simulate the media query change event and this time we pass different media query i.e `media: 'screen and (max-width: 320px)'` and then make assertions that we get different device as intended

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,142 @@
+import { useEffect, useState, useCallback, useMemo } from 'react';
+import type { Breakpoints } from '../../tokens/global';
+
+type DeviceType = 'desktop' | 'mobile';
```
anuraghazra, line 4: Let's use an object here. 

```ts
const device = {
  desktop: "desktop",
  mobile: "mobile",
} as const; 
type DeviceType = keyof typeof device;
```
And remove all hardcoded string "mobile", "desktop" magic values.

packages/blade/src/utils/useBreakpoint/useBreakpoint.web.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      addEventListener: addEventListenerMock,
+    });
+
+    const { result } = renderHook(() => useBreakpoint({ breakpoints }));
+
+    expect(result.current.matchedBreakpoint).toBe('l');
+    expect(result.current.matchedDeviceType).toBe('desktop');
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    const addEventListenerMockCallback = addEventListenerMock.mock.calls[0][1] as (arg: {
```
kamleshchandnani, line 36: @anuraghazra if you can help here 😭

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
kamleshchandnani: what will we gain by doing that?

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,142 @@
...
+                mediaQuery = `screen and (min-width: ${breakpointSize}px)`;
+              } else if (breakpointsArray[index - 1]) {
+                mediaQuery = `screen and (min-width: ${
+                  breakpointsArray[index - 1][1] + 1
+                }px) and (max-width: ${breakpointSize}px)`;
+              } else {
+                mediaQuery = `screen and (max-width: ${breakpointSize}px)`;
+              }
+
+              return [breakpointTokenName, breakpointSize, mediaQuery];
```
anuraghazra, line 33: Will probably be a bit better if we make the structure of this into an object? 

`{tokenName, size, query}`

Arrays are evil, need to keep track of indices.

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
anuraghazra: So that we don't have scattered string values for "mobile" & "desktop". 

Screenshot 2022-03-14 at 11 49 04 AM

packages/blade/src/utils/useBreakpoint/useBreakpoint.web.test.ts
anuraghazra: @kamleshchandnani this will fix it. 

```ts
    const addEventListenerMock = jest
      .fn<void, ((arg: { media: string; }) => void)[]>()
      .mockImplementation((_, callback) => {
        callback({ media: 'screen and (min-width: 769px) and (max-width: 1024px)' });
      });
```

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
anuraghazra: We probably don't need this second `forEach`?? We can attach the event listeners while mapping over the mediaQueries

```ts
    const mediaQueryInstances = breakpointsTokenAndQueryCollection.map(([_, __, mediaQuery]) =>{
      const mediaQueryInstance = window.matchMedia(mediaQuery);
    
      if (mediaQueryInstance.addEventListener) {
        mediaQueryInstance.addEventListener('change', handleMediaQueryChange);
      } else {
        // In older browsers MediaQueryList do not yet inherit from EventTarget, So using addListener as fallback - https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener
        mediaQueryInstance.addListener(handleMediaQueryChange);
      }

      return mediaQueryInstance;
    });
```

packages/blade/src/utils/useBreakpoint/useBreakpoint.web.test.ts
kamleshchandnani: the moment I do that ts-eslint throws warning `void is only valid as a return type or generic type variable` 😭

packages/blade/jest-setup.native.js
```diff
@@ -0,0 +1,3 @@
+Object.defineProperty(global.navigator, 'product', {
+  value: 'ReactNative',
+});
```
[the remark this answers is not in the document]
dhruvdutt: Add this in the comment as well briefly? :D

packages/blade/src/components/BladeProvider/BladeProvider.tsx
```diff
@@ -29,9 +34,12 @@ const BladeProvider = ({
...
 
   const colorMode = `on${toTitleCase(colorScheme)}` as ColorSchemeModes;
-  const deviceType = `on${toTitleCase(getDeviceType())}` as TypographyPlatforms;
+  const onDeviceType = `on${toTitleCase(matchedDeviceType)}` as TypographyPlatforms;
```
dhruvdutt, line 42: Let's follow the same convention for both variables. `onColorMode`.

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
dhruvdutt: Write the last line in code comments.

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,142 @@
...
+      matchedDeviceType = 'desktop';
+    }
+    return matchedDeviceType;
+  }, []);
+
+  const getMatchedBreakpoint = useCallback(
+    (event?: MediaQueryListEvent): Breakpoint => {
+      const matchedBreakpoint =
+        breakpointsTokenAndQueryCollection.find(([_, __, mediaQuery]) => {
+          if (event?.media === mediaQuery) {
```
dhruvdutt, line 65: Which case is this? Didn't get. Also, can you add comments in the code.

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,142 @@
...
+            return true;
+          }
+          if (window.matchMedia(mediaQuery).matches) {
+            return true;
+          }
+          return false;
+        })?.[0] ?? undefined;
```
dhruvdutt, line 72: This `[0]` thing can also be simplified if we store the collection as array of objects.

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,142 @@
...
+          prevBreakpointAndDevice.matchedBreakpoint !== matchedBreakpoint ||
+          prevBreakpointAndDevice.matchedDeviceType !== matchedDeviceType
+        ) {
+          return { matchedBreakpoint, matchedDeviceType };
+        }
+        return prevBreakpointAndDevice;
+      });
+    };
+
+    const mediaQueryInstances = breakpointsTokenAndQueryCollection.map(([_, __, mediaQuery]) =>
```
dhruvdutt, line 107: This `_, __` thing can also be simplified if we store the collection as array of objects.

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
dhruvdutt: +1

Can even do `.map.forEach`.

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,144 @@
...
+      if (matchedBreakpoint && ['xs', 's', 'm'].includes(matchedBreakpoint)) {
+        // tablet is also categorised as mobile
+        matchedDeviceType = deviceType.mobile;
+      } else {
+        matchedDeviceType = deviceType.desktop;
+      }
+    } else if (typeof process !== 'undefined') {
+      // node
+      //@TODO: Check for useragent for node
+      matchedDeviceType = deviceType.desktop;
```
dhruvdutt, line 61: For ssr here, change this to mobile by default

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
anuraghazra: `.map.forEach` would be the same as before, 
What I'm saying is to basically do that inside map itself. That way we won't need to loop twice. once in `map` and then on `forEach`

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
kamleshchandnani: it's self-explanatory no. what shall I add? yaha pe I've added so that it's easier for PR reviewer

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
kamleshchandnani: because it's setState so if I always return the new object it'll trigger unnecessary re-renders hence I return the old state if nothing changes. basically bail-out of re-render

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
kamleshchandnani: hmm ok for web I'm keeping it to desktop for now since majority of our apps are desktop

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,142 @@
...
+      if (mediaQueryInstance.addEventListener) {
+        mediaQueryInstance.addEventListener('change', handleMediaQueryChange);
+      } else {
+        // In older browsers MediaQueryList do not yet inherit from EventTarget, So using addListener as fallback - https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener
+        mediaQueryInstance.addListener(handleMediaQueryChange);
+      }
+    });
+
+    return (): void => {
+      // window.removeEventListener('resize', handleMediaQueryChange);
```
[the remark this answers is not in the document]
kamleshchandnani: nupe i had removed it long back but forgot to push 🤦

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -66,15 +68,17 @@ const useBreakpoint = ({ breakpoints }: { breakpoints: Breakpoints }): Breakpoin
   const getMatchedBreakpoint = useCallback(
     (event?: MediaQueryListEvent): Breakpoint => {
       const matchedBreakpoint =
-        breakpointsTokenAndQueryCollection.find(([_, __, mediaQuery]) => {
-          if (event?.media === mediaQuery) {
+        breakpointsTokenAndQueryCollection.find((breakpointTokenAndQuery) => {
```
dhruvdutt, line 71:
```suggestion
        breakpointsTokenAndQueryCollection.find(({ mediaQuery } ) => {
```

Nitpick

dhruvdutt (approved):

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -0,0 +1,149 @@
...
+              mediaQuery = `screen and (min-width: ${
+                breakpointsArray[index - 1][1] + 1
+              }px) and (max-width: ${screenSize}px)`;
+            } else {
+              mediaQuery = `screen and (max-width: ${screenSize}px)`;
+            }
+
+            return { token, screenSize, mediaQuery };
+          })
+        : []) as {
```
chaitanyadeorukhkar, line 39: We have to do this because `Object.entries(breakpoints)` is inferring the type as `[string, number]`. Can it not retain ` "xs" | "s" | "m" | "l" | "xl" | "max"` instead of a generic string here?

// @anuraghazra any ideas?

chaitanyadeorukhkar (approved):
Lgtm

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
kamleshchandnani: need to add a lot of typings here since I'm returning either empty array or the array of objects with this structure so not worth the efforts

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
anuraghazra: Yes we can do that have to write a stricter objectEntries() function to infer the types. 

Same as we do for objectKeys.
```ts
const objectKeys = <T extends Record<string, unknown>>(obj: T) =>
  (Object.keys(obj) as unknown) as (keyof T)[]
```

anuraghazra: Okay yeah this should work. 

```ts
const objectEntries = <T extends Record<string, any>>(obj: T) =>
  (Object.entries(obj) as unknown) as [keyof T, T[keyof T]][];
```

PR #433, opened 2022-03-26
feat(rfc): add motion RFC

chaitanyadeorukhkar:
### [View Formatted RFC](https://github.com/razorpay/blade/blob/rfc/2022-03-22-motion-rfc/rfcs/2022-03-22-motion-rfc.md)


rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+> You can use `transition-delay` CSS property in the exit animation of the sub-menu to achieve this.
+
+<img alt="rotate" src="./images/motion/dropdown.gif" width="300px">
+
+#### Tokens - Delay
+We will be storing these tokens in `blade/src/tokens/global/motion.ts` as a `string` of milliseconds.
+
+> *Note: The naming for these tokens is not finalized yet. We will be updating this in the future.*
+
+Here is the **list of tokens** that we will store in Blade for `Delay`:
```
saurav12, line 127: How did we decide on these values or any rationale behind them?
PS: Can we link Dushyant's RFC/document here if any?

saurav12:
Very well written. Looks good to me 🙌
Just one point, it would be great if we can add references to the research we did in analysing the token values.

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+
+# Adoption strategy
+- Consumers can start using motion tokens as they would with other DS tokens.
+- Consumers would need to be aware of the drawback mentioned above at [Library specific tokens](#library-specific-tokens) before they start using them
+- Blade's developers can use this RFC as a guideline to add motion to their components
+
+# How do we educate people?
+- Have a detailed guideline on our documentation site
+
+# Open Questions
```
sidag95, line 602: Should we also include a section on how to handle user preference of `prefers-reduced-motion` to stop these motion animations?

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+
+# Drawbacks/Constraints
+### Library specific tokens
+Consumers would need to be aware that certain tokens are library specific and would not work with other libraries
+- For web, we expose `duration` as a `string` of *milliseconds* which works well with CSS's `transition-duration` property but will not work with Framer motion since it requires a `number` in *seconds*
+- For web, we expose `easing` as a `string` of `cubic-bezier(..)` which works well with CSS's `transition-timing-function` property but will not work with Framer motion since it requires either,
+  - The name of an existing easing function.
+  - An array of four numbers to define a cubic bezier curve.
+  - An easing function, that accepts and returns a value 0-1.
+- For native, we expose `easing` as a function of `Easing.bezier(..)` imported from `react-native-reanimated` which won't work when used with React Native's `Animated` API.
```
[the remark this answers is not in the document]
sidag95: Again, as this is not in scope at the moment, maybe this can also be explored in the Real-time motion RFC.

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+## Scope of Motion
+As of this RFC, we will be defining rules of motion only for **Non-Realtime** interactions. 
+
+Few examples of non-realtime motion are:
+1. Changing the size of a button on mouse hover
+2. Changing the opacity of a card on press
+3. Fading out a modal when the user closes it
+
+
+## Actions of Motion
```
nidhi-tandon, line 83: Actions of motion can be a combination of 2 like scale and fill.

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+### Library specific tokens
+Consumers would need to be aware that certain tokens are library specific and would not work with other libraries
+- For web, we expose `duration` as a `string` of *milliseconds* which works well with CSS's `transition-duration` property but will not work with Framer motion since it requires a `number` in *seconds*
+- For web, we expose `easing` as a `string` of `cubic-bezier(..)` which works well with CSS's `transition-timing-function` property but will not work with Framer motion since it requires either,
+  - The name of an existing easing function.
+  - An array of four numbers to define a cubic bezier curve.
+  - An easing function, that accepts and returns a value 0-1.
+- For native, we expose `easing` as a function of `Easing.bezier(..)` imported from `react-native-reanimated` which won't work when used with React Native's `Animated` API.
+
+### React Native `v0.62`
```
kamleshchandnani, line 495: wouldn't this be a biggest problem? since we now have React 18 out which also exposes certain things we might want to use and if we are limited to this version then that might also cause issue?

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+    100% {
+      transform: translateY(100px);
+    }
+}
+```
+
+- In the 2-step keyframe, easing is applied for the animation taking place from 0% to 100%.
+- In the multi-step keyframe, easing is applied for each step of the animation. That is, if we use `ease-in-out` easing, it would be applied to the animation happening from `0%` to `50%` and then again from `50%` to `100%`.
+- This is not supported in React Native Reanimated _directly_ since we apply easing to the `timing` function.
+
```
kamleshchandnani, line 530: can you attach a side by side example gif to make it visually understandable?

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+    shakeAnimation.value = withTiming(100, { 
+      duration: motionToken.duration.duration3, 
+      easing: motionToken.easing.standard.effective, 
+    });
+  ```
+
+- We apply easing to the timing function as `shakeAnimation` value goes from 0 to 100. In this case if we apply an `ease-in-out` easing, it would be applied to the entire animation happening from `0%` to `100%` regardless of the fact that our interpolate's input & output range is more than 2 steps (array of more than 2 elements).
+- If we _really_ want to replicate the web's behaviour on react native, we would need to apply easing on the `interpolate` function. 
+- As of `v2.3.0` of `react-native-reanimated`, there is no support to add an easing to `interpolate`. 
+- To be able to achieve this, we would need to implement the animation using React native's `Animated` API since it supports applying easing to interpolation like this:
```
kamleshchandnani, line 541: so finally what are we concluding? we'll use Reanimated or native animated? above you have mentioned that we'll use reanimated but that has constraints of being used on `version: >0.62` and it also doesn't support the similar behavior to the animations on web.

kamleshchandnani (changes requested):

rfcs/2022-03-22-motion-rfc.md
chaitanyadeorukhkar: I've added the link to his RFC in references

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+- In the above example, we are using `useAnimatedStyle` to create a worklet that will be executed on the UI thread and has a **shared value** `translateXOffset` that is shared by JS as well as native realms.
+- React Native reanimated is able to achieve this using React Native's `TurboModules` feature.
+- Since **we want to leverage these benefits** of **React Native Reanimated**, **we will be using it** for our native **animations on Blade**.
+
+
+# Drawbacks/Constraints
+### Library specific tokens
+Consumers would need to be aware that certain tokens are library specific and would not work with other libraries
+- For web, we expose `duration` as a `string` of *milliseconds* which works well with CSS's `transition-duration` property but will not work with Framer motion since it requires a `number` in *seconds*
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: We might want to use framer or react spring at some point in the future. We'd have to relook at the tokens then. For those use cases, we could have helpers in place and we can be okay but for use cases where we'd wanted to use `Easing.bezier` from react native's `Animated` API instead of Easing.bezier` of react native reanimated, we might have to make a good amount of refactor since we're planning to expose the token as a function of `Easing.bezier(...)` where `Easing` is imported from a specific library.

rfcs/2022-03-22-motion-rfc.md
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah, we'd need a solution that allows us to extend usage to multiple libraries.

rfcs/2022-03-22-motion-rfc.md
chaitanyadeorukhkar: Yeah, it can be combinations as well. I'll add that here.

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+```
+
+> For React Native, we would need to store delay as `number` instead of `string`. Eg: `delay1: 70`
+
+### Duration
+Duration is the time taken to complete any transition, interaction and animation. Each animation can run for a duration of time depending on factors like size of the object, distance of travel, direction of travel, etc. 
+
+*Example: The card takes 300ms to open and 250ms to close*
+
+<img alt="rotate" src="./images/motion/card-duration.gif" width="300px">
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Achieving shared animation is not the goal of this RFC. That's a separate concern on its own. I put this example here to demonstrate how this real-world example that we've commonly seen uses duration for entry & exit states.

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,612 @@
...
+
+Here is the **list of tokens** that we will store in Blade for `Delay`:
+```js
+delay: {
+  delay1: '70ms',
+  delay2: '120ms',
+  delay3: '180ms',
+  delay4: '3000ms',
+  delay5: '5000ms',
+}
```
anuraghazra, line 136: Is this finalized? 
How will we access this? delay.delay1? seems redundant. 🤔 maybe we should do `delay[100]`

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,612 @@
...
+
+#### Tokens - Easing
+> Note: *The naming for these tokens is not finalized yet. We will be updating this in the future.*
+
+- We will be storing **Easing tokens** in `blade/src/tokens/global/motion.ts` as a `cubic-bezier()` function.
+
+- After thorough research & experimentation, our design team has created 3 broad **intents** of easing: 
+  - `standard`
+  -  `entrance`
+  -   `exit`
```
anuraghazra, line 228: **Naming opinion:**

To me standard is a bit confusing, the intent isn't obvious when should I use. 

A better naming would be "idle" where it hints to the consumer that this token is used for objects which are idle and not entering or exiting. It conveys they intent much better IMO.

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,612 @@
...
+  },
+  exit: {
+    effective: 'cubic-bezier(0.17, 0, 1, 1)',
+    revealing: 'cubic-bezier(0.5, 0, 1, 1)',
+    attentive: 'cubic-bezier(0.7, 0, 0.5, 1)',
+  },
+}
+```
+
+For **React Native**, we need to store easing as `Easing.bezier` where `Easing` can be imported from react-native's `Animated` or `react-native-reanimated` depending on the library you are using.
```
anuraghazra, line 277: We should update this section with the discussion we did today morning on call about using makeBezier function.

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,612 @@
...
+  entrance: {
+    effective: 'cubic-bezier(0, 0, 0.2, 1)',
+    revealing: 'cubic-bezier(0, 0, 0, 1)',
+    attentive: 'cubic-bezier(0.5, 0, 0.3, 1.5)',
+  },
+  exit: {
+    effective: 'cubic-bezier(0.17, 0, 1, 1)',
+    revealing: 'cubic-bezier(0.5, 0, 1, 1)',
+    attentive: 'cubic-bezier(0.7, 0, 0.5, 1)',
+  },
```
anuraghazra, line 273: **Implementation thoughts:**

How do we provide a way for component to unmount & remount while waiting for enter & exit animations? FramerMotion and other third party libs automatically handle these stuff.

rfcs/2022-03-22-motion-rfc.md
chaitanyadeorukhkar: Naming isn't finalised

rfcs/2022-03-22-motion-rfc.md
chaitanyadeorukhkar: `standard` is just the standard easing we'll be using for everything. `idle` might not be the case every time. We could go with `default` but that seems more from the sense that if you don't apply anything, this would be the default anyway. `standard` seems like an intent you'd have for all easings that aren't entering or exiting

rfcs/2022-03-22-motion-rfc.md
chaitanyadeorukhkar: That's okay. That's implementation details. Even after adding a helper on top of it, the end state would still be the same. If I add that here, it would just complicate it for readers.

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+
+const AnimatedBlock = styled.div` /* Create a styled div */
+  :hover {
+    animation: 
+      shake /* Set a custom keyframe animation */
+      ${motionToken.duration.duration3} /* Set the duration of the animation using our motion tokens */
+      ${motionToken.easing.standard.effective}; /* Set the easing of the animation using our motion tokens */
+  }
+
+  @keyframes shake { /* Define the custom keyframe animation */
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: I don't think we've reached that point yet where we publish some common animations. That's a big bet to make right now. Cuz if we start publishing shake function for web, we'll have to do it for native too. Let
's figure out if we want to do this if we start seeing a lot of repeating animations. Even in that case, we'd internally abstract it for blade but not expose it to consumers

rfcs/2022-03-22-motion-rfc.md
```diff
@@ -0,0 +1,611 @@
...
+      <AnimatedBlock />
+    <div>
+)
+```
+
+- Different 3rd party libraries have **different approaches** for creating the **same animation**.
+- In our comparison between Framer Motion & native CSS Animations, we found that for **most non-realtime** animation use-cases where we want to create **simple transitions** & **keyframe** animations on **events** such as mouse **hover**, **CSS animations** are **sufficient** and easy to implement.
+- Implementation becomes a bit **complicated** when we want to **trigger** certain **animations** from **Javascript**. 
+- With **Framer Motion**, we can use it's **animation control** created with `useAnimation` to **trigger** the animation. Eg) `move.start()`
+- For **CSS**, we would either need to **manipulate** component **state** or **manipulate** **CSS** **classes** list with `document.getElementById().classList.add()` / `document.getElementById().classList.remove()` that can dynamically add and remove an animated class from an element to **trigger** an animation.
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: It's just a different approach. Libraries like framer give us an easier way to do this but doing it this way isn't a drawback. Just highlighting here that it is relatively more complicated than framer. 

But open to suggestions here, I can add to drawbacks if you think it really is a drawback

rfcs/2022-03-22-motion-rfc.md
chaitanyadeorukhkar: Not limited to this version. Any version greater than or equal to 0.62. I'll rephrase it to make it clear

rfcs/2022-03-22-motion-rfc.md
chaitanyadeorukhkar: Not too sure about this but I'm guessing you unmount _after_ the motion ends and remount _before_ the motion starts

rfcs/2022-03-22-motion-rfc.md
chaitanyadeorukhkar: If we for some reason want to replicate the keyframes behaviour on react native, we'll use the Animated API. For all other use cases, react native reanimated works well for us. 

The constraint of the `0.62` version is not a big issue. All our internal apps above `0.62` and react native's latest is `0.68`

rfcs/2022-03-22-motion-rfc.md
chaitanyadeorukhkar: This is an edge case where someone would want to replicate the keyframes behavior on native. I've added a link to [this blog](https://www.joshwcomeau.com/animation/keyframe-animations/#multi-step-animations) above that talks more about multi-step animations. Going into examples here would increase the complexity of the RFC.

The intention to keep this documented here is so that _if_ someone wants to do this specific thing in the future, they can refer to this RFC and understand that we were already aware of this edge case and have a workaround described here as well.

rfcs/2022-03-22-motion-rfc.md
chaitanyadeorukhkar: Good question. Not sure where this fits in right now. 

1. Would consumers take care of implementing alternative "less animated" versions using different tokens that we'll expose?
2. Would we change the values of the tokens depending on `prefers-reduced-motion`? 
3. Would we need to change motion for the components we expose from blade if we detect `prefers-reduced-motion`? 
4. Should we figure these out when we explore with a11y?

Let me discuss this with the rest of the team and get back to you.

// @kamleshchandnani @saurav12

kamleshchandnani (approved):

PR #437, opened 2022-04-08
feat(blade): add Text decisions

kamleshchandnani:


packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
...
+
+- `Text`
+
+  This component is the one that we'll use to render normal text on the screen(a.k.a as paragraphs or body text)
+
+  Text will internally use `BaseText` to render. Natively we'll use html `<p></p>` tag for web and native `<Text></Text>` react-native to render the headings.
+
+  ```jsx
+  <Text
+    variant=" medium | large | caption"
```
anuraghazra, line 67: no small variant for <Text />?

packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
...
+    weight="regular | bold"
+    type="normal | subtle | subdued | muted | placeholder"
+    shouldTruncate="true | false"
```
anuraghazra, line 70: We might need to be consistent of these naming conventions where we prefix booleans with "should" or "is". 

Feel like isTruncated is a better here. 🤔

packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
...
+  This component is just meant to be used for internal purposes for Blade
+  Here's how the API will look like
+
+  ```jsx
+  <BaseText
+    color="surface.text | feedback.text | button.text"
+    fontFamily="fontfamily.name"
+    fontSize="fontSize.name"
+    fontWeight="fontWeight.name"
+    lineWeight="lineWeight"
```
chaitanyadeorukhkar, line 25:
```suggestion
    lineHeight="lineHeight"
```

packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
...
+
+Here's how the hierarchy will looks like
+![Component Breakdown](./component-breakdown.png)
+
+The APIs for all the above component will look something like below
+
+- `BaseText`
+
+  This component is just meant to be used for internal purposes for Blade
+  Here's how the API will look like
```
chaitanyadeorukhkar, line 17:
```suggestion
 
  Here's how the API will look like:
```

packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
+The text component is categorised into multiple parts
```
anuraghazra, line 1: Could restructure the folder a bit better. 

/spec/decisions.md
/specs/accessibility.md - in Future.

packages/blade/src/components/Text/Decisions/Decisions.md
chaitanyadeorukhkar: Should we just call it `truncate`? 

`truncate` sounds more like a command we're giving to the component and makes `true` seem implicit. Since the default is `false`, the presence or absence of `truncate` will act as a decision-maker. `shouldTruncate` sounds more like a question that needs an answer - should truncate? yes.  Also, `truncate` is much smaller to write 😛 

```jsx
<Text
  truncate
/>
```

packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
...
+
+  Text will internally use `BaseText` to render. Natively we'll use html `<p></p>` tag for web and native `<Text></Text>` react-native to render the headings.
+
+  ```jsx
+  <Text
+    variant=" medium | large | caption"
+    weight="regular | bold"
+    type="normal | subtle | subdued | muted | placeholder"
+    shouldTruncate="true | false"
+    numberOfLines="" // not yet finalised but might need for truncating on react-native
```
chaitanyadeorukhkar, line 71: We'll definitely need this for react-native. Why isn't it finalised?

packages/blade/src/components/Text/Decisions/Decisions.md
anuraghazra: truncate is fine too, although i will lean a bit towards prefixing with `is`. 

reason being: 

- prefixing with `is` hints user that its a boolean without knowing the types.
- `truncate` can be a bit confusing in non-TS projects like: "What does it take? truncate={5} will this truncate 5lines? or what?"
- Seeing most third party design systems also follow this naming convention

packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
...
+  >
+    Some Text
+  </Text>
+  ```
+
+- `Code`
+
+  This component will be used to highlight code content in the paragraph. This not the actual code block component. the scope is to just highlight inline content which can refer code.
```
saurabhdaware, line 79: > This not the actual code block component.

Should we call this something else in that case? Maybe `<Tag>`? From the name I assumed that it is a code block component 🤔

packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
...
+  ```jsx
+  <BaseText
+    color="surface.text | feedback.text | button.text"
+    fontFamily="fontfamily.name"
+    fontSize="fontSize.name"
+    fontWeight="fontWeight.name"
+    lineWeight="lineWeight"
+    letterSpacing="letterSpacing"
+    formatting="bold/italic/strike"
+    as="h1-h6/p/span"
```
saurabhdaware, line 28: Will all exported components support `as`?

saurabhdaware (commented):
Yay! Thanks for opening it up for reviews 🎉

packages/blade/src/components/Text/Decisions/Decisions.md
saurabhdaware: Or can we make decision based on the value? 

`<Code>SENTRY_AUTH_TOKEN</Code>` is rendered as inlined code tag

```js
<Code>
const a = 3;
const b = 9;
</Code>
```

will be rendered as codeblock

packages/blade/src/components/Text/Decisions/Decisions.md
chaitanyadeorukhkar: Makes sense. `is` as a prefix makes more sense to me than `should`. For instance, a button can have `disabled` or `isDisabled` but never `shouldDisable`

packages/blade/src/components/Text/Decisions/Decisions.md
chaitanyadeorukhkar: Had the same confusion. Tag would be a bit more confusing imo. We'll probably have `CodeBlock` in the future as a component and that might resolve the confusion but I don't mind renaming `Code` component to `InlineCode` to be clearer

packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
...
+  />
```
PujanShah22, line 29: Can we also include `textAlign`? There are many use-cases of it.

packages/blade/src/components/Text/Decisions/Decisions.md
PujanShah22, line 67: How font size will be defined? Based on variants or type?
In blade old we had `size` prop which was supporting different font sizes for design. Just curious to know how it will be handled here?

kamleshchandnani: not right now.

packages/blade/src/components/Text/Decisions/Decisions.md
kamleshchandnani: there's a difference in tense when you refer to `is` and `should`. `is` basically says it will instantly be true/false or on action for eg: `isDisabled` for a button that might be toggled based on the form being submitted or based on a user role. when you talk about `should` it basically says the behavior shall be toggled for something that might happen in the future eg: `shouldComponentUpdate` lifecycle method. Similarly, over here we are saying `shouldTruncate` which dictates that the text shall be truncated after some lines so if I say `isTruncate` that conveys that it'll immediately be truncated but ideally it's not.

packages/blade/src/components/Text/Decisions/Decisions.md
kamleshchandnani: decisions will have all sorts of decisions, be it API, be it Accessibility, or anything else. otherwise, it's challenging to discover these files. since it's mostly meta-information i would prefer to keep it under one file. if this had been a component or any other file the recommendation would have been valid.

packages/blade/src/components/Text/Decisions/Decisions.md
kamleshchandnani: yes, i just need to identify it's counterpart in web

packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
...
+
+  Title will internally use `BaseText` to render. We'll use native html heading tags to render title and native text element for react-native to render the Title.
+
+  ```jsx
+  <Title variant="small | medium | large" type="normal | subtle | subdued | muted | placeholder">
+    Some Title
+  </Title>
+  ```
+
+- `Heading`
```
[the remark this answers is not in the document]
kamleshchandnani: we basically want to free people from deciding H1-H6 as we are using a design language to streamline this vocabulary. so H1-H6 is understandable for web folks but not by designers and react native developers. So since we'll also be publishing for react-native there's no such thing as H1-H6 for react native it's all plain text. Regarding, your question about how we'll identify which H1 to render on the web we have variants if you see them in the API. We'll have a variant to equivalent Heading tag internally and take that decision internally while rendering instead of consumers taking that call. So as a consumer(dev/designer) I'll just say I need small heading, medium heading or large heading.

packages/blade/src/components/Text/Decisions/Decisions.md
```diff
@@ -0,0 +1,87 @@
...
+    fontSize="fontSize.name"
+    fontWeight="fontWeight.name"
+    lineWeight="lineWeight"
+    letterSpacing="letterSpacing"
+    formatting="bold/italic/strike"
+    as="h1-h6/p/span"
+  />
+  ```
+
+- `Title`
```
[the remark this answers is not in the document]
kamleshchandnani: >Can we not have this as a variant of Heading?

Actually no because they have different meaning. Titles are mostly the texts that are free form on the landing pages, the big texts you saw vs Heading which always are like section breakpoints and will be followed by paragraph or subheadings.


>Also can this be confused with HTML's <title>? 

hmm, interesting. any other suggestions you might suggest? you can look the hierarchy diagram and understand the semantic behind each component and propose any suggestion for Title?

packages/blade/src/components/Text/Decisions/Decisions.md
kamleshchandnani: Might keep it as `Code` and enhance it when we start supporting multi-line code block and at that time we can just make this component accept a prop called `inline=true`. Keeping it close to `<code>` native HTML element

packages/blade/src/components/Text/Decisions/Decisions.md
kamleshchandnani: nupe only base text which again will only be used internally

packages/blade/src/components/Text/Decisions/Decisions.md
kamleshchandnani: can you share some example?

packages/blade/src/components/Text/Decisions/Decisions.md
kamleshchandnani: it'll be decided based on combination of both variant and type.

packages/blade/src/components/Text/Decisions/Decisions.md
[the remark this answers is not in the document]
saurabhdaware: aha that sounds better. We can see how it acts out in actual applications and make changes later if accessibility is getting imacted.

packages/blade/src/components/Text/Decisions/Decisions.md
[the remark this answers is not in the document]
saurabhdaware: > Actually no because they have different meaning. Titles are mostly the texts that are free form on the landing pages, the big texts you saw vs Heading which always are like section breakpoints and will be followed by paragraph or subheadings.

It can still be variant of Heading right? For example, it still is a 'Heading' of the page. Having paragraph isn't a requirement right? Even in terms of semantics, it will be `<h1 />`.

I think Vercel has `<Text title>` component for their page headings, Chakra uses `<Heading />` itself.


If we still want separate components, I can think of these names - `<H1 />`, `<ScreenTitle />`, `<PageTitle />`, `<MainHeading />`.

packages/blade/src/components/Text/Decisions/Decisions.md
saurabhdaware: For example this section in razorpay.com-

image

Here "Explore Razorpay Business Banking" is h2

Both, "Manage your company’s finances with Razorpay X Business Banking" and those labels like "Current Account", "Capital & Credits" they are all h3 even though they are styled completely differently.

Are we going to limit such usecases from design itself?

packages/blade/src/components/Text/Decisions/Decisions.md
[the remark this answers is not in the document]
kamleshchandnani: Re: variant of heading - it can be a variant of heading but as I mentioned we would want the internal system to take care of that instead of consumers making these choices that what HTML heading tag to use. so once we have semantic named components like Title, Heading, Text, Code which dictates the intent and what to render internally is the decision left up to the underlying system

Re: Title - I think we shall start thinking from interfaces in general and not map it to native HTML as we have audiences such as designers and devs from different backgrounds like React native. so Sticking with title for now.

packages/blade/src/components/Text/Decisions/Decisions.md
kamleshchandnani: yes that won't be a problem right since we have `Heading` tag to deal with these kind of use cases

PR #438, opened 2022-04-11
feat(blade): add motion tokens

chaitanyadeorukhkar:
Adds motion tokens for
- Duration
- Delay
- Easing

Find a detailed RFC for these tokens here: [View Formatted RFC](https://github.com/razorpay/blade/blob/master/rfcs/2022-03-22-motion-rfc.md)


packages/blade/src/utils/makeBezier/makeBezier.native.ts
```diff
@@ -0,0 +1,7 @@
+import { Easing, EasingFunctionFactory } from 'react-native-reanimated';
+
+const makeBezier = (x1: number, y1: number, x2: number, y2: number): EasingFunctionFactory => {
+  return Easing.bezier(x1, y1, x2, y2);
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Bascially this
```ts
type EasingFunction = (value: number) => number;
type EasingFunctionFactory = { factory: () => EasingFunction };

```

packages/blade/src/tokens/global/motion.ts
```diff
@@ -0,0 +1,132 @@
+import makeBezier from '../../utils/makeBezier';
+
+export type Motion = Readonly<{
+  delay: {
+    /** `70` milliseconds */
```
anuraghazra, line 5: Man!! I really wish typescript had a way to derive these values from the code. If we change the value this wont be updated. :(

chaitanyadeorukhkar: Yes :(

packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.native.ts
```diff
@@ -0,0 +1,1126 @@
...
+      quick3: 200,
+      moderate1: 250,
+      moderate2: 300,
+      gentle1: 400,
+      gentle2: 600,
+      gentle3: 900,
+    },
+    easing: {
+      standard: {
+        //@ts-expect-error Need to figure out a way to have a different Theme type for native
```
chaitanyadeorukhkar, line 1045: @kamleshchandnani @anuraghazra having to do this is a big bummer. Since our `Theme` type has this as a `string` which is valid for web but not for native. I could add a `string | EasingFactoryFunction` but that would lead to confusion for consumers. 

The best we could to do is add `string | EasingFactoryFunction` and mention it in the description that it would return `string` on web & `EasingFactoryFunction` on native?

anuraghazra: Yeah agree on the consumer confusion point, probably we could separate out the Theme types but need to think it through. 

> The best we could to do is add string | EasingFactoryFunction and mention it in the description that it would return string on web & EasingFactoryFunction on native?

Yeah this sounds fine for now.

chaitanyadeorukhkar: @kamleshchandnani thoughts?

kamleshchandnani (changes requested):

packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.native.ts
chaitanyadeorukhkar: It returns an `EasingFunctionFactory` which is `type EasingFunctionFactory = { factory: () => EasingFunction };` and `EasingFunction` is `type EasingFunction = (value: number) => number;`

chaitanyadeorukhkar: I've changed the type for easing to `string | { factory: () => (value: number) => number };`

Avoiding importing a type from `react-native-reanimated` since this would cause issues on web

packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.native.ts
```diff
@@ -0,0 +1,1126 @@
...
+      '2xshort': 70,
+      xshort: 120,
+      short: 180,
+      long: 3000,
+      xlong: 5000,
+    },
+    duration: {
+      quick1: 70,
+      quick2: 150,
+      quick3: 200,
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: The issue would be with `moderate`, what would we call them? `moderate` & `xmoderate`? Which one would be moderate and which one would be `xmoderate`?

packages/blade/src/tokens/global/motion.ts
```diff
@@ -0,0 +1,132 @@
...
+    /** `900` milliseconds */
+    gentle3: number;
+  };
+  easing: {
+    /** Easings for all standard animations*/
+    standard: {
+      /** `cubic-bezier(0.5, 0, 0.3, 1.5)`
+       *
+       *  Returns a `string` for web & react-native-reanimated's Easing Function of type `EasingFunctionFactory` for native */
+      attentive: string;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Not sure what we can do for `Easing.Bezier` since the return type is always `{ factory: () => (value: number) => number };`. It stores the bezier values internally. It's sorta like this `Easing.Bezier(1,0,1,0)(0)`.

packages/blade/src/tokens/global/motion.ts
```diff
@@ -0,0 +1,132 @@
...
+      /** `cubic-bezier(0, 0, 0, 1)`
+       *
+       *  Returns a `string` for web & react-native-reanimated's Easing Function of type `EasingFunctionFactory` for native */
+      revealing: string;
+    };
+    /** Easings for all exit animations*/
+    exit: {
+      /** `cubic-bezier(0.7, 0, 0.5, 1)`
+       *
+       *  Returns a `string` for web & react-native-reanimated's Easing Function of type `EasingFunctionFactory` for native */
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Doesn't look great while typing but once expanded it looks good.
Screenshot 2022-04-18 at 3 59 21 PM

Screenshot 2022-04-18 at 3 59 47 PM

packages/blade/src/tokens/global/motion.ts
```diff
@@ -0,0 +1,138 @@
...
+  xlong: 5000;
+};
+
+type Easing = {
+  /** Easings for all standard animations*/
+  standard: {
+    /** `cubic-bezier(0.5, 0, 0.3, 1.5)`
+     *
+     *  Returns a `string` of `"cubic-bezier(...)"` for web & react-native-reanimated's Easing Function of type `EasingFunctionFactory` for native */
+    attentive: 'cubic-bezier(0.5, 0, 0.3, 1.5)' | { factory: () => (value: number) => number };
```
anuraghazra, line 41: Extract this factory function type to top, we are repeating it a lot.

packages/blade/src/tokens/global/motion.ts
```diff
@@ -0,0 +1,138 @@
...
+    /** `cubic-bezier(0.3, 0, 0.2, 1)`
+     *
+     *  Returns a `string` of `"cubic-bezier(...)"` for web & react-native-reanimated's Easing Function of type `EasingFunctionFactory` for native */
+    effective: 'cubic-bezier(0.3, 0, 0.2, 1)' | { factory: () => (value: number) => number };
```
anuraghazra, line 45: In here we are hard coding the string type. 
But in the actual easing values we are using that makeBezier() function which returns the narrowed template literal string. 

If you want you can extract the type and use it here, looks much cleaner 

```ts
export type MakeBezier<
  X1 extends number,
  Y1 extends number,
  X2 extends number,
  Y2 extends number
> = `cubic-bezier(${X1}, ${Y1}, ${X2}, ${Y2})`;

const makeBezier = <X1 extends number, Y1 extends number, X2 extends number, Y2 extends number>(
  x1: X1,
  y1: Y1,
  x2: X2,
  y2: Y2,
): MakeBezier<X1, Y1, X2, Y2> => {
  return `cubic-bezier(${x1}, ${y1}, ${x2}, ${y2})`;
};

```

Screenshot 2022-04-18 at 4 34 43 PM

chaitanyadeorukhkar: I think that's okay since these are constant values. Similar to  how `duration.quick3` has a type of `200`, this will have a type of `cubic-bezier(...)`

My goal is to ensure that `entrance.effective` should _always_ be `cubic-bezier(0, 0, 0, 1)`. If someone changes the definition of what `MakeBezier` is to something like this:
```js
export type MakeBezier<
  X1 extends number,
  Y1 extends number,
  X2 extends number,
  Y2 extends number
> = `some-incorrect-usage(${X1}, ${Y1}, ${X2}, ${Y2})`;
```

This will never throw any errors since my expectation and outcome will always be of type `MakeBezier`. Since we know what our expectation is, it's better to keep it as a constant `cubic-bezier(0, 0, 0, 1)`

anuraghazra: Okay cool then. makes sense.

anuraghazra (approved):
Looks mostly fine to me.

packages/blade/src/tokens/global/motion.ts
```diff
@@ -0,0 +1,140 @@
...
+  quick3: 200,
+  moderate1: 250,
+  moderate2: 300,
+  gentle1: 400,
+  gentle2: 600,
+  gentle3: 900,
+};
+
+/* makeBezier returns a string of `cubic-bezier()` for web & a react-native-reanimated Easing Function of type `EasingFunctionFactory` for native */
+const easing: Easing = {
```
kamleshchandnani, line 115: the type and autosuggestion of `Easing` how will that be used in action actually? I see you are directly using `makeBezier` function.

packages/blade/src/utils/makeBezier/makeBezier.native.ts
```diff
@@ -0,0 +1,8 @@
+import { Easing } from 'react-native-reanimated';
+import type { EasingFunctionFactory } from 'react-native-reanimated';
+
+const makeBezier = (x1: number, y1: number, x2: number, y2: number): EasingFunctionFactory => {
```
kamleshchandnani, line 4: nit: can we make it accept object as argument? i guess we have been following this for function accepting more than one arguments
```suggestion
const makeBezier = ({x1, y1, x2, y2}:{x1: number, y1: number, x2: number, y2: number}): EasingFunctionFactory => {
```

kamleshchandnani: no generic input type for `makeBezier` function? I see you did it for web

```
const makeBezier = <X1 extends number, Y1 extends number, X2 extends number, Y2 extends number>(
  x1: X1,
  y1: Y1,
  x2: X2,
  y2: Y2,
)
```

kamleshchandnani (changes requested):

packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.native.ts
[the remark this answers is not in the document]
kamleshchandnani: good question 🙈. from here: https://simple.wiktionary.org/wiki/moderate 

`lessModerate, moderate, moreModerate, mostModerate`

the issue is this scale of  1, 2, 3 looks very weird, and we shall try to find alternatives wherever possible

packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.native.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { Theme } from '../../';
+import paymentLightThemeWeb from './paymentLightTheme.web';
+
+const paymentLightTheme: Theme = {
+  ...paymentLightThemeWeb,
+  motion: {
+    ...paymentLightThemeWeb.motion,
+    easing: {
+      standard: {
+        attentive: Easing.bezier(0.5, 0, 0.3, 1.5),
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Since this was sort of a 'mock' theme, I wanted to put in raw values. Especially because we use this theme to do the check `expect(result.current.theme).toEqual(paymentLightTheme);` - here we should have our "expectations" as raw as possible instead of making it dependent on another function that can change over time leading to our "expectations" changing over time. So in case, makeBezier evolves & changes, it should force the person making the change to relook at these tests and change it here as well. Thoughts?

packages/blade/src/tokens/global/motion.ts
```diff
@@ -0,0 +1,140 @@
+import makeBezier from '../../utils/makeBezier';
+
+type Duration = {
+  /** `70` milliseconds */
+  quick1: 70;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: No, native needs `number` and web needs `string` with 'ms'. We'll need a make function to do this in this PR itself.

packages/blade/src/tokens/global/motion.ts
chaitanyadeorukhkar: Didn't get your point here. But if you're wondering how the typing helps here - if I pass a wrong input to `makeBezier`, it will return an incorrect value that won't fit with the expected easing and throw an error. For example,

Screenshot 2022-04-21 at 10 59 45 AM

packages/blade/src/utils/makeBezier/makeBezier.native.ts
chaitanyadeorukhkar: We can do this but I don't see the value in doing it for native. On web we're defining these generics so that we can form a return type of `cubic-bezier(${X1}, ${Y1}, ${X2}, ${Y2})` out of it. On native the return type is always `EasingFunctionFactory`. Am I missing something here?

packages/blade/src/tokens/global/motion.ts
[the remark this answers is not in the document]
chaitanyadeorukhkar: I've added `makeMotionTime`. Open to suggestions for a better name 😅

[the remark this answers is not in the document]
anuraghazra: `makeMotionUnit` ??

packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.native.ts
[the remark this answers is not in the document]
kamleshchandnani: makes complete sense 👍

packages/blade/src/utils/makeBezier/makeBezier.native.ts
kamleshchandnani: then to I would suggest why do we even need generics for web? we can just do

```
const makeBezier = (
  x1: number,
  y1: number,
  x2: number,
  y2: number,
): `cubic-bezier(${number}, ${number}, ${number}, ${number})` => {
  return `cubic-bezier(${x1}, ${y1}, ${x2}, ${y2})`;
};
```

packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.native.ts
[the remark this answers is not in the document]
chaitanyadeorukhkar: In that case we'll have to go in this sequence
`lessModerate, moreModerate, mostModerate, moderate` or `moderate, mostModerate, moreModerate, lessModerate`

The closer you are to moderate, the more moderate you are on either the higher or lower side.

Looking at just `lessModerate` I can't tell whether its faster or slower than `moderate`

----
This might make more sense in comparison then - at least it would follow the same syntax as `2xquick` & that of `delay`,
`2xquick, xquick, quick, moderate, xmoderate, gentle, xgentle, 2xgentle`

packages/blade/src/tokens/global/motion.ts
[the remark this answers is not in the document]
chaitanyadeorukhkar: Too vague :/

`makeMotionTime` at least makes me aware that this is a unit of time and I should use it for `delay`& `duration` but not for `easing`

packages/blade/src/utils/makeBezier/makeBezier.native.ts
chaitanyadeorukhkar: Generics makes sense for web since I can enforce that my return value consists the input values in the exact expected sequence of `x1, y1, x2, y2`

packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.native.ts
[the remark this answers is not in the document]
kamleshchandnani: > Looking at just lessModerate I can't tell whether its faster or slower than moderate

this same confusion will also exist for `moderate` and `xmoderate` 😞

here's my another suggestion

```javascript
moreQuick: 70,
quick: 150,
lessQuick: 200,
moderate: 250,
lessModerate: 300,
moreGentle: 400,
gentle: 600,
lessGentle: 900,
```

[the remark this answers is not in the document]
chaitanyadeorukhkar: I don't mind this but the design team might want to add more values here. For example, 900 was recently added when they saw a use case. This naming convention makes it very rigid and we won't be able to extend it in any way

packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.web.ts
```diff
@@ -1030,14 +1030,14 @@ const paymentLightTheme: Theme = {
...
-      moderate1: 250,
-      moderate2: 300,
-      gentle1: 400,
-      gentle2: 600,
-      gentle3: 900,
+      '2xquick': 70,
+      xquick: 150,
+      quick: 200,
+      moderate: 250,
+      xmoderate: 300,
```
anuraghazra, line 1037: Okay so bit confused here. 

In quick, `x`quick is reducing to 150 from 200ms

But in moderate, `x`moderate is increased to 300 from 250ms

How is the numbers are reducing and increasing with the same naming convention applied, is it relative or what?

anuraghazra: is `x` here means 1x or `small` as in xxs?

chaitanyadeorukhkar: It is relative, yes.

`gentle` is `400` but "extra gentle" implies more gently which means an increase in duration which makes `xgentle` as `600`

`quick` is `200` but "extra quick" implies more quickly which means a decrease in duration which makes `xquick` as `200`

In case of moderate, there is no obvious to decide what would be `xmoderate` - assumption is that it means "more moderate" similar to "more gentle" & "more quick" which would make `xmoderate` a value thats a little more than moderate. Hence `xmoderate` as `300`. 

We could think in reverse for moderate and make it lesser but that's just a hard call we need to make on which direction to take it 😓

anuraghazra (approved):

anuraghazra (approved):

kamleshchandnani (approved):

PR #442, opened 2022-04-13
fix(blade): infra changes and removal of examples directory

kamleshchandnani:


packages/blade/src/components/BladeProvider/BladeProvider.tsx
```diff
@@ -57,7 +58,11 @@ const BladeProvider = ({
     setColorScheme,
   };
 
-  return <ThemeContext.Provider value={themeContextValue}>{children}</ThemeContext.Provider>;
+  return (
+    <ThemeContext.Provider value={themeContextValue}>
```
chaitanyadeorukhkar, line 62: I'm guessing the usage of our `ThemeContext.Provider` is now limited to colorScheme since we're now able to consume our theme from styled components?

chaitanyadeorukhkar (approved):

packages/blade/src/components/BladeProvider/BladeProvider.tsx
kamleshchandnani: the consumers will still use our hooks and not styled-components hooks. this provider is essentially to help us write styled components with theme in context

chaitanyadeorukhkar: Makes sense

PR #443, opened 2022-04-13
fix(blade): add BaseText component

kamleshchandnani:


packages/blade/.storybook/react-native/Storybook.tsx
```diff
@@ -1,18 +1,33 @@
...
+    //   previewBorderColor: '#b3b3b3',
+    //   buttonTextColor: '#999999',
+    //   buttonActiveTextColor: '#444444',
+    // },
+  });
+
+  return (
+    <BladeProvider
+      // key={`${context.globals.themeTokens}-${context.globals.colorScheme}`}
+      themeTokens={paymentTheme}
```
kamleshchandnani, line 26: not putting efforts to add a theme switcher and color mode switcher like we have on web. might add later once the need arises

packages/blade/src/components/Typography/BaseText/BaseText.d.ts
```diff
@@ -0,0 +1,33 @@
+import type { ReactNode } from 'react';
+import type { ColorContrast } from '../../../tokens/theme/theme';
+import type { Theme } from '../../BladeProvider';
+
+type DotNotationStringToken<TokenType> = {
```
kamleshchandnani, line 5: this is to generate the `color` prop which supports values `action.text.primary.default`, `surface.text.muted.highContrast`, etc.

anuraghazra: We should put it in the jsdoc itself. 

```
/**
 * @template TokenType object
 * @description Tokenises objects to dot notation strings, eg: `a.b.c`
 */
```

packages/blade/src/components/Typography/BaseText/BaseText.d.ts
```diff
@@ -0,0 +1,32 @@
...
+  [K in keyof TokenType]: `${K & string}.${TokenType[K] extends Record<
+    string,
+    ColorContrast | string
+  >
+    ? Extract<keyof TokenType[K], number | string>
+    : DotNotationStringToken<TokenType[K]>}`;
+}[keyof TokenType];
+
+type FeedbackColors = `feedback.text.${DotNotationStringToken<
```
anuraghazra, line 14: Not very fond of not having type checking in a file which has complex types. 

Either we can change this file to `.ts` or add 

```ts
  "compilerOptions": {
    "skipLibCheck": false
  },
```

in tsconfig to enable typechecking of d.ts files

packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -0,0 +1,45 @@
...
+  lineHeight,
+  as,
+  children,
+  textAlign,
+}: BaseTextProps): ReactElement => {
+  const { theme } = useTheme();
+  const textColor = getIn(theme.colors, color);
+
+  if (isEmpty(textColor)) {
+    throw new Error(`[Blade:BaseText]: Invalid value ${color} for color prop passed`);
```
anuraghazra, line 24: Are you getting

```
Invalid type "BaseTextProps" of template literal expression.
```
here?

packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  return (
+    <StyledBaseText
+      theme={theme}
+      color={textColor}
```
anuraghazra, line 30: Okay so this is wrong, 

the textColor is a plain string of the color value `hsla(...)` but `StyledBaseText` is set to only accept the tokenized color strings (action.primary.brand)

`StyledBaseText` & `getBaseTextStyles` 's color prop should only be a plain string. 

This is not caught because `getIn` is returning `any`

packages/blade/src/components/Typography/BaseText/StyledBaseText.native.tsx
```diff
@@ -0,0 +1,30 @@
+import styled from 'styled-components/native';
+import getBaseTextStyles from './getBaseTextStyles';
+import type { BaseTextProps } from './BaseText.d';
+
+const StyledBaseText = styled.Text<BaseTextProps>(
```
anuraghazra, line 5: For some cases can't we just detect platform and based on that conditionally change logic instead of copy pasting the similar code for native and web? 

like `Platform.isNative ? styled.Text : styled.div`

packages/blade/src/components/Typography/BaseText/getBaseTextStyles.ts
```diff
@@ -0,0 +1,25 @@
...
+  theme,
+  color,
+  fontFamily,
+  fontSize,
+  fontWeight,
+  fontStyle,
+  textDecorationLine,
+  lineHeight,
+  textAlign,
+}: BaseTextProps): string => `
```
anuraghazra, line 14: Could make it a bit cleaner 

```ts
  const fonts = theme.typography.fonts;
  return `
    color: ${color};
    font-family: ${fonts.family[fontFamily]};
    font-size: ${makeTypography(fonts.size[fontSize])};
    font-weight: ${fonts.weight[fontWeight]};
    font-style: ${fontStyle};
    text-decoration-line: ${textDecorationLine};
    line-height: ${makeTypography(theme.typography.lineHeights[lineHeight])};
    ${textAlign ? `text-align: ${textAlign};` : ''}
  `;
```

packages/blade/src/components/Typography/BaseText/getBaseTextStyles.ts
```diff
@@ -0,0 +1,25 @@
...
+  color: ${color};
+  font-family: ${theme.typography.fonts.family[fontFamily]};
+  font-size: ${makeTypography(theme.typography.fonts.size[fontSize])};
```
anuraghazra, line 17: Not sure how you feel but I think it would be a lot more cleaner to use object here. 

```ts
  return {
    color,
    fontFamily: fonts.family[fontFamily],
    fontSize: makeTypography(fonts.size[fontSize]),
    fontWeight: fonts.weight[fontWeight],
    fontStyle,
    textDecorationLine,
    lineHeight: makeTypography(theme.typography.lineHeights[lineHeight]),
    ...(textAlign ? { textAlign } : {}),
  };
```

anuraghazra: Styled component does support this syntax nah? 🤔  https://styled-components.com/docs/advanced#style-objects

packages/blade/.babelrc.js
```diff
@@ -8,6 +8,7 @@ const configs = {
           {
             alias: {
               '@storybook/react': '@storybook/react-native',
+              '^styled-components$': 'styled-components/native',
```
chaitanyadeorukhkar, line 11: I see you've already done this here 😅 

Styled components recommend this approach: https://styled-components.com/docs/basics#simpler-usage-with-the-metro-bundler

#449

packages/blade/.storybook/react-native/storybook.requires.js
```diff
@@ -22,7 +22,11 @@ if (parameters) {
 argsEnhancers.forEach((enhancer) => addArgsEnhancer(enhancer));
 
 const getStories = () => {
-  return [require('../../src/components/Button/Button.stories.tsx')];
+  return [
+    require('../../src/components/Button/Button.stories.tsx'),
+    require('../../src/components/Typography/BaseText.stories.tsx'),
+    require('../../src/storybook-recipes/SimpleCardWithTokens/CardWithTokens.stories.tsx'),
```
chaitanyadeorukhkar, line 28: This crashes on react native since it internally uses html elements. Can we somehow ignore this for native?

packages/blade/package.json
```diff
@@ -45,9 +45,9 @@
     "test:react": "FRAMEWORK=REACT jest -c ./jest.web.config.js",
     "test:react-native": "FRAMEWORK=REACT_NATIVE jest -c ./jest.native.config.js",
     "react-native:get-stories": "sb-rn-get-stories --config-path=./.storybook/react-native",
-    "android": "yarn react-native:get-stories && FRAMEWORK=REACT_NATIVE react-native run-android",
-    "ios": "yarn react-native:get-stories && FRAMEWORK=REACT_NATIVE react-native run-ios",
-    "start": "yarn react-native:get-stories && FRAMEWORK=REACT_NATIVE react-native start --reset-cache",
+    "react-native:storybook:android": "yarn react-native:get-stories && FRAMEWORK=REACT_NATIVE react-native run-android",
+    "react-native:storybook:ios": "yarn react-native:get-stories && FRAMEWORK=REACT_NATIVE react-native run-ios",
+    "react-native:start": "yarn react-native:get-stories && FRAMEWORK=REACT_NATIVE react-native start --reset-cache",
```
chaitanyadeorukhkar, line 50:
```suggestion
    "react-native:storybook:start": "yarn react-native:get-stories && FRAMEWORK=REACT_NATIVE react-native start --reset-cache",
```

chaitanyadeorukhkar:
@kamleshchandnani add changeset

chaitanyadeorukhkar:
BaseText renders weirdly on iOS. 

Screenshot 2022-04-21 at 4 45 59 PM

packages/blade/src/components/Typography/BaseText/BaseText.tsx
anuraghazra: `any` is a top type, can be assigned to anything. thus didn't throw error.

chaitanyadeorukhkar:
This is happening because of line-heights. I guess the lineheight isn't enough for the font to fit. LineHeight works differently on web & mobile 😓 
https://github.com/razorpay/blade/pull/443#issuecomment-1105077261

packages/blade/.babelrc.js
kamleshchandnani: yeah but I'm not doing it for all the files. I'm basically doing where react-native will refer to a common component(common for web and native). rest at all the places we shall explicitly write `/native` since there are many places jaha pe types also fail since they are diff for web and native :()

packages/blade/src/components/Typography/BaseText/BaseText.d.ts
kamleshchandnani: but that will enable checks for tall the internal libs as well right?

packages/blade/src/components/Typography/BaseText/BaseText.tsx
kamleshchandnani: when?

packages/blade/src/components/Typography/BaseText/BaseText.tsx
kamleshchandnani: 😮 good catch. but i'm wondering the same how come TS didn't catch this error 🤔

packages/blade/src/components/Typography/BaseText/getBaseTextStyles.ts
kamleshchandnani: tried this but then again TS kicks in since now styled components default types need to be figured out as it'll be object and again web and native things might have diff things for type definitions then again I'll have to split this. however, I'll try to do this in later components maybe. right now just little bit scared with a level of abstraction we are doing

packages/blade/src/components/Typography/BaseText/StyledBaseText.native.tsx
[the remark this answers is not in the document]
kamleshchandnani: ideally it should be dot notation. let me fix this

kamleshchandnani: > BaseText renders weirdly on iOS.
> 
> Screenshot 2022-04-21 at 4 45 59 PM

as we has discussed today, we won't control this as `BaseText` is meant to be flexible but with tokens constraints in place

packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -18,21 +55,25 @@ const BaseText = ({
   ...rest
 }: BaseTextProps): ReactElement => {
   const { theme } = useTheme();
-  const textColor = getIn(theme.colors, color);
+  const textColor: string = getIn(theme.colors, color, '');
+  const themeFontfamily = theme.typography.fonts.family[fontFamily];
+  const themeFontSize = makeTypographySize(theme.typography.fonts.size[fontSize]);
+  const themeFontWeight = theme.typography.fonts.weight[fontWeight];
+  const themeLineHeight = makeTypographySize(theme.typography.lineHeights[lineHeight]);
```
anuraghazra, line 62: We are skipping the runtime checks for these theme accesses?

packages/blade/src/components/Typography/BaseText/BaseText.tsx
anuraghazra: *Resolved in meeting*

packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -18,21 +55,25 @@ const BaseText = ({
   ...rest
```
anuraghazra, line 55: Destructure the children here only and remove `rest` if not needed.

packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -2,8 +2,45 @@ import type { ReactElement } from 'react';
...
+ * @description Tokenises objects to dot notation strings, eg: `surface.text.normal.lowContrast`
+ */
+type DotNotationStringToken<TokenType> = {
+  [K in keyof TokenType]: `${K & string}.${TokenType[K] extends Record<
+    string,
+    ColorContrast | string
+  >
+    ? Extract<keyof TokenType[K], number | string>
+    : DotNotationStringToken<TokenType[K]>}`;
+}[keyof TokenType];
```
anuraghazra, line 21: **Self note:** 
We are using the slow version here, There is actually a bit of lag I can see on intellisence, We might need to move to faster one if the union is large later on.

packages/blade/src/components/Typography/BaseText/BaseText.tsx
anuraghazra: Fine now with the getInvalidColorPropValueError function

packages/blade/src/components/Typography/BaseText/BaseText.tsx
kamleshchandnani: nupe actually I forgot to push 🤦‍♂️

packages/blade/src/components/Typography/BaseText/BaseText.tsx
kamleshchandnani: the faster one is this only right? instead of directly doing keyOf on the object itself?

anuraghazra: No no the `keyof` version is faster, this one was the slower. 
Anyway I think lets keep it as is for now, we can later on change it if we face performance issues. No need to micro optimize rn.

packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -0,0 +1,128 @@
...
+  fontWeight: keyof Theme['typography']['fonts']['weight'];
+  fontStyle?: 'italic' | 'normal';
+  textDecorationLine?: 'line-through' | 'none';
+  lineHeight: keyof Theme['typography']['lineHeights'];
+  as?: 'code' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p' | 'span';
+  textAlign?: 'center' | 'justify' | 'left' | 'right';
+  children?: React.ReactNode | undefined;
+};
+
+export const getInvalidPropValueError = ({
```
chaitanyadeorukhkar, line 42: extract this into utils and accept another param for `component`?

packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -0,0 +1,128 @@
...
+  propName,
+  propValue,
+}: {
+  propName: string;
+  propValue: number | string;
+}): string => `[Blade:BaseText]: Invalid value ${propValue} passed for ${propName} prop`;
```
chaitanyadeorukhkar, line 48:
```suggestion
}): string => `[Blade: BaseText]: Invalid value ${propValue} passed for ${propName} prop`;
```

packages/blade/src/components/Typography/BaseText/StyledBaseText.native.tsx
```diff
@@ -0,0 +1,35 @@
+import styled from 'styled-components/native';
+import getBaseTextStyles from './getBaseTextStyles';
+import type { StyledBaseTextProps } from './StyledBaseText.d';
+
+export const ERROR_AS_PROP_NOT_SUPPORTED = `[Blade:BaseText]: "as" prop is not supported for BaseText for React Native platform`;
```
chaitanyadeorukhkar, line 5:
```suggestion
export const ERROR_AS_PROP_NOT_SUPPORTED = `[Blade: BaseText]: "as" prop is not supported for BaseText for React Native platform`;
```

packages/blade/src/components/Typography/BaseText/StyledBaseText.native.tsx
```diff
@@ -0,0 +1,35 @@
...
+const StyledBaseText = styled.Text<StyledBaseTextProps>(
```
chaitanyadeorukhkar, line 6: New line
```suggestion

const StyledBaseText = styled.Text<StyledBaseTextProps>(
```

packages/blade/src/components/Typography/BaseText/StyledBaseText.native.tsx
```diff
@@ -0,0 +1,35 @@
...
+    fontSize,
+    fontWeight,
+    fontStyle,
+    textDecorationLine,
+    lineHeight,
+    textAlign,
+    as,
+  }) => {
+    if (as) {
+      throw new Error(ERROR_AS_PROP_NOT_SUPPORTED);
```
chaitanyadeorukhkar, line 19: Lets not use a const here and export that const to be re-used in tests. 

If I change this const's value into something non-meaningful, our tests will never catch it. Let's hardcode the test to ensure the right error is being thrown.

packages/blade/src/components/Typography/BaseText/__tests__/BaseText.native.test.tsx
```diff
@@ -0,0 +1,287 @@
+import React from 'react';
+import renderWithTheme from '../../../../_helpers/testing/renderWithTheme.native';
```
chaitanyadeorukhkar, line 2: Out of scope from this PR but we need aliases soon.

packages/blade/src/components/Typography/BaseText/__tests__/BaseText.native.test.tsx
```diff
@@ -0,0 +1,287 @@
...
+      >
+        {displayText}
+      </BaseText>,
+    ).toJSON();
+    expect(renderedTree).toMatchSnapshot();
+  });
+
+  it('should render text in italics style with line-through', () => {
+    const displayText = 'Displaying some text';
+    const renderedTree = renderWithTheme(
```
chaitanyadeorukhkar, line 29: No more container being exported from testing library :(

packages/blade/src/components/Typography/BaseText/__tests__/BaseText.native.test.tsx
```diff
@@ -0,0 +1,287 @@
...
+          lineHeight="m"
+          fontSize={25}
+          fontWeight="regular"
+          fontStyle="italic"
+          textDecorationLine="line-through"
+          as="p"
+        >
+          {displayText}
+        </BaseText>,
+      ).toJSON();
```
chaitanyadeorukhkar, line 61: Why do you need to do `toJSON` here?
```suggestion
      );
```

packages/blade/src/components/Typography/BaseText/__tests__/BaseText.native.test.tsx
```diff
@@ -0,0 +1,287 @@
...
+          fontFamily="text"
+          lineHeight="m"
+          fontSize={25}
+          fontWeight="regular"
+          fontStyle="italic"
+          textDecorationLine="line-through"
+        >
+          {displayText}
+        </BaseText>,
+      ).toJSON();
```
chaitanyadeorukhkar, line 86:
```suggestion
      )
```

packages/blade/src/components/Typography/BaseText/__tests__/BaseText.native.test.tsx
```diff
@@ -0,0 +1,287 @@
...
+    } catch (error: unknown) {
+      if (error instanceof Error) {
+        expect(error.message).toEqual(
+          getInvalidPropValueError({ propName: 'color', propValue: color }),
```
chaitanyadeorukhkar, line 90: We should do a hard check here to make sure that even if someone messes with the function logic, our errors remain intact

packages/blade/src/utils/makeTypographySize/makeTypographySize.native.ts
```diff
@@ -0,0 +1,5 @@
+const makeTypographySize = (size: number): `${number}px` => {
```
chaitanyadeorukhkar, line 1:
```suggestion
const makeTypographySize = (size: T extends number): `${T}px` => {
```

packages/blade/src/utils/makeTypographySize/makeTypographySize.web.ts
```diff
@@ -0,0 +1,6 @@
+const makeTypographySize = (size: number): `${number}rem` => {
```
chaitanyadeorukhkar, line 1:
```suggestion
const makeTypographySize = <T extends number>(size: T): `${T}rem` => {
```

packages/blade/src/components/Typography/BaseText/BaseText.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+    fontFamily: 'text',
+    fontSize: 700,
+    fontWeight: 'regular',
+    fontStyle: 'normal',
+    textAlign: 'left',
+    textDecorationLine: 'none',
+    lineHeight: '5xl',
+    name: 'Storybook',
+    children: 'Base Text',
+    myprop: 'p',
```
chaitanyadeorukhkar, line 19: Remove this
```suggestion
```

packages/blade/src/components/Typography/BaseText/BaseText.tsx
kamleshchandnani: yes that's the plan, don't want all the abstractions in one shot

packages/blade/src/components/Typography/BaseText/__tests__/BaseText.native.test.tsx
kamleshchandnani: yesss

chaitanyadeorukhkar (approved):
LETSSS GOOOOO 🚀

anuraghazra (approved):

PR #451, opened 2022-04-22
feat: add icon components

chaitanyadeorukhkar:
- Add the following components that would act as building blocks for our icons:
	  1. `Svg`
	  2. `Path`
	  3. `Rect`
	  4. `Defs`
	  5. `ClipPath`
	  6. `G`
- Add `CreditCardIcon` component
- Add `RupeeIcon` component

Decisions PR: https://github.com/razorpay/blade/pull/452
> We can implement other SVG components like `Circle`, `Line`, `Ellipse`, etc. as and when the need arises 

**Story Preview:** 
Screenshot 2022-05-04 at 3 51 34 PM

PR #454, opened 2022-04-27
feat(blade): add text component

kamleshchandnani:
This PR adds `Text` component


packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -35,7 +37,9 @@ export type BaseTextProps = {
   lineHeight: keyof Theme['typography']['lineHeights'];
   as?: 'code' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p' | 'span';
   textAlign?: 'center' | 'justify' | 'left' | 'right';
-  children?: React.ReactNode | undefined;
+  truncateAfterLines?: number;
+  className?: string;
```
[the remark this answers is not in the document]
dhruvdutt: Shouldn't this have `_` in prefix?
`_className`

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+type TextCommonProps = {
+  type: TextTypes;
+  truncateAfterLines?: number;
+  children?: React.ReactNode;
+};
+
+type TextBodyVariant = TextCommonProps & {
+  variant: 'body';
+  weight: keyof Theme['typography']['fonts']['weight'];
```
dhruvdutt, line 17: Noice 🔥

packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -35,7 +37,9 @@ export type BaseTextProps = {
   lineHeight: keyof Theme['typography']['lineHeights'];
   as?: 'code' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p' | 'span';
   textAlign?: 'center' | 'justify' | 'left' | 'right';
-  children?: React.ReactNode | undefined;
+  truncateAfterLines?: number;
```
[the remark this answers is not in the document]
anuraghazra: Nice.

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,97 @@
...
+};
+
+type TextCaptionVariant = TextCommonProps & {
+  variant: 'caption';
+  weight: keyof Pick<Theme['typography']['fonts']['weight'], 'regular'>;
```
dhruvdutt, line 22: Evaluated type here is a [this](https://github.com/razorpay/blade/blob/57635d733d4d0635a2a6072c55629d6f3e560d3d/packages/blade/src/tokens/global/typography.ts#L5)? And above in `TextBodyVariant` is an [object](https://github.com/razorpay/blade/blob/57635d733d4d0635a2a6072c55629d6f3e560d3d/packages/blade/src/tokens/global/typography.ts#L4-L7)?
Seems like inconsistent for the same prop name `weight`.

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,97 @@
...
+});
+
+const Text = <T extends { variant: 'body' | 'caption' }>({
+  variant,
+  weight,
+  type,
+  truncateAfterLines,
+  children,
+}: TextProps<T>): ReactElement => {
+  const props = { truncateAfterLines, ...getProps({ variant, type, weight }) };
```
dhruvdutt, line 93: The default for `truncateAfterLines` is `undefined`? Is the behaviour the same for RN and Web in this case?
The way it overflows or other behaviors?

packages/blade/src/utils/getPlatform/getPlatform.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  if (typeof document !== 'undefined') {
+    return 'browser';
+  }
+
+  if (typeof process !== 'undefined') {
+    return 'node';
+  }
+
+  return undefined;
```
dhruvdutt, line 16: How about throwing an error here? If it's none of the above platforms, could be some bug actually?

packages/blade/src/components/Typography/BaseText/BaseText.tsx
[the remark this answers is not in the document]
kamleshchandnani: Nupe since no one is explicitly adding this class(yes consumers can abuse this), it's styled-component that is adding this class. also this is anyways `BaseText` whose job is to let people customise things when our regular text doesn't fits their need. None of the other Text components have this prop so it's safe.

packages/blade/src/components/Typography/Text/Text.tsx
kamleshchandnani: yes

packages/blade/src/utils/getPlatform/getPlatform.ts
kamleshchandnani: could be any other platform too in the future like watchOS or something. regardless i need to return since it's a function expecting a type, else TS will beat me off :D 

I'll add `console.error` though.

packages/blade/src/components/Typography/BaseText/BaseText.stories.tsx
```diff
@@ -1,9 +1,11 @@
+/* eslint-disable react/display-name */
```
chaitanyadeorukhkar, line 1: See this https://github.com/razorpay/blade/pull/451/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R102

packages/blade/src/components/Typography/BaseText/BaseText.stories.tsx
```diff
@@ -17,10 +19,32 @@ export default {
...
+          <Subtitle>
+            The BaseText component is a wrapper component based on Blade&apos;s token system and
+            should be used only in rarest exceptions where your Typography needs are not satisfied
+            by Title, Heading or Text typography components(eg: highlighting a word in the Title
+            with gradients in a landing page).
+          </Subtitle>
+          <Title>Example</Title>
+          <Primary />
+          <Title>Properties</Title>
+          <ArgsTable story={PRIMARY_STORY} />
```
chaitanyadeorukhkar, line 36: Same as `ArgsTable` but seems more readable
```suggestion
          <Props story={PRIMARY_STORY} />
```

packages/blade/src/components/Typography/BaseText/BaseText.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Nice

packages/blade/src/components/Typography/Text/Text.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+import type { TextProps } from './';
+import TextComponent from './';
+//The Text component is a wrapper component that will apply typography styles to the text inside.
+
+const TextStoryMeta: Meta<TextProps<{ variant: 'body' | 'caption' }>> = {
+  title: 'Components/Typography/Text',
+  component: TextComponent,
+  args: {
+    variant: 'body',
+    weight: 'bold',
```
chaitanyadeorukhkar, line 13: Why bold by default? Feels weird. Keep regular only?

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,97 @@
...
+  forwardedAs?: BaseTextProps['as'];
+};
+
+const getProps = <T extends { variant: 'body' | 'caption' }>({
+  variant,
+  type,
+  weight,
+}: Pick<TextProps<T>, 'type' | 'variant' | 'weight'>): BaseTextProps & TextForwardedAs => {
+  const props: BaseTextProps & TextForwardedAs = {
+    color: `surface.text.${type}.lowContrast`,
```
chaitanyadeorukhkar, line 45: Getting an eslint error here. Are you not getting this?

Screenshot 2022-05-04 at 5 08 35 PM

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const Text = <T extends { variant: 'body' | 'caption' }>({
+  variant,
+  weight,
+  type,
+  truncateAfterLines,
+  children,
+}: TextProps<T>): ReactElement => {
+  const props = { truncateAfterLines, ...getProps({ variant, type, weight }) };
+  return <StyledText {...props}>{children}</StyledText>;
```
chaitanyadeorukhkar, line 94: Why is the github security bot not annoying you here? I got a lot of warnings for dangerously spreading props in the Icon PR

packages/blade/src/components/Typography/Text/__tests__/Text.native.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+        {displayText}
+      </Text>,
+    ).toJSON();
+    expect(renderedTree).toMatchSnapshot();
+  });
+
+  it('should throw error when variant is "caption" but weight "bold" is passed', () => {
+    try {
+      const displayText = 'Displaying some text';
+      // @todo: when variant is caption then weight should not even show bold
```
chaitanyadeorukhkar, line 32: This is done, right?

packages/blade/src/components/Typography/Text/Text.stories.tsx
kamleshchandnani: was playing around and forgot to remove it 🤦

packages/blade/src/components/Typography/Text/Text.tsx
kamleshchandnani: nupe, even the validate check passed 🤔

packages/blade/src/components/Typography/BaseText/BaseText.stories.tsx
kamleshchandnani: can't find its documentation. can you point me to one?

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,97 @@
...
+  if (truncateAfterLines) {
+    if (getPlatform() === 'react-native') {
+      return {
+        numberOfLines: truncateAfterLines,
+      };
+    }
+    return {
+      overflow: 'hidden',
+      display: '-webkit-box',
+      '-webkit-line-clamp': `${truncateAfterLines}`,
```
anuraghazra, line 79: Only supported in webkit? What's the browser compat here?

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,97 @@
...
+  weight,
+}: Pick<TextProps<T>, 'type' | 'variant' | 'weight'>): BaseTextProps & TextForwardedAs => {
+  const props: BaseTextProps & TextForwardedAs = {
+    color: `surface.text.${type}.lowContrast`,
+    fontSize: 25,
+    fontWeight: weight,
+    fontStyle: 'normal',
+    lineHeight: 's',
+    fontFamily: 'text',
+    forwardedAs: getPlatform() !== 'react-native' ? 'p' : undefined,
```
anuraghazra, line 51: I mean <Text> can also be span right? or are we not allowing that? What if I want span like inline behaviour?

packages/blade/src/components/Typography/Text/Text.tsx
anuraghazra, line 93: Should probably explicitly say type the props as`BaseTextProps & TextForwardedAs`, just in case someone adds extra props {hello: string}, because spreading props wouldn't do typechecking here.

packages/blade/src/components/Typography/Text/Text.tsx
kamleshchandnani: i'll add `line-clamp` too maybe just forgot. the browser compact is pretty good: https://caniuse.com/?search=line-clamp

packages/blade/src/utils/getPlatform/getPlatform.ts
anuraghazra: Would probably be better if we return consistent string here? 

Like `return "unknown"`

That we atleast we don't have to check for undefined when using this function. We can just assert on string itself.

packages/blade/src/utils/getPlatform/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from './getPlatform';
```
anuraghazra, line 1: We named the file `getPlatform` and even using it as `getPlatform()` but the function name is `getPlatformType`

packages/blade/src/components/Typography/Text/Text.tsx
kamleshchandnani: Right now we have capped it to `p` only.

packages/blade/src/utils/getPlatform/index.ts
kamleshchandnani: nice catch. the intention was `getPlatformType` itself. fixing this

packages/blade/src/components/Typography/Text/Text.tsx
anuraghazra: Nice then.

packages/blade/src/components/Typography/Text/Text.tsx
anuraghazra: lol this looks soo overkill because there's just two values 'regular' & 'bold' :p

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,101 @@
...
+  variant: 'body';
+  weight: keyof Theme['typography']['fonts']['weight'];
+};
+
+type TextCaptionVariant = TextCommonProps & {
+  variant: 'caption';
+  weight: keyof Pick<Theme['typography']['fonts']['weight'], 'regular'>;
+};
+
+export type TextProps<T> = T extends {
```
anuraghazra, line 25: We should probably add a small description here, Why we are doing it this way instead of overloads or unions. 

```ts
/**
 * Conditionally changing props based on variant
 * Because overloads or union gives wrong intellisense.
 */
```

anuraghazra (dismissed):
Mostly LGTM!

packages/blade/src/components/Typography/Text/Text.tsx
kamleshchandnani: aye sir

kamleshchandnani: done

anuraghazra (approved):

packages/blade/src/components/Typography/Text/Text.tsx
kamleshchandnani: consistency 🙈

chaitanyadeorukhkar (approved):

PR #461, opened 2022-05-05
feat(Blade): add `Title`component

kamleshchandnani:
This PR adds `Title` component

The API for `Title` component can be found under [Typography/Text/_decisions](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Typography/_decisions/decisions.md)


packages/blade/src/@types/globals.d.ts
```diff
@@ -2,7 +2,7 @@
 import 'styled-components';
 import 'styled-components/native';
 import type { Theme } from '../components/BladeProvider';
-// and extend them!
+import '@testing-library/jest-dom';
```
kamleshchandnani, line 5: adding to make the types globally available for test files, otherwise I've to include this in every test file else `expect().toBeInTheDocument()` kind of method gives TS errors `cannot call method on 'any'`

packages/blade/src/components/Typography/Title/Title.tsx
```diff
@@ -0,0 +1,48 @@
...
+    props.lineHeight = '6xl';
+    props.as = getPlatform() !== 'react-native' ? 'h1' : undefined;
+  }
+
+  return props;
+};
+
+const Title = ({ variant, type, children }: TitleProps): ReactElement => {
+  const props = getProps({ variant, type });
+  return <BaseText {...props}>{children}</BaseText>;
```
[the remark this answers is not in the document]
kamleshchandnani: @chaitanyadeorukhkar tumhari khwaish puri hui 😆

nashcheez (dismissed):

packages/blade/src/components/Typography/Title/Title.tsx
```diff
@@ -0,0 +1,45 @@
...
+    fontWeight: 'bold',
+    fontStyle: 'normal',
+    lineHeight: 's',
+    fontFamily: 'text',
+  };
+
+  if (variant === 'small') {
+    props.fontSize = 600;
+    props.lineHeight = '4xl';
+    props.as = getPlatformType() !== 'react-native' ? 'h3' : undefined;
```
anuraghazra, line 26: Maybe we should instead check for `getPlatformType() === 'web'`

Only for web we will set h1,2,3 else undefined, otherwise for `unknown` platform too it will set h1,2,3

packages/blade/src/components/Typography/Title/__tests__/__snapshots__/Title.web.test.tsx.snap
```diff
@@ -0,0 +1,46 @@
...
+</div>
+`;
+
+exports[`<Title /> should render Title with variant "small" 1`] = `
+<div>
+  <h3
+    class="StyledBaseTextweb__StyledBaseText-sc-1t5p7u8-0 jOhoak"
+    color="hsla(217, 56%, 17%, 1)"
+    font-family="Lato, -apple-system, BlinkMacSystemFont, \\"Segoe UI\\", Roboto, Helvetica, sans-serif, \\"Apple Color Emoji\\", \\"Segoe UI Emoji\\", \\"Segoe UI Symbol\\""
+    font-size="1.5625rem"
```
anuraghazra, line 39: Hmmm, I never noticed this. But these props should not be in the html. 🤔

anuraghazra: So styled components seem to injecting this, I wonder why. Only for handful of props it does, if the prop name is `fontSize`, 'fontWeight' etc. Won't do it for other props like `lineHeight`

packages/blade/src/components/Typography/Title/Title.tsx
```diff
@@ -0,0 +1,45 @@
+import type { ReactElement } from 'react';
+import type { TextTypes } from '../../../tokens/theme/theme';
+import getPlatformType from '../../../utils/getPlatformType';
```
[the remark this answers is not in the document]
kamleshchandnani: yes it is deprecated, we are not using it for web we just use it to detect for React Native because that's how React Native does it internally and exposes that to us - [Ref code from React Native](https://github.com/facebook/react-native/blob/6b6adcc111/Libraries/Core/setUpNavigator.js)

packages/blade/src/components/Typography/Title/Title.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+  },
+  parameters: {
+    docs: {
+      page: () => (
+        <>
+          <StorybookTitle />
+          <Subtitle>
+            The Title Component make a bold visual statement. Use them to create impact when the
+            main goal is visual storytelling. For example, use Title as marketing content on landing
+            pages or to capture attention during onboarding.
```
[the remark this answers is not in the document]
kamleshchandnani: Yeah, the terminology and intent is the same as a hero but the generic term is Title which is generic. Could be hero sections or could be a free form section in the page

packages/blade/src/components/Typography/Title/Title.tsx
kamleshchandnani: it's browser or node(for ssr) so I'll have to write two conditions to evaluate this.

packages/blade/src/components/Typography/Title/__tests__/__snapshots__/Title.web.test.tsx.snap
kamleshchandnani: yep we need to filter these props using [`shouldForwardProp`](https://styled-components.com/docs/api#shouldforwardprop) but since it's not affecting anything I didn't wanted to add another layer to this.

packages/blade/src/components/Typography/Title/Title.tsx
anuraghazra: Oh okay i see, fine then.

packages/blade/src/components/Typography/Title/Title.tsx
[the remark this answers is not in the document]
nashcheez: Nice! Funny they decided to go with "ReactNative" instead of "Gecko". One would assume the distinction for the platform being made on the basis of `userAgent` and not on a property that was being demised. 😄

packages/blade/src/components/Typography/Title/Title.tsx
kamleshchandnani: made the changes, even I was not liking that previously :D

anuraghazra (approved):

PR #463, opened 2022-05-09
docs(rfc): add accessibility RFC

anuraghazra:
## [View Rendered Text](https://github.com/razorpay/blade/blob/anu/a11y-rfc/rfcs/2022-04-09-accessibility.md)


rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,167 @@
...
+- [References](#references)
+
+# Summary
+
+
+## What is Accessibility?
+
+Accessibility is the practice of making websites usable by people of all abilities and disabilities.   
+This includes people with visual, auditory, physical, and cognitive impairments, as well as those with temporary disabilities such as broken bones, color blindness, and temporary loss of vision, but the practice of making sites accessible also benefits other groups such as people using mobile phones, people with temporary disabilities, situational limitations.
+
```
divyanshu013, line 27: Commonly I see people confuse a11y with something that isn't important for everyone. Perhaps we should add a note on how / why it's crucial for everything - for example filling forms, closing modal, etc.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,167 @@
...
+- Unlabeled password toggle icon 
+- Wrongly focus on text input's placeholder text
+
+
+https://user-images.githubusercontent.com/35374649/167380507-71e60b7f-d36e-42b3-b649-911400e2c66f.mp4
+
+
+Even though only a few of the teams are on boarded to blade, when more teams will migrate this will become a problem since every app/site will be inaccessible if we don’t implement accessibility in blade components. 
+
+Adding accessible components to blade will iron out all these inconsistencies and teams can build more accessible apps without thinking about it.
```
divyanshu013, line 121: Think people will still need to think about a11y though they won't have to worry about implementation for each and every usecase. Somethings which we can't handle correctly in blade itself for example - tabbing order of inputs in forms

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Why are we doing this? 
+
+Without proper accessibility, certain functionality used in web sites is not available to some users with disabilities, especially people who rely on screen readers and people who cannot use a mouse. 
+
+Making our products accessible is important for two reasons.  
+- it's the right thing to do. We believe that everyone deserves to be able to use our products, regardless of their abilities.
+- It's good for business & branding. By making our products accessible, we can expand our potential customer base and make our products more appealing to a wider range of people, plus it helps create a good faith between consumers and our product. 
+
+**Making customers happy & respecting human rights:**
```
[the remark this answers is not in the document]
anuraghazra: Yep, In india only government sites have mandatory a11y requirements by law. We could add few lines about the law too.

rfcs/2022-04-09-accessibility.md
anuraghazra: Correct, will add few lines to highlight why accessibility is for everyone.

rfcs/2022-04-09-accessibility.md
anuraghazra: Hmm yeah i meant that 🤔 I think wording can be improved here. For example even if we provide a <Select /> element users will still have to provide a accessible label to define what elements are in that slelect, like "aria-label='Country codes'"

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,167 @@
...
+**Making customers happy & respecting human rights:**  
+
+In 2020, A 29 year old banker who is visually impaired opened petitions to make [zomato](https://www.change.org/p/zomato-make-zomato-accessible-with-screen-readers-for-visually-challenged-people) & [swiggy](https://www.change.org/p/swiggy-make-swiggy-accessible-with-screen-readers-in-android-for-visually-challenged-people) accessible because he couldn’t order food from it due to their inaccessible apps. 
+Over **15,000 & 27,000 people had signed the petition.**
+Later zomato, swiggy listened to the people and [improved their accessibility.](https://theprint.in/tech/why-the-visually-challenged-are-signing-petitions-addressed-to-apps-such-as-zomato-swiggy/507282/) 
+
+
+## The current state  
+
+In current state, blade components are not very accessible out of the box, neither are some of our production grade mobile applications.
```
kamleshchandnani, line 104:
```suggestion
In the current state, blade components are not very accessible out of the box, and neither are our existing frontend products. 
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,167 @@
...
+https://user-images.githubusercontent.com/35374649/167380507-71e60b7f-d36e-42b3-b649-911400e2c66f.mp4
+
+
+Even though only a few of the teams are on boarded to blade, when more teams will migrate this will become a problem since every app/site will be inaccessible if we don’t implement accessibility in blade components. 
+
+Adding accessible components to blade will iron out all these inconsistencies and teams can build more accessible apps without thinking about it. 
+
+## What is the expected outcome? 
+
+The goal of this RFC is to lay the foundations for implementing accessibility right into blade, so that blade consumers don’t have to think about accessibility while building their products:
```
kamleshchandnani, line 125:
```suggestion
The goal of this RFC is to lay the foundations so that we can ship blade components that are accessible out of the box such that consumers don't have to worry about every minute detail. Here are couple of areas that this RFC will touch upon
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,167 @@
...
+
+
+Even though only a few of the teams are on boarded to blade, when more teams will migrate this will become a problem since every app/site will be inaccessible if we don’t implement accessibility in blade components. 
+
+Adding accessible components to blade will iron out all these inconsistencies and teams can build more accessible apps without thinking about it. 
+
+## What is the expected outcome? 
+
+The goal of this RFC is to lay the foundations for implementing accessibility right into blade, so that blade consumers don’t have to think about accessibility while building their products: 
+
```
kamleshchandnani, line 126: the outcomes in form of functionality like what all things we are scoping to right now for example, Keyboard navigation, screen readers, focus management are not added here. do you plan to add them later?

divyanshu013:
Sharing a couple of bookmarks I'd on the topic, might find something useful there :)
- [facebook.com a11y](https://engineering.fb.com/2020/07/30/web/facebook-com-accessibility/)
- [Inclusive design - some good resrouces](https://github.com/inclusive-design-and-development/handbook/tree/main/leadership/bringing_others_along/ingredients_for_case)

rfcs/2022-04-09-accessibility.md
anuraghazra: Yes we will be adding that, outcome I've kept at high level.

rfcs/2022-04-09-accessibility.md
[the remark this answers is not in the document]
anuraghazra: Added

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,167 @@
...
+  - Example: Using your phone one handedly in a crowded metro station
+
+<img src="./images/accessibility/ms-inclusive-design.png" alt="Microsoft Inclusive Design Shows Spectrum Of Disabilities" width="25%" />
+
+# Motivation
+- Why are we doing this?
+- What use cases does it support?
+- What is the expected outcome?
+
+## Why are we doing this?
```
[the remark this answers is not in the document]
nashcheez: More rules and policies are laid out [here](https://www.w3.org/WAI/policies/).

rfcs/2022-04-09-accessibility.md
anuraghazra: Will add this video https://www.youtube.com/watch?v=3f31oufqFSM

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+Making our products accessible is important for two reasons.  
+- it's the right thing to do. We believe that everyone deserves to be able to use our products, regardless of their abilities.
+- It'll help build trust with our customers. By making our products accessible, we can expand our potential customer base and make our products more appealing to a wider range of people. 
+
+> [The Business Case Study for Accessibility](https://www.w3.org/WAI/business-case/)
+
+**Making customers happy & respecting human rights:**  
+
+> In 2020, A 29 year old banker who is visually impaired opened petitions to make [zomato](https://www.change.org/p/zomato-make-zomato-accessible-with-screen-readers-for-visually-challenged-people) & [swiggy](https://www.change.org/p/swiggy-make-swiggy-accessible-with-screen-readers-in-android-for-visually-challenged-people) accessible because they couldn’t order food from it due to their inaccessible apps.
```
nashcheez, line 114:
```suggestion
> In 2020, A 29 year old banker who is visually impaired opened petitions to make [Zomato](https://www.change.org/p/zomato-make-zomato-accessible-with-screen-readers-for-visually-challenged-people) & [Swiggy](https://www.change.org/p/swiggy-make-swiggy-accessible-with-screen-readers-in-android-for-visually-challenged-people) accessible because they couldn’t order food from it due to their inaccessible apps.  
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+- it's the right thing to do. We believe that everyone deserves to be able to use our products, regardless of their abilities.
+- It'll help build trust with our customers. By making our products accessible, we can expand our potential customer base and make our products more appealing to a wider range of people. 
+
+> [The Business Case Study for Accessibility](https://www.w3.org/WAI/business-case/)
+
+**Making customers happy & respecting human rights:**  
+
+> In 2020, A 29 year old banker who is visually impaired opened petitions to make [zomato](https://www.change.org/p/zomato-make-zomato-accessible-with-screen-readers-for-visually-challenged-people) & [swiggy](https://www.change.org/p/swiggy-make-swiggy-accessible-with-screen-readers-in-android-for-visually-challenged-people) accessible because they couldn’t order food from it due to their inaccessible apps.  
+> Over **15,000 & 27,000 people had signed the petition.**  
+> Later zomato, swiggy listened to the people and [improved their accessibility.](https://theprint.in/tech/why-the-visually-challenged-are-signing-petitions-addressed-to-apps-such-as-zomato-swiggy/507282/)
```
nashcheez, line 116:
```suggestion
> Later Zomato & Swiggy listened to the people and [improved their accessibility.](https://theprint.in/tech/why-the-visually-challenged-are-signing-petitions-addressed-to-apps-such-as-zomato-swiggy/507282/) 
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+**Making customers happy & respecting human rights:**  
+
+> In 2020, A 29 year old banker who is visually impaired opened petitions to make [zomato](https://www.change.org/p/zomato-make-zomato-accessible-with-screen-readers-for-visually-challenged-people) & [swiggy](https://www.change.org/p/swiggy-make-swiggy-accessible-with-screen-readers-in-android-for-visually-challenged-people) accessible because they couldn’t order food from it due to their inaccessible apps.  
+> Over **15,000 & 27,000 people had signed the petition.**  
+> Later zomato, swiggy listened to the people and [improved their accessibility.](https://theprint.in/tech/why-the-visually-challenged-are-signing-petitions-addressed-to-apps-such-as-zomato-swiggy/507282/) 
+
+**Accessibility laws & policies** 
+
+While in India, there's no mandatory law to building accessible web apps,
```
nashcheez, line 120:
```suggestion
While in India, there are no mandatory laws to build accessible web apps, 
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+**Making customers happy & respecting human rights:**  
+
+> In 2020, A 29 year old banker who is visually impaired opened petitions to make [zomato](https://www.change.org/p/zomato-make-zomato-accessible-with-screen-readers-for-visually-challenged-people) & [swiggy](https://www.change.org/p/swiggy-make-swiggy-accessible-with-screen-readers-in-android-for-visually-challenged-people) accessible because they couldn’t order food from it due to their inaccessible apps.  
+> Over **15,000 & 27,000 people had signed the petition.**  
+> Later zomato, swiggy listened to the people and [improved their accessibility.](https://theprint.in/tech/why-the-visually-challenged-are-signing-petitions-addressed-to-apps-such-as-zomato-swiggy/507282/) 
+
+**Accessibility laws & policies** 
+
+While in India, there's no mandatory law to building accessible web apps, 
+There's still [policies/laws](https://www.w3.org/WAI/policies/india/) which are passed to ensure rights of disabled people remains intact.
```
nashcheez, line 121:
```suggestion
there are still [policies/laws](https://www.w3.org/WAI/policies/india/) that are passed to ensure that the rights of disabled people remain intact.
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+- Missing accessibilityLabel in input box.
+- Missing accessibilityRole=button in buttons
+- Missing accessibilityRole=link in links
+- Grouping text together so they get announced at once.
+- Unlabeled password toggle icon 
+- Wrongly focus on text input's placeholder text
+
+https://user-images.githubusercontent.com/35374649/167380507-71e60b7f-d36e-42b3-b649-911400e2c66f.mp4
+
+Even though only a few of the teams are on boarded to blade, when more teams will migrate this will become a problem since every app/site will be inaccessible if we don’t implement accessibility in blade components.
```
nashcheez, line 141:
```suggestion
Even though only a few teams are onboarded to blade, this could become a major problem when more teams migrate since every app/site won't be completely accessible if we don’t implement accessibility centrally in the core blade components. 
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+https://user-images.githubusercontent.com/35374649/167380507-71e60b7f-d36e-42b3-b649-911400e2c66f.mp4
+
+Even though only a few of the teams are on boarded to blade, when more teams will migrate this will become a problem since every app/site will be inaccessible if we don’t implement accessibility in blade components. 
+
+Adding accessible components to blade will iron out all these inconsistencies and teams can build more accessible apps with minimal effort.
+
+## What is the expected outcome? 
+
+The goal of this RFC is to lay the foundations so that we can ship blade components that are accessible out of the box such that consumers don't have to worry about every minute detail. Here are couple of areas that this RFC will touch upon:
```
nashcheez, line 147:
```suggestion
The goal of this RFC is to lay the foundations so that we can ship accessible components out of the box so that consumers don't have to worry about every minute detail. Here are couple of areas that this RFC will touch upon: 
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+  We will pick up motion accessibility at a later point, we are first prioritizing foundations.
+
+# Keyboard Accessibility
+
+> [Specification](https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation.html)
+
+**Target:** Everyone and especially people with motor disabilities who use a keyboard to navigate.  
+**Goal:** Ensure users who cannot use the mouse (blind, motor disabilities) can access the crucial parts of the app through keyboard.  
+
+For a web page to be accessible, all interactive elements must be operable via the keyboard, Not all users are able to use a mouse to navigate a web page, Keyboard-only and screen reader users rely on navigating and using a web page with only a keyboard.
```
nashcheez, line 198:
```suggestion
For a web page to be accessible, all interactive elements must be operable via the keyboard. Not all users are able to use a mouse to navigate a web page, keyboard-only and screen reader users rely on navigating and using a web page with a physical or a virtual keyboard.
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+- If necessary, use CSS to control the visual presentation of the elements on the page.
+- [Avoid using tabindex](https://www.w3.org/WAI/WCAG21/Techniques/failures/F44) values of 1 or greater to change the default keyboard navigation order.
+
+## Using tabindex
+
+`tabindex` attribute authors can make the element focusable and appear in [sequential focus order](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation)
+
+Setting tabindex to any non-negative integer makes elements focusable, allow or prevent them from being sequentially focusable, and determine their relative ordering
+
+When set to `0`, the element becomes focusable by keyboard and via programatic means with `focus()`.
```
nashcheez, line 229:
```suggestion
When set to `0`, the element becomes focusable by keyboard and via programmatic means with `focus()`. 
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+- [Avoid using tabindex](https://www.w3.org/WAI/WCAG21/Techniques/failures/F44) values of 1 or greater to change the default keyboard navigation order.
+
+## Using tabindex
+
+`tabindex` attribute authors can make the element focusable and appear in [sequential focus order](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation)
+
+Setting tabindex to any non-negative integer makes elements focusable, allow or prevent them from being sequentially focusable, and determine their relative ordering
+
+When set to `0`, the element becomes focusable by keyboard and via programatic means with `focus()`. 
+When set to `-1`, the element becomes focusable programatically, but it does not become part of the keyboard focus order.
```
nashcheez, line 230:
```suggestion
When set to `-1`, the element becomes focusable programmatically, but it does not become part of the keyboard focus order.
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+The following table describes tabindex behavior in modern browsers:
+
+| tabindex attribute |	Focusable with mouse or programatically |	Tab navigable |
+| :-- | :-- | :-- |
+| not present	| The user agent will decide | The user agent will decide |
+| Negative (tabindex="-1")	| Yes | No, can only be focused programatically |
+| Zero (tabindex="0") | Yes | Yes, In tab order relative to element's position in document | 
+| Positive (tabindex="2") | Yes | Yes, tabindex value determines where this element is positioned in the tab order | 
+
+> Warning: avoid using positive values for tabindex. Using positive values means authors will have to set (and maintain) tabindex values for all focusable elements on the page whenever they use one or more positive values for tabindex.
```
nashcheez, line 241:
```suggestion
> Warning: Avoid using positive values for tabindex. Using positive values means authors will have to set (and maintain) tabindex values for all focusable elements on the page whenever they use one or more positive values for tabindex.  
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+> *[Failure of Success Criterion 2.4.3](https://www.w3.org/WAI/WCAG21/Techniques/failures/F44) due to using tabindex to create a tab order that does not preserve meaning and operability*
+
+
+## Skip navigations
+
+> [Specification](https://www.w3.org/TR/2013/NOTE-WCAG20-TECHS-20130905/G1)
+
+A skip navigation link provides a way for keyboard and screen reader users to skip to main content of a webpage.  
+Without skip links, keyboard and screen reader users must navigate a long list of navigation links and other elements before ever arriving at the main content, This can be particularly difficult for users with some forms of motor disabilities.
```
nashcheez, line 251:
```suggestion
Without skip links, keyboard and screen reader users must navigate a long list of navigation links and other elements before ever arriving at the main content. This can be particularly difficult for users with some forms of motor disabilities.
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+**Temporarily hidden skip links**
+
+Most common type of skip links are which that is hidden until the user navigates to it with a keyboard.
```
nashcheez, line 255:
```suggestion
Most common type of skip links are the ones that are hidden until the user navigates to it with a keyboard.
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+```
+
+The `skip-link` class here is a screen reader only [visually hidden](https://webaim.org/techniques/css/invisiblecontent/) class, which is only accessible through screen readers. By using the `:focus` pesudo class it becomes visible when focused.
+
+- [ReachUI SkipNav](https://reach.tech/skip-nav/)
+- [ChakraUI SkipNav](https://chakra-ui.com/docs/components/navigation/skip-nav)
+
+## Composite widgets
+
+*What are composite widgets?*
```
nashcheez, line 288: We can skip the question here maybe? I don't think we mentioned a question in the "Skip Navigations" or the "Using tabindex" sections, plus this is self explanatory.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+A composite is a widget that may contain navigable descendants or childrens. Composite widgets should only have single navigation tab stop.
+Once the composite widget has focus, It's required to provide a separate navigation mechanism for users to navigate to elements that are descendants (generally with arrow keys).
```
nashcheez, line 291:
```suggestion
Once the composite widget has focus, it should provide for a separate navigation mechanism for users to navigate to elements that are descendants (generally with arrow keys).
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+- Pressing <kbd>Home</kbd> or <kbd>PageUp</kbd> moves focus to the first element.
+- Pressing <kbd>End</kbd> or <kbd>PageDown</kbd> moves focus to the last element.
+
+## Implementations
+
+There are various third party library implementations for roving tabindex pattern.  
+
+We tried out 3 of them to understand the pros and cons for each and decided on X.
+
+- FluentUI - [Demo](https://codesandbox.io/s/blade-comp-fluentui-focuszone-0sjyv1?file=/src/App.tsx) | [Doc](https://developer.microsoft.com/en-us/fluentui#/controls/web/focuszone) | [Pros and cons](https://codesandbox.io/s/blade-comp-fluentui-focuszone-0sjyv1?file=/src/pros-and-cons.md)
```
nashcheez, line 333: What's the status of this? Have we decided on which third party library we are proceeding with? The pros and cons of whichever one we decide should come here in the rfc.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+  - Do all keyboard interactive elements display visual keyboard focus?
+  - Is the visual keyboard focus easy to identify?
+- During the navigation, are there any instances when you become trapped in an element? ([Failure of Success Criterion](https://www.w3.org/WAI/WCAG21/Techniques/failures/F10.html))
+- Make sure hidden popuops or titles are accessible through keyboard. 
+
+# Focus management
+
+> Specification
+
+**Target:** Everyone and especially people with visual impairments or cognitive limitations, motor disabilities.
```
nashcheez, line 358:
```suggestion
**Target:** Everyone, especially people with visual impairments or cognitive limitations & motor disabilities.  
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+  - Is the visual keyboard focus easy to identify?
+- During the navigation, are there any instances when you become trapped in an element? ([Failure of Success Criterion](https://www.w3.org/WAI/WCAG21/Techniques/failures/F10.html))
+- Make sure hidden popuops or titles are accessible through keyboard. 
+
+# Focus management
+
+> Specification
+
+**Target:** Everyone and especially people with visual impairments or cognitive limitations, motor disabilities.  
+**Goal:** Setting guidelines for general focus behaviours & Providing screen reader & keyboard users a smooth experience by managing focus behavior for certain elements like Modals/Page transition etc.
```
nashcheez, line 359:
```suggestion
**Goal:** Setting guidelines for general focus behaviours & providing screen reader & keyboard users a smooth experience by managing focus behavior for certain elements like Modals/Page transition etc. 
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+- [Focus order](#keyboard-tab-order)
+- [Focus ring styling](#focus-ring-styling)
+- [Focus traps](#focus-trap)
+  - Focus restoration
+- Page transitions
+
+## Focus order
+
+Focus order or tab order we already covered in [keyboard tab order section](#keyboard-tab-order)
```
nashcheez, line 375:
```suggestion
We have already covered focus order or tab order in the [keyboard tab order section](#keyboard-tab-order)
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+## Focus ring styling
+
+The focus ring must be visible to all users,
```
nashcheez, line 379:
```suggestion
The focus ring must be visible to all users. 
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+- Provide focus styles that are highly visible.
+- Make sure that a visible element has focus at all times when using a keyboard.
+- Avoid using `*:focus { outline: none }` snippet to hide focus
+- Design focus rings such that it has proper contrast
+
+> Not providing visible focus ring is [Failure of Success Criterion 2.4.7](https://www.w3.org/WAI/WCAG21/Techniques/failures/F78)
+
+**Keyboard only focus rings**
+
+While focus indicators are neccesary,
```
nashcheez, line 393: Remove the line break.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+- Make sure that a visible element has focus at all times when using a keyboard.
+- Avoid using `*:focus { outline: none }` snippet to hide focus
+- Design focus rings such that it has proper contrast
+
+> Not providing visible focus ring is [Failure of Success Criterion 2.4.7](https://www.w3.org/WAI/WCAG21/Techniques/failures/F78)
+
+**Keyboard only focus rings**
+
+While focus indicators are neccesary,  
+Mouse clicks also changes the focus, for example clicking on a button also makes the focus ring visible.
```
nashcheez, line 394:
```suggestion
mouse clicks also change the focus, for example clicking on a button also makes the focus ring visible.  
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+**Implementations**
+
+- CSS
+- ReactAria
+
+**CSS:**  
+
+> [Demo](https://codesandbox.io/s/blade-a11y-css-focus-ring-051xff?file=/src/App.tsx)
+
+CSS provides us with a pesudo class called [`:focus-visible`](https://css-tricks.com/keyboard-only-focus-styles). Browser support for focus-visible is [good enough.](https://caniuse.com/?search=%3Afocus-visible).
```
nashcheez, line 409:
```suggestion
CSS provides us with a pesudo class called [`:focus-visible`](https://css-tricks.com/keyboard-only-focus-styles). Browser support for focus-visible is [good enough](https://caniuse.com/?search=%3Afocus-visible).  
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+CSS provides us with a pesudo class called [`:focus-visible`](https://css-tricks.com/keyboard-only-focus-styles). Browser support for focus-visible is [good enough.](https://caniuse.com/?search=%3Afocus-visible).  
+And we can also use the [official pollyfill](https://github.com/WICG/focus-visible) for older browsers.
+
+**ReactAria:**  
+
+> [Demo](https://codesandbox.io/s/blade-a11y-reactaria-focus-ring-hc3cub?file=/src/App.tsx)
+
+ReactAria provides a [FocusRing](https://react-spectrum.adobe.com/react-aria/FocusRing.html) component which solves this. 
+
+While this works great, this solution can be a bit overkill since react aria does this all with javascript and have it's own event handling system.
```
nashcheez, line 418:
```suggestion
While this works great, this solution can be a bit of an overkill since ReactAria does this all with JavaScript and has its own event handling system.  
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+And we can also use the [official pollyfill](https://github.com/WICG/focus-visible) for older browsers.
+
+**ReactAria:**  
+
+> [Demo](https://codesandbox.io/s/blade-a11y-reactaria-focus-ring-hc3cub?file=/src/App.tsx)
+
+ReactAria provides a [FocusRing](https://react-spectrum.adobe.com/react-aria/FocusRing.html) component which solves this. 
+
+While this works great, this solution can be a bit overkill since react aria does this all with javascript and have it's own event handling system.  
+I think it will be better and easier if we `use the platform™`
```
nashcheez, line 419: These sentence is a little vague and does not explain what we mean by "using the platform".

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+## Focus trap
+
+> [Specification]()
+
+Trapping focus is a behaviour we usually want when there is modality in a page. 
+
+**Why focus traps?**
+
+Focus traps are essential to communicate a modal's bounds to groups of people who are visually impaired or have mobility issues, And helps inform what is and is not modal content.
```
nashcheez, line 429:
```suggestion
Focus traps are essential to communicate a modal's bounds to people who are visually impaired or have mobility issues, and it helps inform them what is and is not contained within a modal.
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+The idea is that if for part of the site we prevent clicks, we should also prevent focus.
```
nashcheez, line 431:
```suggestion
The idea is that if for parts of the site where we prevent clicks, we should also prevent focus on the same elements.
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+
+- People with low vision who rely on screen readers need to know when the modal opens.
+- People with low vision who cannot properly see the focus ring can get lost if focus isn't contained in the modal, This can be confusing and disorienting.
```
nashcheez, line 434:
```suggestion
- People with low vision who cannot properly see the focus ring can get lost if focus isn't contained in the modal, this can be confusing and disorienting.
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+- People with mobility issues using keyboard could tab out of the modal and have a hard time getting back into the modal.
+
+**Best practices for focus traps**
+
+- Ensure users don't get permanently trapped in the focus trap, If the user can open the modal with keyboard, they should also have the ability to exit the modal with keyboard.
+- The focus should loop When reaching the last or first element inside modal.
```
nashcheez, line 440:
```suggestion
- The focus should loop when reaching the last or first element inside modal.
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+> [Demo](https://codesandbox.io/s/blade-a11y-reactaria-focustrap-6z537u?file=/src/App.tsx)
+
+ReactAria's `FocusScope` component is a fully fledged solution with other edge cases handled for us. 
+
+It also handles: 
+
+- Focus restoration to trigger
+- Auto focusing to children elements
+
+Considering the `inert` attribute is not supported in major browsers & we will have to write custom logic to handle other edge cases. It's better to use the ReactAria's component in this case.
```
nashcheez, line 474:
```suggestion
Considering the `inert` attribute is not supported in major browsers & we will have to write custom logic to handle other edge cases, it's better to use the ReactAria's component in this case. 
```

rfcs/2022-04-09-accessibility.md
anuraghazra: Will do.

rfcs/2022-04-09-accessibility.md
anuraghazra: > The pros and cons of whichever one we decide should come here in the rfc.  

yes.

We haven't decided on which library to pick up yet. Needs discussion with @kamleshchandnani @chaitanyadeorukhkar, plus the decision could be dependent on other factors like focus management & screen reader support.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+- **Cognitive:**  
+  Concentration, memory, judgment, problem solving, logic skills.
+- **Mobility:**  
+  Anything that affects movement in a body.
+- **Auditory:**  
+  Disabilities related to sound or audio.
+
+Other subcategories based on [microsoft’s inclusive design](https://www.microsoft.com/design/inclusive):  
+
+- **Permanent or disability that does not go away:**:
```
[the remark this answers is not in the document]
anuraghazra: 🙈

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,505 @@
...
+- Make text readable and understandable.
+- Make content appear and operate in predictable ways.
+- Help users avoid and correct mistakes.
+
+**Robust**
+
+- Maximize compatibility with current and future user tools.
+
+# Motivation
+- Why are we doing this?
```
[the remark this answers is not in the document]
anuraghazra: Sure

rfcs/2022-04-09-accessibility.md
anuraghazra: Sure will rephrase.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,495 @@
...
+- Hidden by default
+- Accessible to keyboard navigation
+- Visible when it is focused
+- Set focus to the main content area when activated
+
+### Implementations
+
+Skip nav implementations are fairly simple: 
+
+> [Live Demo](https://stackblitz.com/edit/react-ts-nnnymu?file=App.tsx)
```
kamleshchandnani, line 261: the demo isn't working? 🤔

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,495 @@
...
+```
+
+The `skip-link` class here is a screen reader only [visually hidden](https://webaim.org/techniques/css/invisiblecontent/) class, which is only accessible through screen readers. By using the `:focus` pesudo class it becomes visible when focused.
+
+- [ReachUI SkipNav](https://reach.tech/skip-nav/)
+- [ChakraUI SkipNav](https://chakra-ui.com/docs/components/navigation/skip-nav)
+
+## Composite widgets
+
+A composite is a widget that may contain navigable descendants or childrens. Composite widgets should only have single navigation tab stop.
```
kamleshchandnani, line 281: too much content but difficult to visualise as to what you mean here. probably a small gif or video would help understand this clearly

rfcs/2022-04-09-accessibility.md
kamleshchandnani: Can we add the pros and cons in collapsible section in this readme itself?

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,495 @@
...
+`tabindex` attribute authors can make the element focusable and appear in [sequential focus order](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation)
+
+Setting tabindex to any non-negative integer makes elements focusable, allow or prevent them from being sequentially focusable, and determine their relative ordering
+
+When set to `0`, the element becomes focusable by keyboard and via programmatic means with `focus()`. 
+When set to `-1`, the element becomes focusable programmatically, but it does not become part of the keyboard focus order.
+
+The following table describes tabindex behavior in modern browsers:
+
+| tabindex attribute |	Focusable with mouse or programmatically |	Tab navigable |
```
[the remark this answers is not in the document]
anuraghazra: clicking it will make it focusable. https://codesandbox.io/s/white-meadow-0o7chs?file=/src/App.tsx

See focus ring section for more info.

rfcs/2022-04-09-accessibility.md
anuraghazra: Yes open the window directly, stackblitz might be messing up your focus https://react-ts-nnnymu.stackblitz.io/#main-content

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,495 @@
...
+| :-- | :-- | :-- |
+| not present	| The user agent will decide | The user agent will decide |
+| Negative (tabindex="-1")	| Yes | No, can only be focused programmatically |
+| Zero (tabindex="0") | Yes | Yes, In tab order relative to element's position in document | 
+| Positive (tabindex="2") | Yes | Yes, tabindex value determines where this element is positioned in the tab order | 
+
+> Warning: Avoid using positive values for tabindex. Using positive values means authors will have to set (and maintain) tabindex values for all focusable elements on the page whenever they use one or more positive values for tabindex.
```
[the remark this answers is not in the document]
anuraghazra: We discourage using positive values because positive values can cause wrong tab order 

for example: 

```
<button tabIndex="1">one</button>
<button tabIndex="2">one</button>
<button tabIndex="3">one</button>
```

The above code might look fine but since tabindex is tracked globally these buttons will always be the first ones to get focus even if they are at the end of the page, plus someone might mess up the order and cause confusion. 

>  then we shall also discourage negative values right since that will stop tab navigation

No that depends, in certain situations we want to prevent tab navigation to reduce tab stops in the page, see composite widgets section. https://github.com/razorpay/blade/blob/anu/a11y-rfc/rfcs/2022-04-09-accessibility.md#composite-widgets

We want to reduce the number of tab stops in certain elements like select-menu/combobox/radio groups etc, And let users use Arrowkeys to move between the group. 

> I'm thinking we shall discourage using tabindex altogether

In app level consumers probably wouldn’t need to use tabindex, but in component level we need to manage that to ensure focus management is handled correctly.

rfcs/2022-04-09-accessibility.md
anuraghazra: Sure.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,495 @@
...
+> [Specification](https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation.html)
+
+**Target:** Everyone and especially people with motor disabilities who use a keyboard to navigate.  
+**Goal:** Ensure users who cannot use the mouse (blind, motor disabilities) can access the crucial parts of the app through keyboard.  
+
+For a web page to be accessible, all interactive elements must be operable via the keyboard. Not all users are able to use a mouse to navigate a web page, keyboard-only and screen reader users rely on navigating and using a web page with a physical or a virtual keyboard.
+
+### Areas to cover
+
+- [Keyboard tab order](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation-order)
```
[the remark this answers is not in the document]
anuraghazra: yeah will change those.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,495 @@
...
+  </main>
+</body>
+```
+
+The `skip-link` class here is a screen reader only [visually hidden](https://webaim.org/techniques/css/invisiblecontent/) class, which is only accessible through screen readers. By using the `:focus` pesudo class it becomes visible when focused.
+
+- [ReachUI SkipNav](https://reach.tech/skip-nav/)
+- [ChakraUI SkipNav](https://chakra-ui.com/docs/components/navigation/skip-nav)
+
+## Composite widgets
```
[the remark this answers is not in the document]
anuraghazra: Yes will add gifs in places for easier understanding.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,495 @@
...
+- Set accessibility guidelines to make sure products built with blade are accessible by people with various disabilities. 
+- Document the component development lifecycle for blade developers on building accessible components. (Similar to a checklist)
+
+By baking in accessibility at the foundational level we will ensure our products are accessible to as many people as possible, and this includes people with disabilities.
+
+# Detailed Design
+
+## Scope 
+
+- Web
```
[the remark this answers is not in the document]
anuraghazra: Yes. we can do that.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,495 @@
...
+> Check out this youtube [video about Web Accessibility Perspectives](https://www.youtube.com/watch?v=3f31oufqFSM) by W3C for more details about how accessibility can help wide range of people with or without disabilities.
+
+## Various types disabilities
+
+- **Visual:**  
+  Anything that deals with sight, eg: people who have different types of sight like color blindness.
+- **Cognitive:**  
+  Concentration, memory, judgment, problem solving, logic skills.
+- **Mobility:**  
+  Anything that affects movement in a body.
```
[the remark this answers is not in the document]
anuraghazra: Okay will rephrase this, what this means that say people who have mobility issues eg: broken/missing arm should also be able to operate our sites. 

This goes under having proper keyboard accessibility because people with mobility issues might not be as accurate with mouse.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,606 @@
...
+ReactAria's `FocusScope` component is a fully fledged solution with other edge cases handled for us.
+
+It also handles:
+
+- Focus restoration to trigger
+- Auto focusing to children elements
+
+Considering the `inert` attribute is not supported in major browsers & we will have to write custom logic to handle other edge cases, it's better to use the ReactAria's component in this case.
+
+**Native:**
```
kamleshchandnani, line 562: can you attach a demo and video for this on native?

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,606 @@
...
+
+To simulate `inert` behaviour in react-native we have platform specific accessibility props:
+
+- For android: [`importantForAccessibility`](https://reactnative.dev/docs/accessibility#importantforaccessibility-android)
+- For iOS: [`accessibilityViewIsModal`](https://reactnative.dev/docs/accessibility#accessibilityviewismodal-ios) & [`accessibilityElementsHidden`](https://reactnative.dev/docs/accessibility#accessibilityelementshidden-ios)
+
```
kamleshchandnani, line 568: once you're done with all the evaluations, let's put a table for a summary of what to use for what platform for a particular a11y thing. it'll basically a snapshot of the information you've already put above. for eg: we'll use reac-aria for focus trap on web and on native we'll use the platform specific attributes, etc.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,606 @@
...
+- Easy to use
+- Handles roving index by default
+- Can handle nested focus zones
+- Provides tabbable focus, means we can enable use of tab + arrow keys
+- Support layout grid behavior (we might not need this)
+
+**Cons:**
+
+- Not much flexibility to write our own logic, doesn’t provide any focus manager
```
kamleshchandnani, line 297: how will it be useful? maybe can you elaborate it a little in the content?

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,606 @@
...
+
+**Cons:**
+
+- Doesn’t provide roving index behavior by default
+- A lot of manual logic and code has to be written to implement a proper Roving index component which handles edge cases,
+- The implementation in the codesandbox has various bugs & edge cases, for eg: initially the elements don't have tabIndex set, these kinds of edge cases we need to solve for.
+
+</details>
+
+**Ariakit** - [Demo](https://codesandbox.io/s/blade-comp-ariakit-composite-ewc9i4?file=/src/App.tsx) | [Doc](https://reakit.io/docs/composite/)
```
kamleshchandnani, line 331: I think ariakit looks fine to me in this current context based on the flexibility and capabilities it provides. Although documentation can be quite tricky I think if a few basic/advanced interactions are sorted then we shall be ok to take this bet.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,606 @@
...
+> [Demo](https://codesandbox.io/s/blade-a11y-reactaria-focus-ring-hc3cub?file=/src/App.tsx)
+
+ReactAria provides a [FocusRing](https://react-spectrum.adobe.com/react-aria/FocusRing.html) component which solves this.
+
+While this works great, this solution can be a bit of an overkill since ReactAria does this all with JavaScript and has its own event handling system.  
+I think it will be better and easier if we stick to `:focus-visible` CSS property which is becoming standard in browser & a platform feature.
+
+**Native:**
+
+In native we don't need to handle focus-visible since the focus will only be visible when the screen reader is open.
```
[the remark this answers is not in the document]
anuraghazra: On native focus ring styling is based on the screen reader application.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,606 @@
...
+- The new version of ariakit lacks documentation
+- No separate package scope for their focus utilities, we will have to rely on treeshaking to be able to shave all other unnecessary components.
+
+</details>
+
+**Native:**
+
+For native we are scoping out keyboard accessibility, because it will only be effective in scenarios when using a hardware keyboard. We will revisit this once we have usecases for it.
+
+#### aria-activedescendant <!-- omit in toc -->
```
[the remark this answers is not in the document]
anuraghazra: It's under composite widgets, a variation of roving index pattern.

rfcs/2022-04-09-accessibility.md
anuraghazra: Yes it's pretty nice :( but the unsuitability is the issue I'm worried about, I don't think we should take this bet. We will be in a cycle of fixing breaking changes.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,606 @@
...
+| Zero (tabindex="0")      | Yes                                      | Yes, In tab order relative to element's position in document                     |
+| Positive (tabindex="2")  | Yes                                      | Yes, tabindex value determines where this element is positioned in the tab order |
+
+> Warning: Avoid using positive values for tabindex. Using positive values means authors will have to set (and maintain) tabindex values for all focusable elements on the page whenever they use one or more positive values for tabindex.
+
+> _[Failure of Success Criterion 2.4.3](https://www.w3.org/WAI/WCAG21/Techniques/failures/F44) due to using tabindex to create a tab order that does not preserve meaning and operability_
+
+**Native: Using `accessible` prop**
+
+In react-native there is no `tabindex` like behaviour for components, instead we have access to the [`accessible`](https://reactnative.dev/docs/accessibility#accessible) prop, setting `accessible` to true elements get grouped into a single selectable component.
```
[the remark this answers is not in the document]
anuraghazra: Added

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+  - Focus trap [best practices](#best-practices-for-focus-traps)
+  - Semantic HTML [best practices](#best-practices-to-using-semantic-html)
+  - Dynamic content [best practices](#best-practices-for-dynamic-contents)
+  - Design [best practices](#best-practices-when-using-color-styles-and-making-them-accessible)
+- Implement [automated testing tools](#tools--infrastructure) which discussed
+- Do manual testing for [keyboard accesibility](#keyboard-accessibility), [focus management & screen readers.](#tools--infrastructure) 
+  
+# Drawbacks/Constraints
+
+Why should we _not_ do this? Maybe try to consider the following constraints
```
chaitanyadeorukhkar, line 933:
```suggestion
```

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+**Cons:**
+
+- The biggest issue with ariakit is that it’s not stable yet :(  
+  So ariakit is the successor to reakit, Haz has been developing it for over 4 years now.
+  But as the doc says ariakit is better but still the API is not stable and can have breaking changes
+- The new version of ariakit lacks documentation
+- No separate package scope for their focus utilities, we will have to rely on treeshaking to be able to shave all other unnecessary components.
+
+</details>
+
```
divyanshu013, line 372: Which one are we going ahead with? Needs a conclusion section here. Or if we've open question on this.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+Examples of composite widgets are:
+
+- Radio Groups
+- Tabs
+- Listbox
+- Toolbars
+
+#### Roving Tabindex
+
+> [Specification](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_roving_tabindex)
```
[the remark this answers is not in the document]
anuraghazra: God damn, actually WAI-ARIA spec has been revamped with a new website, but they don't redirect to new content.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+
+One way to manage focus within `composite` widgets is roving tabindex pattern.
```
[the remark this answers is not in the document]
anuraghazra: Yes aria-activedescendant 
https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_focus_activedescendant

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+**Native:**
+
+To simulate `inert` behaviour in react-native we have platform specific accessibility props:
+
+**iOS:**
+
+iOS has better accessibility support for modals than android,
+In iOS we can set [`accessibilityViewIsModal`](https://reactnative.dev/docs/accessibility#accessibilityviewismodal-ios) which will cause voice overs to ignore any other elements outside the modal.
+
+**Android:**
```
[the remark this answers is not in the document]
anuraghazra: I'm not really sure actually, I didn't find any focus scope utility for android and even saw some component libraries like native-base & reactnative-paper. 

So what should we do in this case is for cases where we want modality and trap the focus we should use the modal from `react-native` because that is accessible out of the box.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+
+By baking in accessibility at the foundational level we will ensure our products are accessible to as many people as possible, and this includes people with disabilities.
+
+# Detailed Design
+
+## Keyboard Accessibility
+
+> [Specification](https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation.html)
+
+**Accessibility Principal:** Operable
```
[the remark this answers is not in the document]
anuraghazra: lol my english. xD

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+
+In web we can simply add `aria-live` to any element or use [roles which have implicit live region attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes)
+
+> [Demo](https://codesandbox.io/s/blade-a11y-rfc-arialive-demo-h53z61?file=/src/App.tsx)
+
+https://user-images.githubusercontent.com/35374649/170882365-47b9caa8-eff8-4572-a319-d0f63785f497.mov
+
+We can also use the [@react-aria/live-announcer](https://github.com/adobe/react-spectrum/tree/main/packages/%40react-aria/live-announcer) package which gives us programmatic way of triggering announcements with `announce()` function, by creating temporary live-regions.
+
+**Native:**
```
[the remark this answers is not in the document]
anuraghazra: Yes, every alert should this by default. 

In fact we can just add role="alert" in those scenarios.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+- Provide text alternatives for non-text content.
+- Provide captions and other alternatives for multimedia.
+- Create content that can be presented in different ways, including by assistive technologies, without losing meaning.
+- Make it easier for users to see and hear content.
+
+**Operable**
+
+- Make all functionality available from a keyboard.
+- Give users enough time to read and use content.
+- Do not use content that causes seizures or physical reactions.
```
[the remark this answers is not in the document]
anuraghazra: Actually motion is out of scope of accessibility, atleast for now.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+  - Dynamic content [best practices](#best-practices-for-dynamic-contents)
+  - Design [best practices](#best-practices-when-using-color-styles-and-making-them-accessible)
+- Implement [automated testing tools](#tools--infrastructure) which discussed
+- Do manual testing for [keyboard accesibility](#keyboard-accessibility), [focus management & screen readers.](#tools--infrastructure) 
+  
+# Drawbacks/Constraints
+
+Why should we _not_ do this? Maybe try to consider the following constraints
+
+- Implementation cost, both in terms of code size and complexity.
```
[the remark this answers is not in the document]
anuraghazra: This is just the default template RFC text, I didn't added these. We can actually talk about bundle size but need to decide which library to use first.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+
+- Ensuring keyboard accessibility helps people with temporary/permanent mobility but also improves the experience for normal users by providing a better user experience for components like Select or Modals.
+- Large buttons and links helps people with reduced dexterity but also improve the user experience of mobile devices.
+- Providing captions for a video can benefit users who watch videos in loud environments or in a library.
+- Sufficient color contrast is good for different lighting conditions and people with visual impairments.
+- Older people with changing abilities due to aging can benefit from accessible experiences.
+
+> Check out this youtube [video about Web Accessibility Perspectives](https://www.youtube.com/watch?v=3f31oufqFSM) by W3C for more details about how accessibility can help wide range of people with or without disabilities.
+
+## Various types of disabilities
```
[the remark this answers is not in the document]
anuraghazra: Not really sure how would I format it in table. 

Like this whole section? 

```

- **Visual:**  
  Anything that deals with sight, eg: people who have different types of sight like color blindness.
- **Cognitive:**  
  Difficulties with concentration, memory, judgment, problem solving, eg: older people with changing abilities.
- **Mobility:**  
  Anything that affects physical movement, eg: broken/missing arm.
- **Auditory:**  
  Disabilities related to sound or audio, eg: someone who is deaf.

Other subcategories based on [Microsoft’s inclusive design](https://www.microsoft.com/design/inclusive):

- **Permanent, disability that does not go away:**
  - Example: Someone who is deaf
- **Temporary, disability that will go away in time:**
  - Example: Someone who has an ear infection
- **Situational, Context based disabilities:**
  - Example: Someone who is at a loud sporting event
  - Example: Using the web app in bright sunlight
  - Example: Using your phone one handedly in a crowded metro station
```

What would be the headers and such?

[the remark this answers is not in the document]
divyanshu013: Type of disability, description, example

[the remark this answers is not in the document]
anuraghazra: That looks great thanks!

rfcs/2022-04-09-accessibility.md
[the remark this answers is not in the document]
anuraghazra: Don't need this section will update if necessary with bundlesize and all that info.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+
+## The current state
+
+In the current state, blade components are not very accessible out of the box, and neither are our existing frontend products.
+
+Here’s the screen reader accessibility of Razorpay mobile app
+
+- Missing `accessibilityLabel` in input box.
+- Missing `accessibilityRole=button` in buttons
+- Missing `accessibilityRole=link` in links
```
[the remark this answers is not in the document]
anuraghazra: 🤔 not sure if this would be the right place tho. What do you recommend?

rfcs/2022-04-09-accessibility.md
anuraghazra: done.

rfcs/2022-04-09-accessibility.md
anuraghazra: Attached video for most of the native demos, but this one is tricky. see
https://github.com/razorpay/blade/pull/463#discussion_r886387385

rfcs/2022-04-09-accessibility.md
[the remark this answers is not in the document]
anuraghazra: Fixed added a header for both.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,984 @@
...
+
+While building each blade component we should read upon these patterns and ensure they behave as expected.
+
+## Design Accessibility
+
+Designing with accessibility in mind is an important part of the process, ensuring that our designs have proper color contrast, visual hierearchy & legible text is important.
+
+#### Best practices when using color styles and making them accessible: <!-- omit in toc -->
+
+- Use colors which are dark enough and atleast pass the AA standards.
```
saurav12, line 858: @anuraghazra We can also attach this link here - [https://usecontrast.com/guide](https://usecontrast.com/guide)

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,984 @@
...
+- Don’t use color alone to make critical information understandable
+- Don’t use color as the only visual means of conveying information
+- Design [legible & usable focus states](#focus-ring-styling)
+
+<img src="./images/accessibility/typography-contrast-ratio.png" alt="Typography color contrast ratio table" width="70%" />
+
+#### Best practices when using typography:  <!-- omit in toc -->
+
+- Always use the primary font (Lato). It’ very readable if used in a hiererchy.
```
saurav12, line 867: @anuraghazra Don't call out `Lato` here. It might change going forward.

rfcs/2022-04-09-accessibility.md
anuraghazra: great resource! Thanks.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,984 @@
...
+- Make sure that the base font-size is used in the majority of the page. We’ve kept the base-font-size to 14px which works well with our use case.
+- Use larger font sizes on titles, and make sure to establish a hiererchy if there are more than 2 sections on a page so that user can easily scan the page.
+- Don’t use underlines in any state if they are not links.
+- Limit the font-style variations. Try to keep things consistent. Do not overuse Bolds, All caps, etc.
+- Use markup wherever required. Do not use it for anything except code snippets.
+- Don’t make the content (block of text) width too long. Wrap,
+  - Titles: at max. of 70% of the page width.
+  - Subtitles & Paras: at max. of 50%-60% of the page width.
+- Make sure the text doesn’t overlap in between two different sections.
+
```
saurav12, line 877: @anuraghazra We can also add:
- For legibility, always use the same width numbers. As we deal in the financial ecosystem, it makes it more readable when we read numbers in combination. We should always use a monospaced number font. It should not necessarily be a coding font, it can be a generic font as well.

rfcs/2022-04-09-accessibility.md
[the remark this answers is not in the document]
divyanshu013: Optional, I though it would be nice to know for readers to know. Perhaps one liner. Feel free to skip if you feel otherwise.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,1001 @@
...
+- Does the tab order match the logical reading order of the page? ([Failure of Success Criterion](https://www.w3.org/WAI/WCAG21/Techniques/failures/F44.html))
+- Focus visible ([Failure Of Success Criterion](https://www.w3.org/WAI/WCAG21/Techniques/failures/F78.html))
+  - Do all keyboard interactive elements display visual keyboard focus?
+  - Is the visual keyboard focus easy to identify?
+- During the navigation, are there any instances when you become trapped in an element? ([Failure of Success Criterion](https://www.w3.org/WAI/WCAG21/Techniques/failures/F10.html))
+- Make sure hidden popuops or titles are accessible through keyboard.
+
+## Focus management
+
+> Specification
```
kamleshchandnani, line 462: link?

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,1001 @@
...
+
+**Web:**
+
+In web we can simply add `aria-live` to any element or use [roles which have implicit live region attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes)
+
+> [Demo](https://codesandbox.io/s/blade-a11y-rfc-arialive-demo-h53z61?file=/src/App.tsx)
+
+https://user-images.githubusercontent.com/35374649/170882365-47b9caa8-eff8-4572-a319-d0f63785f497.mov
+
+We can also use the [@react-aria/live-announcer](https://github.com/adobe/react-spectrum/tree/main/packages/%40react-aria/live-announcer) package which gives us programmatic way of triggering announcements with `announce()` function, by creating temporary live-regions.
```
kamleshchandnani, line 818: how do we decide whether to use `aria-live` attribute or this utility?

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,1001 @@
...
+```ts
+React.useEffect(() => {
+  AccessibilityInfo.announceForAccessibility('Some content to be announced');
+}, []);
+```
+
+https://user-images.githubusercontent.com/35374649/170883127-b81bc324-3d5c-4b9d-905f-14733fcc821d.mp4
+
+
+### ARIA Patterns
```
kamleshchandnani, line 835: this section doesn't seems like fitting under Screen readers, feels like this is more generic?

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,1001 @@
...
+- Follow [WAI-ARIA Authoring Practices Guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/) for component specific accessibility testing
+- Do [manual keyboard testing](#manual-testing)
+- Do manual [screen reader testing](https://dequeuniversity.com/screenreaders/voiceover-ios-shortcuts)
+- Follow a checklist for general page wide accessibility testing
+  - [18F checklist](https://accessibility.18f.gov/checklist/)
+  - [a11y project checklist](https://www.a11yproject.com/checklist/)
+  - [gov.uk](https://www.gov.uk/government/publications/doing-a-basic-accessibility-check-if-you-cant-do-a-detailed-one/doing-a-basic-accessibility-check-if-you-cant-do-a-detailed-one) basic accessibility tests
+  - or use WCAG [evaluation tool](https://www.w3.org/WAI/eval/report-tool/)
+
+# Actionable Items
```
kamleshchandnani, line 930: can you make them checklist so we can tick them off when completed

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,1001 @@
...
+
+**Manual Testing:**
+
+Automated accessibility tools only [pick up around 40% of errors automatically](https://alphagov.github.io/accessibility-tool-audit/) at best, to ensure blade components are accessible it's essential to do manual testing
+
+- Follow preliminary guide for ["A First Review of Web Accessibility"](https://www.w3.org/WAI/test-evaluate/preliminary/) by w3c
+- Follow [WAI-ARIA Authoring Practices Guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/) for component specific accessibility testing
+- Do [manual keyboard testing](#manual-testing)
+- Do manual [screen reader testing](https://dequeuniversity.com/screenreaders/voiceover-ios-shortcuts)
+- Follow a checklist for general page wide accessibility testing
```
kamleshchandnani, line 924: can we create a simple checklist optimized for our use cases and put it in some central place? every time we build a new component it's difficult for devs to go through so many things. Basically a checklist of items that all the devs shall use when building a component.

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,1001 @@
...
+Ideally only two people needs to be educated: 
+
+- Blade design system designers
+- Blade design system developers
+
+This document provides enough insights, guides & best practices to ensure that designers & developers are well educated about the ins and out of building accessible components, although other teams also need to be proactive about certain things like providing proper accessibility label, alt text etc.
+
+# Open Questions
+
+- Which third party library for roving tabindex pattern should we choose?
```
kamleshchandnani, line 973: i think this is already answered. let's remove it?

rfcs/2022-04-09-accessibility.md
anuraghazra: So the difference this announce() utility will be handy because this utility will create hidden DOM nodes with VisuallyHidden programatically thus it won't be visible to users. But with aria-live we have to manually make sure it's hidden. 

It's based on our usecases.

rfcs/2022-04-09-accessibility.md
anuraghazra: This is kind of fits, I don't know where to put it. It will just hang out in the middle of the document otherwise.

rfcs/2022-04-09-accessibility.md
anuraghazra: Yes that makes sense, I was in fact planning to write a doc for that. But not entirely sure if Accessibility RFC is the right place to do this. Might be better if we create something like `component-development-lifecycle.md` file to document our process.

rfcs/2022-04-09-accessibility.md
kamleshchandnani: it can become it's own topic maybe while we are talking about ARIA things since this is setting the context and basic information.

rfcs/2022-04-09-accessibility.md
kamleshchandnani: makes sense. but can't button loading have 

`<button aria-live={isLoading? 'assertive':undefined}>Submit Form</button>

rfcs/2022-04-09-accessibility.md
```diff
@@ -0,0 +1,996 @@
...
+- [x] Implement `<VisuallyHidden />` component for [hidden content](#hidden-content)
+- [ ] Ensure each component adhers to the [WAI-ARIA Authoring Pratices Guide](https://www.w3.org/WAI/ARIA/apg/patterns/) 
+- [ ] Ensure general accessibility guidelines are followed while building blade components, we discussed best practices for each category: 
+  - Keyboard accessibility [best practices](#best-practices-for-logical-tab-order)
+  - Focus ring [best practices](#best-practices-for-focus-styling)
+  - Focus trap [best practices](#best-practices-for-focus-traps)
+  - Semantic HTML [best practices](#best-practices-to-using-semantic-html)
+  - Dynamic content [best practices](#best-practices-for-dynamic-contents)
+  - Design [best practices](#best-practices-when-using-color-styles-and-making-them-accessible)
+- [x] Implement [automated testing tools](#tools--infrastructure) which discussed
```
kamleshchandnani, line 946:
```suggestion
- [x] Implement [automated testing tools](#tools--infrastructure) which we discussed
```

kamleshchandnani:
@anuraghazra apart from this everything else LGTM

https://github.com/razorpay/blade/pull/463#discussion_r896596888

rfcs/2022-04-09-accessibility.md
anuraghazra: It will work, but it's not the correct way to handle live announcements, live announcements should always have their own independent container... putting aria-live directly inside the button you are coupling the announcement to the button's text tightly and there is no way to have different announcement and different text for button. 

You'll need to do this
```ts
<button aria-live={isLoading? 'assertive':undefined}>{isLoading ? "Loading" : "Submit button"}</button>
```

With this example how would you let users know that button has stopped loading? You cannot announce "Finished loading" since if you do the button's text will also say "Finished loading"

rfcs/2022-04-09-accessibility.md
anuraghazra: Decided to decouple this as a different task to not block a11y RFC.

rfcs/2022-04-09-accessibility.md
anuraghazra: Made it it's own section instead of subsection of screen readers. makes more sense.

kamleshchandnani (approved):

divyanshu013:
🥳

PR #464, opened 2022-05-09
fix(blade): add `Heading` component

kamleshchandnani:
This PR adds `Heading` component

The API for `Heading` component can be found under [Typography/Text/_decisions](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Typography/_decisions/decisions.md)


.eslintrc.js
```diff
@@ -95,6 +95,7 @@ module.exports = {
         '@typescript-eslint/no-unnecessary-condition': 'off',
         '@typescript-eslint/consistent-type-imports': ['error', { prefer: 'type-imports' }],
         '@typescript-eslint/unbound-method': 'off',
+        '@typescript-eslint/sort-type-union-intersection-members': 'off',
```
kamleshchandnani, line 98: this auto-sorting of union keys was annoying a lot.

For eg: If I type following thing `const HeadingVariants = 'small' | 'medium' | 'large' | 'subheading'` it will sort the union keys in alphabetical order but I wanted it to be the way I declare because it's easier to read that way and most of the cases devs do this so turning off this rule

anuraghazra: Yeah mostly find this rule useless.

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -0,0 +1,85 @@
+import type { ReactElement } from 'react';
+import type { TextTypes } from '../../../tokens/theme/theme';
+import getPlatformType from '../../../utils/getPlatformType';
+import type { Theme } from '../../BladeProvider';
+import BaseText from '../BaseText';
+import type { BaseTextProps } from '../BaseText/BaseText';
+
+type HeadingCommonProps = {
+  type: TextTypes;
+  children?: React.ReactNode;
```
anuraghazra, line 10: Should we make the childrens required? 

Whats the usecase when any text/heading component will be used without children or ie: be empty?

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -0,0 +1,85 @@
...
+    fontWeight: weight,
+    fontStyle: 'normal',
+    lineHeight: 'm',
+    fontFamily: 'text',
+  };
+
+  if (variant === 'small') {
+    props.fontSize = 200;
+    props.lineHeight = 'xl';
+    props.as = getPlatformType() === 'browser' || getPlatformType() === 'node' ? 'h6' : undefined;
```
anuraghazra, line 54: Extract this to a local var, we repeating it a lot. 

```ts
const isBrowserLike = getPlatformType() === 'browser' || getPlatformType() === 'node'; 
```

anuraghazra:
Screenshot 2022-05-10 at 2 39 28 PM

Are you also getting these two unwanted intellisense? Seems like styled components & emotion is injecting them through namespace

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -0,0 +1,85 @@
...
+  variant: infer Variant;
+}
+  ? Variant extends 'small' | 'medium' | 'large'
+    ? HeadingNormalVariant
+    : Variant extends 'subheading'
+    ? HeadingSubHeadingVariant
+    : T
+  : T;
+
+const getProps = <T extends { variant: 'small' | 'medium' | 'large' | 'subheading' }>({
```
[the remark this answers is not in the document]
anuraghazra: type HeadingVariants = 'small' | 'medium' | 'large' | 'subheading'; 

```
<T extends { variant: HeadingVariants }>
```

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -0,0 +1,85 @@
...
```
anuraghazra, line 37: Question: 

Why are we extracting out this `getProps` function instead of inlining it directly on the component? Feels like if we are not using it else where better to keep it inline.

packages/blade/src/components/Typography/Heading/Heading.tsx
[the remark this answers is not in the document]
nashcheez: Since both `subheading` and `small` are `h6`, `subheading` being the more diminutive element with the lesser fontSize and lineHeight, I feel it would make more sense to start with `subHeading` and go in an ascending order of hierarchy and end with `large`.

`<T extends { variant: 'subheading' | 'small' | 'medium' | 'large' }>`

packages/blade/src/components/Typography/Heading/Heading.tsx
nashcheez: Or make this into another util/helper inside `getPlatformType`? Seen this being used in [Title.tsx](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Typography/Title/Title.tsx#L26) and imagine it will be used extensively in other components too.

packages/blade/src/components/Typography/Heading/__tests__/Heading.web.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <Heading type="subdued" variant="large" weight="regular">
+        {displayText}
+      </Heading>,
+    );
+    expect(getByRole('heading', { level: 4 })).toBeInTheDocument();
+    expect(getByText('Get Started With Payment Gateway')).toBeInTheDocument();
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should throw error when variant is "subheading" but weight "regular" is passed', () => {
```
nashcheez, line 45: nit: Maybe we should also have a test for `variant="subheading"` and `weight="bold"` before this? Same for `native.test.tsx`.

packages/blade/src/components/Typography/Heading/Heading.tsx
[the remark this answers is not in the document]
kamleshchandnani: arre sorting is fine, I'm more concerned about the variant `subheading` itself. like `small`, `medium`, `large` and `subheading` aren't going well together

packages/blade/src/components/Typography/Heading/Heading.tsx
kamleshchandnani: Kept it intentionally, need to fix this for the entire Typography suite.

packages/blade/src/components/Typography/Heading/Heading.tsx
kamleshchandnani: just kept it as a pattern for readability, at few places we do some modifications inside this function. Since this function is responsible as an abstraction for our shared styles and conditional styles, felt it would be better to keep it out from the Component function

packages/blade/src/components/Typography/Heading/Heading.tsx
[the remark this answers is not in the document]
anuraghazra: Oh I see, you are saying that `subheading` doesn't sound like a variant? 

But to be honest, subheading is the only one which sounds like a variant 😆 `small` `medium` `large` is more like sizes for the title. 

Although agree that the naming aren't going together intuitively.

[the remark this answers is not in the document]
nashcheez: I know chakra does it with `xs` to `xl` but they call their prop `size` instead of `variant`.
https://chakra-ui.com/docs/components/typography/heading

packages/blade/src/components/Typography/Heading/Heading.tsx
kamleshchandnani: Addressed in https://github.com/razorpay/blade/pull/467

packages/blade/src/components/Typography/Heading/__tests__/Heading.native.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const { toJSON, getByText } = renderWithTheme(
+      <Heading type="subdued" variant="large" weight="regular">
+        {displayText}
+      </Heading>,
+    );
+    expect(getByText('Get Started With Payment Gateway')).toBeTruthy();
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render Heading with variant "large"', () => {
```
chaitanyadeorukhkar, line 42: Move this case before the `should render Heading with variant "subheading" and weight "bold"` so that we have small, medium, large tests in sequence?

packages/blade/src/components/Typography/Heading/__tests__/Heading.native.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+      </Heading>,
+    );
+    expect(getByText('Get Started With Payment Gateway')).toBeTruthy();
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render Heading with variant "medium"', () => {
+    const displayText = 'Get Started With Payment Gateway';
+    const { toJSON, getByText } = renderWithTheme(
+      <Heading type="muted" variant="medium" weight="regular">
```
chaitanyadeorukhkar, line 23: Assuming all other tests for `type` prop is not added because we're only passing it down & BaseText takes care of it and it is tested there?

packages/blade/src/components/Typography/Heading/__tests__/Heading.web.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+describe('<Heading />', () => {
+  it('should render Heading with variant "small"', () => {
+    const displayText = 'Get Started With Payment Gateway';
+    const { container, getByRole, getByText } = renderWithTheme(
+      <Heading type="normal" variant="small" weight="regular">
+        {displayText}
+      </Heading>,
+    );
+    expect(getByRole('heading', { level: 6 })).toBeInTheDocument();
```
chaitanyadeorukhkar, line 16: Nice!

packages/blade/src/components/Typography/Heading/__tests__/Heading.web.test.tsx
chaitanyadeorukhkar: What happens if I don't pass `weight="bold"`? What would be the default weight for a `subheading` variant? I can't see any weight being assigned when variant is subheading in the code. 

`h1-h6` tags are bold by default but what about for native? Won't we need to explicitly make it bold by default?

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -0,0 +1,87 @@
...
+type HeadingSubHeadingVariant = HeadingCommonProps & {
+  variant: 'subheading';
+  weight: keyof Pick<Theme['typography']['fonts']['weight'], 'bold'>;
+};
+
+/**
+ * Conditionally changing props based on variant.
+ * Overloads or union gives wrong intellisense.
+ */
+export type HeadingProps<T> = T extends {
```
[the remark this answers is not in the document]
kamleshchandnani: didn't get you? can't see `fontWeight` in the sense?

packages/blade/src/components/Typography/Heading/__tests__/Heading.native.test.tsx
kamleshchandnani: yes

packages/blade/src/components/Typography/Heading/__tests__/Heading.native.test.tsx
kamleshchandnani: copy paste 🤦

packages/blade/src/components/Typography/Heading/Heading.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+  Subtitle,
+  Primary,
+  Props,
+  Stories,
+  PRIMARY_STORY,
+} from '@storybook/addon-docs';
+import type { HeadingProps } from './';
+import HeadingComponent from './';
+
+const TitleStoryMeta: Meta<
```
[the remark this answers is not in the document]
kamleshchandnani: copy paste 😄

packages/blade/src/components/Typography/Heading/Heading.tsx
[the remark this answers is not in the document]
kamleshchandnani: added in b209162117341eb6837234bfe411256e515f692f

chaitanyadeorukhkar (approved):

nashcheez (approved):

PR #467, opened 2022-05-11
fix(blade): make children prop across Typography suite better

kamleshchandnani:
This PR streamlines the type of children that each component of the `Typography` family accepts. Here are the decisions that were taken:

### BaseText
`children: React.ReactNode`
* Children are mandatory
* We could accept any valid ReactNode which is one of `ReactElement | ReactText | ReactFragment | ReactPortal | boolean | null | undefined`.
* The reason to accept `ReactNode` for `BaseText` is since this is a raw structure for the consumers they can satisfy their use case of building anything custom for rare use cases. For example, composing BaseText components by nesting them to make one word bold in a sentence. 
```
<BaseText>
	<BaseText weight="bold">Current <BaseText/>
	  Accounts supercharged by RazorpayX come with integrated tools and integrations that make financial management and accounting simple
</BaseText>
```
* This gives flexibility and also makes sure people are still under the boundaries of DS

### Text
`children: React.ReactNode`
* Children are mandatory
* We could accept any valid ReactNode which is one of `ReactElement | ReactText | ReactFragment | ReactPortal | boolean | null | undefined`.
* The reason to accept `ReactNode` and not `string` for `Text` is since it's a body text which will mostly be used in paragraphs and the main content on the page so it's better to have the flexibility for consumers because usually in the body text there are use cases to add some code snippets inline or highlight a word within the paragraph without explicitly building the `Text` component from scratch using `BaseText`.

### Title / Heading
`children: string`
* Children are mandatory
* We would accept only String for Title/Heading components since there's no use case where people would want to have special formatting in Title or Heading.
* There are use cases on our landing pages where we highlight or animate certain parts of the word or even add a logo inside the title. For those use cases, I guess consumers have to use two components and place them side by side regardless of using some layout utilities(of course since heading and title are block elements).

### DIFF
--- a/packages/blade/src/components/Typography/BaseText/BaseText.stories.tsx
+++ b/packages/blade/src/components/Typography/BaseText/BaseText.stories.tsx
@@ -9,14 +9,16 @@ export default {
   args: {
     color: 'surface.text.normal.lowContrast',
     fontFamily: 'text',
-    fontSize: 700,
+    fontSize: 200,
     fontWeight: 'regular',
     fontStyle: 'normal',
     textAlign: 'left',
     textDecorationLine: 'none',
-    lineHeight: '5xl',
+    lineHeight: 'l',
     name: 'Storybook',
-    children: 'Base Text',
+    as: 'span',
+    children:
+      'Current Accounts supercharged by RazorpayX come with integrated tools and integrations that make financial management and accounting simple',
   },
   parameters: {
     docs: {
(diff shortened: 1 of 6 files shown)


anuraghazra:
So we are only going to accept string then? 



https://github.com/razorpay/blade/pull/464#discussion_r869881146

anuraghazra:
So we are only going to accept string then? 



https://github.com/razorpay/blade/pull/464#discussion_r869881146

kamleshchandnani:
> So we are only going to accept string then?
> 
> [#464 (comment)](https://github.com/razorpay/blade/pull/464#discussion_r869881146)

@anuraghazra it was in draft mode because I had to update the description with the decisions 😆. Anyways, I've updated the description

chaitanyadeorukhkar (dismissed):

chaitanyadeorukhkar:
@kamleshchandnani can you just copy the PR Description and add it to `decisions.md`?

nashcheez (approved):

PR #469, opened 2022-05-15
Fix/text input internal state maintained removed

AditiTiwariRazorpay:
Following are read directly from props
1. Input value
2. onChange


packages/blade-old/src/atoms/TextInput/TextInput.native.js
```diff
@@ -395,7 +375,7 @@ const TextInput = React.forwardRef(
                                   placeholderTextColor={placeholderTextColor}
                                   onFocus={onTextInputFocus}
                                   onBlur={onTextInputBlur}
-                                  onChangeText={onChangeText}
+                                  onChangeText={onChange}
```
pritishvaidya, line 378: `onChange` and `onChangeText` are two different functions , one contains nativeEvent properties and shouldn't be mergd together even in blade

packages/blade-old/src/atoms/TextInput/TextInput.native.js
```diff
@@ -405,7 +385,7 @@ const TextInput = React.forwardRef(
                                   hasLeftIcon={hasLeftIcon}
                                   maxLength={maxLength}
                                   onLayout={onTextInputLayout}
-                                  value={input}
+                                  value={value}
```
pritishvaidya, line 388: _TextInput is a controlled component, which means the native value will be forced to match this value prop if provided_
This is a breaking change and value would need to be provided for every TextInput in the app.

pritishvaidya (changes requested):

packages/blade-old/src/atoms/TextInput/TextInput.native.js
AditiTiwariRazorpay: onchange we are getting here is a prop of blade component which works as onChangeText. (this was used in same way earlier as well).

packages/blade-old/src/atoms/TextInput/TextInput.native.js
AditiTiwariRazorpay: Already did a sanity for all the input once merge will give QA for sanity as well.

In my knowledge in design system we should not control value internally. This should be upto the client. Please do correct me if I am wrong

kamleshchandnani: @AditiTiwariRazorpay we need to cross-check with PG mobile devs too, this is a breaking change for them too and we shall provide a migration path for existing consumers before merging this PR

@shubhnik @shridhar52war

kamleshchandnani:
@AditiTiwariRazorpay can you detail out what was the reason of this change in your PR description?

kamleshchandnani:
@AditiTiwariRazorpay @sumit-gupta91 do we still need this or we shall close it and revisit later?

kamleshchandnani:
Closing this based on the discussions on Slack since it has a lot of breaking changes. Kindly, re-open this once we have a full-fledged solution along with the migration path.

PR #470, opened 2022-05-16
feat: add button decisions

chaitanyadeorukhkar:


packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+- What about small/large screen?
+- Component-level Actions look mostly like primary except Focus, what do we do?
+- Link & Plain Buttons are the same?
+- Should Plain/Link behave as `<a>` tag?
+- Should we expose a `type` prop for `button`, `reset`, `submit`?
+- Do we need `leadingIcon`/`leftIcon` `trailingIcon`/`rightIcon` props?
+- Should we call it onClick or onPress?
+- Scope of A11y?
+  - [WAI-ARI Button](https://www.w3.org/TR/wai-aria-practices-1.2/#button)
+  - [WAI-ARI Link](https://www.w3.org/TR/wai-aria-practices-1.2/#link)
```
anuraghazra, line 67: - Accessible label for icon only buttons. 

<Button accessibilityLabel="Search articles" icon={Search} />

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+   -  **`size`**: `large` | `medium` | `small` | `xsmall`
+      - **default**: `medium`
+   -  **`children`**: `string`
+   -  **`icon`**: `IconComponent`
+      -  **default**: `undefined`
+   -  **`iconPosition`**: `left` | `right`
+      -  **default**: `right`
+   -  **`isDisabled`**: `true` | `false`
+      -  **default**: `false`
+      -  Should we call this `disabled` instead?
```
kamleshchandnani, line 32: My vote for `is*` convention for both `isDisabled` and `isFullWidth`

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Design
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=5200%3A0) to all variants of the Button component
+
+## Internal BaseButton Component
+-  Internal component that exposes certain extra props to enable creation of Component Level Action Buttons like `positive`, `negative`, `neutral`
+-  **BaseButton API**:
+   -  **`variant`**: `primary` | `secondary` | `tertiary` | `link`
+      -  **default**: `primary`
+      -  Link will look like an inline Link component but will have the role of a `button`
```
saurabhdaware, line 16: > component but will have the role of a `button`

Do we mean it will be `<a>` but with role="button"? or will it be rendered as `<button />`?

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Component Breakdown
+
+<img src="./component-breakdown.png" width="300px"/>
+
+## Open Questions
+- ~Should we have `variant+contrast` props or `variant+action+contrast` props?~ We will go ahead with `variant+intent+contrast` props for now.
+- ~What about small/large screen?~ This would be handled internally
+- ~Component-level Actions look mostly like primary except Focus, what do we do?~ This was a Figma issue, it is intended to look exactly like the `primary` variant
+- ~Should Plain/Link behave as `<a>` tag?~ We will have a `Link` variant but that would also be a `button`. We would have another `Link` component that will be an `<a>` tag. We do this to maintain the correct roles for button & link components.
```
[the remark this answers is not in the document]
saurabhdaware: Oh so `<Link />` component will always be `<a />` and `<Button />` component will always be `<button />`. Is that correct?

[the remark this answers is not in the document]
anuraghazra: Yes, Link and Button will be two separate components. And we won't be exposing any as prop to consumers.

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+      -  Link will look like an inline Link component but will have the role of a `button`
+   -  **`intent`**: `positive` | `negative` | `notice` | `info` | `neutral`
+      -  **default**: `undefined`
+      -  We will use this to set the `intent` for internal Blade components that have positive, negative, notice, info, or neutral buttons like an `Alert` component
+   -  **`contrast`**: `high` | `low`
+      -  **default**: `low`
+      -  We need this for internal Blade components since a `positive` intent could have a `high` as well as a `low` contrast button
+   -  **`size`**: `large` | `medium` | `small` | `xsmall`
+      - **default**: `medium`
+   -  **`children`**: `string`
```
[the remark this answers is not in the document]
anuraghazra: I think we will be restricting it to string, because otherwise anyone can put any Icon or other components as prefix/suffix. 

// cc @chaitanyadeorukhkar

packages/blade/src/components/Button/_decisions/decisions.md
anuraghazra: It will always render as button.

packages/blade/src/components/Button/_decisions/decisions.md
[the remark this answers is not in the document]
saurabhdaware: Yup right. Makes sense. Can't think of any usecase where a nested element will be required inside a button 🤔

packages/blade/src/components/Button/_decisions/decisions.md
saurabhdaware: aha nice!

saurabhdaware:
```suggestion
      -  Link will look like an inline Link component but will be rendered as `button` element
```

can we rephrase this? got a bit confused because of `role of a button` line

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+   -  **`onClick`**: `Function`
+      -  **default**: `undefined`
+   -  **`type`**: `button` | `submit` | `reset`
+      -  **default**: `button`
+      -  This prop will be ignored for react-native
+
+## Button Component
+- This will be the Button component that is exposed to our consumers and will be a wrapper on top of `BaseButton` component
+-  **Button API**:
+   -  **`variant`**: `primary` | `secondary` | `tertiary`
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah, missed it here. Added it now

packages/blade/src/components/Button/_decisions/decisions.md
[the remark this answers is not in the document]
sumitd94: If we are accepting `children` as string, that means it will basically be a text which we want to display on the button.

Rather than exposing `children`, why not expose `label` and `sub label`? What do you think?

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+   -  **`icon`**: `IconComponent`
+      -  **default**: `undefined`
+   -  **`iconPosition`**: `left` | `right`
+      -  **default**: `right`
+   -  **`isDisabled`**: `true` | `false`
+      -  **default**: `false`
+      -  Should we call this `disabled` instead?
+   -  **`isFullWidth`**: `true` | `false`
+      -  **default**: `false`
+      -  Should we call this `fullWidth` instead?
```
sumitd94, line 35: If the `isFullWidth` is false, then how are we going to calculate the width of the button? Or will it just take the width depending on the content (icon, label, etc)?

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+   -  **`intent`**: `positive` | `negative` | `notice` | `info` | `neutral`
+      -  **default**: `undefined`
+      -  We will use this to set the `intent` for internal Blade components that have positive, negative, notice, info, or neutral buttons like an `Alert` component
+   -  **`contrast`**: `high` | `low`
+      -  **default**: `low`
+      -  We need this for internal Blade components since a `positive` intent could have a `high` as well as a `low` contrast button
+   -  **`size`**: `large` | `medium` | `small` | `xsmall`
+      - **default**: `medium`
+   -  **`children`**: `string`
+   -  **`icon`**: `IconComponent`
```
sumitd94, line 26: is this going to be a component or just the name of the Icon?
I think the component should be used inside the `BaseButton` and let's only expose the name of the icon that needs to be displayed .. what do you think?

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=5200%3A0) to all variants of the Button component
+
+## Internal BaseButton Component
+-  Internal component that exposes certain extra props to enable creation of Component Level Action Buttons like `positive`, `negative`, `neutral`
+-  **BaseButton API**:
+   -  **`variant`**: `primary` | `secondary` | `tertiary` | `link`
+      -  **default**: `primary`
+      -  Link will look like an inline Link component but will be rendered as `button` element
+   -  **`intent`**: `positive` | `negative` | `notice` | `info` | `neutral`
+      -  **default**: `undefined`
```
blenderskool, line 18: I feel the default value of a prop should only be one of the values allowed in the prop. Because it is unclear which `intent` would the Button be by default.

packages/blade/src/components/Button/_decisions/decisions.md
anuraghazra: isFullWidth will take up the parent container's full width.

packages/blade/src/components/Button/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: > Rather than exposing children, why not expose label and sub label? What do you think?

Yes, But at the end of the day it's a matter of API choices, both will work as long as we are consistent with the API.  

With children it's more intuitive and easy to teach people. Plus one less prop to worry about thus lesser API surface. (children can also be used as a prop but you get the point)

packages/blade/src/components/Button/_decisions/decisions.md
anuraghazra: It will be the React component element itself. 

Will probably look something like this

```ts
<Button icon={SearchIcon} />
```

anuraghazra: > let's only expose the name, typography, etc of the icon that needs to be displayed .. what do you think?

So exposing only the name as issues. [Icons won't be treeshakable](https://github.com/razorpay/blade/issues/363)

packages/blade/src/components/Button/_decisions/decisions.md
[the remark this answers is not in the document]
aditya-razorpay: I think there are genuine cases where we could need nested markup inside Button. Like suppose we need to show a pill inside button as shown in GitHub.

If we're saying that we'll have a rigid api and all these will be allowed only when design system supports a custom prop then ok.

image

packages/blade/src/components/Button/_decisions/decisions.md
aditya-razorpay: this would mean if consumer won't be able to pass any custom prop to the icon. Is that ok?

packages/blade/src/components/Button/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: > If we're saying that we'll have a rigid api and all these will be allowed only when design system supports a custom prop then ok.

Yes, GitHub might have the use case, but it's depends on if our products have the use case, if it does we can look into making it more flexible. 

Saurav from the design side has audited components to collect most of the usecases.

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+## Internal BaseButton Component
+-  Internal component that exposes certain extra props to enable creation of Component Level Action Buttons like `positive`, `negative`, `neutral`
+-  **BaseButton API**:
+   -  **`variant`**: `primary` | `secondary` | `tertiary` | `link`
+      -  **default**: `primary`
+      -  Link will look like an inline Link component but will have the role of a `button`
+   -  **`intent`**: `positive` | `negative` | `notice` | `info` | `neutral`
+      -  **default**: `undefined`
+      -  We will use this to set the `intent` for internal Blade components that have positive, negative, notice, info, or neutral buttons like an `Alert` component
+   -  **`contrast`**: `high` | `low`
```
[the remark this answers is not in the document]
divyanshu013: Seems like we're not exposing this as a prop on the main Button component (for consumers). How would they configure it?

divyanshu013 (commented):
Suggestion - use tables for easier parsing of information 😄

packages/blade/src/components/Button/_decisions/decisions.md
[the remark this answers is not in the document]
chaitanyadeorukhkar: @aditya-razorpay we don't have any use-cases for such buttons as of now. We do have Buttons we icons and we intend to support that. Adding more accessories to button can be done later as and when we see more common patterns being introduced in our system

packages/blade/src/components/Button/_decisions/decisions.md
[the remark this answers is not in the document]
chaitanyadeorukhkar: The Button component will only have a `variant` prop that would allow you to choose the kind of button you want. `contrast` is only required when we are using it along with `intent` for internal Blade components like an `Alert` that could have a `low` as well as `high` contrast `positive` intent button. 

Our assumption is that our products will not require a standalone `low contrast positive` button (except a couple edge cases where we would encourage using custom buttons). We're doing this to ensure these different combinations of buttons don't get over-used within our products leading to breakage in the design language. 

But the fact that we have these capabilities built in the `BaseButton` itself would ensure that we _can_ expose these to our consumers when we see a massive demand for it

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+      -  **default**: `low`
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Because all the variants are made use `low` contrast, it's only when we use it along with `intent` we need `high` contrast. `low` covers a lot more of our current use-cases than `high`

packages/blade/src/components/Button/_decisions/decisions.md
chaitanyadeorukhkar: @sumitd94, yes it will calculate the width based on content & padding.  When we enable `isFullWidth` it will take up the parent's width

packages/blade/src/components/Button/_decisions/decisions.md
chaitanyadeorukhkar: @aditya-razorpay that's the intention. We wouldn't want someone to use a `xlarge` icon with a `small` button. Do you forsee any alarming use-cases where we might need to pass some custom props to the icon directly?

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
...
+- [Component Breakdown](#component-breakdown)
+- [Open Questions](#open-questions)
+
+## Design
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=5200%3A0) to all variants of the Button component
+
+## Internal BaseButton Component
+-  Internal component that exposes certain extra props to enable creation of Component Level Action Buttons like `positive`, `negative`, `neutral`
+-  **BaseButton API**:
+   -  **`variant`**: `primary` | `secondary` | `tertiary` | `link`
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: `BaseButton` is really a Button component with added flexibility. `Button` is a wrapper on top of it which adds some rigidity to the `BaseButton` to avoid misuse. We intend to use `BaseButton` in certain internal blade components like `Alert` or `Modal`. These components would need specific intents & contrasts of buttons that a general button in a product wouldn't.

packages/blade/src/components/Button/_decisions/decisions.md
aditya-razorpay: the only thing I'm worried about is that as the Icon component api widens, the Button api would also widen. If Button took an iconElement then Button and Icon would still be composable without Button having to extend Icon api.

However I do see your point that we want to place some restrictions on what props are passed to Icon when used inside Button which can't be done without React.cloneElement (and not a big fan of that)

chaitanyadeorukhkar: I don't see the icon api getting more complex than what it is today except maybe supporting multicolor icon at some point in the future. I think adding another prop to button would still seem better than to open up the icon prop and let the consumer configure it since that would lead to (allow) misuse from day 1 😓

packages/blade/src/components/Button/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: Makes sense, perhaps add it as a note to the md file?

packages/blade/src/components/Button/_decisions/decisions.md
chaitanyadeorukhkar: That makes sense. If you see the Figma file, we're using intents only to tweak the variants further. Having an `undefined` intent makes it seem like something is missing. How about 
- `intent="none"`  (Seems awfully close to the existing `neutral`)
- `intent="action"` (I like this better since a default button is always used for some action but also confusing since positive, negative, notice, info, neutral are also used for an action)

Any suggestions?

blenderskool: `none` sounds better to me as all kinds of buttons inherently perform some action. But I do agree it is somewhat close to `neutral`.

blenderskool (commented):
Would the `Button` component also expose native `<button>` element attributes(and perhaps the equivalent of react-native)?
It would be primarily useful for `aria-*` props. I have used `title` and `aria-label` attributes in purely icon based buttons in the past to give more context about what the button does to the user and screen reader respectively.

anuraghazra:
> It would be primarily useful for aria-* props. I have used title and aria-label attributes in purely icon based buttons in the past to give more context about what the button does to the user and screen reader respectively.

@blenderskool we will have to expose accessibilityLabel prop https://github.com/razorpay/blade/pull/470#discussion_r874460282

chaitanyadeorukhkar:
@anuraghazra what props to expose would be dependent on which library we decide to use for a11y and how their API needs us to expose things. I think we'll know more in the coming week from your POC

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,64 @@
...
+
+### BaseButton API
+
+| Prop | Type | Required | Default | Description |
+|---|---|---|---|---|
+| variant | `primary`,  `secondary`,  `tertiary`,  `link` | No | `primary` | - The variant of the button to be rendered.<br><br>Note: `link` will look like an inline Link component but will be rendered as a `button` element |
+| intent | `positive`,`negative`, `notice`, `info`, `neutral` | No | undefined | The intent of the button to be rendered.<br><br>  Note: <br> - We will use this to set the intent for internal Blade components that have positive, negative, notice, info, or neutral buttons like an Alert component<br> - All the variants of the button will not be using any intent since it would only be required for internal components. |
+| contrast | `low`, `high` | No | `low` | The contrast of the button to be rendered.<br><br>  Note: We need this for internal Blade components since a `positive` intent could have a `high` as well as a `low` contrast button |
+| size | `large`, `medium`, `small`, `xsmall` | No | `medium` | The size of the button to be rendered. |
```
kamleshchandnani, line 23: are these sizes also good enough to accommodate cases for toasts, alerts, badges?

chaitanyadeorukhkar: I'm assuming, yes. Got these sizes from Figma. Saurav has already used these buttons for creating Alerts

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,64 @@
...
+| children | `string` | No | undefined | The text to be rendered within the button. |
```
kamleshchandnani, line 24: why is children optional?

anuraghazra:
> @anuraghazra what props to expose would be dependent on which library we decide to use for a11y and how their API needs us to expose things

Accessibility props for aria won't depend on that. We will have our own API for that similar to [react-native-web](https://necolas.github.io/react-native-web/docs/accessibility/)

Like this: https://github.com/razorpay/blade/pull/481

packages/blade/src/components/Button/_decisions/decisions.md
chaitanyadeorukhkar: For when we have icon-only button

packages/blade/src/components/Button/_decisions/decisions.md
chaitanyadeorukhkar: @kamleshchandnani @anuraghazra thoughts?

packages/blade/src/components/Button/_decisions/decisions.md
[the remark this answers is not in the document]
nashcheez: @aditya-razorpay Nice observation! Given the nature of how GitHub as a product is it assigns numbers/counters as a composite component to many of its components like buttons, tabs, sidebar etc. Check [ref](https://primer.style/react/Button#counter-component). However, if you assess other commonly used design systems or even sites in Razorpay a pill inside a button is quite rare.

packages/blade/src/components/Button/_decisions/decisions.md
nashcheez: `intent="none"` sounds betters. Should it also be added as a type then?

packages/blade/src/components/Button/_decisions/decisions.md
chaitanyadeorukhkar: Makes sense. We'll make `children` optional only when `icon` is defined and make it mandatory when `icon` isn't defined with some TS magic 🪄

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,64 @@
...
+| Prop | Type | Required | Default | Description |
+|---|---|---|---|---|
+| variant | `primary`,  `secondary`,  `tertiary`,  `link` | No | `primary` | - The variant of the button to be rendered.<br><br>Note: `link` will look like an inline Link component but will be rendered as a `button` element |
+| size | `large`, `medium`, `small`, `xsmall` | No | `medium` | The size of the button to be rendered. |
+| children | `string` | No | undefined | The text to be rendered within the button. |
+| icon | `Icon` | No | undefined | The Blade `Icon` component to be rendered within the button. |
+| iconPosition | `left`, `right` | No | `right` | The position of the rendered icon. |
+| isDisabled | `boolean` | No | `false` | Control whether the button is disabled or not. |
+| isFullWidth | `boolean` | No | `false` | Allows the button to take up the full width of its parent. |
+| onClick | `Function` | No | undefined | The function to be called when the button is clicked. |
```
divyanshu013, line 46: So even for native we'll call this `onClick` rather than `onPress`?

kamleshchandnani:
We probably missed 1 attribute on Button i.e `form`. Not sure how it behaves on React Native but on web we have this attribute so that when we want to submit a form on click of this button we basically use `form` attribute. Read more [about this here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-form)

packages/blade/src/components/Button/_decisions/decisions.md
chaitanyadeorukhkar: Yeah 😓 

We want to keep the API consistent across platforms as much as possible

chaitanyadeorukhkar:
> We probably missed 1 attribute on Button i.e `form`. Not sure how it behaves on React Native but on web we have this attribute so that when we want to submit a form on click of this button we basically use `form` attribute. Read more [about this here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-form). Not sure if we shall add this now or later but this is an important thing. maybe you can explore other DS as well on how they do it

@kamleshchandnani most of the design systems are not exposing the `form` attribute since the usage of it is only relevant when you are using a button outside a form element but within the same document and you want that button to be associated/act on a specific form element

If we don't specify the `form` attribute, it automatically associates itself with the closest ancestor form element.

Do you think we still need to expose a form attribute?

kamleshchandnani:
> > We probably missed 1 attribute on Button i.e `form`. Not sure how it behaves on React Native but on web we have this attribute so that when we want to submit a form on click of this button we basically use `form` attribute. Read more [about this here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-form). Not sure if we shall add this now or later but this is an important thing. maybe you can explore other DS as well on how they do it
> 
> @kamleshchandnani most of the design systems are not exposing the `form` attribute since the usage of it is only relevant when you are using a button outside a form element but within the same document and you want that button to be associated/act on a specific form element
> 
> If we don't specify the `form` attribute, it automatically associates itself with the closest ancestor form element.
> 
> Do you think we still need to expose a form attribute?

cool, in that case we'll descope it for now and re-visit later. I'm merging this PR then, LGTM ✅

packages/blade/src/components/Button/_decisions/decisions.md
chaitanyadeorukhkar: Works

chaitanyadeorukhkar:
@kamleshchandnani dont merge. Pushing a commit

chaitanyadeorukhkar:
@kamleshchandnani you can check now

packages/blade/src/components/Button/_decisions/decisions.md
kamleshchandnani: `intent: none` sounds like a new kind of intent altogether 😑. It's basically `undefined | one of positive, negative, neutral, info, notice`. we do it for all the properties because adding this at one place will force us introduce this pattern across of setting value of undefined props to be meaningful

chaitanyadeorukhkar: With most places we are sticking to the pattern of "default value should be one of the defined types" for instance, `isDisabled` is `false` by default whereas it could've been `undefined` by default and still worked. If we start considering `undefined` as a valid value instead of a "lack of value", we'd have to reconsider our defaults in places like `isDisabled` & `isFullWidth`

Although, `isDisabled={undefined}` will also be allowed as a valid value and will behave exactly like `isDisabled={false}`. By that logic, we should be considering `undefined` as a valid value and in this case having `intent={undefined}` should be valid and we could get rid of `intent="none"`

anuraghazra: I'll go with `undefined | possible unions` where undefined means it can be omitted and in that case the button will not have any intent specific styling as such.

chaitanyadeorukhkar: Leaning more toward having `undefined` as a valid value instead of `none`.

chaitanyadeorukhkar: Upvote this comment for `none`

kamleshchandnani: > I'll go with `undefined | possible unions` where undefined means it can be omitted and in that case the button will not have any intent specific styling as such.
> 
> Edit: also when we say `undefined | possible unions` the undefined key should not be explicitly the part of the union the whole prop should be optional. Otherwise it will force you to set `intent={undefined}` (Could be good or bad depending on if you want this or not.)

yep I love TS philosophy here: `intent?: one of FeedbackVariants` which basically says it's either undefined(doesn't exist) but if it exists then it has to be one of the variants value and internally in JS it's just the matter of if the prop exists then use that intent and apply respected styles else just ignore and none of the styles will be applied for intent

blenderskool: I agree with `undefined` too after understanding the behaviour of `intent` and the analogy given by @kamleshchandnani.
`"none"` at this point is just us defining another form of `undefined`, but without the awesome TS, docs support that `undefined` comes with.

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Internal BaseButton Component
+Internal component that exposes certain extra props to enable creation of Component Level Action Buttons like `positive`, `negative`, `neutral`
+
+### BaseButton API
+
+| Prop | Type | Required | Default | Description |
+|---|---|---|---|---|
+| variant | `primary`,  `secondary`,  `tertiary`,  `link` | No | `primary` | - The variant of the button to be rendered.<br><br>Note: `link` will look like an inline Link component but will be rendered as a `button` element |
+| intent | `none`, `positive`,`negative`, `notice`, `info`, `neutral` | No | `none` | The intent of the button to be rendered.<br><br>  Note: <br> - We will use this to set the intent for internal Blade components that have positive, negative, notice, info, or neutral buttons like an Alert component<br> - All variants of the button will be using the `none` intent which would indicate there is no effect added by the `intent` prop |
```
kamleshchandnani, line 21: @chaitanyadeorukhkar change the values for `intent` here?

chaitanyadeorukhkar: Done

kamleshchandnani (approved):

PR #471, opened 2022-05-16
fix(blade): set defaults for all typography components

kamleshchandnani:


packages/blade/src/components/Typography/Title/Title.tsx
```diff
@@ -15,32 +15,33 @@ const getProps = ({
   type,
 }: Pick<TitleProps, 'variant' | 'type'>): Omit<BaseTextProps, 'children'> => {
   const props: Omit<BaseTextProps, 'children'> = {
-    color: `surface.text.${type}.lowContrast`,
-    fontSize: 25,
```
chaitanyadeorukhkar, line 19: Why were we using 25 before? 🤔

kamleshchandnani: it was just the default value for an object because TS was crying as I'd given the type to it. This wouldn't have been rendered(unless used in JS project and people explicitly didn't pass props) in any case since all the props were required

chaitanyadeorukhkar (dismissed):

chaitanyadeorukhkar:
@kamleshchandnani add a changeset

anuraghazra (dismissed):

anuraghazra (dismissed):

PR #479, opened 2022-05-20
chore(modal): add testID to modal

shamshad17:
- add testID to modal

### DIFF
--- a/.changeset/breezy-icons-obey.md
+++ b/.changeset/breezy-icons-obey.md
@@ -1,6 +1,0 @@
----
-"@razorpay/blade": patch
-"@razorpay/blade-old": patch
----
-
-chore(modal): add testID to bottomsheet modal
--- a/.changeset/breezy-icons-obey.md
+++ b/.changeset/breezy-icons-obey.md
@@ -1,0 +1,6 @@
+---
+"@razorpay/blade": patch
+"@razorpay/blade-old": patch
+---
+
+chore(modal): add testID to bottomsheet modal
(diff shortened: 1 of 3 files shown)


packages/blade-old/src/molecules/Modal/Modal.native.js
```diff
@@ -118,6 +118,7 @@ const Modal = ({ children, variant, visible, onClose, onBackdropClick }) => {
       <Space margin={[0]}>
         <Flex justifyContent="flex-end">
           <RNModal
+            testID="bottomsheet-modal"
```
gupta-ji6, line 121:
```suggestion
            testID="ds-bottomsheet-modal"
```
I think we add `ds` as a prefix as a convention to DS components, check once.

Also, how about adding `testID` as a prop & give a default value to it?

chaitanyadeorukhkar, line 121: Why not accept `testID` from props and do this? We're doing this for other components as well. Refer Button or Card
```suggestion
             {...automation(testID)}
```

gupta-ji6 (dismissed):

packages/blade-old/src/molecules/Modal/Modal.native.js
gupta-ji6: looks good now, resolving

chaitanyadeorukhkar (approved):

.changeset/breezy-icons-obey.md
```diff
@@ -0,0 +1,6 @@
+---
+"@razorpay/blade": patch
```
chaitanyadeorukhkar, line 2: Why are you bumping blade?
```suggestion
```

chaitanyadeorukhkar (changes requested):

.changeset/breezy-icons-obey.md
shamshad17: deleted

chaitanyadeorukhkar (approved):
Added a changeset to only bump blade-old

PR #481, opened 2022-05-20
feat(blade): add makeAccessible function

anuraghazra:
POC

- Added a mapProps function which will act as a compatibility layer between the aria attributes between two platforms.


packages/blade/src/components/NativeAria/NativeAria.web.tsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import mapA11yProps from '../../utils/mapProps';
+import { CloseIcon, EyeIcon } from '../Icons';
+
+const Checkbox: React.FC<{ label: string; checked?: boolean; disabled?: boolean }> = ({
```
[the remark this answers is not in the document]
anuraghazra: Only for testing. the main meat is the [mapProps](https://github.com/razorpay/blade/pull/481/files#diff-a020dcd9e63384659e96412db025c5e4b7651eeae3482c4f7b3f3e758e1a1579) function. 

Oh yeah try on react-native, to run on web change this: 

```diff
//  NativeAria.stories.tsx
- import NativeAria from './';
+ import NativeAria from './NativeAria.web';
```

packages/blade/src/utils/mapProps/a11yMap.ts
```diff
@@ -0,0 +1,93 @@
+export const accessibilityValue = {
+  accessibilityValueMax: 'aria-valuemax',
+  accessibilityValueMin: 'aria-valuemin',
+  accessibilityValueNow: 'aria-valuenow',
+  accessibilityValueText: 'aria-valuetext',
+};
```
kamleshchandnani, line 6: why not have map for `.web` and `.native`? with that you can just have one mapProps file and then the specific platforms will render that a11y key

packages/blade/src/utils/mapProps/mapProps.native.ts
```diff
@@ -0,0 +1,53 @@
...
+  // eslint-disable-next-line guard-for-in
+  for (const key in props) {
+    const propKey = key as keyof AccessibilityMap;
+
+    // Group accessibilityState props
+    if (Object.keys(accessibilityState).includes(propKey)) {
+      const groupKey = accessibilityState[propKey as keyof typeof accessibilityState].replace(
+        'aria-',
+        '',
+      );
```
kamleshchandnani, line 17: all this will go away once you have map for respective platforms.

packages/blade/package.json
```diff
@@ -54,7 +54,9 @@
     "chromatic": "npx chromatic"
   },
   "dependencies": {
-    "@babel/runtime": "7.16.5"
+    "@babel/runtime": "7.16.5",
+    "@react-native-aria/checkbox": "^0.2.3",
+    "@react-stately/toggle": "^3.2.7"
```
kamleshchandnani, line 59: 📌 pin it
```suggestion
    "@react-native-aria/checkbox": "0.2.3",
    "@react-stately/toggle": "3.2.7"
```

kamleshchandnani: Also don't actually see these package import in the files 🤔

packages/blade/src/components/NativeAria/NativeAria.native.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <EyeIcon color="action.icon.link.active" size="small" />
+      ) : (
+        <CloseIcon color="action.icon.link.active" size="small" />
+      )}
+      <Pressable
+        onPress={toggleChecked}
+        {...mapA11yProps({
+          accessibilityDisabled: disabled,
+          accessibilityRole: 'checkbox',
+          accessibilityChecked: isChecked ? 'true' : 'false',
```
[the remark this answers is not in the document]
anuraghazra: I think what you are suggesting is something similar to accessibility hooks. 

where we can do `useCheckboxA11y` and this hook will return the necessary accessibility props. 

But even with that we will need some sort of compatibility layer because that useCheckboxA11y hook needs to return proper props based on platform. so in a way this getProps function can be used internally within the useCheckboxA11y.

packages/blade/src/utils/mapProps/a11yMap.ts
anuraghazra: Actually we are not doing bi-directional mapping. 

Say you have `accessibilityLabel=hello` 
In native this prop will just be directly be returned,
but in web this will be converted to `aria-label`

We are not doing aria-label -> accessibilityLabel conversion.

packages/blade/package.json
anuraghazra: Experimental things.

packages/blade/src/components/NativeAria/NativeAria.native.tsx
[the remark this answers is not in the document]
anuraghazra: Also mapA11yProps function is a general low level utility, not a component specific function. 

Usage as follows

```
mapA11yProps({ accessibiltyRole: 'button', accessibilityLabel: "Search" })

// web: -> { role: 'button', 'aria-label': 'search' } 
// native: -> { accessibiltyRole: 'button', accessibilityLabel: "Search" }
```

packages/blade/src/utils/mapProps/a11yMap.ts
kamleshchandnani: Correct. That's what I'm suggesting, instead of doing the map and changing things there why not create 2 maps in 2 files(`.native`, `.web`)

so you can have 2 files 

```
// accessibilityProps.native.ts

export const accessibilityValue = {
  accessibiltyRole: 'accessibiltyRole',
  accessibilityLabel: 'accessibilityLabel',
};


// accessibilityProps.web.ts

export const accessibilityValue = {
  accessibiltyRole: 'role',
  accessibilityLabel: 'aria-label',
};
```
with this you'll get rid of all the map functions you have to do this programatically.

packages/blade/package.json
anuraghazra: Just experimental things will remove it once done.

packages/blade/src/utils/mapProps/a11yMap.ts
anuraghazra: I see, okay I'll try this approach out and see which works best. 

Although 

```ts
export const accessibilityValue = {
  accessibiltyRole: 'accessibiltyRole',
  accessibilityLabel: 'accessibilityLabel',
};
```

is a bit redundant + now we need to maintain two accessibility maps.

packages/blade/package.json
kamleshchandnani: is it required now since we are in the final stages to merge this

packages/blade/src/utils/mapProps/a11yMap.ts
chaitanyadeorukhkar: Can we remove the word `accessibility` from the keys of `mapA11yProps` & call it `makeAccessible`?

Something like this:
```
makeAccessible({ role, label, checked, valueText, valueMin, valueMax })
```

packages/blade/src/components/NativeAria/NativeAria.native.tsx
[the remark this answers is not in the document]
kamleshchandnani: bumping this so we can discuss this

packages/blade/src/components/NativeAria/NativeAria.native.tsx
```diff
@@ -0,0 +1,66 @@
...
+    </>
+  );
+};
+
+const NativeAria = (): React.ReactElement => {
+  return (
+    <>
+      <Text nativeID="id-group-label">Fruites</Text>
+      <View
+        {...mapA11yProps({
```
kamleshchandnani, line 45: I think we shall have a better demo which shows our workflow. 
* Code sharing
* One component that works on both web and native
* a11y props on the component that resolves based on the platform.

Basically we can have the common checkbox and show the `mapA11y` function with that. It'll give better understanding

packages/blade/src/components/NativeAria/NativeAria.native.tsx
```diff
@@ -0,0 +1,66 @@
...
+          accessibilityRole: 'group',
+          accessibilityLabelledBy: 'id-group-label',
+        })}
+      >
+        <View accessibilityRole="list">
```
kamleshchandnani, line 50: `accessibilityRole` this should have also come from `mapA11y` function

packages/blade/src/utils/mapProps/a11yMap.ts
kamleshchandnani: atleast it's understandable IMO compare to reading all those maps and iterations

anuraghazra: > makeAccessible({ role, label, checked, valueText, valueMin, valueMax })

It's fine but this will clash with lot of other props if we ever create a wrapper component like I proposed in <Box \/> 
Plus I want to make it similar to how react-native calls them.

packages/blade/src/components/NativeAria/NativeAria.native.tsx
anuraghazra: This function will be internal right? Why do we need to have a full showcase? Or do we make an internal storybook?
And this NativeAria.tsx example is just for development purposes.

packages/blade/src/utils/mapProps/mapProps.web.ts
```diff
@@ -0,0 +1,23 @@
+import { accessibilityMap } from './a11yMap.web';
+import type { AccessibilityMap } from './a11yMap.web';
+
+const mapA11yProps = (props: AccessibilityMap): Record<string, unknown> => {
+  const newProps: Record<string, unknown> = {};
+
+  // eslint-disable-next-line guard-for-in
+  for (const key in props) {
+    const propKey = key as keyof AccessibilityMap;
```
kamleshchandnani, line 9:
```suggestion
    const accessibilityPropKey = key as keyof AccessibilityMap;
```

packages/blade/src/utils/mapProps/mapProps.web.test.ts
```diff
@@ -0,0 +1,51 @@
+import mapProps from './mapProps';
+
+describe('mapProps', () => {
+  it('should work', () => {
```
kamleshchandnani, line 4: can we have better contextual names for test?

these are simple keys so we can mention something like

`should return correct accessibility attributes for flat accessibility keys`

packages/blade/src/utils/mapProps/mapProps.web.test.ts
```diff
@@ -0,0 +1,51 @@
...
+        accessibilityLabelledBy: 'id1',
+        accessibilityRole: 'button',
+      }),
+    ).toStrictEqual({
+      'aria-labelledby': 'id1',
+      'aria-label': 'hello world',
+      role: 'button',
+    });
+
+    expect(mapProps({ accessibilityLiveRegion: 'polite' })).toStrictEqual({
```
kamleshchandnani, line 17: can have another `it` which can have title 

`should return correct accessibility attributes for accessibilityLiveRegion`

packages/blade/src/utils/mapProps/mapProps.web.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      'aria-live': 'polite',
+    });
+
+    expect(
```
kamleshchandnani, line 21: add another `it` which can have title something like

`should return correct accessibility attributes for accessibility states`

kamleshchandnani: similarly can do for other tests

packages/blade/src/utils/mapProps/mapProps.native.ts
```diff
@@ -0,0 +1,64 @@
...
+  const newProps: Record<string, any> = {};
+
+  // loop through all the incoming props and map them
+  for (const key in props) {
+    const propKey = key as keyof AccessibilityMap;
+    const propValue = props[propKey];
+    const a11yKey = accessibilityMap[propKey];
+
+    // group accesibilityState prop for native
+    if (isAccessibilityStateProp(propKey)) {
```
kamleshchandnani, line 29: Arre anurag sir, we are just using this at one single place can just do `accessibilityStateKeys.includes(propKey)` in here only na? why to create one more function?

packages/blade/src/utils/mapProps/mapProps.native.ts
```diff
@@ -0,0 +1,64 @@
...
+      console.warn('No mapping found for', propKey);
+    }
+  }
+
+  // ignore unsupported roles in native
+  if (
+    newProps.accessibilityRole &&
+    !supportedRolesInNative.includes(newProps.accessibilityRole as string)
+  ) {
+    delete newProps.accessibilityRole;
```
kamleshchandnani, line 58: 👏 

shall we add a console warning? just to give some feedback to consumers?

kamleshchandnani:
Overall the PR is pretty much understandable and readable now. kudos 👏

packages/blade/src/components/NativeAria/NativeAria.native.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
```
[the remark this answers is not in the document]
anuraghazra: 😆 no idea, suggest something better

packages/blade/src/utils/mapProps/a11yMap.web.ts
```diff
@@ -0,0 +1,62 @@
...
+  accessibilityRole: 'role',
+  accessibilityRoleDescription: 'aria-roledescription',
+  accessibilityRowCount: 'aria-rowcount',
+  accessibilityRowIndex: 'aria-rowindex',
+  accessibilityRowSpan: 'aria-rowspan',
+  accessibilitySetSize: 'aria-setsize',
+  accessibilitySort: 'aria-sort',
+};
+
+export type AccessibilityMap = Partial<
```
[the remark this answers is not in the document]
anuraghazra: Oh yeah I'm in process of fixing the types please ignore this.

packages/blade/src/utils/mapProps/a11yMap.native.ts
```diff
@@ -0,0 +1,96 @@
+export const accessibilityValue = {
```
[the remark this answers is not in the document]
anuraghazra: Yeah sure.

packages/blade/src/components/NativeAria/useCheckboxA11yProps.ts
```diff
@@ -0,0 +1,14 @@
+import mapA11yProps from '../../utils/mapProps';
+
+type CheckboxA11yProps = { isDisabled?: boolean; isChecked?: boolean };
+
+export const useCheckboxA11yProps = ({
```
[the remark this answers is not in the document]
anuraghazra: Hmmmm.... I kinda resonate with the hooks mindset, a custom accessibility hook. 🤔  Although let's go with `get` for now we'll refine the patterns later no point discussing it right now.

packages/blade/src/utils/mapProps/mapProps.web.test.ts
anuraghazra: yeah i just skimmed through them will improve these descriptions, thanks!

packages/blade/src/utils/mapProps/mapProps.native.ts
anuraghazra: Yeah good idea.

anuraghazra:
Resolved most of the comments. 

TODO: Fix types

packages/blade/src/components/NativeAria/NativeAria.native.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: `AriaDemo`?

packages/blade/src/components/NativeAria/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from './NativeAria.native';
+export * from './NativeAria.native';
```
chaitanyadeorukhkar, line 2: Why is this exporting `.native` files? It should just do this
```suggestion
export { default } from './NativeAria';
export * from './NativeAria';
```

packages/blade/src/components/NativeAria/getCheckboxA11yProps.ts
```diff
@@ -0,0 +1,14 @@
...
+type CheckboxA11yProps = { isDisabled?: boolean; isChecked?: boolean };
+
+export const getCheckboxA11yProps = ({
+  isDisabled,
+  isChecked,
+}: CheckboxA11yProps): Record<string, unknown> => {
+  return mapA11yProps({
+    accessibilityDisabled: isDisabled,
+    accessibilityRole: 'checkbox',
+    accessibilityChecked: isChecked ? 'true' : 'false',
```
chaitanyadeorukhkar, line 12: Seems to be some incorrect type here

packages/blade/src/components/NativeAria/NativeAria.native.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Can we move all of this to `storybook-recipes`? Seems weird to keep this in `components`.

packages/blade/src/utils/mapProps/accessibilityMap.web.ts
```diff
@@ -0,0 +1,68 @@
...
+  accessibilityRowCount: 'aria-rowcount',
+  accessibilityRowIndex: 'aria-rowindex',
+  accessibilityRowSpan: 'aria-rowspan',
+  accessibilitySetSize: 'aria-setsize',
+  accessibilitySort: 'aria-sort',
+  accessibilityCurrent: 'aria-current',
+  accessibilityDropEffect: 'aria-dropeffect',
+  accessibilityGrabbed: 'aria-grabbed',
+  accessibilityLevel: 'aria-level',
+  accessibilityRelevant: 'aria-relevant',
```
chaitanyadeorukhkar, line 64: Why are we appending `accessibility` to all the keys? Lets just remove these keys? Much simpler API. The function name `mapA11yProps` or whatever we decide to call it should be a good enough context on what these keys are

anuraghazra: Not sure how expressive that would be, With `accessibility` prefixes the API is same as [`react-native-web`](https://necolas.github.io/react-native-web/docs/accessibility/), [`react-native`](https://reactnative.dev/docs/accessibility) & web also prefixes with `aria-` 

Just that working with `aria` and accessibility it just feel kinda unusual to not have the prefix of some kind. 
```ts
  return mapAccessibilityProps({
    role: 'checkbox',
    disabled: isDisabled,
    checked: isChecked,
  });
```

packages/blade/src/components/NativeAria/NativeAria.native.tsx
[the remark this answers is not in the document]
kamleshchandnani: moving to recipes is a good idea actually since stories are for direct components, for such demos we have recipes section

packages/blade/src/utils/mapProps/accessibilityMap.web.ts
anuraghazra: Decided to go with makeAccessible and no accessibility prefix.

packages/blade/.storybook/react-native/storybook.requires.js
```diff
@@ -24,6 +24,7 @@ argsEnhancers.forEach((enhancer) => addArgsEnhancer(enhancer));
 const getStories = () => {
   return [
     require('../../src/components/Icons/Icons.stories.tsx'),
+    require('../../src/components/NativeAria/NativeAria.stories.tsx'),
```
kamleshchandnani, line 27: we need to update this right based on the restructure? just run the native storybook and this will be updated

packages/blade/src/storybook-recipes/AccessibilityLayer/AccessibilityLayer.native.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
```
kamleshchandnani, line 1: nit(better names sorry but not sorry): can we rename this to `AccessibilityInteropDemo`? We are actually showing a demo and not the layer iteself

packages/blade/src/storybook-recipes/AccessibilityLayer/AccessibilityLayer.native.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { CloseIcon, EyeIcon } from '../../components/Icons';
+import { getCheckboxA11yProps } from './getCheckboxA11yProps';
+
+const CheckboxItem: React.FC<{ label: string; checked?: boolean; disabled?: boolean }> = ({
+  label,
+  checked,
+  disabled,
+}) => {
+  const [isChecked, setChecked] = React.useState(checked);
+  const checkboxA11yProps = getCheckboxA11yProps({ isDisabled: disabled, isChecked });
```
kamleshchandnani, line 13: nit: give some mehnat to you finger sand make this name accessible :P 

```suggestion
  const checkboxAccessibilityProps = getCheckboxAccessibilityProps({ isDisabled: disabled, isChecked });
```

packages/blade/src/storybook-recipes/AccessibilityLayer/AccessibilityLayerDemo.stories.tsx
```diff
@@ -0,0 +1,11 @@
+import type { ComponentStory, Meta } from '@storybook/react';
```
kamleshchandnani, line 1: see yaha pe you have named file correctly :D

packages/blade/src/storybook-recipes/AccessibilityLayer/AccessibilityLayerDemo.stories.tsx
```diff
@@ -0,0 +1,11 @@
...
+import NativeAria from './AccessibilityLayer';
```
kamleshchandnani, line 2: fix `NativeAria`

packages/blade/src/utils/index.ts
```diff
@@ -7,3 +7,4 @@ export { default as merge } from './merge';
 export { default as toTitleCase } from './toTitleCase';
 export { default as useBreakpoint } from './useBreakpoint';
 export { default as useColorScheme } from './useColorScheme';
+export { default as mapA11yProps } from './makeAccessible';
```
kamleshchandnani, line 10: fix import name

packages/blade/src/utils/makeAccessible/accessibilityMapNative.ts
```diff
@@ -0,0 +1,99 @@
+import type { AccessibilityMap } from './makeAccessible.d';
+
+export const accessibilityValue = {
```
kamleshchandnani, line 3: no explicity type for this?

packages/blade/src/utils/makeAccessible/accessibilityMapNative.ts
```diff
@@ -0,0 +1,99 @@
...
+  valueMax: 'max',
+  valueMin: 'min',
+  valueNow: 'now',
+  valueText: 'text',
+};
+
+export const accessibilityState = {
```
kamleshchandnani, line 10: same for this - no explicit type?

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
```diff
@@ -0,0 +1,278 @@
...
+  | 'tabpanel'
+  | 'term'
+  | 'textbox'
+  | 'timer'
+  | 'toolbar'
+  | 'tooltip'
+  | 'tree'
+  | 'treegrid'
+  | 'treeitem'
+  | (string & Record<string, unknown>);
```
kamleshchandnani, line 76: question: why this?

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
```diff
@@ -0,0 +1,278 @@
...
+
+export type AccessibilityProps = CommonAccessibilityProps;
+export type CommonAccessibilityKeys = keyof CommonAccessibilityProps;
+export type AccessibilityMap = Record<CommonAccessibilityKeys, string>;
```
kamleshchandnani, line 80: nit: can we move this to below `CommonAccessibilityProps` basically using the types after they are declared. not sure why the tseslint didn't trigger error 🤔  it does for vars though

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
```diff
@@ -0,0 +1,278 @@
...
+
+export type CommonAccessibilityProps = {
+  role: AriaRoles;
+  /** Identifies the currently active element when DOM focus is on a composite widget, textbox, group, or application. */
+  activeDescendant?: string;
+  /** Indicates whether assistive technologies will present all, or only parts of, the changed region based on the change notifications defined by the aria-relevant attribute. */
+  atomic?: boolean;
+  /**
+   * Indicates whether inputting text could trigger display of one or more predictions of the user's intended value for an input and specifies how predictions would be
+   * presented if they are made.
```
kamleshchandnani, line 90:
```suggestion
   * Indicates whether inputting text could trigger display of one or more predictions of the user's intended value for an input and specifies how predictions would be presented if they are made.
```

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
```diff
@@ -0,0 +1,278 @@
...
+  | 'treeitem'
+  | (string & Record<string, unknown>);
+
+export type AccessibilityProps = CommonAccessibilityProps;
+export type CommonAccessibilityKeys = keyof CommonAccessibilityProps;
+export type AccessibilityMap = Record<CommonAccessibilityKeys, string>;
+
+export type CommonAccessibilityProps = {
+  role: AriaRoles;
+  /** Identifies the currently active element when DOM focus is on a composite widget, textbox, group, or application. */
```
kamleshchandnani, line 84: nit(can ignore): let's make these consistent?
```suggestion
  /**
  * Identifies the currently active element when DOM focus is on a composite widget, textbox, group, or application.
  */
```

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
```diff
@@ -0,0 +1,278 @@
...
+  activeDescendant?: string;
+  /** Indicates whether assistive technologies will present all, or only parts of, the changed region based on the change notifications defined by the aria-relevant attribute. */
```
kamleshchandnani, line 86: same as above

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
```diff
@@ -0,0 +1,278 @@
...
+  atomic?: boolean;
+  /**
+   * Indicates whether inputting text could trigger display of one or more predictions of the user's intended value for an input and specifies how predictions would be
+   * presented if they are made.
+   */
+  autoComplete?: 'none' | 'inline' | 'list' | 'both';
+  /** Indicates an element is being modified and that assistive technologies MAY want to wait until the modifications are complete before exposing them to the user. */
```
kamleshchandnani, line 93: same as above

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
```diff
@@ -0,0 +1,278 @@
...
+  | 'tree'
+  | 'treegrid'
+  | 'treeitem'
+  | (string & Record<string, unknown>);
+
+export type AccessibilityProps = CommonAccessibilityProps;
+export type CommonAccessibilityKeys = keyof CommonAccessibilityProps;
+export type AccessibilityMap = Record<CommonAccessibilityKeys, string>;
+
+export type CommonAccessibilityProps = {
```
kamleshchandnani, line 82: why is it called commona11yprops? are there uncommon props?

packages/blade/src/utils/makeAccessible/webToNativeRole.ts
```diff
@@ -0,0 +1,27 @@
+const webToAccessibilityRole = {
```
kamleshchandnani, line 1: what is the intent of this var? finding it hard to understand 🤔

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
```diff
@@ -0,0 +1,76 @@
...
+  const newProps: Record<string, any> = {};
+
+  // loop through all the incoming props and map them
+  for (const key in props) {
+    const propKey = key as keyof AccessibilityMap;
+    const propValue = props[propKey];
+    const accessibilityAttribute = accessibilityMap[propKey];
+
+    // group accesibilityState prop for native
+    if (isAccessibilityStateProp(propKey)) {
```
kamleshchandnani, line 31: can we inline this?
`accessibilityStateKeys.includes(prop)`

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
```diff
@@ -0,0 +1,76 @@
...
+      newProps.accessibilityState = {
+        ...newProps.accessibilityState,
+        [accessibilityAttribute]: propValue,
+      };
+      continue;
+    }
+
+    // group accesibilityValue prop for native
+    if (isAccessibilityValueProp(propKey)) {
```
kamleshchandnani, line 40: can we inline this?
`isAccessibilityValueProp`

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
```diff
@@ -0,0 +1,76 @@
...
+      console.warn(
+        `[Blade: makeAccessible]: No mapping found for ${propKey}. Make sure you have entered valid key`,
+      );
+    }
+  }
+
+  if (newProps.accessibilityRole) {
+    // map web to native overlapping roles
+    const role = webToAccessibilityRole(newProps.accessibilityRole);
+    newProps.accessibilityRole = role;
```
kamleshchandnani, line 60: can you share an example? didn't quite understand this?

packages/blade/src/utils/makeAccessible/makeAccessible.web.ts
```diff
@@ -0,0 +1,25 @@
+import { accessibilityMap } from './accessibilityMapWeb';
+import type { AccessibilityMap, AccessibilityProps } from './makeAccessible.d';
+
+const makeAccessible = (props: Partial<AccessibilityProps>): Record<string, unknown> => {
+  const newProps: Record<string, unknown> = {};
```
kamleshchandnani, line 5: why is it `Record<string, unknown>`? the new props shall have `key:value` pairs which are valid right?

packages/blade/src/utils/makeAccessible/accessibilityMapNative.ts
anuraghazra: Not needed we are already adding explicit type in the map object and spreading this object onto it.

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
anuraghazra: I copy pasted this union from the React d.ts file. We can remove it

We don't need this but since you asked what this does is, say you want a union to have specific set of values but also accept any arbitrary string value but if you simply do this: 

Screenshot 2022-06-17 at 4 32 19 PM

You will lose autocomplete, adding `string` in union collapsed the whole union to just string.


But the trick here is if you add`(string & {})` it will accept any value without error but also show autocomplete

Screenshot 2022-06-17 at 4 35 19 PM

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
anuraghazra: in typescript there's no execution context, all types are hoisted to a symbol table.

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
anuraghazra: We can call it `AriaAttributes` these are 'WAI-ARIA 1.1 attributes'

packages/blade/src/utils/makeAccessible/makeAccessible.web.ts
anuraghazra: It is defining key:value pairs. just that we don't know what the value would be. It can be either boolean, string, 'mixed' or anything. I should change this to `any` actually.

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
anuraghazra: There are common overlaps between some of the `roles` between web and native. for example in web `slider` maps to `adjustable` in native. 

Example: 
https://github.com/razorpay/blade/pull/481/files#diff-b10f9269d534b1f77c6207692521ca38f0790ea3da3d0d45252deb29d1a8f81dR18

packages/blade/src/storybook-recipes/AccessibilityInterop/AccessibilityInteropDemo.stories.tsx
```diff
@@ -0,0 +1,11 @@
+import type { ComponentStory, Meta } from '@storybook/react';
+import AccessibilityInteropDemo from './AccessibilityInteropDemo';
+
+export default {
+  title: 'Recipes/AccessibilityInterop',
+  component: AccessibilityInteropDemo,
+} as Meta;
+
+export const MapAccessibilityProps: ComponentStory<typeof AccessibilityInteropDemo> = () => {
```
kamleshchandnani, line 9:
```suggestion
export const AccessibilityInteropDemoTemplate: ComponentStory<typeof AccessibilityInteropDemo> = () => {
```

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
kamleshchandnani: 👀 ohoooo interesting, but why is that the case that `string` vanishes the autocomplete?

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
kamleshchandnani: let's do that

packages/blade/src/utils/makeAccessible/webToNativeRole.ts
kamleshchandnani: so this var is basically the web a11y roles? if yes, then we can probably rename it to `webAccessibilityRoles`?

packages/blade/src/utils/makeAccessible/makeAccessible.d.ts
anuraghazra: 'hello' | 'world' | string

If you put 'hello' | 'world' a literal type union and add `string` which is wider than both `hello` | `world` it will widen the whole union to just be `string` type. 

tldr; `string` type is wider than `hello` | `world` both literals.

packages/blade/src/utils/makeAccessible/webToNativeRole.ts
anuraghazra: No not web. It's `web -> native`

[This](https://github.com/razorpay/blade/pull/481#discussion_r900022908)

For example: 
Let's say you are building a <Slider \/> component and you have to put proper aria role for it. 

In web we have the [`slider`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/slider_role) role
but in native it's called [`adjustable`](https://reactnative.dev/docs/accessibility#accessibilityrole)

So we are unifying out makeAccessible() API to be consistent. 

if we say:
```
makeAccessible({ role: 'slider' }) 
```
This will return `slider` for web and `adjustable` for native.

kamleshchandnani: let's discuss this in person

packages/blade/src/utils/makeAccessible/webToNativeRole.ts
```diff
@@ -0,0 +1,24 @@
...
```
kamleshchandnani, line 1:
```suggestion
const webToNativeAccessibilityRoleMap = {
```

packages/blade/src/utils/makeAccessible/webToNativeRole.ts
```diff
@@ -0,0 +1,24 @@
...
+const webToNativeRole = (role: keyof typeof webToAccessibilityRole): string => {
+  if (role) {
+    const inferredRole = webToAccessibilityRole[role];
+    if (inferredRole !== null) {
+      // ignore roles that don't map to web
+      return inferredRole || role;
+    }
+  }
+  return role;
+};
```
kamleshchandnani, line 22:
```suggestion
const webToNativeAccesibilityRole = (role: keyof typeof webToAccessibilityRole): string => {
   const nativeRole = webToAccessibilityRole?.[role];
    // ignore roles that don't map to web
    return nativeRole || role;
};
```

packages/blade/src/utils/makeAccessible/accessibilityMapNative.ts
```diff
@@ -0,0 +1,99 @@
...
+  setSize: 'accessibilitySetSize',
+  sort: 'accessibilitySort',
+  current: 'accessibilityCurrent',
+  dropEffect: 'accessibilityDropEffect',
+  grabbed: 'accessibilityGrabbed',
+  level: 'accessibilityLevel',
+  relevant: 'accessibilityRelevant',
+};
+
+export const supportedAccessibilityRoles = [
```
kamleshchandnani, line 70: maybe move this to `webToNativeAccessibilityRole.ts` file since it better belongs there

kamleshchandnani:
```
const accessibilityRoleMap = {
  slider: 'adjustable',
  alert: 'alert',
  button: 'button',
  checkbox: 'checkbox',
  combobox: 'combobox',
  header: 'header',
  image: 'image',
  imagebutton: 'imagebutton',
  keyboardkey: 'keyboardkey',
  link: 'link',
  menu: 'menu',
  menubar: 'menubar',
  menuitem: 'menuitem',
  presentation: 'none',
  progressbar: 'progressbar',
  radio: 'radio',
  radiogroup: 'radiogroup',
  scrollbar:'scrollbar',
  search:'search',
  spinbutton:'spinbutton',
  region:'summary',
  switch:'switch',
  tab: 'tab',
  tablist: 'tablist',
  text: 'text',
  timer: 'timer',
  togglebutton: 'togglebutton',
  toolbar: 'toolbar',
  region: 'summary',
}
```

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -89,6 +90,7 @@ const Heading = <T extends { variant: HeadingVariant }>({
   children,
 }: HeadingProps<T>): ReactElement => {
   const props = getProps({ variant, type, weight, contrast });
+  console.log(props);
```
kamleshchandnani, line 93: remove log 😄

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
```diff
@@ -0,0 +1,78 @@
...
+        `[Blade: makeAccessible]: No mapping found for ${propKey}. Make sure you have entered valid key`,
+      );
+    }
+  }
+
+  if (newProps.accessibilityRole) {
+    // map web to native overlapping roles
+    type Roles = keyof typeof accessibilityRoleMap;
+    const role = accessibilityRoleMap[newProps.accessibilityRole as Roles];
+    newProps.accessibilityRole = role;
```
kamleshchandnani, line 63: shouldn't this `newProps.accessibilityRole` only be assigned when the role exists and is valid? basically after the if condition?

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // ignore unsupported roles
+    if (!role) {
+      const validRoles = Object.keys(accessibilityRoleMap).join(', ');
+      console.log(
+        `[Blade: makeAccessible]: Unsupported accessibilityRole for native, Valid roles are: ${validRoles}`,
```
kamleshchandnani, line 69:
```suggestion
        `[Blade: makeAccessible]: Unsupported accessibility role for react-native. Expected one of ${validRoles} but found ${newProps.accessibilityRole}`
```

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
anuraghazra: We are not bailing out with `return` inside `if (!role)` thus if we put it after it it will be assigned to undefined. putting `return newProps` will introduce another return path for the function.

kamleshchandnani (approved):

PR #482, opened 2022-05-20
feat: add `BaseButton` & `Button` components

chaitanyadeorukhkar:
Adds `BaseButton` & `Button` components


packages/blade/src/components/Button/BaseButton/BaseButton.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+  Stories,
+  PRIMARY_STORY,
+  Description,
+} from '@storybook/addon-docs';
+import iconMap from '../../Icons/iconMap';
+import type { BaseButtonIcon, BaseButtonProps } from './BaseButton';
+import BaseButtonComponent from './BaseButton';
+
+export default {
+  title: 'Components/Button/BaseButton (Internal)',
```
kamleshchandnani, line 16: 💡 Possibly what we can also do is rename the internal stories with `*.stories.internal.tsx` and then in our path mapping add a dev flag which will enable these kind of stories on dev mode and will be stripped off in prod story so we can have better DX

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+type BaseButtonWithIconProps = BaseButtonCommonProps & {
+  icon: IconComponentType;
+  children?: string;
+};
+
+export type BaseButtonProps<T> = T extends {
+  icon?: infer Icon;
+}
+  ? Icon extends Extract<BaseButtonIcon, IconComponentType>
```
kamleshchandnani, line 35: why extract?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+    ? BaseButtonWithIconProps
+    : Icon extends Extract<BaseButtonIcon, undefined | null>
+    ? BaseButtonWithoutIconProps
+    : T
```
kamleshchandnani, line 39: @anuraghazra can we have `if` conditions for typing instead of tertiary. the nested tertiary is so hard to read 😢

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+    contrast,
+    size,
+    icon,
+    iconPosition,
+    isDisabled,
+    isFullWidth,
+    onClick,
+    type,
+    children,
+  });
```
kamleshchandnani, line 72: remove this

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const BaseButton = <T extends { icon: BaseButtonIcon }>({
+  variant = 'primary',
+  intent,
+  contrast = 'low',
+  size = 'medium',
+  icon,
+  iconPosition = 'left',
+  isDisabled = false,
+  isFullWidth = false,
```
kamleshchandnani, line 50: why to give default values to this? it's as equivalent of checking if this prop exists then use that value?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+    ? BaseButtonWithoutIconProps
+    : T
+  : T;
+
+const BaseButton = <T extends { icon: BaseButtonIcon }>({
+  variant = 'primary',
+  intent,
+  contrast = 'low',
+  size = 'medium',
+  icon,
```
kamleshchandnani, line 47:
```suggestion
  icon: Icon,
```

you can title case it inline only

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+  iconPosition = 'left',
+  isDisabled = false,
+  isFullWidth = false,
+  onClick,
+  type = 'button',
+  children,
+}: BaseButtonProps<T>): ReactElement => {
+  const { theme } = useTheme();
+  const buttonColor = getIn(theme.colors, 'action.background.primary.default');
```
kamleshchandnani, line 56: how about contrast?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+    <StyledBaseButton
+      color={buttonColor}
+      hoverColor={hoverColor}
+      onClick={(): void => {
+        console.log('clicked');
+      }}
+      disabled={isDisabled}
+      activeColor={activeColor}
+    >
+      {IconComponent ? <IconComponent size="small" color="action.icon.primary.default" /> : null}
```
kamleshchandnani, line 83: shouldn't we check for icon position too?

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -0,0 +1,31 @@
+import type { StyledBaseButtonProps } from './StyledBaseButton';
+
+const getBaseButtonStyles = ({
+  color,
+  hoverColor,
+  activeColor,
+}: Omit<StyledBaseButtonProps, 'children' | 'onClick'>): string => `
+  display:flex;
+  flex-direction: row;
```
kamleshchandnani, line 9: do we really need this? 🤔

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -0,0 +1,31 @@
...
+  height: 48px;
+  cursor: pointer;
+  background-color: ${color};
+  border: 2px solid ${color};
+  color: white;
+  border-radius: 2px;
+  justify-content: center;
+  align-items: center;
+  gap: 8px;
```
kamleshchandnani, line 18: use the spacing token?

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -0,0 +1,31 @@
...
+  align-self: center;
+  padding: 0px 20px 0px 20px;
```
kamleshchandnani, line 20: same for padding use spacing token?

packages/blade/src/components/Icons/iconMap.ts
```diff
@@ -8,10 +8,10 @@ import DollarIconComponent from './DollarIcon';
 import EyeIconComponent from './EyeIcon';
 import EyeOffIconComponent from './EyeOffIcon';
 import RupeeIconComponent from './RupeeIcon';
-import type { IconProps } from '.';
+import type { IconComponentType } from '.';
```
kamleshchandnani, line 11:
```suggestion
import type { IconComponent } from '.';
```

nit: Let's keep it as `IconComponent` as while defining and importing we already use `type` keyword

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -0,0 +1,31 @@
...
+  color,
+  hoverColor,
+  activeColor,
+}: Omit<StyledBaseButtonProps, 'children' | 'onClick'>): string => `
+  display:flex;
+  flex-direction: row;
+  height: 48px;
+  cursor: pointer;
+  background-color: ${color};
+  border: 2px solid ${color};
```
kamleshchandnani, line 13: we also have a border token so use that?

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -0,0 +1,31 @@
...
+  color: white;
+  border-radius: 2px;
```
kamleshchandnani, line 15: we have a border radius token too

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: Wanted it to represent that `Icon` needs to be one of the types from `BaseButtonIcon`. I could keep it `Icon extends IconComponentType` but it seems very detached from the `BaseButtonIcon`. Imagine `IconComponentType` evolves over time but we fail to change it here

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: Since in our API we have mentioned that these would be `false` by default. Just a good practice to keep it boolean since the type is boolean

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
chaitanyadeorukhkar: Ignore all the styles, they'll all change 😅

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
anuraghazra: no 👅  Thats the `if` condition.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+  icon?: undefined;
+  children: string;
+};
+
+type BaseButtonWithIconProps = BaseButtonCommonProps & {
+  icon: IconComponentType;
+  children?: string;
+};
+
+export type BaseButtonProps<T> = T extends {
```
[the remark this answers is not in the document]
anuraghazra: Yes it's not needed I think. Because there is no union type. and children/icon both wouldn't have any intelliSense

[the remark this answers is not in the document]
chaitanyadeorukhkar: Haha yes, cleaned it up now

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
kamleshchandnani: why not?

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
kamleshchandnani: why not?

packages/blade/src/components/Button/BaseButton/BaseButton.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+    },
+  },
+} as Meta<BaseButtonProps<{ icon: BaseButtonIcon }>>;
+
+const BaseButtonTemplate: ComponentStory<typeof BaseButtonComponent> = ({
+  icon,
+  children,
+  ...args
+}) => {
+  const IconComponent = iconMap[(icon as unknown) as string];
```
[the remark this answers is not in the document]
kamleshchandnani: @chaitanyadeorukhkar ??

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <StyledBaseButton
+      color={buttonColor}
+      hoverColor={hoverColor}
+      onClick={(): void => {
+        console.log('clicked');
+      }}
+      disabled={isDisabled}
+      activeColor={activeColor}
+    >
+      {Icon ? <Icon size="small" color="action.icon.primary.default" /> : null}
```
kamleshchandnani, line 77: shouldn't we check for the icon position? or you've implemented that in later PRs?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+      <BaseText
+        lineHeight="s"
+        fontSize={100}
+        fontWeight="bold"
+        textAlign="center"
+        color="action.text.primary.default"
```
kamleshchandnani, line 83: hardcoded value? just for minimalistic implementation?

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
chaitanyadeorukhkar: Ignore all the styles, they'll all change. Will take care of all of this in chained PRs. Check the scope of this PR in the description. This is just a dummy implementation

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <StyledPressable
+      onPress={onClick}
+      color={color}
+      hoverColor={hoverColor}
+      activeColor={activeColor}
+      style={({ pressed }): { backgroundColor: string; borderColor: string } => ({
+        backgroundColor: pressed ? activeColor : color,
+        borderColor: pressed ? activeColor : color,
+      })}
```
kamleshchandnani, line 26: can't this be part of `getBaseButtonStyles`? or this is something explicitly that the `StyledPressable` needs as attributes?

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+describe('<BaseButton />', () => {
+  it('should render button with default properties', () => {
+    const buttonText = 'Pay Now';
+    const { toJSON, getByText } = renderWithTheme(<BaseButton>{buttonText}</BaseButton>);
+    expect(toJSON()).toMatchSnapshot();
+    expect(getByText('Pay Now')).toBeTruthy();
+  });
+  it('should throw an error when there is no icon or text passed to the button', () => {
+    try {
+      // @ts-expect-error
```
kamleshchandnani, line 17: can you mention the description about the expected error here? or is it temporary and will be handled in subsequent PRs?

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+afterAll(() => jest.restoreAllMocks());
+
+describe('<BaseButton />', () => {
+  it('should render button with default properties', () => {
+    const buttonText = 'Pay Now';
+    const { container, getByText, getByRole } = renderWithTheme(
+      <BaseButton>{buttonText}</BaseButton>,
+    );
+    expect(container).toMatchSnapshot();
+    expect(getByRole('button')).toBeInTheDocument();
```
kamleshchandnani, line 15: ❤️

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+import renderWithTheme from '../../../../_helpers/testing/renderWithTheme.web';
+import BaseButton from '../BaseButton';
+
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
+afterAll(() => jest.restoreAllMocks());
+
+describe('<BaseButton />', () => {
+  it('should render button with default properties', () => {
```
kamleshchandnani, line 9: add test for icon-only button? and then you can do a negative assertion for a11y things below

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: Handled in later PRs

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: Handled in later PRs

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
chaitanyadeorukhkar: Will handle this better in later PRs

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
chaitanyadeorukhkar: I've added tests for icon only button in later PRs. Specifically this one -> #484

chaitanyadeorukhkar: > and then you can do a negative assertion for a11y things below

Not sure what you mean by this?

packages/blade/src/components/Button/BaseButton/BaseButton.stories.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Not sure what you mean here. I'm changing the type here since the actual `icon` prop is of type `IconComponent` but here in the stories I'm using it as a `string` to give a better experience for when developers are trying icons in the story

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -0,0 +1,31 @@
+import type { StyledBaseButtonProps } from './StyledBaseButton';
+
+const getBaseButtonStyles = ({
+  color,
+  hoverColor,
+  activeColor,
+}: Omit<StyledBaseButtonProps, 'children' | 'onClick'>): string => `
+  display:flex;
+  flex-direction: row;
+  height: 48px;
```
kamleshchandnani, line 10: should be `min-height` else the zooming and default font size change will break layouts. Check this [demo app](https://codesandbox.io/s/szi8i?file=/src/styles.css)  from [RFC](https://github.com/razorpay/blade/blob/master/rfcs/2021-02-19-units-for-typography-and-layout.md#what-will-work-for-us)

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -0,0 +1,31 @@
...
+  background-color: ${color};
+  border: 2px solid ${color};
+  color: white;
+  border-radius: 2px;
+  justify-content: center;
+  align-items: center;
+  gap: 8px;
+  align-self: center;
+  padding: 0px 20px 0px 20px;
+  &:hover {
```
kamleshchandnani, line 21: shall we try out using https://styled-components.com/docs/advanced#style-objects. Think it's better than plan interpolated strings. Realised late after releasing Typography.

chaitanyadeorukhkar: Yup, lets do this. Much better

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -0,0 +1,31 @@
...
+    background-color: ${hoverColor};
+    border: 2px solid ${hoverColor};
+  }
+  &:active {
```
kamleshchandnani, line 25: what happens to these pseudo selectors on native? they don't break?

chaitanyadeorukhkar: They get ignored

anuraghazra: What is the alternative in react native then? There is no "active" state?

chaitanyadeorukhkar: There is an active state, it's when you press the button to indicate it's being pressed. There is no focus state though. The `pressed` state we have to handle in `StyledBaseButton.native.tsx` since that's the only way they've enabled it. Ref: https://reactnative.dev/docs/pressable#style

packages/blade/src/components/Button/BaseButton/BaseButton.stories.tsx
chaitanyadeorukhkar: Done in #490

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,91 @@
+import type { ReactElement } from 'react';
+import getIn from '../../../utils/getIn';
+import BaseText from '../../Typography/BaseText';
+import { useTheme } from '../../BladeProvider';
+import type { IconComponent } from '../../Icons';
+import StyledBaseButton from './StyledBaseButton';
+
+export type BaseButtonIcon = undefined | null | IconComponent;
```
kamleshchandnani, line 8: why do you need this?

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -0,0 +1,89 @@
...
+    disabled,
+    focusBorderColor,
+    focusColor,
+    focusRingColor,
+    hoverBorderColor,
+    hoverColor,
+    isFullWidth,
+    borderWidth,
+    borderRadius,
+  }: Omit<StyledBaseButtonProps, 'children' | 'onClick'>) =>
```
anuraghazra, line 22: You are anyway destructuring the props inside getBaseButtonStyles and not using any ...spreading right? then we can directly do this 

```ts

const StyledPressable = styled.Pressable(
  (props: Omit<StyledBaseButtonProps, 'children' | 'onClick'>) => getBaseButtonStyles(props),
);
```

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+};
+
+type BaseButtonWithIconProps = BaseButtonCommonProps & {
+  icon: IconComponent;
+  children?: string;
+};
+
+export type BaseButtonProps = BaseButtonWithIconProps | BaseButtonWithoutIconProps;
+
+const StyledBaseText = styled(BaseText)(
```
kamleshchandnani, line 42:
```suggestion
const ButtonText = styled(BaseText)(
```
Better to not confuse with BaseText or StyledBaseText? this name is anyways not conflicting and contextual too

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+      borderRadius={borderRadius}
+    >
+      {Icon && iconPosition == 'left' ? <Icon size={iconSize} color={iconColor} /> : null}
+      {text && (
+        <StyledBaseText
+          lineHeight={lineHeight}
+          fontSize={fontSize}
+          fontWeight="bold"
+          textAlign="center"
+          color={textColor}
```
kamleshchandnani, line 246: don't we explicitly need to add these properties to the `StyledBaseText` argument? we just have mentioned few of the props there

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+          iconPosition={iconPosition}
+          hasIcon={!!Icon}
+          iconSpacing={iconSpacing}
+        >
+          {text}
+        </StyledBaseText>
+      )}
+      {Icon && iconPosition == 'right' ? <Icon size={iconSize} color={iconColor} /> : null}
```
kamleshchandnani, line 254: the icon right is not working on the storybook, can you please check? Also not sure what all props are functional on the storybook, a couple of them don't work. 


Screenshot 2022-06-14 at 12 33 31 PM

packages/blade/src/components/Button/BaseButton/StyledBaseButton.d.ts
```diff
@@ -0,0 +1,24 @@
+import type { BaseButtonProps } from './BaseButton';
+import type { ButtonMinHeight } from './buttonTokens';
+
+export type StyledBaseButtonProps = Omit<BaseButtonProps, 'icon' | 'children'> & {
+  activeBorderColor: string;
+  activeColor: string;
+  borderColor: string;
```
kamleshchandnani, line 7: since we are prefixing all other states then let's prefix this to `defaultBorderColor` to keep it consistent and avoid confusion?

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
kamleshchandnani: i think it's fine to be explicit, since if there's any bug that might be introduced in BaseButton and any dev on DS passes any prop which shouldn't be passed then it might be bypassed if we add `props`.

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
kamleshchandnani: any updates here?

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
kamleshchandnani: bumping this

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const buttonText = 'Pay Now';
+    const { toJSON } = renderWithTheme(<BaseButton size="medium">{buttonText}</BaseButton>);
+    expect(toJSON()).toMatchSnapshot();
+  });
+  it('should render small size button', () => {
+    const buttonText = 'Pay Now';
+    const { toJSON } = renderWithTheme(<BaseButton size="small">{buttonText}</BaseButton>);
+    expect(toJSON()).toMatchSnapshot();
+  });
+  it('should render xsmall size button', () => {
```
kamleshchandnani, line 44: nit(can ignore): tests for `xsmall` ➡️ `small` ➡️ `medium` ➡️ so on

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+  it('should render disabled button', () => {
+    const buttonText = 'Pay Now';
+    const onClick = jest.fn();
+    const { toJSON, getByText } = renderWithTheme(
+      <BaseButton isDisabled={true} onClick={onClick}>
+        {buttonText}
+      </BaseButton>,
+    );
+    const button = getByText(buttonText);
+    fireEvent.press(button);
```
kamleshchandnani, line 90: can we simulate a test to check the styles when pressed? Basically tests which can test different states(focus, active, etc) of button

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
kamleshchandnani: we don't have something similar to test this on React Native too?

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const { container } = renderWithTheme(<BaseButton isFullWidth={true}>{buttonText}</BaseButton>);
+    expect(container).toMatchSnapshot();
+  });
+  it('should render disabled button', () => {
+    const buttonText = 'Pay Now';
+    const { container, getByRole } = renderWithTheme(
+      <BaseButton isDisabled={true}>{buttonText}</BaseButton>,
+    );
+    expect(container).toMatchSnapshot();
+    expect(getByRole('button')).toBeDisabled();
```
kamleshchandnani, line 93: we don't have similar thing on native? over there we have to simulate that with fireEvent.

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+          `[Blade: BaseButton]: Cannot render a BaseButton without an icon or text`,
+        );
+      }
+    }
+  });
+  it('should render large size button', () => {
+    const buttonText = 'Pay Now';
+    const { toJSON } = renderWithTheme(<BaseButton size="large">{buttonText}</BaseButton>);
+    expect(toJSON()).toMatchSnapshot();
+  });
```
kamleshchandnani, line 33: nit(please don't ignore): add a line break after every `it` because right now things look very cluttered

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -0,0 +1,50 @@
+import type { ThemeTokens } from '../../../tokens/theme/theme.d';
+import type { TypographyPlatforms } from '../../../tokens/global/typography';
+
+export type ButtonMinHeight = '48px' | '40px' | '32px' | '28px';
+
+export type ButtonTypography = {
+  [key in TypographyPlatforms]: {
+    fonts: {
+      size: Record<string, keyof ThemeTokens['typography'][key]['fonts']['size']>;
```
kamleshchandnani, line 9: `Record<string, keyof ThemeTokens['typography'][key]['fonts']['size']>;` in this instead of `string` it should be union of allowed sizes right? 

Basically 👇 ?

`Record<'xsmall' | 'small' | 'medium' | 'larger', keyof ThemeTokens['typography'][key]['fonts']['size']>;`

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -0,0 +1,50 @@
...
+    };
+    lineHeights: Record<string, keyof ThemeTokens['typography'][key]['lineHeights']>;
+  };
+};
+
+const typography: ButtonTypography = {
+  onDesktop: {
+    fonts: {
+      size: {
+        large: 200,
```
kamleshchandnani, line 19: was wondering what are we getting from using the key here? can't we directly use the token here? `typography.onDesktop.fonts.size[200]`

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -0,0 +1,50 @@
...
+        medium: 100,
+        small: 50,
+        xsmall: 50,
+      },
+    },
+    lineHeights: {
+      large: 'm', // this is 24px but we need 20px
```
kamleshchandnani, line 42: is this fixed?

packages/blade/src/components/Icons/RupeeIcon/RupeeIcon.tsx
```diff
@@ -1,9 +1,8 @@
-import type { ReactElement } from 'react';
 import { Svg, Path } from '../_Svg';
-import type { IconProps } from '..';
+import type { IconComponent } from '..';
 import useIconProps from '../useIconProps';
 
-const RupeeIcon = ({ size, color }: IconProps): ReactElement => {
+const RupeeIcon: IconComponent = ({ size, color }) => {
```
kamleshchandnani, line 5: what is the difference between writing it this way vs that way?

I understand with this we are typing the variable and in the earlier approach we were typing the object arg as well as the function arg, but when to use which approach?

cc: @anuraghazra

packages/blade/src/utils/makeBorderSize/makeBorderSize.ts
```diff
@@ -0,0 +1,10 @@
+function makeBorderSize<T extends number>(s: T): `${T}px`;
```
kamleshchandnani, line 1:
```suggestion
function makeBorderSize<T extends number>(size: T): `${T}px`;
```

packages/blade/src/utils/makeBorderSize/makeBorderSize.ts
```diff
@@ -0,0 +1,10 @@
...
+function makeBorderSize<T extends string>(s: T): T;
+function makeBorderSize<T extends number | string>(size: T): `${T}px` | T {
```
kamleshchandnani, line 3: if we are already overloading types above then why are we again typing things inline here? shouldn't it infer since the function name is same

also, with this inline type defintion we might be creating a bug? because this says the input can be `string` or `number` and then output can be `{T}px` or `T` which means 

`makeBorderSize('50%')` ➡️ `50px` is a valid return type which logically is incorrect. 

correct me if I didn't get this correctly?

packages/blade/src/utils/makeBorderSize/makeBorderSize.test.ts
```diff
@@ -0,0 +1,17 @@
...
+
+describe('makeBorderSize', () => {
+  it('should return the border width in `px`', () => {
+    const space = makeBorderSize(border.width.thin);
+    expect(space).toEqual('1px');
+  });
+  it('should return the border radius in `px`', () => {
+    const space = makeBorderSize(border.radius.small);
+    expect(space).toEqual('2px');
+  });
```
kamleshchandnani, line 12: also i think we shall restrict this use case and only stick with `%` or number

packages/blade/src/utils/makeBorderSize/makeBorderSize.test.ts
```diff
@@ -0,0 +1,17 @@
...
+  it('should return the border radius in `%`', () => {
+    const space = makeBorderSize(border.radius.round);
+    expect(space).toEqual('50%');
+  });
```
kamleshchandnani, line 16: add a negative test case where we can check that `makeBorderSize('50%')` should never return `50px` or `makeBorderSize(50)` should never return `50`(without units)

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
anuraghazra: Yeah agree about being explicit, the point was that we are already being explicit inside the getBaseButtonStyles function. So seems like double work to explicitly destructure props twice.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+    focusColor: getIn(theme.colors, 'action.background.primary.focus'),
+    focusBorderColor: getIn(theme.colors, 'action.border.primary.focus'),
+    focusRingColor: getIn(theme.colors, 'brand.primary.400'),
+    borderWidth: makeBorderSize(theme.border.width.thin),
+    borderRadius: makeBorderSize(theme.border.radius.small),
+  };
+
+  switch (size) {
+    case 'large':
+      props.iconSize = 'medium';
```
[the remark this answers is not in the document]
anuraghazra: small/medium/large was the conclusion.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+      props.lineHeight = buttonTypographyTokens.lineHeights.xsmall;
+      props.minHeight = '28px';
+      props.iconSpacing = makeSpace(theme.spacing[1]);
+      props.spacing = `${makeSpace(theme.spacing[0])} ${makeSpace(theme.spacing[2])}`;
+      props.text = children?.trim().toUpperCase();
+      break;
+    default:
+  }
+
+  if (isDisabled) {
```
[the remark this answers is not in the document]
anuraghazra: In here we are natively passing the html `disabled` prop to be true when isDisabled is true right? @chaitanyadeorukhkar?? In that case focus styles won't even apply because focus will be lost.

packages/blade/src/components/Icons/RupeeIcon/RupeeIcon.tsx
anuraghazra: Depends, One is a function declaration, where you are typing the function when defining it's implementation. 

And the second one is a function type signature or function call signature (whatever you wanna call it)

Here's the difference: https://tsplay.dev/wEPRvW

packages/blade/src/utils/makeBorderSize/makeBorderSize.ts
anuraghazra: > if we are already overloading types above then why are we again typing things inline here?

You have to provide the overlapping function arguments but you can omit the return type (although eslint will yell)

But it's best to type the return type explicitly here otherwise it won't throw if we do something like this: 

```ts
function makeBorderSize(size: number | string) {
  if (typeof size === 'number') {
    return `${size}whatever`; // <- should throw error, but doesn't. 
    // In runtime the value will be `whatever` but in type level it's showing `px` if you hover over it, which is really bad.
  }
  return size;
}
```

> also, with this inline type defintion we might be creating a bug? because this says the input can be string or number and then output can be {T}px or T which means
> makeBorderSize('50%') ➡️ 50px is a valid return type which logically is incorrect.

No it won't do that, thats the whole point of having overloads and doing a runtime check of `typeof size === 'number'` inside the function, TypeScript is smart to discriminate the return type based on the argument passed either number or string.

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
chaitanyadeorukhkar: There is a `getByRole` for react native but we have to explicitly pass `accessibilityRole='button'` to our component. Held this off since we weren't doing a11y in this PR. We could probably add a test for it once we add some a11y props to the button

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: Missed it. Removing this.

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
chaitanyadeorukhkar: We're being explicit with `getBaseButtonStlyes`. It makes sense to not destruct it here and use `props`.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+  size?: 'large' | 'medium' | 'small' | 'xsmall';
+  iconPosition?: 'left' | 'right';
+  isDisabled?: boolean;
+  isFullWidth?: boolean;
+  onClick?: () => void;
+  type?: 'button' | 'reset' | 'submit';
+};
+
+type BaseButtonWithoutIconProps = BaseButtonCommonProps & {
+  icon?: undefined;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Intentional. This is the `BaseButtonWithoutIconProps` type where there is no icon in which case `children` is mandatory.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+import makeSpace from '../../../utils/makeSpace';
+import type { TypographyPlatforms } from '../../../tokens/global/typography';
+import makeBorderSize from '../../../utils/makeBorderSize';
+import StyledBaseButton from './StyledBaseButton';
+import type { ButtonMinHeight, ButtonTypography } from './buttonTokens';
+import { typography } from './buttonTokens';
+
+export type BaseButtonIcon = undefined | null | IconComponent;
+
+type BaseButtonCommonProps = {
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: We'll add it later when we have the functionality

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+    Pick<BaseButtonProps, 'iconPosition'> & { hasIcon: boolean; iconSpacing: string }) => ({
+    paddingLeft: hasIcon && iconPosition === 'left' ? iconSpacing : '0px',
+    paddingRight: hasIcon && iconPosition === 'right' ? iconSpacing : '0px',
+  }),
+);
+
+type BaseButtonStyleProps = {
+  iconSize: IconSize;
+  fontSize: keyof Theme['typography']['fonts']['size'];
+  lineHeight: keyof Theme['typography']['lineHeights'];
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: This is not exposed to the consumers, this is only saying that these are the valid and acceptable `lineHeights` we're enforcing boundaries using `buttonTokens` and not using global lineHeights directly

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+  minHeight: ButtonMinHeight;
+  iconSpacing: string;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Because the value of `iconSpacing` would be something like `4px`. This is why I had put comments earlier that some of these props are raw values and not token values. We need raw values in place where we're directly using it with a styled component. We use token values to pass it down to a blade component like `BaseText`.

https://github.com/razorpay/blade/pull/482/files#diff-7cae5a72597e5debaf6982c2f7afd5d7e4a19abbdc3e598a8e16c772a1dce75dR95

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+  iconColor: IconProps['color'];
+  textColor: BaseTextProps['color'];
+  spacing: string;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: https://github.com/razorpay/blade/pull/482#discussion_r897534974

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+  text?: string;
+  color: string;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: `color` is the button color. `textColor` is the text color for BaseText. We had it named as `buttonColor` before but changed it to `color` since it should be implicit that its for button unless explicitly mentioned like `textColor`

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+  borderColor: string;
+  hoverColor: string;
+  hoverBorderColor: string;
+  activeColor: string;
+  activeBorderColor: string;
+  focusColor: string;
+  focusBorderColor: string;
+  focusRingColor: string;
+  borderWidth: string;
+  borderRadius: string;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Same like earlier, this would be the raw value we get after computing `borderRadius: makeBorderSize(theme.border.radius.small)`

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,259 @@
...
+  children?: string;
+  isDisabled: boolean;
+  theme: Theme;
+  size: BaseButtonCommonProps['size'];
+}): BaseButtonStyleProps => {
+  const props: BaseButtonStyleProps = {
+    iconSize: 'medium',
+    fontSize: buttonTypographyTokens.fonts.size.medium,
+    lineHeight: buttonTypographyTokens.lineHeights.medium,
+    minHeight: '40px',
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: We actually don't. We have `makeSpace`, `makeBorderSize`, `makeTypographySize`, `makeMotionTime`.

Since this was `px` in both platforms, I used it directly. Should we have a `makeDimensionSize`?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yes, small medium large.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yup thats correct. Focus styles won't apply when disabled is true

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: Not sure what you mean here. Its an instance of `BaseText`, we're only extending it with a styled wrapper

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: Working for me 🤔 

Except `contrast` & `intent`, all other props should work - as they are on localhost for me. CI Is failing, not able to check if its working there (I'll debug the failing CI)
Screenshot 2022-06-15 at 10 40 22 AM

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
chaitanyadeorukhkar: Yeah, tried doing this. Its a bit weird on native. We have to pass `testOnly_pressed={true}` for us to simulate a `pressed` behaviour 🤦 

This means passing `testOnly_pressed` from the root `BaseButton` leading to the same prop being exposed to consumers as well as web which doesn't seem like a good enough tradeoff

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
chaitanyadeorukhkar: Don't have the similar equivalent there. We could do `container.props.disabled` assertion but its as good as doing a snapshot check which will check for prop as well as styles

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
chaitanyadeorukhkar: Yup, makes more sense

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
chaitanyadeorukhkar: We want this to get computed from a theme, right? We use a key here which the component will then use to access the correct font size in its theme.

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
chaitanyadeorukhkar: Yup, fixed Removed the comments in other PR, forgot here

packages/blade/src/components/Icons/RupeeIcon/RupeeIcon.tsx
chaitanyadeorukhkar: In my non-ts expert understanding - Its now easier for our consumers to create a new component by using `IconComponent` type directly instead of specifying the props & return type

This also allows us to use `IconComponent` in places like Button where want to accept a `icon` prop and we can make it's type as `icon: IconComponent` so that our expectations are clearly mentioned

packages/blade/src/utils/makeBorderSize/makeBorderSize.test.ts
chaitanyadeorukhkar: What do you mean? `border.radius.small` is a `number` here. So we are already doing % or number

packages/blade/src/utils/makeBorderSize/makeBorderSize.test.ts
chaitanyadeorukhkar: We already have test cases for both 

**Test A:** `expect(makeBorderSize(2)).toEqual('2px');`
and
**Test B:** `expect(makeBorderSize('50%')).toEqual('50%');`

If for any reason `makeBorderSize('50%')` returned `50px`, **Test B** would fail and similarly if for any reason `makeBorderSize(2)` returned `2`, **Test A** would fail

Not sure we we need negative test here?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: CI is working but just realised that you cant see it in CI since its a prod build and this is an internal story. Can you check again on your local?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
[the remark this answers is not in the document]
kamleshchandnani: it can just be `makeSize`

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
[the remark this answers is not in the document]
kamleshchandnani: didn't understand what do you mean by button color? ideally color and textColor are same thing, color in CSS world is color of the text. is there any possibility both of the these values will be different?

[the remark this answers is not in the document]
chaitanyadeorukhkar: Here color means the color of the button. I'm not mapping it exactly to the CSS property since it's the color of the button (`background-color`), hence `defaultColor`, `hoverColor`, `activeColor`, etc.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: You're confusing two types here. This is `BaseButtonWithoutIconProps` where the case is that the icon will not exist but the children will need to be mandatory. There is another type called `BaseButtonWithIconProps` and that is where we'll have `icon: IconComponent`. [See this line](https://github.com/razorpay/blade/pull/482/files#diff-7cae5a72597e5debaf6982c2f7afd5d7e4a19abbdc3e598a8e16c772a1dce75dR34)

Later when we have variants, intents, contrasts it becomes a bit more complex and I've added comments to explain it better. Check this [block of code](https://github.com/razorpay/blade/pull/506/files#diff-7cae5a72597e5debaf6982c2f7afd5d7e4a19abbdc3e598a8e16c772a1dce75dR27-R70).

chaitanyadeorukhkar:
@kamleshchandnani check the latest commit https://github.com/razorpay/blade/pull/482/commits/0a977e89c4dafd19d2dc09c15487ec9891fb026c

The switch case was getting a bit unnecessary so removed it, moved things to button tokens, and controlled it via `size`. See if it makes sense.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,231 @@
...
+      hoverBorderColor={hoverBorderColor}
+      hoverColor={hoverColor}
+      isFullWidth={isFullWidth}
+      onClick={onClick}
+      type={type}
+      borderWidth={borderWidth}
+      borderRadius={borderRadius}
+    >
+      {Icon && iconPosition == 'left' ? <Icon size={iconSize} color={iconColor} /> : null}
+      {text && (
```
kamleshchandnani, line 212: nit: follow one pattern?

above you're using ternary operator for conditional rendering, here you're using && operator?

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -0,0 +1,102 @@
...
+        medium: 100,
+        small: 75,
+        xsmall: 75,
+      },
+    },
+    lineHeights: {
+      large: 's',
+      medium: 'l',
+      small: 'l',
+      xsmall: 'l',
```
kamleshchandnani, line 38: nit(sorry but not sorry): sort the value? xsmall ➡️ small ➡️ medium and so on for all the objects in this file. 🙈

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  it('should render large size button', () => {
+    const buttonText = 'Pay Now';
+    const { toJSON } = renderWithTheme(<BaseButton size="large">{buttonText}</BaseButton>);
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render button with icon without text', () => {
+    const { toJSON } = renderWithTheme(<BaseButton icon={CreditCardIcon} />);
```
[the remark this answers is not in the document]
anuraghazra: I'll cover a11y related tests on my PRs.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah, we should have stricter types. I've added this commit [`2851947` (#482)](https://github.com/razorpay/blade/pull/482/commits/28519470a7b186e809b118fa25a019374ef4b124) that increases type safety for most of these. Can you see if this makes sense? Had to change the type of tokens a bit too

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -0,0 +1,56 @@
...
+  focusBorderColor,
+  borderWidth,
+  borderRadius,
+}: StyledBaseButtonProps): React.ReactElement => {
+  return (
+    <StyledPressable
+      onPress={onClick}
+      style={({ pressed }): { backgroundColor: string; borderColor: string } => ({
+        backgroundColor: pressed ? activeColor : defaultColor,
+        borderColor: pressed ? activeColor : defaultColor,
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: No, this implementation will change in the chained PR for variants

packages/blade/src/components/Button/BaseButton/StyledBaseButton.web.tsx
```diff
@@ -0,0 +1,42 @@
...
+    focusBorderColor,
+    focusColor,
+    focusRingColor,
+    hoverBorderColor,
+    hoverColor,
+    isFullWidth,
+    borderWidth,
+    borderRadius,
+  }: Omit<StyledBaseButtonProps, 'children' | 'onClick'>) =>
+    getBaseButtonStyles({
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Has been done in the chained PR

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
```diff
@@ -0,0 +1,114 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+import { fireEvent } from '@testing-library/react-native';
+import renderWithTheme from '../../../../_helpers/testing/renderWithTheme.native';
+import { CreditCardIcon } from '../../../Icons';
+import BaseButton from '../BaseButton';
+
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: @anuraghazra will be doing it

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -0,0 +1,102 @@
...
+export type ButtonTypography = {
+  [key in TypographyPlatforms]: {
+    fonts: {
+      size: Record<
+        Required<BaseButtonProps['size']>,
+        keyof ThemeTokens['typography'][key]['fonts']['size']
+      >;
+    };
+    lineHeights: Record<
+      Required<BaseButtonProps['size']>,
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Removed `undefined` & `null` from the type in cases where we want to enforce the type while ensuring that the value exists. Size by default can be undefined in the props but since we assign it a default value as `medium`, we want to remove `undefined` from the type here

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
chaitanyadeorukhkar: Will do this later on, I'll add a comment in the file to remember. Its not just here but in other places too where I need to sort things by name. Avoiding such things since it causes a huge mess in rebasing chained PRs.

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -0,0 +1,102 @@
...
+    topBottom: 0,
+    rightLeft: 2,
+  },
+  small: {
+    topBottom: 0,
+    rightLeft: 3,
+  },
+  medium: {
+    topBottom: 0,
+    rightLeft: 5,
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Eg. `padding: "2px 4px"`

TopBottom is 2px
RightLeft is 4px

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -0,0 +1,51 @@
...
+    borderColor: hoverBorderColor,
+  },
+  '&:active': {
+    backgroundColor: activeColor,
+    borderColor: activeBorderColor,
+  },
+  '&:focus': {
+    backgroundColor: focusColor,
+    borderColor: focusBorderColor,
+    boxShadow: `0px 0px 0px 4px ${focusRingColor}`,
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: This actually doesn't fit any of our shadow tokens. 

1. We don't have `0px` in `offsetY`
2. We don't have `0px` in `blurRadius`
3. We don't have a `spread` token in shadow

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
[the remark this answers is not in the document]
anuraghazra: You can remove the Required type (it also clashes with the built-in Required type). 

and replace with the built-in [`NonNullable`](https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype) type. It does the same thing.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -0,0 +1,231 @@
...
+    activeBorderColor: getIn(theme.colors, 'action.border.primary.active'),
+    focusBackgroundColor: getIn(theme.colors, 'action.background.primary.focus'),
+    focusBorderColor: getIn(theme.colors, 'action.border.primary.focus'),
+    focusRingColor: getIn(theme.colors, 'brand.primary.400'),
+    borderWidth: makeBorderSize(theme.border.width.thin),
+    borderRadius: makeBorderSize(theme.border.radius.small),
+  };
+
+  if (isDisabled) {
+    const disabledColor = getIn(theme.colors, 'action.background.primary.disabled');
```
kamleshchandnani, line 121:
```suggestion
    const disabledBackgroundColor = getIn(theme.colors, 'action.background.primary.disabled');
```

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -0,0 +1,56 @@
...
+  focusBorderColor,
+  borderWidth,
+  borderRadius,
+}: StyledBaseButtonProps): React.ReactElement => {
+  return (
+    <StyledPressable
+      onPress={onClick}
+      style={({ pressed }): { backgroundColor: string; borderColor: string } => ({
+        backgroundColor: pressed ? activeBackgroundColor : defaultBackgroundColor,
+        borderColor: pressed ? activeBackgroundColor : defaultBackgroundColor,
```
kamleshchandnani, line 33: can't we have this directly in the styled func instead of passing it via style attribute?

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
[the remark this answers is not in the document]
kamleshchandnani: if it doesn't fits any of our tokens then we shall either update this or our tokens, that's the intent

packages/blade/src/_helpers/types.ts
```diff
@@ -11,4 +11,11 @@ type DotNotationColorStringToken<TokenType> = {
     : DotNotationColorStringToken<TokenType[K]>}`;
 }[keyof TokenType];
 
-export { DotNotationColorStringToken };
+type Required<T> = Exclude<T, undefined | null>;
+
+type ValueOf<
```
kamleshchandnani, line 16: what is the significane of this? and what does this do?

packages/blade/src/components/Button/BaseButton/StyledBaseButton.d.ts
```diff
@@ -0,0 +1,25 @@
+import type { Theme } from '../../BladeProvider';
+import type { BaseButtonProps } from './BaseButton';
+import type { ButtonMinHeight } from './buttonTokens';
+
+export type StyledBaseButtonProps = Omit<BaseButtonProps, 'icon' | 'children'> & {
+  activeBorderColor: string;
+  activeBackgroundColor: string;
+  defaultBorderColor: string;
+  minHeight: `${ButtonMinHeight}px`;
+  spacing: `${ValueOf<Theme['spacing']>}px ${ValueOf<Theme['spacing']>}px`;
```
kamleshchandnani, line 10: this might work here but later we might need to have a makeSpace equivalent at Type level basis that the types would return px/rem

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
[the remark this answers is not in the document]
kamleshchandnani: @chaitanyadeorukhkar 👆

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
chaitanyadeorukhkar: This is not the final implementation for the styling. [See this](https://github.com/razorpay/blade/pull/498/files#diff-4adb37d058188f426fa85fcce5f5ea64bcd0b62448fa8eba65cee240c606c479R37) for the final implementation

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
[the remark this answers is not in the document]
chaitanyadeorukhkar: This is not really a shadow, so not sure if this should even go in our shadow tokens. Lets discuss it on call.

packages/blade/src/_helpers/types.ts
chaitanyadeorukhkar: This will extract the values of all the keys inside an object/array

```
{
	"a": 1,
	"b": 2,
	"c": 3,
}
```

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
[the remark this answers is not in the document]
kamleshchandnani: discussed and tried why it won't work.

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
```diff
@@ -123,4 +122,228 @@ describe('<BaseButton />', () => {
...
+    );
+    expect(container).toMatchSnapshot();
+  });
+  it('should render disabled negative intent low contrast button', () => {
+    const buttonText = 'Pay Now';
+    const { container } = renderWithTheme(
+      <BaseButton intent="negative" contrast="low" isDisabled={true}>
+        {buttonText}
+      </BaseButton>,
+    );
```
kamleshchandnani, line 194: tests for checking a11y? using that axe tool with no violations and testing all the roles and a11y labels?

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -1,24 +1,23 @@
 import type { ThemeTokens } from '../../../tokens/theme/theme.d';
 import type { TypographyPlatforms } from '../../../tokens/global/typography';
-import type { Required } from '../../../_helpers/types';
 import type { IconProps } from '../../Icons';
 import type { BaseButtonProps } from './BaseButton';
 
 //TODO: Sort keys in all objects
 
-export type ButtonMinHeight = 48 | 40 | 32 | 28;
+export type ButtonMinHeight = 48 | 36 | 32 | 28;
```
kamleshchandnani, line 8:
```suggestion
export type ButtonMinHeight = 28 | 32 | 36 | 48;
```

nit(sorry OCD triggering): sort these values?

kamleshchandnani (approved):

PR #483, opened 2022-05-23
fix: remove `surface.action.icon.link.*` colors from `theme`

chaitanyadeorukhkar:


packages/blade/src/_helpers/types.ts
```diff
@@ -1,13 +1,11 @@
-import type { ColorContrast } from '../tokens/theme/theme.d';
-
 /**
  * @template TokenType token type generic
  * @description Tokenises objects to dot notation strings, eg: `surface.text.normal.lowContrast`
  */
 type DotNotationColorStringToken<TokenType> = {
   [K in keyof TokenType]: `${Extract<K, number | string>}.${TokenType[K] extends Record<
     string,
-    ColorContrast | string
```
chaitanyadeorukhkar, line 10: Don't need this anymore since we have separate types for `ActionVariantsWithContrast` & `ActionVariants`

packages/blade/src/tokens/theme/theme.d.ts
```diff
@@ -102,9 +125,9 @@ export type Colors = {
     background: Record<'level1' | 'level2' | 'level3', ColorContrast>;
     border: Record<'normal' | 'subtle', ColorContrast>;
     text: Record<TextTypes, ColorContrast>;
-    action: {
-      icon: Pick<ActionVariants, 'link'>;
-    };
+    // action: {
+    //   icon: Pick<ActionStates, 'link'>;
+    // };
```
chaitanyadeorukhkar, line 130: Removing `surface.action.icon` types since we won't have any standalone clickable icon on any of the surfaces. It would always be a button icon for which the relevant types exist

packages/blade/src/components/Icons/index.tsx
```diff
@@ -4,14 +4,40 @@ import type { DotNotationColorStringToken } from '../../_helpers/types';
 type FeedbackColors = `feedback.icon.${DotNotationColorStringToken<
   Theme['colors']['feedback']['icon']
 >}`;
-type SurfaceColors = `surface.icon.${DotNotationColorStringToken<
-  Theme['colors']['surface']['action']['icon']
+
+// TODO: Figure out a better way to handle Feedback**Colors types without having to repeat them
```
anuraghazra, line 8: This should work. 

```ts
type FeedbackColorTypes = 'positive' | 'negative' | 'information' | 'notice' | 'neutral';
type FeedbackVariantColors = `feedback.${FeedbackColorTypes}.action.icon.${DotNotationColorStringToken<
  Theme['colors']['feedback'][FeedbackColorTypes]['action']['icon']
>}`;
```

packages/blade/src/tokens/theme/theme.d.ts
```diff
@@ -43,20 +43,43 @@ export type ColorContrast = {
...
+  disabled: ColorContrast;
+};
+
+// export type ActionStates = {
+//   default: ColorContrast;
+//   hover: ColorContrast;
+//   focus: ColorContrast;
+//   active: ColorContrast;
+//   disabled: ColorContrast;
+// };
```
kamleshchandnani, line 67: remove this?

packages/blade/src/tokens/theme/theme.d.ts
kamleshchandnani: confirm with Saurav once and then delete this entirely

packages/blade/src/components/Icons/index.tsx
```diff
@@ -4,14 +4,17 @@ import type { DotNotationColorStringToken } from '../../_helpers/types';
 type FeedbackColors = `feedback.icon.${DotNotationColorStringToken<
   Theme['colors']['feedback']['icon']
 >}`;
-type SurfaceColors = `surface.icon.${DotNotationColorStringToken<
-  Theme['colors']['surface']['action']['icon']
+
+type FeedbackVariants = 'positive' | 'negative' | 'information' | 'notice' | 'neutral';
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Calling it `FeedbackActions`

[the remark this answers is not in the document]
chaitanyadeorukhkar: There is a similar type defined in theme.d. Importing that now

kamleshchandnani (dismissed):

kamleshchandnani:
The changeset shall be patch?

chaitanyadeorukhkar:
> The changeset shall be patch?

@kamleshchandnani  kept it as minor since we're commenting out certain tokens. If someone was using them for any reason, we don't want to break their usage with a patch. Thoughts?

kamleshchandnani:
> > The changeset shall be patch?
> 
> @kamleshchandnani kept it as minor since we're commenting out certain tokens. If someone was using them for any reason, we don't want to break their usage with a patch. Thoughts?

Sure then we shall add proper guidelines mentioning that

chaitanyadeorukhkar:
@kamleshchandnani done

PR #484, opened 2022-05-25
feat: add icon & size support to `BaseButton` (Button [2])

chaitanyadeorukhkar:
**Things to review**:
- Spacing
- Height & Width
- `size` prop & its effect on font, dimensions, & spacing
- `icon` prop & its effect on spacing
- `iconPosition` prop & its effect on spacing
- `buttonTokens` for component level tokens & its consumption

**Things not to review**
- Other props
- Variant specific styling
- States of button


packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -31,6 +35,106 @@ type BaseButtonWithIconProps = BaseButtonCommonProps & {
 
 export type BaseButtonProps = BaseButtonWithIconProps | BaseButtonWithoutIconProps;
 
+const StyledBaseText = styled(BaseText)(
```
chaitanyadeorukhkar, line 38: Need to do this since we don't have an enhancer component for `Spacing` which we need to add before the button. I've added support for extending BaseText with `styled` (this means accepting a `style` prop for it to work on native). I could have done the same with `Icon` component but that would means accepting and passing down `style` prop for _every_ new icon component that is created (Eg. CreditCardIcon, ChevronDownIcon, etc.).

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -31,6 +35,106 @@ type BaseButtonWithIconProps = BaseButtonCommonProps & {
...
+    iconPosition,
+    iconSpacing,
+  }: BaseTextProps &
+    Pick<BaseButtonProps, 'iconPosition'> & { hasIcon: boolean; iconSpacing: string }) => `
+    padding-left: ${hasIcon && iconPosition === 'left' ? iconSpacing : 0};
+    padding-right: ${hasIcon && iconPosition === 'right' ? iconSpacing : 0};
+`,
+);
+
+type ConfigProps = {
```
chaitanyadeorukhkar, line 50: Here `prop` indicates that it is a prop that needs to be returned and passed to a DS component like `BaseText` whereas `raw value` indicates that it is a raw value that needs to be passed down to a `styled` component like `StyledBaseButton` to be consumed directly without computations. 

This is causing quite a confusion since `iconSpacing` will be a raw value like `0rem 1rem 0rem 1rem` but `iconColor` would be a prop value like `action.icon.primary.default`

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -31,6 +35,106 @@ type BaseButtonWithIconProps = BaseButtonCommonProps & {
...
+    fontSize: 100,
+    lineHeight: 'l',
+    buttonHeight: '40px',
+    iconSpacing: makeSpacingSize(2),
+    iconColor: 'action.icon.primary.default',
+    textColor: 'action.text.primary.default',
+    buttonSpacing: `${makeSpacingSize(theme.spacing[0])} ${makeSpacingSize(
+      theme.spacing[5],
+    )} ${makeSpacingSize(theme.spacing[0])} ${makeSpacingSize(theme.spacing[5])}`,
+    text: children,
```
chaitanyadeorukhkar, line 82: returning `children` as is except when the size is `xsmall` and we need it all in upper-case. We could use `text-transform` but this seems fine too imo

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -4,20 +4,20 @@ const getBaseButtonStyles = ({
   color,
   hoverColor,
   activeColor,
+  buttonHeight,
+  buttonSpacing,
 }: Omit<StyledBaseButtonProps, 'children' | 'onClick'>): string => `
-  display:flex;
+  display: flex;  
   flex-direction: row;
```
chaitanyadeorukhkar, line 11: Need this for native since it defaults to `column`

packages/blade/src/utils/makeSpacingSize/makeSpacingSize.web.ts
```diff
@@ -0,0 +1,6 @@
+const makeSpacingSize = (size: number): `${number}rem` => {
```
chaitanyadeorukhkar, line 1: This is exactly the same as `makeTypographySize`, should we club them both together and call it `makeSize`?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -1,8 +1,12 @@
 import type { ReactElement } from 'react';
+import styled from 'styled-components';
```
chaitanyadeorukhkar, line 2: This import won't fail since our storybook babel is configured to read this as `styled-components/native` but this would fail where people have not configured it. But that's okay since this is the exact same case in `BladeProvider` as well. We'll just have to ensure people configure babel correctly.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -73,17 +187,23 @@ const BaseButton = ({
...
+      buttonSpacing={buttonSpacing}
     >
-      {Icon ? <Icon size="small" color="action.icon.primary.default" /> : null}
-      <BaseText
-        lineHeight="s"
-        fontSize={100}
+      {Icon && iconPosition == 'left' ? <Icon size={iconSize} color={iconColor} /> : null}
+      <StyledBaseText
+        lineHeight={lineHeight}
+        fontSize={fontSize}
```
chaitanyadeorukhkar, line 196: TS doesn't give any hints here as to what scale represent what size in `px/rem`. This makes the dev experience difficult because now we need to open up the `typography.ts` file and check what scale is what value. Figma also doesn't help here since it shows the font size directly in pixel and not the scale we made. This issue stands true on `BaseText.stories.tsx` as well

Screenshot 2022-05-25 at 4 46 15 PM

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -31,6 +35,105 @@ type BaseButtonWithIconProps = BaseButtonCommonProps & {
...
+  theme,
+  size,
+  children,
+}: {
+  theme: Theme;
+  size: BaseButtonCommonProps['size'];
+  children?: string;
+}): ConfigProps => {
+  const props: ConfigProps = {
+    iconSize: 'medium',
```
chaitanyadeorukhkar, line 72: We're following `small`, `medium`, etc for icon size & button size but `s`, `m`, etc. for `lineHeight`. Should We change one of them? Which one?

packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -26,6 +26,7 @@ export type BaseTextProps = {
   textAlign?: 'center' | 'justify' | 'left' | 'right';
   truncateAfterLines?: number;
   className?: string;
+  style?: React.CSSProperties;
```
chaitanyadeorukhkar, line 29: Need to pass this to ensure `styled(BaseText)` works on native

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
kamleshchandnani: why do you need to explicitly have that distinction? the type suffice what we are doing right?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
kamleshchandnani: I think we are also using `small`, `medium` for Typography so maybe better to change everything to `s`, `m`, `l` we can also change everything to `small`, `medium` but that might tricky when we reach `2xlarge`, `3xlarge` which looks weird in reading compared to `2xl`,`3xl`

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -31,6 +36,103 @@ type BaseButtonWithIconProps = BaseButtonCommonProps & {
...
+    case 'large':
+      props.iconSize = 'medium';
+      props.fontSize = buttonTypographyTokens.fonts.size.large;
+      props.lineHeight = buttonTypographyTokens.lineHeight.large;
+      props.buttonHeight = '48px';
+      props.iconSpacing = makeSpacingSize(theme.spacing[2]);
+      props.buttonSpacing = `${makeSpacingSize(theme.spacing[0])} ${makeSpacingSize(
+        theme.spacing[5],
+      )} ${makeSpacingSize(theme.spacing[0])} ${makeSpacingSize(theme.spacing[5])}`;
+      break;
```
kamleshchandnani, line 98: nit: add new line

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -73,17 +188,25 @@ const BaseButton = ({
       }}
       disabled={isDisabled}
       activeColor={activeColor}
+      buttonHeight={buttonHeight}
```
kamleshchandnani, line 191:
```suggestion
      height={buttonHeight}
```
how about just `height`? since button is already in context. ideally it shall be `minHeight`

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -73,17 +188,25 @@ const BaseButton = ({
...
+      buttonSpacing={buttonSpacing}
```
kamleshchandnani, line 192:
```suggestion
      spacing={buttonSpacing}
```

spacing?

packages/blade/src/components/Button/BaseButton/StyledBaseButton.d.ts
```diff
@@ -1,9 +1,13 @@
-export type StyledBaseButtonProps = {
+import type { BaseButtonProps } from './BaseButton';
+
+export type StyledBaseButtonProps = Omit<BaseButtonProps, 'icon' | 'children'> & {
   children: string;
   onClick: () => void;
   color: string;
   hoverColor: string;
   activeColor: string;
+  buttonHeight: '48px' | '40px' | '32px' | '28px';
```
kamleshchandnani, line 9: put it under button tokens?

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
```diff
@@ -24,4 +25,51 @@ describe('<BaseButton />', () => {
...
+  it('should render button with icon with left iconPosition', () => {
+    const buttonText = 'Pay Now';
+    const { toJSON } = renderWithTheme(
+      <BaseButton iconPosition="left" icon={CreditCardIcon}>
+        {buttonText}
+      </BaseButton>,
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
+  it('should render button with icon with right iconPosition', () => {
```
kamleshchandnani, line 66: do we want to write tests for all permutation combinations? I wonder that's just too much to bloat up things. I think we discussed something similar last time. don't remember the conclusion 🤔

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
```diff
@@ -27,4 +28,53 @@ describe('<BaseButton />', () => {
...
+    const buttonText = 'Pay Now';
+    const { container } = renderWithTheme(<BaseButton size="medium">{buttonText}</BaseButton>);
+    expect(container).toMatchSnapshot();
+  });
+  it('should render small size button', () => {
+    const buttonText = 'Pay Now';
+    const { container } = renderWithTheme(<BaseButton size="small">{buttonText}</BaseButton>);
+    expect(container).toMatchSnapshot();
+  });
+  it('should render xsmall size button', () => {
```
kamleshchandnani, line 46: nit: maybe let's sort the tests? render `xsmall`, `small`, `medium` and so on?

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -0,0 +1,36 @@
+const typography = {
+  onDesktop: {
+    fonts: {
+      size: {
+        large: 200,
```
kamleshchandnani, line 5: wondering if we shall use the key from the global token here? so that we never miss this. Although TS will fail if this messes up regardless but just to keep things in sync? thoughts? I just realized that even in typography I've missed this

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -0,0 +1,36 @@
...
+        medium: 100,
+        small: 50,
+        xsmall: 50,
+      },
+    },
+    lineHeight: {
+      large: 'm', // this is 24px but we need 20px
+      medium: 'm', // this is 24px but we need 20px
```
kamleshchandnani, line 29: @saurav12 something to look into?

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -4,20 +4,20 @@ const getBaseButtonStyles = ({
...
-  height: 48px;
+  align-items: center;
+  align-self: center;
+  height: ${buttonHeight};
```
kamleshchandnani, line 14:
```suggestion
  min-height: ${buttonHeight};
```

packages/blade/src/utils/makeSpacingSize/makeSpacingSize.web.ts
kamleshchandnani: nupe. The reason is that we use `rems` in typography but in layout we need to use `px`. [Refer RFC](https://github.com/razorpay/blade/blob/master/rfcs/2021-02-19-units-for-typography-and-layout.md#what-will-work-for-us)

kamleshchandnani: Also we can just rename this to `makeSpace`. Size is something that will come separately and can be something as `makeSize` and will affect things like `height`, `width`, etc family

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -31,6 +36,103 @@ type BaseButtonWithIconProps = BaseButtonCommonProps & {
...
+const getProps = ({
+  theme,
+  size,
+  children,
+  buttonTypographyTokens,
+}: {
+  theme: Theme;
+  size: BaseButtonCommonProps['size'];
+  children?: string;
+  buttonTypographyTokens: typeof typography.onDesktop | typeof typography.onMobile;
```
kamleshchandnani, line 72: put the typings under tokens file itself and then use it here?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: Not for the consumers. We're using `BaseText` inside `BaseButton` which needs to pass down "extra styles" (padding) for `BaseText` when there is an icon.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: It does. It's just annoying to have `iconColor` as a dot notation string vs `buttonColor`(see the next PR) as a direct string color value

packages/blade/src/utils/makeSpacingSize/makeSpacingSize.web.ts
chaitanyadeorukhkar: Oh got it. I'll rename this and return `px`

chaitanyadeorukhkar: Also, since `makeSpace` would be the same for native & web I'm keeping only 1 implementation

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: `StyledBaseText` is `BaseText` wrapped with styled. But even if its not wrapped with style, the same issue persists

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: After every `break`? Won't that look odd?

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
chaitanyadeorukhkar: Yes, had to discuss this with @saurav12. I'll bring it up with him during standup

packages/blade/src/components/Button/BaseButton/StyledBaseButton.d.ts
chaitanyadeorukhkar: Will do this in the #487 PR. Would cause a lot of rebasing headache otherwise.

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
chaitanyadeorukhkar: Don't remember either 😅 But if we don't ensure tests for at least the commonly used combinations, it would be open to breakage

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
chaitanyadeorukhkar: Check this commit. See if this makes a bit more sense [`368d21f` (#487)](https://github.com/razorpay/blade/pull/487/commits/368d21f4b574230b99dfdb82cfc6fc178cac8d02)

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -31,6 +36,103 @@ type BaseButtonWithIconProps = BaseButtonCommonProps & {
...
+      break;
+    case 'xsmall':
+      props.iconSize = 'xsmall';
+      props.fontSize = buttonTypographyTokens.fonts.size.xsmall;
+      props.lineHeight = buttonTypographyTokens.lineHeight.xsmall;
+      props.minHeight = '28px';
+      props.iconSpacing = makeSpace(theme.spacing[1]);
+      props.spacing = `${makeSpace(theme.spacing[0])} ${makeSpace(theme.spacing[2])} ${makeSpace(
+        theme.spacing[0],
+      )} ${makeSpace(theme.spacing[2])}`;
```
anuraghazra, line 127: Extract it to a function? 

```ts
  const getSpacing = (one: keyof Theme['spacing'], two: keyof Theme['spacing']): string => {
    return `${makeSpace(theme.spacing[one])} ${makeSpace(theme.spacing[two])} ${makeSpace(
      theme.spacing[one],
    )} ${makeSpace(theme.spacing[two])}`;
  };
```

would make things much cleaner.

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
anuraghazra: That's too many snapshots. :( 

We can reduce some of them. 

1. For icons we should instead do testing by checking if icon exists in the document. 

Eg: 

```diff
+    expect(getByTestId('icon-left')).toBeInTheDocument()
+    expect(getByTestId('icon-right')).not.toBeInTheDocument()
-    expect(container).toMatchSnapshot();

+    expect(getByTestId('icon-right')).toBeInTheDocument()
+    expect(getByTestId('icon-left')).not.toBeInTheDocument()
-    expect(container).toMatchSnapshot();
```

2. for button type prop, just assert the type prop, don't need snapshot

```ts
    expect(getByRole('button')).toHaveAttribute('type', 'reset');
```

packages/blade/src/components/Button/BaseButton/StyledBaseButton.d.ts
chaitanyadeorukhkar: Done in #487

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: `sm`, `md`, `lg` looks weird. We'd then need `2xlg`, `3xlg`?. ` s`, `m`, `l` still gives you an idea of t-shirt sizing. My issue with all of these are they are not verbose enough. We don't have a character limit and `small`, `medium`, `large` looks pretty descriptive. Its like naming a variable `i` vs `index`. Closer it is to the English language, the more readable it is

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -44,25 +146,38 @@ const BaseButton = ({
   type = 'button',
   children,
 }: BaseButtonProps): ReactElement => {
-  const { theme } = useTheme();
+  const { theme, platform } = useTheme();
+  const buttonTypographyTokens = typography[platform];
```
nashcheez, line 150: Seeing a eslint `no-unsafe-member-access` warning here - `Computed name [platform] resolves to an any value.`

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
nashcheez: I think `xs, sm, md, lg, xl, 2xl` are standards today and widely acceptable because of [Bootstrap's grid system](https://getbootstrap.com/docs/4.0/layout/grid/) convention and [Tailwind](https://tailwindcss.com/docs/responsive-design). The t-shirt sizing logic makes sense, but yes they are not the most verbose nor readable. `small, medium & large` are readable but they are big prop names for what they do.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: That would be an overkill imo. Spacing isn't always a 2 input 4 output function. It could be 1, or 2, or 4 output function. Generalizing it would mean that we sometimes use it and sometimes don't. `makeSpace` is already an abstraction. I don't think wrapping it further makes sense. Although I did use the shorthand now and made it only 2 units instead of 4

packages/blade/src/utils/makeSpace/makeSpace.ts
```diff
@@ -0,0 +1,5 @@
+const makeSpace = <T extends number>(size: T): `${T}px` => {
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: `makeSpace` seems better imo. Aligns well with other things like `makeTypographySize`, `makeMotionUnit`

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
anuraghazra: +1 `xs, sm, md, lg, xl, 2lx` they are very standard names,  

Used by 
- [Chakra](https://chakra-ui.com/docs/components/form/button)
- [Carbon](https://carbondesignsystem.com/components/button/usage/)

The pattern which I'm seeing is component libraries which have only 3 variants for size mostly opt for `small`, `medium`, `large` (for eg: materialui, vercel) but if the variants are more than that xs, sm , md, lg.

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
anuraghazra: > Would want to have snapshot test for icon because padding of text changes based on icon presence and position.

I feel like thats a bit overkill just for testing padding. Rather we can test individually for the padding by accessing the computed styles.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
anuraghazra: Makes sense but I'm not saying to generalise globally, just locally it will clean things up in the switch statement. 
Though you can make a call here as you find it fitting.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: Damn. Huge bummer. This would've been very helpful

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.native.test.tsx
chaitanyadeorukhkar: What's the harm doing a snapshot test? Lets relook at this if our tests feel bloated? Right now it seems fine

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
kamleshchandnani: > Unions can't have jsdoc descriptions. it's keyof nah..

@anuraghazra didn't get this? 🤔

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
kamleshchandnani: it's getting very confusing to context switch and build context as to what is what in which PR 😭

chaitanyadeorukhkar: We can do everything in a single PR but it would be too huge to review then

PR #485, opened 2022-05-27
feat: add fullWidth support to BaseButton (Button [3])

chaitanyadeorukhkar:
**Things to review**
- `fullWidth` prop
- `fullWidth` behaviour on web & native
- Tests for `fullWidth`


packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -6,12 +6,15 @@ const getBaseButtonStyles = ({
   activeColor,
   buttonHeight,
   buttonSpacing,
+  isFullWidth,
```
kamleshchandnani, line 9: so so beautiful. Just pass the prop down to one single function, and things work out of the box on both platforms. ❤️. Any addition/deletion is just central

PR #487, opened 2022-05-30
feat: add onClick, isDisabled, & type support (Button [4])

chaitanyadeorukhkar:
**Things to review**
- `onClick` prop
- `isDisabled` prop
- `type` prop

**Things to not review**
- variant-specific stylings (all stylings are based on primary for now including `isDisabled` styling. I'll incorporate handling of disabled styling when we implement variants


packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -58,31 +58,75 @@ type ConfigProps = {
   textColor: BaseTextProps['color']; // prop
   buttonSpacing: string; // raw value
   text?: string;
+  buttonColor: string; // raw value
```
kamleshchandnani, line 61: we have textcolor on the top, what is buttoncolor?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -58,31 +58,75 @@ type ConfigProps = {
...
+  buttonBorderColor: string; // raw value
```
kamleshchandnani, line 62: nit: let's remove `button` as a context in front of these props?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -58,31 +58,75 @@ type ConfigProps = {
...
 }: {
+  buttonTypographyTokens: typeof typography.onDesktop | typeof typography.onMobile;
+  children?: string;
+  isDisabled: boolean;
   theme: Theme;
   size: BaseButtonCommonProps['size'];
-  children?: string;
-  buttonTypographyTokens: typeof typography.onDesktop | typeof typography.onMobile;
 }): ConfigProps => {
   const props: ConfigProps = {
```
kamleshchandnani, line 85: nit: `ConfigProps` sounds very generic. Let's rename this to something more specific? Maybe `BaseButtonStyleProps`?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -58,31 +58,75 @@ type ConfigProps = {
...
     iconSpacing: makeSpacingSize(theme.spacing[2]),
-    iconColor: 'action.icon.primary.default',
-    textColor: 'action.text.primary.default',
+    iconColor: isDisabled ? 'action.icon.primary.disabled' : 'action.icon.primary.default',
+    textColor: isDisabled ? 'action.text.primary.disabled' : 'action.text.primary.default',
     buttonSpacing: `${makeSpacingSize(theme.spacing[0])} ${makeSpacingSize(
       theme.spacing[5],
     )} ${makeSpacingSize(theme.spacing[0])} ${makeSpacingSize(theme.spacing[5])}`,
     text: children?.trim(),
+    buttonColor: getIn(
```
kamleshchandnani, line 97: since the only differentiating prop is `disabled` here? then maybe abstract it out at a top-level and put it under the `if` condition and compute the disabled props, then just spread that object over here

packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
```diff
@@ -82,4 +83,30 @@ describe('<BaseButton />', () => {
     const { container } = renderWithTheme(<BaseButton isFullWidth={true}>{buttonText}</BaseButton>);
     expect(container).toMatchSnapshot();
   });
+  it('should render disabled button', () => {
+    const buttonText = 'Pay Now';
+    const { container, getByRole } = renderWithTheme(
+      <BaseButton isDisabled={true}>{buttonText}</BaseButton>,
+    );
+    expect(container).toMatchSnapshot();
+    expect(getByRole('button')).toBeDisabled();
```
kamleshchandnani, line 92: amazing 🔥

packages/blade/src/components/Button/BaseButton/__tests__/__snapshots__/BaseButton.web.test.tsx.snap
```diff
@@ -1,10 +1,53 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`<BaseButton /> should render button of type "reset" 1`] = `
+<div>
+  <button
+    class="StyledBaseButtonweb__StyledBaseButton-sc-26bt38-0 cmQBah"
+    color="hsla(213, 89%, 56%, 1)"
+    type="reset"
+  >
+    <div
```
kamleshchandnani, line 10: why is this rendering a `div` here? 🤔

packages/blade/src/components/Button/BaseButton/getBaseButtonStyles.ts
```diff
@@ -15,20 +22,27 @@ const getBaseButtonStyles = ({
...
   background-color: ${color};
-  border: 2px solid ${color};
+  border: 2px solid ${borderColor};
   color: white;
   border-radius: 2px;
   padding: ${buttonSpacing};
   &:hover {
     background-color: ${hoverColor};
-    border: 2px solid ${hoverColor};
+    border: 2px solid ${hoverBorderColor};
```
kamleshchandnani, line 33: border token use krle yaha pe? or it will come later? 🙈

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -58,31 +58,75 @@ type ConfigProps = {
...
+    ),
+    focusColor: getIn(
+      theme.colors,
+      isDisabled ? 'action.background.primary.disabled' : 'action.background.primary.focus',
+    ),
+    focusBorderColor: getIn(
+      theme.colors,
+      isDisabled ? 'action.border.primary.disabled' : 'action.border.primary.focus',
+    ),
+    focusRingColor: getIn(theme.colors, 'brand.primary.400'),
```
kamleshchandnani, line 129: @saurav12 shall we add focus ring color semantic token? or we shall keep using brand color?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -58,38 +58,82 @@ type BaseButtonStyleProps = {
...
+      isDisabled ? 'action.border.primary.disabled' : 'action.border.primary.default',
+    ),
+    hoverColor: getIn(
+      theme.colors,
+      isDisabled ? 'action.background.primary.disabled' : 'action.background.primary.hover',
+    ),
+    hoverBorderColor: getIn(
+      theme.colors,
+      isDisabled ? 'action.border.primary.disabled' : 'action.border.primary.hover',
+    ),
```
anuraghazra, line 112: I have an urge to refactor this to a function 😢 but not sure if that would be a good idea. 

```ts
const props = {
  color: getVariant(isDisabled, 'action.background.primary', 'default'),
  borderColor: getVariant(isDisabled, 'action.border.primary', 'default'),
  hoverColor: getVariant(isDisabled, 'action.background.primary', 'hover'),
  hoverBorderColor: getVariant(isDisabled, 'action.border.primary', 'hover')
}
```

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: I've cleaned it up a bit, see if it makes sense

packages/blade/src/components/Button/BaseButton/__tests__/__snapshots__/BaseButton.web.test.tsx.snap
chaitanyadeorukhkar: Because there's a text inside the button?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
anuraghazra: Oh this looks better https://github.com/razorpay/blade/pull/487/commits/731f8d7a034db03161757d2874c9169110c9a059

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -49,32 +51,43 @@ const StyledBaseText = styled(BaseText)(
...
   text?: string;
+  color: string;
+  borderColor: string;
+  hoverColor: string;
+  hoverBorderColor: string;
+  activeColor: string;
+  activeBorderColor: string;
+  focusColor: string;
+  focusBorderColor: string;
+  focusRingColor: string;
```
nashcheez, line 71: What is the difference between `focusBorderColor` and `focusRingColor`? 🤔

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -1,4 +1,18 @@
-const typography = {
+import type { ThemeTokens } from '../../../tokens/theme/theme.d';
+import type { TypographyPlatforms } from '../../../tokens/global/typography';
+
+export type ButtonMinHeight = '48px' | '40px' | '32px' | '28px';
```
nashcheez, line 4: Nice!

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -58,38 +58,82 @@ type BaseButtonStyleProps = {
...
   text?: string;
+  color: string; // raw value
+  borderColor: string; // raw value
+  hoverColor: string; // raw value
+  hoverBorderColor: string; // raw value
+  activeColor: string; // raw value
+  activeBorderColor: string; // raw value
+  focusColor: string; // raw value
+  focusBorderColor: string; // raw value
+  focusRingColor: string; // raw value
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Removed these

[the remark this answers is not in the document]
chaitanyadeorukhkar: Removed these

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
anuraghazra: Why do we need both?

nashcheez: Is the `box-shadow` used in focus different from the one that might be used in hover (if we decide to keep one)? If not, then we can consolidate it in a single var.  Also, does it make sense to have this one called `focusShadowColor` instead of `focusRingColor` for more clarity?

chaitanyadeorukhkar: >Why do we need both?

@anuraghazra Border always exists even on default state. We can't treat border as a focus ring. See the tertiary variant design for example where you have border along with a focus ring

chaitanyadeorukhkar: @nashcheez For now we're not keeping box-shadow on hover. We would have to remove the hover box-shadow on focus to show the focus ring instead. Could be a problem maybe if design wants shadow along with focus ring. But we can discuss that with design separately. 

Right now, the color is acting as a `focusRing` so lets call it that. If we do want shadows later on we can rename this. But as of now, it's better to keep the naming as per its usage.

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -1,4 +1,18 @@
...
+
+export type ButtonTypography = {
+  [key in TypographyPlatforms]: {
```
anuraghazra, line 7: Cool. 

```suggestion
  [Key in TypographyPlatforms]: {
```

PR #488, opened 2022-05-31
fix(typography): added accessibilityRole in heading and title component

anuraghazra:
- Improved the accessibility of heading and title component by adding accessibilityRole prop in BaseText

before: 

SR announces "Inaccessible heading example"

https://user-images.githubusercontent.com/35374649/171153739-fed9fd82-6b70-43ce-b07f-3e4dbf2d611f.mov

After: 

SR announces "Accessible heading example, heading"

https://user-images.githubusercontent.com/35374649/171153818-729fbde5-f9e7-4b0d-a047-6f62ded3a8fa.mov

These headers will also be available as landmark signposts.

### DIFF
--- a/packages/blade/src/components/Typography/BaseText/BaseText.tsx
+++ b/packages/blade/src/components/Typography/BaseText/BaseText.tsx
@@ -27,6 +27,10 @@ export type BaseTextProps = {
   truncateAfterLines?: number;
   className?: string;
   children: React.ReactNode;
+  /**
+   * @platform native
+   */
+  accessibilityRole?: string;
 };
 
 const BaseText = ({
@@ -42,6 +46,7 @@ const BaseText = ({
   children,
   truncateAfterLines,
   className,
+  accessibilityRole,
 }: BaseTextProps): ReactElement => {
   const { theme } = useTheme();
   const textColor = getIn(theme.colors, color);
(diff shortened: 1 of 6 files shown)


kamleshchandnani:
@anuraghazra the idea was to add a111y attributes with our new a11y layer.

anuraghazra:
> @anuraghazra the idea was to add a111y attributes with our new a11y layer.

We can use that layer but in this case it's unnecessary, that a11y layer is for compatibility, in this case we are only adding accessibilityRole for native only. for web we don't need it. 

> plus, that PR is not ready yet.

packages/blade/src/components/Typography/BaseText/BaseText.tsx
```diff
@@ -27,6 +27,10 @@ export type BaseTextProps = {
   truncateAfterLines?: number;
   className?: string;
   children: React.ReactNode;
+  /**
+   * @platform native
+   */
+  accessibilityRole?: string;
```
chaitanyadeorukhkar, line 33: I thought we'll use the mapA11yProps function to do this on all components? Why are we doing this directly?

packages/blade/src/components/Typography/BaseText/StyledBaseText.d.ts
```diff
@@ -11,4 +11,5 @@ export type StyledBaseTextProps = {
   as?: 'code' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p' | 'span';
   textAlign?: 'center' | 'justify' | 'left' | 'right';
   numberOfLines?: number;
+  accessibilityRole?: string;
```
chaitanyadeorukhkar, line 14: Why string? shouldn't it have specific values?

PR #489, opened 2022-06-01
test: enable snapshot testing to capture component styling

chaitanyadeorukhkar:
Take snapshots along with styling to catch exact snapshot changes instead of vague class name changes


anuraghazra (approved):
Looks good! we also want this to properly test the VisuallyHidden component since that depends on CSS.

kamleshchandnani (approved):

PR #491, opened 2022-06-01
feat(blade): add VisuallyHidden component

anuraghazra:
- Adds visually hidden component

RFC Reference - [Hidden Content](https://github.com/razorpay/blade/blob/anu/a11y-rfc/rfcs/2022-04-09-accessibility.md#visually-hidden-)

### DIFF
--- a/.changeset/twenty-grapes-kneel.md
+++ b/.changeset/twenty-grapes-kneel.md
@@ -1,0 +1,9 @@
+---
+"@razorpay/blade": minor
+---
+
+feat(blade): add VisuallyHidden component
+
+This component is used specifically when you want to hide certain things visually for people who are not visually impaired but also want to make your content is accessible to visually impaired people via assistive technologies.
+
+You can play around with it on [Storybook](https://master--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/docs/components-accessibility-visuallyhidden--visually-hidden)


packages/blade/src/components/VisuallyHidden/VisuallyHidden.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+import BaseText from '../Typography/BaseText';
+import type { VisuallyHiddenProps } from './VisuallyHidden';
+import VisuallyHiddenComponent from './';
+
+const Page = (): ReactElement => {
+  return (
+    <>
+      <StorybookTitle />
+      <Subtitle>
+        The VisuallyHidden component makes content hidden from sighted users but available for
```
nashcheez, line 22:
```suggestion
        The VisuallyHidden component hides its content visually while keeping them visible to screen readers and other assistive technologies.
```

packages/blade/src/components/VisuallyHidden/VisuallyHidden.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+    docs: {
+      page: () => <Page />,
+    },
+  },
+};
+
+const VisuallyHiddenTemplate: ComponentStory<typeof VisuallyHiddenComponent> = () => {
+  return (
+    <button>
+      <EyeIcon color="action.icon.link.active" size="large" />
```
nashcheez, line 52: Should we use the eye icon as an example? It would probably be better to give a real life example where we could have a button or a nav element with the text "Skip to content" which when entered on leads to a content.
For e.g. check the [bbc website](https://www.bbc.com/) or even JIRA

packages/blade/src/components/VisuallyHidden/VisuallyHidden.stories.tsx
anuraghazra: Changed

packages/blade/src/components/VisuallyHidden/VisuallyHidden.stories.tsx
anuraghazra: I guess the example isn't quite optimal, let me think of something better.  🤔 

> be better to give a real life example where we could have a button or a nav element with the text "Skip to content" 
We can do that 🤔 but we will also be having a separate <SkipNav /> component don't you think the storybook examples will get a bit redundant? 

Button with an hidden icon + visually hidden text is also a valid usecase tho.

anuraghazra: I tried out this example, CustomCheckbox with hidden input 😅 feels too complex tho, Let's just have Icon button only. I think it conveys the usage. 

```ts
const VisuallyHiddenTemplate: ComponentStory<typeof VisuallyHiddenComponent> = () => {
  const [checked, setChecked] = React.useState(false);
  return (
    <span>
      <VisuallyHiddenComponent>
        <input
          checked={checked}
          onChange={(): void => setChecked(!checked)}
          type="checkbox"
          id="c1"
        />
      </VisuallyHiddenComponent>
      <span style={{ outline: '1px solid red' }}>
        {checked ? '✅ ' : '❎ '}
        <label htmlFor="c1">Custom checkbox with hidden input</label>
      </span>
    </span>
  );
};
```

packages/blade/src/components/VisuallyHidden/__tests__/VisuallyHidden.native.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+describe('<VisuallyHidden />', () => {
+  // TODO: Enable css testing
+  it('should render VisuallyHidden content', () => {
+    const displayText = 'Text only visible to screen readers';
+    const { toJSON, getByText } = renderWithTheme(
+      <VisuallyHidden>
+        <Text>{displayText}</Text>
+      </VisuallyHidden>,
+    );
+    expect(getByText(displayText)).toBeTruthy();
```
kamleshchandnani, line 15: how is this returning truthy? or is that expected?

kamleshchandnani (changes requested):

packages/blade/src/components/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import StyledVisuallyHidden from './StyledVisuallyHidden';
+
+export type VisuallyHiddenProps = {
+  children: React.ReactNode;
+};
+
+const VisuallyHidden = ({ children }: VisuallyHiddenProps): JSX.Element => {
```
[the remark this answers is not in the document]
anuraghazra: Yeah that works. 


:p Need a box 😉 
<Box as={{ web: 'div', native: 'View' }} styles={{...}} />

packages/blade/src/components/VisuallyHidden/VisuallyHidden.stories.tsx
anuraghazra: Done.

packages/blade/src/components/VisuallyHidden/VisuallyHidden.stories.tsx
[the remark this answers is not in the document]
anuraghazra: Added the instruction in the example itself & in docs added a link to the hidden content RFC is people want to deep dive into it.

anuraghazra:
@kamleshchandnani This is ready to be merged.

packages/blade/src/components/VisuallyHidden/__tests__/VisuallyHidden.native.test.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Text } from 'react-native';
+import renderWithTheme from '../../../_helpers/testing/renderWithTheme.native';
+import VisuallyHidden from '../VisuallyHidden.native';
+
+describe('<VisuallyHidden />', () => {
+  // TODO: Enable css testing
```
kamleshchandnani, line 7:
```suggestion
```

packages/blade/src/components/VisuallyHidden/VisuallyHidden.d.ts
```diff
@@ -0,0 +1,6 @@
+export { default } from './VisuallyHidden.web';
+export * from './ScreenReaderStyles';
+
+export type VisuallyHiddenProps = {
+  children: React.ReactNode;
```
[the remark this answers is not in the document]
anuraghazra: There's nothing to support explicitly about `null | undefined`, they are just valid react nodes considered by React itself, I don't think we need to worry about that causing any issues, React.ReactNode is pretty standard. Should be fine.

packages/blade/src/components/VisuallyHidden/__tests__/VisuallyHidden.web.test.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import renderWithTheme from '../../../_helpers/testing/renderWithTheme.web';
+import VisuallyHidden from '../VisuallyHidden.web';
+
+describe('<VisuallyHidden />', () => {
+  // TODO: Enable css testing
```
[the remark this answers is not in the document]
anuraghazra: Yep we added css testing, so can remove this.

packages/blade/src/components/VisuallyHidden/ScreenReaderStyles.ts
```diff
@@ -0,0 +1,17 @@
+import type { CSSObject } from 'styled-components';
+
+const screenReaderStyles: CSSObject = {
+  border: 0,
+  clip: 'rect(0 0 0 0)',
+  clipPath: 'inset(50%)',
+  height: 1,
```
chaitanyadeorukhkar, line 7: 1 what? What units are they mapping to in web & native?

packages/blade/src/components/VisuallyHidden/VisuallyHidden.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+        See Hidden Content RFC
+      </Link>
+      <br />
+      <br />
+      <StorybookTitle>Usage</StorybookTitle>
+      <Highlight language="tsx">{`import { VisuallyHidden } from '@razorpay/blade/components' \nimport type { VisuallyHiddenProps } from '@razorpay/blade/components'`}</Highlight>
+      <StorybookTitle>Example</StorybookTitle>
+      <Primary />
+      <StorybookTitle>Properties</StorybookTitle>
+      <ArgsTable story={PRIMARY_STORY} />
```
chaitanyadeorukhkar, line 38: If there are no props, lets remove this section?
```suggestion
```

chaitanyadeorukhkar: Actually it does accept `children` as a prop. Why is that not showing up in the story?

packages/blade/src/components/VisuallyHidden/VisuallyHidden.d.ts
```diff
@@ -0,0 +1,6 @@
+export { default } from './VisuallyHidden.web';
```
chaitanyadeorukhkar, line 1: Can we have a structure like this?

We can get rid of having to create components twice, having to create `VisuallyHidden.d.ts`, and this matches our component creation pattern

Screenshot 2022-06-15 at 3 37 42 PM

chaitanyadeorukhkar: Attaching more details of each file
Screenshot 2022-06-15 at 3 39 23 PM

chaitanyadeorukhkar: Screenshot 2022-06-15 at 3 40 07 PM

chaitanyadeorukhkar: Screenshot 2022-06-15 at 3 40 38 PM

chaitanyadeorukhkar: Screenshot 2022-06-15 at 3 40 52 PM

packages/blade/src/components/VisuallyHidden/VisuallyHidden.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const VisuallyHiddenTemplate: ComponentStory<typeof VisuallyHiddenComponent> = () => {
+  return (
+    <>
+      <Text>
+        This is an example for VisuallyHidden component, enable voiceover and focus on the checkbox
+        to hear it's invisible label.
+      </Text>
+      <VisuallyHiddenComponent>
+        <label htmlFor="darkmode">Toggle dark mode</label>
```
chaitanyadeorukhkar, line 63: This is crashing in native. Can you check?

anuraghazra: Yeah if course it would we don't have cross platform Label or Input component. I'll have to create two files for this example then.

packages/blade/src/components/VisuallyHidden/VisuallyHidden.d.ts
anuraghazra: That's what I did earlier, changed it after this comment. https://github.com/razorpay/blade/pull/491#discussion_r893313104

cc @kamleshchandnani which one should we use?

packages/blade/src/components/VisuallyHidden/ScreenReaderStyles.ts
anuraghazra: It would be 1px in web i know. We can be explicit.

packages/blade/src/components/VisuallyHidden/VisuallyHidden.stories.tsx
anuraghazra: Because I haven't passed the children in.

packages/blade/src/components/VisuallyHidden/VisuallyHidden.d.ts
kamleshchandnani: @chaitanyadeorukhkar understand your point of following pattern consistency but this is just overwhelming hence if you remember we had decided that we'll take this pattern case by case basis as to where the abstraction lies. so i had commented last time to reduce that and have only VisuallyHidden<platform>.tsx

packages/blade/src/components/VisuallyHidden/__tests__/VisuallyHidden.native.test.tsx
kamleshchandnani: yeah correct, the test is partially implemented then because what we need to also make sure is that it's visually hidden. can we add an assertion for that as well? maybe [this might help](https://github.com/testing-library/jest-dom#tobevisible)?

`expect(getByText(displayText)).not.toBeVisible();`

kamleshchandnani: while snapshot does that but it's just implying the CSS not the expectation

packages/blade/src/components/VisuallyHidden/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './VisuallyHidden';
+export * from './VisuallyHidden.d';
+export { default } from './VisuallyHidden';
```
kamleshchandnani, line 3: can you also add it to exports on `src/components/index.ts`?

anuraghazra: Right

packages/blade/src/components/VisuallyHidden/VisuallyHidden.stories.tsx
kamleshchandnani: I think we can proceed for now it's just a story and we don't publish native storybook. Just to be confident please just test it once on emulator if things are working fine and won't break in consumer's codebase

packages/blade/src/components/VisuallyHidden/VisuallyHidden.d.ts
chaitanyadeorukhkar: Okay, lets keep it this way here then

kamleshchandnani (approved):

PR #497, opened 2022-06-07
feat(a11y): added skip nav component

anuraghazra:
- Added <SkipNavLink \/> & <SkipNavContent \/> components

See RFC - [Skip Navigation](https://github.com/razorpay/blade/blob/anu/a11y-rfc/rfcs/2022-04-09-accessibility.md#skip-navigations)


package.json
```diff
@@ -22,6 +22,8 @@
     "release": "changeset publish",
     "preinstall": "git config --global url.\"https://github.com/\".insteadOf git://github.com/",
     "build": "lerna run --scope @razorpay/blade build",
+    "test:web": "yarn workspace @razorpay/blade test:react",
+    "test:native": "yarn workspace @razorpay/blade test:react-native",
     "start:web": "yarn workspace @razorpay/blade react:storybook",
     "start:ios": "yarn workspace @razorpay/blade react-native:storybook:ios",
     "start:android": "yarn workspace @razorpay/blade react-native:storybook:android"
```
anuraghazra, line 29: I'm sick of writing `yarn workspace @razorpay/blade` every time and then remember all the permutations of 

```ts
"test:react:ios" | "test:react:android" | "test:react:start" | "test:react:build" | "test:react-native:ios" | "test:react-native:android" | "test:react-native:start" | "test:react-native:build"
```

packages/blade/src/components/SkipNav/SkipNav.web.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import type { CSSObject } from 'styled-components';
+import styled from 'styled-components';
+import BaseText from '../Typography/BaseText';
+
+export const fallbackId = 'blade-skip-nav';
```
kamleshchandnani, line 6: why to export this?

packages/blade/src/components/SkipNav/SkipNav.web.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <StyledLink href={`#${id}`}>
+      <BaseText color="action.text.link.default" lineHeight="s">
+        {children}
+      </BaseText>
+    </StyledLink>
+  );
+};
+
+type SkipNavContentProps = {
+  id?: string;
```
kamleshchandnani, line 56: shouldn't id be required prop?

packages/blade/src/components/SkipNav/SkipNav.web.tsx
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const SkipNavContent = ({ id = fallbackId }: SkipNavContentProps): JSX.Element => {
```
kamleshchandnani, line 59: this id and link id has to be same right? else this won't work?

package.json
kamleshchandnani: but why don't you make your current working directory as `packages/blade`?

kamleshchandnani: also install https://fig.io/ and you won't have to remember all the P&C you'll get intellisense in your teminal

packages/blade/src/components/SkipNav/SkipNav.web.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return <div tabIndex={-1} id={id} style={{ outline: 0 }} data-testid="skipnav-content" />;
```
kamleshchandnani, line 60: probably the high time to introduce [this helper](https://github.com/razorpay/blade/blob/master/packages/blade-old/src/_helpers/automation-attributes.web.js) which we had in blade old

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+describe('<SkipNav />', () => {
+  it('<SkipNavLink /> should render correctly', () => {
+    const { getByRole } = renderWithTheme(<SkipNavLink />);
+
+    expect(getByRole('link')).toHaveTextContent('Skip to content');
+  });
+
+  it('<SkipNavLink /> should render correctly with children', () => {
+    const { getByRole } = renderWithTheme(<SkipNavLink children="Skip" />);
```
kamleshchandnani, line 14: why children as a prop? and not `<SkipNavLink>Skip</SkipNavLink>`?

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    expect(getByRole('link')).toHaveTextContent('Skip');
+  });
+
+  it('<SkipNavContent /> should render correctly', () => {
+    const { getByTestId } = renderWithTheme(<SkipNavContent />);
+
+    expect(getByTestId('skipnav-content')).toHaveAttribute('tabindex', '-1');
```
kamleshchandnani, line 22: will this by default have `tabindex: -1` ?

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+  });
+
```
kamleshchandnani, line 24: we shall also add at least 1 negative test case? for eg: what happens when the id mismatches the link and content.

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+          <SkipNavContent id="custom-id" />
+          <button>main content</button>
+        </main>
+      </div>,
+    );
+
+    // TODO: move to ariakit-utils after this is fixed:
+    // https://github.com/ariakit/ariakit/issues/1483
+    await user.tab();
+    expect(getByRole('link')).toHaveFocus();
```
kamleshchandnani, line 45: let's add an assertion for the text content too? because we shall also be sure right that we have focus on the correct link

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    await user.tab();
+    expect(getByRole('button', { name: 'one' })).toHaveFocus();
+
+    await user.tab({ shift: true });
+    expect(getByRole('link')).toHaveFocus();
+
+    await user.click(getByRole('link'));
+    expect(window.location.hash).toBe('#custom-id');
+
+    // Can't test for focus because jsdom doesn't move focus to fragment anchor elements
```
kamleshchandnani, line 56: isn't there any other jest util that does this?

packages/blade/src/components/SkipNav/index.ts
```diff
@@ -0,0 +1 @@
+export * from './SkipNav.web';
```
kamleshchandnani, line 1: put this export under `src/components/index.ts` too so that this will be available for consumers too

kamleshchandnani (changes requested):

package.json
anuraghazra: Why have monorepo in that case? 🤔 If i'm going to CD into it anyways. Like won't installing deps without specifying the workspace mess things up?

packages/blade/src/components/SkipNav/SkipNav.web.tsx
```diff
@@ -0,0 +1,63 @@
...
+  top: theme.spacing[6],
+  left: theme.spacing[6],
+  color: theme.colors.action.text.link.default,
+  ':focus': {
+    padding: theme.spacing[2],
+    clip: 'auto',
+    clipPath: 'unset',
+    width: 'auto',
+    height: 'auto',
+  },
```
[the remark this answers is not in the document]
anuraghazra: The only visibly important part here is the color, which I used the link.default.

packages/blade/src/components/SkipNav/SkipNav.web.tsx
anuraghazra: Removed.

packages/blade/src/components/SkipNav/SkipNav.web.tsx
anuraghazra: No because generally in most cases you have one <SkipNav /> by default in a page, so requiring to add an ID for it is a suboptimal DX. Instead we fallback to `fallbackId`

packages/blade/src/components/SkipNav/SkipNav.web.tsx
anuraghazra: Yes.

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
anuraghazra: Both are same thing tho. children is a prop too. I just like the self closing jsx for simple components while testing.

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
anuraghazra: Yes. TabIndex -1 so that the element isn't focusable through keyboard but can be focused by other means (mouse, .focus etc)

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
anuraghazra: Yeah but currently we can't test that, because of this https://github.com/razorpay/blade/pull/497#discussion_r893386542

packages/blade/src/components/SkipNav/index.ts
anuraghazra: Right.

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
anuraghazra: No I don't think we can solve it from user end. I might open an bug report in jsdom. 

Basically fragment anchors don't focus the element it's referencing.

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
anuraghazra: I'm already doing text content assertion in above two test cases. Don't think it's necessary again. 
We are sure the focus is on the correct link, since we only have one link in the example JSX inside the renderWithTheme

package.json
```diff
@@ -21,7 +21,12 @@
     "lint": "eslint --ext js,jsx,ts,tsx packages",
     "release": "changeset publish",
     "preinstall": "git config --global url.\"https://github.com/\".insteadOf git://github.com/",
-    "build": "lerna run --scope @razorpay/blade build"
+    "build": "lerna run --scope @razorpay/blade build",
+    "test:web": "yarn workspace @razorpay/blade test:react",
+    "test:native": "yarn workspace @razorpay/blade test:react-native",
+    "start:web": "yarn workspace @razorpay/blade react:storybook",
+    "start:ios": "yarn workspace @razorpay/blade react-native:storybook:ios",
+    "start:android": "yarn workspace @razorpay/blade react-native:storybook:android"
```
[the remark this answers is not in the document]
anuraghazra: So I can execute them from the blade root folder. without CD into any pacakge or doing yarn workspace command.

package.json
kamleshchandnani: it's a monorepo since we wanted to have `old-blade` in this repo itself. it's just that we are sharing the eslint, tsconfig and few other things via this setup(debt we want to remove). and CD you can just do once. Also installing dependencies won't create mess since it won't let you install dependencies without mentioning the workspace name(it'll just fail, you can try out)

packages/blade/src/components/SkipNav/SkipNav.native.tsx
```diff
@@ -0,0 +1,9 @@
+const SkipNavLink = (): never => {
+  throw new Error('[Blade: SkipNav]: SkipNavContent is not available on React Native');
```
kamleshchandnani, line 2:
```suggestion
  throw new Error('[Blade: SkipNav]: SkipNavLink is not available on React Native');
```

packages/blade/src/components/SkipNav/SkipNav.web.tsx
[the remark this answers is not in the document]
kamleshchandnani: how about the background color?

packages/blade/src/components/SkipNav/SkipNav.web.tsx
kamleshchandnani: >in most cases you have one by default in a page

didn't understand this part?

packages/blade/src/components/SkipNav/SkipNav.web.tsx
kamleshchandnani: @anuraghazra 👆

packages/blade/src/components/SkipNav/SkipNav.web.tsx
kamleshchandnani: since we have default fallback id, imagine I've 2 skipnavcontent and 2 skipnav link how will the navigation work since the id is being set to default fallback

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
kamleshchandnani: I see, but then shouldn't we have a test to add maybe 2 SkipNav components like you have in the storybook? that will give us the actual confidence IMO

anuraghazra: Oh yeah that makes sense. I'll add a test for it.

packages/blade/src/components/SkipNav/SkipNav.web.tsx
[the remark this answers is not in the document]
anuraghazra: Background color is tertiary, (will be white). 
Should we ask @saurav12 to create a design for this or just go ahead with white bg, link text?

packages/blade/src/components/SkipNav/SkipNav.web.tsx
anuraghazra: Oh I meant, in most websites you will have at least one skip navigation for sure. Thus we can set a default ID for it so that consumers don't have to worry about it. 

They can just do:  

```ts
<SkipNavLink>Skip link</SkipNavLink>
<SkipNavContent />
```

packages/blade/src/components/SkipNav/SkipNav.web.tsx
anuraghazra: That's where you'll add `id` prop in one of them. In most cases there will be one skipnav only. Two skip navs are rare

packages/blade/src/components/SkipNav/SkipNav.web.tsx
anuraghazra: Done.

packages/blade/src/components/SkipNav/SkipNav.native.tsx
```diff
@@ -0,0 +1,9 @@
+const SkipNavLink = (): never => {
+  throw new Error('[Blade: SkipNav]: SkipNavLink is not available on React Native');
```
chaitanyadeorukhkar, line 2:
```suggestion
  throw new Error('[Blade: SkipNav]: SkipNav is not available on React Native');
```

packages/blade/src/components/SkipNav/SkipNav.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+import type { ReactElement } from 'react';
+import { Text } from '../Typography';
+import { SkipNavContent, SkipNavLink } from '.';
+
+const Page = (): ReactElement => {
+  return (
+    <>
+      <StorybookTitle />
+      <Subtitle>
+        The SkipNav component lets users skip the navigation and jump to the main content of the
```
chaitanyadeorukhkar, line 20: Can we mention here that this is only available for web?

packages/blade/src/components/SkipNav/SkipNav.web.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Lets take some input from @saurav12 since this would be visible to users

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+describe('<SkipNav />', () => {
+  it('<SkipNavLink /> should render correctly', () => {
+    const { getByRole } = renderWithTheme(<SkipNavLink />);
+
+    expect(getByRole('link')).toHaveTextContent('Skip to content');
+  });
+
+  it('<SkipNavLink /> should render correctly with children', () => {
+    const { getByRole } = renderWithTheme(<SkipNavLink children="Skip" />);
```
chaitanyadeorukhkar, line 14: **nit**: can we do `<SkipNavLink>{"Skip"}</SkipNavLink>`. Looks cleaner.

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+    await user.click(getByRole('link'));
+    expect(window.location.hash).toBe('#custom-id');
+
+    // Can't test for focus because jsdom doesn't move focus to fragment anchor elements
+    // expect(getByTestId('skipnav-content')).toHaveFocus();
+    expect(getByTestId('skipnav-content')).toHaveAttribute('id', 'custom-id');
+  });
+
+  it('should correctly work with two skip navs', async () => {
+    jest.useRealTimers();
```
chaitanyadeorukhkar, line 60: Why do we need real timers here? Do we need to do some cleanup here after the tests run for timers?

packages/blade/src/utils/testId/testId.native.ts
```diff
@@ -0,0 +1,5 @@
+const testId = (defaultKey: string): { testID: string } => ({
```
chaitanyadeorukhkar, line 1: Capitalise ID?
```suggestion
const testID = (defaultKey: string): { testID: string } => ({
```

chaitanyadeorukhkar: Should we do something else for this like we did for blade-old [here](https://github.com/razorpay/blade/blob/master/packages/blade-old/src/_helpers/automation-attributes.web.js)

We might want to add more meta data and could reuse such functions

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
anuraghazra: Yep not needed, was experimenting why that anchor link wasn't getting focused.

packages/blade/src/utils/testId/testId.native.ts
anuraghazra: Yeah we can even have a common function for this. like how we did mapA11yProps. But lets not do it this PR, we'll pick it up later.

packages/blade/src/components/SkipNav/SkipNav.stories.tsx
anuraghazra: Yes

packages/blade/src/components/SkipNav/SkipNav.web.tsx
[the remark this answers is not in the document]
kamleshchandnani: Closing in favor of following discussion posted on slack:

>Discussed the use case and the target users for SkipNav, basically who are the group of people going to use SkipNav.
We decided to currently go ahead with the current styling itself with white bg, padding and link color.
Saurav would need some time to research and create the final styling, we will have two variants for it banking theme and payment theme. So we will pick this up probably next quarter, until then we will move ahead and not block the accessibility things.

packages/blade/src/components/SkipNav/__tests__/SkipNav.web.test.tsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
```
kamleshchandnani, line 1: we shall add test for native to check that the errors are thrown correctly

packages/blade/src/components/SkipNav/__tests__/SkipNav.native.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+import renderWithTheme from '../../../_helpers/testing/renderWithTheme.web';
+import { SkipNavContent, SkipNavLink } from '../SkipNav';
+
+describe('<SkipNav />', () => {
+  it('<SkipNavLink /> should throw error on native', () => {
+    expect(() => renderWithTheme(<SkipNavLink />)).toThrow();
+  });
+
+  it('<SkipNavContent /> should throw error on native', () => {
+    expect(() => renderWithTheme(<SkipNavContent />)).toThrow();
```
kamleshchandnani, line 11: sir, make it verbose 

[Ref](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Typography/BaseText/__tests__/BaseText.native.test.tsx#L52-L75)

anuraghazra: Done

kamleshchandnani (approved):

PR #500, opened 2022-06-14
feat(blade): add aria hidden to icons

anuraghazra:
- Added support for accessibilityHidden in mapA11yProps.
- Added aria hidden properties in Svg.


packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
```diff
@@ -37,6 +37,18 @@ const makeAccessible = (props: Partial<AccessibilityProps>): Record<string, unkn
       continue;
     }
 
+    // handle accessibilityHidden
+    // in react-native we have to platform specific values for it.
+    if (propKey === 'hidden') {
+      if (propValue === true) {
+        newProps.accessibilityElementsHidden = true;
+        newProps.importantForAccessibility = 'no-hide-descendants';
+      }
```
kamleshchandnani, line 46: shouldn't these props be part of the map and if it's present you directly use that value?

also, why are we deleting  `accessibilityHidden` if propKey === 'hidden'?

anuraghazra: > shouldn't these props be part of the map and if it's present you directly use that value?
This is special case, no way to put it in the map

We don't need `accessibilityHidden` to be passed in the components thus we delete it, we only need to passthrough accessibilityElementsHidden, importantForAccessibility

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
```diff
@@ -37,6 +37,18 @@ const makeAccessible = (props: Partial<AccessibilityProps>): Record<string, unkn
       continue;
     }
 
+    // handle accessibilityHidden
+    // in react-native we have to platform specific values for it.
```
kamleshchandnani, line 41:
```suggestion
    // in react-native we have 2 platform-specific values for it.
```

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
kamleshchandnani: is `accessibilityHidden` is just the key to enable these two props and hence we are deleting it?

anuraghazra: Yes.

kamleshchandnani (approved):

PR #501, opened 2022-06-15
fix(typography): added accessibilityRole in heading and title component v2

anuraghazra:
- Improved the accessibility of heading and title component by adding accessibilityRole prop in BaseText

before: 

SR announces "Inaccessible heading example"

https://user-images.githubusercontent.com/35374649/171153739-fed9fd82-6b70-43ce-b07f-3e4dbf2d611f.mov

After: 

SR announces "Accessible heading example, heading"

https://user-images.githubusercontent.com/35374649/171153818-729fbde5-f9e7-4b0d-a047-6f62ded3a8fa.mov

These headers will also be available as landmark signposts.


packages/blade/src/components/Typography/Title/Title.tsx
```diff
@@ -24,6 +25,7 @@ const getProps = ({
     fontStyle: 'normal',
     lineHeight: '4xl',
     fontFamily: 'text',
+    accessibilityRole: makeAccessible({ role: 'heading' }).accessibilityRole as string,
```
kamleshchandnani, line 28: we can pass `accessibilityProps` which will accept the a11y props we support and then call `makeAccessible` in the base component or the component where all the magic happens.

anuraghazra: @kamleshchandnani this is a lot cleaner but now in web we have role="heading" in the final html output :(

anuraghazra: Fixed it by checking the platform: https://github.com/razorpay/blade/pull/501/commits/85d38cc52f03e2db69600e405cd3e6d7162aa43a#diff-5bd329389c4afd8a481956fafa1cedbe640850ef03783703e887172939c4af41R58

PR #514, opened 2022-06-21
feat: add Button component

chaitanyadeorukhkar:
Adds the `Button` component which will be exposed to our consumers


packages/blade/src/components/Button/Button/__tests__/Button.web.test.tsx
```diff
@@ -0,0 +1,114 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
chaitanyadeorukhkar, line 1: I've added some tests that seemed critical to me but not sure if what tests should be here since all of these(and more) are already covered on BaseButton

packages/blade/src/components/Button/Button/Button.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+      type: 'select',
+      options: Object.keys(iconMap),
+    },
+  },
+  parameters: {
+    docs: {
+      page: () => (
+        <>
+          <Title />
+          <Subtitle>This is the Button component which comes in 3 variants.</Subtitle>
```
chaitanyadeorukhkar, line 34: Need to improve the story to document things better, will do it on a chained PR

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -0,0 +1,57 @@
...
+  isFullWidth?: boolean;
+  onClick?: () => void;
+  type?: 'button' | 'reset' | 'submit';
+};
+
+/*
+  Mandatory children prop when icon is not provided
+  */
+type ButtonWithoutIconProps = ButtonCommonProps & {
+  icon?: undefined;
```
kamleshchandnani, line 18: check [this](https://www.typescriptlang.org/play?ssl=26&ssc=1&pln=27&pc=1#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4AocmATzCTgCEBXGGCAOwGFcQOAFHGADOcALxwA3uThwAbiijAU7GAH4AXATCKQC6vjgAfAkKRoOAEz0Hj+GElhKo+ijKHAAXkg0EANgoBzJBsCECQLYCYQEPwhXV9fGIAPOJQEsmk4YHN2Pgh3GGAOH3xfJEwYGMUAgAtK1yyhABFgIRQAIzKLH3aICDLlBtaAMSYEgHVgCxganr6B9gaOTl9sgGsfAAoASjEAPjkIKYaaOhL2ljZ2KqRTSqMTJnaQYHryAF8KKlp6ZlYOSYzCAsACSOQEEGEYkYl2WPH4ghEADJJJlssVNEx2BZysB2OEGmgasBfBYiOxNEIYIp2AEKJ9KKdfrD2ICamCEZCROI-lduCBeLlEXAUVIZOiKXAqTS6ZkiSSyUh2D5pXjZQzvnQYf8hVzobyAa8asCYBzdVDjAbWUazRDhF8kElILA4DkqdqrporXbuaiZPLSeTNPgVus4ABZJAAQnwABoPhQgA) out? Will this work to get rid of `icon?: undefined`?

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -0,0 +1,57 @@
...
+  isDisabled = false,
+  isFullWidth = false,
+  onClick,
+  size = 'medium',
+  type = 'button',
+  variant = 'primary',
+}: ButtonProps): React.ReactElement => {
+  return (
+    <BaseButton
+      {...(icon ? { icon, children } : { icon, children })}
```
kamleshchandnani, line 45: what is this piece doing? can't understand?

packages/blade/src/components/Button/Button/__tests__/Button.web.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+  });
+
+  it('should render disabled tertiary variant button', () => {
+    const buttonText = 'Pay Now';
+    const { container } = renderWithTheme(
+      <Button variant="tertiary" isDisabled={true}>
+        {buttonText}
+      </Button>,
+    );
+    expect(container).toMatchSnapshot();
```
kamleshchandnani, line 103: for web can we add assertion to check that the button is not clickable in disabled state?
```
const button = getByRole('button');
fireEvent.click(button);
expect(onClick).toHaveBeenCalledTimes(0);
```

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+/*
+   Optional children prop when icon is provided
+  */
+type ButtonWithIconProps = ButtonCommonProps & {
+  icon: IconComponent;
+  children?: string;
+};
+
+export type ButtonProps = ButtonWithoutIconProps | ButtonWithIconProps;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Discussed, we'll repeat this to avoid unintended things propagating upwards from BaseButton

packages/blade/src/components/Button/Button/Button.tsx
chaitanyadeorukhkar: Discussed TS magic. Will make this

```
      {...(icon ? { icon, children } : { children })}
```

PR #522, opened 2022-06-23
fix(blade): typo in theme.d.ts file

anuraghazra:
- minor typo, missed because of d.ts file


chaitanyadeorukhkar (approved):

anuraghazra:
This is a patch release i think.

chaitanyadeorukhkar:
I don't think this color is being used anywhere right now. Can we sneak in this fix with some other release later? Not sure if this needs a release in itself

@kamleshchandnani thoughts?

kamleshchandnani:
That's okay not a big problem. we can release this. let's not tweak our release process case by case, else we'll be in a loop of automated release but customised manually 😄

kamleshchandnani (approved):

PR #527, opened 2022-06-28
feat(blade): added InternalStoryAddon

anuraghazra:
This addon will provide ability to hide and show internal components in storybook. 

Just prefix any story name with `Internal` and it will be handled by this addon. 
You can also prefix a single story component to be internal by changing the exported story's name. 

> NOTE: Storybook doesn't provide a way for us to hide [stories from sidebar](https://github.com/storybookjs/storybook/issues/9209) thus had to use CSS workaround.

https://user-images.githubusercontent.com/35374649/176144599-4ee5e51e-edf2-403c-840b-511fd3083016.mov

### DIFF
--- a/packages/blade/.storybook/react/manager.js
+++ b/packages/blade/.storybook/react/manager.js
@@ -1,7 +1,8 @@
 // .storybook/manager.js
-
-import { addons } from '@storybook/addons';
-import { themes } from '@storybook/theming';
+import React from 'react';
+import { addons, types } from '@storybook/addons';
+import { useGlobals } from '@storybook/api';
+import { Icons, IconButton } from '@storybook/components';
 
 export const theme = {
   base: 'light',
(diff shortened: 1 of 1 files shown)


packages/blade/.storybook/react/manager.js
```diff
@@ -41,6 +42,47 @@ export const theme = {
...
+document.head.append(hiddenStoryStyle);
+
+const InternalStoryAddon = () => {
+  const [{ showInternalComponents }, updateGlobals] = useGlobals();
+
+  const toggleVisibility = React.useCallback(() => {
+    updateGlobals({
+      showInternalComponents: !showInternalComponents,
+    });
+    hiddenStoryStyle.disabled = showInternalComponents ? undefined : 'disabled';
```
kamleshchandnani, line 62: what does `disabled` do here?

anuraghazra: It disables the stylesheet 
https://developer.mozilla.org/en-US/docs/Web/API/StyleSheet/disabled

packages/blade/.storybook/react/manager.js
```diff
@@ -41,6 +42,47 @@ export const theme = {
...
+  }
+`;
+document.head.append(hiddenStoryStyle);
+
+const InternalStoryAddon = () => {
+  const [{ showInternalComponents }, updateGlobals] = useGlobals();
+
+  const toggleVisibility = React.useCallback(() => {
+    updateGlobals({
+      showInternalComponents: !showInternalComponents,
```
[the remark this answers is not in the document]
anuraghazra: updateGlobals is the state management hook [provided by storybook](https://storybook.js.org/docs/react/essentials/toolbars-and-globals#updating-globals-from-within-an-addon) 

It also sets this `showInternalComponents=false` to the URL parameter

kamleshchandnani (approved):
LGTM 🚀

PR #538, opened 2022-07-04
docs: add Link component decisions

chaitanyadeorukhkar:
### [View Formatted API Decisions](https://github.com/razorpay/blade/blob/docs/link-decisions/packages/blade/src/components/Link/_decisions/decisions.md)


packages/blade/src/components/Link/_decisions/decisions.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Link API
+
+| Prop | Type | Required | Default | Description |
+|---|---|---|---|---|
+| variant | `anchor`, `button` | No | `anchor` | The variant of the Link to be rendered.<br><br>Note: `button` will look like an inline Link component(no spacing; can be rendered alongside other text) but will be rendered as a `button` element |
+| children | `string` | No | undefined | The text to be rendered within the link. |
+| icon | `Icon` | No | undefined | The Blade `Icon` component to be rendered within the link. |
+| iconPosition | `left`, `right` | No | `right` | The position of the rendered icon. |
+| isDisabled | `boolean` | No | `false` | Control whether the link is disabled or not. |
```
anuraghazra, line 23: When the variant is `anchor` how does the `isDisabled` behaviour would work from a implementation standpoint?

packages/blade/src/components/Link/_decisions/decisions.md
```diff
@@ -0,0 +1,33 @@
...
+- [Open Questions](#open-questions)
+
+## Design
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=12699%3A147155) to the **Link** component
+
+
+## Link Component
+A Link component that can act as a `<button>` as well as an `<anchor>` element
+
+### Link API
```
anuraghazra, line 15: How would we define external links with `target=__blank` we might need another prop like `isExternal` which will add that in.

chaitanyadeorukhkar: Yup. This needs to be added. Let me think this through

packages/blade/src/components/Link/_decisions/decisions.md
chaitanyadeorukhkar: From an implementation standpoint, we could remove the href from the component and add other visual cues (color, cursor). We could also set `aria-disabled` as true in this case

packages/blade/src/components/Link/_decisions/decisions.md
```diff
@@ -0,0 +1,33 @@
...
+| onClick | `Function` | No | undefined | The function to be called when the link is clicked. |
+| href | `string` | No | undefined | The href for the link component.<br><br>Note: This will be ignored when the variant is `button` |
+
+     
+## Component Breakdown
+
+<img src="./component-breakdown.png" width="300px"/>
+
+## Open Questions
+NA
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Why would we need to think of that? This should be a barebone component. The consumer can pass proper `href`/`onClick` that can handle where it needs to go

packages/blade/src/components/Link/_decisions/decisions.md
```diff
@@ -0,0 +1,33 @@
...
+### Link API
+
+| Prop | Type | Required | Default | Description |
+|---|---|---|---|---|
+| variant | `anchor`, `button` | No | `anchor` | The variant of the Link to be rendered.<br><br>Note: `button` will look like an inline Link component(no spacing; can be rendered alongside other text) but will be rendered as a `button` element |
+| children | `string` | No | undefined | The text to be rendered within the link. |
+| icon | `Icon` | No | undefined | The Blade `Icon` component to be rendered within the link. |
+| iconPosition | `left`, `right` | No | `right` | The position of the rendered icon. |
+| isDisabled | `boolean` | No | `false` | Control whether the link is disabled or not. |
+| onClick | `Function` | No | undefined | The function to be called when the link is clicked. |
```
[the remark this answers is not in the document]
anuraghazra: +1. We might need to accept both because of scenarios like analytics libraries where user might need to send some event when the link is Clicked. 

> Should we consider scenarios where devs want the onClick to override the href but still want it as a link and in that case the href can be equated to javascript:void(0)?

If we devs want the onClick to override the href they can just `return false` from the onClick right?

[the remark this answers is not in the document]
chaitanyadeorukhkar: `href` should do what it natively does, take the user to a specific place. `onClick` should be called irrespective of `href`. Consumers would need this in cases where they might want to track click events with some analytics service after the link is clicked and that logic can go inside `onClick`.

packages/blade/src/components/Link/_decisions/decisions.md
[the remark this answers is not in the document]
nashcheez: What even is a disabled link? That imo would be un-semantic. Design could have an option, but dev should rather create a `button` if required with the `disabled` attribute (a more likelier option for the `button` variant). Let's discuss with @saurav12 in details.

packages/blade/src/components/Link/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: How would consumers use this Link component with ReactRouter or Next.js Link component? We don't want users creating their own styled wrapper around the third party libraries right?

nashcheez (dismissed):
Looks good, feel free to open this up for review by everyone else.

packages/blade/src/components/Link/_decisions/decisions.md
```diff
@@ -0,0 +1,41 @@
...
+| href | `string` | No | undefined | The href for the link component.<br><br>Note: This prop is only valid for `anchor` variant. |
+| target | `string` | No | undefined | The target for the link component.<br><br>Note: This prop is only valid for `anchor` variant. |
+| accessibilityLabel | `string` | No | undefined | The `aria-label` (web) & `accessibilityLabel` (native) of the link component. |
+
+     
+## Component Breakdown
+
+<img src="./component-breakdown.png" width="300px"/>
+
+## Accessibility
```
chaitanyadeorukhkar, line 35: @anuraghazra @nashcheez Check this?

nashcheez: Looks fine, just ensure the aria attributes and a11y considerations that we have done for the button component are carried forward to the `variant = button`.

anuraghazra: LGTM.

anuraghazra:
LGTM, We can open for further public reviews.

packages/blade/src/components/Link/_decisions/decisions.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Link API
+
+| Prop | Type | Required | Default | Description |
+|---|---|---|---|---|
+| variant | `anchor`, `button` | *No* | `anchor` | The variant of the Link to be rendered.<br><br>Note: `button` will look like an inline Link component (no spacing; can be rendered alongside other text) but will be rendered as a `button` element. |
+| children | `string` | When `icon` is passed: *No*<br>When `icon` is not passed: *Yes*   | undefined | The text to be rendered within the link. |
+| icon | `Icon` | When `children` is passed: *No*<br>When `children` is not passed: *Yes* | undefined | The Blade `Icon` component to be rendered within the link. |
+| iconPosition | `left`, `right` | *No* | `right` | The position of the rendered icon. |
+| isDisabled | `boolean` | *No* | `false` | Control whether the link is disabled or not.<br><br>Note: This prop is only valid for `button` variant. |
```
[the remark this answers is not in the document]
anuraghazra: [Yeah it can](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AocmATzCTgGFcQIA7AGWFYGsAFHMAGc4AXjgBvcnDgA3FFGApWMAFwElaABbR8cAD4EARgFcYMNmSlxNRTAH41gmAtYBzCgF9KNOnE49+CCEAQVYtaFEJKzkFJVV1MO0oS2k2ABtqfz4BR2cud3IvH3oswKEAIVNzVkjJaRjFZTV8EzMLCmlgQQARLpRDNKQAEwc4QwgIQaVPb1oSrmyg4TEmEBYOBbLhADI4AApSgUFQ8Kh9P02jyrbWAEoKNDYnC55IvbAjtUOl29EAPii0iIMGMUBqrGMaTSM3IpAAPFlZPJGjARAAiDRJNFwLq9QT9QZDODpTKXILo7EAej+sIRCyRsWU6Na1WxuL6A2GxNYGW+YApcGpQA) we are already using a lot of conditional typescript props in BaseText, Button and other components.

packages/blade/src/components/Link/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: Since clicking with `href` can take people do a different page, do we handle waiting for the `onClick` handler to finish before the page unmounts?

packages/blade/src/components/Link/_decisions/decisions.md
divyanshu013: Should also add noopener noreferrer I suppose with external link.

packages/blade/src/components/Link/_decisions/decisions.md
```diff
@@ -0,0 +1,62 @@
...
+
+
+## Internal BaseLink Component
+Internal component that exposes certain extra props to enable creation of Component Level Action Link like `positive`, `negative`, `neutral`, `information`, `notice`
+
+### BaseLink API
+| Prop | Type | Required | Default | Description |
+|---|---|---|---|---|
+| variant | `anchor`, `button` | *No* | `anchor` | The variant of the Link to be rendered.<br><br>Note: `button` will look like an inline Link component (no spacing; can be rendered alongside other text) but will be rendered as a `button` element. |
+| intent |  `positive`,`negative`, `notice`, `information`, `neutral` | *No* | undefined | The intent of the link to be rendered.<br><br>  Note: <br> - We will use this to set the intent for internal Blade components that have positive, negative, notice, info, or neutral link buttons like an Alert component. |
```
divyanshu013, line 22: How would default `undefined` intent translate design wise 🤔

divyanshu013 (commented):
It seems we're using the `Link` as both a button and anchor. What's the rationale for having it work like a button too? In other words, why would someone want to use `Link` if all they need is button functionality.
From my general understanding, I would expect Link to always work like an anchor tag / link which are associated with an `href` 🤔

chaitanyadeorukhkar:
> It seems we're using the `Link` as both a button and anchor. What's the rationale for having it work like a button too? In other words, why would someone want to use `Link` if all they need is button functionality. From my general understanding, I would expect Link to always work like an anchor tag / link which are associated with an `href` 🤔

The are cases when you need to use a something inline that looks like a link but isn't semantically a link. For example,
"Forgot Password? **Click to reset**" or 
"Haven't received an OTP? **Resend OTP**"

In such cases, people generally use a link and let it behave as a button while being semantically incorrect (`a` tag in dom instead of `button`) or use some `inline` variant of a button which looks mostly like a button but has no spacing. 

Button supports different sizes, whereas Link is always inline and matches the size of the body text. Since it's structure and usage matches closer to Link, we decided to go ahead with creating a variant of Link that can behave as a button.

packages/blade/src/components/Link/_decisions/decisions.md
chaitanyadeorukhkar: When the intent is not defined, it'll behave like a normal link (blue in color) but when there is an intent, it will look as per the intent (red for negative, yellow for notice, etc.)

packages/blade/src/components/Link/_decisions/decisions.md
divyanshu013: Would it also accept `rel` prop for noopener noreferrer?

packages/blade/src/components/Link/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: `href` is useful for navigation, SEO, keyboard press

divyanshu013:
Semantically we might be rendering a button correctly but usability wise, this seems weird to me:

```js
<Link onClick={}>Resend OTP</Link>
```

It seems Link here just creates a "different looking button" whereas I would assume this would've been a different button variant 🤔

chaitanyadeorukhkar:
> Semantically we might be rendering a button correctly but usability wise, this seems weird to me:
> 
> ```js
> <Link onClick={}>Resend OTP</Link>
> ```
> 
> It seems Link here just creates a "different looking button" whereas I would assume this would've been a different button variant 🤔

Yeah it does look a bit weird but when you think of a "Button" you'd be expecting an element that takes up a good amount of real estate with proper padding whereas links are used right between a body text to "link" a user somewhere. 

We could have moved this to be a variant of a button but we'd also need to ensure that we restrict usage of `size`, `isLoading`, `type` & `fullWidth` when the variant is `inline` (or `link`)

Its a very subjective decision where neither seems a 100% accurate but making this a variant of Link seemed more appropriate given how it's used from design pov as a link (design had this as a variant of button called "link" initially) and it's usage & restrictions match more with Link than with Button

packages/blade/src/components/Link/_decisions/decisions.md
chaitanyadeorukhkar: Added!

packages/blade/src/components/Button/_decisions/decisions.md
```diff
@@ -17,8 +17,8 @@ Internal component that exposes certain extra props to enable creation of Compon
 
 | Prop | Type | Required | Default | Description |
 |---|---|---|---|---|
-| variant | `primary`,  `secondary`,  `tertiary`,  `link` | No | `primary` | - The variant of the button to be rendered.<br><br>Note: `link` will look like an inline Link component but will be rendered as a `button` element |
-| intent | `positive`,`negative`, `notice`, `info`, `neutral` | No | undefined | The intent of the button to be rendered.<br><br>  Note: <br> - We will use this to set the intent for internal Blade components that have positive, negative, notice, info, or neutral buttons like an Alert component<br> - All the variants of the button will not be using any intent since it would only be required for internal components. |
+| variant | `primary`,  `secondary`,  `tertiary` | No | `primary` | - The variant of the button to be rendered. |
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: No no, just removed it from button decisions

packages/blade/src/components/Link/_decisions/decisions.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+## Link Component
+A Link component that can act as a `<button>` as well as an `<anchor>` element. This will be the component exported to the consumers. This is a wrapper on top of `BaseLink` component to avoid exposing `intent` & `contrast` to consumers.
+
+### Link API
+
+| Prop | Type | Required | Default | Description |
+|---|---|---|---|---|
+| variant | `anchor`, `button` | *No* | `anchor` | The variant of the Link to be rendered.<br><br>Note: `button` will look like an inline Link component (no spacing; can be rendered alongside other text) but will be rendered as a `button` element. |
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Better if we make it explicit otherwise it could lead to unintended changes for the consumer.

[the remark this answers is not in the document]
kamleshchandnani: I would recommend keeping the API surface as lean as possible.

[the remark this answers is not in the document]
chaitanyadeorukhkar: Yes, but this is a very critical differentiation on whether it will be rendered as an `anchor` element or a `button` element. 

For example, a user passes `onClick` and decides to handle link navigation by themselves by adding `e.preventDefault()` on the onClick and then setting `window.location.href`. In this case the consumer is expecting an `a` tag to be rendered but hasn't passed `href` cuz they have decided to handle it themselves. But since we internally decided to render it as a `button` since `href` is missing, the consumer will end up having an unintended `button` element instead of a `a` element. 

This is why I think we should let the consumer explicitly handle the switch between `a` & `button`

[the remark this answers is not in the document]
kamleshchandnani: if the consumer wants to do something nasty they'll figure out anyway to do it. this seems to be a very rare edge case, as an API designer i would not tradeoff a new prop and learning curve just for such a use case. even if we have a variant(which is not a required prop) consumers can still achieve the use case you have mentioned because they didn't care enough to go through the documentation.

kamleshchandnani:
Only https://github.com/razorpay/blade/pull/538/files#r925567533 comment, else we are good to go

packages/blade/src/components/Link/_decisions/decisions.md
```diff
@@ -0,0 +1,64 @@
...
+| children | `string` | When `icon` is passed: *No*<br>When `icon` is not passed: *Yes*   | undefined | The text to be rendered within the link. |
+| icon | `Icon` | When `children` is passed: *No*<br>When `children` is not passed: *Yes* | undefined | The Blade `Icon` component to be rendered within the link. |
+| iconPosition | `left`, `right` | *No* | `right` | The position of the rendered icon. |
+| isDisabled | `boolean` | *No* | `false` | Control whether the link is disabled or not.<br><br>Note: This prop is only valid for `button` variant. |
+| onClick | `Function` | No | undefined | The function to be called when the link is clicked. |
+| href | `string` | No | undefined | The href for the link component.<br><br>Note: This prop is only valid for `anchor` variant. |
+| target | `string` | No | undefined | The target for the link component.<br><br>Note: This prop is only valid for `anchor` variant. |
+| rel | `string` | No | undefined | The rel for the link component.<br><br>Note: This prop is only valid for `anchor` variant. |
```
kamleshchandnani, line 31: we can basically say if `target="_blank"` at that time we can define the default value as `noopener noreferrer`

divyanshu013: Shouldn't we keep it `undefined` by default? Similar to `a` tag?

packages/blade/src/components/Link/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: I see the point chaitanya is making. But if consumer wants to do that "nasty" navigation they can pass it like this

<Link href="#" onClick={e => e.preventDefault()} /> 

> Also there's nothing called a link without href. If you don't pass href to link. It renders like a plain text :lol:

packages/blade/src/components/Link/_decisions/decisions.md
kamleshchandnani: depends what we are solving for. here we solving for some better practices given to the consumers as out of the box while they use blade components vs a wrapper over `a` tag with just styles and then consumers decide what props they want to pass

packages/blade/src/components/Link/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: we can rather add a TS check which tells that at least one of `href` or `onClick` is required and then we'll be all good

packages/blade/src/components/Link/_decisions/decisions.md
divyanshu013: If we're modifying the HTML defaults then I feel it would make more sense to have a different prop like `isExternal` for setting both target and rel accordingly. Makes the intent clearer imo.

kamleshchandnani (approved):

packages/blade/src/components/Link/_decisions/decisions.md
kamleshchandnani: but then how we will support other target values `target="_blank|_self|_parent|_top|framename"` if we add a boolean prop?

divyanshu013: Non blocker: yes, I think we can provide `target` and `rel` for overriding things. `isExternal` should be able to cover majority usecases.

divyanshu013 (commented):
There are certain web only properties like `target`, `rel`, how will those be handled for native?

chaitanyadeorukhkar:
> There are certain web only properties like `target`, `rel`, how will those be handled for native?

@divyanshu013 for such properties, we ignore them on native. The education effort will be done via documentation where we'll mention these are web only props. Once we have support for [TS Module suffixes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#resolution-customization-with-modulesuffixes), we will leverage that as well to help consumers understand what props are available for their platform

PR #539, opened 2022-07-05
docs(blade): add Checkbox API decisions

anuraghazra:
## [View Rendered Text](https://github.com/razorpay/blade/blob/anu/checkbox-api/packages/blade/src/components/Checkbox/_decisions/decisions.md)


packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,262 @@
...
+| value (html native) | `string`   | No       | undefined | The value of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) |
+| isDisabled          | `boolean`  | No       | `false`   | Control whether the checkbox is disabled or not.                                                                                                         |
+| isReadOnly          | `boolean`  | No       | `false`   | Control whether the checkbox is readonly or not.                                                                                                         |
+| isInvalid           | `boolean`  | No       | `false`   | Control whether the checkbox is invalid or not.                                                                                                          |
+| isRequired          | `boolean`  | No       | `false`   | Control whether the checkbox is required or not.                                                                                                         |
```
nashcheez, line 83: We only have designs for `Disabled` and `Negative`. How will the `isReadOnly=true` and `isRequired=true` checkboxes look from a design standpoint? (assuming `isInvalid` would be the `Negative` state). Do we even have use cases for these in our products? @saurav12

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,262 @@
...
+
+
+## `CheckboxGroup` API
+
+
+| Prop          | Type                        | Required | Default     | Description                                                                                                                        |
+| ------------- | --------------------------- | -------- | ----------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| children      | React.ReactNode             | No       | `undefined` | Accepts multiple Checkboxes                                                                                                        |
+| label         | `string`                    | No       | `undefined` | The label of the group                                                                                                             |
+| labelPosition | `top                        | left`    | No          | `top`                                                                                                                              | The position of the rendered label |
```
nashcheez: Need to re-format this as `top | left`, currently the operator is being treated as the ending of the cell.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,262 @@
...
+    >
+      <Checkbox value="use-ts">I will use TS</Checkbox>
+      <Checkbox value="no-any">I won't use any</Checkbox>
+      <Checkbox value="use-js">I won't use JS</Checkbox>
+    </CheckboxGroup>
+  );
+}
+```
+
+And lastly users can also individually set `isRequired`, `isInvalid` props on the <Checkbbox \/> components instead of the parent CheckboxGroup for greater fine control.
```
nashcheez:
```suggestion
And lastly users can also individually set `isRequired`, `isInvalid` props on the <Checkbox \/> components instead of the parent CheckboxGroup for greater fine control. 
```

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,262 @@
...
+      <Checkbox value="chaitanya">Chaitanya</Checkbox>
+    </CheckboxGroup>
+  )
+}
+```
+
+
+### Note on CheckboxGroup
+
+For properties `isDisabled`, `isReadOnly`, `isInvalid`, `isRequired` we want the CheckboxGroup to pass down these props to all the Checkboxes inside of it, but there is a gotcha.
```
nashcheez: Should we pass these props from the `CheckboxGroup` or at individual 'Checkbox'? I don't remember the last time when I saw all the checkboxes in a group being disabled, readonly or invalid at once. Generally individual checkboxes can have those states.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,262 @@
...
+<Checkbox helperText="Hint: yes">is blade awesome?</Checkbox>
+// uncontrolled
+<Checkbox defaultChecked={true}>is blade awesome?</Checkbox>
+// controlled
+<Checkbox isChecked={true | false} onChange={isChecked => {}}>is blade awesome?</Checkbox>
+```
+
+### Validations
+
+with `isInvalid` & `isRequired` will pass down `required` and `aria-invalid` attributes accordingly
```
nashcheez, line 102:
```suggestion
`isRequired` & `isInvalid` will pass down `required` and `aria-invalid` attributes respectively
```

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,373 @@
...
+      <CheckboxHelperText />
+    </>
+  );
+};
+```
+
+CheckboxGroup Component's Internal Anatomy:
+
+```tsx
+const CheckboxGrop = () => {
```
chaitanyadeorukhkar, line 57:
```suggestion
const CheckboxGroup = () => {
```

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,262 @@
...
+| ------------------- | ---------- | -------- | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| children            | `string`   | No       | undefined | The text to be rendered as CheckboxLabel                                                                                                                 |
+| helperText          | `string`   | No       | undefined | The helper text to be rendered                                                                                                                           |
+| isChecked           | `boolean`  | No       | false     | state of the checkbox, if explicitly set the checkbox will be controlled                                                                                 |
+| defaultChecked      | `boolean`  | No       | false     | default state of the checkbox, if isChecked is not provided the checkbox will be uncontrolled                                                            |
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: `defaultChecked` will act as an initializer for when we want the checkbox to be uncontrolled. If consumer passes `isChecked={true}` it will _always_ remain checked even when a user tries to toggle the checkbox until consumer changes it to `isChecked={false}` but in case of `defaultChecked={true}` it will start at an initial state of being checked but when a user clicks on the checkbox, it will get unchecked even though `defaultChecked` remains true.

packages/blade/src/components/Checkbox/_decisions/decisions.md
chaitanyadeorukhkar: +1. Not sure what the use-case for isReadOnly is here and why do we need it

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,373 @@
...
+| name (html native)  | `string`   | No       | undefined | The name of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name)            |
+| value (html native) | `string`   | No       | undefined | The value of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) |
+| isDisabled          | `boolean`  | No       | `false`   | Control whether the checkbox is disabled or not.                                                                                                         |
+| isReadOnly          | `boolean`  | No       | `false`   | Control whether the checkbox is readonly or not.                                                                                                         |
+| isInvalid           | `boolean`  | No       | `false`   | Control whether the checkbox is invalid or not.                                                                                                          |
```
chaitanyadeorukhkar, line 82: Since this indicates an error state. Invalid seems too specific whereas error seems a bit generic. It might make a consumer think that setting it as invalid might make the checkbox invalid and make it disappear from the checkbox group. Error is a widely accepted negative state. Thoughts?
```suggestion
| hasError           | `boolean`  | No       | `false`   | Control whether the checkbox is in a error state.                                                                                                          |
```

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,373 @@
...
+<Checkbox>is blade awesome?</Checkbox>
+// Helper text
+<Checkbox helperText="Hint: yes">is blade awesome?</Checkbox>
+// uncontrolled
+<Checkbox defaultChecked={true}>is blade awesome?</Checkbox>
+// controlled
+<Checkbox isChecked={true | false} onChange={isChecked => {}}>is blade awesome?</Checkbox>
+```
+
+#### Validations
```
chaitanyadeorukhkar, line 100: Some of these validations seem a bit complex to understand at first glance. You've explained it well here but our API won't be able to communicate it when a consumer is using it.  Lets discuss if we can reduce some complexity here?

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,373 @@
...
+      <Checkbox value="no-any">I won't use any (you can use it)</Checkbox>
+      <Checkbox value="use-js" isRequired>
+        I won't use JS
+      </Checkbox>
+    </CheckboxGroup>
+  );
+}
+```
+
+## Component Architecture
```
chaitanyadeorukhkar, line 263: Good details but I'd suggest removing these from the decision doc since this might attract a lot of feedbacks & suggestions that might not be relevant at API decision phase. We can review this in the implementation PR instead and you might realise some other things while implementing that you might not be able to at this time which could change this architecture too.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,373 @@
...
+```
+
+**CheckboxGroupContext:**
+
+CheckboxGroupContext will pass down the <CheckboxGroup \/> properties like `isDisabled`, `isReadOnly` to the Checkboxes.
+
+## Accessibility
+
+- Checkbox - https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/
+- CheckboxGroup - The group's accessibility is bit tricky to get right, thus I'll implement and refine it and then document it here.
```
chaitanyadeorukhkar, line 342: Translates to - "I'll handle it, don't worry" 

Nice 😂

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,373 @@
...
+  isInvalid={isInvalid}
+  helperText={
+    isInvalid
+    ? 'Invalid username provided, contains special chars' 
+    : 'Enter valid username'
+  }
+/>
+```
+
+Will it be better to have a `errorText` prop instead which will automatically be switched internally is `isInvalid` is true?
```
chaitanyadeorukhkar, line 363: I think this should be the case. Whenever `isInvalid` (I prefer `hasError`) is true, it will ignore the helperText & render the errorText instead. We could even enforce consumers to pass an `errorText` whenever they are using `isInvalid` (`hasError`)

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Both will be supported you can pass them to Group or individually too. Since <Checkbox /> itself can also have all those common properties.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,263 @@
...
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=13227%3A163026) to all variants of the Checkbox component
+
+## Anatomy
+
+- **Checkbox**
+  - **CheckboxInput** - Hidden native <input \/> element
+  - **CheckboxIcon** - Checked/Unchecked & indeterminate icon
+  - **CheckboxLabel** - The Checkbox's <label \/> element
+  - **CheckboxLabelText** - The cross-platform Text component
+  - **CheckboxHelpText** - The helper text component
```
chaitanyadeorukhkar, line 29: We should look for a way to abstract out `Label` & `HelperText` components so that they can be used for other components that are coming up as well. Anyway - for now let's keep it this way in the API decisions here. We'll figure this out while implementing in the PR

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,263 @@
...
+const CheckboxGroup = () => {
+  return (
+    <CheckboxGroupFieldset>
+      <CheckboxGroupLegend>title</CheckboxGroupLegend>
+      {children}
+      <CheckboxGroupHintText variant="help | error" />
+    </CheckboxGroupFieldset>
+  );
+};
+```
```
chaitanyadeorukhkar, line 66: Lets remove this for now? We can take care of this in the implementation PR.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,263 @@
...
+| name (html native)  | `string`   | No       | undefined | The name of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name)            |
+| value (html native) | `string`   | No       | undefined | The value of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) |
+| isDisabled          | `boolean`  | No       | `false`   | Control whether the checkbox is disabled or not.                                                                                                         |
+| hasError           | `boolean`  | No       | `false`   | Control whether the checkbox is invalid or not.                                                                                                          |
+| isRequired          | `boolean`  | No       | `false`   | Control whether the checkbox is required or not.                                                                                                         |
```
chaitanyadeorukhkar, line 82: Didn't we conclude that we'll remove this?

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,263 @@
...
+| labelPosition | `top, left`                 | No       | `top`       | The position of the rendered label                                                                                                 |
+| helpText    | `string`                    | No       | `undefined` | The helper text to be rendered                                                                                                     |
+| errorText    | `string`                    | No       | `undefined` | The error text to be rendered                                                                                                     |
+| defaultValue  | `string[]`                  | No       | `[]`        | The initial value of the checkbox group                                                                                            |
+| value         | `string[]`                  | No       | `[]`        | The value of the checkbox, if present will act as controlled component                                                             |
```
chaitanyadeorukhkar, line 117: Should these always accept an array or can it take a string when only 1 value needs to be passed? 🤔

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+## Accessibility
+
+- Checkbox - https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/
+- CheckboxGroup - The group's accessibility is bit tricky to get right, thus I'll implement and refine it and then document it here. 
+
+## Open Questions
+
+1. Currently we only have `helpText` in Checkbox & CheckboxGroup and when we set `hasError` it becomes the `negative` variant. The problem is the content of the `helpText` should also change in certain scenarios to convey the exact error.
```
chaitanyadeorukhkar, line 234: We concluded that we'll go ahead with `helpText` as well as `errorText`. Can we strike off this open question now?

chaitanyadeorukhkar (changes requested):

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Always array. `string | string[]` will just make things complicated.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: I've added "conclusion" in the end.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Yeah, We will once we have more form components and have a common API for that.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Hmm.. missed this part but we concluded to remove it on the CheckboxGroup in addition of showOptionalLabel, 
The user can mark individual Checkbox isRequired right if not used with CheckboxGroup?

nashcheez: I think we decided to remove `isRequired` from everywhere and just have `showOptionalLabel` on the CheckboxGroup. All the validations for if a checkbox has to be checked or not will be dependent on the logic provided by consumers.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,217 @@
...
+<Checkbox helpText="Hint: yes">is blade awesome?</Checkbox>
+// uncontrolled
+<Checkbox defaultChecked={true}>is blade awesome?</Checkbox>
+// controlled
+<Checkbox isChecked={true | false} onChange={isChecked => {}}>is blade awesome?</Checkbox>
+```
+
+#### Validations
+
+And `hasError` will pass `aria-invalid` attributes to indicate invalid state.
```
nashcheez, line 55:
```suggestion
`hasError` will pass `aria-invalid` attributes to indicate invalid state.
```

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,217 @@
...
+| ------------- | --------------------------- | -------- | ----------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| children      | `React.ReactNode`             | Yes       | `undefined` | Accepts multiple Checkboxes                                                                                                        |
+| label         | `string`                    | Yes       | `undefined` | The label of the group                                                                                                             |
+| labelPosition | `top, left`                 | No       | `top`       | The position of the rendered label                                                                                                 |
+| helpText    | `string`                    | No       | `undefined` | The helper text to be rendered                                                                                                     |
```
nashcheez, line 68: I am assuming the only difference between `helpText` and `errorText` at the Checkbox group level is that one appears in grey while the other appears in red, they both have the same positioning. What happens if a user provides values for both? Are they going to stack up on one another. Don't remember seeing that in the designs though.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,217 @@
...
+    </CheckboxGroup>
+  );
+};
+```
+
+### Note on CheckboxGroup
+
+For properties `isDisabled`, `hasError` we want the CheckboxGroup to pass down these props to all the Checkboxes inside of it.
+
+For showing optional label in `CheckboxGroup` we will introduce a prop `showOptionalLabel` which will only render a `(optional)` tag in the CheckboxGroupLabel,
```
nashcheez, line 131: I think we can add a screenshot here from design of how the `optional` label will look so that it's clearer.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Oh 🤔 

We also discussed along the lines of "The checkbox will not be used outside of CheckboxGroup" something like that right? If so then it's fine to remove isRequired. 

But if say the checkbox can be used outside of <CheckboxGroup then we should consider isRequired. for html forms and screen readers.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Added

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| isIndeterminate     | `boolean`  | No       | false     | mixed state of the checkbox                                                                                                                              |
+| name (html native)  | `string`   | No       | undefined | The name of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name)            |
+| value (html native) | `string`   | No       | undefined | The value of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) |
+| isDisabled          | `boolean`  | No       | `false`   | Control whether the checkbox is disabled or not.                                                                                                         |
+| hasError           | `boolean`  | No       | `false`   | Control whether the checkbox is invalid or not.                                                                                                          |
```
avinashdvv, line 35: question: can't we use the `errorText` to find if CheckBox is valid or not. Why we need to use `hasError`. It will put the source of the truth for error in multiple places.

anuraghazra: Yes from code side we can but it won't be ergonomic to use. 
If we only provide errorText users will have to manually handle logic to set or unset it. 

```tsx
const { errors } = useSomeFormLibrary();

<CheckboxGroup errorText={errors.length > 0 ? "Opps something went wrong" : null} />
```

While with hasError it's much cleaner

```tsx
const { errors } = useSomeFormLibrary();

<CheckboxGroup 
   hasError={errors.length > 0}
   errorText="Opps something went wrong" 
 />
```

hasError in this case will just be a boolean toggle if set to true `errorText` will be shown

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| helpText          | `string`   | No       | undefined | The helper text to be rendered                                                                                                                           |
+| isChecked           | `boolean`  | No       | false     | state of the checkbox, if explicitly set the checkbox will be controlled                                                                                 |
+| defaultChecked      | `boolean`  | No       | false     | default state of the checkbox, if isChecked is not provided the checkbox will be uncontrolled                                                            |
+| onChange            | `Function` | No       | undefined | The function to be called when the checkbox state changes.                                                                                               |
+| isIndeterminate     | `boolean`  | No       | false     | mixed state of the checkbox                                                                                                                              |
```
divyanshu013, line 31: In which cases can there be an indeterminate state? Let's add in description too

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+
+Example of only one checkbox needs to be checked:
+
+```tsx
+function AnyOneHasToBeSelected() {
+  const [selected, setSelected] = React.useState([]);
+
+  return (
+    <CheckboxGroup
+      label="Do you love any of the fruites?"
```
divyanshu013, line 147:
```suggestion
      label="Do you love any of the fruits?"
```

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+      helpText="You can select multiple"
+      value={selected}
+      onChange={setSelected}
+      hasError={selected.length === 0}
```
divyanshu013, line 151: This check seems contrasting with the description - example of only one checkbox needs to be checked and the helpText. With this check, we can have multiple selected right? Should it be `selected.length !== 1`

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Will do, isIndeterminate will be an independent state boolean. which the user will handle the logic. 

[Live demo: ](https://61c19ee8d3d282003ac1d81c-evllbcmcsj.chromatic.com/?path=/story/components-checkbox--indeterminate-checkbox&globals=measureEnabled:false)

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: lol it's always this spelling mistake 🙈 I should start using other examples like `Select your fav typescript feature`.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: The example is `AnyOneHasToBeSelected` thus seems fine? If none of them are selected we will set error to true.

divyanshu013: Oh I see, got me confused. You should reword it - <s>only one</s> any one / atleast one 😄

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
+# Checkbox Decisions <!-- omit in toc -->
```
kamleshchandnani, line 1: image ko render to krdo bhai markdown me, can't find it 🙈

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| name (html native)  | `string`   | No       | undefined | The name of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name)            |
+| value (html native) | `string`   | No       | undefined | The value of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) |
+| isDisabled          | `boolean`  | No       | `false`   | Control whether the checkbox is disabled or not.                                                                                                         |
+| hasError           | `boolean`  | No       | `false`   | Control whether the checkbox is invalid or not.                                                                                                          |
+| isRequired          | `boolean`  | No       | `false`   | Control whether the checkbox is required or not.                                                                                                         |
```
kamleshchandnani, line 36: can't find this prop on Checkbox Group 🤔

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| helpText    | `string`                    | No       | `undefined` | The helper text to be rendered                                                                                                     |
+| errorText    | `string`                    | No       | `undefined` | The error text to be rendered                                                                                                     |
+| defaultValue  | `string[]`                  | No       | `[]`        | The initial value of the checkbox group                                                                                            |
+| value         | `string[]`                  | No       | `[]`        | The value of the checkbox, if present will act as controlled component                                                             |
+| onChange      | `(value: string[]) => void` | No       | `undefined` | The function to be called when any checkbox's state changes                                                                        |
```
kamleshchandnani, line 72: i think we shall also pass `event` and `name` as args along with `value`

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### Basic
+
+```tsx
+<CheckboxGroup
+  label="Developers"
+  helpText="Pick blade developers"
+  defaultValue={['anurag', 'kamlesh']}
+>
+  <Checkbox value="anurag">Anurag</Checkbox>
```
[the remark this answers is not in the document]
anuraghazra: You can't really. At least not from TS. 
We can either filter out only the <Checkbox /> or throw an error on runtime.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| defaultChecked      | `boolean`  | No       | false     | default state of the checkbox, if isChecked is not provided the checkbox will be uncontrolled                                                            |
+| onChange            | `Function` | No       | undefined | The function to be called when the checkbox state changes.                                                                                               |
+| isIndeterminate     | `boolean`  | No       | false     | mixed state of the checkbox                                                                                                                              |
+| name (html native)  | `string`   | No       | undefined | The name of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name)            |
+| value (html native) | `string`   | No       | undefined | The value of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) |
```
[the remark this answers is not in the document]
anuraghazra: Yes. This value & name prop will be mandatory if you want uncontrolled forms which some form libraries do.

packages/blade/src/components/Checkbox/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Yes, we have a CheckboxGroupHint which has variant of `'error' | 'help'`

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| children            | `string`   | No       | undefined | The text to be rendered as CheckboxLabel                                                                                                                 |
+| helpText          | `string`   | No       | undefined | The helper text to be rendered                                                                                                                           |
+| isChecked           | `boolean`  | No       | false     | state of the checkbox, if explicitly set the checkbox will be controlled                                                                                 |
+| defaultChecked      | `boolean`  | No       | false     | default state of the checkbox, if isChecked is not provided the checkbox will be uncontrolled                                                            |
+| onChange            | `Function` | No       | undefined | The function to be called when the checkbox state changes.                                                                                               |
```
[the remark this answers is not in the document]
anuraghazra: Sure

packages/blade/src/components/Checkbox/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: No doesn't depend on isChecked. isIndeterminate will override the isChecked state. 

So if you have 
<Checkbox isChecked={true} isIndeterminate={true} />

The checkbox will be in indeterminate state and because we don't want. 
example: https://github.com/razorpay/blade/pull/544/files#diff-ecad1a0cb9d580749fedae8066d59e12d65733f666f0f39cc9389af1bfd057ffR129

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+
+```tsx
+// Label text
+<Checkbox>is blade awesome?</Checkbox>
+// Helper text
+<Checkbox helpText="Hint: yes">is blade awesome?</Checkbox>
+// uncontrolled
+<Checkbox defaultChecked={true}>is blade awesome?</Checkbox>
+// controlled
+<Checkbox isChecked={true | false} onChange={isChecked => {}}>is blade awesome?</Checkbox>
```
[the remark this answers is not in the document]
anuraghazra: Currently only passing the isChecked state, not sure if we need value since it will be static anyways.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: > helpText and users can themselves do helpText = hasError? errortext:helptext?

In here we already have the information if the CheckboxGroup is in error state or not via hasError. and you also need this to change the helper text color right? Users should not handle all these logic IMO. 

So this will look kinda odd: 
<CheckboxGroup hasError helperText={hasError ? "Err" : "Ok" } />

packages/blade/src/components/Checkbox/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: can't we restrict the children to be of certain types only? at least with react's helper we can do runtime, but isn't that information available at type level?

packages/blade/src/components/Checkbox/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: what do you mean by static?

kamleshchandnani:
@anuraghazra a lot of comments I'd put seems to not have been addressed, can you check once again?

packages/blade/src/components/Checkbox/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: value won't change even if checkbox is checked or unchecked

packages/blade/src/components/Checkbox/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: No we are already executing the JSX elements by the time we accept the children. 

<Checkbox />
Translates to: 
React.createElement(Checkbox, {})

Every JSX element will just return `JSX.Element` as it's type. 
For now [TS can't handle it](https://github.com/microsoft/TypeScript/issues/13618)

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: We need some discussion around this 
https://github.com/razorpay/blade/pull/539#discussion_r918765931

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+
+## `Checkbox` API
+
+| Prop                | Type       | Required | Default   | Description                                                                                                                                              |
+| ------------------- | ---------- | -------- | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| children            | `string`   | No       | undefined | The text to be rendered as CheckboxLabel                                                                                                                 |
+| helpText          | `string`   | No       | undefined | The helper text to be rendered                                                                                                                           |
+| isChecked           | `boolean`  | No       | false     | state of the checkbox, if explicitly set the checkbox will be controlled                                                                                 |
```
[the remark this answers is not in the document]
anuraghazra: Done

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+function AllOfThemHasToBeSelected() {
+  const [selected, setSelected] = React.useState([]);
+
+  return (
+    <CheckboxGroup
+      label="Do you agree to all the terms?"
+      helpText="Select all"
+      value={selected}
+      onChange={setSelected}
+      hasError={selected.length < 2}
```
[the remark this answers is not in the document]
anuraghazra: Correct.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Oh that Image is for "component architecture" I added that section initially in the decision file but to moved it to the PR later. https://github.com/razorpay/blade/pull/544#issue-1301661653

anuraghazra:
@kamleshchandnani addressed the comments you can check again.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+  hasError={hasError}
+  helpText={
+    hasError
+    ? 'Invalid username provided, contains special chars' 
+    : 'Enter valid username'
+  }
+/>
+```
+
+Will it be better to have a `errorText` prop instead which will automatically be switched internally is `hasError` is true?
```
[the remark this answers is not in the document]
anuraghazra: Discussed this already, we are keeping errorText on individual checkboxes too.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Not sure if we have the usecase. Also there is no 'event' for CheckboxGroup since it's custom managed state.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: @chaitanyadeorukhkar @nashcheez Let's keep isRequired. since now we have a better API `neccessityIndicator=required|optional` we can propagate it to the checkboxes to have better screen reader accessibility and set the native `required` prop.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: We will need these anyway. and now we decided to have validationState prop so it will stay as is. Let me know if you have any ideas to reduce the complexity further.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: We can do only string[] array now since for single checkboxes we can now use individual checkbox.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Update: we have changed hasError to validationState which takes `'error' | undefined` to be consistent with other Input Elements which takes `validationState=error|success|normal`

packages/blade/src/components/Checkbox/_decisions/decisions.md
saurabhdaware: Will renaming it to something generic help? like `caption={hasError ? 'boooo' : 'yay'}` sort of API? other alternatives for names `subText`, `text`, `validityMessage`, etc 🤔 ?

 > why do we have errorText at Checkbox group only and not individual checkbox level? is that intentional?

cause I also got confused in this part. I would imagine API on all components from design system to be similar otherwise people will have to remember each and every decision while working no?

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| labelPosition | `top, left`                 | No       | `top`       | The position of the rendered label                                                                                                 |
+| helpText    | `string`                    | No       | `undefined` | The helper text to be rendered                                                                                                     |
+| errorText    | `string`                    | No       | `undefined` | The error text to be rendered                                                                                                     |
+| defaultValue  | `string[]`                  | No       | `[]`        | The initial value of the checkbox group                                                                                            |
+| value         | `string[]`                  | No       | `[]`        | The value of the checkbox, if present will act as [controlled component](https://reactjs.org/docs/forms.html#controlled-components)                                                             |
```
saurabhdaware, line 71: Can we not figure out defaultValue from `value` prop itself? 🤔

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| onChange      | `(value: string[]) => void` | No       | `undefined` | The function to be called when any checkbox's state changes                                                                        |
+| name          | `string`                    | No       | `undefined` | The name of the checkbox group, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name) |
+| isDisabled    | `boolean`                   | No       | `false`     | Control whether the checkbox group is disabled or not.                                                                             |
+| validationState     | `undefined | 'error'`                   | No       | `undefined`     | Control whether the checkbox group is invalid or not.                                                                              |
+| necessityIndicator    | `'optional' | 'required' | undefined`                   | No       | `undefined`     | Renders optional or required indicators in the group, if set to undefined renders nothing |
```
saurabhdaware, line 76: Checkbox has `isRequired` but CheckboxGroup has this `necessityIndicator`. Can we standardise `isRequired` for everything?

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+
+1. Currently we only have `helpText` in Checkbox & CheckboxGroup and when we set `hasError` it becomes the `negative` variant. The problem is the content of the `helpText` should also change in certain scenarios to convey the exact error. 
+
+With this approach users will have to handle the logic
+
+```tsx
+const [hasError, setInvalid] = React.useState(false);
+
+<Checkbox 
+  hasError={hasError}
```
saurabhdaware, line 199: a bit confused. Don't see this prop in API reference chart above. Are we use `validationState: 'error'` or `hasError` 🤔 ?

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+  helpText={'Enter valid username'}
+  errorText={'Invalid username provided, contains special chars'}
+/>
+```
+
+**Conclusion:** No `errorText` on the `Checkbox` itself, we will only have it on `CheckboxGroup`.
```
saurabhdaware, line 220: I think this will be updated right https://github.com/razorpay/blade/pull/539/files#r924185463?

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: >  like caption={hasError ? 'boooo' : 'yay'}

yes but we will still need to know the error state. to be able to set proper aria-invalid & colors. 

> cause I also got confused in this part

We've changed this, now single checkbox will also have errorText

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: We can't because the CheckboxGroup can also have `(optional)` and `(required)` tags in the label. 

1. If in a form all inputs are required and only few are optional we will render optional in those instead of polluting all the inputs with required tag. 
2. Similarly if in a form all inputs are optional and only few are required we will only set required on those inputs only.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Right

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: value can be dynamically be set by the user and be changed. We can't effectively guess what user wants depending on if they provide onChange or not. 

For example. What If i want default values but also like to listen for onChange? 

```tsx
// this is default value or controlled value? 
<Checkbox value={true} /> 

// is this controlled or uncontrolled?
<Checkbox value={true} onChange={e => only want to listening to the change event} /> 
```

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: validationState. Will fix it.

packages/blade/src/components/Checkbox/_decisions/decisions.md
saurabhdaware: > We can't because the CheckboxGroup can also have (optional) and (required) tags in the label.

didn't understand this. So `isRequired={true}` will be same as `necessityIndicator="required"` and `isRequired={false}` will be same as `necessityIndicator="optional"` right?

> If in a form all inputs are required and only few are optional we will render optional in those instead of polluting all the inputs with required tag.
> Similarly if in a form all inputs are optional and only few are required we will only set required on those inputs only.

We can still do both of these from `isRequired` right? basically see the props of children and make decision based on that. I am guessing that is how we will do it even in `necessityIndicator` case?

saurabhdaware: Alsoo what happens in this case-
```ts
<CheckboxGroup neccessityIndicator="optional">
  <Checkbox isRequired={true} />
  <Checkbox isRequired={false} />
  <Checkbox isRequired={false} />
</CheckboxGroup>
```

Is the first checkbox here now required or optional?

anuraghazra: > didn't understand this. So isRequired={true} will be same as necessityIndicator="required" and isRequired={false} will be same as necessityIndicator="optional" right?

True but technically there are 3 states. 
The last one is the absence of `necessityIndicator` will render nothing. the `necessityIndicator` is mostly a visual prop for the group. and it happens to just propagate the isRequired prop for accessibility.

saurabhdaware: oh got it. makes sense.

What happens in this case btw https://github.com/razorpay/blade/pull/539#discussion_r924360265

anuraghazra: @saurabhdaware yes need to discuss on that. 

Currently if user set onChange, defaultChecked, isChecked on individual checkboxes inside the group we throw error. Maybe we also need to do this for isRequired.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+function AllOfThemHasToBeSelected() {
+  const [selected, setSelected] = React.useState([]);
+
+  return (
+    <CheckboxGroup
+      label="Do you agree to all the terms?"
+      helpText="Select all"
+      value={selected}
+      onChange={setSelected}
+      hasError={selected.length < 3 ? 'error' : undefined}
```
kamleshchandnani, line 174: this example needs to be updated with `validationState` prop? and we also need to add `errorText` prop?

packages/blade/src/components/Checkbox/_decisions/decisions.md
kamleshchandnani: can we update this open questions section? right now after our discussions this seems misleading

packages/blade/src/components/Checkbox/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Okay in that case we can add that in. Similar to how you have done in TextInput

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Updated.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| isIndeterminate     | `boolean`  | No       | false     | mixed state of the checkbox                                                                                                                              |
+| name (html native)  | `string`   | No       | undefined | The name of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name)            |
+| value (html native) | `string`   | No       | undefined | The value of the input field in a checkbox, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) |
+| isDisabled          | `boolean`  | No       | `false`   | Control whether the checkbox is disabled or not.                                                                                                         |
+| validationState           | `undefined | 'error'`  | No       | `undefined`   | Control whether the checkbox is invalid or not.                                                                                                          |
```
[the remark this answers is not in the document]
anuraghazra: Yes. github being weird.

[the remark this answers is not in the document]
anuraghazra: Resolved.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+## Design
+
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=13227%3A163026) to all variants of the Checkbox component
+
+## `Checkbox` API
+
+| Prop                | Type       | Required | Default   | Description                                                                                                                                              |
+| ------------------- | ---------- | -------- | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| children            | `string`   | No       | undefined | The text to be rendered as CheckboxLabel                                                                                                                 |
+| helpText          | `string`   | No       | undefined | The helper text to be rendered                                                                                                                           |
```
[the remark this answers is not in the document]
anuraghazra: Sure

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| isChecked           | `boolean`  | No       | false     | state of the checkbox, if explicitly set the checkbox will become [controlled component](https://reactjs.org/docs/forms.html#controlled-components)                                                                                 |
+| defaultChecked      | `boolean`  | No       | false     | default state of the checkbox, if isChecked is not provided the checkbox will become [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html)                                                            |
+| onChange            | `(isChecked: boolean) => void` | No       | undefined | The function to be called when the checkbox state changes.                                                                                               |
```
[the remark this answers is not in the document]
anuraghazra: Added: event, value, isChecked

packages/blade/src/components/Checkbox/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Not sure exactly. I think value should be fine since CheckboxGroup will always have multiple checkbox and works like a single Form Field. Similar to how <select multiple value=[] /> in native html.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+| children      | `React.ReactNode`             | Yes       | `undefined` | Accepts multiple Checkboxes                                                                                                        |
+| label         | `string`                    | Yes       | `undefined` | The label of the group                                                                                                             |
+| labelPosition | `top, left`                 | No       | `top`       | The position of the rendered label                                                                                                 |
+| helpText    | `string`                    | No       | `undefined` | The helper text to be rendered                                                                                                     |
+| errorText    | `string`                    | No       | `undefined` | The error text to be rendered                                                                                                     |
```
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,220 @@
...
+Example of only one checkbox needs to be checked:
+
+```tsx
+function AnyOneHasToBeSelected() {
+  const [selected, setSelected] = React.useState([]);
+
+  return (
+    <CheckboxGroup
+      label="Do you love any of the fruits?"
+      helpText="You can select multiple"
```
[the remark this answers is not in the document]
anuraghazra: Updated. Altho ideally we need content guidelines from design for these cases what to put in helperText and errorText.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Updated.

packages/blade/src/components/Checkbox/_decisions/decisions.md
kamleshchandnani: didn't get you? how can someone look at this image? apart from navigating directly from the directory?

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,224 @@
...
+| isDisabled          | `boolean`  | No       | `false`   | Control whether the checkbox is disabled or not.                                                                                                         |
+| isRequired          | `boolean`  | No       | `false`   | Control whether the checkbox is required or not.                                                                                                         |
+| helpText          | `string`   | No       | undefined | The helper text to be rendered                                                                                                                           |
+| errorText          | `string`   | No       | undefined | The error text to be rendered                                                                                                                           |
+| validationState           | `undefined \| 'error'`  | No       | `undefined`   | Control whether the checkbox is invalid or not.                                                                                                          |
```
kamleshchandnani, line 35: do we really need `undefined` as a value?

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Design
+
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=13227%3A163026) to all variants of the Checkbox component
+
+## `Checkbox` API
+
+| Prop                | Type       | Required | Default   | Description                                                                                                                                              |
+| ------------------- | ---------- | -------- | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| children            | `string`   | No       | undefined | The text to be rendered as CheckboxLabel                                                                                                                 |
```
kamleshchandnani, line 26: nit(OCD triggering): can you format all the `undefined` to look same? half of them are code formatted other half are not, follow one pattern please? :D

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,224 @@
...
+
+#### Validations
+
+And `validationState` will pass `aria-invalid` attributes to indicate invalid state.
+
+```tsx
+<Checkbox validationState={'error' | undefined}>is blade awesome?</Checkbox>
+```
+
+## `CheckboxGroup` API
```
kamleshchandnani, line 62: no `isRequired` prop on checkboxgroup? that shall be a general validation for all input fields. thoughts?

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,224 @@
...
+| isDisabled    | `boolean`                   | No       | `false`     | Control whether the checkbox group is disabled or not.                                                                             |
+| helpText    | `string`                    | No       | `undefined` | The helper text to be rendered                                                                                                     |
+| errorText    | `string`                    | No       | `undefined` | The error text to be rendered                                                                                                     |
+| validationState     | `undefined \| 'error'`                   | No       | `undefined`     | Control whether the checkbox group is invalid or not.                                                                              |
+| necessityIndicator    | `'optional' \| 'required' \| undefined`                   | No       | `undefined`     | Renders optional or required indicators in the group, if set to undefined renders nothing |
```
kamleshchandnani, line 77:
```suggestion
| necessityIndicator    | `'optional' \| 'required' \| undefined`                   | No       | `undefined`     | Renders `${label} (optional)` for `optional` and `${label} *` for `required`. if set to undefined renders nothing |
```

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,224 @@
...
+
+<Checkbox 
+  validationState={validationState}
+  helpText={'Enter valid username'}
+  errorText={'Invalid username provided, contains special chars'}
+/>
+```
+
+**Conclusion:**  
+Individual Checkboxes will have `errorText`, and if user sets `validationState` on the individual Checkbox when inside the CheckboxGroup we will throw a runtime error.
```
kamleshchandnani, line 224:
```suggestion
Individual Checkboxes will have `errorText`, and if user sets `validationState` on the individual Checkbox when inside the CheckboxGroup we will throw at runtime error.
```

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Nope. no required prop on checkbox. Passing down required prop from checkboxgroup to individual checkboxes doesn't make sense since we don't know if user wants all/some or atleast one to be required or not. 

We only have neccessityIndicator=required

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: They won't cuz I'll delete it 🤣 There's no reference to this image anymore in the Decision file. 

We had before when [we had the "component architecture" section.](https://github.com/razorpay/blade/blob/91c15c6bd2ce889f4ec866499dc028e0fd22120c/packages/blade/src/components/Checkbox/_decisions/decisions.md#component-architecture)

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Yes we should add undefined as value. Since we are treating it as a value. 

For example in code also we should do this: 

```ts
// explicitly setting undefined in the union
type CheckboxProps = {
  validationState?: 'error' | undefined;
}
```

**But Why?** 

because if consumers set [exactOptionalPropertyTypes](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) in their tsconfig.json they will get error that undefined is not assignable. 

```tsx
<Checkbox validationState={undefined}>Hello</Checkbox>;
^
Type '{ children: string; validationState: undefined; }' is not assignable to type 'CheckboxProps' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Types of property 'validationState' are incompatible.
    Type 'undefined' is not assignable to type '"error"'.ts(2375)
```

**[More info in this Slack Thread ](https://razorpay.slack.com/archives/C02FRV92MPH/p1651860634820809?thread_ts=1651851415.416079&cid=C02FRV92MPH)**

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Sure.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/Checkbox/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Added

packages/blade/src/components/Checkbox/_decisions/decisions.md
kamleshchandnani: okay then if that's the case i would recommend to have a semantic value for undefined state(and we can use this everywhere then even in text input I'll update).
basically,

```
validationState = 'error' | 'success' | 'none'
```

that would be better to write in code too

```
validationState={text.length < 10 ? 'error' : 'none'}
```

kamleshchandnani:
@anuraghazra would suggest to add architectural image as part of the markdown. it was giving better visual corelation

packages/blade/src/components/Checkbox/_decisions/decisions.md
anuraghazra: Yes that's what I'm also pushing for https://github.com/razorpay/blade/pull/552#discussion_r926472221 

'none' seems fine. cc @nashcheez

anuraghazra:
Oh haha okay sure. Will put it in after the anatomy?

packages/blade/src/components/Checkbox/_decisions/decisions.md
```diff
@@ -0,0 +1,228 @@
...
+| name          | `string`                    | No       | `undefined` | The name of the checkbox group, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name) |
+| isDisabled    | `boolean`                   | No       | `false`     | Control whether the checkbox group is disabled or not.                                                                             |
+| helpText    | `string`                    | No       | `undefined` | The helper text to be rendered                                                                                                     |
+| errorText    | `string`                    | No       | `undefined` | The error text to be rendered                                                                                                     |
+| validationState     | `'none' \| 'error'`                   | No       | `undefined`     | Control whether the checkbox group is invalid or not.                                                                              |
```
kamleshchandnani, line 80:
```suggestion
| validationState     | `'none' \| 'error'`                   | No       | `none`     | Control whether the checkbox group is invalid or not.                                                                              |
```

anuraghazra: fixed

kamleshchandnani (approved):

PR #542, opened 2022-07-07
feat: add internal Box component

chaitanyadeorukhkar:
Add Box component only for internal usage. We can think about exposing this to the consumers at a later point


packages/blade/src/components/Box/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './Box';
```
anuraghazra, line 1: Maybe let's use named exports.
It's 10x easier to reason about named import and avoid bugs and confusion. 

```
export * from './Box';
```

packages/blade/src/components/Box/Box.d.ts
```diff
@@ -0,0 +1,21 @@
...
+  justifyContent?: CSSObject['justifyContent'];
+  alignSelf?: CSSObject['alignSelf'];
+  paddingTop?: DotNotationSpacingStringToken;
+  paddingBottom?: DotNotationSpacingStringToken;
+  paddingLeft?: DotNotationSpacingStringToken;
+  paddingRight?: DotNotationSpacingStringToken;
+  minHeight?: number;
+  minWidth?: number;
+  maxHeight?: number;
+  maxWidth?: number;
```
anuraghazra, line 18: We don't need `width`, `height`?

packages/blade/src/components/Box/__tests__/__snapshots__/Box.web.test.tsx.snap
```diff
@@ -0,0 +1,36 @@
...
+  min-height: 48px;
+  min-width: 48px;
+  max-height: 56px;
+  max-width: 56px;
+}
+
+<div>
+  <div
+    class="c0"
+    display="flex"
```
anuraghazra, line 33: The display probably should not get leaked into the dom.

packages/blade/src/components/Box/index.ts
anuraghazra: In consumer side we are anyway doing named export. That's no the issue. 

```
export { default as Box } from './Box' 
```

The main thing is the internal blade components. 
- With default exports it's hard to reason about exports
- If we rename a component we won't get any error that we renamed and might continue using the older name


```
// Box.d.ts
export { default as Box } from './Box.d';
export * from './Box';
```

vs

```
// Box.d.ts
export * from './Box';
```

packages/blade/src/components/Box/__tests__/__snapshots__/Box.web.test.tsx.snap
anuraghazra: Just doesn't feel right to pollute the html elements, in certain cases I believe react also warns about these if they overlap with any native props. 

It's just weird why styled-components work like this, They have transient props but they [look super ugly](https://styled-components.com/docs/api#transient-props)

packages/blade/src/components/Box/index.ts
chaitanyadeorukhkar: >With default exports it's hard to reason about exports

Not sure what you mean here

>If we rename a component we won't get any error that we renamed and might continue using the older name

Yes, that's a blessing and a curse 😅 


Take a look at BaseText or Icons

We do
```
import BaseText from '../../BaseText';
import CreditCardIcon from '../../CreditCardIcon';
```

Honestly, I don't mind named exports - just as long as we do the same things everywhere. 

If we have a strong enough reason to move to named exports, lets do that everywhere. But don't want to do a default export in some places and named in others. Lets take this up separately on whether or not we want to move everything to named exports? That would require change in multiple places and should be handled in another PR

anuraghazra: Yeah we can decide upon that separately and should also add eslint no default exports once we go full in with named exports.

packages/blade/src/components/Box/__tests__/__snapshots__/Box.web.test.tsx.snap
chaitanyadeorukhkar: > Just doesn't feel right to pollute the html elements

Agreed. I'm guessing we'll face this issue as we add more things to props and they start showing up in the DOM 😢

chaitanyadeorukhkar: Transient props do look ugly. 

We could do this
```
import type { BoxProps } from './Box.d';
import getBoxStyles from './getBoxStyles';

const StyledBox = styled.div<Omit<BoxProps, 'display'> & { $display: BoxProps['display'] }>(
  ({ $display, ...props }) => ({
    ...getBoxStyles({ ...props, display: $display }),
  }),
);

const Box = ({ display, ...props }: BoxProps): ReactElement => {
  return <StyledBox $display={display} {...props} />;
};

export default Box;
```

But I'm not sure if it's worth it. Lets move to doing this if it causes any issues?

packages/blade/src/components/Box/index.ts
chaitanyadeorukhkar: Yeah, lets have a discussion around this. Would like to understand better why we'd prefer named exports for Box, BaseText, Icons, etc over default

packages/blade/src/components/Box/__tests__/__snapshots__/Box.web.test.tsx.snap
anuraghazra: Yeah for now let's skip it we can relook it later. I've seen some solutions in stackoverflow that destructuring props while accepting them also works.

packages/blade/src/components/Box/Box.web.tsx
```diff
@@ -0,0 +1,7 @@
+import styled from 'styled-components';
+import type { BoxProps } from './Box.d';
+import getBoxStyles from './getBoxStyles';
+
+const Box = styled.div<BoxProps>(getBoxStyles);
```
kamleshchandnani, line 5: This will always add a dummy new element(`div` in this case) in the dom whenever wrapped around anything?

chaitanyadeorukhkar: Yes, there will be a div added. The idea is to use this as a wrapper in places and not as an enhancer.

packages/blade/src/components/Box/index.ts
divyanshu013: I would personally prefer using default exports for this purpose 😄

divyanshu013:
[This one](https://github.com/razorpay/blade/pull/542#discussion_r926512301) I think we should confirm / fix before merging. Rest looks good to me.

packages/blade/src/components/Box/__tests__/__snapshots__/Box.web.test.tsx.snap
chaitanyadeorukhkar: Don't see any sort of warnings in the console 🤔

packages/blade/src/components/Box/Box.web.tsx
chaitanyadeorukhkar: We can but lets avoid doing it unless we see valid use-cases

packages/blade/src/components/Box/Box.d.ts
```diff
@@ -0,0 +1,21 @@
+import type { CSSObject } from 'styled-components';
+import type { DotNotationSpacingStringToken } from '../../_helpers/types';
+
+export type BoxProps = {
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: No reason to not add more props here as we see more valid use-cases. We're starting with props we need _now_ and can extend this later

packages/blade/src/components/Box/Box.web.tsx
kamleshchandnani: what does as="tag" do?

kamleshchandnani (approved):

packages/blade/src/components/Box/Box.web.tsx
divyanshu013: Renders the component using the html primitive specified, for example `<Box as="h1">`

packages/blade/src/components/Box/__tests__/__snapshots__/Box.web.test.tsx.snap
divyanshu013: Does it render `flex` property on the div in DOM?

chaitanyadeorukhkar: It has display as flex but doesn't show in the DOM
Screenshot 2022-07-21 at 5 51 06 PM

divyanshu013 (approved):
Looks good to me.

PR #549, opened 2022-07-13
feat(tokens): add new tokens

rzpcibot:

### DIFF
--- a/packages/blade/src/tokens/theme/theme.d.ts
+++ b/packages/blade/src/tokens/theme/theme.d.ts
@@ -50,6 +50,10 @@ export type ActionStates = {
   disabled: string;
 };
 
+export type LinkActionStates = ActionStates & {
+  visited: string;
+};
+
 export type ActionStatesWithContrast = {
   default: ColorContrast;
   hover: ColorContrast;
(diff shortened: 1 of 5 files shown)


chaitanyadeorukhkar (dismissed):

packages/blade/src/tokens/theme/theme.d.ts
```diff
@@ -50,6 +50,10 @@ export type ActionStates = {
   disabled: string;
 };
 
+export type LinkActionStates = ActionStates & {
```
chaitanyadeorukhkar, line 53: @kamleshchandnani can you take a look at this once? Have updated the types to accommodate a new `visited` token

packages/blade/src/tokens/theme/theme.d.ts
```diff
@@ -96,7 +96,7 @@ export type FeedbackActions = {
 
 export type Colors = {
   brand: {
-    primary: Record<300 | 400 | 500 | 600 | 700, string>;
+    primary: Record<300 | 400 | 500 | 600 | 700 | 800, string>;
```
chaitanyadeorukhkar, line 99: This update came from the plugin. I think we added 800 to brand primary colors

kamleshchandnani (approved):

chaitanyadeorukhkar (approved):

PR #552, opened 2022-07-18
docs(blade): add input api decisions

kamleshchandnani:
## [View Formatted File](https://github.com/razorpay/blade/blob/feat/input-field/packages/blade/src/components/Input/_decisions/decisions.md)


packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+| validationState                                                                                    | `error` , `success`                                                       | No                                                                                  |                                                           | Applies the style to the input field based on validation status and shows either `errorText` or `successText` respectively                                                                                                                                                            |
```
anuraghazra, line 61: validationState will be error or success right? and If not present (undefined) it will be nothing. 

We are also doing the same on Checkbox. That if validationState is undefined we set nothing. 

Why not have a common value for that "undefined"? Maybe like `idle` or `default`

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+| type                                                                                               | `text`, `search`, `tel`, `email`, `url`, `numeric`                        | No                                                                                  | `text`                                                    | type of text to be rendered. this will also determine the keyboard type to be show on mobile devices                                                                                                                                                                                  |
```
anuraghazra, line 45: What should we do for `numeric` input? 

Do we show those up/down spinner icons? Do we have designs for this?

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+| label                                                                                              | `string`                                                                  | (confirm if we have a use case for input field without labels(search field maybe?)) |                                                           | Determines the label of an input field. Also used as `aria-label`                                                                                                                                                                                                                     |
```
saurabhdaware, line 46: Can we do some typescript magic to make `ariaLabel` property required when `label` is not given? like one of `label` or `ariaLabel` is required 🤔

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+| -------------------------------------------------------------------------------------------------- | --------------------------------- | ----------------------------------------------------------------------------------- | --------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| label                                                                                              | `string`                          | (confirm if we have a use case for input field without labels(search field maybe?)) |                                                           | Determines the label of an input field. Also used as `aria-label`                                                                                                                                                                                                                     |
```
saurabhdaware, line 113: we can make this required in other inputs other than `TextInput` right? For example, I don't see usecase of `PasswordInput` not requiring label.

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+| <kbd>Previous</kbd> | `enterkeyhint="previous"` | `returnKeyType="previous"`(android only) |
+| <kbd>Search</kbd>   | `enterkeyhint="search"`   | `returnKeyType="search"`                 |
+| <kbd>Send</kbd>     | `enterkeyhint="send"`     | `returnKeyType="send"`                   |
+
+## PasswordInput
+
+**Sample Usage**
+
+```tsx
+<PasswordInput
```
saurabhdaware, line 102: How do we handle "Show Password" option here?

image

packages/blade/src/components/Input/_decisions/decisions.md
nashcheez, line 46: I think we should make `label` a required field for `TextInput`. Design systems generally have a different component like Search or QueryField to be used for the SearchField. It's better to keep them separate as they serve different purposes and most likely the SearchField will have additional props. In my previous experience, we had to learn from our mistakes and then built it out separately [here](https://design.gs.com/components/search). 😄

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+| isDisabled                                                                                         | `boolean`                                                                 | No                                                                                  | `false`                                                   | Marks the input field as disabled. It'll also loose keyboard focus.                                                                                                                                                                                                                   |
```
nashcheez, line 52: Also, `aria-disabled` should be set to `true` here.

packages/blade/src/components/Input/_decisions/decisions.md
nashcheez: `undefined` for a state is a good enough value, what's the benefit of adding an extra value?

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+| fullWidth                                                                                          | `boolean`                                                                 | No                                                                                  | `false`                                                   | **Discuss if we want to add support for full-width input fields.**                                                                                                                                                                                                                    |
```
nashcheez, line 66: When we say `fullWidth` what do we mean? Are we going to provide a large fixed width or a `min-width` to the input field (assuming the normal input field would be have a `width: 100%` and would take the width of its container).

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+- character counter(only input field)
+- clear icon button
+- prefix/suffix
+
+## Open questions
+
+- [ ] Do we have a usecase of using textfield without label or form level things?
+- [ ] Do we need to expose `autoCapitalize` for native?
+- [ ] Do we need to expose `onSubmitEditing` for native?
+- [ ] Do we need to fix the height for reserving space for help/error/success text?
```
nashcheez, line 147: I think we should not fix the height and let this wrap, there could be sensitive/important information as part of an input field validation that would need to be displayed fully.

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+- [ ] motion for the textfield animation in active state
+- [ ] fullWidth inputField do we have a use case?
+- [ ] textContentType for autocomplete on native ios only
+  - https://reactnative.dev/docs/textinput#textcontenttype-ios
+- [ ] what is the default position for loader in loading state? following are some options
```
nashcheez, line 152: Generally should be towards the right end of the input so as to not obscure the text that could be in the input field itself. If a clear button is present it could replace the clear button and show the loading state (as there is no usage of the clear button during the loading state). Putting it on the left will cause the text to move to the right when the loader appears if we decide to have the text during loading.

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+| isDisabled                                                                                         | `boolean`                         | No                                                                                  | `false`                                                   | Marks the input field as disabled. It'll also loose keyboard focus.                                                                                                                                                                                                                   |
+| showClearButton                                                                                    | `boolean`                         | No                                                                                  | `false`                                                   | Defines if we want to show the clear icon button on the right side of the password field                                                                                                                                                                                              |
```
nashcheez, line 120: Should we be showing a clear button on the password field? I think we already have the eye icon towards the right for our password fields, adding an extra icon would make it a little messy. I don't see the clear button in our designs though.

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,212 @@
...
+## TextInput API
+
+**Sample Usage**
+
+```tsx
+<TextInput
+  label="Enter Name"
+  labelPosition="left"
+  placeholder="Enter your name"
+  leadingIcon={UserIcon}
```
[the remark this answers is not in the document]
nashcheez: +1 if it has to be only icons.

But should the content only be limited to icons? I think there can be many use cases where consumers would want to have text instead. Could be `@gmail.com`, `Rs.`, telephone codes as `+91` etc. I think in that case we should have props `trailingContent` & `leadingContent` that can accept `Icon` or `string`.

packages/blade/src/components/Input/_decisions/decisions.md
nashcheez: On second thoughts, I can see consumers would want to use the input fields in conjunction with data tables or as filters where they may want to skip the label. Could make it optional, entirely depends on the guidelines we would want to set.

packages/blade/src/components/Input/_decisions/decisions.md
kamleshchandnani: it'll automatically be placed for all the password input fields, user don't have to decide that, system will take that decision

packages/blade/src/components/Input/_decisions/decisions.md
kamleshchandnani: making it required for now. once we have a use case for making it optional we can add that as an enhancement. changing required to optional is an easy migration path but reverse is not true and scary 😄

packages/blade/src/components/Input/_decisions/decisions.md
kamleshchandnani: yes it was pending on design, we will remove this prop and all our input fields will be width:100% to the container.

packages/blade/src/components/Input/_decisions/decisions.md
kamleshchandnani: pending on design, saurav will test this visually and conclude

packages/blade/src/components/Input/_decisions/decisions.md
kamleshchandnani: clear button will only be shown when the input field is in active state so i think should be fine vs not providing it at all which is valid case for all the input fields

packages/blade/src/components/Input/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: update based on design: we decided to only accept `icon` as a prop and always fix it to left because that's the general design decision we want to impose. right side is always preserved for showing only actions(show password, clear button) on and suffix content like `.00` or `@gmail.com`

>I think in that case we should have props trailingContent & leadingContent that can accept Icon or string.

yes we do have `prefix` and `suffix` props already listed in the table. this is just a minimalistic example for demo purpose

packages/blade/src/components/Input/_decisions/decisions.md
kamleshchandnani: check the table where I've mentioned the input type and their corresponding input mode. this case is covered there. we don't show those up/down arrow

packages/blade/src/components/Input/_decisions/decisions.md
kamleshchandnani: why do we need a value for that? if it doesn't exist it's not there. and for cases where you want to set it to none we can do this always right

```
validationState={text.lenght < 10: 'error' : undefined}
```

will adding another state will make sense?

packages/blade/src/components/Input/_decisions/decisions.md
kamleshchandnani: yes it was pending on design

packages/blade/src/components/Input/_decisions/decisions.md
anuraghazra: undefined is fine. Just that we are implicitly agreeing that undefined is a value, it's a bit harder to communicate the intent with undefined.

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+| isDisabled                                                                                        | `boolean`                                                                                                                                                                                                                                              | No       | `false`                                                   | Marks the input field as disabled. It'll also set `aria-disabled` to true and hence the input field will loose keyboard focus.                                                                                                                                                        |
```
anshulsahni, line 49: **Question:** Possible to have an alias to this properly -  `disabled` - since this would be in conjuncture to the original HTML API to <input/> element

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+| onClearButtonClicked                                                                              | `(event) => {}`                                                                                                                                                                                                                                        | No       |                                                           | Used in conjunction with `showClearButton` prop. Provides the capability to the user to handle/control the behavior of clear icon button. Can be used to send analytics event and can also be used to clear the content when the input field is `controlled`                          |
```
anshulsahni, line 54: **Question:** Will this override the functionality or extend it?

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+| helpText                                                                                          | `string`                                                                                                                                                                                                                                               | No       |                                                           | Shown when we want to add some hint to the input field. Displayed under the input field. Only one of `helpText`, `errorText` or `successText` is shown at a time in the priority order as `errorText`, `successText`, `helpText`                                                      |
```
anshulsahni, line 61: **Suggestion:** Removing the help text in case there is an error or some other message visible, doesn't seem like a good experience. If user is entering invalid value then this help text would help the user in entering the correct value but with correct flow the error text would remove the help text from screen thus increasing the chances of invalid text more

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+| successText                                                                                       | `string`                                                                                                                                                                                                                                               | No       |                                                           | Shown when the `validationState` of the input field is set to Success. Only one of `helpText`, `errorText` or `successText` is shown at a time in the priority order as `errorText`, `successText`, `helpText`                                                                        |
```
anshulsahni, line 63: **Question:** Why no option of warningState?

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+| necessityIndicator                                                                                 | `optional`, `required`                                                    | No       |                                                           | Indicator to be shown next to the label for emphasizing the necessity of the field. if `optional` is provided then it'll show `${label} (optional)` and if `required` is provided then it'll show `${label} *`                                                                        |
+| icon                                                                                               | `Icon`                                                                    | No       |                                                           | Icon to be displayed at the start of the input field.                                                                                                                                                                                                                                 |
```
anshulsahni, line 108: **Question:** Do we have `<Icon/>` component already, if not, in that case till then are thinking to accept in this prop?

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+| isLoading                                                                                          | `boolean`                                                                 | No       | `false`                                                   | Used to show a spinner when the the data is being loaded. **Discuss the position of the loader. Check options mentioned in the open questions below**                                                                                                                                 |
+| maxCharacters                                                                                      | `number`                                                                  | No       |                                                           | Used to show a character counter under the input field                                                                                                                                                                                                                                |
```
anshulsahni, line 114: **Question:** Same question as for `isDisabled`, possible to have an alias which is the same as the original API maxlength?

anshulsahni (commented):
Added few comments

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+| autofillSuggestionType                                                                            | `password`, `passwordNew`, `none` | No       |                                                           | Specifies autocomplete hints for the system, so it can provide autofill                                                                                                                                                                                                               |
+| keyboardReturnKeyType [Check this table for full reference](#return-key-types-for-web-and-native) | `go`, `done`, `next`, `send`      | No       | Closest based on the `type` attribute                     | Determines how the return key should look on the keyboard on mobile devices or virtual keyboard                                                                                                                                                                                       |
+
```
divyanshu013, line 187: What about usecase for showing what you type in password field? Like the eye icon we generally see 👁️

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+- label
+- helpText/errorText/successText
+- character counter(only input field)
+- clear icon button
+- prefix/suffix
+
+## Open questions
+
+- [ ] Do we need to expose `autoCapitalize` for native?
```
divyanshu013, line 200: I think it can be useful because the default is capitalize first letter of each sentence

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+  - confirm once with mobile devs. design side we shall not touch this
+- [ ] Do we need to expose `onSubmitEditing` for native?
```
divyanshu013, line 202: I would presume yes because it can be helpful UX for changing focus to the next input field while filling a form (since Tab key doesn't work in mobile)

pritishvaidya: This is required for submission of keyboard done button to trigger an action

packages/blade/src/components/Input/_decisions/decisions.md
pritishvaidya: It should be added since native already provides a prop for it without need to calculate the same in JS to avoid jitter

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+| placeholder                                                                                       | `string`                                                                                                                                                                                                                                               | No       | Default to the label if label provided else `Enter Value` | Placeholder text to be displayed inside the input field                                                                                                                                                                                                                               |
```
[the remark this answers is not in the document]
divyanshu013: In that case shouldn't we make placeholder required? Otherwise if people leave this as it's it might unintentionally creep into UIs as "Enter value" 🤔

packages/blade/src/components/Input/_decisions/decisions.md
anshulsahni: @kamleshchandnani few work-around I can think of at this time
 is that incase where helptext & errorText both are going configured,
 -  there help text can be made part of something like a under info symbol or a tooltip - _something which isn't visible all thetim_
 - help text in some way can be configured with error text like a very small cta with error text `show help` so that the user has the scope of seeing the help text even when input is in error state
 - Error text dismissable so that after dismissing the error message user can see the help text

packages/blade/src/components/Input/_decisions/decisions.md
anshulsahni: Well, I don't know whether it's implemented or not, but in certain places in our system amount is accepted in paise and in some cases it's in Rupees, so can be used in one of those cases to warn the users

packages/blade/src/components/Input/_decisions/decisions.md
anshulsahni: ooops sorry 😅

packages/blade/src/components/Input/_decisions/decisions.md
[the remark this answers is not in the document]
anshulsahni: +1 to @divyanshu013's point, if we really want to make placeholder necessary then we should make the prop mandatory & you feel people will ignore type-check then just throw exception if the prop is not present

packages/blade/src/components/Input/_decisions/decisions.md
nikhilk1993: +1 `onSubmitEditing` needs to be added..used in both pg & x apps

nikhilk1993: `autoCapitalize` also needs to be added.. use cases are for activation forms where we 
capitalise characters for ifsc (autoCapitalize="characters")

nikhilk1993: @shubhnik we have used `contextMenuHidden` prop in AmountInput..do you think that is required?

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+- [OTPInput]()
+- [CardInput]()
+- [TextArea]()
+
+## BaseInput API
+
+The BaseInput is the underlying input field that will in turn be used to render all types of input. For now this will be an **internal component** and won't be exposed to consumers. We need this so we can have the rendering of text input and interactions at one central place but can restrict/add more control to consumer components based on their type.
+
+| Prop                                                                                              | Type                                                                                                                                                                                                                                                   | Required | Default                                                   | Description                                                                                                                                                                                                                                                                           |
```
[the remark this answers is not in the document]
kamleshchandnani: >autoCapitalize
What is the use case you have seen for this?

>onSubmitEditiong
what is the use case for this? people adding information and hitting the submit/return button on the keyboard and the form shall submit?

>maxLength
same as maxCharacters

>multiline
will evaluate once we reach text area. this is right now scoped to TextInput and PasswordInput

>editable
covered as `readOnly`

>placeholderTextColor
why do we need this? ideally this would be automatically handled by the input field based on the other colors. it's the basic styling that isn't configurable by the consumer.

packages/blade/src/components/Input/_decisions/decisions.md
kamleshchandnani: >* there help text can be made part of something like behind a info symbol or a tooltip - something which isn't visible all thetim
>* help text in some way can be configured with error text like a very small cta with error text show help so that the user has the scope of seeing the help text even when input is in error state

both the options will lead the user to perform an extra step for checking help text, we want it to be upfront.

>Error text dismissable so that after dismissing the error message user can see the help text
I guess this indeirectly will always exist, once the user has validated the actual text the help text will appear again.


also, what are the use cases where you think the error text will be out of context of help text? for eg:
If the helptext for password will be "Password must contain 8 characters" the error text will automatically be in that context already "Password should be of atleast 8 characters"

packages/blade/src/components/Input/_decisions/decisions.md
kamleshchandnani: we'll enhance later on then once we have a use case

packages/blade/src/components/Input/_decisions/decisions.md
```diff
@@ -0,0 +1,295 @@
...
+<PasswordInput
+  label="Enter Password"
+  labelPosition="top"
+  placeholder="Enter your password"
+  errorText="Password cannot be blank"
+  validationState={!inputValue.length ? 'error' : undefined}
+/>
+```
+
+| Prop                                                                                              | Type                              | Required | Default                                                   | Description                                                                                                                                                                                                                                                                           |
```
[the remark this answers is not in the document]
kamleshchandnani: secureTextEntry will be automatically set when you use `<PasswordInput/>`

packages/blade/src/components/Input/_decisions/decisions.md
anshulsahni: We're going with assumption that error text & help text would be related, but that's not going to be the case most of the time.

Help text always doesn't contain the information about the validation, let's say someone is writing TAN Number, now help text could be telling the user which document one should check for looking for getting their TAN number. There could be an error that `TAN is not valid` - _this could be due to some static validation_, in this case, we removed the help text and made visible the error text. Here our above assumption doesn't fit

kamleshchandnani:
Merging this, will raise another PR for further enhancements

PR #560, opened 2022-07-21
feat: add support for CSS theme variables

chaitanyadeorukhkar:
**Generates CSS variables by theme & device:**
- paymentThemeLightDesktop
- paymentThemeLightMobile
- paymentThemeDarkDesktop
- paymentThemeDarkMobile
- bankingThemeLightDesktop
- bankingThemeLightMobile 
- bankingThemeDarkDesktop
- bankingThemeDarkMobile 

**How does this work?**
1. Rollup generates a `generated/themeBundle.js` with `rollup-themes.config` which converts our typescript themes into `cjs`
2. During build, we call the script `generateCssThemeVariables`
3. `generateCssThemeVariables` prepares `paymentTheme` & `bankingTheme` to be used as inputs for [`style-dictionary`](https://amzn.github.io/style-dictionary/). Style Dictionary expects the theme input to be in a [specific format](https://amzn.github.io/style-dictionary/#/tokens?id=design-tokens).
4. We generate 8 CSS outputs using 8 different configs of Styled Dictionary
5. These 8 CSS files are generated into `build/css` which makes them available for `npm publish`

**Pending:**
- ~Can we club mobile and desktop themes together? ` --typography-on-desktop-fonts-size-10` & ` --typography-on-mobile-fonts-size-10`?~ Discussed with the consumer team and concluded that having 2 different files would be more convenient for them
- ~How do we handle units for different tokens? `--border-radius-small` is `2` but should we ship it as `--border-radius-small: 2px`?~ We are now adding units where its needed

**Publish dry run:**
Screenshot 2022-07-21 at 5 42 26 PM

**Example:**
Screenshot 2022-07-22 at 12 43 11 PM


packages/blade/scripts/generateCSSTokens.js
```diff
@@ -0,0 +1,169 @@
+const { getLeaves } = require('any-leaf');
+const StyleDictionary = require('style-dictionary');
+const set = require('lodash/fp/set');
```
chaitanyadeorukhkar, line 3: I am using the [Function Programming version](https://github.com/lodash/lodash/wiki/FP-Guide) of `set` which does not mutate the original theme object. I spent quite some time figuring out how to avoid this mutation causing issues while generating other themes but couldn't really work it out. Non-mutative method is much more straightforward

kamleshchandnani: did you try using deepClone from lodash? that deep clones object and then you can use it regularly and it'll not have any mutations

packages/blade/package.json
```diff
@@ -37,11 +37,13 @@
     "build": "run-s build:*",
     "build:clean": "rm -rf build",
     "build:generate-types": "tsc -p ./tsconfig-generate-types.json",
+    "build:generate-css-tokens": "FRAMEWORK=REACT rollup -c ./rollup-themes.config.js && node ./scripts/generateCSSTokens.js",
     "build:copy-declarations": "copyfiles -u 1 \"src/**/*.d.ts\" build/types",
     "build:react": "FRAMEWORK=REACT rollup -c && node ./scripts/generateRootImports.js",
     "build:react-native": "FRAMEWORK=REACT_NATIVE rollup -c && node ./scripts/generateRootImports.js",
     "build:declarations": "rollup -c && node ./scripts/generateRootImports.js",
     "build:clean-declarations": "rm -rf build/types",
+    "build:clean-theme-bundle": "rm -rf generated",
```
kamleshchandnani, line 46: why not use `build/css-vars`?

packages/blade/rollup-themes.config.js
```diff
@@ -0,0 +1,40 @@
...
+  '.mobile.ts',
+  '.mobile.tsx',
+  '.js',
+  '.ts',
+  '.jsx',
+  '.tsx',
+  '.mjs',
+];
+
+export default {
```
kamleshchandnani, line 23: why do you need this? you can add this config to the existing rollup config too?

packages/blade/package.json
chaitanyadeorukhkar: This only cleans the themeBundle.js that is created by rollup. We dont need to publish it so it can't be in build folder

packages/blade/rollup-themes.config.js
chaitanyadeorukhkar: Makes sense. Done

packages/blade/package.json
kamleshchandnani: i do that for types as well. I've build/types temporary and then clean it.

packages/blade/scripts/generateCSSTokens.js
chaitanyadeorukhkar: I'll check this

chaitanyadeorukhkar: Yup this works well!

packages/blade/scripts/generateCssThemeVariables.js
```diff
@@ -0,0 +1,196 @@
...
+            destination: `${outputFileName}.css`,
+            format: 'css/variables',
+          },
+        ],
+      },
+    },
+  };
+};
+
+const makeValue = ({ path, value }) => {
```
divyanshu013, line 47: We're doing some fuzzy checks here with `.includes`. Can this cause any issues considering there are ton of css properties? For example - say we've border.color, border.radius -> both pass for the first check 🤔

divyanshu013 (commented):
Let's also add some context on why we're generating these CSS theme variables in the PR description. Also, with CSS variables we're enforcing some minimum browser versions that will be supported. Just wanted to confirm if that is a conscious decision.

divyanshu013:
A couple of questions in last review, otherwise looks good to me.

packages/blade/scripts/generateCssThemeVariables.js
chaitanyadeorukhkar: Yeah, it could cause issues. We will have to put some kind of checks somewhere in order to parse it. Fuzzy checks seem much easier at the moment, if we see more regressions coming in the future we can add more strict checks for parsing values.

chaitanyadeorukhkar: I'll fix the border issue now

chaitanyadeorukhkar:
@divyanshu013 It's available for 96.51% of browsers. Compatibility will anyway be handled by consumers who would need these since we're not using CSS variables anywhere on blade. I will still add this in usage guidelines.

Screenshot 2022-07-28 at 11 14 22 AM

divyanshu013 (approved):
Looks good to me.

chaitanyadeorukhkar:
@divyanshu013 @kamleshchandnani I've also some documentation for CSS variables. Can you take a look?

packages/blade/docs/tokens/CssVariables.stories.mdx
```diff
@@ -0,0 +1,813 @@
...
+```
+
+> **Note:** Since we depend on CSS variables for our tokens, please ensure that you check for the minimum supported browser versions for CSS variables and handle that approriately.
+> Refer: [caniuse.com/css-variables](https://caniuse.com/css-variables)
+
+## Example
+
+This is an example CSS file for `paymentThemeLightDesktop.css`
+
+> **Note**: This is not the actual `paymentThemeLightDesktop.css` but just a dummy version of it. The actual values of these variables can change over time as we evolve our themes & tokens.
```
divyanshu013, line 81: My recommendation here is to avoid including the actual values below, makes the doc too long (and also harder to maintain) and people interested in knowing the actual variables and values would rather look at the actual file. I think few properties as an example should be fine to give a general idea.

packages/blade/docs/tokens/CssVariables.stories.mdx
```diff
@@ -0,0 +1,813 @@
...
+
+```css
+:root {
+  --border-radius-none: 0px;
```
divyanshu013, line 85: Maybe include few properties here and add a comment:

```css
:root {
  /* mention some properties */

  /* and so on... */
}
```

divyanshu013 (approved):
Looks good to me. A minor recommendation.

packages/blade/docs/tokens/CssVariables.stories.mdx
chaitanyadeorukhkar: Makes sense. Done!

packages/blade/rollup.config.js
```diff
@@ -123,9 +124,35 @@ const getDeclarationsConfig = ({ exportCategory }) => ({
...
+    }),
+    aliases,
+  ],
+});
+
 const config = () => {
   const framework = process.env.FRAMEWORK;
+  const generateCSSVariables = process.env.GENERATE_CSS_VARIABLES;
 
+  if (generateCSSVariables == 'true' && framework === 'REACT') {
```
nashcheez, line 151: nit: @chaitanyadeorukhkar Any specific reason to not do a strict check here for `generateCSSVariables`?

chaitanyadeorukhkar: No reason. This is derived from env variables which in general are always transformed into a string.

PR #573, opened 2022-07-26
docs(Alert): add API design and decisions

divyanshu013:
Contains API design and decisions.

- Please read [formatted document](https://github.com/razorpay/blade/blob/div/alerts-decisions/packages/blade/src/components/Alert/_decisions/decisions.md)


packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,162 @@
...
+| isFullWidth   | `boolean`                                | `false`     | Spans the entire width of container, otherwise max width is restricted to 584px by default |          |
+| variant       | `info`, `positive`, `notice`, `negative` | `info`      | Sets the color tone of entire alert. Icon is set automatically based on variant \*         |          |
+| isBorderless  | `boolean`                                | `false`     | Removes borders and border radii, useful for creating a full bleed layout                  |          |
+
+We'll also expose the following sub components on `Alert` which can be passed as children to `Alert`. The order of passing these as children doesn't matter:
+
+| Sub component           | Description                                                          | Required |
+| ----------------------- | -------------------------------------------------------------------- | -------- |
+| `Alert.Description`     | For setting body content                                             | ✅       |
```
nashcheez, line 73: What is the logic behind having `Description` as a compound component as compared to `Title` being a prop? There can be certain content structures where the title may get longer than the description.

nashcheez, line 73: Also, does `Alert.Description` need to be required? I can imagine many scenarios where certain Alerts could just have the `title`.

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Alert.SecondaryAction
+
+`Alert.SecondaryAction` renders and accepts a subset of `Link` (button) props:
+
+| Prop       | Type          | Default     | Description | Required |
+| ---------- | ------------- | ----------- | ----------- | -------- |
+| `children` | `string`, JSX | `undefined` | Content     | ✅       |
+| `href`     | `string`      | `undefined` | Hyperlink   |          |
+| `onClick`  | `function`    | `undefined` | Callback    |          |
```
nashcheez, line 100: A simple note here should be that `href` and `onClick` are mutually exclusive and cannot be used together.

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,162 @@
...
+  <Alert.Description>
+    Use vendor payouts to quickly generate invoices.{' '}
+    <Link href="https://razorpay.com">Know More</Link>
+  </Alert.Description>
+</Alert>
+```
+
+## Open questions
+
+**Q1.** Should we use a different variant naming convention in code like `success`, `warn`, `info`, `error`? Currently this is same as design
```
nashcheez, line 141: Valid argument, engineering generally should go with `success`, `warn`, `info`, `error` but in Blade, we have maintained prop consistency with design so that navigation and usability across Figma and code is seamless. Need to decide on this one as a team.

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,162 @@
...
+
+**A1.** TBD
+
+**Q2.** It's possible to pass any JSX as content but ideally we want to restrict it to few components (like `Link`, `string`, `List`). Doing these on code side would mean lot of checks. Similarly, restriction on `Alert.SecondaryAction` which should always be used with `Alert.PrimaryAction`. How do we do these - TS / runtime checks, etc.? One simpler alternative is to rely on design side restrictions for these but it makes the API more flexible should someone want to try things different from design.
```
kamleshchandnani, line 145: we can put the type of children as `children: Link | List | string`? and that might balance off flexibility+constraints?

For restricting the Primary and Secondary action thing - i think that might be the roadblock for compound component in this case. because the pattern itself proposes that things shall be composable. if we decide to move away from compound component then we can probably accept one prop similar to reshaped for actions and then add restrictions on that prop

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,162 @@
...
+
+**A2.** TBD
+
+## Pending
+
+- Document a11y related things above
+
```
kamleshchandnani, line 152: can we add `a11y` related things?

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,162 @@
...
+### Alert
+
+We'll expose an `Alert` component with the following API:
+
+| Prop          | Type                                     | Default     | Description                                                                                | Required |
+| ------------- | ---------------------------------------- | ----------- | ------------------------------------------------------------------------------------------ | -------- |
+| `children`    | `string`, Alert sub components           | `undefined` | Body content                                                                               | ✅       |
+| title         | `string`                                 | `undefined` | A brief heading                                                                            |          |
+| isDismissable | `boolean`                                | `true`      | Shows a dismiss button                                                                     |          |
```
kamleshchandnani, line 63: we probably might need an onClick handler for this to maybe capture some analytics?

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: The design guidelines specify title should be concise and scannable (few words). Detailed description should come below that. This was the rationale behind making description a child (since it can render any JSX) and title (which only renders a string) as a prop

divyanshu013: Yes, description will always be required as per design guidelines. So a title only alert wouldn't exist in our design language.

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: These can be used together (href is good for SEO reasons and onClick for handling actions). The secondary action will render a link button:
- [Decision docs for link button](https://github.com/razorpay/blade/blob/docs/link-decisions/packages/blade/src/components/Link/_decisions/decisions.md#link-component)
- [Related thread on Link component doc PR](https://github.com/razorpay/blade/pull/538#discussion_r912915149)

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: Right, keeping the terminology same as design is the driving factor here. Will update A1 for this Q1.

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,162 @@
...
+| contrast      | `high`, `low`                            | `low`       | Can be set to `high` for more prominent look _(not related to a11y)_                       |          |
+| isFullWidth   | `boolean`                                | `false`     | Spans the entire width of container, otherwise max width is restricted to 584px by default |          |
```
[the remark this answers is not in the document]
divyanshu013: It would be max-width because the alert shouldn't take more space than the screen size. However on larger screens the alert shouldn't take too much width.

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,162 @@
...
+</Alert>
+```
+
+### Full bleed
+
+> **Note**
+>
+> Design update pending to remove all borders and border radii
+
+This full bleed layout works for all desktop, mobile and native when the `isBorderless` prop is passed
```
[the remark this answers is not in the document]
divyanshu013: For setting width 100% we have the `isFullWidth` prop, so I think we should keep that decoupled from `isBorderless`.

We're not changing position CSS with `isFullWidth` 🤔 
I think absolute, relative shouldn't be needed because this will take entire width of container right? So users should customize the container's styles. Lemme know if I'm missing something.

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: Defining types on `children` is a good start, though I'm not sure if it'll be able to cover all possible combinations. For example - there can be a `Link` inside `List` content, so we would need sort of nested checks, not to forget that these are TS only checks so people using JS wouldn't get any constraints on these checks 😅 

Having an `actions` prop does help with restriction on preventing users from having a secondary action without primary action but it comes with its own learning curve:
- Actions should be an array of max 2 objects (first element goes as primary action and second element goes as secondary action)
- Secondary action can also have `href` because it's a link button, so the second object in actions should have a key to pass that (so users need to remember the order first is always a primary button and second is always a link button)
- Users might think they can pass another object to render 3 or more actions in the row 😅 

I put these as compound components so it feels more natural (*in declarative JSX way*) to compose and pass props to them like `onClick`, `children`, `href`.

But compound components do have their own learning curve so I don't have strong opinions to choose one over another; though personally favoring them slightly more. Open to other opinions, we can decide and go forward either way.

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,162 @@
...
+
+![Alert breakdown](./alert-breakdown.png)
+
+Sample usage:
+
+```jsx
+import { Alert, Link } from '@razorpay/blade';
+
+<Alert title="International Payments Only">
+  <Alert.Description>
```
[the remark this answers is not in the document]
divyanshu013: Made it compound to go well with primary and secondary action (related discussion in below thread). You're right consumers can pass anything as children here but that would be possible even if we've this as a prop (say `description` which accepts some JSX), wouldn't it 🤔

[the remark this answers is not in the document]
kamleshchandnani: so 2 things. one is the general API design and the second is JSX for action. Here, I'm referring to the former. the compound component pattern for Alert in general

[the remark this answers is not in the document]
divyanshu013: Added some thoughts in [this comment below](https://github.com/razorpay/blade/pull/573#discussion_r930692213), we can go with simple props as well like `description`, `actions` (or `primaryAction`, `secondaryAction`).

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -104,6 +104,17 @@ There'll also be an internal `BaseAlert` component that handles rendering web or
 
 ## a11y
 
+Web:
```
anuraghazra, line 107: Link to the APG pattern.
https://www.w3.org/WAI/ARIA/apg/patterns/alert/

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: Added

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: Yes, referenced this, could you please review the section and add thoughts

packages/blade/src/components/Alert/_decisions/decisions.md
anuraghazra: > not to forget that these are TS only checks so people using JS wouldn't get any constraints on these checks

In fact, you can't do TS checks on JSX childrens. Since children types get widened to JSX.Element so you won't be able to restrict it statically. This is a current design limitation of how JSX is integrated with TS. 

> Having an actions prop does help with restriction on preventing users from

Why accept array here? If we accept object most the the issues which you pointed will be solved. 

```tsx
<Alert actions={{ primary: {}, secondary: {} }} />
```
Or 

```tsx
<Alert primaryAction={} secondaryAction={} />
```

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,181 @@
...
+
+![Alert breakdown](./alert-breakdown.png)
+
+Sample usage:
+
+```jsx
+import { Alert, Link } from '@razorpay/blade';
+
+<Alert title="International Payments Only">
+  <Alert.Description>
```
anuraghazra, line 33: Also I'm not a great fan of the dot notation components. 

The problem with this approach is: 

1. It now becomes a backpack component. Say for example users never use Alert.SecondaryAction they will still carry the bundle size with them. They are not treeshakable. 
2. Discoverability issues are there with this approach, since now users have no idea that pressing "." dot after writing alert will have other components inside of it. 

I'll rather just do this: 

```tsx
<Alert>
  <AlertDescription>hello<AlertDescription>
  <AlertPrimaryAction>click</AlertPrimaryAction>
</Alert>
```


But I do still see why it can be useful, tho we first have to discuss wether we want this pattern in all components or not. If lets say we use this pattern only for Alert and not for other components end users will just get confused on which standard practices we follow.

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: Using array for actions would push people towards always specifying the primary action before using secondary action because primary action would be first element in the array.

divyanshu013: If we can't do TS checks to restrict passed content to link, list and string then runtime checks seem like the only option (which I feel might be an overkill right now 😅 )

anuraghazra: Arey yeah I got it. 
But with object we will be in control. 

Doesn't matter if users do `{ secondary: {}, primary: {} }` or `{ primary: {}, secondary: {} }`
We will render it our way inside the component.

divyanshu013: Ideally we want to restrict using secondary only when primary is also being used (not on its own)

packages/blade/src/components/Alert/_decisions/decisions.md
anuraghazra: Yes the a11y seems to be fine. 💯

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,204 @@
...
+import { Alert, Link } from '@razorpay/blade';
+
+<Alert
+  title="International Payments Only"
+  description={
+    <>
+      Currently you can only accept payments in international currencies using PayPal. You cannot
+      accept payments in INR (₹) using PayPal. <Link href="https://razorpay.com">Know More</Link>
+    </>
+  }
```
chaitanyadeorukhkar, line 39: Shouldn't this be a function?
```suggestion
  description={() =>
    <>
      Currently you can only accept payments in international currencies using PayPal. You cannot
      accept payments in INR (₹) using PayPal. <Link href="https://razorpay.com">Know More</Link>
    </>
  }
```

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,204 @@
...
+  actions={{
+    primary: { text: 'Primary Action', onClick: {() => { /* do something */ }} },
+    secondary: { text: 'Link', onClick: {() => { /* do something */ }}, href: 'https://razorpay.com'},
```
chaitanyadeorukhkar, line 42: Would we also need to accept target & rel for secondary?

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+### Alert
+
+We'll expose an `Alert` component with the following API:
+
+| Prop          | Type                                                        | Default     | Description                                                                                | Required |
+| ------------- | ----------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------------------ | -------- |
+| description   | `string`, `JSX`                                             | `undefined` | Body content                                                                               | ✅       |
+| title         | `string`                                                    | `undefined` | A brief heading                                                                            |          |
```
chaitanyadeorukhkar, line 54: Shouldn't title be required by design? Would look odd without one

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,204 @@
...
+
+Renders and accepts a subset of `Link` (button) props as keys:
+
+| Key       | Type       | Default     | Description | Required |
+| --------- | ---------- | ----------- | ----------- | -------- |
+| `text`    | `string`   | `undefined` | Content     | ✅       |
+| `href`    | `string`   | `undefined` | Hyperlink   |          |
+| `onClick` | `function` | `undefined` | Callback    |          |
+
+There'll also be an internal `BaseAlert` component that handles rendering web or native version similar to other Blade components. API would be similar to `Alert`.
```
chaitanyadeorukhkar, line 84: Why do we need an internal `BaseAlert`? For Link & Button we needed `Base*` since there were certain props we didn't want to expose to the consumer. I can't see that use-case here 🤔

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,204 @@
...
+
+## a11y
```
chaitanyadeorukhkar, line 86: Will we use LiveAnnouncer for this? 

// @anuraghazra

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,204 @@
...
+      Use vendor payouts to quickly generate invoices.{' '}
+      <Link href="https://razorpay.com">Know More</Link>
+    </>
+  }
+/>
+```
+
+## Open questions
+
+**Q1.** Should we use a different variant naming convention in code like `success`, `warn`, `info`, `error`? Currently this is same as design
```
chaitanyadeorukhkar, line 139: We should stick to design since that's the nomenclature we follow on tokens and rest of the action level components (Check intent prop for Button & Link)

packages/blade/src/components/Alert/_decisions/decisions.md
anuraghazra: Not really needed. Tho in android we might need to since last time I checked the aria role alert did not announce the changes. Otherwise role=alert should suffice.

packages/blade/src/components/Alert/_decisions/decisions.md
chaitanyadeorukhkar: But having said that, this is not intent. This is variant - we can choose to go ahead with something else here if it makes sense

packages/blade/src/components/Alert/_decisions/decisions.md
nashcheez: We have use cases for Alerts without a title - [ref](https://github.com/razorpay/blade/blob/div/alerts-decisions/packages/blade/src/components/Alert/_decisions/decisions.md#without-title)

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: From the React Native [docs](https://reactnative.dev/docs/accessibility#accessibilityrole) it should work ideally but yes if it doesn't we can definitely do this as an enhancement.

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: I think there isn't a need for that, JSX should be enough. Do you forsee a usecase for passing a function?

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: Same feelings, added the summary in the corresponding answer **A1**

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,211 @@
...
+/>;
+```
+
+### Alert
+
+We'll expose an `Alert` component with the following API:
+
+| Prop          | Type                                                        | Default     | Description                                                                                | Required |
+| ------------- | ----------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------------------ | -------- |
+| description   | `string`, `JSX`                                             | `undefined` | Body content                                                                               | ✅       |
```
blenderskool, line 53: Curious as to why description is passed as `description` prop and not `children` as I don't see it being used here for anything else.

packages/blade/src/components/Alert/_decisions/decisions.md
chaitanyadeorukhkar: Not sure. Trying to think if we should have a general guideline for such a thing where a prop would accept JSX or a function. For button and link we do this. I'm worried if consumers would be confused on when to use what

```jsx
<Button icon={CreditCardIcon} />
```

packages/blade/src/components/Alert/_decisions/decisions.md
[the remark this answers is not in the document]
sumit-gupta91: In general, what is the use case for Full Bleed ?

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: This was done to make it consistent with other props which affects what content gets rendered in the `Alert` - `title`, `actions`. These are passed inline.
Passing `description` as `children` was making it inconsistent (eg. why is title not passed as children).

packages/blade/src/components/Alert/_decisions/decisions.md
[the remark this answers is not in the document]
inaseem: I think we should leave the management of alert width on the consumers(developers). Ideally we should create all components assuming that they will occupy all the remaining horizontal space in a container.

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,211 @@
...
+| isDismissable | `boolean`                                                   | `true`      | Shows a dismiss button                                                                     |          |
+| onDismiss     | `function`                                                  | `undefined` | Callback when the dismiss button is pressed                                                |          |
+| contrast      | `high`, `low`                                               | `low`       | Can be set to `high` for more prominent look _(not related to a11y)_                       |          |
+| isFullWidth   | `boolean`                                                   | `false`     | Spans the entire width of container, otherwise max width is restricted to 584px by default |          |
+| variant       | `info`, `positive`, `notice`, `negative`                    | `info`      | Sets the color tone of entire alert. Icon is set automatically based on variant            |          |
+| isBorderless  | `boolean`                                                   | `false`     | Removes borders and border radii, useful for creating a full bleed layout                  |          |
```
kamleshchandnani, line 60: ideally since we have isBorderless for a very specific use case, we shall try to restrict people to use it even for non-full width/fleed layouts.

**proposal:** maybe, we shall remove this prop and only have `isFullWidth` which will also control the hiding of border and removing radii. that way we can say that the max-width of the alert will be set to 540px but if you want full width that will always become full bleed and usually full bleed is mostly for cases like banner only.

kamleshchandnani: **question:** will isBorderless make the alert absolute positioned, stick at the top, outside the entire dom tree? basically with higher z-index?

packages/blade/src/components/Alert/_decisions/decisions.md
kamleshchandnani: if the alert is something which is shown dynamically based on some validation fetched from API call, then shouldn't we announce them with live-announcer? that was the use case for live announcer right?

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: That would make it hard to build alerts for desktop where they're expected to span full width of container (which might be more than 584px on desktop). Example - the yellow alert on top below:

image

divyanshu013: `isBorderless` doesn't change positioning at all, because if we did we would also need to provide a way to set the position for users (top: 0, etc.)

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: `role="alert"` would do that automatically, wouldn't it? Also we'll make it `polite` for blue and green alerts.

packages/blade/src/components/Alert/_decisions/decisions.md
divyanshu013: I think this should be JSX, because a function isn't needed. If this were a `children` prop we would've expected to pass JSX, expanding same analogy here. Related [composition docs](https://reactjs.org/docs/composition-vs-inheritance.html#containment).

Regarding the point about `icon`. Generally if I pass JSX I would expect it to get rendered without modifications, passing something like `CreditCardIcon` it looks like that might get modified before rendering.

Though if you feel strongly about this approach, let's perhaps discuss and finalize on a call?

packages/blade/src/components/Alert/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: This can lead to inconsistencies which is one of the primary goals of resolving with blade :)

For example - we'll have to manually ensure across projects that all of them follow a consistent max-width, etc. for all these components

packages/blade/src/components/Alert/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: For building end to end alerts - generally the ones which need to stick to the top of container / app. Example:

image

packages/blade/src/components/Alert/_decisions/decisions.md
chaitanyadeorukhkar: > Generally if I pass JSX I would expect it to get rendered without modifications, passing something like CreditCardIcon it looks like that might get modified before rendering.

For Icon, it does need to get modified before rendering since the `size` & `color` of the icon depends on the state, intent & variant of button/link. Maybe that could be a decision factor for us - if it needs to be rendered as is then accept JSX otherwise pass it like we pass icon. 

I don't feel strongly about not using JSX here - just concerned that we're taking different approaches in which case we should be aware of what to use when. Lets take it case by case. JSX makes more sense here

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,211 @@
...
+| title         | `string`                                                    | `undefined` | A brief heading                                                                            |          |
+| isDismissable | `boolean`                                                   | `true`      | Shows a dismiss button                                                                     |          |
+| onDismiss     | `function`                                                  | `undefined` | Callback when the dismiss button is pressed                                                |          |
+| contrast      | `high`, `low`                                               | `low`       | Can be set to `high` for more prominent look _(not related to a11y)_                       |          |
+| isFullWidth   | `boolean`                                                   | `false`     | Spans the entire width of container, otherwise max width is restricted to 584px by default |          |
+| variant       | `info`, `positive`, `notice`, `negative`                    | `info`      | Sets the color tone of entire alert. Icon is set automatically based on variant            |          |
```
chaitanyadeorukhkar, line 59:
```suggestion
| variant       | `information`, `positive`, `notice`, `negative`                    | `info`      | Sets the color tone of entire alert. Icon is set automatically based on variant            |          |
```

Ref: https://github.com/razorpay/blade/pull/595#discussion_r936247945

chaitanyadeorukhkar (approved):
LGTM. Just need to finalize whether to call it `variant` or `intent` since this is been used across `BaseButton`, `BaseLink`, `Alert` & `Badge`. Possibly would be used in more components later.

packages/blade/src/components/Alert/_decisions/decisions.md
nashcheez: Would vote too to keep `isBorderless` independent of any kind of positioning. The decisions for positioning (absolute, sticky, fixed etc.) should be entirely left to the consumer if they want it wrapped inside a container or to the viewport. Even for the full bleed example I can see use cases where consumers might want the Alert popping up from the bottom of the screen, so a `position: fixed; top: 0` wouldn't make sense then.

nashcheez (approved):
Looks good overall, let's decide on the `variant/intent` piece and close the decisions by EOW.

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,215 @@
...
+<img src="./alerts-thumbnail.png" width="380" alt="" />
+
+## Design
+
+- [Figma - Alerts](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=6824%3A0)
+
+> **Note**
+>
+> The following needs work:
+>
```
saurav12, line 16: These all are done from the Design side. Can you please re-check it once and update it?

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,215 @@
...
+**A2.** We can do some static checks with TypeScript where feasible, for example enforcing `SecondaryAction` to be used only with `PrimaryAction` but some checks are outside scope of TypeScript and in the realm of runtime checks which at the moment seems like an overkill, for example restricting the content type for `description` prop.
+
+**Q3.** Should we make alerts focussable so they're discoverable by tabbing? It would mean also showing a visible keyboard focus indicator. Not doing this currently because I can't think of possible usecases right now, maybe we can revisit this later if / when needed.
+
+**A3.** TBD
+
+**Q4.** Whether to use `intent` or `variant`?
+
+**A4.** The rationale to use `intent` is because it signifies what the alert is meant for (success, error et al.). Also, we might need to extend alerts later to have different _variants_ like outline, ghost, etc. for which having a `variants` prop reserved for later would make more sense. There are some alternate terminology instead of `intent` that could be considered for example `colorScheme`.
+
```
saurav12, line 161: One more open question from the Design side is how we will render the links inside the `description` prop (in JSX)?

packages/blade/src/components/Alert/_decisions/decisions.md
saurav12: @kamleshchandnani @divyanshu013 Shall we remove links inside the description for now? Until we have a live use case?

divyanshu013: Thanks, updated the list. Only one remains now:

> the borderless intent mentioned in full bleed layout shouldn't have any borders and border radii

packages/blade/src/components/Alert/_decisions/decisions.md
saurav12: @kamleshchandnani @divyanshu013 Shall we remove links inside the description for now? Until we have a live use case?
// @nashcheez

divyanshu013: We can pass `Link` component as part of the JSX for that in code. Hope that's what you meant:

```jsx
<Alert
  description={
    <>
      Currently you can only accept payments in international currencies using PayPal. You cannot
      accept payments in INR (₹) using PayPal. <Link href="https://razorpay.com">Know More</Link>
    </>
  }
  ...
/>;
```

- [More details in the doc](https://github.com/razorpay/blade/blob/div/alerts-decisions/packages/blade/src/components/Alert/_decisions/decisions.md#api)

nashcheez: Good reminder @saurav12! 
@divyanshu013 links inside the description in the high contrast version are one thing we haven't solved yet design-wise, let's add it to the open questions and close this.

divyanshu013: Added in https://github.com/razorpay/blade/pull/573/commits/008e599e1ff67b63f5094b85c0d43e83e440f74b

packages/blade/src/components/Alert/_decisions/decisions.md
saurav12: @divyanshu013 This is also done. Please check.

> the borderless intent mentioned in full bleed layout shouldn't have any borders and border radii

divyanshu013: I think these ones are remaining:

image
 
image

saurav12: @divyanshu013 Thanks for pointing it out. Done!

packages/blade/src/components/Alert/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: Fullbleed (`isBorderless`) can be there without being full width (flat alerts). So kept both decoupled. Thoughts?

image

image

packages/blade/src/components/Alert/_decisions/decisions.md
```diff
@@ -0,0 +1,217 @@
...
+**A2.** We can do some static checks with TypeScript where feasible, for example enforcing `SecondaryAction` to be used only with `PrimaryAction` but some checks are outside scope of TypeScript and in the realm of runtime checks which at the moment seems like an overkill, for example restricting the content type for `description` prop.
+
+**Q3.** Should we make alerts focussable so they're discoverable by tabbing? It would mean also showing a visible keyboard focus indicator. Not doing this currently because I can't think of possible usecases right now, maybe we can revisit this later if / when needed.
+
+**A3.** TBD
+
+**Q4.** Whether to use `intent` or `variant`?
+
+**A4.** The rationale to use `intent` is because it signifies what the alert is meant for (success, error et al.). Also, we might need to extend alerts later to have different _variants_ like outline, ghost, etc. for which having a `variants` prop reserved for later would make more sense. There are some alternate terminology instead of `intent` that could be considered for example `colorScheme`.
```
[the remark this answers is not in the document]
divyanshu013: This was my first thought but since there aren't any other "variants" (usecases), didn't make it a `variant`. Example - there could be a ghost alert which is borderless (expanding upon current designs). Thoughts?

packages/blade/src/components/Alert/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: confirm with @saurav12 once. I'm okay whatever design takes the decision here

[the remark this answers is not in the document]
saurav12: @divyanshu013 I agree with @kamleshchandnani here. Full-bleed will be used with `fullWidth` only.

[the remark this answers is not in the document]
divyanshu013: Based on feedback, on design side whenever we specify `isBorderless`, it would automatically make the alert `isFullWidth`. Added an open question **Q6** for further info.

packages/blade/src/components/Alert/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: Added in open questions as **Q6** based on discussions. Please check.

nashcheez (approved):
I think this should be good to close now, don't see any other open questions.

PR #574, opened 2022-07-26
docs: added radio api decisions

anuraghazra:


packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,101 @@
+# Checkbox Decisions <!-- omit in toc -->
```
chaitanyadeorukhkar, line 1:
```suggestion
# Radio Decisions <!-- omit in toc -->
```

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,101 @@
...
+
```
chaitanyadeorukhkar, line 2: **Nit:** Add a cover image for Radio similar to how Divyanshu added for Alert?
image

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,101 @@
...
+- [`RadioGroup` API](#radiogroup-api)
+  - [Examples:](#examples)
+    - [Basic](#basic)
+    - [Controlled, Uncontrolled](#controlled-uncontrolled)
+- [Accessibility](#accessibility)
+- [Open Questions](#open-questions)
+
+## Design
+
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=13133%3A160709) to all variants of the Checkbox component
```
chaitanyadeorukhkar, line 14:
```suggestion
- [Figma Link - Radio](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=13133%3A160709)
```

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,101 @@
...
+
+## `Radio` API
```
chaitanyadeorukhkar, line 16:
```suggestion
## `Radio`
A Radio is used to present each item in a list of options where users must make a single selection
```

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,101 @@
...
+| helpText            | `string`                                | `undefined` | The helper text to be rendered                                                                                                                                 |          |
+| validationState     | `'none' \| 'error'`                      | `none`      | Control whether the radio is invalid or not.                                                                                                                   |          |
+| name (html native)  | `string`                                | `undefined` | The name of the input field in a radio, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name)                     |          |
+| value (html native) | `string`                                | `undefined` | The value of the input field in a radio, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#value)             |          |
+
+
+## `RadioGroup` API
```
chaitanyadeorukhkar, line 32:
```suggestion
## `RadioGroup`
A Radio Group is a list of grouped radio buttons. It is used when users have to choose a single option from a list of options.
```

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <RadioGroup
+      name="developers"
+      label="Developers (controlled)"
+      value={selected}
+      onChange={setSelected}
+    >
+      <Radio value="anurag">Anurag</Radio>
+      <Radio value="kamlesh">Kamlesh</Radio>
+      <Radio value="chaitanya">Chaitanya</Radio>
```
chaitanyadeorukhkar, line 81: 😡 @divyanshu013 !?!?
```suggestion
      <Radio value="anurag">Anurag</Radio>
      <Radio value="kamlesh">Kamlesh</Radio>
      <Radio value="chaitanya">Chaitanya</Radio>
      <Radio value="divyanshu">Divyanshu</Radio>
```

packages/blade/src/components/Radio/_decisions/decisions.md
chaitanyadeorukhkar, line 16: The whole API is similar to that of `Checkbox`, right? Only the minor differences of multi-select vs single-select?

anuraghazra: Yep! that's correct. (and keyboard handling which is by using arrow keys, tho handled by browser)

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Examples:
+
+#### Basic
+
+```tsx
+<RadioGroup
+  name="developers"
+  label="Developers"
+  helpText="Pick preferred blade developer"
```
divyanshu013, line 62: 😹

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,113 @@
...
+```tsx
+const Controlled = () => {
+  const [selected, setSelected] = React.useState('kamlesh');
+
+  return (
+    <RadioGroup
+      name="developers"
+      label="Developers (controlled)"
+      value={selected}
+      onChange={setSelected}
```
divyanshu013, line 83: This wouldn't work right? Checking the params received by `onChange` callback 🤔

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,113 @@
...
+| value (html native) | `string`                                | `undefined` | The value of the input field in a radio, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#value)             |          |
+
+## `RadioGroup`
+
+| Prop               | Type                                        | Default     | Description                                                                                                                      | Required |
+| ------------------ | ------------------------------------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| children           | `React.ReactNode`                           | `undefined` | Accepts multiple radios                                                                                                          | ✅       |
```
divyanshu013, line 41: Should also accept an array right? Because multiple `Radio` children?

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,113 @@
...
+| children            | `string`                                | `undefined` | The text to be rendered as RadioLabel                                                                                                                          | ✅       |
+| isChecked           | `boolean`                               | `false`     | state of the radio, if explicitly set the radio will become [controlled component](https://reactjs.org/docs/forms.html#controlled-components)                  |          |
+| defaultChecked      | `boolean`                               | `false`     | default state of the radio, if isChecked is not provided the radio will become [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html) |          |
+| onChange            | `({ event, isChecked, value }) => void` | `undefined` | The function to be called when the radio state changes.                                                                                                        |          |
```
divyanshu013, line 29: Do we need `event`? Also do we need both `isChecked` and `value` here 🤔

divyanshu013 (commented):
Looks good mostly. Some questions.

packages/blade/src/components/Radio/_decisions/decisions.md
nashcheez: +1, noticed this inconsistency in the Checkbox API where we are passing the event in Checkbox but not in the Checkbox group. Just thinking what possible use-cases will the consumer have with the event object in either cases (not sure if we have discussed this earlier). 

We will need both `isChecked` and `value`. `value` here refers to the value of the input field in the radio button, check [ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#value)

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,113 @@
...
+    - [Basic](#basic)
+    - [Controlled, Uncontrolled](#controlled-uncontrolled)
+- [Accessibility](#accessibility)
+- [Open Questions](#open-questions)
+
+## Design
+
+- [Figma Link - Radio](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=13133%3A160709)
+
+## `Radio`
```
nashcheez, line 22: This is also a question for design but will we ever have a single radio button? Radio buttons always exist in groups and only one of them should be selectable in which case shouldn't they be all part of a `RadioGroup`?

We had an unusual case for Checkbox where a single checkbox could exist (eg. the terms and conditions check), but I can't think of where we will have a use case for a single radio button?

packages/blade/src/components/Radio/_decisions/decisions.md
anuraghazra: Will fix.

packages/blade/src/components/Radio/_decisions/decisions.md
anuraghazra: React.ReactNode type defs cover that.

packages/blade/src/components/Radio/_decisions/decisions.md
anuraghazra: Correct we won't have single Radio outside of RadioGroup. Since a radio will only work if it's grouped together with multiple radios by binding the `name` html prop.

packages/blade/src/components/Radio/_decisions/decisions.md
anuraghazra: > noticed this inconsistency in the Checkbox API where we are passing the event in Checkbox but not in the Checkbox group

@nashcheez The reason is in a single Checkbox it has it's own `<input />` which can have `event` but CheckboxGroup and RadioGroups are abstractions which we built to provide state management and other features, thus it can't have an `event` object. 

> We will need both isChecked and value

Yes initially on the first draft I only kept `isChecked` but kamlesh suggested that for analytics and other services they could want the `value` too. [ref](https://github.com/razorpay/blade/pull/539#discussion_r925639702)

packages/blade/src/components/Radio/_decisions/decisions.md
```diff
@@ -0,0 +1,109 @@
...
+
+> Note: Radio component can't be used outside of RadioGroup.
+
+| Prop                | Type                                    | Default     | Description                                                                                                                                                    | Required |
+| ------------------- | --------------------------------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| children            | `string`                                | `undefined` | The text to be rendered as RadioLabel                                                                                                                          | ✅       |
```
divyanshu013, line 28: What is RadioLabel? Seems like a component which isn't mentioned here

divyanshu013 (approved):
Looks good to me. One question.

packages/blade/src/components/Radio/_decisions/decisions.md
anuraghazra: Internal component accidentally mentioned here. Will remove.

sumitd94, line 28: just curious to know, why are we going with `children` here and not accepting a prop for a label?
so basically Radio can expose a prop - `labelText` and we can pass what label we want to display.

anuraghazra: No particular reason, just how we designed the API with the approach. Intuitively It looks and feels more declarative. 
Most third party libraries also use similar API (chakra, spectrum, reshaped etc). 

image

packages/blade/src/components/Radio/_decisions/decisions.md
jaywalker21: @anuraghazra Noticed that for a single checkbox, we are not sending out `name` in the `onChange` handler (maybe, can be extracted from `event`, as `event.target.name`), but for `RadioGroup`, we are having `name` in the `onChange`. I know this PR is about `Radio`, but given `Radio` cannot exist w/o `RadioGroup` and we are sending out the name in `RadioGroup`, thought we should be doing the same for a single checkbox from analytics perspective. It will even make the analytics handler similar as it will always get the `{ name, value }` in the change handler.

Given how `Input` component's [decisions](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Input/_decisions/decisions.md) are taken, should we send out the `name` in single checkbox too for uniformity purpose in the `onChange` prop?

anuraghazra: Yeah sure we'll consider this and make the API consistent. For the time being we're freezing this API.

PR #587, opened 2022-08-01
refactor(blade): storybook examples

anuraghazra:
Changes in this PR

- Sorted storybook to keep components first and then story recipes 
- Changed SkipNav to use BaseLink component
- Improved SkipNav stories to use newer components like Button, Box, Link etc
- Improved VisuallyHidden storybook with Checkbox
- Refactored Accessibility demo recipe 
- **Major styling refactor of CardsWithToken demo**
  - Fixed React-native card (now it renders)
  - Improved/refactored duplicate styles


divyanshu013 (commented):
Why are the imports not being auto imported from `~components` instead of importing them individually from their paths like `~components/Box`, `~components/Icon`, etc.

anuraghazra:
> Why are the imports not being auto imported from `~components` instead of importing them individually from their paths like `~components/Box`, `~components/Icon`, etc.

We did not configured it that way. 

The path alias is 

```json
"~components/*": ["src/components/*"],
```

It will work only if we add this

```json
"~components": ["src/components"],
"~components/*": ["src/components/*"],
```

packages/blade/src/storybook-recipes/SimpleCardWithTokens/Card.native.tsx
```diff
@@ -1,96 +1,17 @@
+/* eslint-disable babel/new-cap */
```
divyanshu013, line 1: Why is this needed 🤔

packages/blade/src/storybook-recipes/SimpleCardWithTokens/styles.ts
```diff
@@ -0,0 +1,104 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
```
divyanshu013, line 2: Probably we can make this rule required only for exported modules

divyanshu013 (approved):
Looks good to me.

packages/blade/src/storybook-recipes/SimpleCardWithTokens/styles.ts
anuraghazra: Not sure if there's any rule like that exists.

packages/blade/src/storybook-recipes/SimpleCardWithTokens/Card.native.tsx
anuraghazra: Because I named the style functions with uppercase. 

> A function with a name starting with an uppercase letter should only be used as a constructor.eslint[babel/new-cap](https://eslint.org/docs/rules/new-cap)

packages/blade/src/storybook-recipes/SimpleCardWithTokens/styles.ts
divyanshu013: Non blocker: In short making it required only for [module boundary types](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md) and disabling it for other functions which aren't exported. Needs discussion within team though.

chaitanyadeorukhkar (changes requested):

packages/blade/src/components/SkipNav/SkipNav.web.tsx
```diff
@@ -1,24 +1,21 @@
...
 
 const fallbackId = 'blade-skip-nav';
 type SkipNavLinkProps = {
   id?: string;
   children?: string;
 };
 
-// TODO: Replace with Link component
-const StyledLink = styled.a(({ theme }) => ({
+const StyledLink = styled(BaseLink)(({ theme }) => ({
```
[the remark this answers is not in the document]
anuraghazra: It's not for spacing only. 

I want all the features of Link. focus, hover, colors, isVisited state, interactions.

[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah but you're extending Link's style just to add spacing. Use box instead? Don't want to spread props on link or allow it's style to be overriden

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -248,6 +249,7 @@ const BaseLink = ({
       motionDuration={motionDuration}
       motionEasing={motionEasing}
       setCurrentInteraction={setCurrentInteraction}
+      {...props}
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: @anuraghazra ^ We shouldn't do this. It will allow others to override styles of Link and create design inconsistencies. It could also lead to passing of dangerous attributes like `as`

packages/blade/src/components/SkipNav/SkipNav.web.tsx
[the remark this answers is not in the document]
anuraghazra: The important part here is that I want composition. Open for suggestion on how we can utilise the existing Link to extend the functionality. 

I'm extending BaseLink's styling to add `screenReaderStyles` & `:focus` to get the SkipNav behaviour. 
One way would be to instead using the `StyledBaseLink`.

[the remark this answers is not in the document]
anuraghazra: @chaitanyadeorukhkar can you check again. 

I think this is a good tradeoff to be had.

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/SkipNav/SkipNav.stories.tsx
```diff
@@ -55,29 +58,36 @@ const SkipNavTemplate: ComponentStory<typeof SkipNavLink> = () => {
...
+          <Button size="small">Button 2</Button>
+        </Box>
         <SkipNavContent id="second" />
+        <Box marginTop="spacing.1" />
         <Text>Second Main content of the page</Text>
-        <button>Button 3</button>
-        <button>Button 4</button>
-      </main>
+        <Box marginTop="spacing.1" />
+        <Box gap="spacing.1" display="flex">
```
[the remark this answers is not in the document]
anuraghazra: I'll fix it separately on the Checkbox patch PR

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -187,6 +187,10 @@ const BaseLink = ({
   intent,
   contrast = 'low',
   accessibilityLabel,
+  // @ts-expect-error avoiding exposing to public
+  className,
+  // @ts-expect-error avoiding exposing to public
+  style,
```
chaitanyadeorukhkar, line 193: This will only stop TS projects. Lets not allow overriding styles. The use-case here is not strong enough to allow style overriding. We just want to add spacing in stories and that's not a good reason to expose this 😓

chaitanyadeorukhkar (changes requested):

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
anuraghazra: Yep totally got your point. 

But I'm not just adding padding. 

```ts
const StyledLink = styled(BaseLink)(({ theme }) => ({
  ...screenReaderStyles,      // screen reader styles
  top: theme.spacing[4],     // padding
  left: theme.spacing[4],     // padding
  ':focus': {    // on focus -> reset screen reader styles
    padding: theme.spacing[1],
    clip: 'auto',
    clipPath: 'unset',
    width: 'auto',
    height: 'auto',
  },
}));
```

chaitanyadeorukhkar: Can't we do all of these styling on a wrapper on top of Link?

```jsx
const SkipNavContainer = styled.div({
...screenReaderStyles,
})

<SkipNavContainer>
  <Link />
</SkipNavContainer>
```

anuraghazra: Unfortunately we can't since:

- now the focus will be on the div so we also need to add :focus on the div. 
- then add tabindex={0} to make it focusable
- focus styling will be different 
- Pressing enter will not trigger the link navigation but rather the div will be clicked. 

:(

chaitanyadeorukhkar: Okay, let's move ahead with this then. I'll take a final review and approve the PR

chaitanyadeorukhkar (approved):

PR #593, opened 2022-08-02
fix(Typography): expose text styles

kamleshchandnani:
This PR exposes Text styles as object so we can use them in places where we can't use text as JSX. Eg: styling for `placeholder` text for Input Field

With this, I can now style the text with just CSS without deviating from the system and get the same text as we get via JSX
```js
const StyledInput = styled.input((props) => ({
  ...getTextStyles({
    size: 'medium',
    variant: 'body',
    type: 'subtle',
    weight: 'regular',
    contrast: 'low',
    theme: props.theme,
  }),
  '::placeholder': getTextStyles({
    size: 'medium',
    variant: 'body',
    type: 'placeholder',
    weight: 'regular',
    contrast: 'low',
    theme: props.theme,
  }),
  backgroundColor: props.theme.colors.brand.gray[200],
  width: '100%',
}));
```


divyanshu013:
The snapshots are changing from actual css values to tokens. Is that intentional? I thought we wanted css values in snapshots to catch discrepancies 🤔

packages/blade/src/components/Typography/BaseText/BaseText.web.tsx
```diff
@@ -0,0 +1,69 @@
...
+      fontSize={fontSize}
+      fontWeight={fontWeight}
+      fontStyle={fontStyle}
+      textDecorationLine={textDecorationLine}
+      lineHeight={lineHeight}
+      as={as}
+      textAlign={textAlign}
+      numberOfLines={truncateAfterLines}
+      className={className}
+      style={style}
```
chaitanyadeorukhkar, line 62: Non blocker: We might want to consider removing `style` prop here. Earlier we (I) introduced this because there was no mechanism of adding padding between icon & text of Button. With the Box component, this need is resolved and its usage is also removed with [this PR](https://github.com/razorpay/blade/pull/543). 

Exposing style directly is a bit risky and we should be exposing this in only extreme situations

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -83,7 +83,7 @@ const getProps = <T extends { variant: HeadingVariant }>({
   return props;
 };
 
-const Heading = <T extends { variant: HeadingVariant }>({
+export const Heading = <T extends { variant: HeadingVariant }>({
```
chaitanyadeorukhkar, line 86: So we're moving to named exports everywhere now?

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -1,8 +1,8 @@
 /* eslint-disable @typescript-eslint/no-unnecessary-type-assertion */
 import type { ReactElement } from 'react';
 import styled from 'styled-components';
-import BaseText from '../BaseText';
-import type { BaseTextProps } from '../BaseText/BaseText';
+import { BaseText } from '../BaseText';
```
chaitanyadeorukhkar, line 4: **Non-Blocker:** Named export look weird tbh 😢

packages/blade/src/components/Typography/index.ts
```diff
@@ -1,6 +1,6 @@
-export { default as Title } from './Title';
+export { Title } from './Title';
```
chaitanyadeorukhkar, line 1: Why do we need 
```jsx
export { Title } from './Title';
```
if we're also doing
export * from './Title';

Same for Heading & Text

packages/blade/src/components/Typography/Heading/Heading.tsx
kamleshchandnani: the problem is right now we end up in a lot of mess sometimes because of named export and the weird `.web.tsx` and `.native.tsx` and then to resolve them there's no option without having named exports, i guess both me and Anurag have struggled because of this. so trying if we can use named everywhere then. because in the end for consumers everything is named import only. it's just for internals, open for the conclusion we come to mutually

packages/blade/src/components/Typography/index.ts
kamleshchandnani: wait, i fixed this already i think on my local, could be diff branch. let me fix this

kamleshchandnani:
> The snapshots are changing from actual css values to tokens. Is that intentional? I thought we wanted css values in snapshots to catch discrepancies 🤔

the problem here is the snapshots are at all levels so the abstractions will now yield to tokens now after my refactor. previously we were evaluating the tokens and sending the raw values to the StyledBaseText but now because of the requirement i have in input we need to send tokens else there will be token evaluations at all levels hence this snapshot change. I can do one thing i can write a test for getBaseTextStyles functions and snapshot the return type so at least at the final bottom abstraction level we'll see CSS things

packages/blade/src/components/Typography/BaseText/BaseTextTypes.ts
```diff
@@ -0,0 +1,48 @@
...
+>}`;
+type FeedbackActionColors = `feedback.${Feedback}.action.text.${DotNotationColorStringToken<
+  Theme['colors']['feedback'][Feedback]['action']['text']
+>}`;
+type SurfaceColors = `surface.text.${DotNotationColorStringToken<
+  Theme['colors']['surface']['text']
+>}`;
+type ActionColors = `action.text.${DotNotationColorStringToken<Theme['colors']['action']['text']>}`;
+
+export type BaseTextProps = {
```
divyanshu013, line 17: Can probably add JSDoc comments on non react native props so they will show up in editor

packages/blade/src/components/Typography/Heading/Heading.tsx
divyanshu013: We can probably enforce this inside `components/*` with a linting rule

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Typography/BaseText/BaseText.web.tsx
kamleshchandnani: Probably we can do it in another PR?

packages/blade/src/components/Typography/Heading/Heading.tsx
chaitanyadeorukhkar: Enforce named exports?

chaitanyadeorukhkar (approved):

PR #594, opened 2022-08-02
fix(blade): checkbox errorText alignment

anuraghazra:
Changes in this PR

- Updated errorText placement.
- Added new FormGroup component which will handle error, success, help text rendering.


packages/blade/src/components/Checkbox/Checkbox.tsx
```diff
@@ -177,33 +177,34 @@ const Checkbox = ({
...
+        <CheckboxIcon
+          isChecked={state.isChecked}
+          isIndeterminate={isIndeterminate}
+          isDisabled={_isDisabled}
+          isNegative={_hasError}
+        />
+        <Box>
+          <CheckboxLabelText>{children}</CheckboxLabelText>
+          {showSupportingText && (
+            <SupportingText id={ids?.helpTextId} variant="help">
```
divyanshu013, line 201: What should be the scope of formgroup (example - supporting text is not in form group)? Also, aren't we planning to expose form group to consumers?

anuraghazra: SupportText is a very special usecase for Checkbox. 

I've extracted out helpText, errorText, successText logic into FormGroup so that it can be used in Input too. 

> Also, aren't we planning to expose form group to consumers?
Not right now, We have that as part of Milestone 3 in September. That time we will look to make the API more refined for public usage.

divyanshu013 (approved):
Looks good to me. Kamlesh is working on some form components with Input, you might want to check those too to avoid conflicts.

packages/blade/src/components/Form/FormHintText.tsx
```diff
@@ -42,23 +75,72 @@ const FormHintText = ({
...
+              as={isReactNative ? undefined : 'span'}
+              color={colors[state]}
+              fontSize={50}
+              lineHeight="s"
+              fontStyle="italic"
+              fontFamily="text"
+            >
+              {helpText}
+            </BaseText>
+          </FormHintTextWrapper>
```
kamleshchandnani, line 101: this you can abstract it to a function above in this file only? since it's used at three places back to back here

packages/blade/src/components/Form/FormHintText.tsx
```diff
@@ -9,15 +9,48 @@ import { getPlatformType } from '~utils';
...
 
 const FormHintText = ({
-  id,
-  children,
-  variant = 'help',
+  state,
+  errorText,
+  successText,
+  helpText,
+  helpTextId,
```
kamleshchandnani, line 51: why do we need ids? for a11y?

packages/blade/src/components/Form/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './FormHintText';
```
kamleshchandnani, line 1: explicitly export what you need?

kamleshchandnani:
resolve conflicts

packages/blade/src/components/Form/FormHintText.tsx
anuraghazra: Yes

kamleshchandnani: what exactly does the ids do? we don't write logic for them right?

packages/blade/src/components/Form/index.ts
```diff
@@ -0,0 +1,2 @@
+export { FormHintText } from './FormHintText';
```
kamleshchandnani, line 1: do you think it would make sense to have `FormHint` and `FormLabel` components?

PR #595, opened 2022-08-02
docs: add `Badge` API decisions

chaitanyadeorukhkar:
### [View Formatted Decisions](https://github.com/razorpay/blade/blob/docs/badge-decisions/packages/blade/src/components/Badges/_decisions/decisions.md)


packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,61 @@
...
+- Should we call it `labelWeight`, `labelStyle` or `fontWeight`, `fontStyle` or `textWeight`, `textStyle`?
+- Should we call it `variant` or `intent`?
+- What should be the default `variant`? Kept it as `neutral` for now since it seems more _neutral_ and appropriate
+- Scope of A11y?
+  - NA
+
+#### Design
+- Do we need badges where the text can overflow? Ref: [Mantine Badge](https://mantine.dev/core/badge/#full-width-and-overflow)
+- Do we need to support icon only badges?
+- Should we auto-capitalize label when `labelWeight` is bold?
```
chaitanyadeorukhkar, line 50: @saurav12 would need your help on these. I've also added these comments on Figma

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,61 @@
...
+</Badge>
+```
+
+## Open Questions
+#### Dev
+- Should the label be automatically capitalized when `labelWeight` is `bold`
+- Should we call it `labelWeight`, `labelStyle` or `fontWeight`, `fontStyle` or `textWeight`, `textStyle`?
+- Should we call it `variant` or `intent`?
+- What should be the default `variant`? Kept it as `neutral` for now since it seems more _neutral_ and appropriate
+- Scope of A11y?
```
chaitanyadeorukhkar, line 44: @anuraghazra I don't think we explicitly need to do anything here. Do you think we might need to do something extra?

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,61 @@
...
+  labelWeight='bold'
+>
+  New
+</Badge>
+```
+
+## Open Questions
+#### Dev
+- Should the label be automatically capitalized when `labelWeight` is `bold`
+- Should we call it `labelWeight`, `labelStyle` or `fontWeight`, `fontStyle` or `textWeight`, `textStyle`?
```
saurabhdaware, line 41: I like `fontStyle` and `fontWeight` as it is similar to css properties.

What is the standard in other components?

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,61 @@
...
+  labelStyle='italic' 
+  labelWeight='bold'
+>
+  New
+</Badge>
+```
+
+## Open Questions
+#### Dev
+- Should the label be automatically capitalized when `labelWeight` is `bold`
```
saurabhdaware, line 40: This is design question no?

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,61 @@
...
+- Should we call it `labelWeight`, `labelStyle` or `fontWeight`, `fontStyle` or `textWeight`, `textStyle`?
+- Should we call it `variant` or `intent`?
```
saurabhdaware, line 42: +1 to `variant` since that is the one used in `Alert` component. Not sure which one is used in other components. Do we use `intent` anywhere else?

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Design
+- [Figma - Badge](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=8110%3A417)
+
+
+## API
+
+| Prop | Type | Required | Default | Description |
+|---|---|---|---|---|
+| variant | `positive`,`negative`, `notice`, `information`, `neutral` | No | `neutral` | The variant of the badge to be rendered. |
```
saurabhdaware, line 14:
```suggestion
| variant | `positive`,`negative`, `notice`, `info`, `neutral` | No | `neutral` | The variant of the badge to be rendered. |
```

can we make this `info`? Alert has `info`

chaitanyadeorukhkar: Our tokens have `information`. BaseButton & Link `intent` has `information`. We should keep it similar everywhere. @divyanshu013 should we make it `information` for Alert?

packages/blade/src/components/Badges/_decisions/decisions.md
chaitanyadeorukhkar: Yup. Added in both places by mistake 🙈

chaitanyadeorukhkar: Yup. Added in both places by mistake 🙈

packages/blade/src/components/Badges/_decisions/decisions.md
chaitanyadeorukhkar: We're using a prop called `intent` that can be `positive`, `negative`, `neutral`, `information`, `notice` for `BaseButton` & `BaseLink`

- [BaseButton API](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Button/_decisions/decisions.md#basebutton-api)
- [BaseLink API](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Link/_decisions/decisions.md#baselink-api)

chaitanyadeorukhkar: cc: @divyanshu013 another possible decision that we need to take for both `Badge` & `Alert`. Calling it `variant` or `intent`?

packages/blade/src/components/Badges/_decisions/decisions.md
chaitanyadeorukhkar: On further digging. Should we make the aria-role as `status`. It acts as a `live-region`

Ref: 
- https://www.w3.org/TR/wai-aria-1.1/#status
- https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22#example-2-updating-the-shopping-cart-status
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role

packages/blade/src/components/Badges/_decisions/decisions.md
divyanshu013: I think we can call it `variant` which is a more conventional prop in design libraries. `intent` could be kept for internal components and props which we aren't exposing to users.

Regardless, we should keep it consistent b/w all exported components.

packages/blade/src/components/Badges/_decisions/decisions.md
chaitanyadeorukhkar: @saurabhdaware Divyanshu changed it to `information` on Alert. Lets stick to `information` since the convention would be similar across components and design

packages/blade/src/components/Badges/_decisions/decisions.md
chaitanyadeorukhkar: There would be a possibility that we might expose `intent` as a prop on Button to our consumers when use-cases pop up.  This means at that point, we will have `variant` & `intent` on Button where `intent` will be similar to `variant` of Alert & Badge. 

I agree variant seems more conventional for Badge & Alert but if we're treating `positive`, `negative`, `information`, etc. as _intents_ (intentions) then why not call it `intent` on Badge & Alert as well?

divyanshu013: We have a `variant` on button for primary, etc. So variant on some components and intent on some could also be a point of confusion. On some libraries they have separate prop for controlling the color part (which we're doing with intent).

divyanshu013: Let's add the points we discussed and the final call for historical purposes :)

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,60 @@
...
+- Should we call it `labelWeight`, `labelStyle` or `fontWeight`, `fontStyle` or `textWeight`, `textStyle`?
+- Should we call it `variant` or `intent`?
+- What should be the default `variant`? Kept it as `neutral` for now since it seems more _neutral_ and appropriate
+- Scope of A11y?
+  - NA
+
+#### Design
+- Do we need badges where the text can overflow? Ref: [Mantine Badge](https://mantine.dev/core/badge/#full-width-and-overflow)
+- Do we need to support icon only badges?
+- Should we auto-capitalize label when `labelWeight` is bold?
```
divyanshu013, line 49: I think, not on dev side, keeping control for users.

packages/blade/src/components/Badges/_decisions/decisions.md
divyanshu013: I think if we're considering icon only badges then a label would definitely be needed

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,60 @@
...
+```jsx
+import { Badge } from '@razorpay/components';
+
+<Badge 
+  variant='positive'
+  contrast='high'
+  icon={InfoIcon} 
+  size='large' 
+  labelStyle='italic' 
+  labelWeight='bold'
```
divyanshu013, line 32: Isn't the code getting autoformatted to use double quotes?

divyanshu013 (commented):
Looks good overall, we can summarize the points from variants discussion and include in doc.

packages/blade/src/components/Badges/_decisions/decisions.md
anuraghazra: Yes icon only badges need that. 

Tho @chaitanyadeorukhkar we don't need role=status in a badge. Mostly badges are static elements. And that Add to cart example is very specific which can be achieved with our announce() function. 

Unfortunately there's not pattern for badges thus better not to invent one.

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+## API
+
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| children | `string` | `undefined` | The label to be rendered within the badge. | ✅ |
+| variant | `positive`,`negative`, `notice`, `information`, `neutral` | `neutral` | The variant of the badge to be rendered. |  |
+| contrast | `low`, `high` | `low` | The contrast of the badge to be rendered. |  |
+| size | `medium`, `small` | `medium` | The size of the badge to be rendered. |  |
```
nashcheez, line 17: What is the difference b/w the medium and small sizes? Shouldn't the badge wrap its content and have a `max-width`?

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,77 @@
...
+- But having `intent` accept semantic only(`positive`, `negative`, `notice`, `information`, & `neutral`) values on BaseButton, BaseLink & Alert and accepting semantic + non-semantic(`purple`, `indigo` or whatever new requirement comes in) values on Badge did not make sense. For such free form values, `variant` prop would make more sense.
+- Now we concluded to use the `variant` prop that accepts semantic + non-semantic values. We would follow this rule: If the prop accepts semantic only values then we would `intent` but if the prop accepts semantic as well as non-semantic values, we would use `variant`
+- The tradeoff right now is that until we get more requirements for non-semantic values, we would have a `variant` prop on Badge that accepts `positive`, `negative`, `notice`, `information`, & `neutral` but we are future-proofing the component by leaving room to accept more non-semantic values.
```
nashcheez, line 66: https://c.tenor.com/ZRZPJ-LJ-D0AAAAd/saturnoir.gif

packages/blade/src/components/Badges/_decisions/decisions.md
chaitanyadeorukhkar: We won't be having icon only badge so this won't be a concern

packages/blade/src/components/Badges/_decisions/decisions.md
chaitanyadeorukhkar: Added

packages/blade/src/components/Badges/_decisions/decisions.md
chaitanyadeorukhkar: It's not 😅

packages/blade/src/components/Badges/_decisions/decisions.md
chaitanyadeorukhkar: Different heights. Small is of 20px and Medium is 24px

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,77 @@
...
+
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| children | `string` | `undefined` | The label to be rendered within the badge. | ✅ |
+| variant | `positive`,`negative`, `notice`, `information`, `neutral` | `neutral` | The variant of the badge to be rendered. |  |
+| contrast | `low`, `high` | `low` | The contrast of the badge to be rendered. |  |
+| size | `medium`, `small` | `medium` | The size of the badge to be rendered. |  |
+| icon | `Icon` | `undefined` | The Blade `Icon` component to be rendered for the badge. |  |
+| fontStyle | `italic`, `normal` | `normal` | The font style of the badge's font. |  |
+| fontWeight | `bold`, `regular` | `regular` | The font weight of the badge's font. |  |
```
kamleshchandnani, line 20: why do we need to expose this? 🤔

chaitanyadeorukhkar: Part of design, we need to let the consumers make it bold or italic

packages/blade/src/components/Badges/_decisions/decisions.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Sample Usage
+```jsx
+import { Badge } from '@razorpay/components';
+
+<Badge 
+  variant='positive'
+  contrast='high'
+  icon={InfoIcon}
```
sidag95, line 29: **nit(non-blocking):** How are we importing InfoIcon here? Can we add the import for that here since we only allow Blade icons here?‏

sidag95: **question:** ‏How and why are we restricting this component to use only Blade icons? Given that the variant is a free-form prop, shall the icon also be a free-form prop here?

packages/blade/src/components/Badges/_decisions/decisions.md
saurav12: @chaitanyadeorukhkar No, it's not like that. Capitalisation can take place regardless of the font-weight. I've also changed the type-token for labels. Now we have only one token for all labels `BodySmallRegular` (Regular/Bold).

_PS: I've commented the same on Figma._

packages/blade/src/components/Badges/_decisions/decisions.md
chaitanyadeorukhkar: Added import for icon. We want consumers to use Blade Icons. Generally, we support all feather icons, but if there are any missing and the requirement comes in, we can add the icon directly to blade. 

Icons also have to adhere to a design language - for instance, we only support monochrome icons right now. Making it free-form could lead to unintended usage

sidag95: Understood and agreed. So how are we enforcing the Icons to only be BladeIcons here? Seems to be guidelines right now 🤔

chaitanyadeorukhkar: Yup. Guidelines and usage examples. We thought about adding it as `iconName` and letting the consumer pass the icon name instead of the component but that would affect tree-shaking negatively. Adding guidelines seem to be the best approach. We still need to figure out a better way to handle icons across components but this seemed the best tradeoff for now

PR #604, opened 2022-08-03
fix: button styling for native

chaitanyadeorukhkar:


packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -144,7 +144,7 @@ const getProps = ({
     fontSize: buttonTypographyTokens.fonts.size[size],
     lineHeight: buttonTypographyTokens.lineHeights[size],
     minHeight: makeSize(buttonMinHeight[size]),
-    iconPadding: hasIcon ? `spacing.${textPadding[size]}` : undefined,
+    iconPadding: hasIcon && children?.trim() ? `spacing.${textPadding[size]}` : undefined,
```
chaitanyadeorukhkar, line 147: We should be adding padding only when there is icon _and_ text. Without this, Icon Only buttons had an extra padding

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -354,7 +354,7 @@ const BaseButton = ({
         flexDirection="row"
         alignItems="center"
         justifyContent="center"
-        alignSelf="center"
+        flex={1}
```
chaitanyadeorukhkar, line 357: Need to specify the flex for the content other native would not take up the whole real-estate available within button and restrict content to the top

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -8,7 +8,10 @@ import { useTheme } from '~components/BladeProvider';
 
 const StyledPressable = styled(Animated.createAnimatedComponent(Pressable))<
   Omit<StyledBaseButtonProps, 'onClick' | 'accessibilityProps'>
->((props) => getStyledBaseButtonStyles(props));
+>((props) => ({
+  ...getStyledBaseButtonStyles(props),
+  alignSelf: 'center',
```
chaitanyadeorukhkar, line 13: Need to let the native button align itself to center otherwise it takes up all the available space and spreads itself

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
anuraghazra: Why is this happening suddenly? Previously it was fine.

chaitanyadeorukhkar: The PR was open for too long, I must've missed checking this on native 🤦‍♂️

anuraghazra (approved):

PR #605, opened 2022-08-03
feat(blade): Single radio - PR1

anuraghazra:
Not ready for review. 

Added single radio component.

### DIFF
--- a/packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
+++ b/packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
@@ -38,10 +38,12 @@ const variants = {
 
 export type CheckboxRectProps = Omit<CheckboxIconProps, 'state'> & {
   isChecked: boolean;
+  isFullRounded?: boolean;
 };
 
 const getCheckboxIconWrapperStyles = ({
   theme,
+  isFullRounded,
   isChecked,
   isDisabled,
   isNegative,
@@ -63,7 +65,7 @@ const getCheckboxIconWrapperStyles = ({
     borderStyle: 'solid',
     margin: '0px',
     marginTop: '3px',
-    borderRadius: makeSize(theme.border.radius.small),
+    borderRadius: makeSize(isFullRounded ? theme.border.radius.max : theme.border.radius.small),
     marginRight: makeSpace(theme.spacing[2]),
     backgroundColor: getIn(theme, backgroundColor),
     borderColor: getIn(theme, borderColor),
--- a/packages/blade/src/components/Icons/_Svg/Circle/Circle.d.ts
+++ b/packages/blade/src/components/Icons/_Svg/Circle/Circle.d.ts
@@ -1,0 +1,11 @@
+export type CircleProps = {
+  fill?: string;
+  stroke?: string;
+  r?: string;
+  cx?: string;
+  cy?: string;
+  x?: string;
+  y?: string;
+};
+
+export { default } from './Circle.web';
--- a/packages/blade/src/components/Radio/Radio.stories.tsx
+++ b/packages/blade/src/components/Radio/Radio.stories.tsx
@@ -1,0 +1,81 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+import React from 'react';
+import type { ComponentStory, Meta } from '@storybook/react';
+import { Button, Text } from '..';
+import type { RadioProps } from './Radio';
+import { Radio as RadioComponent } from './Radio';
+import Box from '~components/Box';
+
+export default {
+  title: 'Components/Radio/Radio',
+  component: RadioComponent,
+  args: {
+    defaultChecked: undefined,
+    validationState: undefined,
+    isChecked: undefined,
+    isDisabled: undefined,
+    isIndeterminate: undefined,
+    isRequired: undefined,
+    name: undefined,
+    onChange: undefined,
+    value: undefined,
+    helpText: undefined,
+    errorText: undefined,
+    labelPosition: 'top',
+    children: 'Toggle checkbox',
+  },
+} as Meta<RadioProps>;
+
+const RadioTemplate: ComponentStory<typeof RadioComponent> = ({ children, ...args }) => {
+  return <RadioComponent {...args}>{children}</RadioComponent>;
+};
+
+export const Default = RadioTemplate.bind({});
+Default.storyName = 'Default';
+
+export const Checked = RadioTemplate.bind({});
+Checked.storyName = 'Checked';
+Checked.args = {
+  isChecked: true,
+};
+
+export const Disabled = RadioTemplate.bind({});
+Disabled.storyName = 'Disabled';
+Disabled.args = {
+  isDisabled: true,
+};
+
+export const Invalid = RadioTemplate.bind({});
+Invalid.storyName = 'Invalid';
+Invalid.args = {
+  validationState: 'error',
+};
+
+const ControlledAndUncontrolledComp = () => {
+  const [checked, setChecked] = React.useState(false);
+
+  return (
+    <>
+      <RadioComponent name="fruits">Uncontrolled</RadioComponent>
+      <Text>&nbsp;</Text>
+      <RadioComponent
+        isChecked={checked}
+        onChange={({ isChecked }) => {
+          setChecked(isChecked);
+        }}
+        name="fruits"
+      >
+        Controlled
+      </RadioComponent>
+      <Text>Checked: {checked ? 'True' : 'False'}</Text>
+      <Box marginBottom="spacing.2" />
+      <Button size="small" onClick={() => setChecked((prev) => !prev)}>
+        Toggle checked
+      </Button>
+    </>
+  );
+};
+const _ControlledAndUncontrolled: ComponentStory<typeof RadioComponent> = () => {
+  return <ControlledAndUncontrolledComp />;
+};
+export const ControlledAndUncontrolled = _ControlledAndUncontrolled.bind({});
--- a/packages/blade/src/components/Radio/Radio.tsx
+++ b/packages/blade/src/components/Radio/Radio.tsx
@@ -1,0 +1,122 @@
+/* eslint-disable @typescript-eslint/no-shadow */
+import React from 'react';
+import { useRadio } from './useRadio';
+import { RadioIcon } from './RadioIcon';
+import { FormHintText } from '~components/Form/FormHintText';
+import Box from '~components/Box';
+import { CheckboxLabel } from '~components/Checkbox/CheckboxLabel';
+import { CheckboxInput } from '~components/Checkbox/CheckboxInput';
+import { CheckboxLabelText } from '~components/Checkbox/CheckboxLabelText';
+
+type OnChange = ({
+  isChecked,
+  event,
+  value,
+}: {
+  isChecked: boolean;
+  event?: React.ChangeEvent;
+  value?: string;
+}) => void;
+
+type RadioProps = {
+  /**
+   * If `true`, The Radio will be checked. This also makes the Radio controlled
+   * Use `onChange` to update its value
+   *
+   * @default false
+   */
+  isChecked?: boolean;
+  /**
+   * If `true`, the Radio will be initially checked. This also makes the Radio uncontrolled
+   *
+   * @default false
+   */
+  defaultChecked?: boolean;
+  /**
+   * The callback invoked when the checked state of the `Radio` changes.
+   */
+  onChange?: OnChange;
+  /**
+   * Sets the label text of the Radio
+   */
+  children: string;
+  /**
+   * Help text for the Radio
+   */
+  helpText?: string;
+  /**
+   * The name of the input field in a Radio
+   * (Useful for form submission).
+   */
+  name?: string;
+  /**
+   * The value to be used in the Radio input.
+   * This is the value that will be returned on form submission.
+   */
+  value?: string;
+  /**
+   * If `true`, the Radio will be disabled
+   *
+   * @default false
+   */
+  isDisabled?: boolean;
+  /**
+   * If `true`, the Radio input is marked as required,
+   * and `required` attribute will be added
+   *
+   * @default false
+   */
+  isRequired?: boolean;
+  /**
+   * If `error`, the Radio input is marked as invalid,
+   * and `invalid` attribute will be added
+   */
+  validationState?: 'error' | 'none';
+};
+
+const Radio = ({
+  defaultChecked,
+  isChecked,
+  isDisabled,
+  isRequired,
+  name,
+  value,
+  onChange,
+  children,
+  helpText,
+  validationState,
+}: RadioProps): React.ReactElement => {
+  const { state, ids, inputProps } = useRadio({
+    defaultChecked,
+    isChecked,
+    isDisabled,
+    isRequired,
+    name,
+    value,
+    onChange,
+  });
+  const _hasError = validationState === 'error';
+  const showHelpText = !_hasError && helpText;
+
+  return (
+    <CheckboxLabel inputProps={state.isReactNative ? inputProps : {}}>
+      <CheckboxInput
+        isChecked={state.isChecked}
+        isDisabled={isDisabled}
+        isNegative={_hasError}
+        inputProps={inputProps}
+      />
+      <RadioIcon isChecked={state.isChecked} isDisabled={isDisabled} isNegative={_hasError} />
+      <Box>
+        <CheckboxLabelText>{children}</CheckboxLabelText>
+        {showHelpText && (
+          <FormHintText id={ids?.helpTextId} variant="help">
+            {helpText}
+          </FormHintText>
+        )}
+      </Box>
+    </CheckboxLabel>
+  );
+};
+
+export { Radio, RadioProps };
--- a/packages/blade/src/components/Radio/RadioIcon.tsx
+++ b/packages/blade/src/components/Radio/RadioIcon.tsx
@@ -1,0 +1,45 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+import React from 'react';
+import { useTheme } from '~components/BladeProvider';
+import Svg from '~components/Icons/_Svg';
+import { getIn } from '~utils';
+import { CheckboxIconWrapper } from '~components/Checkbox/CheckboxIcon/CheckboxIconWrapper';
+import Circle from '~components/Icons/_Svg/Circle';
+
+const CheckedIcon = ({ color }: { color: string }) => {
+  return (
+    <Svg width="16px" height="16px" viewBox="0 0 16 16" fill="none">
+      <Circle cx="8" cy="8" r="7.5" fill={color} stroke={color} />
+      <Circle cx="8" cy="8" r="4" fill="white" />
+    </Svg>
+  );
+};
+
+export type RadioIconProps = {
+  isFocused?: boolean;
+  isDisabled?: boolean;
+  isNegative?: boolean;
+  isChecked?: boolean;
+};
+
+const RadioIcon = ({ isChecked, isDisabled, isNegative, isFocused }: RadioIconProps) => {
+  const { theme } = useTheme();
+  const defaultIconColor = getIn(theme, 'colors.brand.primary.200');
+  const disabledIconColor = getIn(theme, 'colors.brand.gray.500');
+  const iconColor = isDisabled ? disabledIconColor : defaultIconColor;
+
+  return (
+    <CheckboxIconWrapper
+      isFullRounded={true}
+      isFocused={isFocused}
+      isDisabled={isDisabled}
+      isNegative={isNegative}
+      isChecked={!!isChecked}
+    >
+      {isChecked ? <CheckedIcon color={iconColor} /> : null}
+    </CheckboxIconWrapper>
+  );
+};
+
+export { RadioIcon };
--- a/packages/blade/src/components/Radio/useRadio.ts
+++ b/packages/blade/src/components/Radio/useRadio.ts
@@ -1,0 +1,106 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+import React from 'react';
+import type { GestureResponderEvent } from 'react-native';
+import { useControllableState } from '../../hooks/useControllable';
+import { useId } from '../../hooks/useId';
+import { getPlatformType, makeAccessible } from '../../utils';
+import type { RadioProps } from './Radio';
+
+type UseRadioProps = Pick<
+  RadioProps,
+  'isChecked' | 'defaultChecked' | 'isDisabled' | 'isRequired' | 'onChange' | 'name' | 'value'
+> & { hasError?: boolean };
+
+const useRadio = ({
+  isChecked,
+  defaultChecked,
+  isDisabled,
+  isRequired,
+  hasError,
+  onChange,
+  name,
+  value,
+}: UseRadioProps) => {
+  const inputRef = React.useRef<HTMLInputElement>(null);
+
+  const isReactNative = getPlatformType() === 'react-native';
+  if (isChecked && defaultChecked) {
+    throw new Error(
+      `[Blade useRadio] Do not provide both 'isChecked' and 'defaultChecked' to useRadio. Consider if you want this component to be controlled or uncontrolled.`,
+    );
+  }
+
+  const [radioState, setRadioStateChange] = useControllableState({
+    value: isChecked,
+    defaultValue: defaultChecked ?? false,
+  });
+
+  const handleOnChange = (event: React.ChangeEvent<HTMLInputElement> | GestureResponderEvent) => {
+    if (isDisabled) {
+      event.stopPropagation();
+      event.preventDefault();
+      return;
+    }
+
+    setRadioStateChange((checked) => {
+      onChange?.({
+        isChecked: !checked,
+        event: event as React.ChangeEvent,
+        value,
+      });
+      return !checked;
+    });
+  };
+
+  const state = {
+    isReactNative,
+    isChecked: radioState,
+    setChecked: setRadioStateChange,
+  };
+
+  const idBase = useId('radio');
+  const helpTextId = useId(`${idBase}-helptext`);
+
+  const accessibilityProps = makeAccessible({
+    role: 'radio',
+    required: Boolean(isRequired),
+    hidden: !isReactNative,
+    invalid: Boolean(hasError),
+    disabled: Boolean(isDisabled),
+    checked: radioState,
+    describedBy: helpTextId,
+  });
+
+  if (isReactNative) {
+    return {
+      state,
+      inputProps: {
+        onPress: handleOnChange,
+        name,
+        value,
+        ...accessibilityProps,
+      },
+    };
+  }
+
+  return {
+    state,
+    ids: { helpTextId },
+    inputProps: {
+      ref: inputRef,
+      onChange: handleOnChange,
+      type: 'radio',
+      name,
+      value,
+      checked: radioState,
+      disabled: isDisabled,
+      required: isRequired,
+      ...accessibilityProps,
+    },
+  };
+};
+
+export type InputProps = ReturnType<typeof useRadio>['inputProps'];
+
+export { useRadio };
--- a/packages/blade/.storybook/react-native/storybook.requires.js
+++ b/packages/blade/.storybook/react-native/storybook.requires.js
@@ -27,6 +27,9 @@ const getStories = () => {
     require('../../src/components/Checkbox/Checkbox.stories.tsx'),
     require('../../src/components/Checkbox/CheckboxGroup.stories.tsx'),
     require('../../src/components/Icons/Icons.stories.tsx'),
+    require('../../src/components/Link/BaseLink/BaseLink.stories.tsx'),
+    require('../../src/components/Link/Link/Link.stories.tsx'),
+    require('../../src/components/Radio/Radio.stories.tsx'),
     require('../../src/components/SkipNav/SkipNav.stories.tsx'),
     require('../../src/components/Typography/BaseText/BaseText.stories.tsx'),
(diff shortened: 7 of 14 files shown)


packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,118 @@
...
+    isRequired,
+    name,
+    value,
+    onChange,
+  });
+  const _hasError = validationState === 'error';
+  const showHelpText = !_hasError && helpText;
+
+  return (
+    <SelectorLabel inputProps={state.isReactNative ? inputProps : {}}>
```
chaitanyadeorukhkar, line 102: Instead of returning `isReactNative` from the `useRadio` hook just directly use `getPlatformType()`. `isReactNative` is not logically a state for a Radio, right?
```suggestion
    <SelectorLabel inputProps={getPlatformType() === 'react-native' ? inputProps : {}}>
```

packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <CheckboxInput
```
chaitanyadeorukhkar, line 103: Why are we using `CheckboxInput`. If this is a common component then shouldn't we move this to Forms & call it `Input` and use it in both checkbox & radio?

packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,118 @@
...
+        isChecked={state.isChecked}
+        isDisabled={isDisabled}
+        isNegative={_hasError}
```
chaitanyadeorukhkar, line 106: Why not just have the prop as `hasError`? `isNegative` is another verbatim that we'll have to remember. For input whenever it's negative it has error so might as well just call is `hasError`?
```suggestion
        hasError={_hasError}
```

packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,118 @@
...
+        inputProps={inputProps}
+      />
+      <RadioIcon isChecked={state.isChecked} isDisabled={isDisabled} isNegative={_hasError} />
```
chaitanyadeorukhkar, line 109: Same here. Should we call this `hasError` too? There are 3 variables controlling an error state
- validationState
- hasError
- isNegative

I understand why we don't want to use validation state since it's a string check and boolean is more convenient so at least lets stick to one boolean only? `validationState='error'` & `hasError` seem related. isNegative seems a bit out of the odd. 
```suggestion
      <RadioIcon isChecked={state.isChecked} isDisabled={isDisabled} hasError={_hasError} />
```

packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <Box>
```
chaitanyadeorukhkar, line 110: Do we need a `Box` here or can we just use `React.Fragment` here?

packages/blade/src/components/Radio/RadioIcon.tsx
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+import React from 'react';
+import { useTheme } from '~components/BladeProvider';
+import Svg from '~components/Icons/_Svg';
```
chaitanyadeorukhkar, line 5: nit: We should do named exports here too since we've decided to do it everywhere 😅 

We can do this later as well. Not in this PR's scope

packages/blade/src/components/Radio/RadioIcon.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { getIn } from '~utils';
+import { CheckboxIconWrapper } from '~components/Checkbox/CheckboxIcon/CheckboxIconWrapper';
```
chaitanyadeorukhkar, line 7: Lets we extract this to a common place now? `InputIconWrapper`?

packages/blade/src/components/Radio/useRadio.ts
```diff
@@ -0,0 +1,106 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+import React from 'react';
+import type { GestureResponderEvent } from 'react-native';
```
chaitanyadeorukhkar, line 4: Importing from `react-native` in a file that can be used on web is dangerous. This will cause failures on projects that are react only and wouldn't be able to find `react-native`. Lets not do this?

packages/blade/src/components/Radio/useRadio.ts
```diff
@@ -0,0 +1,106 @@
...
+import { useControllableState } from '../../hooks/useControllable';
```
chaitanyadeorukhkar, line 5: use import aliases?

packages/blade/src/components/Radio/useRadio.ts
```diff
@@ -0,0 +1,106 @@
...
+  onChange,
+  name,
+  value,
+}: UseRadioProps) => {
+  const inputRef = React.useRef<HTMLInputElement>(null);
+
+  const isReactNative = getPlatformType() === 'react-native';
+  if (isChecked && defaultChecked) {
+    throw new Error(
+      `[Blade useRadio] Do not provide both 'isChecked' and 'defaultChecked' to useRadio. Consider if you want this component to be controlled or uncontrolled.`,
```
chaitanyadeorukhkar, line 30: If I as a consumer pass `isChecked` and `defaultChecked` will I get the same error? Seems misleading for a consumer since it talks about `useRadio`. If this is consumer facing then we should rephrase this to make it more generic for Radio

packages/blade/src/components/Radio/useRadio.ts
```diff
@@ -0,0 +1,106 @@
...
+    );
+  }
+
+  const [radioState, setRadioStateChange] = useControllableState({
```
chaitanyadeorukhkar, line 34: What does `setRadioStateChange` mean? The name seems too confusing. Can we have a more meaningful name here or add comments to indicate what this does?

packages/blade/src/components/Radio/useRadio.ts
```diff
@@ -0,0 +1,106 @@
...
+    value: isChecked,
+    defaultValue: defaultChecked ?? false,
+  });
+
+  const handleOnChange = (event: React.ChangeEvent<HTMLInputElement> | GestureResponderEvent) => {
+    if (isDisabled) {
+      event.stopPropagation();
+      event.preventDefault();
```
chaitanyadeorukhkar, line 42: These might not be available on react native. Can you check and ensure this doesn't break on react native?

packages/blade/src/components/Radio/useRadio.ts
```diff
@@ -0,0 +1,106 @@
...
+        isChecked: !checked,
+        event: event as React.ChangeEvent,
+        value,
+      });
+      return !checked;
+    });
+  };
+
+  const state = {
+    isReactNative,
```
chaitanyadeorukhkar, line 57: Don't think this belongs here. Let the consumer component call `getPlatformType` if it needs to figure out the platform

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
```diff
@@ -38,10 +38,12 @@ const variants = {
 
 export type CheckboxRectProps = Omit<CheckboxIconProps, 'state'> & {
   isChecked: boolean;
+  isFullRounded?: boolean;
```
chaitanyadeorukhkar, line 41: Just `isRounded`?
```suggestion
  isRounded?: boolean;
```

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
```diff
@@ -38,10 +38,12 @@ const variants = {
 
 export type CheckboxRectProps = Omit<CheckboxIconProps, 'state'> & {
```
chaitanyadeorukhkar, line 39: We should extract this component out into a common place and not import from Checkbox for Radio

chaitanyadeorukhkar (changes requested):

packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,118 @@
...
+  defaultChecked,
+  isChecked,
+  isDisabled,
+  isRequired,
+  name,
+  value,
+  onChange,
+  children,
+  helpText,
+  validationState,
```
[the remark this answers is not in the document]
anuraghazra: Will do, tho can't put on defaultChecked, isChecked since absence of them (undefined) will determine the behaviour too.

packages/blade/src/components/Radio/Radio.tsx
anuraghazra: Yeah right I guess we should do that.

packages/blade/src/components/Radio/Radio.tsx
anuraghazra: Fixed in next PRs which has FormLabel stuff

packages/blade/src/components/Radio/Radio.tsx
anuraghazra: 🤔 Did it because in here isNegative is a intent not a state. It just changes the visual appearance.

packages/blade/src/components/Radio/Radio.tsx
anuraghazra: Okay then, that makes sense. I'll made this change on #607 since these components are mostly either moved or renamed. I'll add a TODO.

packages/blade/src/components/Radio/Radio.tsx
anuraghazra: We do need Box, since we need to control the flex layout.

packages/blade/src/components/Radio/RadioIcon.tsx
anuraghazra: Yeah let's do this on a separate PR.

packages/blade/src/components/Radio/RadioIcon.tsx
anuraghazra: Yep already done in #607

packages/blade/src/components/Radio/useRadio.ts
anuraghazra: It's a `type` import.

packages/blade/src/components/Radio/useRadio.ts
anuraghazra: done

packages/blade/src/components/Radio/useRadio.ts
anuraghazra: Sure.

packages/blade/src/components/Radio/useRadio.ts
anuraghazra: Yeah right 😆

packages/blade/src/components/Radio/useRadio.ts
anuraghazra: Nope doesn't seem to break, This code is also present on Checkbox. I think React's synthetic event system mocks these on native.

packages/blade/src/components/Radio/useRadio.ts
anuraghazra: Removed.

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
anuraghazra: Done in PR2

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
```diff
@@ -38,10 +38,12 @@ const variants = {
...
   isChecked: boolean;
+  isFullRounded?: boolean;
 };
 
 const getCheckboxIconWrapperStyles = ({
   theme,
+  isFullRounded,
```
divyanshu013, line 46: Think you need to update the instances of `isFullRounded`

packages/blade/src/components/Icons/_Svg/Circle/Circle.d.ts
```diff
@@ -0,0 +1,11 @@
+export type CircleProps = {
```
divyanshu013, line 1: Would all be optional?

packages/blade/src/components/Radio/Radio.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+import type { ComponentStory, Meta } from '@storybook/react';
+import { Button, Text } from '..';
+import type { RadioProps } from './Radio';
+import { Radio as RadioComponent } from './Radio';
+import Box from '~components/Box';
+
+export default {
+  title: 'Components/Radio/Radio',
+  component: RadioComponent,
+  args: {
```
divyanshu013, line 12: I thought we don't need to specify `undefined` props below 🤔

packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,120 @@
+/* eslint-disable @typescript-eslint/no-shadow */
```
divyanshu013, line 1: Instead shouldn't we use variables with different names in local scope?

packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,120 @@
...
+    isChecked,
+    isDisabled,
+    isRequired,
+    name,
+    value,
+    onChange,
+  });
+  const _hasError = validationState === 'error';
+  const showHelpText = !_hasError && helpText;
+  const isReactNative = getPlatformType() === 'react-native';
```
divyanshu013, line 101: Can we move platform specific logic to respective files? .web, .native i.e.

packages/blade/src/components/Radio/useRadio.ts
divyanshu013: But it's coming from `react-native` package which won't exist on web only projects

divyanshu013: I'm in favor of having two different modules .web, .native for this. Thoughts?

packages/blade/src/components/Radio/useRadio.ts
```diff
@@ -0,0 +1,105 @@
...
+  isDisabled,
+  isRequired,
+  hasError,
+  onChange,
+  name,
+  value,
+}: UseRadioProps) => {
+  const inputRef = React.useRef<HTMLInputElement>(null);
+
+  const isReactNative = getPlatformType() === 'react-native';
```
divyanshu013, line 27: Can we move this logic to appropriate .web / .native?

divyanshu013 (commented):
Some queries. Also it seems the checkbox and text aren't perfectly horizontal center aligned, is it just me?

image

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
anuraghazra: Why?

packages/blade/src/components/Icons/_Svg/Circle/Circle.d.ts
anuraghazra: Probably some will be required.

packages/blade/src/components/Radio/Radio.stories.tsx
anuraghazra: Yeah wouldn't matter if we don't, just the sorting will be random.

packages/blade/src/components/Radio/Radio.tsx
anuraghazra: Not really feasible. 

We are just adding a prop: 

```
inputProps={isReactNative ? inputProps : {}}
```

Duplicating and maintaining two identical component just for a simple prop can be tricky.

packages/blade/src/components/Radio/useRadio.ts
anuraghazra: > But it's coming from react-native package which won't exist on web only projects

type imports are stripped away and won't even be imported/bundled on the web project. 

Anyway we will have this on each and every component when we have TS 4.7 module suffixes.

packages/blade/src/components/Radio/useRadio.ts
anuraghazra: Again, not feasible. 
Imagine maintaining useRadio twice with identical logic. 

.web / .native files makes more sense when there's little to know overlap or lets say when we use the Styled* components.

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
divyanshu013: Because we're calling it `isRounded`? [Ref](https://github.com/razorpay/blade/pull/605#discussion_r944391741)

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Radio/useRadio.ts
chaitanyadeorukhkar: Since its only type and its going to be stripped off, I think its okay. With TS 4.7 we'll see what to do

chaitanyadeorukhkar (approved):

packages/blade/src/components/Form/Selector/SelectorGroupField.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import type { CSSObject } from 'styled-components';
+import { getPlatformType, makeAccessible } from '~utils';
+import Box from '~components/Box';
+
+type SelectorGroupFieldProps = {
+  children: React.ReactNode;
+  labelledBy: string;
```
chaitanyadeorukhkar, line 8: Can we stick to american english since we're doing that everywhere? 😅 
```suggestion
  labeledBy: string;
```

chaitanyadeorukhkar: Or are we doing this cuz of a11y props?

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
```diff
@@ -50,12 +52,12 @@ const StyledInput = styled.input<HoverProps>(({ theme, isChecked, isDisabled, is
...
 }));
 
-const CheckboxInput = ({
+const SelectorInput = ({
   inputProps,
   isChecked,
   isDisabled,
   isNegative,
-}: HoverProps & { inputProps: InputProps }): React.ReactElement => {
+}: HoverProps & { inputProps: any }): React.ReactElement => {
```
chaitanyadeorukhkar, line 60: Oh no :(

chaitanyadeorukhkar: No other way around?

packages/blade/src/components/Radio/RadioGroup/useRadioGroup.ts
```diff
@@ -0,0 +1,102 @@
...
+  const idBase = useId('radio-group');
+  const labelId = `${idBase}-label`;
+  const fallbackName = name ?? idBase;
+  const [checkedValue, setValue] = useControllableState({
+    value,
+    defaultValue,
+    onChange: (v: string) => onChange?.({ value: v, name: fallbackName }),
+  });
+
+  // FIX React native bug, where it unchecks the radio on clicking it again
```
chaitanyadeorukhkar, line 54: Is this fixed?

packages/blade/src/components/Radio/RadioGroup/useRadioGroup.ts
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        setValue(() => v);
+      },
+      removeValue(): void {
+        if (isDisabled) {
+          return;
+        }
+
+        // @ts-expect-error TODO: fix this
```
chaitanyadeorukhkar, line 70: Looks like a pending todo. Can you take a look?

packages/blade/src/components/Radio/RadioIcon/Fade.native.tsx
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+import React from 'react';
+import type { EasingFn } from 'react-native-reanimated';
```
chaitanyadeorukhkar, line 4:
```suggestion
import type { EasingFunctionFactory } from 'react-native-reanimated';
```

packages/blade/src/components/Radio/RadioIcon/Fade.native.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return {
+    ...styles,
+  };
+});
+
+const Fade = ({ children, show, styles }: FadeProps) => {
+  const { theme } = useTheme();
+
+  const enterEasing = (theme.motion.easing.entrance.effective as unknown) as EasingFn;
+  const exitEasing = (theme.motion.easing.exit.effective as unknown) as EasingFn;
```
chaitanyadeorukhkar, line 21:
```suggestion
  const enterEasing = (theme.motion.easing.entrance.effective as unknown) as EasingFunctionFactory;
  const exitEasing = (theme.motion.easing.exit.effective as unknown) as EasingFunctionFactory;
```

packages/blade/src/components/Form/Selector/SelectorGroupField.tsx
anuraghazra: > Can we stick to american english since we're doing that everywhere? 😅

our `makeAccessible` follows the WAI-ARIA props which takes british `labelledBy`, let's keep it as is for consistency sake. 

> Or are we doing this cuz of a11y props?

Yes so that the group has a label reference of it's child text. It works similarly to `fieldset` & `legend`

```tsx
<fieldset>
  <legend>hello</legend>
</fieldset>

<SelectorGroupField labelledBy="id-1">
  <Text id="id-1">hello</Text>
</SelectorGroupField>
```

packages/blade/src/components/Radio/RadioIcon/Fade.web.tsx
```diff
@@ -0,0 +1,74 @@
...
+    animation: ${scaleIn} ${makeMotionTime(duration)}
+      ${theme.motion.easing.entrance.effective as string};
+  `;
+
+  const exit = css`
+    animation: ${fadeOut} ${makeMotionTime(duration)}
+      ${theme.motion.easing.exit.effective as string};
+  `;
+
+  const { isMounted, isVisible } = usePresence(!!show, {
```
chaitanyadeorukhkar, line 58: Can you add some comments here to help understand what we're doing with this?

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
anuraghazra: Could be, but fine for now. Anyways this component is only shared in Checkbox & radio for now. which has similar APIs

packages/blade/src/components/Radio/RadioGroup/useRadioGroup.ts
anuraghazra: Oh yeah I did fixed it, let me check. Maybe merge kerne ke time pe it got messed up.

packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
```diff
@@ -0,0 +1,76 @@
...
+  return {
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: '16px',
+    height: '16px',
+    borderWidth: '1px',
+    borderStyle: 'solid',
+    margin: '0px',
+    marginTop: '3px',
```
chaitanyadeorukhkar, line 64: Margin of 3px? Where is this?

packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
```diff
@@ -0,0 +1,76 @@
...
+  const backgroundColor = variants[variant].background[checked];
+  const borderColor = variants[variant].border[checked];
+
+  return {
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: '16px',
+    height: '16px',
+    borderWidth: '1px',
```
chaitanyadeorukhkar, line 61: Can we not use border tokens?

packages/blade/src/components/Radio/RadioIcon/Fade.native.tsx
anuraghazra: Can't do that. Keyframe() takes EasingFn only.

packages/blade/src/components/Radio/RadioIcon/Fade.web.tsx
anuraghazra: Will do.

packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
```diff
@@ -0,0 +1,76 @@
...
+    borderStyle: 'solid',
+    margin: '0px',
```
chaitanyadeorukhkar, line 63:
```suggestion
    margin: makeSpace(theme.spacing[0])
```

packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
anuraghazra: Not a token from figma. It's to vertical align the Icon to text. We can't use flex since It has to stay in the center of text not the whole wrapper.

packages/blade/src/components/Radio/useRadio.ts
```diff
@@ -0,0 +1,158 @@
...
+  if (isChecked && defaultChecked) {
+    throw new Error(
+      `[Blade Radio] Do not provide both 'isChecked' and 'defaultChecked' to useRadio. Consider if you want this component to be controlled or uncontrolled.`,
```
chaitanyadeorukhkar, line 78: Will this error be shown to the user if they pass `isChecked` & `defaultChecked` together? If yes then we need to rephrase this. User shouldn't be worried about what `useRadio` is

```suggestion
      `[Blade: Radio] Do not provide both 'isChecked' and 'defaultChecked' to useRadio. Consider if you want this component to be controlled or uncontrolled.`,
```

packages/blade/src/components/Radio/Radio.stories.tsx
```diff
@@ -0,0 +1,159 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+import type { ComponentStory, Meta } from '@storybook/react';
+import React from 'react';
+import { Text } from '../Typography';
+import type { RadioGroupProps } from './RadioGroup/RadioGroup';
+import { RadioGroup as RadioGroupComponent } from './RadioGroup/RadioGroup';
+import { Radio as RadioComponent } from './Radio';
+
+export default {
```
chaitanyadeorukhkar, line 9: Check the stories for `Checkbox` once. There are some things missing here like Subtitle, Usage & Example. Lets add that to make the documentation consistent

packages/blade/src/components/Radio/Radio.stories.tsx
```diff
@@ -0,0 +1,159 @@
...
+  title: 'Components/Radio',
```
chaitanyadeorukhkar, line 10: Lets also mention RadioGroup here so consumers know that this is a combined doc?
```suggestion
  title: 'Components/Radio & RadioGroup',
```

packages/blade/src/components/Radio/RadioGroup/useRadioGroup.ts
anuraghazra: @chaitanyadeorukhkar Yeah I did fixed it. In [here ](https://github.com/razorpay/blade/pull/605/files#diff-82055f61296921ac0c865ece1e8c254f6ebd3bdf0e2dee1c33d2244cfdac1634R95-R96)

packages/blade/src/components/Radio/RadioGroup/useRadioGroup.ts
anuraghazra: Just a TS error. setValue() doesn't accept undefined but I want it to, to be able to remove the selected radio. Should be fine.

packages/blade/src/components/Radio/RadioGroup/useRadioGroup.ts
chaitanyadeorukhkar: Remove the comment then?

packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
chaitanyadeorukhkar: This could be troublesome. 3px might not stay true to align it to the center in various circumstances - platform (android, ios, web), pixel density, a11y font changes, etc. Have you verified the alignment on Android as well as iOS?

packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
anuraghazra: We can.

packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
anuraghazra: I see yeah could be,  I don't see any difference in native. Anyway are we doing pixel density calculation in any of our spacing tokens?

chaitanyadeorukhkar: No but we shouldn't use margin to center align it. We should rely on some form of layout

PR #613, opened 2022-08-04
feat: add IconButton (internal)

divyanshu013:
## Description

Adds IconButton. Lot of snapshot changes bumping up the diff stats.

## Open things

Few Todos that needs inputs:
- color tokens for surfaceAction do not exist [Related thread](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1659520391997739?thread_ts=1659424387.058829&cid=G01B3LQ9H0W)
- currently I'm setting currentColor for svg and path using css since there isn't a way to pass currentColor to Icon. Should we allow that for Icons to enable this? [Related thread](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1659447713147139)
  - exposed `currentColor` on icon
- there are discrepancies in size between code and figma [Related thread](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1659517877637939?thread_ts=1659434153.316999&cid=G01B3LQ9H0W)

### Web

- [Storybook](https://61c19ee8d3d282003ac1d81c-oeanmowhaz.chromatic.com/?path=/story/components-iconbutton-internal--icon-button&globals=measureEnabled:false)

| Default | Focus | High contrast |
| - | - | - |
|  image | image | image |

### Native

1. low contrast
2. high contrast

| Default | Active (pressed) |
| - | - |
| Simulator Screen Shot - iPhone 13 - 2022-08-10 at 21 26 53 | Simulator Screen Shot - iPhone 13 - 2022-08-10 at 21 27 23 |
| Simulator Screen Shot - iPhone 13 - 2022-08-10 at 21 27 47 | Simulator Screen Shot - iPhone 13 - 2022-08-10 at 21 27 49 |

## Recording

Hard to see color diff on pressing button

https://user-images.githubusercontent.com/6682655/183979904-5a9ea324-b390-44e3-a15d-30f8e77b0c7d.mp4


divyanshu013:
@chaitanyadeorukhkar @anuraghazra @kamleshchandnani could you please take an early look for feedback?

packages/blade/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { makeAccessible } from '~src/utils/makeAccessible';
+
+export type IconButtonProps = {
+  /** Icon component to be rendered, eg. `CloseIcon` */
+  icon: IconComponent;
+  onClick: () => void;
+  size?: 'medium' | 'large'; // todo: discrepancy in size between code and figma
+  contrast?: 'low' | 'high';
+  /** **a11y**: Sets aria-label to help users know what the action does, eg 'Dismiss alert' */
+  accessibilityLabel: string;
```
anuraghazra, line 14: Let's format this correctly, and use a consistent jsdoc comment habit.

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
```diff
@@ -0,0 +1,46 @@
...
+    background: 'transparent',
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    color: linkColorToken.default[contrastToken],
+    transition: `all ${motionToken.duration.xquick} ${motionToken.easing.standard.effective}`,
+
+    'svg, path': {
+      // todo: should be ideally set in Icon component
+      fill: 'currentColor',
```
anuraghazra, line 27: How does currentColor work in react-native?

divyanshu013: [Support](https://github.com/react-native-svg/react-native-svg/blob/main/USAGE.md#svg) is there. Though I haven't implemented the native code *yet*. Should work in theory 🤞

kamleshchandnani:
>currently I'm setting currentColor for svg and path using css since there isn't a way to pass currentColor to Icon. Should we allow that for Icons to enable this? [Related thread](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1659447713147139)

isn't the color of the iconbutton same as our existing icons?

divyanshu013:
No, the tokens seem different on Figma. Please check this part:

> color tokens for surfaceAction do not exist [Related thread](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1659520391997739?thread_ts=1659424387.058829&cid=G01B3LQ9H0W)

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
```diff
@@ -0,0 +1,46 @@
...
+    },
+
+    '&:hover': {
+      color: linkColorToken.hover[contrastToken],
+    },
```
kamleshchandnani, line 32: most of the styles overlap with button, do you think it'll be better to abstract it in a function(like I did for typography) and expose and use it here as well or it might be premature abstraction?

packages/blade/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,39 @@
+import type { ReactElement } from 'react';
```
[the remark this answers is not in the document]
divyanshu013: Thinking `components/IconButton`. But we can consider something else for internal components too.

packages/blade/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+import type { IconComponent } from '..';
+
+import StyledIconButton from './StyledIconButton.web';
+import { makeAccessible } from '~src/utils/makeAccessible';
```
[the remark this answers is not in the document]
divyanshu013: VSCode auto import 😅

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
```diff
@@ -0,0 +1,46 @@
...
+    border: 'none',
+    cursor: 'pointer',
+    padding: 0,
+    borderRadius: theme.border.radius.small,
+    background: 'transparent',
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    color: linkColorToken.default[contrastToken],
+    transition: `all ${motionToken.duration.xquick} ${motionToken.easing.standard.effective}`,
```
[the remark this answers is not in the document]
divyanshu013: Color transition on hover / focus / active

packages/blade/src/components/IconButton/IconButton.tsx
divyanshu013: Right, we can definitely consider that.

Though for this PR, to keep things scoped, I'll format the JSDocs correctly.

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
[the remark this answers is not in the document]
divyanshu013: Definitely. Just wanted a 👍 for the approach. Will abstract along with native implementation. Was planning to do that in a separate PR to keep things small?

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
divyanshu013: I think based on our [discussion](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1659352774861379?thread_ts=1659348526.339609&cid=G01B3LQ9H0W) plan was to make this not adhere to Button's constraints. I think we can consider an abstraction later perhaps.

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+    '&:hover': {
+      color: linkColorToken.hover[contrastToken],
+    },
+
+    '&:focus': {
+      outline: 'none',
+      boxShadow: `0px 0px 0px 4px ${theme.colors.brand.primary[400]}`,
+      color: linkColorToken.focus[contrastToken],
+    },
```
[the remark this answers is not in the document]
divyanshu013: There are slight deviations here, for example considering button:
- We don't show a border here
- We don't change background

So if we want to abstract only focus ring we can consider moving `outline` and `boxShadow` separately, which does make sense because on Figma this is a separate common abstraction. Will dig around other components and update.

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
chaitanyadeorukhkar: React native doesn't support `&:hover`, `&:active` so even if we use `currentColor` there, we'll have to change the color with Pressable's `pressed` from it's style or children (similar to the implementation on Link). 

This approach looks cleaner for web though, we can avoid JS here 👌

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah. At the very least have a `getFocusRingStyles()`?

packages/blade/src/components/IconButton/IconButton.tsx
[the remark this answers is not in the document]
nashcheez: We should extract out common components into a separate directory and just keep the components we expose inside `components/*`. This could be similar to how we have the `utils` folder. Maybe have a `common/` that can include `IconButton`, `Label`, `ValidationState` etc. ?

packages/blade/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+/**
+ * Internal component for making clickable icons
+ */
+const IconButton = ({
+  icon: Icon,
+  onClick,
+  size = 'medium',
+  contrast = 'low',
```
nashcheez, line 24: I was of the impression we were only scoping a transparent icon button as part of this, why include a `contrast` then? Is the usage of the `contrast` specific to Alerts?

nashcheez: In our use cases in the dark background I see the `IconButton` will have a transparent background (like in Alerts). Maybe I am missing something?

Screenshot 2022-08-05 at 5 57 37 PM

packages/blade/src/components/IconButton/IconButton.tsx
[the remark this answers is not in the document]
divyanshu013: We also have other internal components which reside in the same `components/` directory like BaseButton, BaseInput, etc. 🤔

packages/blade/src/components/IconButton/IconButton.tsx
divyanshu013: For high contrast components, the colors are different:

| Low contrast | High contrast |
| - | - |
| `Action/Surface/Icon/lowContrast/Default`  | `Action/Surface/Icon/highContrast/Default` |
| image | image |

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
divyanshu013: So it seems with `react-native-svg` it can only derive the `currentColor` on elements inside `Svg`, not directly on `Svg`, having `color="currentColor"` crashes it. So it seems for native we'll have to use similar approach we did earlier with Animated and Pressable.

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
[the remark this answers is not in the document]
divyanshu013: I see that we're using focus ring in different ways, in some components as CSS and at some places via helper functions. Probably can be taken up as a separate PR.

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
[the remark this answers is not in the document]
divyanshu013: Don't think we need any abstraction here, please have a look at the native implementation.

packages/blade/src/components/IconButton/IconButton.tsx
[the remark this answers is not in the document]
kamleshchandnani: Would rather suggest to keep it closer where it belongs so for Label we have a Form directory where we have kept it and all other utils and helpers for form also go there so here i think we can move it to `components/Button`. thoughts?

packages/blade/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,54 @@
+import type { ReactElement } from 'react';
+
+import type { IconComponent } from '..';
+import StyledIconButton from './StyledIconButton';
+
+export type IconButtonProps = {
```
kamleshchandnani, line 6: how are we deciding the color?

packages/blade/src/components/IconButton/StyledIconButton.native.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+type GetIconColorTokenParams = {
+  isPressed: boolean;
+  contrast: ColorContrastTypes;
+};
+
+type State = 'active' | 'default';
+type IconColorToken = `feedback.neutral.action.icon.link.${State}.${ColorContrastTypes}Contrast`;
+
+const getIconColorToken = ({ isPressed, contrast }: GetIconColorTokenParams): IconColorToken => {
```
kamleshchandnani, line 17: not a blocker: inline it since it's used anywhere else?

packages/blade/src/components/Icons/index.tsx
```diff
@@ -16,7 +16,10 @@ type ActionIconColors = `action.icon.${DotNotationColorStringToken<
 
 export type IconSize = 'large' | 'medium' | 'small' | 'xlarge' | 'xsmall' | 'xxsmall';
 export type IconProps = {
-  color: ActionIconColors | FeedbackIconColors | FeedbackActionIconColors;
+  /**
+   * Color token (not to be confused with actual hsla value)
+   */
+  color: ActionIconColors | FeedbackIconColors | FeedbackActionIconColors | 'currentColor'; // currentColor is useful for letting the SVG inherit color property from its container
```
kamleshchandnani, line 22: should we rather add suffix as `Token`? if it wasted your whole day to debug 😅

kamleshchandnani (commented):
overall looks good to me, just few open questions.

also, there's an empty file - `packages/blade/src/components/IconButton/styles.ts` not sure if that's intended

kamleshchandnani:
@saurav12 @divyanshu013  we also need a small size icon button so we can use that in input field. let's sort the icon sizes issue before we merge this

packages/blade/src/components/IconButton/IconButton.tsx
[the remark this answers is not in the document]
divyanshu013: Sounds good, it's anyway a button variation. Will move it

packages/blade/.storybook/react-native/storybook.requires.js
```diff
@@ -1,15 +1,20 @@
...
+  addDecorator,
+  addParameters,
+  addArgsEnhancer,
+} from "@storybook/react-native";
 
-import '@storybook/addon-ondevice-notes/register';
-import '@storybook/addon-ondevice-controls/register';
-import '@storybook/addon-ondevice-backgrounds/register';
-import '@storybook/addon-ondevice-actions/register';
+import "@storybook/addon-ondevice-notes/register";
```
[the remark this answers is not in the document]
divyanshu013: It auto generated this with double quotes when I ran `yarn react-native:storybook:ios` 😅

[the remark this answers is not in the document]
kamleshchandnani: yeah but i guess then prettier makes it single quote on pre-commit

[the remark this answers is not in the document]
chaitanyadeorukhkar: +1. Try going to the file and saving. It should format this

packages/blade/.storybook/react-native/storybook.requires.js
```diff
@@ -1,15 +1,20 @@
 /* do not change this file, it is auto generated by storybook. */
```
chaitanyadeorukhkar, line 1: Need to run prettier here

packages/blade/src/components/IconButton/IconButton.stories.internal.tsx
```diff
@@ -0,0 +1,66 @@
...
+const meta: Meta<IconButtonProps> = {
+  title: 'Components/IconButton (Internal)',
+  component: IconButtonComponent,
+  args: {
+    size: 'medium',
+    contrast: 'low',
+    accessibilityLabel: 'Close',
+  },
+  argTypes: {
+    onClick: { action: 'onClick' },
```
chaitanyadeorukhkar, line 40: I don't think we need to specify this. It should assume all onClicks as actions, right?

packages/blade/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,54 @@
+import type { ReactElement } from 'react';
+
+import type { IconComponent } from '..';
```
chaitanyadeorukhkar, line 3: Lets be specific here?
```suggestion
import type { IconComponent } from './IconButton.d';
```

packages/blade/src/components/IconButton/__tests__/IconButton.native.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+  it('should render', () => {
+    const noop = () => {};
+    const { toJSON } = renderWithTheme(
+      <IconButton accessibilityLabel="Close" icon={CloseIcon} onClick={noop} />,
+    );
+
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should handle callback', () => {
```
chaitanyadeorukhkar, line 20: Do you mean should handle onClick? Callback would be the function passed as 1st argument to `onClick`. We're not asserting that here anywhere? Refer Button/Link's tests for callback vs onclick tests

packages/blade/src/components/IconButton/__tests__/IconButton.native.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const { getByRole } = renderWithTheme(
+      <IconButton accessibilityLabel="Close" icon={CloseIcon} onClick={onClick} />,
+    );
+    const iconButton = getByRole('button');
+    fireEvent.press(iconButton);
+
+    expect(onClick).toHaveBeenCalledTimes(1);
+  });
+
+  it('should have a11y label', () => {
```
chaitanyadeorukhkar, line 31: nit: 
```suggestion
  it('should have accessibility label', () => {
```

packages/blade/src/components/IconButton/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './IconButton';
```
chaitanyadeorukhkar, line 1: Named export? 😅 
```suggestion
export { IconButton, IconButtonProps } from './IconButton';
```

packages/blade/src/components/Icons/useIconProps/useIconProps.tsx
```diff
@@ -33,7 +33,7 @@ function useIconProps({
 }: IconProps): { height: string; width: string; iconColor: string } {
   const { theme } = useTheme();
   const { height, width } = getIconDimensions({ size });
-  const iconColor = get(theme.colors, color, '');
+  const iconColor = color === 'currentColor' ? 'currentColor' : get(theme.colors, color, '');
```
chaitanyadeorukhkar, line 36: nit
```suggestion
  const iconColor = color === 'currentColor' ? color : get(theme.colors, color, '');
```

chaitanyadeorukhkar (changes requested):

packages/blade/src/components/IconButton/StyledIconButton.native.tsx
```diff
@@ -0,0 +1,55 @@
...
+  onClick,
+  size,
+  contrast,
+  accessibilityLabel,
+}: StyledIconButtonProps): ReactElement => {
+  const [isPressed, setIsPressed] = useState(false);
+  const iconColorToken = getIconColorToken({ isPressed, contrast });
+
+  return (
+    <StyledPressable
```
chaitanyadeorukhkar, line 43: The clickable area is too small for a thumb to be able to record a press. Should add a [`hitSlop`](https://reactnative.dev/docs/pressable#hitslop) here?

divyanshu013: [Similar thoughts](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1660135022356289) but I think we want to wait for some research on this before taking a call.

packages/blade/src/components/Icons/useIconProps/useIconProps.tsx
divyanshu013: This was more explicit 😄

packages/blade/src/components/IconButton/index.ts
divyanshu013: Oh, are we doing this for all components now?

packages/blade/src/components/IconButton/__tests__/IconButton.native.test.tsx
divyanshu013: I think `a11y` is more accessible to read and type than `accessibility` 😅

packages/blade/src/components/IconButton/__tests__/IconButton.native.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import { fireEvent } from '@testing-library/react-native';
+
+import IconButton from '../';
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
+import { CloseIcon } from '~components/Icons';
+
+describe('<IconButton />', () => {
+  it('should render', () => {
+    const noop = () => {};
```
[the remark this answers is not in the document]
divyanshu013: Intentionally passing explicit function because I don't want to test the mock function in this block. Added a separate test case for mock function callback.

packages/blade/src/components/IconButton/StyledIconButton.native.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return `feedback.neutral.action.icon.link.${state}.${contrastType}`;
+};
+
+type StyledPressableProps = {
+  contrast: ColorContrastTypes;
+};
+
+const StyledPressable = styled.Pressable<StyledPressableProps>({
+  alignSelf: 'flex-start', // ensure button only takes needed width
```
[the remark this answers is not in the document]
divyanshu013: I see, my rationale was if I place this button in a full width box it should appear to the left (start) and not automatically center.

packages/blade/src/components/IconButton/IconButton.tsx
divyanshu013: VSCode auto import 😅

packages/blade/src/components/IconButton/IconButton.stories.internal.tsx
divyanshu013: Right, can be skipped, thought to make it explicit

packages/blade/.storybook/react-native/storybook.requires.js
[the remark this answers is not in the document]
divyanshu013: Oh okay, will do, the file has a warning at the top so didn't touch it by hand 😬 

```js
/* do not change this file, it is auto generated by storybook. */
```

packages/blade/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+   */
+  icon: IconComponent;
+  onClick: () => void;
+
+  /**
+   * Icon size
+   *
+   * @default 'medium'
+   */
+  size?: 'medium' | 'large'; // todo: discrepancy in size between code and figma
```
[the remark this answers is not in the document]
divyanshu013: Not yet, @chaitanyadeorukhkar has taken up a task for that. Will also convert this comment to an issue for tracking.

packages/blade/src/components/IconButton/IconButton.tsx
divyanshu013: With this approach, the consumer will also need to pass hover, active, focus colors, etc. Was replicating the [design](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=12702%3A149930) here. In figma the tokens are coming from `action/surface` 🤔 /cc @saurav12

packages/blade/src/components/IconButton/IconButton.tsx
divyanshu013: Discussing [above](https://github.com/razorpay/blade/pull/613#discussion_r944196472)

packages/blade/src/components/Icons/index.tsx
divyanshu013: That might be a good idea but it would be a breaking change so let's get some more opinions perhaps /cc @anuraghazra @chaitanyadeorukhkar

packages/blade/src/components/IconButton/StyledIconButton.native.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { makeAccessible } from '~utils';
+import type { ColorContrastTypes } from '~tokens/theme/theme';
+
+type GetIconColorTokenParams = {
+  isPressed: boolean;
+  contrast: ColorContrastTypes;
+};
+
+type State = 'active' | 'default';
+type IconColorToken = `feedback.neutral.action.icon.link.${State}.${ColorContrastTypes}Contrast`;
```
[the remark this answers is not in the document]
divyanshu013: Discussing [above](https://github.com/razorpay/blade/pull/613#discussion_r944196472).

packages/blade/src/components/IconButton/IconButton.tsx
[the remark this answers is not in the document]
divyanshu013: Moved in https://github.com/razorpay/blade/pull/613/commits/963d8b8d376877a21c3cc3b37953cf486ad341bb

packages/blade/src/components/IconButton/StyledIconButton.native.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah, that makes sense too. We can't predict what the consumer would want most of the times. I've been habitual to making it center because that has fit most use-cases in the past. Anyway, we've done center for Button, Link & Badge too. Would be better if we kept the same experience across components so that the consumer doesn't suddenly see some different behaviour for some components right?

packages/blade/src/components/IconButton/index.ts
chaitanyadeorukhkar: I think so, yes 😅

packages/blade/src/components/IconButton/StyledIconButton.native.tsx
chaitanyadeorukhkar: Responded there. We do need some hitSlop at least for IconButton. This would be too small for a human thumb to be able to press it

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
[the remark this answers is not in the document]
divyanshu013: Resolved in https://github.com/razorpay/blade/pull/613/commits/a7c992a1205f00da61795e47039c46eb4457f64a

packages/blade/src/components/IconButton/StyledIconButton.native.tsx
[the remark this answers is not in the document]
divyanshu013: Resolved in https://github.com/razorpay/blade/pull/613/commits/a7c992a1205f00da61795e47039c46eb4457f64a

packages/blade/src/components/IconButton/__tests__/IconButton.native.test.tsx
divyanshu013: Resolved in https://github.com/razorpay/blade/pull/613/commits/a7c992a1205f00da61795e47039c46eb4457f64a

packages/blade/src/components/IconButton/IconButton.tsx
divyanshu013: After discussions, plan is to go ahead with current implementation and pick the colors off the new surface.action tokens. Will create a followup once the tokens are updated on figma and code.

packages/blade/src/components/IconButton/StyledIconButton.native.tsx
divyanshu013: Resolved in https://github.com/razorpay/blade/pull/613/commits/1522ef438febf3bd4460334b869493793813b17d

packages/blade/src/components/IconButton/IconButton.stories.internal.tsx
divyanshu013: So it turns out without this, the click action doesn't work on android. Confirmed for buttons too. Works fine on ios. So we'll need these.

packages/blade/src/components/IconButton/StyledIconButton.native.tsx
divyanshu013: Holding on to this so we've design inputs first. Think we can merge this in for the moment and revisit later with more design inputs.

packages/blade/src/components/IconButton/index.ts
divyanshu013: Done in https://github.com/razorpay/blade/pull/613/commits/aaf6244a6828a7dd0e0c8278c201e6ab43ced2e1

packages/blade/src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Icon contrast
+   *
+   * @default 'low'
+   */
+  contrast?: 'low' | 'high';
```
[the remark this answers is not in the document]
divyanshu013: Discussing https://github.com/razorpay/blade/pull/613#discussion_r944196472.

chaitanyadeorukhkar (approved):

packages/blade/src/components/IconButton/StyledIconButton.web.tsx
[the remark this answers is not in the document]
divyanshu013: Tracking https://github.com/razorpay/blade/issues/638

packages/blade/src/components/IconButton/StyledIconButton.native.tsx
divyanshu013: Tracking https://github.com/razorpay/blade/issues/639

packages/blade/src/components/Button/IconButton/StyledIconButton.native.tsx
```diff
@@ -0,0 +1,49 @@
...
+  contrast,
+  accessibilityLabel,
+}: StyledIconButtonProps): ReactElement => {
+  const [isPressed, setIsPressed] = useState(false);
+  const getIconColorToken = (): IconColorToken => {
+    const contrastType = contrast === 'high' ? 'highContrast' : 'lowContrast';
+    const state = isPressed ? 'active' : 'default';
+
+    return `feedback.neutral.action.icon.link.${state}.${contrastType}`;
+  };
```
anuraghazra, line 33: We should extract this out of the component.

packages/blade/src/components/Icons/index.tsx
anuraghazra: @kamleshchandnani you mean. ActionIconColorToken?

packages/blade/src/components/Button/IconButton/StyledIconButton.native.tsx
divyanshu013: Did that first but later inlined it. [Ref](https://github.com/razorpay/blade/pull/613#discussion_r944114607)

anuraghazra (approved):
LGTM

packages/blade/src/components/Button/IconButton/StyledIconButton.native.tsx
anuraghazra: Fine I guess. Just it will get recreated on each render.

packages/blade/src/components/Icons/index.tsx
divyanshu013: The `color` prop -> `colorToken`

PR #614, opened 2022-08-04
chore: improve the button story with icons to show proper usage

chaitanyadeorukhkar:
Since we're since some incorrect usage of Icon with Buttons, wanted to try and enhance the story a bit to be clearer on how to use these icons. Not sure if this will have any impact at all but better this than nothing for now. 

**Before:**
Screenshot 2022-08-04 at 3 27 21 PM
**After:**
Screenshot 2022-08-04 at 3 26 51 PM


divyanshu013 (approved):
Keeping this in sync would be tough but much better now 👍

PR #621, opened 2022-08-05
infra: disable dependabot version update PRs

divyanshu013:
Only security updates will happen now
- [Docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit)

image


snitin315 (approved):
🔪 🤖

PR #622, opened 2022-08-05
chore: add appropriate types for onClick of Button & BaseButton

chaitanyadeorukhkar:
Add proper types for onClick of `Button` & `BaseButton`

### DIFF
--- a/packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
+++ b/packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
@@ -126,6 +126,31 @@ describe('<BaseButton />', () => {
     expect(onClick).toHaveBeenCalledTimes(1);
   });
 
+  it('should call function on click of button with event callback', () => {
+    const buttonText = 'Pay Now';
+    const onClick = jest.fn();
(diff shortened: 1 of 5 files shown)


packages/blade/src/components/Button/BaseButton/__tests__/BaseButton.web.test.tsx
```diff
@@ -126,6 +126,31 @@ describe('<BaseButton />', () => {
...
+    const button = getByRole('button');
+    fireEvent(
+      button,
+      new MouseEvent('click', {
+        bubbles: true,
+        cancelable: true,
+        screenX: 20,
+        screenY: 20,
+      }),
+    );
```
anuraghazra, line 142: import {userEvent} from 'react-testing-library/events'
const user = userEvent.setup()

user.click()

chaitanyadeorukhkar: Need to assert if callback has been called. Couldn't find a way to add that assertion with userEvents. Have you added such an assertion before? If not we can just keep it this way as long as our test is able to assert on the callback

anuraghazra (approved):

PR #602, opened 2022-08-03
feat(BaseInput): add BaseInput field

kamleshchandnani:
This PR implements `BaseInput` for web and react-native

**[API Decisions Doc](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Input/_decisions/decisions.md)**

Capabilities
**PR1(earlier this PR)**
- [x] `placeholder` text
- [x] `type` of input field
- [x] basic styling
- [x] `defaultValue`
- [x] `name` for form fields
- [x] `onChange`

**PR2 - #624(which is merged in this PR now)**
- [x] capability to make the field controlled
- [x] capability to make the field isDisabled
- [x] `label` for input field
- [x] handle validations on input field
- [x] capability for `isRequired`

**PR3 - #632(which is merged in this PR now)** 
 - [x] provision for `leadingIcon`
 - [x] provision for  `prefix`
 - [x] provision for  `interactionElement`(could be button or loader)
 - [x] provision for `suffix`
 - [x] provision for  `trailingIcon`
 - [x] Group storybook controls
 
**PR4 - #649 (which is merged in this PR now)**
 - [x] `autoFocus` input  
 - [x] `textAlign` the input field content
 - [x] keyboard return key type
 - [x] autofill suggestion type
 - [x] keyboard type

**PR5 - #665  (which is merged in this PR now)**
- [x] accessibility

**PR6 - #676   (which is merged in this PR now)**
- [x] animations

Web-Default State
Screenshot 2022-08-03 at 12 16 33 PM

Web-Focussed State
Screenshot 2022-08-03 at 12 16 38 PM

Native-Default State
Screenshot 2022-08-03 at 12 17 30 PM

Native-Focussed State
Screenshot 2022-08-03 at 12 17 43 PM

PR #624, opened 2022-08-05
feat(Input): make baseinput field controlled - PR2

kamleshchandnani:
This PR adds the following capabilities to the input field
- [x] capability to make the field controlled
- [x] capability to make the field isDisabled
- [x] label for input field
- [x] handle validations on input field
- [x] capability for isRequired

[API Decisions Doc](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Input/_decisions/decisions.md)

Previous PR link for some context
PR1 - https://github.com/razorpay/blade/pull/602

Remaining capabilities
PR3 - #632


packages/blade/src/components/Box/getBoxStyles.ts
```diff
@@ -4,14 +4,17 @@ import type { BoxProps } from './Box';
 import { getIn, makeSize, makeSpace } from '~utils';
 
 // allows 'auto' as spacing value
-const getAutoSpacingValue = <T extends string | undefined>({
+const getAutoSpacingValue = <T extends string | number | undefined>({
```
anuraghazra, line 7: Let's rename this function to "getSpacingValue" then.

packages/blade/src/components/Checkbox/useCheckbox.ts
```diff
@@ -5,7 +5,7 @@ import type { GestureResponderEvent } from 'react-native';
 import { useControllableState } from '../../hooks/useControllable';
 import { getPlatformType, makeAccessible } from '../../utils';
 import type { CheckboxProps } from './Checkbox';
-import { useFormId } from '~src/hooks/useFormId';
+import { useFormId } from '~components/Form/useFormId';
```
anuraghazra, line 8: Why put hooks in Form folder?

anuraghazra: Oh I see. the useFormId hook. I guess makes sense then.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -1,73 +1,166 @@
...
-type OnChange = ({
-  event,
-  name,
-  value,
-}: {
-  event?: React.ChangeEvent<HTMLInputElement>;
-  name?: string;
-  value?: string;
-}) => void;
+export type OnChange = ({ name, value }: { name?: string; value?: string }) => void;
```
anuraghazra, line 19: event?

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -1,73 +1,166 @@
...
+  handleOnChange: HandleOnChange;
+} => {
+  if (value && defaultValue) {
+    throw new Error(
+      `[Blade Input]: Either 'value' or 'defaultValue' shall be passed. This decides if the input field is controlled or uncontrolled`,
+    );
+  }
+
+  const handleOnChange: HandleOnChange = React.useCallback(
+    ({ inputName, inputValue }) => {
```
anuraghazra, line 123: Why did we called it `inputName` `inputValue` instead of just `name` `value`

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -1,73 +1,166 @@
...
+      let _inputValue = '';
+
+      if (getPlatformType() === 'react-native' && typeof inputValue === 'string') {
+        _inputValue = inputValue;
+      } else if (typeof inputValue !== 'string') {
+        // it's weird but TS forced me to write this much code where I could have just done "getPlatformType() === 'react-native' ? inputValue : inputValue?.target.value" :(
+        _inputValue = inputValue?.target.value ?? '';
+      }
```
anuraghazra, line 131: Does this look a bit more cleaner? 

> also uplift the `const isReactNative = getPlatformType() === 'react-native';` check to component body, no need to check on each and every keystroke. 

```ts
      const isValueString = typeof inputValue === 'string';
      if (isReactNative && isValueString) {
        _inputValue = inputValue;
      }
      if (!isValueString) {
        // it's weird but TS forced me to write this much code where I could have just done "getPlatformType() === 'react-native' ? inputValue : inputValue?.target.value" :(
        _inputValue = inputValue?.target.value ?? '';
      }
```

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -1,73 +1,166 @@
...
+  );
+
+  return { handleOnChange };
+};
+
+export const getHintType = ({
+  _validationState,
+  hasHelpText,
+}: {
+  _validationState: BaseInputProps['validationState'];
```
anuraghazra, line 148: No need of _ prefix

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -1,49 +1,39 @@
...
-    label,
-    labelPosition,
-    onChange,
-  });
+  isRequired,
+  isDisabled,
+  handleOnChange,
+  ...props
+}: StyledBaseInputProps): ReactElement => {
+  const [isFocused, setisFocused] = useState(false);
```
anuraghazra, line 27: We can utilize our useInteraction hook. 

```ts
  const { currentInteraction, onBlur, onFocus } = useInteraction();
  const isFocused = currentInteraction === 'focus';
```

packages/blade/src/components/Input/BaseInput/getBaseInputStyles.ts
```diff
@@ -1,26 +1,74 @@
...
-const getBaseInputStyles = ({ theme }: { theme: Theme }): CSSObject => ({
+export const getInputBackgroundAndBorderStyles = ({
+  theme,
+  isFocused,
+  isDisabled,
+  validationState,
+}: Pick<BaseInputProps, 'isDisabled' | 'validationState'> & {
+  isFocused?: boolean;
+  theme: Theme;
+}): CSSObject => {
```
anuraghazra, line 15: Let's not inline complex types on function body, looks very cluttered. 

Also `getInputBackgroundAndBorderStyles` & `getBaseInputStyles` has the exact same type defs, check if we can just reuse it.

```ts
type GetBaseInputStylesProps = Pick<BaseInputProps, 'isDisabled' | 'validationState'> & {
  isFocused?: boolean;
  theme: Theme;
};
```

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -77,26 +170,60 @@ export const BaseInput = ({
...
+      <Box
+        display="flex"
+        flexDirection={labelPosition === 'left' ? 'row' : 'column'}
+        justifyContent={labelPosition === 'left' ? 'center' : 'normal'}
+        alignItems={labelPosition === 'left' ? 'center' : 'normal'}
+      >
+        <FormLabel
+          as="label"
+          neccessityIndicator={neccessityIndicator}
+          id={labelId}
```
anuraghazra, line 197: Oh actually ID isn't needed when using it `as=label` can you remove it and make the ID prop on FormLabel optional.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -77,26 +170,60 @@ export const BaseInput = ({
...
+      </Box>
+      <Box marginLeft={labelPosition === 'left' ? 120 : 'auto'}>
+        <FormHint
+          type={getHintType({ _validationState: validationState, hasHelpText: Boolean(helpText) })}
+          helpText={helpText}
+          errorText={errorText}
+          successText={successText}
+          helpTextId={helpTextId}
+          errorTextId={errorTextId}
+          successTextId={successTextId}
```
anuraghazra, line 224: > TODO: write up a11y ids to inputs

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -1,73 +1,166 @@
...
+   * Renders when `validationState` is set to 'success'
    */
-  name?: string;
+  successText?: string;
   /**
-   * The callback function to be invoked when the value of the input field changes
+   * If `error`, the input is marked as invalid,
+   * and `invalid` attribute will be added
+   *
+   * If `success`, the input is marked as invalid,
```
anuraghazra, line 59: If success marked as invalid?

packages/blade/src/components/Form/FormLabel.tsx
```diff
@@ -112,7 +118,7 @@ const FormLabel = ({
   const width = position === 'left' && isDesktop ? '120px' : 'auto';
 
   return (
-    <Component htmlFor={htmlFor} style={{ width }} id={id}>
+    <Component htmlFor={htmlFor} style={{ width, flexShrink: 0 }} id={id}>
       <Box marginBottom="spacing.1">{textNode}</Box>
```
anuraghazra, line 122: Oops actually I forgot to do this: 

The margin bottom should only apply when position is top: 

```ts
position === 'left' ? 'spacing.0' : 'spacing.1'
```

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
kamleshchandnani: yeah some spaghetti code, fixed

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
kamleshchandnani: let me check that

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
kamleshchandnani: it'll be needed right? because we are not wrapping the text input inside label so we need to associate the label with the input id

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
kamleshchandnani: A11y is next milestone - PR4 maybe

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
kamleshchandnani: typo

packages/blade/src/components/Form/FormLabel.tsx
kamleshchandnani: ideally shouldn't matter but fine i'll fix it

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
kamleshchandnani: don't mind doing that but honestly i don't see any value of doing that. it's just moving things around 1 level up where we don't even get any other advantage. if `getPlatformType` and `typeof` operation would have been very heavy i would have done that but they are very very very light weight operations so wouldn't make any diff by moving up at the cost of readability.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
kamleshchandnani: scope clash 😢 
trying to fix it now since a lot of changes have gone through let me check

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
kamleshchandnani: don't need event in case of input field. just sending out name and value to the consumer. also event is diff on react native and essentially there's no use as of now

kamleshchandnani:
@anuraghazra addressed all your comments, can you re-review and then change the label to L2 once you're done? @divyanshu013 tagged you for L2 review which you can start in parallel

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
anuraghazra: label doesn't need ID. it only needs htmlFor. 

input needs ID.

packages/blade/src/components/Box/getBoxStyles.ts
```diff
@@ -4,14 +4,17 @@ import type { BoxProps } from './Box';
 import { getIn, makeSize, makeSpace } from '~utils';
 
 // allows 'auto' as spacing value
-const getAutoSpacingValue = <T extends string | undefined>({
+const getSpacingValue = <T extends string | number | undefined>({
```
divyanshu013, line 7: Why are we allowing numbers? Won't that break the purpose of tokens 🤔 
If there's a legit usecase we should add a comment to explain the intent

packages/blade/src/components/Form/FormLabel.tsx
```diff
@@ -108,12 +114,13 @@ const FormLabel = ({
   }
 
   const Component = as;
+  const isLabelLeftPositioned = position === 'left';
   // only set 120px label when device is desktop
-  const width = position === 'left' && isDesktop ? '120px' : 'auto';
+  const width = isLabelLeftPositioned && isDesktop ? '120px' : 'auto';
```
divyanshu013, line 119: Optional: I think we should move this width to somewhere on the top as a `SOME_CONSTANT` for ease of readability.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -1,73 +1,167 @@
...
+    throw new Error(
+      `[Blade Input]: Either 'value' or 'defaultValue' shall be passed. This decides if the input field is controlled or uncontrolled`,
+    );
+  }
+
+  const handleOnChange: HandleOnChange = React.useCallback(
+    ({ name, value }) => {
+      let _value = '';
+
+      if (getPlatformType() === 'react-native' && typeof value === 'string') {
```
divyanshu013, line 127: Can we move platform specific logic to .web / .native files?

divyanshu013 (commented):

packages/blade/src/components/Box/getBoxStyles.ts
chaitanyadeorukhkar: +1. Now that we don't need all the "absolute" hacks. Can we get avoid allowing free-form margins?

packages/blade/src/components/Form/FormLabel.tsx
chaitanyadeorukhkar: We should have a `formTokens.js` where we can store this. Doing the same for other components where with fixed dimensions

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
kamleshchandnani: okay i confused these 2 props.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
kamleshchandnani: what is the benefit of this useInteraction?

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
kamleshchandnani: so at every step you kind of ask this question that shall we move this to a new file or same file with a basic check. In this case it's just 2-3 things which are diff rest the entire thing is same so putting a check is less noisy and less duplication

packages/blade/src/components/Box/getBoxStyles.ts
kamleshchandnani: will fix this in Pr-3 since the everything resides there for this so will check

packages/blade/src/components/Form/FormLabel.tsx
kamleshchandnani: it's just one tiny thing don't think that creating a whole file for just this will add any value.

chaitanyadeorukhkar:
Cant render this PR on react native @kamleshchandnani 
Screenshot 2022-08-23 at 4 53 17 PM

divyanshu013 (approved):
Looks good to me.

kamleshchandnani:
han could be merge conflicts after icon thing 😢 

fixing it now

> Cant render this PR on react native @kamleshchandnani Screenshot 2022-08-23 at 4 53 17 PM

chaitanyadeorukhkar (approved):

PR #630, opened 2022-08-10
feat: add `Badge` component

chaitanyadeorukhkar:
Adds `Badge` component 

Link to API Decisions: 
- #595


packages/blade/src/components/Badge/badgeTokens.ts
```diff
@@ -0,0 +1,12 @@
+import type { BadgeProps } from './Badge';
+
+type BadgeMinHeight = 20 | 24;
+
+const minHeight: Record<NonNullable<BadgeProps['size']>, BadgeMinHeight> = {
+  small: 20,
+  medium: 24,
+};
+
+const maxWidth = 100;
```
chaitanyadeorukhkar, line 10: Waiting for @saurav12 to confirm `maxWidth`

packages/blade/src/components/Box/Box.d.ts
```diff
@@ -9,6 +9,7 @@ export type BoxProps = {
   alignItems?: CSSObject['alignItems'];
   justifyContent?: CSSObject['justifyContent'];
   alignSelf?: CSSObject['alignSelf'];
+  overflow?: CSSObject['overflow'];
```
chaitanyadeorukhkar, line 12: Need overflow on `Box` for truncating text on `Badge`

packages/blade/src/components/Button/Button/Button.stories.tsx
```diff
@@ -52,6 +60,8 @@ const Page = (): ReactElement => {
       <Primary />
       <Title>Properties</Title>
       <ArgsTable story={PRIMARY_STORY} />
+      <Description markdown=">Note: `icon` prop accepts an `IconComponent` of Blade which should be used as:" />
```
chaitanyadeorukhkar, line 63: I was adding a note for `icon` usage on `Badge` and thought I'd do the same for `Button`

packages/blade/src/components/Badge/badgeTokens.ts
saurav12: Please check for `maxWidth` for each device on [Platform](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=8110%3A157) & [Characteristics](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=8110%3A50) section on Figma.

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,76 @@
...
+import styled from 'styled-components';
+import { StyledBadge } from './StyledBadge';
+import type { IconComponent } from '~components/Icons';
+import { BaseText } from '~components/Typography/BaseText';
+import { getPlatformType } from '~utils';
+import Box from '~components/Box';
+
+type BadgeProps = {
+  children: string;
+  variant?: 'positive' | 'negative' | 'notice' | 'information' | 'neutral';
```
anuraghazra, line 12: Let's try to set these jsdoc comments for consumers, inline docs ftw: 

```
  /**
   * Sets the variant of badge
   *
   * @default 'neutral'
   */
```

> we should probably also add these in other components like Text and etc.

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,76 @@
...
+    throw new Error('[Blade: Badge]: Text as children is required for Badge.');
+  }
+
+  return (
+    <StyledBadge variant={variant} contrast={contrast} size={size}>
+      <Box
+        paddingRight="spacing.2"
+        paddingLeft="spacing.2"
+        display="flex"
+        flex={1}
```
anuraghazra, line 50: Is this needed?

packages/blade/src/components/Badge/__tests__/Badge.native.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
+afterAll(() => jest.restoreAllMocks());
+
+describe('<Badge />', () => {
+  it('should render Badge with default props', () => {
+    const label = 'Label';
+    const { toJSON } = renderWithTheme(<Badge>{label}</Badge>);
+    expect(toJSON()).toMatchSnapshot();
+  });
+  it('should throw an error when no children are passed', () => {
```
anuraghazra, line 14: Let's move the runtime error check at the end. 

and also 🙈 
Add space between each test. 

```diff
+ });
+      
+ it('should render small size Badge', () => {
```

packages/blade/src/components/Badge/Badge.tsx
anuraghazra: @chaitanyadeorukhkar withIcon story isn't working as intended

image

chaitanyadeorukhkar: Can you check the CI link generated for chromatic? Local messes up sometimes. It looks fine on my local

packages/blade/src/components/Badge/__tests__/Badge.native.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+  });
+  it('should render high contrast negative variant Badge', () => {
+    const label = 'Label';
+    const { toJSON } = renderWithTheme(
+      <Badge variant="negative" contrast="high">
+        {label}
+      </Badge>,
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
```
anuraghazra, line 79: @chaitanyadeorukhkar what do you think about using test.each function from jest? 

```ts

  const variants = ['positive', 'negative', 'notice', 'information', 'neutral'];
  test.each(variants)('should render low contrast %p variant Badge', (variant) => {
    const label = 'Label';
      <Badge variant={variant as BadgeProps['variant']} contrast="low">
        {label}
      </Badge>,
    );
  });

  test.each(variants)('should render high contrast %p variant Badge', (variant) => {
    const label = 'Label';
      <Badge variant={variant as BadgeProps['variant']} contrast="high">
        {label}
      </Badge>,
    );
  });
```

All of the manual permutations can be deleted.

packages/blade/src/components/Badge/Badge.tsx
anuraghazra: Yeah that's weird. Working on chromatic. Why is this happening?

packages/blade/src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+  children: 'Label',
+  variant: 'neutral',
+  fontWeight: 'regular',
+  contrast: 'low',
+  size: 'small',
+};
+Badge.storyName = 'Default';
+
+const BadgesWithVariantTemplate: ComponentStory<typeof BadgeComponent> = ({ ...args }) => {
+  const variants: Array<NonNullable<BadgeProps['variant']>> = [
```
anuraghazra, line 106: const variants = ['positive', 'negative', 'notice', 'information', 'neutral'] as const;

packages/blade/src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+    'positive',
+    'negative',
+    'notice',
+    'information',
+    'neutral',
+  ];
+
+  const getLabel = (label: string): string => {
+    return args.fontWeight === 'bold' ? label.toUpperCase() : capitalize(label);
+  };
```
anuraghazra, line 116: Should we add this logic in story? Won't people think just by looking at the story that if fontWeight is bold the badge automatically makes it uppercase?

packages/blade/src/components/Badge/Badge.tsx
anuraghazra: Oh i think storybook is messing up because in argsType we are defining the control as `select` with option with `Object.keys`. 

But then in `BadgesWithVariantTemplate` we are not doing: 
```ts
const IconComponent = iconMap[(args.icon as unknown) as string];
```

chaitanyadeorukhkar: Generally I clear storybook's cache and everything works okay

packages/blade/src/components/Badge/__tests__/Badge.native.test.tsx
chaitanyadeorukhkar: Looks interesting. I'll play around with this to see how it works. But I'm guessing this is a non-blocker. I'll move to L2 meanwhile I try this

packages/blade/src/components/Badge/Badge.stories.tsx
chaitanyadeorukhkar: Haha thanks! 🚀

packages/blade/src/components/Badge/Badge.stories.tsx
chaitanyadeorukhkar: They won't. If they see the code below the story, they can see that the label text is capitalized. Anyway, even if they do the moment they use it they'll realise it doesn't work that way so its okay. Not a huge learning curve there

packages/blade/.storybook/react-native/storybook.requires.js
```diff
@@ -23,10 +23,13 @@ argsEnhancers.forEach((enhancer) => addArgsEnhancer(enhancer));
 
 const getStories = () => {
   return [
+    require('../../src/components/Badge/Badge.stories.tsx'),
     require('../../src/components/Button/Button/Button.stories.tsx'),
     require('../../src/components/Checkbox/Checkbox.stories.tsx'),
     require('../../src/components/Checkbox/CheckboxGroup.stories.tsx'),
     require('../../src/components/Icons/Icons.stories.tsx'),
+    require('../../src/components/Link/BaseLink/BaseLink.stories.tsx'),
```
divyanshu013, line 31: When this runs on my machine it generates this file with double quotes 😬

packages/blade/src/components/Button/Button/Button.stories.tsx
divyanshu013: Nice, wish there was a way for storybook to automatically show this correctly in generated code 🤞

packages/blade/src/components/Badge/__tests__/Badge.web.test.tsx
```diff
@@ -0,0 +1,149 @@
+import { Badge } from '../Badge';
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.web';
+import { InfoIcon } from '~components/Icons';
+
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
```
divyanshu013, line 5: Why do we do this 🤔

packages/blade/src/components/Badge/__tests__/Badge.native.test.tsx
divyanshu013: I think we don't need snapshots for all these cases, makes the snapshots huge and difficult to review 😅

packages/blade/src/components/Badge/Badge.stories.tsx
divyanshu013: Non blocker: personally in favor of keeping storybooks vanilla, without any special logic

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,107 @@
...
+   * Sets the fontWeight of the label.
+   *
+   * @default 'regular'
+   */
+  fontWeight?: 'regular' | 'bold';
+};
+
+const StyledBaseText = styled(BaseText)(
+  (): CSSObject => {
+    if (getPlatformType() !== 'react-native') {
```
divyanshu013, line 51: Optional: Can we move platform specific styles / logic to `.web` module? In favor of keeping conditional logic cleaner

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,107 @@
...
+        paddingLeft="spacing.2"
+        display="flex"
+        flex={1}
+        flexDirection="row"
+        justifyContent="center"
+        alignItems="center"
+        overflow="hidden"
+      >
+        {Icon ? (
+          <Box paddingRight={Boolean(Icon) ? 'spacing.1' : 'spacing.0'} display="flex">
```
divyanshu013, line 88: This is a pain, creating a wrapper box for spacing et al 😅

packages/blade/src/components/Badge/StyledBadge.native.tsx
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components/native';
+import { getStyledBadgeStyles } from './getStyledBadgeStyles';
+import type { StyledBadgeProps } from './StyledBadge.d';
+
+const StyledBadge = styled.View<StyledBadgeProps>((props) => ({
+  ...getStyledBadgeStyles(props),
+  alignSelf: 'center',
```
divyanshu013, line 7: Why is this `center` and not `flex-start` 🤔

divyanshu013 (commented):
Looks good overall, some questions, some non blockers

packages/blade/src/components/Badge/Badge.tsx
anuraghazra: Wait I never noticed this. But we are only importing styled from `styled-components` right? 
So when this code runs on native doesn't it throw some kind of warning that we are not importing from `styled-components/native`? 

@chaitanyadeorukhkar

divyanshu013: I thought it automatically handled that 😹

packages/blade/.storybook/react-native/storybook.requires.js
chaitanyadeorukhkar: Yeah same for me. We just need to run prettier on it. I added it to the script now

packages/blade/src/components/Badge/__tests__/Badge.web.test.tsx
chaitanyadeorukhkar: Because there's an error that's thrown in one of the test cases. The test case is to ensure that the error is thrown (when no text is passed) but if we don't mock it, it will throw an error in the jest process itself

packages/blade/src/components/Badge/Badge.tsx
chaitanyadeorukhkar: Haha for react native, we alias `styled-components` to `styled-components/native` 😅  

So for such small cases its okay. I prefer doing `styled-components/native` when we're styling a native specific thing like View or Pressable. But for a component like BaseText its okay since we know BaseText internally will handle platform specific things

packages/blade/src/components/Badge/Badge.tsx
chaitanyadeorukhkar: Yes 😢

packages/blade/src/components/Badge/StyledBadge.native.tsx
chaitanyadeorukhkar: Same discussion like the one we had for IconButton. Here centering it makes even more sense since when a badge is placed in a row in a table, the most usual use-case is to center align it.

divyanshu013: On web, it is aligned to start 😀 
image

divyanshu013 (approved):
Looks good to me.

PR #634, opened 2022-08-11
fix(blade): Checkbox design changes 

anuraghazra:
New changes according to spacing spec: https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=15622%3A243373

- Fixed padding issues
- Fixed gutter between FormLabel
- Fixed alignment issues

### DIFF
--- a/.changeset/clever-crews-smell.md
+++ b/.changeset/clever-crews-smell.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix(blade): Checkbox design changes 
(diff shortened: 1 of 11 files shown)


divyanshu013 (commented):
Related to gap, we're planning to remove it right?

Also I think we're about to add the token for 2px so this change won't be needed right?

anuraghazra:
> Related to gap, we're planning to remove it right?
> 
> Also I think we're about to add the token for 2px so this change won't be needed right?

Regardless of 2px this change is needed, So what happened was there was few discrepancy between design and dev plus few margin issues which needed to be fixed. 

You can check [visual diff here](https://www.chromatic.com/test?appId=61c19ee8d3d282003ac1d81c&id=62f4f9232418add56d956a04)

divyanshu013 (approved):
Looks good to me. We should fix the spacing back to tokens once the values are updated for 2px token.

anuraghazra:
> Looks good to me. We should fix the spacing back to tokens once the values are updated for 2px token.

Yes that's a breaking change we need to do.

PR #637, opened 2022-08-16
chore: added override theme example

anuraghazra:


packages/blade/src/tokens/theme/overrideTheme.stories.tsx
```diff
@@ -0,0 +1,196 @@
...
+      <Title>Usage</Title>
+      <Highlight language="tsx">{outdent`
+        const customTheme = overrideTheme({
+          baseThemeTokens: paymentTheme, // theme to override
+          overrides: {
+            colors: {
+              onLight: {
+                brand: {
+                  primary: {
+                    '500': 'red',
```
chaitanyadeorukhkar, line 36: Lets pass some `hsla` color and not `red`. We should encourage people to use hsla just like we've used for our own colors

packages/blade/src/tokens/theme/overrideTheme.stories.tsx
```diff
@@ -0,0 +1,196 @@
...
+    },
+  },
+} as Meta;
+
+const OverrideThemeExample = (props: { hue: number }): React.ReactElement => {
+  // Global parameters
+  const hue = props.hue;
+  const primaryColor = `hsla(${hue}, 100%, 50%, 1)`;
+  const secondaryColor = `hsla(${hue}, 100%, 99%, 1)`;
+  const tertiaryColor = `hsla(${hue}, 100%, 96%, 1)`;
```
chaitanyadeorukhkar, line 86: Would it be also possible to create `primaryHoverColor`, `primaryActiveColor`, etc? Right now there are no hover and active states in the example. We shouldn't encourage people to override without hover & active states

packages/blade/src/tokens/theme/overrideTheme.stories.tsx
```diff
@@ -0,0 +1,196 @@
...
+  );
+};
+
+export default {
+  title: 'Utilities/overrideTheme',
+  args: {
+    hue: 28,
+  },
+  argTypes: {
+    hue: {
```
chaitanyadeorukhkar, line 63: Brilliant stuff 🔥

packages/blade/src/tokens/theme/overrideTheme.stories.tsx
anuraghazra: Yeah should be possible.

anuraghazra: Added

packages/blade/package.json
```diff
@@ -58,6 +58,7 @@
   },
   "dependencies": {
     "@babel/runtime": "7.16.5",
+    "outdent": "0.8.0",
```
chaitanyadeorukhkar, line 61: Do we really need this? Do you forsee us using this elsewhere too?

anuraghazra: We can use this on all the Highlight's in storybook. 

Writing in a single line is a bit messy. This override example is the biggest tho 

```tsx
      <Highlight language="tsx">
        {outdent`
          import { Badge } from '@razorpay/blade/components';
          import type { BadgeProps } from '@razorpay/blade/components';
        `}
      </Highlight>
```

chaitanyadeorukhkar: Okay. Can we have it as devDependencies then?

chaitanyadeorukhkar (approved):
LGTM. See if we can move outdent to devDependencies.

packages/blade/package.json
anuraghazra: I think that would break on storybook deploy right?

PR #641, opened 2022-08-17
fix: button spinner layout

chaitanyadeorukhkar:
Button spinner was broken on native but seemed okay on web. Had to modify some layout styles to keep it consistent across web & native. 

**Added props to Box**: 
- position
- top
- right
- bottom
- left

**Result**: 
Screenshot 2022-08-17 at 3 43 45 PM


packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -345,7 +345,16 @@ const BaseButton = ({
...
-        <Box display="flex" justifyContent="center">
+        <Box
+          display="flex"
+          justifyContent="center"
+          alignItems="center"
+          position="absolute"
+          top={0}
+          left={0}
+          bottom={0}
+          right={0}
```
anuraghazra, line 356: Hope you tested on web.

chaitanyadeorukhkar: Attached screenshot

anuraghazra: Oh you did.cool!

anuraghazra (approved):
LGTM, let's release add a patch.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
chaitanyadeorukhkar: Also, check the story https://61c19ee8d3d282003ac1d81c-cfzrfrubtk.chromatic.com/?path=/story/components-button--button-loading&globals=measureEnabled:false

chaitanyadeorukhkar: Also, check the story https://61c19ee8d3d282003ac1d81c-cfzrfrubtk.chromatic.com/?path=/story/components-button--button-loading&globals=measureEnabled:false

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -345,7 +345,16 @@ const BaseButton = ({
       motionEasing={motionEasing}
     >
       {isLoading ? (
-        <Box display="flex" justifyContent="center">
```
divyanshu013, line 348: This was working fine on web earlier right?

divyanshu013: Why wasn't this working on native? And how did absolute position solve that 🤔

chaitanyadeorukhkar: The box needs to take up all the available width & height. For it do so, it needs to be absolute. In fact the earlier implementation was incorrect since it wasn't absolute. The idea is that our actual button content will be behind this absolute box and the loader on top will take up the content's space and show a spinner in between

divyanshu013 (approved):

PR #646, opened 2022-08-18
fix: icon sizes for `Icon`, `IconButton`, `Button`, `Link` & `Spinner` components

chaitanyadeorukhkar:
**This PR fixes the following:**

For `Button` component, 
- Icon size for all sizes of Button with Text
- Icon size for all sizes of Button without Text
- Spinner size for all sizes of Button
- Horizontal padding for Icon only Button of all sizes

For `Link` component,
- Icon size for Link with Text
- Icon sizes for Link without Text

For `IconButton` component,
- The sizes for IconButton now map exactly to the ones on Figma

**❗️This PR also introduces a breaking change for the `Icon` component's `size` prop**
Earlier, the `size` prop had the following size to pixel mapping:
- **xxsmall:** 10px
- **xsmall**: 12px
- **small**: 16px
- **medium**: 20px
- **large**: 24px
- **xlarge**: 32px

Now, the correct `size` prop will have the following size to pixel mapping:
- **xsmall**: 8px
- **small**: 12px
- **medium**: 16px
- **large**: 20px
- **xlarge**: 24px
- **2xlarge**: 32px

> ⚠️ `xxsmall` is not an accepted value anymore. Instead, we have a new acceptable value of `2xlarge`.


.changeset/great-bobcats-suffer.md
```diff
@@ -0,0 +1,26 @@
+---
+"@razorpay/blade": minor
```
chaitanyadeorukhkar, line 2: Should we make a major version bump? Even though its a breaking change, I'd prefer not bumping the major version since we're at an early stage anyway

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -69,21 +69,61 @@ const buttonPadding: Record<
...
 
+const buttonIconOnlyPadding: Record<
+  NonNullable<BaseButtonProps['size']>,
+  Record<'top' | 'bottom' | 'left' | 'right', keyof Theme['spacing']>
+> = {
+  xsmall: {
+    top: 0,
+    bottom: 0,
+    left: 2, // should be `6px` as per design but we're making it `8px` since `6px` is not available as a spacing token
+    right: 2, // should be `6px` as per design but we're making it `8px` since `6px` is not available as a spacing token
```
divyanshu013, line 80: Should we add the missing spacing token?

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
chaitanyadeorukhkar: Not yet. This will be a longer exercise since it `6px` sits between 4 & 8 which would mean changing the spacing tokens and ensuring we change it everywhere 😅

chaitanyadeorukhkar: Spoke to Saurav, we shall keep it 8px here but pick this refactor up later

anuraghazra (approved):

PR #660, opened 2022-08-23
feat: add `blue` variant to `Badge`

chaitanyadeorukhkar:
Adds support for `blue` variant of `Badge`

**Blue variant:**
Screenshot 2022-08-23 at 3 31 29 PM


packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -46,6 +64,40 @@ type BadgeProps = {
...
+  contrast: NonNullable<BadgeProps['contrast']>;
+}): ColorProps => {
+  // const feedbackVariants = ['information', 'negative', 'neutral', 'notice', 'positive'] as const;
+  const props: ColorProps = {
+    iconColor: 'feedback.icon.neutral.lowContrast',
+    textColor: 'feedback.text.neutral.lowContrast',
+    backgroundColor: 'feedback.background.neutral.lowContrast',
+    borderColor: 'feedback.border.neutral.lowContrast',
+  };
+  if (variant != 'blue') {
```
chaitanyadeorukhkar, line 87: @anuraghazra know any better way to handle this? I ideally wanted to do 
```jsx
  const feedbackVariants = ['information', 'negative', 'neutral', 'notice', 'positive']

  if (feedbackVariants.includes(variant)) {
    props.iconColor = `feedback.icon.${variant}.${contrast}Contrast`;
     ...
  }
```
But when I do this `variant` doesn't strip away the value `blue` from it which leads to a TS error on `props.iconColor` since it doesn't recognize a valid value with `blue` as variant. When I do `if (variant != 'blue')`, it strips away the type properly.

anuraghazra:
@chaitanyadeorukhkar can you delete the "Badges" folder and move the API Decisions in "Badge"

packages/blade/src/components/Badge/Badge.tsx
anuraghazra: If you don't want to do variant !== 'blue'

you can add a type guard checking for feedbackVariants

```ts
const isFeedbackVariant = (variant: string): variant is Feedback => {
  const feedbackVariants = ['information', 'negative', 'neutral', 'notice', 'positive'] as const;
  return feedbackVariants.includes(variant);
};


if (isFeedbackVariant(variant)) {
  props.iconColor = `feedback.icon.${variant}.${contrast}Contrast`;
  props.textColor = `feedback.text.${variant}.${contrast}Contrast`;
  props.backgroundColor = `feedback.background.${variant}.${contrast}Contrast`;
  props.borderColor = `feedback.border.${variant}.${contrast}Contrast`;
} else {
  props.iconColor = `badge.icon.${variant}.${contrast}Contrast`;
  props.textColor = `badge.text.${variant}.${contrast}Contrast`;
  props.backgroundColor = `badge.background.${variant}.${contrast}Contrast`;
  props.borderColor = `badge.border.${variant}.${contrast}Contrast`;
}
```

anuraghazra (approved):
LGTM.

packages/blade/src/components/Badge/Badge.tsx
chaitanyadeorukhkar: Nice

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -19,7 +22,7 @@ type BadgeProps = {
    *
    * @default 'neutral'
    */
-  variant?: 'positive' | 'negative' | 'notice' | 'information' | 'neutral';
+  variant?: Feedback | 'blue';
```
divyanshu013, line 25: Confused: We also call these intent at some places, should we use that terminology instead of `Feedback`? If not when should we call these intent and when feedback?

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -46,6 +49,46 @@ type BadgeProps = {
...
+};
+
+const getColorProps = ({
+  variant,
+  contrast,
+}: {
+  variant: NonNullable<BadgeProps['variant']>;
+  contrast: NonNullable<BadgeProps['contrast']>;
+}): ColorProps => {
+  // const feedbackVariants = ['information', 'negative', 'neutral', 'notice', 'positive'] as const;
```
divyanshu013, line 71: Stale comment?

divyanshu013 (approved):
Looks good, minor comments.

PR #662, opened 2022-08-24
docs: add `Spinner` decisions

chaitanyadeorukhkar:
### [View Formatted File](https://github.com/razorpay/blade/blob/docs/spinner-decisions/packages/blade/src/components/Spinner/_decisions/decisions.md)

> Note: Some of these decisions are expected to change since design is iterating over spinner at the moment


packages/blade/src/components/Spinner/_decisions/decisions.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+### Accessibility
+- We'll accept a prop `accessibilityLabel` which will be passed on as `aria-label` for web & `accessibilityLabel` for native to aide screen readers
+- By default the value of `accessibilityLabel` will be `Loading` and can be overwritten with something more meaningful by the consumer
+- We will also set `aria-live` as `assertive` to inform a user with a screen-reader that something is loading
+
+
+## Open Questions
+- Should we set `aria-role` as `alert` for accessibility? [Carbon DS](https://carbondesignsystem.com/components/loading/accessibility#how-it-works) does this as well.
```
divyanshu013, line 52: Might need to do some research on this but can we instead use aria busy? alert doesn't sound like the best thing (this is not exactly an alert - and probably doesn't need to be this assertive). 

Regardless I think aria-live is a good idea to announce changes in label. Also can consider aria-hidden for the animating component.

divyanshu013 (commented):
Some queries, looks good otherwise.

packages/blade/src/components/Spinner/_decisions/decisions.md
```diff
@@ -0,0 +1,62 @@
...
+
+
+## BaseSpinner
+A BaseSpinner will contain some extra props like `intent`. This will not be exposed to the consumer and will only be used for internal components.
+
+### API
+
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| intent | `positive`,`negative`, `notice`, `information`, `neutral` | `undefinedg` | The intent of the spinner to be rendered. |  |
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: `neutral` is an intent. When the intent is `undefined` we use the `brand.gray` colors for Spinner

packages/blade/src/components/Spinner/_decisions/decisions.md
```diff
@@ -0,0 +1,62 @@
...
+| contrast | `low`, `high` | `low` | The contrast of the spinner to be rendered. |  |
+| size | `small`, `medium`, `large` | `medium` | The size of the spinner to be rendered. |  |
+| accessibilityLabel | `string` | `"Loading"` | The accessibility label (aria-label) for the spinner. |
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah we could make it mandatory for BaseSpinner for now and see how it affects our internal usage. Making it mandatory for consumers could lead to some friction

packages/blade/src/components/Spinner/_decisions/decisions.md
```diff
@@ -0,0 +1,62 @@
...
+
+
+## Spinner
+A wrapper around Spinner which will be exposed to the consumer to be used as a Spinner. This component will not allow passing `intent` to the Spinner.
+
+### API
+
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| accessibilityLabel | `string` | `"Loading"` | The accessibility label (aria-label) for the spinner. |
```
nashcheez, line 30: This is more of a question for design but have we considering adding an option for a label below the Spinner? There are many use cases of spinners being accompanied by labels like "Loading", "Searching" etc. and instead of a developer adding an extra div and then centrally aligning it to match the positioning of the spinner (which will be a repeated effort), maybe the design system could expose a prop for that? @saurav12 

This is also helpful for accessible users who can't perceive motion and rely on text content to grasp meaning.

packages/blade/src/components/Spinner/_decisions/decisions.md
```diff
@@ -0,0 +1,62 @@
...
+| contrast | `low`, `high` | `low` | The contrast of the spinner to be rendered. |  |
+| size | `small`, `medium`, `large` | `medium` | The size of the spinner to be rendered. |  |
```
nashcheez, line 32: Many design systems allow consumers to change the timing of the animation. Certain user flows may require a slower/faster animation than what we are providing as default. Or have we decided to be rigid when it comes to the timing? Given we have clearly defined motion tokens I think we can entertain timing as a prop  - a question for @saurav12 to pose to Dushyant.

packages/blade/src/components/Spinner/_decisions/decisions.md
chaitanyadeorukhkar: Yeah I'll look some more into what makes sense for a11y here. We wouldn't need additional element with aria-hidden if we have aria-label already right?

packages/blade/src/components/Spinner/_decisions/decisions.md
```diff
@@ -0,0 +1,62 @@
...
+
+### API
+
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| contrast | `low`, `high` | `low` | The contrast of the spinner to be rendered. |  |
+| size | `small`, `medium`, `large` | `medium` | The size of the spinner to be rendered. |  |
+| accessibilityLabel | `string` | `"Loading"` | The accessibility label (aria-label) for the spinner. |
+
+### Sample Usage
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Even in those cases, a declarative API should suffice. Do you think there are use-cases where a Declarative API just won't make sense for a Spinner?

packages/blade/src/components/Spinner/_decisions/decisions.md
chaitanyadeorukhkar: There was text before but they've removed it. @saurav12 can add more

packages/blade/src/components/Spinner/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: Right, a declarative API will definitely work too. Was thinking imperative one could be simpler to use in response to user actions / status updates like `spinner.show()`, `spinner.hide()`, `spinner.update()` etc.

We can go ahead with declarative for now and can reconsider later if needed.

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Spinner/_decisions/decisions.md
```diff
@@ -0,0 +1,62 @@
...
+
+
+## BaseSpinner
+A BaseSpinner will contain some extra props like `intent`. This will not be exposed to the consumer and will only be used for internal components.
+
+### API
+
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| accessibilityLabel | `string` | `undefined` | The accessibility label (aria-label) for the spinner. | ✅ |
```
saurav12, line 17: Why is it that in `BaseSpinner` the **`accessibilityLabel`** is required but not required on Normal `Spinner`?

saurav12: Does that mean a `label` is always required?

chaitanyadeorukhkar: `accessibilityLabel` is only for a11y purposes not for actually showing some text label. Making it mandatory for BaseSpinner to encourage our internal developers (blade developers) to add an `accessibilityLabel` wherever they use the BaseSpinner but we don't yet want to enforce this on external consumers. 

Ref: https://github.com/razorpay/blade/pull/662#discussion_r955699523

PR #663, opened 2022-08-24
chore(blade): added icon plop script & docs

### LINKED ISSUE #647: Add docs for adding new icons to storybook

anuraghazra:
- Added [plop](https://github.com/plopjs/plop) template and script to automate adding icons
- Added docs for adding new icons

fixes #647

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,7 +135,7 @@ Run
 yarn workspace @razorpay/blade plop
 ```
 
-This command automate the process of generating boilerplate code and automatically adding neccessary imports & exports. 
+This command automates the process of generating boilerplate code and automatically adds neccessary imports & exports. 
 
 First enter the icon name in Title Case.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,15 @@
+- [Contributing guidelines for `blade`](#contributing-guidelines-for-blade)
+  - [Development setup](#development-setup)
+    - [Prerequisites](#prerequisites)
+    - [Setting up web](#setting-up-web)
+    - [Setting up iOS](#setting-up-ios)
+    - [Setting up Android](#setting-up-android)
+  - [Adding New Icons](#adding-new-icons)
+    - [Step 1 - Copy SVG](#step-1---copy-svg)
+    - [Step 2 - Run Generator](#step-2---run-generator)
+    - [Step 3 - Finalize](#step-3---finalize)
+
+
 # Contributing guidelines for `blade`
 
 ## Development setup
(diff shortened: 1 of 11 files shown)


divyanshu013 (commented):
I think, instead of adding this, can we do this with [svgr](https://react-svgr.com/playground/?native=true) instead? Advantage - lesser things to maintain for us and it has a nice GUI so people don't have to do lot of CLI things, etc.

divyanshu013:
Example playground [link](https://react-svgr.com/playground/?expandProps=none&native=true&replaceAttrValues=%23132644%3D%7BiconColor%7D&svgProps=width%3D%7Bwidth%7D%2Cheight%3D%7Bheight%7D&svgo=false)

anuraghazra:
Will check it out. Maybe we can make the svg->jsx conversation with svgr, though if you are talking about full svg -> react-component then there will still be few things to change.

divyanshu013:
Yes, just SVG to JSX is fine imo (even with just the GUI). The issue with adding another script is with maintenance, because it would still need some manual steps in between. Thoughts @chaitanyadeorukhkar ?

chaitanyadeorukhkar:
Plop seems to be not working at all for me. Just gets stuck. Tried with multiple icons. Instead of running from root, I'm running from `packages/blade` 🤔 
Screenshot 2022-08-29 at 12 34 36 PM

chaitanyadeorukhkar:
SVGR has a CLI. Have we explored that? Ref: https://react-svgr.com/docs/cli/

chaitanyadeorukhkar:
Regarding the CLI vs GUI approach. I'm okay with using a CLI since it does everything with a single command. I have more confidence on the GUI approach because SVGR has been reliable for a long time. If we can use the SVGR CLI, it would be the best of both worlds - reliability, and automation.

divyanshu013:
We can also do this in phases, right now add docs with svgr gui and later add cli based automation (if it takes time)

anuraghazra:
We can do one thing. Keep the plop templates 
it will generate the boilerplate test files folder etc. And let user replace the SVGs

anuraghazra:
> Tried with multiple icons. Instead of running from root, I'm running from packages/blade 🤔

@chaitanyadeorukhkar I see I think it failed to open a text editor.

chaitanyadeorukhkar:
> > Tried with multiple icons. Instead of running from root, I'm running from packages/blade 🤔
> 
> @chaitanyadeorukhkar I see I think it failed to open a text editor.

It did open a text editor, I pasted the SVG and then it got stuck

anuraghazra:
@chaitanyadeorukhkar ohh i see. You need to save the file and exit the TextEditor app from the app tray. (not just close the editor window)

chaitanyadeorukhkar:
> @chaitanyadeorukhkar ohh i see. You need to save the file and exit the TextEditor app from the app tray. (not just close the editor window)

That seems like a very specific ask from the consumer. Can we not paste it in the terminal itself instead? Or VScode?

anuraghazra:
>  Can we not paste it in the terminal itself instead?

Pasting multines in terminal is a mess. Didn't worked. 
Tried vim too, it conflicts with the terminal already open. 

I don't know why macos is like this, closing the windows doesn't close the editor, i can try vscode.

divyanshu013:
Seems like this workflow might cause some confusions. I'm leaning more towards simpler approach with svgr GUI and later we can consider automation with svgr cli 🤔

anuraghazra:
Yeah that's what I'm suggesting. 

Let's just generate the minimal possible boilerplate and let users do the svg manipulation stuff

> We can do one thing. Keep the plop templates
it will generate the boilerplate test files folder etc. And let user replace the SVGs

divyanshu013:
How would the workflow look like with this approach?

anuraghazra:
> How would the workflow look like with this approach?

It will ask just the Icon name thats it. 
And it will create the whole boilerplate. 

And then users will copy paste the svg and make changes to the svg either svgr or however they like.

chaitanyadeorukhkar:
I think vscode should work. Git does it for commit messages. If we're just generating skeletons then I'm not sure if its worth the maintenance overhead since consumers would still need to refer the docs, follow precise steps and ensure they've manipulated the correct files.

divyanshu013:
I think we can just add some docs for now and later revisit. We can just reference svgr for now in the docs.

chaitanyadeorukhkar:
Agreed. @anuraghazra shall we close this PR then?

PR #671, opened 2022-08-26
chore: move outdent to devDependencies

chaitanyadeorukhkar:
Moves outdent to devDependencies


packages/blade/package.json
```diff
@@ -154,4 +154,4 @@
     "react": "17.0.2",
     "styled-components": "^5"
   }
-}
+}
```
divyanshu013, line 157: No newline at end of file?

divyanshu013 (approved):

packages/blade/package.json
chaitanyadeorukhkar: Ugh prettier messes this up every time

divyanshu013: Will add an `.editorconfig` file

Issue #648, opened 2022-08-18
TypeScript should throw errors if people try to use standard svg properties on icons like `fill-rule`, `clip-rule`


divyanshu013:
I think these should be `fillRule`, `clipRule` (TS should ideally be warning these). Similarly at other places.

_Originally posted by @divyanshu013 in https://github.com/razorpay/blade/pull/645#discussion_r948940625_

divyanshu013:
Looks like this can't be done with TS, might have to add some tests - https://github.com/microsoft/TypeScript/issues/32447

PR #673, opened 2022-08-26
chore: fix path exact types

anuraghazra:
- Adds an Exact<> type which will fix the edge cases in SVG Path props

### DIFF
--- a/packages/blade/src/components/Icons/_Svg/Path/Path.d.ts
+++ b/packages/blade/src/components/Icons/_Svg/Path/Path.d.ts
@@ -1,6 +1,8 @@
-export type NonDashedProps<T> = {
-  [K in keyof T]: K extends `${string}-${string}` ? never : T[K];
-};
+// https://github.com/microsoft/TypeScript/issues/12936#issuecomment-368244671
+export type Exact<T, X extends T> = T &
+  {
+    [K in keyof X]: K extends keyof T ? X[K] : never;
+  };
 
 export type PathProps = {
   clipPath?: string;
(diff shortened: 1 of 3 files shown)


packages/blade/src/components/Icons/_Svg/Path/Path.d.ts
```diff
@@ -1,6 +1,8 @@
-export type NonDashedProps<T> = {
-  [K in keyof T]: K extends `${string}-${string}` ? never : T[K];
-};
+// https://github.com/microsoft/TypeScript/issues/12936#issuecomment-368244671
```
divyanshu013, line 1: Can add some more explanation inline here perhaps, hard to understand the linked comment 😅

anuraghazra: Will try. Kinda hard to explain that huge discussion in a few lines though.

chaitanyadeorukhkar: Just a description on what Exact does. Don't need to add all the context on why we need and all

chaitanyadeorukhkar (approved):
LGTM. Just add a description on what the Exact is doing

packages/blade/src/components/Icons/_Svg/Path/Path.d.ts
anuraghazra: Added 

image

anuraghazra:
Reference https://github.com/razorpay/blade/issues/648

PR #675, opened 2022-08-26
feat(blade): Improve platform types with TS 4.7

anuraghazra:
Changes in this PR: 

1. Upgrades to TS 4.8
2. [RollUp.config changes
](https://github.com/razorpay/blade/pull/675/files#diff-0f3a84aa9c55612e700bb36db9fc13a096abb22d16f485a0ebb9f701e365e144)
3. Imports/file names/type changes

# TypeScript Module Suffixes Migration

## Goals

- TypeScript 4.7+ versioned consumers should get correct type defs depending on their projects 
- Ensure backward compatibility with older versions of TypeScript
- Provide some standard type utilities to help write cross platform types for blade developers 

## Build Process 

1. Generate types for web/native with `tsconfig` specific to platform
2. Copy types for web/native into `build`
3. Output and bundle both types into `build` folder with rollup

In step `1` it is important to exclude other platform's files. For example if we are building for `web` we should ignore `.native` files. 

## Backward Compatibility

To ensure that our types work in older TypeScript version we outputed our `.web` targetted types from rollup to emit `index.d.ts` file instead of `index.web.d.ts`. 
This way even on older version TypeScript can pick up `index.d.ts` file which will give users `web` types only. 

And for native we are emitting `index.native.d.ts` which TypeScript 4.7+ will pick it up based on `moduleSuffixes` option in `tsconfig.json`

## Blade DX 

With this system in place, blade developers have to be bit more deligent while writing types. 

Tips: 

1. If you have platform specific types or files ensure they are properly suffixed `.native` & `.web`
2. If you don't have `filename.web.ts` suffix then unsuffixed `filename.ts` will picked up.
3. Also no more `export default from './Button.web'` in `Button.d.ts` files

### Writing Types

To ease the pain of writing cross-platform types we've added few utilities `Platform.Select`, `Platform.CastWe` `Platform.CastNative`

In cases when writing cross-platform component which has no platform suffix and you need to switch between types on correct platform use the `Platform.Select` utility.

```tsx
// blade internal
type ButtonProps = {
  onClick?: Platform.Select<{
    web: (event: React.MouseEvent<HTMLButtonElement>) => void;
  }>;
}
const Button = (props: ButtonProps) => {
  return <>
}

// consumer end: on native `e` is GestureResponderEvent
<Button onClick={e => {}} />

// consumer end: on web `e` is MouseEvent
<Button onClick={e => {}} />
```

In some cases we need to cast to web or native types to ensure there are no platform specific erorrs, especially happens with `styled-component` types in those cases use `Platform.Cast*` types.

```ts
// blade internal: 
type ButtonProps = {
  onClick?: Platform.Select<{
    web: (event: React.MouseEvent<HTMLButtonElement>) => void;
  }>;
}

// blade internal: web.ts
const Styled = styled.div();
<Styled onClick={onClick as Platform.CastNative<typeof onClick>} />

// blade internal: native.ts
const Styled = styled.div();
<Styled onClick={onClick as Platform.CastWeb<typeof onClick>} />
```

**Module Resolution Examples:**

Here are few examples of module resolution. We should avoid few of these patterns see below: 

**Example 1:**

- Button.web.ts
- Button.native.ts

`.web` & `.native` platforms will be independantly pick the types. 

> **Note**
> Recommended, prefer explicit platform specific files

---

**Example 2**

- Button.ts
- Button.native.ts

For web the unsuffixed `Button.ts` will be taken. 
For native the `.native` file

> **Note**
> Okay but better to suffix when you know `Button.ts` is gonna be picked for web

---

**Example 3**

- Button.ts
- Button.web.ts

For native the unsuffixed `Button.ts` will be taken. 
For web the `.web` file

> **Note**
> HIGHLY not recommended, this will cause lot of confusion

---

**Example 4:**

- Button.ts

For both platforms Button.ts will be taken. 

> **Note**
> Okay, but blade developers have to make sure that all types work in both platforms.

---


packages/blade/package.json
```diff
@@ -37,16 +37,25 @@
     "prepublish": "yarn build",
     "build": "run-s build:*",
     "build:clean": "rm -rf build",
-    "build:generate-types": "tsc -p ./tsconfig-generate-types.json",
-    "build:generate-css-variables": "GENERATE_CSS_VARIABLES=true FRAMEWORK=REACT rollup -c && node ./scripts/generateCssThemeVariables.js",
-    "build:copy-declarations": "copyfiles -u 1 \"src/**/*.d.ts\" build/types",
+    "// Generate types & move type declarations": "",
```
anuraghazra, line 40: :p poor-mans-packagejson-comments. 

Let me know if it's fine otherwise I'll remove it.

packages/blade/rollup.config.js
```diff
@@ -164,7 +179,8 @@ const config = () => {
   return exportCategories
     .map((exportCategory) => [
       // bundle our declarations for each category `components`, `tokens` and `utils` and place it next to each category under `build`
-      getDeclarationsConfig({ exportCategory }),
+      getDeclarationsConfig({ exportCategory, isNative: false }),
+      getDeclarationsConfig({ exportCategory, isNative: true }),
```
anuraghazra, line 183: This is where we generate two typing files for native & web.

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -1,15 +1,15 @@
 import type { ReactElement } from 'react';
 import type { CSSObject } from 'styled-components';
 import styled from 'styled-components';
-import type { StyledBadgeProps } from './StyledBadge';
+import type { StyledBadgeProps } from './StyledBadge.d';
```
anuraghazra, line 4: Every `.d.ts` file imports now has to explicitly state the `.d` part since StyleBadge in here will point to `.web` or `.native`

packages/blade/src/components/Box/Box.d.ts
```diff
@@ -29,5 +29,3 @@ export type BoxProps = {
   maxHeight?: number;
   maxWidth?: number;
 };
-
-export { default } from './Box.web';
```
anuraghazra, line 33: No more .web reexports 🎉

packages/blade/src/components/Button/BaseButton/StyledBaseButton.web.tsx
```diff
@@ -1,6 +1,6 @@
 import styled from 'styled-components';
 import getStyledBaseButtonStyles from './getStyledBaseButtonStyles';
-import type { StyledBaseButtonProps } from './StyledBaseButton.d';
+import type { StyledBaseButtonProps } from './types';
```
anuraghazra, line 3: I'm just generally avoiding .d.ts files

Reason being: 

1. They are not type-checked
2. They were could interfere with typescript's resolution if we accidentally reexport .web files

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
```diff
@@ -36,7 +36,7 @@ const getHoverStyles = ({
     borderColor: getIn(theme, backgroundColor),
     backgroundColor: getIn(theme, backgroundColor),
     transitionTimingFunction: theme.motion.easing.standard.effective as string,
-    transitionDuration: makeMotionTime(theme.motion.duration['2xquick']),
+    transitionDuration: (makeMotionTime(theme.motion.duration['2xquick']) as unknown) as string,
```
anuraghazra, line 39: Make motion time returns `string` on web & `number` on native but Styled components only takes strings. 

While type checking native will fail, thus we need `as unknown as string`

packages/blade/tsconfig-generate-types.native.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "./tsconfig-generate-types.json",
+  "exclude": [
+    "**/**.test.ts",
+    "**/**.test.tsx",
+    "__tests__",
+    "**/**.stories.tsx",
+    "**/**.stories.internal.tsx",
+    "**/**.web.ts",
+    "**/**.web.tsx"
```
anuraghazra, line 10: Excluding .web files when generating types for native.

packages/blade/tsconfig.json
```diff
@@ -7,7 +7,8 @@
       "~utils": ["src/utils"],
       "~components/*": ["src/components/*"],
       "~tokens/*": ["src/tokens/*"],
-      "~src/*": ["src/*"],
-    }
+      "~src/*": ["src/*"]
+    },
+    "moduleSuffixes": [".all", ".web", ""]
```
anuraghazra, line 12: While developing internally for blade we also have this special `.all` file which we only use it for `Platform.Select`

What it does is, when resolving to this `.all` file `Platform.Select<{web, native}>` will return the union of both platform types instead of giving us either web or native. This helps while developing the components since now the platform specific type checking won't throw error in vscode.

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -1,13 +1,14 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
divyanshu013, line 1: Why do we need this?

packages/blade/src/components/Button/BaseButton/StyledBaseButton.web.tsx
divyanshu013: Sounds good to me, anyway `types` is a clearer name for the module

packages/blade/src/components/Checkbox/CheckboxIcon/Fade.native.tsx
```diff
@@ -6,7 +6,7 @@ import Animated, { Keyframe } from 'react-native-reanimated';
 import type { CSSObject } from 'styled-components';
 import styled from 'styled-components/native';
 import { useTheme } from '../../BladeProvider';
-import type { FadeProps } from './Fade';
+import type { FadeProps } from './Fade.d';
```
divyanshu013, line 9: Using `.d` here instead of `types`?

packages/blade/src/components/Form/Selector/SelectorGroupField.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import type { CSSObject } from 'styled-components';
-import type { AriaRoles } from '~utils';
 import { getPlatformType, makeAccessible } from '~utils';
 import Box from '~components/Box';
+import type { AriaRoles } from '~src/utils/makeAccessible/types';
```
divyanshu013, line 5: What's happening here? It's going `src/utils` instead of `utils` 😕

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
divyanshu013: This is quite confusing. Can we do without `makeMotionTime`? Let styled-components take care of things?

packages/blade/src/components/Icons/_Svg/Rect/Rect.native.tsx
```diff
@@ -1,6 +1,6 @@
 import type { ReactElement } from 'react';
 import { Rect as RectNative } from 'react-native-svg';
-import type { RectProps } from './Rect';
+import type { RectProps } from './Rect.d';
```
divyanshu013, line 3: At some places you're using `.d`. Are they meant to be moved to `types` too?

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -258,6 +260,7 @@ const BaseLink = ({
       motionDuration={motionDuration}
       motionEasing={motionEasing}
       setCurrentInteraction={setCurrentInteraction}
+      // @ts-ignore
```
divyanshu013, line 263: Why do we need this?

packages/blade/src/storybook-recipes/AccessibilityInterop/AccessibilityInteropDemo.web.tsx
```diff
@@ -39,6 +41,7 @@ const Checkbox: React.FC<{ label: string; checked?: boolean; disabled?: boolean
         {...checkboxAccessibilityProps}
         tabIndex={0}
         style={{
+          // @ts-ignore
```
divyanshu013, line 44: Why is this needed? And why did this uncover now? Due to TS upgrade?

packages/blade/src/utils/isReactNative/platform.all.ts
```diff
@@ -0,0 +1,32 @@
+// THIS .all file is special only for platform.ts file. Do not use anywhere else
```
divyanshu013, line 1:
```suggestion
// This .all file is special only for platform.ts file. Do not use anywhere else
```

packages/blade/src/utils/isReactNative/platform.all.ts
```diff
@@ -0,0 +1,32 @@
...
+// This file only contains type level code and putting any runtime code will cause errors since our
```
divyanshu013, line 2: So if I add a noop function here `const noop = () => {}` will it throw error in build time or run time?

packages/blade/src/utils/isReactNative/platform.all.ts
```diff
@@ -0,0 +1,32 @@
...
+  export type Name = 'web' | 'native';
+  // Unionify both web & native types, this will help us blade developers internally
+  export type Select<Options extends { web: any; native: any }> = IsFunction<
+    Options['native']
+  > extends true
+    ? IsFunction<Options['web']> extends true
+      ? // functions has to be intersected
+        Brand<Options['native'] & Options['web'], 'platform-all'>
+      : Brand<Options['native'], 'platform-native'> | Brand<Options['web'], 'platform-web'>
+    : Brand<Options['native'], 'platform-native'> | Brand<Options['web'], 'platform-web'>;
```
divyanshu013, line 19: Suggestion: can we simplify this? Break into multiple intermediate types?

packages/blade/src/utils/isReactNative/types.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable @typescript-eslint/no-namespace */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+// export type Brand<Base, Branding, ReservedName extends string = '__type__'> = Base &
+//   { [K in ReservedName]?: Branding };
```
divyanshu013, line 4:
```suggestion
```

packages/blade/tsconfig.json
divyanshu013: We don't add `.native` here?

divyanshu013 (commented):
Can we try publishing from this branch? And validate what consumers are getting? Can be done locally with `yalc` I think.

Also, do consumers need anything special on their TS sides for module suffixes?

packages/blade/src/components/Icons/_Svg/Rect/Rect.native.tsx
anuraghazra: I guess we can do it for the consistency sake. But they didn't really cause any issues so didn't bothered migrating them.  
I only faced issues in few of the components like StyledButton while I was migrating (tho that time didn't figure out the module resolution)

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
anuraghazra: It's because of this: https://github.com/razorpay/blade/pull/675/files/21e5af143380e3c1117e77b1114c873633dfeaf3#diff-8df6f3ac8deb957fe315382616c9e40354240ff966d0f5f089583dade523a644R196-R198

(This was added earlier so that we can extend the BaseLink with `styled(BaseLink)` but still don't have to expose the typescript API to public, [reference](https://github.com/razorpay/blade/pull/587#discussion_r938456767))

packages/blade/src/utils/isReactNative/platform.all.ts
anuraghazra: Will try.

packages/blade/src/utils/isReactNative/platform.all.ts
anuraghazra: Unless you import & use them.  

For example
Let's say you add `moduleSuffix: [".all", ".native"]` 

Now when you import `import noop from './example'` typescript will resolve the `.all` file with correct types but if your bundler isn't setup to handle the .all file extension this will fail.

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
anuraghazra: removed.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -1,6 +1,8 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
anuraghazra: Not even using any 😎 removed.

packages/blade/src/components/Box/Box.d.ts
```diff
@@ -28,6 +28,5 @@ export type BoxProps = {
   minWidth?: number;
   maxHeight?: number;
   maxWidth?: number;
+  className?: string;
```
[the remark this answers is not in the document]
anuraghazra: We are using className in BaseLink's box. `content-container`

package.json
```diff
@@ -60,7 +60,7 @@
     "stylelint-config-recommended": "5.0.0",
     "stylelint-config-styled-components": "0.1.1",
     "stylelint-processor-styled-components": "1.10.0",
-    "typescript": "4.6.3"
+    "typescript": "4.8.2"
```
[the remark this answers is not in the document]
anuraghazra: Apart from upgrading to 4.7+ 
If consumer is using web target then no change is needed. Since by default it will be picked. 
If they use react-native then they need to add that `moduleSuffix` option. I'll add docs for this.

packages/blade/tsconfig.json
anuraghazra: This tsconfig is blade's tsconfig. We can't have both web&native types resolution at once. No possible. We either have to have web or native target here. 

That's why we also added that .all, Platform.Select etc utilties.

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
anuraghazra: Agreed, thoughts @chaitanyadeorukhkar ?

chaitanyadeorukhkar: Still going through the PR and building context. Will come back to this 😁

packages/blade/src/components/Form/Selector/SelectorGroupField.tsx
anuraghazra: Ahh forgot to reexport the types. Done.

packages/blade/src/storybook-recipes/AccessibilityInterop/AccessibilityInteropDemo.web.tsx
anuraghazra: This is Due to TS upgrade. 

It has to do with Box's style prop definition different from react-native to web. So ts-ignored for now.

packages/blade/src/utils/isReactNative/platform.all.ts
anuraghazra: Done.

packages/blade/src/utils/isReactNative/platform.all.ts
anuraghazra: I've simplified it a bit, let me know if it's more readable. 

```ts
  // Unionify both web & native types, this will help us blade developers internally
  export type Select<Options extends { web: any; native: any }> = IsBothFunction<
    Options['native'],
    Options['web']
  > extends true
    ? // functions has to be intersected
      Brand<Options['native'] & Options['web'], 'platform-all'>
    : Brand<Options['native'], 'platform-native'> | Brand<Options['web'], 'platform-web'>;

```

packages/blade/tsconfig.json
divyanshu013: Suggestion to maybe add some comments in the code. Me from future will thank you 😸

packages/blade/src/components/Box/Box.d.ts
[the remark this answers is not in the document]
divyanshu013: Oh, didn't see in the Box implementation explicitly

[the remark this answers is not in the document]
divyanshu013: Anyway `className` won't work in native right?

packages/blade/src/components/Icons/_Svg/Rect/Rect.native.tsx
divyanshu013: I see, maybe we can have a followup issue to keep track of this for consistency

packages/blade/tsconfig.json
anuraghazra: Will do, good idea.

packages/blade/src/components/Box/Box.d.ts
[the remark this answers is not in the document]
anuraghazra: Yes but style will do, we need both for making this cross platform.

packages/blade/src/utils/isReactNative/platform.all.ts
divyanshu013: Basically it would fail the at the build step itself then?

packages/blade/src/utils/isReactNative/platform.all.ts
divyanshu013: A bit better

packages/blade/src/components/Box/Box.d.ts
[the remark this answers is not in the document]
anuraghazra: Yes and it's a web specific API. For native I think we are doing something else. It's just easier with className to reference the component and style that way.

divyanshu013:
Please also check [these comments](https://github.com/razorpay/blade/pull/675#pullrequestreview-1094394329):
> Can we try publishing from this branch? And validate what consumers are getting? Can be done locally with yalc I think.

> Also, do consumers need anything special on their TS sides for module suffixes?

anuraghazra:
> Can we try publishing from this branch? And validate what consumers are getting? Can be done locally with yalc I think.

@divyanshu013 Yes I've testing out with yalc already. 

> Also, do consumers need anything special on their TS sides for module suffixes?

Yes
1. Upgrade to 4.7+
2. in `tsconfig.json` add `moduleSuffix: ['.web', '']` or ``moduleSuffix: ['.native', '']`` (depending on the platform)

divyanshu013:
Nice, we can add the docs / changelog too

packages/blade/src/utils/isReactNative/platform.all.ts
anuraghazra: Yeah right.

packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
```diff
@@ -72,7 +72,7 @@ const getRadioIconWrapperStyles = ({
     marginRight: makeSpace(theme.spacing[2]),
     backgroundColor: getIn(theme, backgroundColor),
     borderColor: getIn(theme, borderColor),
-    transitionDuration: makeMotionTime(theme.motion.duration.xquick),
+    transitionDuration: (makeMotionTime(theme.motion.duration.xquick) as unknown) as string,
```
chaitanyadeorukhkar, line 75: Now that we have proper type support shouldn't we be getting rid of such hacks?

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-unnecessary-type-assertion */
 import type { ReactElement } from 'react';
 import { BaseText } from '../BaseText';
-import type { BaseTextProps } from '../BaseText';
+import type { BaseTextProps } from '../BaseText/BaseTextTypes';
```
chaitanyadeorukhkar, line 4: Call this just `types`?

chaitanyadeorukhkar (commented):
Left some comments but wasn't able to do a thorough review. Still need to spend some time on this PR to understand a few things better. 

Things like motion tokens still don't seem flacky and might need to see how we can fix this

packages/blade/package.json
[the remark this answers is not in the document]
anuraghazra: Sure. Was getting out of hand while I was developing so added these temporarily. Will remove.

packages/blade/package.json
```diff
@@ -37,16 +37,25 @@
...
+    "build:generate-types": "run-s types:generate-types:*",
```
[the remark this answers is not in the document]
anuraghazra: Good idea. Will do.

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -1,15 +1,15 @@
...
 import { StyledBadge } from './StyledBadge';
 import { useTheme } from '~components/BladeProvider';
 import type { IconComponent, IconProps } from '~components/Icons';
-import type { BaseTextProps } from '~components/Typography/BaseText';
 import { BaseText } from '~components/Typography/BaseText';
 import { getPlatformType } from '~utils';
 import Box from '~components/Box';
 import type { Feedback } from '~tokens/theme/theme';
+import type { BaseTextProps } from '~components/Typography/BaseText/BaseTextTypes';
```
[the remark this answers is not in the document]
anuraghazra: Yeah we can. will do, the file was already there didn't noticed.

packages/blade/src/components/Button/IconButton/StyledIconButton.native.tsx
```diff
@@ -2,7 +2,7 @@ import type { ReactElement } from 'react';
 import { useState } from 'react';
 import styled from 'styled-components/native';
 
-import type { StyledIconButtonProps } from './StyledIconButton';
+import type { StyledIconButtonProps } from './types';
```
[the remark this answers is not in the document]
anuraghazra: I haven't done for all the files, I mean if you and @divyanshu013 both agree upon this we can do it. 

Check this comment: https://github.com/razorpay/blade/pull/675#discussion_r962521415

packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
anuraghazra: We have proper types support for consumers, from blade side we have to ensure that our components are compatible with both platform types. so here makeMotion time returns different type on different platform but the file is isomorphic (without any extension).

packages/blade/src/components/Typography/Heading/Heading.tsx
anuraghazra: Will do

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
anuraghazra: @chaitanyadeorukhkar yes that's actually weird it should not throw when doing build. Since i'm ignoring the web files on native compilation. I'm investigating this.

anuraghazra: Oh damn. Got it. It was failing because we had `SelectorInput.d.ts` which was reexport the web file. 

export * from './SelectorInput.web';

package.json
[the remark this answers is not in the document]
anuraghazra: Yes we should.

packages/blade/src/components/Button/IconButton/StyledIconButton.native.tsx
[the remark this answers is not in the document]
anuraghazra: @chaitanyadeorukhkar lets do `type.ts` everywhere. 

They are so buggy. 
https://github.com/razorpay/blade/pull/675#discussion_r964793026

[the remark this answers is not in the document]
divyanshu013: Yeah, have been hurt by `.d.ts` a few times 🥲

package.json
```diff
@@ -60,12 +60,12 @@
...
   },
   "husky": {
     "hooks": {
       "prepare-commit-msg": "exec < /dev/tty && git cz --hook || true",
-      "pre-commit": "lint-staged && yarn tsc:blade"
+      "pre-commit": "lint-staged"
```
chaitanyadeorukhkar, line 68: Lets add tsc for both native and web in pre-commit since the editor will only run tsc for one platform?

packages/blade/src/components/BladeProvider/emptyTheme.ts
```diff
@@ -1,3 +1,5 @@
+// TODO: test this better
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import type { Theme } from './';
 
 const emptyTheme: Theme = {
```
chaitanyadeorukhkar, line 5: I thought we are removing emptyTheme

packages/blade/src/tokens/global/motion.ts
```diff
@@ -34,56 +35,60 @@ type Delay = {
 };
 
 export type EasingFunctionFactory = { factory: () => (value: number) => number }; // similar to EasingFunctionFactory of `react-native-reanimated`
+type EasingType<Value extends string> = Platform.Select<{
```
chaitanyadeorukhkar, line 38: Nice!

packages/blade/src/utils/isReactNative/castUtils.ts
```diff
@@ -0,0 +1,11 @@
+import type { Platform } from './';
+
+const castWeb = <T>(value: T): Platform.CastWeb<T> => {
```
chaitanyadeorukhkar, line 3: Lets call this `castWebType` & `castNativeType` to be more explicit on what "casting" is happening here. Folks unaware of the context might get confused why certain things are wrapped around castWeb. 

Also can you add proper description her in comments for both `castWebType` & `castNativeType` so that if anyone visits these files will know what these functions do in detail? Its a fairly complex thing to understand with fresh eyes

packages/blade/src/utils/isReactNative/index.ts
```diff
@@ -0,0 +1,3 @@
+export { isReactNative } from './isReactNative';
+export { Platform } from './platform';
```
chaitanyadeorukhkar, line 2: Can you move all of these into `utils/platform.ts`? `isReactNative` doesn't seem like the right file to import these utils for Platform from

packages/blade/src/utils/isReactNative/platform.all.ts
```diff
@@ -0,0 +1,32 @@
...
+// bundler is not setup to handle this extension.
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-namespace */
+import type { Brand, IsBothFunction, NativeOrWebBrand } from './types';
+
+namespace Platform {
+  export type Name = 'web' | 'native';
+
+  // Unionify both web & native types, this will help us blade developers internally
```
chaitanyadeorukhkar, line 12: Can we be more descriptive here to explain what this does?

packages/blade/src/utils/makeMotionTime/makeMotionTime.all.ts
```diff
@@ -0,0 +1,5 @@
+import type { MakeMotionTime } from './types';
```
chaitanyadeorukhkar, line 1: Unsure about why we need `makeMotionTime.all`. Tried things without this and cast* functions seem to be returning things appropriately. Let's catch up for this?

chaitanyadeorukhkar (changes requested):

packages/blade/package.json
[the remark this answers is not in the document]
chaitanyadeorukhkar: I think this was missed out @anuraghazra

packages/blade/package.json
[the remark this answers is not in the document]
chaitanyadeorukhkar: Can we remove these @anuraghazra ?

packages/blade/src/utils/isReactNative/index.ts
anuraghazra: Done.

packages/blade/src/utils/isReactNative/castUtils.ts
anuraghazra: Done.

packages/blade/src/utils/isReactNative/platform.all.ts
anuraghazra: Done.

packages/blade/src/utils/makeMotionTime/makeMotionTime.all.ts
anuraghazra: It's not actually returning the correct types. 

Let's remove the .all file and then cast the type to native. 

image

You'll see that castNativeType is returning `never`, It's returning never since without the all type the module resolution will only resolve the `.web` type and the Platform.CastNative's `Extract<>` will fail. 

And this will happen vice versa, when your module resolution is set to `.native` if you try to do `castWebType` the type will return `never`

----

This is particularly bad also because `never` can be assignable to all the types and will bite us later, since it won't throw error:

```ts
const t: number = 'this is not a number' as never;
```

packages/blade/src/components/BladeProvider/emptyTheme.ts
anuraghazra: will remove it.

package.json
anuraghazra: Done!

packages/blade/package.json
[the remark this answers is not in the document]
anuraghazra: Done.

rfcs/writing-cross-platform-typescript.md
```diff
@@ -0,0 +1,134 @@
+# Writing Cross Platform TypeScript
```
divyanshu013, line 1: Let's link this in contributing guide too

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -423,7 +423,7 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
             <BaseInputVisuals leadingIcon={leadingIcon} prefix={prefix} isDisabled={isDisabled} />
             <StyledBaseInput
               id={inputId}
-              ref={ref}
+              ref={ref as never}
```
divyanshu013, line 426: Didn't understand why do we need `as never` here?

packages/blade/.storybook/react/main.js
```diff
@@ -44,6 +51,11 @@ module.exports = {
       }),
     ];
 
+    // const tsConfigIndex = config.plugins.findIndex(
+    //   (v) => v.constructor.name === 'ForkTsCheckerWebpackPlugin',
+    // );
+    // config.plugins.splice(tsConfigIndex, 1);
+
```
divyanshu013, line 58: Shall we remove these instead?
```suggestion
```

divyanshu013 (approved):
Looks good to me. Minor comments. Also we should add changelog for consumers to upgrade and docs if needed.

rfcs/writing-cross-platform-typescript.md
anuraghazra: Will do.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
anuraghazra: meant to do `as any`

packages/blade/.storybook/react/main.js
anuraghazra: Done.

rfcs/writing-cross-platform-typescript.md
anuraghazra: Done.

chaitanyadeorukhkar (approved):
Mostly LGTM 

@kamleshchandnani can you take a quick view?

divyanshu013:
Please pull in the latest master before merging so we run new TS on all the code

PR #681, opened 2022-09-01
chore: added textarea api doc

anuraghazra:
## [View Formatted](https://github.com/razorpay/blade/blob/anu/textarea-api/packages/blade/src/components/TextArea/_decisions/decisions.md)


packages/blade/src/components/TextArea/_decisions/decisions.md
```diff
@@ -0,0 +1,66 @@
...
+| neccessityIndicator  | `required, optional`             |         | Indicator to be shown next to the label for emphasizing the necessity of the field. if optional is provided then it'll show `(optional)` and if required is provided then it'll show `*`                                                                     |          |
+| showClearButton      | `boolean`                        | `true`  | Defines if we want to show the clear icon button on the right side of the password field                                                                                                                                                                     |          |
+| onClearButtonClicked | `() => {}`                       |         | Used in conjunction with showClearButton prop. If clicked it removes the contents of the TextArea. Users can also hook their own analytics code or do something else.                                                                                        |          |
```
divyanshu013, line 42: How about `onClickClearButon`?

packages/blade/src/components/TextArea/_decisions/decisions.md
```diff
@@ -0,0 +1,66 @@
...
+| errorText            | `string`                         |         | Shown when the validationState of the input field is set to Error.                                                                                                                                                                                           |          |
+| successText          | `string`                         |         | Shown when the validationState of the input field is set to Success                                                                                                                                                                                          |          |
+| autoFocus            | `boolean`                        | `false` | The autofocus global attribute is a Boolean attribute indicating that an element should be focused on page load. Web Reference, Native Reference                                                                                                             |          |
+
+
+## Open Questions: 
+
+- [ ] Should the `onClearButton` by default clear the TextArea contents? Or should users handle that logic?
```
divyanshu013, line 54: I think it should just clear the text, we shouldn't make that part configurable. But users should be able to hook in their callback to call additionally to the above.

packages/blade/src/components/TextArea/_decisions/decisions.md
```diff
@@ -0,0 +1,66 @@
...
+- [ ] Does the `maxCharacters` just a suggestion or a hard limit? Like if it's set to 20 can users go above 20, like Twitter?
```
divyanshu013, line 55: How are we doing this for other inputs?

packages/blade/src/components/TextArea/_decisions/decisions.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+## Open Questions: 
+
+- [ ] Should the `onClearButton` by default clear the TextArea contents? Or should users handle that logic?
+- [ ] Does the `maxCharacters` just a suggestion or a hard limit? Like if it's set to 20 can users go above 20, like Twitter? 
+- [ ] If the above is true, Should the textarea become invalid when `maxCharacters` count is exceeded? 
+- [ ] How will the `maxLines` behave? In design we set it to minimum of 2 and maximum of 5 lines, when we add `maxLines={4}` will it stay in `2` lines by default and then grow as needed based on the text content? or should it be just `4` lines. If it just stays in `4` lines no matter what then `numberOfLines` or `rows` (native html attribute like) would be a better name.
+- [ ] In design we have that small resize handler on the bottom right corner of the textarea what would happen when user tries to resize the textarea given that we also want to enforce `maxLines`
```
divyanshu013, line 58: Do we also need resize props? Allows resize horizontal / vertical / both?

packages/blade/src/components/TextArea/_decisions/decisions.md
anuraghazra: Seems kinda reversed 😅 
In BaseText input we also have the same name. OnClearButtonClicked.

anuraghazra: just `onClear` sounds okay to me too though. Bit shorter.

packages/blade/src/components/TextArea/_decisions/decisions.md
anuraghazra: Okay so if users do this: 

<TextArea onClearButtonClicked={() => { alert('cleared') }} />

This will clear the textarea and then show the alert. 

This?

packages/blade/src/components/TextArea/_decisions/decisions.md
anuraghazra: @saurav12 @kamleshchandnani ^

packages/blade/src/components/TextArea/_decisions/decisions.md
divyanshu013: Yes, that's how I would expect it to work.

packages/blade/src/components/TextArea/_decisions/decisions.md
saurav12: > Do we also need resize props? Allows resize horizontal / vertical / both?

@divyanshu013 @anuraghazra Yes, we need it so that users can customize the height & width. This helps with the overall accessibility.

packages/blade/src/components/TextArea/_decisions/decisions.md
divyanshu013: Oh, I think we should have the same convention in that case. Would favor `onClick...` because we follow present tense at some places like `onDismiss` in `Alert` and `onClick` in `Button`

divyanshu013: Or we need to think about how should we approach such props naming convention 🤔

packages/blade/src/components/TextArea/_decisions/decisions.md
```diff
@@ -0,0 +1,66 @@
...
+| onBlur               | `({ event, name, value }) => {}` |         | Function called when the input loses focus                                                                                                                                                                                                                   |          |
+| isDisabled           | `boolean`                        | `false` | Marks the input field as disabled. It'll also set aria-disabled to true and hence the input field will loose keyboard focus                                                                                                                                  |          |
+| isRequired           | `boolean`                        | `false` | Marks the input field as required. if the errorText is provided then that will be shown if the field is required and not filled, else ${label} is required                                                                                                   |          |
```
kamleshchandnani, line 39: do we really need to show the `${label} is required`? also don't think this is supported by our `FormHint` component yet

packages/blade/src/components/TextArea/_decisions/decisions.md
```diff
@@ -0,0 +1,66 @@
...
+| maxLines             | `number`                         | `2`     | Sets the number of lines (or rows) of the text area, minimum is `2` maximum is `5`                                                                                                                                                                           |          |
+| maxCharacters        | `number`                         |         | Used to show a character counter under the input field                                                                                                                                                                                                       |          |
+| validationState      | `error , success`                |         | Applies the style to the input field based on validation status and shows either errorText or successText respectively                                                                                                                                       |          |
```
kamleshchandnani, line 45: add `none` too

packages/blade/src/components/TextArea/_decisions/decisions.md
kamleshchandnani: what I've done for textinput is
1. if it is uncontrolled then clear it
2. if it is controlled then do nothing and let the users clear it since they will be managing the state in their application and if we clear and they don't then there will be discrepancy in their state but visually the input will be empty

packages/blade/src/components/TextArea/_decisions/decisions.md
kamleshchandnani: it's a hard limit unlike twitter. it basically sets maxLength on the input field natively

packages/blade/src/components/TextArea/_decisions/decisions.md
kamleshchandnani: we discussed to descope this for now because of trailing visual elements and it's visual issues

packages/blade/src/components/TextArea/_decisions/decisions.md
```diff
@@ -0,0 +1,66 @@
...
+| autoFocus            | `boolean`                        | `false` | The autofocus global attribute is a Boolean attribute indicating that an element should be focused on page load. Web Reference, Native Reference                                                                                                             |          |
+
+
+## Open Questions: 
+
+- [ ] Should the `onClearButton` by default clear the TextArea contents? Or should users handle that logic?
+- [ ] Does the `maxCharacters` just a suggestion or a hard limit? Like if it's set to 20 can users go above 20, like Twitter? 
+- [ ] If the above is true, Should the textarea become invalid when `maxCharacters` count is exceeded? 
+- [ ] How will the `maxLines` behave? In design we set it to minimum of 2 and maximum of 5 lines, when we add `maxLines={4}` will it stay in `2` lines by default and then grow as needed based on the text content? or should it be just `4` lines. If it just stays in `4` lines no matter what then `numberOfLines` or `rows` (native html attribute like) would be a better name.
```
kamleshchandnani, line 57: don't grow whatever the lines are render them exactly

packages/blade/src/components/TextArea/_decisions/decisions.md
anuraghazra: Sorry this line is a bit confusing I guess. What i meant was that the input will be set to `aria-required/required` prop. 
Visually we handle the required prop via `neccessityIndicator`

packages/blade/src/components/TextArea/_decisions/decisions.md
```diff
@@ -0,0 +1,70 @@
...
+```tsx
+<TextArea
+  label="Enter Message"
+  labelPosition="left"
+  placeholder="Enter your message"
+  showClearButton
+  helpText="Send message to author"
+  errorText="Message cannot be blank"
+  successText="Message sent"
+  validationState={textareaValue.length < 10 ? 'error' : 'success'}
```
avinashdvv, line 24: question: why this is not boolean instead of `string: 'error' | 'success'`?

anuraghazra: It might look like we have two states in first glance but we actually have 3 states. 

IDLE: 

image

ERROR: 

image

SUCCESS: 

image


So if we make the validationState into a boolean, then we will have to rely on the `undefined` value or the non-existence of the prop to map to the IDLE state. Which isn't very scalable. 

```ts
<Input validationState={isInvalid ? true : isValid ? true : undefiend} /> // <--- Weird ternaries
```

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/TextArea/_decisions/decisions.md
kamleshchandnani: don't mind changing in BaseInput to `OnClearButtonClick`. it didn't click me at that time 😅 . `onClickClearButton` should also work.

kamleshchandnani (approved):
LGTM

packages/blade/src/components/TextArea/_decisions/decisions.md
snehasish27, line 24: can we have a the warning state also for text area ?

packages/blade/src/components/TextArea/_decisions/decisions.md
```diff
@@ -0,0 +1,70 @@
...
+| showClearButton      | `boolean`                        | `true`  | Defines if we want to show the clear icon button on the right side of the password field                                                                                                                                                                     |          |
+| onClearButtonClicked | `() => {}`                       |         | Used in conjunction with showClearButton prop. If clicked it removes the contents of the TextArea. Users can also hook their own analytics code or do something else.                                                                                        |          |
+| maxLines             | `number`                         | `2`     | Sets the number of lines (or rows) of the text area, minimum is `2` maximum is `5`                                                                                                                                                                           |          |
```
snehasish27, line 43: for max line can we not set a upper limit ?

and can have a extra param call `autoGrow` which will automatically grow the text area to the `maxLines` value which consumer will specify

and if `autoGrow` is false we can honour the `maxLines` prop and stick the text area to that no of row only.

packages/blade/src/components/TextArea/_decisions/decisions.md
```diff
@@ -0,0 +1,70 @@
...
+| maxCharacters        | `number`                         |         | Used to show a character counter under the input field                                                                                                                                                                                                       |          |
+| validationState      | `error , success, none`          |         | Applies the style to the input field based on validation status and shows either errorText or successText respectively                                                                                                                                       |          |
```
snehasish27, line 45: can we have the warning state also

snehasish27 (commented):
some suggestion

packages/blade/src/components/TextArea/_decisions/decisions.md
anuraghazra: Design doesn't have warning state for any inputs. 
https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=11115%3A166743

packages/blade/src/components/TextArea/_decisions/decisions.md
anuraghazra: Good idea! Yes we could add that but right now we've scoped this feature, until requirement/usecase arises. 

Check the Open Questions at the end.

packages/blade/src/components/TextArea/_decisions/decisions.md
anuraghazra: https://github.com/razorpay/blade/pull/681#discussion_r971835438

PR #685, opened 2022-09-01
feat: add `Spinner` component

chaitanyadeorukhkar:
Adds `Spinner` & `BaseSpinner` components


packages/blade/src/components/Spinner/BaseSpinner/BaseSpinner.tsx
```diff
@@ -0,0 +1,83 @@
...
+   * Sets the aria-label for web & accessibilityLabel react-native.
+   *
+   */
+  accessibilityLabel: string;
+  /**
+   * Indicates that an element will be updated, and describes the types of updates the user agents, assistive technologies, and user can expect from the live region.
+   *
+   *  @default 'assertive'
+   */
+  accessibilityLiveRegion?: 'off' | 'assertive' | 'polite';
```
anuraghazra, line 34: Let's get derive from the makeAccessible types: 

accessibilityLiveRegion: AccessibilityProps['liveRegion']

packages/blade/src/components/Spinner/BaseSpinner/BaseSpinner.tsx
```diff
@@ -0,0 +1,83 @@
...
+  accessibilityLiveRegion = 'assertive',
+}: BaseSpinnerProps): React.ReactElement => {
+  const { theme } = useTheme();
+  return (
+    <SpinningBox
+      display="flex"
+      {...makeAccessible({
+        label: accessibilityLabel,
+        liveRegion: accessibilityLiveRegion,
+        atomic: true,
```
anuraghazra, line 72: Don't think we need atomic here do we?

packages/blade/src/components/Spinner/BaseSpinner/SpinningBox.web.tsx
```diff
@@ -0,0 +1,23 @@
...
+  }
+  to {
+    transform: rotate(360deg);
+  }
+`;
+
+const SpinningBox = styled(Box)(({ theme }) => {
+  return css`
+    width: max-content;
+    animation: ${rotate} ${makeMotionTime(getIn(theme.motion, motion.duration))}
```
anuraghazra, line 18: Same here. 

getIn(theme.motion, theme.motion)?

packages/blade/src/components/Spinner/BaseSpinner/SpinningBox.native.tsx
```diff
@@ -0,0 +1,52 @@
...
+} from 'react-native-reanimated';
+import { motion } from './spinnerTokens';
+import { useTheme } from '~components/BladeProvider';
+import { getIn, makeMotionTime } from '~utils';
+import Box from '~components/Box';
+
+const SpinningBox = ({ children }: { children: React.ReactNode }): React.ReactElement => {
+  const { theme } = useTheme();
+  const duration = (makeMotionTime(getIn(theme.motion, motion.duration)) as unknown) as number;
+  const easing = getIn(theme.motion, motion.easing) as EasingFunctionFactory;
```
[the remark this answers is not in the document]
anuraghazra: Oh my bad it's `getIn(theme.motion, motion.easing)` and the `motion.easing` is the spinner tokens. Kinda hard to figure out at a glance. Can we call it maybe `spinnerMotion`?

packages/blade/src/components/Spinner/BaseSpinner/BaseSpinner.tsx
chaitanyadeorukhkar: We do, otherwise screenreaders will go inside the wrapper & icon. With atomic, it considers the whole spinner as one item

.changeset/wet-cows-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": minor
+---
+
+feat: add `Spinner` and `BaseSpinner` components
```
divyanshu013, line 5: I think we should specify internal components because these go in changelog

```suggestion
feat: add `Spinner` component
- Also adds an internal `BaseSpinner` component
```

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
```diff
@@ -120,10 +120,10 @@ const buttonSizeToSpinnerSizeMap: Record<
   NonNullable<BaseButtonProps['size']>,
   SpinnerProps['size']
 > = {
-  xsmall: 'medium',
+  xsmall: 'small',
   small: 'medium',
-  medium: 'large',
-  large: 'xlarge',
+  medium: 'medium',
+  large: 'large',
```
divyanshu013, line 126: Why are these changing 🤔

packages/blade/src/components/Spinner/BaseSpinner/BaseSpinner.tsx
```diff
@@ -0,0 +1,84 @@
...
+  size?: 'small' | 'medium' | 'large';
+  /**
+   * Sets the aria-label for web & accessibilityLabel react-native.
+   *
+   */
+  accessibilityLabel: string;
+  /**
+   * Indicates that an element will be updated, and describes the types of updates the user agents, assistive technologies, and user can expect from the live region.
+   *
+   *  @default 'assertive'
```
divyanshu013, line 33: For my curiosity, why are we going ahead with assertive and not polite?

packages/blade/src/components/Spinner/BaseSpinner/SpinnerIcon.tsx
```diff
@@ -0,0 +1,28 @@
+import { Svg, Path } from '~components/Icons/_Svg';
+
+type SpinnerIconProps = {
+  dimensions: string;
+  color: string;
+};
+
+const SpinnerIcon = ({ dimensions, color }: SpinnerIconProps): React.ReactElement => {
```
divyanshu013, line 8: Don't we want to add this in `icons/`?

packages/blade/src/components/Spinner/BaseSpinner/SpinningBox.native.tsx
```diff
@@ -0,0 +1,54 @@
...
+      withTiming(360, {
+        duration,
+        easing,
+      }),
+      -1,
+    );
+    return (): void => {
+      cancelAnimation(rotation);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
divyanshu013, line 44: Can cause potential stale closure issues

packages/blade/src/components/Spinner/BaseSpinner/SpinningBox.web.tsx
```diff
@@ -0,0 +1,23 @@
...
+const rotate = keyframes`
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+`;
+
+const SpinningBox = styled(Box)(({ theme }) => {
```
divyanshu013, line 15: I think this should be inline element, `inline-block` or `inline-flex` to help with styling other nearby content, thoughts?

packages/blade/src/components/Spinner/BaseSpinner/SpinnerIcon.tsx
anuraghazra: imo let's keep it here. Specific to spinner only. No need for consumers to expose this.

packages/blade/src/components/Button/BaseButton/buttonTokens.ts
chaitanyadeorukhkar: Earlier we made up these sizes as placeholders until spinner was finalized on design. These are the new correct size mappings

packages/blade/src/components/Spinner/BaseSpinner/BaseSpinner.tsx
chaitanyadeorukhkar: Removed this now

packages/blade/src/components/Spinner/BaseSpinner/SpinnerIcon.tsx
chaitanyadeorukhkar: Yeah. Don't see a use-case for consumers to be using this icon standalone

packages/blade/src/components/Spinner/BaseSpinner/SpinningBox.web.tsx
chaitanyadeorukhkar: Yeah makes sense. Made it an `inline-flex`

packages/blade/src/components/Spinner/BaseSpinner/SpinningBox.native.tsx
chaitanyadeorukhkar: Hmmm. What else could we do here? We want it to be triggered on mount and cancel the animation on unmount

divyanshu013 (approved):
Looks good to me. One [minor comment](https://github.com/razorpay/blade/pull/685#discussion_r963546992).

packages/blade/src/components/Spinner/BaseSpinner/SpinningBox.native.tsx
divyanshu013: Either using refs inside or passing the variables in dependency array, whichever would make more sense and looks more readable.

chaitanyadeorukhkar: Added

packages/blade/src/components/Spinner/BaseSpinner/BaseSpinner.tsx
anuraghazra: Conclusion: Fixed a11y issues with role:progressbar

packages/blade/src/components/Spinner/BaseSpinner/BaseSpinner.tsx
```diff
@@ -6,9 +6,22 @@ import type { Theme } from '~components/BladeProvider';
...
+import { Text } from '~components/Typography';
 
 type BaseSpinnerProps = {
   intent?: Feedback;
+  /**
+   * Sets the label of the spinner.
+   *
+   * @default 'right'
+   */
+  label?: string;
```
divyanshu013, line 19: Now that we've both label and a11y label. What if user provides different values for both 🤔

packages/blade/src/components/Spinner/BaseSpinner/__tests__/BaseSpinner.native.test.tsx
```diff
@@ -34,6 +34,30 @@ describe('<BaseSpinner />', () => {
...
+  it('should render low contrast BaseSpinner with right label', () => {
+    const { toJSON } = renderWithTheme(
+      <BaseSpinner
+        accessibilityLabel="Loading"
+        contrast="low"
+        label="Loading"
+        labelPosition="right"
+      />,
+    );
+    expect(toJSON()).toMatchSnapshot();
```
divyanshu013, line 46: Suggestion: it might be better to assert for the presence of the labels explicitly instead of snapshot for these tests.

packages/blade/src/components/Spinner/Spinner/Spinner.tsx
```diff
@@ -15,19 +26,29 @@ type SpinnerProps = {
    */
   size?: 'small' | 'medium' | 'large';
   /**
-   * Sets the aria-label for web & accessibilityLabel react-native.
+   * Sets the aria-label for web & accessibilityLabel react-native. Uses `label` as its value if it exists otherwise falls back to "Loading"
```
divyanshu013, line 29:
```suggestion
   * Sets the aria-label for web & accessibilityLabel for react-native. Uses `label` as its value if it exists otherwise falls back to "Loading"
```

divyanshu013 (commented):

packages/blade/src/components/Spinner/BaseSpinner/BaseSpinner.tsx
chaitanyadeorukhkar: That's a valid case. The way this works is, we set a11y label's value as the prop a11yLabel, if it doesn't exist then we set it as the value of the prop label else we fall back to "Loading"

packages/blade/src/components/Spinner/BaseSpinner/__tests__/BaseSpinner.native.test.tsx
chaitanyadeorukhkar: I prefer snapshot here since it will also store the exact styles and for the label. If someone changes the prop of Text for the label from `subdued` to `normal`, we wouldn't be able to catch it with our tests unless we have a snapshot of it

divyanshu013 (approved):
:shipit:

PR #695, opened 2022-09-06
tests(BaseInput): add web tests and fix onBlur

divyanshu013:
- Lot of diff from snapshot
- Adds tests for main functionalities
- Fixes `onBlur`, it wasn't being passed


packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -363,9 +363,10 @@ export const BaseInput = ({
   keyboardReturnKeyType,
   inputMode,
   autoCompleteSuggestionType,
+  onBlur,
```
kamleshchandnani, line 366: good catch, hence tests are good 😞

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should render success validation state', () => {
+    const { container } = renderWithTheme(
+      <BaseInput label="Enter name" id="name" validationState="success" successText="Success" />,
+    );
+
+    expect(container).toMatchSnapshot();
```
kamleshchandnani, line 21: shouldn't we also be checking that the success text is being displayed? snapshot will help us account for any changes in the layout but whether it's rendering in the first place or not won't be captured using snapshot right?

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+    const { container } = renderWithTheme(
+      <BaseInput
+        label="Enter name"
+        placeholder="First Last"
+        id="name"
+        leadingIcon={EyeIcon}
+        trailingIcon={CloseIcon}
+      />,
+    );
+
```
kamleshchandnani, line 42: shall we check if the icons really rendered or not?

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+    const label = 'Enter name';
+    const valueInitial = 'Divyanshu';
+    const valueFinal = 'Divyanshu Maithani';
+
+    const { getByLabelText, rerender } = renderWithTheme(
+      <BaseInput label={label} id="name" value={valueInitial} />,
+    );
+
+    const input = getByLabelText(label);
+    expect(input).toHaveValue(valueInitial);
```
kamleshchandnani, line 132: nit: since input is not being used anywhere else

```suggestion
    expect(getByLabelText(label)).toHaveValue(valueInitial);
```

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+        label="Enter name"
+        id="name"
+        isRequired
+        helpText="First name and last name"
+        validationState="none"
+      />,
+    );
+
+    const input = getByRole('textbox');
+    await assertAccessible(input);
```
kamleshchandnani, line 155: what all does it asserts? right now just seems like a blackbox.

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+  });
```
kamleshchandnani, line 156: we shall also add tests for 
* error text since the input becomes invalid
* same for success text
and check that the input is marked accessible along with necessary attributes which associates errorText and successText with the input

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: Actually such tests would be much better over snapshots, will swap this out. Personally snapshots are very hard to maintain and review.

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: Intentionally did it this way to improve readability 😸

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: It runs `jest-axe` suite. Was thinking we could add a11y specific tests separately.

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: I realized we don't pass any a11y labels for these icons

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: Done in https://github.com/razorpay/blade/pull/695/commits/1b30d066d0edf03c62601f0c8a2dbb64fe5b7b31

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: Added some explicit tests in https://github.com/razorpay/blade/pull/695/commits/1b30d066d0edf03c62601f0c8a2dbb64fe5b7b31

Doesn't include all possible combinations (too many possibilities)

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: Done in https://github.com/razorpay/blade/pull/695/commits/1b30d066d0edf03c62601f0c8a2dbb64fe5b7b31

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
anuraghazra: @kamleshchandnani it checks these rules. 

https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md

I guess it's okay to test a11y specific things separately too. Mostly react-testing-library helps with making tests generally be written with a11y in mind.

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
anuraghazra: Should be fine actually they are mostly for visuals and doesn't have any functionality nor can they take focus.

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+    );
+
+    const input = getByLabelText(label);
+    expect(input).toHaveValue(valueInitial);
+
+    await user.type(input, ' Maithani');
+    expect(input).toHaveValue(valueFinal);
+  });
+
+  it('should set value as a controlled input', () => {
```
anuraghazra, line 134: Can we modify this test a bit? 
It's better to test controlled component as how the user might use it in real-world. 

```ts

  it('should set value as a controlled input', async () => {
    const user = userEvent.setup();
    const label = 'Enter name';
    const valueInitial = 'Divyanshu';
    const valueFinal = 'Divyanshu Maithani';

    const Example = (): React.ReactElement => {
      const [value, setValue] = React.useState(valueInitial);

      return (
        <>
          <BaseInput
            label={label}
            id="name"
            value={value}
            onChange={({ value }) => setValue(value!)}
          />
          <p data-testid="input-value">{value}</p>
        </>
      );
    };

    const { getByLabelText, getByTestId } = renderWithTheme(<Example />);

    const input = getByLabelText(label);
    expect(input).toHaveValue(valueInitial);
    await user.type(input, ' Maithani');

    expect(getByTestId('input-value')).toHaveTextContent(valueFinal);
  });
```

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+        name="name"
+        defaultValue={userName}
+        // eslint-disable-next-line jsx-a11y/no-autofocus
+        autoFocus
+        onBlur={onBlur}
+      />,
+    );
+
+    // shifts user focus and therefore blurs the focussed input
+    await user.tab();
```
anuraghazra, line 112: This might seem unnecessary but it's better to remove autofocus and use the input how the user might use it. first focus it then blur it manually. 

Reason being autofocus even works when the component is set to `tabIndex={-1}` which removes the elements taking focus from keyboard input but still accepts programatic focus (e.focus/autofocus attr)

```ts
    // focus
    await user.tab();
    // shifts user focus and therefore blurs the focussed input
    // blur
    await user.tab();
```

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+    expect(successText).toBeTruthy();
+    expect(input).toHaveAccessibleDescription('Success');
+    expect(input).toBeValid();
+  });
+
+  it('should display error validation state', () => {
+    const label = 'Enter name';
+    const { getByText, getByLabelText } = renderWithTheme(
+      <BaseInput label={label} id="name" validationState="error" errorText="Error" />,
```
anuraghazra, line 33: In all 3 tests testing validationState let's add all the props. helpText/successText/errorText, this will ensure that our component isn't picking some other text when validation state is set.

anuraghazra (approved):
LGTM

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: I think this would be unrelated to current test (since it's for `onBlur`), though maybe we can add another test for what you're mentioning.

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
anuraghazra: We are testing if the value is actually controlled or not. 

If BaseInput for some reason stops changing the onChange value or doesn't take `value` prop as initial or we change some logic inside the BaseInput we will know since `input-value` won't be as we expect then.

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: Added in https://github.com/razorpay/blade/pull/695/commits/8be6b8367acaeaf5f157a5ba38eb97ff0c75f785

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: Added another case in https://github.com/razorpay/blade/pull/695/commits/8be6b8367acaeaf5f157a5ba38eb97ff0c75f785

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.web.test.tsx
divyanshu013: Done in https://github.com/razorpay/blade/pull/695/commits/8be6b8367acaeaf5f157a5ba38eb97ff0c75f785

PR #700, opened 2022-09-08
fix: TextInput imports

kamleshchandnani:
Issue: aliased imports from `~components/Form` breaks the type generation

Workaround(not ideal but unblocking release for now): replace alias imports with relative imports `~components/Form` ➡️ `../../Form`


chaitanyadeorukhkar (approved):

divyanshu013 (approved):
Let's add some comment about the issue perhaps

kamleshchandnani:
add comments where? in the source file?

Issue #668, opened 2022-08-25
Bug: fix occurrences of a11y roles that don't exist on native


divyanshu013:
We've some occurrences where we're using the same roles on web and native but not all web roles exist on native such as `group`. To replicate run `yarn test:react-native`

<img width="1617" alt="image" src="https://user-images.githubusercontent.com/6682655/186612686-216b6a83-3436-4e0a-a124-fa6ab023363e.png">

jainpawan21:
Hi @divyanshu013, Is this issue available to pick?

snitin315:
Fixed in master, I can see only two such logs from the makeAccessible test, which is intentional & expected.

https://github.com/razorpay/blade/blob/8d2105a86eb02ab0b79dd652d565af83c50da9b5/packages/blade/src/utils/makeAccessible/__tests__/makeAccessible.native.test.ts#L39-L44

https://github.com/razorpay/blade/blob/8d2105a86eb02ab0b79dd652d565af83c50da9b5/packages/blade/src/utils/makeAccessible/__tests__/makeAccessible.native.test.ts#L56

<img width="988" alt="Screenshot 2023-10-12 at 10 53 15 AM" src="https://github.com/razorpay/blade/assets/46647141/ff399b52-47c2-494e-a095-6cac1cc36f9c">

PR #702, opened 2022-09-08
feat(blade): added TextArea component

anuraghazra:
- API Decisions #681

### DIFF
--- a/packages/blade/src/components/Form/FormLabel.tsx
+++ b/packages/blade/src/components/Form/FormLabel.tsx
@@ -49,6 +49,8 @@ const FormLabel = ({
     );
   }
 
+  const isLabelLeftPositioned = position === 'left' && isDesktop;
+
   if (necessityIndicator === 'optional') {
     necessityLabel = (
       <Text variant="caption" weight="regular" type="placeholder">
@@ -89,7 +91,7 @@ const FormLabel = ({
       <Text
         type="subdued"
         variant="body"
-        size={position === 'left' ? 'medium' : 'small'}
+        size={isLabelLeftPositioned ? 'medium' : 'small'}
         weight="bold"
       >
         {children}
@@ -110,7 +112,6 @@ const FormLabel = ({
   }
 
   const Component = as;
-  const isLabelLeftPositioned = position === 'left';
   // only set 120px label when device is desktop
   const width = isLabelLeftPositioned && isDesktop ? '120px' : 'auto';
 
--- a/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
+++ b/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
@@ -401,9 +401,10 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
     }
 
     const isTextArea = as === 'textarea';
+    const isReactNative = getPlatformType() === 'react-native';
 
     return (
-      <>
+      <Box>
         <Box
           display="flex"
           flexDirection={isLabelLeftPositioned ? 'row' : 'column'}
@@ -438,7 +439,7 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
           >
             <BaseInputVisuals leadingIcon={leadingIcon} prefix={prefix} isDisabled={isDisabled} />
             <StyledBaseInput
-              as={as}
+              as={isReactNative ? undefined : as}
               id={inputId}
               ref={ref}
               name={name}
@@ -496,7 +497,7 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
             {trailingFooterSlot?.(inputValue)}
           </Box>
         </Box>
-      </>
+      </Box>
     );
   },
 );
--- a/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
+++ b/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
@@ -21,6 +21,10 @@ import useInteraction from '~src/hooks/useInteraction';
 export type BaseInputProps = FormInputLabelProps &
   FormInputValidationProps & {
     /**
+     * Determines if it needs to be rendered as input or textarea
+     */
+    as?: 'input' | 'textarea';
+    /**
      * ID that will be used for accessibility
      */
     id: string;
@@ -158,6 +162,10 @@ export type BaseInputProps = FormInputLabelProps &
      * Element to be rendered on the trailing slot of input field footer
      */
     trailingFooterSlot?: (value?: string) => ReactNode;
+    /**
+     * Sets the textarea's number of lines
+     */
+    numberOfLines?: 2 | 3 | 4 | 5;
   };
 
 const autoCompleteSuggestionTypeValues = [
@@ -316,6 +324,7 @@ const getDescribedByElementId = ({
 export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
   (
     {
+      as = 'input',
       label,
       labelPosition = 'top',
       placeholder,
@@ -346,6 +355,7 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
       autoCompleteSuggestionType,
       trailingHeaderSlot,
       trailingFooterSlot,
+      numberOfLines = 2,
     },
     ref,
   ) => {
@@ -390,6 +400,8 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
       );
     }
 
+    const isTextArea = as === 'textarea';
+
     return (
       <>
         <Box
@@ -403,6 +415,9 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
             display="flex"
             flexDirection={isLabelLeftPositioned ? 'column' : 'row'}
             justifyContent="space-between"
+            alignSelf="flex-start"
+            marginTop={isTextArea && isLabelLeftPositioned ? 'spacing.3' : 'spacing.0'}
+            marginBottom={isTextArea && isLabelLeftPositioned ? 'spacing.3' : 'spacing.0'}
           >
             <FormLabel
               as="label"
@@ -415,6 +430,7 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
             {trailingHeaderSlot?.(inputValue)}
           </Box>
           <BaseInputWrapper
+            isTextArea={isTextArea}
             isDisabled={isDisabled}
             validationState={validationState}
             currentInteraction={currentInteraction}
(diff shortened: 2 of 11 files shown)


packages/blade/src/components/Form/FormLabel.tsx
```diff
@@ -49,6 +49,8 @@ const FormLabel = ({
     );
   }
 
+  const isLabelLeftPositioned = position === 'left' && isDesktop;
```
anuraghazra, line 52: Fixes a bug where if you make the window smaller the label will go to the top but the size of the label will still remain like the desktop version. 

image

packages/blade/src/components/Input/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,158 @@
...
+        onChange?.({ name, value });
+      }}
+      trailingFooterSlot={(value) => {
+        return maxCharacters ? (
+          <Box marginTop="spacing.2" marginRight="spacing.1">
+            <CharacterCounter currentCount={value?.length ?? 0} maxCount={maxCharacters} />
+          </Box>
+        ) : null;
+      }}
+      id=""
```
anuraghazra, line 153: @kamleshchandnani why is this ID mandetory?

packages/blade/src/components/Input/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,158 @@
...
+  showClearButton,
+  onClearButtonClick,
+  autoFocus,
+  numberOfLines = 2,
+}: TextAreaProps): React.ReactElement => {
+  const inputRef = React.useRef<HTMLInputElement | TextInputReactNative>(null);
+  const [shouldShowClearButton, setShouldShowClearButton] = React.useState(false);
+
+  React.useEffect(() => {
+    if (showClearButton && (value?.length || defaultValue?.length)) {
```
kamleshchandnani, line 75: no need of effect now since you can add it as default value?

packages/blade/src/components/Input/TextArea/TextArea.tsx
kamleshchandnani: to generate all the a11y ids specific to that field else everything will be default to default value of `base-input-***` since it's internal component let's be more specific and restrictive about what kind of ids will be generated else it'll be confusing to debug for consumers

packages/blade/src/components/Input/TextArea/__tests__/TestArea.web.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+      />,
+    );
+
+    // shifts user focus and therefore blurs the focussed input
+    await user.tab();
+    expect(onBlur).toHaveBeenCalledTimes(1);
+    expect(onBlur).toHaveBeenCalledWith({ name: 'name', value: userName });
+  });
+
+  it('should be focussable', async () => {
```
kamleshchandnani, line 184:
```suggestion
  it('should be focusable', async () => {
```

packages/blade/src/components/Input/TextArea/__tests__/TestArea.web.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+      );
+    };
+
+    const { getByLabelText, getByTestId } = renderWithTheme(<ControlledInputExample />);
+
+    const input = getByLabelText(label);
+    expect(input).toHaveValue(valueInitial);
+
+    await user.type(input, ' Maithani');
+    expect(getByTestId('input-value')).toHaveTextContent(valueFinal);
```
kamleshchandnani, line 240: why do we need to check the value of paragraph? we shall assert it against the actual textarea right?

packages/blade/src/components/Input/TextArea/__tests__/TextArea.native.test.tsx
```diff
@@ -0,0 +1,130 @@
+import { fireEvent } from '@testing-library/react-native';
+import type { ReactElement } from 'react';
+import React from 'react';
+
+import { TextArea } from '..';
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
+
```
kamleshchandnani, line 7: parity this with `TextInput.native.test`

packages/blade/src/components/Input/TextArea/__tests__/TextArea.native.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const input = getByPlaceholderText(placeholder);
+    expect(input).toBeDisabled();
+  });
+
+  it('should handle onChange', () => {
+    const placeholder = 'First Last';
+    const onChange = jest.fn();
+    const userName = 'Divyanshu';
+
+    const { getByPlaceholderText } = renderWithTheme(
```
kamleshchandnani, line 76: can we update the label on react native to have a11y prop so we can do `getByLabel`?

packages/blade/src/components/Input/TextInput/TextInput.stories.tsx
```diff
@@ -275,3 +276,81 @@ const TextInputControlledTemplate: ComponentStory<typeof TextInputComponent> = (
   );
 };
 export const TextInputControlled = TextInputControlledTemplate.bind({});
+
+const TextInputKitchenSinkTemplate: ComponentStory<typeof TextInputComponent> = () => {
```
kamleshchandnani, line 280: thanks ❤️

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -416,13 +432,15 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
...
           <BaseInputWrapper
+            isTextArea={isTextArea}
             isDisabled={isDisabled}
             validationState={validationState}
             currentInteraction={currentInteraction}
             isLabelLeftPositioned={isLabelLeftPositioned}
           >
             <BaseInputVisuals leadingIcon={leadingIcon} prefix={prefix} isDisabled={isDisabled} />
             <StyledBaseInput
+              as={isReactNative ? undefined : as}
```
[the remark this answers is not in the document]
anuraghazra: Will have to check `browser` and `node` both then. I think it's fine. 

Maybe later on we could add 1 more option to the `getPlatformType` to return `web` or `react-native`

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -85,7 +86,10 @@ const StyledNativeBaseInput = styled.TextInput<
     trailingIcon: props.trailingIcon,
   }),
   lineHeight: undefined,
-  height: '36px',
+  textAlignVertical: 'top',
```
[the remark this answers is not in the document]
anuraghazra: No it shouldn't. I'll check again tho. 

https://reactnative.dev/docs/textinput#multiline
image

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -113,6 +119,8 @@ export const StyledBaseInput = React.forwardRef<TextInput, StyledBaseInputProps>
         // the types of styled-components for react-native is creating a mess, so there's no other option but to type `ref` as any
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         ref={ref as any}
+        multiline={Boolean(numberOfLines)}
```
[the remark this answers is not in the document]
anuraghazra: Works!

packages/blade/src/components/Input/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,158 @@
...
+  showClearButton?: boolean;
+  /**
+   * Event handler to handle the onClick event for clear button.
+   */
+  onClearButtonClick?: () => void;
+};
+
+// need to do this to tell TS to infer type as TextInput of React Native and make it believe that `ref.current.clear()` exists
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const isReactNative = (_textInputRef: any): _textInputRef is TextInputReactNative => {
```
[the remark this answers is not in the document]
anuraghazra: We could, but let's do this next time. Once we get one more usecase we'll refactor it, shouldn’t be high effort.

packages/blade/src/components/Input/TextArea/__tests__/TextArea.native.test.tsx
anuraghazra: Need to fix this #668

packages/blade/src/components/Input/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,158 @@
...
+  placeholder,
+  value,
+  maxCharacters,
+  showClearButton,
+  onClearButtonClick,
+  autoFocus,
+  numberOfLines = 2,
+}: TextAreaProps): React.ReactElement => {
+  const inputRef = React.useRef<HTMLInputElement | TextInputReactNative>(null);
+  const [shouldShowClearButton, setShouldShowClearButton] = React.useState(false);
```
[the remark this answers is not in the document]
anuraghazra: Actually this will work. 
showClearButton && (value?.length || defaultValue?.length)

packages/blade/src/components/Input/TextArea/__tests__/TestArea.web.test.tsx
anuraghazra: Changed.

packages/blade/src/components/Input/TextArea/__tests__/TextArea.native.test.tsx
anuraghazra: Done.

packages/blade/src/components/Input/TextArea/__tests__/TestArea.web.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+    );
+
+    const input = getByLabelText(label);
+    const errorText = getByText('Error');
+
+    expect(errorText).toBeTruthy();
+    expect(input).toHaveAccessibleDescription('Error');
+    expect(input).toBeInvalid();
+  });
+
```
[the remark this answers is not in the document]
anuraghazra: Done

packages/blade/src/components/Input/BaseInput/baseInputStyles.ts
```diff
@@ -94,6 +94,8 @@ export const getBaseInputStyles = ({
     trailingIcon,
   });
 
+  const isReactNative = getPlatformType() === 'react-native';
```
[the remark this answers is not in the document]
anuraghazra: Then need to check for both browser & node. Let's keep it react-native only for now.

packages/blade/src/components/Form/FormLabel.tsx
```diff
@@ -58,11 +58,7 @@ const FormLabel = ({
 
   let necessityLabel: React.ReactNode = null;
 
-  if (!isDesktop && position === 'left') {
```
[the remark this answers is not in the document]
anuraghazra: Don't think it would be very useful since this warning is based on responsiveness.. So if you make the window smaller then also would throw error. Which may or may not be what they expect to happen.

divyanshu013 (approved):
:shipit:

PR #713, opened 2022-09-19
docs: add OTPInputField decisions

chaitanyadeorukhkar:
### [View Formatted File](https://github.com/razorpay/blade/blob/docs/otp-input-decisions/packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md)


packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
```diff
@@ -0,0 +1,92 @@
...
+  errorText="Invalid OTP"
+  successText="Valid OTP"
+  validationState={inputValue.length < 4 ? 'error' : 'success'}
+/>
+```
+
+
+| Prop | Type | Required | Default | Description |
+|---|---|---|---|---|
+| otpLength | `4`, `6`  | No | `4` | Determines the length of the OTP input. |
```
divyanshu013, line 35: How can we be certain this would always be 4 or 6. This depends on backend service right?

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
```diff
@@ -0,0 +1,92 @@
...
+| label | `string` | Yes |  | Determines the label of the input field. Also used as `aria-label` |
+| labelPosition | `top`, `left` | No | `top` | Used to identify if the label of the input field will be placed on the top of the input field or left. `left` is only available for large screen devices. When position is `left` then the width of the label is fixed to `120px` and the content is aligned left. |
+| necessityIndicator | `optional`, `required` | No |  | Indicator to be shown next to the label for emphasizing the necessity of the field. if `optional` is provided then it'll show `${label} (optional)` and if `required` is provided then it'll show `${label} *` |
```
divyanshu013, line 38: Similar to `PasswordField` can an OTP ever be optional?

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
```diff
@@ -0,0 +1,92 @@
...
+| onBlur | `({ name, value }) => {}` | No |  | Function called when the input field loses focus |
+| isDisabled | `boolean` | No | `false` | Marks the input field as disabled. It'll also set `aria-disabled` to true and hence the input field will loose keyboard focus. |
+| isRequired | `boolean` | No | `false` | Marks the input field as required. if the errorText is provided then that will be shown if the field is required and not filled, else `${label} is required` |
+| autoFocus | `boolean` | No | `false` | The autofocus global attribute is a Boolean attribute indicating that an element should be focused on page load. [Web Reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus), [Native Reference](https://reactnative.dev/docs/textinput#autofocus) |
+| keyboardReturnKeyType. [Check this table for full reference](#keyboard-return-key-types-for-web-and-native) | `default`, `go`, `done`, `next`, `search`, `send` | No | Closest based on the `type` attribute | Determines how the return key should look on the keyboard on mobile devices or virtual keyboard |
```
divyanshu013, line 51: Should we specify this? It can't be for example search, send etc.?

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
```diff
@@ -0,0 +1,92 @@
...
+| onOTPFilled | `({ name, value })=> {}` | No |  | Callback function when all fields of the OTPInput are filled |
+
```
divyanshu013, line 53: Maybe keyboard type as well? User can specify which soft keypad the user should be shown depending on whether the otp is numeric or alphanumeric.

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
nashcheez: Based on the [OTP Input design guidelines](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=11534%3A151829), we should not have an `optional` tag, this should always be a mandatory field. Design needs to decide if we should have the `*` or not in the OTP Input field label irrespective.

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
```diff
@@ -0,0 +1,92 @@
...
+| validationState | `error` , `success`, `none` | No |  | Applies the style to the input field based on validation status and shows either `errorText` or `successText` respectively |
+| helpText | `string` | No |  | Shown when we want to add some hint to the input field. Displayed under the input field. Only one of `helpText`, `errorText` or `successText` is shown at a time in the priority order as `errorText`, `successText`, `helpText` |
+| errorText | `string` | No |  | Shown when the `validationState` of the input field is set to Error. Only one of `helpText`, `errorText` or `successText` is shown at a time in the priority order as `errorText`, `successText`, `helpText` |
+| successText | `string` | No |  | Shown when the `validationState` of the input field is set to Success. Only one of `helpText`, `errorText` or `successText` is shown at a time in the priority order as `errorText`, `successText`, `helpText` |
+| onChange | `({ name, value }) => {}` | No |  | Function called when the value of the input field changes |
+| onBlur | `({ name, value }) => {}` | No |  | Function called when the input field loses focus |
```
nashcheez, line 47: I was of the assumption that the OTP Input field is a combination of either 4 or 6 separate inputs. How is the `onBlur` supposed to work here? Will it get triggered when each of the inputs lose focus?

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
```diff
@@ -0,0 +1,92 @@
...
+| validationState | `error` , `success`, `none` | No |  | Applies the style to the input field based on validation status and shows either `errorText` or `successText` respectively |
+| helpText | `string` | No |  | Shown when we want to add some hint to the input field. Displayed under the input field. Only one of `helpText`, `errorText` or `successText` is shown at a time in the priority order as `errorText`, `successText`, `helpText` |
+| errorText | `string` | No |  | Shown when the `validationState` of the input field is set to Error. Only one of `helpText`, `errorText` or `successText` is shown at a time in the priority order as `errorText`, `successText`, `helpText` |
+| successText | `string` | No |  | Shown when the `validationState` of the input field is set to Success. Only one of `helpText`, `errorText` or `successText` is shown at a time in the priority order as `errorText`, `successText`, `helpText` |
+| onChange | `({ name, value }) => {}` | No |  | Function called when the value of the input field changes |
```
nashcheez, line 46: Do we need both `onChange` and `onBlur` here? I can see both events getting triggered parallel in many occasions.

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
```diff
@@ -0,0 +1,92 @@
...
+| onBlur | `({ name, value }) => {}` | No |  | Function called when the input field loses focus |
+| isDisabled | `boolean` | No | `false` | Marks the input field as disabled. It'll also set `aria-disabled` to true and hence the input field will loose keyboard focus. |
+| isRequired | `boolean` | No | `false` | Marks the input field as required. if the errorText is provided then that will be shown if the field is required and not filled, else `${label} is required` |
```
nashcheez, line 49: Won't this be always required? Do we need this prop?

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
chaitanyadeorukhkar: In our use-cases its 4 or 6 and that's what the design is also aligned to. If we see more use-cases, we can always add more or open this up to be a number.

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
chaitanyadeorukhkar: The indicator will depend on the design. We should give them the choice to show it or not

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
chaitanyadeorukhkar: `done`,`next` & `send` seem to be valid cases. This is still an option that people can specify or let the default behavior be as is.  We could remove the option of setting `search` but that seems over-optimizing it.

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
chaitanyadeorukhkar: I don't think we have use-cases for alphanumeric OTP. I was thinking of forcing it to be numeric. 

@saurav12 should we restrict it to numeric?

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
chaitanyadeorukhkar: Yes, `onBlur` of any of the OTP field we should send back the value of the entered OTP

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
chaitanyadeorukhkar: They will get triggered, yes but there are cases where `onChange` won't be triggered but `onBlur` would be. Do you think it's a good idea to not give the onBlur callback at all then?

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
divyanshu013: Makes sense, let's keep it as it's then

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
chaitanyadeorukhkar: Yeah, makes sense. Lets make it always required

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
```diff
@@ -0,0 +1,93 @@
...
+| necessityIndicator | `optional`, `required` | No |  | Indicator to be shown next to the label for emphasizing the necessity of the field. if `optional` is provided then it'll show `${label} (optional)` and if `required` is provided then it'll show `${label} *` |
+| name | `string` | No |  | The name of the input, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name) |
+| value | `string` | No |  | Makes input field [controlled](https://reactjs.org/docs/forms.html#controlled-components) |
+| defaultValue | `string` | No |  | Used to set the default value of input field when it's [uncontrolled](https://reactjs.org/docs/uncontrolled-components.html#default-values) controlled |
+| validationState | `error` , `success`, `none` | No |  | Applies the style to the input field based on validation status and shows either `errorText` or `successText` respectively |
```
akkhil7, line 42: Can this not be handled internally if it's based on otp length? Or are we letting consumer add custom validation also?

chaitanyadeorukhkar: It won't just be based on OTP length, right? A user could input an invalid OTP, and the consumer would also need to show that error. Its better to keep the API similar to rest of the TextInputs as well so that the familiarity and ease of use exists.

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
divyanshu013: On password field we're not allowing optional here, maybe we can follow the same convention here? Need to confirm with design.

divyanshu013 (approved):
A couple of minor comments above, looks good otherwise.

packages/blade/src/components/Input/OTPInput/_decisions/_decisions.md
chaitanyadeorukhkar: Discussed with Saurav, we won't be having any necessity indicator

divyanshu013 (approved):

PR #718, opened 2022-09-20
feat(Code): Implementation Web & Native

saurabhdaware:
## Description

This PR adds Web and Native implementation for `Code` component.

**PR Stack**
- [x] API Decisions (#711)
- [x] Web and Native Implementation (this PR)
- [x] ~Storybook Docs Improvements (todo) (Basic storybook is added in this PR)~ (Added it in this PR itself)
- [x] Tests (#729)
- [ ] Add final export, changeset, and release component (todo)

### Android

image

### iOS
image

### Web

image

### DIFF
--- a/packages/blade/package.json
+++ b/packages/blade/package.json
@@ -55,11 +55,6 @@
     "react": "yarn run react:storybook",
     "react:storybook": "FRAMEWORK=REACT start-storybook -c ./.storybook/react -p 9009",
     "react:storybook:build": "FRAMEWORK=REACT build-storybook -c ./.storybook/react -o storybook-site",
-    "start:ios": "run-p react-native:storybook:start react-native:storybook:ios",
-    "start:android": "run-p react-native:storybook:start react-native:storybook:android",
-    "start:native": "run-p react-native:storybook:start react-native:storybook:android react-native:storybook:ios",
-    "start:web": "yarn react",
-    "start:all": "run-p start:native start:web",
     "chromatic": "npx chromatic"
   },
   "dependencies": {
--- a/packages/blade/package.json
+++ b/packages/blade/package.json
@@ -55,6 +55,11 @@
     "react": "yarn run react:storybook",
     "react:storybook": "FRAMEWORK=REACT start-storybook -c ./.storybook/react -p 9009",
     "react:storybook:build": "FRAMEWORK=REACT build-storybook -c ./.storybook/react -o storybook-site",
+    "start:ios": "run-p react-native:storybook:start react-native:storybook:ios",
+    "start:android": "run-p react-native:storybook:start react-native:storybook:android",
+    "start:native": "run-p react-native:storybook:start react-native:storybook:android react-native:storybook:ios",
+    "start:web": "yarn react",
+    "start:all": "run-p start:native start:web",
     "chromatic": "npx chromatic"
   },
   "dependencies": {
--- a/packages/blade/src/components/Typography/Code/Code.stories.tsx
+++ b/packages/blade/src/components/Typography/Code/Code.stories.tsx
@@ -13,10 +13,9 @@ const CodeStoryMeta: Meta = {
 };
 
 const CodeTemplate: ComponentStory<typeof CodeComponent> = (args) => (
+  // For React Native, use flex to align items correctly
   <Text>
-    Lorem ipsum normal text <CodeComponent {...args} /> something something
-    <br />
-    also something something else lorem ipsum
+    Lorem ipsum normal text <CodeComponent {...args} /> component
   </Text>
 );
 
--- a/packages/blade/src/components/Typography/Code/Code.stories.tsx
+++ b/packages/blade/src/components/Typography/Code/Code.stories.tsx
@@ -1,23 +1,24 @@
 import type { ComponentStory, Meta } from '@storybook/react';
-import { Code } from './Code';
+import { Text } from '../Text';
+import { Code as CodeComponent } from './Code';
 
-const TextStoryMeta: Meta = {
+const CodeStoryMeta: Meta = {
   title: 'Components/Typography/Code',
-  component: Code,
+  component: CodeComponent,
   args: {
-    variant: 'body',
-    weight: 'regular',
     size: 'medium',
-    type: 'normal',
+    type: 'subtle',
     children: 'SENTRY_AUTH_TOKEN',
-    truncateAfterLines: 3,
-    contrast: 'low',
   },
 };
 
-const CodeTemplate: ComponentStory<typeof Code> = (args) => {
-  return <Code {...args}>{args.children}</Code>;
-};
+const CodeTemplate: ComponentStory<typeof CodeComponent> = (args) => (
+  <Text>
+    Lorem ipsum normal text <CodeComponent {...args} /> something something
+    <br />
+    also something something else lorem ipsum
+  </Text>
+);
 
-export default TextStoryMeta;
-export const CodeExport = CodeTemplate.bind({});
+export default CodeStoryMeta;
+export const Code = CodeTemplate.bind({});
--- a/packages/blade/src/components/Typography/Code/Code.stories.tsx
+++ b/packages/blade/src/components/Typography/Code/Code.stories.tsx
@@ -1,0 +1,23 @@
+import type { ComponentStory, Meta } from '@storybook/react';
+import { Code } from './Code';
+
+const TextStoryMeta: Meta = {
+  title: 'Components/Typography/Code',
+  component: Code,
+  args: {
+    variant: 'body',
+    weight: 'regular',
+    size: 'medium',
+    type: 'normal',
+    children: 'SENTRY_AUTH_TOKEN',
+    truncateAfterLines: 3,
+    contrast: 'low',
+  },
+};
+
+const CodeTemplate: ComponentStory<typeof Code> = (args) => {
+  return <Code {...args}>{args.children}</Code>;
+};
+
+export default TextStoryMeta;
+export const CodeExport = CodeTemplate.bind({});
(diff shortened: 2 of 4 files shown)


packages/blade/package.json
```diff
@@ -55,6 +55,11 @@
     "react": "yarn run react:storybook",
     "react:storybook": "FRAMEWORK=REACT start-storybook -c ./.storybook/react -p 9009",
     "react:storybook:build": "FRAMEWORK=REACT build-storybook -c ./.storybook/react -o storybook-site",
+    "start:ios": "run-p react-native:storybook:start react-native:storybook:ios",
```
saurabhdaware, line 58: TODO: pull master to remove this

packages/blade/src/components/Typography/Code/Code.tsx
```diff
@@ -0,0 +1,96 @@
...
+   *
+   * ```tsx
+   * // ❌ Bad
+   * <Code><Box>something</Box></Code>
+   * ```
+   *
+   *```tsx
+   * // ✅ Good
+   * <Code>something</Code>
+   * ```
```
anuraghazra, line 19: Don't think this is needed, since the TS error will be there anyway.

packages/blade/src/components/Typography/Code/Code.tsx
```diff
@@ -0,0 +1,96 @@
...
+  children: React.ReactNode;
+};
+
+const platformType = getPlatformType();
+const isPlatformWeb = platformType === 'browser' || platformType === 'node';
+
+const CodeContainer = styled(Box)<CodeContainerProps>((props) => {
+  const padding =
+    props.size === 'large'
+      ? `${props.theme.spacing[0]}px ${props.theme.spacing[2]}px`
```
anuraghazra, line 47: let's use the makeSpace utility

packages/blade/src/components/Typography/Code/Code.tsx
```diff
@@ -0,0 +1,96 @@
...
+  return {
+    padding,
+    backgroundColor: props.theme.colors.brand.gray[300],
+    borderRadius: props.theme.border.radius.medium,
+    display: isPlatformWeb ? 'inline-block' : undefined,
+  };
+});
+
+/**
+ * Code component can be used for displaying token, variable names, or inlined code snippets.
```
anuraghazra, line 58: Nice.

packages/blade/src/components/Typography/Code/Code.tsx
```diff
@@ -0,0 +1,96 @@
...
+ *   <Code>SENTRY_TOKEN</Code>
+ *   <Text> normal text</Text>
+ * </Box>
+ * ```
+ */
+const Code = ({ children, size = 'medium', type = 'subtle' }: CodeProps): JSX.Element => {
+  return (
+    <CodeContainer size={size} as={isPlatformWeb ? 'span' : undefined}>
+      <BaseText
+        color={`surface.text.${type ?? 'subtle'}.lowContrast`}
```
anuraghazra, line 85: Do we need to do `type ?? subtle`? 

We are already setting default value to type in the param.

packages/blade/src/components/Typography/Code/Code.tsx
```diff
@@ -0,0 +1,96 @@
...
+   * Decides the visibility and color of Code
+   *
+   * @default subtle
+   */
+  type?: TextTypes;
+};
+
+type CodeContainerProps = {
+  size: CodeProps['size'];
+  children: React.ReactNode;
```
anuraghazra, line 38: 🤔 don't think you need to specify children for styled components.

packages/blade/src/components/Typography/Code/Code.tsx
saurabhdaware: I did for TS. It was still thinking that value can be undefined 🤔

packages/blade/src/components/Typography/Code/Code.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+};
+
+const CodeTemplate: ComponentStory<typeof CodeComponent> = (args) => (
+  // For React Native, use flex to align items correctly
+  <Text>
+    Lorem ipsum normal text <CodeComponent {...args} /> component
+  </Text>
+);
+
+export default CodeStoryMeta;
```
anuraghazra, line 22: Maybe we should add a kitchen sink example with all the variants.

packages/blade/src/components/Typography/Code/Code.tsx
saurabhdaware: Oh right. Will remove it.

packages/blade/src/components/Typography/Code/Code.tsx
```diff
@@ -0,0 +1,96 @@
+import styled from 'styled-components';
```
anuraghazra, line 1: Add index.ts and rexport the Code. 

And also reexport on Typography/index.ts

packages/blade/src/components/Typography/Code/Code.stories.tsx
saurabhdaware: Was thinking that but we don't really have kitchen sink for other typography components and also the variants can be changed from docs controller itself and they don't visibility look that different because it's just change in size.

packages/blade/src/components/Typography/Code/Code.tsx
saurabhdaware: Was planning to do that at the end in followup PR. So that this PR can be merged after review. Added the todo for this in description

packages/blade/src/components/Typography/Code/Code.tsx
anuraghazra: Don't think it's throwing error, maybe try restarting the ts server?

saurabhdaware: ohhh right my badd!! My earlier implementation was a bit different where there was in some nested component where it was throwing the error. You're right its not throwing error here. Will remove it.

packages/blade/src/components/Typography/Code/Code.stories.tsx
anuraghazra: Yes in fact we should add other kitchen sink too. 

My main reason of adding kitchen sink examples is to catch visual regressions with chromatic. 
If the storybook doesn't render the size and other variants chromatic won't screenshot them.

packages/blade/src/components/Typography/Code/Code.tsx
saurabhdaware: done

packages/blade/src/components/Typography/Code/Code.tsx
saurabhdaware: done

packages/blade/src/components/Typography/Code/Code.tsx
saurabhdaware: done

packages/blade/src/components/Typography/Code/Code.tsx
saurabhdaware: done

packages/blade/src/components/Typography/Code/Code.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+  component: CodeComponent,
+  args: {
+    size: 'medium',
+    type: 'subtle',
+    children: 'SENTRY_AUTH_TOKEN',
+  },
+};
+
+const CodeTemplate: ComponentStory<typeof CodeComponent> = (args) => (
+  // For React Native, use flex to align items correctly
```
divyanshu013, line 16: Didn't understand the context of this comment, seems unrelated 🤔

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Typography/Code/Code.stories.tsx
saurabhdaware: As in that story will not look properly aligned on react native. For that we have to use flex and do it properly. I thought of conditionally adding a react native layout based on platform but might be overkill since storybook is mostly visited from web.

packages/blade/src/components/Typography/Code/Code.stories.tsx
saurabhdaware: Added the sink 👯

divyanshu013:
You can also click on UI tests and review them. If you're happy with the results you may click approve on Chromatic.

image

chaitanyadeorukhkar (approved):

packages/blade/src/components/Typography/Code/Code.tsx
```diff
@@ -66,13 +59,13 @@ const CodeContainer = styled(Box)<CodeContainerProps>((props) => {
...
-const Code = ({ children, size = 'medium', type = 'subtle' }: CodeProps): JSX.Element => {
+const Code = ({ children, size = 'small' }: CodeProps): JSX.Element => {
   return (
     <CodeContainer size={size} as={isPlatformWeb ? 'span' : undefined}>
       <BaseText
-        color={`surface.text.${type}.lowContrast`}
+        color="surface.text.subtle.lowContrast"
         fontFamily="code"
-        fontSize={size === 'large' ? 100 : 75}
+        fontSize={size === 'small' ? 75 : 100}
```
chaitanyadeorukhkar, line 68: nit: create some function that takes in `size`, has a switch case or if..elseif and returns 75 or 100 depending on the size? This way if we add another size in the future, we will have a cleaner implementation for it instead of just checking for `small`?

chaitanyadeorukhkar (approved):
One feedback, otherwise lgtm

packages/blade/src/components/Typography/Code/Code.tsx
saurabhdaware: done

PR #723, opened 2022-09-22
fix(BaseInput): use cursor not-allowed for disabled inputs

divyanshu013:
## Screenshots

| PasswordField | TextInput |
| - | - |
|  Screen Shot 2022-09-22 at 13 00 43 | Screen Shot 2022-09-22 at 13 00 57 |

### DIFF
--- a/.changeset/spicy-mails-doubt.md
+++ b/.changeset/spicy-mails-doubt.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix(BaseInput): use cursor not-allowed for disabled inputs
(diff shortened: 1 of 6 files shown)


anuraghazra:
We should also add this for TextArea?

nashcheez:
@anuraghazra yes, we should. I was of the assumption that if you add it to BaseInput it should automatically be resolved in TextInput, Password and TextArea?

anuraghazra:
> @anuraghazra yes, we should. I was of the assumption that if you add it to BaseInput it should automatically be resolved in TextInput, Password and TextArea?

Oh LOL yeah.. it's on baseInput, so does also propagate to textarea 😆

anuraghazra (approved):

PR #724, opened 2022-09-22
chore: remove blade old

anuraghazra:
Changes in this PR

- Removed the blade-old folder under packages
- Removed all documentation & other references to blade-old
- Removed .js extensions from eslintrc
- Removed blade-old related references from tsconfig/package.json


CODEOWNERS
```diff
@@ -1,4 +1,3 @@
 /blade/ @chaitanyadeorukhkar @kamleshchandnani @dhruvdutt @anuraghazra @nashcheez @divyanshu013 @saurabhdaware
```
kamleshchandnani, line 1: only keep this? can remove everything else now

tsconfig.json
```diff
@@ -1,7 +1,6 @@
 {
   "exclude": [
     "node_modules",
-    "packages/blade-old/**",
```
kamleshchandnani, line 4: okay so the entire PR is just removing the `blade-old` package. we also need to modify a lot of config files for eg:
1. refactor the eslint files to just have ts things and remove everything else
2. what if we also remove the eslint, tsconfig, etc all the files  from inside the package/blade and keep them outside?
3. do we really need to keep the blade under `packages/blade` what else will we put in `packages`? examples and other stuff?

anuraghazra: > what if we also remove the eslint, tsconfig, etc all the files from inside the package/blade and keep them outside?

Don't think that would span out very well. eslint is already on the root. 

But if we put tsconfig outside then if you add new packages inside packages folder then they will also have all other moduleSuffixes stuff which is blade specific. 

> do we really need to keep the blade under packages/blade what else will we put in packages

Yeah? Even if we removed the blade-old, It's still a mono repo right. In future if we add other packages (figma plugins, examples, playground, docs) we should add them in packages folder.

CODEOWNERS
anuraghazra: Yep removed. not sure why /blade/ was even their on the first place.

tsconfig.json
chaitanyadeorukhkar: Lets keep it a mono-repo for adding playground, docs, etc 👍

chaitanyadeorukhkar (commented):
LGTM. Few pointers @anuraghazra 

- Add a link to the new repository where we're moving `blade-old` in this PR's description for future reference
- Draft an announcement for internal purposes to let the consumers of `blade-old` know the changes
- Let's migrate all of these things to the new repository, test out a release with the new repo, and only then merge this PR. If we merge this PR and the `blade-old` consumers need some immediate patch, it would be an extra effort. Let's ensure we're prepared to ship `blade-old` from the new repo before removing it from here.

README.md
```diff
@@ -19,10 +19,12 @@
 
 ## 📦 Packages
 Blade has 2 packages
```
kamleshchandnani, line 21: remove this line. and we can put the old blade directly without emphasizing over it and putting it as section header?

kamleshchandnani (approved):

README.md
```diff
@@ -17,17 +17,23 @@
...
+
+
+------
+
+> **Note**
+>
+> The blade-old package has been moved to it's own repo.
+> [github.com/razorpay/blade-old](https://github.com/razorpay/blade-old)
+> 
+> It is under **maintenance** and it won't have any major releases. It will be **deprecated** once the newer version (`@razorpay/blade`) is ready for a stable release.
```
kamleshchandnani, line 36:
```suggestion
> It is under **maintenance** and it won't have any major releases.
```

let's remove that line. I'm not sure when will it be deprecated. and new blade is already pushing out stable releases in chunks

chaitanyadeorukhkar (approved):

PR #728, opened 2022-09-26
chore: added counter API

anuraghazra:
## [View Formatted ](https://github.com/razorpay/blade/blob/anu/counter-api/packages/blade/src/components/Counter/_decisions/decisions.md)


packages/blade/src/components/Counter/_decisions/decisions.md
```diff
@@ -0,0 +1,46 @@
...
+  variant='positive'
+  contrast='high'
+/>
+```
+
+## Accessibility
+
+- Only design accessibility needed
+
+## Open Questions
```
divyanshu013, line 37: Question: why do we need a separate counter component when the same UI can be replicated with badges?

divyanshu013 (approved):
Looks good to me.

chaitanyadeorukhkar (approved):

packages/blade/src/components/Counter/_decisions/decisions.md
anuraghazra: Might need some inputs from @saurav12. I'll ask him.

packages/blade/src/components/Counter/_decisions/decisions.md
```diff
@@ -0,0 +1,46 @@
...
+| intent | `positive`,`negative`, `notice`, `information`, `neutral` | `neutral` | The variant of the counter to be rendered. |  |
+| contrast | `low`, `high` | `low` | The contrast of the counter to be rendered. |  |
+
+## Usage Example
+
+```tsx
+import { Counter } from '@razorpay/components';
+
+<Counter 
+  value={150}
```
kamleshchandnani, line 26: why `value` as a prop and not a children similar to Typography components?

packages/blade/src/components/Counter/_decisions/decisions.md
kamleshchandnani: because the api will need tweaks for eg: the max value, the type of content we put it between badges and counters, etc.

packages/blade/src/components/Counter/_decisions/decisions.md
```diff
@@ -0,0 +1,46 @@
...
+/>
+```
+
+## Accessibility
+
+- Only design accessibility needed
+
+## Open Questions
+
+- Similar to badge's [discussion around](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Badge/_decisions/decisions.md#variant-prop) naming the `intent` prop `variant` should we call it `variant`?
```
kamleshchandnani, line 39: don't see a use case for counter that is outside the `intent` scope. confirm with @saurav12 once.

packages/blade/src/components/Counter/_decisions/decisions.md
```diff
@@ -0,0 +1,46 @@
+# Counter 
+
+Counters are visual indicators that contains numerical values, tallies or counts in regards to some context.
```
kamleshchandnani, line 3: how are we thinking to tackle use cases where we might want to show the counters on top of some icon or something?
Screenshot 2022-10-26 at 1 23 30 PM

packages/blade/src/components/Counter/_decisions/decisions.md
```diff
@@ -0,0 +1,46 @@
...
+## Design
+
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=8222%3A70827) to all variants of the Counter component
+
+## API
+
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| value | `number` | `undefined` | The numeric value of the counter which will be shown. | ✅ |
+| max | `number` | `undefined` | The max value of the counter, if the `value` exceededs `max` then the counter will show a `+` at the end. |  |
```
kamleshchandnani, line 16: can we make it more verbose? `max` right now seems vague of what it will refer to? maybe `maxValue` or something better you can think of?

packages/blade/src/components/Counter/_decisions/decisions.md
anuraghazra: _2 reasons:_ 

1. The counters will accept numbers only so the value have to be set to `number`
 
Setting the `children: number` the consumers will have to use JSX expressions to pass the number in children which looks kinda odd, otherwise TS will yell. 

```tsx
<Counter>{20}</Counter>

<Counter>20</Counter> // <-- error
```

2. When the `max` prop is set we also need to append `+` at the end of the `value`. If we use `children` intuitively the expectation might be that whatever we put in the `children` will be rendered but in reality we will modify what's being rendered. 

```tsx
// renders `10+`
<Counter max={10}>{20}</Counter>

// vs

// renders `10+` similar API to `<input value={} max={} />`
<Counter value={20} max={10} />  
```

packages/blade/src/components/Counter/_decisions/decisions.md
anuraghazra: I don't think there's any opinion in this from design side, so right now we don't support it out of the box. So users have to use absolute positioning. 

I don't see any mention about this usecase in figma also. 

// cc @saurav12

packages/blade/src/components/Counter/_decisions/decisions.md
anuraghazra: Wanted to keep things concise like `<input type="number" value="10" max="5" />` 

Tho don't have any strong opinions in this, `maxValue` also sounds okay.

packages/blade/src/components/Counter/_decisions/decisions.md
anuraghazra: 🤔 

"Counters can't be combined with icons". 

image

divyanshu013: This usecase of showing an indicator is designed in [indicators](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=8505%3A55)

packages/blade/src/components/Counter/_decisions/decisions.md
kamleshchandnani: makes sense. i think let it be `max`

kamleshchandnani:
LGTM ✅

kamleshchandnani (approved):
LGTM ✅

packages/blade/src/components/Counter/_decisions/decisions.md
saurav12: Nice catch. There might be similar to Badges.

packages/blade/src/components/Counter/_decisions/decisions.md
saurav12: We might tackle the use-case that @kamleshchandnani mentioned for a counter on top of the icon. I see it as a valid use-case.

PR #729, opened 2022-09-27
tests(Code): add tests for `Code` component

saurabhdaware:
## Stack
- Implementation PR #718 
- Tests PR (this) 
- Release PR #732 

## Changes in this PR

- Add tests for native
- Add tests for web


packages/blade/src/components/Typography/Code/__tests__/Code.native.test.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Code } from '../Code';
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
+
+describe('<Code />', () => {
+  it('should render Code with default properties', () => {
+    const { getByText, toJSON } = renderWithTheme(<Code>TEST_TOKEN</Code>);
+    expect(getByText('TEST_TOKEN')).toBeTruthy();
```
divyanshu013, line 8: Minor (optional): just `getByText` assertion is enough. I realized this later we don't need `toBeTruthy` because the test will fail if `getByText` fails.

packages/blade/src/components/Typography/Code/__tests__/Code.native.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render large Code with different types', () => {
+    const { toJSON } = renderWithTheme(<Code size="small">SMALL</Code>);
+    expect(toJSON()).toMatchSnapshot();
```
divyanshu013, line 14: Can we add an assertion on a property that gets affected by `size="small"`, it would be easier than reviewing snapshot diffs

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Typography/Code/__tests__/Code.native.test.tsx
saurabhdaware: Ohh righttt!! removed the assertion

packages/blade/src/components/Typography/Code/__tests__/Code.native.test.tsx
saurabhdaware: Thanks for this suggestion!! Added `.toHaveStyle` assertions and they are so much better! We might as well remove snapshots and rely on them in future.

divyanshu013: Nice, maybe instead of hard coding 12 you could use the theme token if that seems better

packages/blade/src/components/Typography/Code/__tests__/Code.web.test.tsx
```diff
@@ -4,17 +4,23 @@ import renderWithTheme from '~src/_helpers/testing/renderWithTheme.web';
 
 describe('<Code />', () => {
   it('should render Code with default properties', () => {
-    const { container } = renderWithTheme(<Code>TEST_TOKEN</Code>);
+    const { container, getByText } = renderWithTheme(<Code>TEST_TOKEN</Code>);
+    expect(getByText('TEST_TOKEN')).toHaveStyle('font-size: 0.75rem');
+    expect(container.querySelector('span')).toHaveStyle('padding: 0px 4px');
```
divyanshu013, line 9: These two assertions might be overkill

saurabhdaware: Actually the padding is important in this case since if we break padding it also breaks the line heights and alignments.

saurabhdaware: Oh just realised I added it to all tests. Kept it only in first test and removed it from other tests.

packages/blade/src/components/Typography/Code/__tests__/Code.native.test.tsx
saurabhdaware: done

chaitanyadeorukhkar (approved):

PR #730, opened 2022-09-29
feat(Typography)!: make `Heading`, `Title`, and `Text` API consistent

saurabhdaware:
> **Warning**
>
> Breaking Change. Changes the prop names and values in some Typography components.

## Problem

Slack Discussion - https://razorpay.slack.com/archives/G01B3LQ9H0W/p1663325670139239?thread_ts=1663318580.283669&cid=G01B3LQ9H0W

Currently the API for typography components is not consistent with each other. For e.g. - 
```jsx
<>
  <Title variant="medium">Medium Title</Title>
  <Heading variant="small">Small Heading</Heading>
  <Text size="small">Small Text</Text>
</>
```

As you can see some component use `variant` for deciding the size whereas `Text` uses `size`.

## Changes

- In `Title`: Rename `variant` to `size`
- In `Heading`: Add `variant="regular"` and add size with `small | medium | large` type

## New Code

All typography components will now use `size` prop for deciding the size. 

```jsx
<>
  <Title size="medium">Medium Title</Title>
  <Heading size="small" variant="regular">Small Heading</Heading>
  <Text size="small">Small Text</Text>
</>
```

## Error Handling

This change adds one scenario where someone can add `subheading` and `size` both at the same time which is invalid.

```jsx
<Heading variant="subheading" size="small">Small Heading</Heading>
```

The TS will handle this scenario and show type error
image

### DIFF
--- a/.changeset/five-tomatoes-try.md
+++ b/.changeset/five-tomatoes-try.md
@@ -1,23 +1,27 @@
 ---
-"@razorpay/blade": major
+'@razorpay/blade': major
 ---
 
 feat(Typography): make `size` prop consistent for `Heading`, `Title`, and `Text`
 
 > **Note**
 >
-> Breaking Change! 
+> Breaking Change!
 > This is a breaking change for apps that are using `Title` or `Heading` component from blade. Rest of the apps can upgrade without any migrations.
 
 #### Migration
 
+_**Tip:** If you're using TypeScript, run `yarn tsc` and that should throw errors wherever a change is required._
+
 1. **`<Title />`:** Rename `variant` prop to `size` in Title
+
 ```diff
 - <Title variant="small">Some Title</Title>
 + <Title size="small">Some Title</Title>
 ```
 
-2. **`<Heading />`:** Rename `variant` prop to `size` if the value is `small`, `medium,` or `large`. No change is on `variant="subheading"`.
+2. **`<Heading />`:** Rename `variant` prop to `size` if the value is `small`, `medium,` or `large`. No change is required on `variant="subheading"`.
+
 ```diff
 <Heading variant="subheading">Nothing changes here</Heading> // No change here
 
--- a/.changeset/five-tomatoes-try.md
+++ b/.changeset/five-tomatoes-try.md
@@ -1,0 +1,39 @@
+---
+"@razorpay/blade": major
+---
+
+feat(Typography): make `size` prop consistent for `Heading`, `Title`, and `Text`
+
+> **Note**
+>
+> Breaking Change! 
+> This is a breaking change for apps that are using `Title` or `Heading` component from blade. Rest of the apps can upgrade without any migrations.
+
+#### Migration
+
+1. **`<Title />`:** Rename `variant` prop to `size` in Title
+```diff
+- <Title variant="small">Some Title</Title>
++ <Title size="small">Some Title</Title>
+```
+
+2. **`<Heading />`:** Rename `variant` prop to `size` if the value is `small`, `medium,` or `large`. No change is on `variant="subheading"`.
+```diff
+<Heading variant="subheading">Nothing changes here</Heading> // No change here
+
+- <Heading variant="medium">Medium Heading</Heading>
++ <Heading size="medium">Medium Heading</Heading>
+```
+
+##### Edge Cases
+
+Make sure to follow migration on new component if `Title` or `Heading` from blade is overriden with styled-components.
+
+```diff
+const MyTitle = styled(Title)`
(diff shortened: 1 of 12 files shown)


chaitanyadeorukhkar (approved):

packages/blade/src/components/Typography/Heading/__tests__/Heading.native.test.tsx
```diff
@@ -13,43 +13,43 @@ describe('<Heading />', () => {
     expect(toJSON()).toMatchSnapshot();
   });
 
-  it('should render Heading with variant "small" and contrast "high"', () => {
+  it('should render Heading with size "small" and contrast "high"', () => {
```
anuraghazra, line 16: If there is a possibility of impossible state, let's test that unhappy path.

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -16,11 +17,16 @@ type HeadingCommonProps = {
...
+  size?: HeadingSize;
   weight?: keyof Theme['typography']['fonts']['weight'];
 };
 
 type HeadingSubHeadingVariant = HeadingCommonProps & {
   variant?: Extract<HeadingVariant, 'subheading'>;
+  /**
+   * `size` cannot be used with variant="subheading". Either change to variant="regular" or remove size prop
+   */
+  size?: undefined;
```
anuraghazra, line 29: Well, this I feel like an arguably less ideal that what we had before. This introduces one more impossible state :sad:

saurabhdaware: True but don't see any other way to do this without introducing inconsistency in prop names. This way we can atleast say that size is decided by `size` prop and subheading just becomes an exception where size can't be changed.

packages/blade/src/components/Typography/Heading/__tests__/Heading.native.test.tsx
saurabhdaware: Pretty simple component so can't think of any impossible state as such 🤔 

One case would be having `size` on `variant="subheading"` but TS would handle that right? do we need runtime checks there?

.changeset/five-tomatoes-try.md
```diff
@@ -0,0 +1,43 @@
+---
+'@razorpay/blade': major
+---
+
+feat(Typography): make `size` prop consistent for `Heading`, `Title`, and `Text`
+
+> **Note**
```
divyanshu013, line 7:
```suggestion
> **Warning**
```

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -16,11 +17,16 @@ type HeadingCommonProps = {
...
   variant?: Exclude<HeadingVariant, 'subheading'>;
+  size?: HeadingSize;
   weight?: keyof Theme['typography']['fonts']['weight'];
 };
 
 type HeadingSubHeadingVariant = HeadingCommonProps & {
   variant?: Extract<HeadingVariant, 'subheading'>;
+  /**
+   * `size` cannot be used with variant="subheading". Either change to variant="regular" or remove size prop
+   */
```
divyanshu013, line 28: Suggestion: add a default for JSDoc

divyanshu013 (approved):
Let's also review the chromatic snapshot diffs because this touches various other components. Although I don't see any snapshot diffs which means this would have ideally not created any side effects.

packages/blade/src/components/Typography/Heading/Heading.tsx
saurabhdaware: Done. Added it to the other `HeadingNormalVariant` type since the actual type was defined there. We're just overriding here to undefined to make sure people don't pass size with `variant="subheading"`

.changeset/five-tomatoes-try.md
saurabhdaware: done

packages/blade/src/components/Typography/Heading/__tests__/Heading.native.test.tsx
saurabhdaware: Added the runtime check and unhappy flow for size and subheading variant

anuraghazra: Cool. 
IMO we should all together try to reduce the number of impossible states as much possible. It's always better DX even with TS.

anuraghazra (approved):

PR #731, opened 2022-10-03
feat: add OTPInput component

chaitanyadeorukhkar:
Adds the `OTPInput` component

### Demo
https://user-images.githubusercontent.com/24487274/193798631-42835217-b654-4bf0-8582-cf0847441cdf.mov

Chained PR: #735 

### Todo Checklist
- [x] PR for API Decisions: #713 
- [x] Web implementation
- [x] Native implementation
- [x] Accessibility
- [x] Stories
- [x] Open questions to be discussed with the team
- [x] Tests: #735


packages/blade/src/components/Form/FormLabel.tsx
```diff
@@ -16,7 +16,7 @@ type CommonProps = {
 };
 
 type LabelProps = CommonProps & {
-  htmlFor: string;
+  htmlFor?: string;
```
chaitanyadeorukhkar, line 19: Since I'm adding `FormLabel` outside of the `BaseInput` instead of using `BaseInput`'s `label` prop, I can't be certain on what the generated id would be for the first `BaseInput` and hence can't pass it as `htmlFor`.

> I can't pass `label` directly to `BaseInput` since it will move the first label down vertically causing it to break alignment with rest of the labels and the space for the label would be limited to that of the first input's width if we attach the label with the first `BaseInput`

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -58,6 +59,14 @@ export type BaseInputProps = FormInputLabelProps &
      * The callback function to be invoked when the value of the input field changes
      */
     onChange?: FormInputOnEvent;
+    /**
+     * The callback function to be invoked when the value of the input field has any input
+     */
+    onInput?: FormInputOnEvent;
```
chaitanyadeorukhkar, line 65: `onChange` only triggers when actual value changes but I need to move from one input to another even when the actual value remains the same. For instance, if value of otpInput1 is `7` and I select the value and press `7` again, `onChange` won't be triggered since there is no change but I still need to change the focus to the next input.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -58,6 +59,14 @@ export type BaseInputProps = FormInputLabelProps &
...
+    /**
+     * The callback function to be invoked when the value of the keydown events
+     */
+    onKeyDown?: FormInputHandleOnKeyDownEvent;
```
chaitanyadeorukhkar, line 69: Needed to handle backspace, arrow keys & delete keys

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -168,6 +177,10 @@ export type BaseInputProps = FormInputLabelProps &
      * Sets the textarea's number of lines
      */
     numberOfLines?: 2 | 3 | 4 | 5;
+    /**
+     * Sets the accessibility label for the input
+     */
+    accessibilityLabel?: string;
```
chaitanyadeorukhkar, line 183: Need to pass accessibility label for individual otp input fields as "Character 1", "Character 2", etc.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -132,8 +134,18 @@ export const StyledBaseInput = React.forwardRef<TextInput, StyledBaseInputProps>
...
+        onChangeText={(text): void => {
+          handleOnChange?.({ name, value: text });
+          handleOnInput?.({ name, value: text });
+        }}
         onEndEditing={(event): void => handleOnBlur?.({ name, value: event?.nativeEvent.text })}
+        onKeyPress={(event): void => {
+          handleOnKeyDown?.({
+            name,
+            key: event?.nativeEvent.key,
+            event: (event as unknown) as React.KeyboardEvent<HTMLInputElement>, // TODO: handle platform specific type
```
chaitanyadeorukhkar, line 146: Need to fix this to get proper platform-specific types. Will do it while PR is reviewed

packages/blade/src/components/Input/BaseInput/baseInputStyles.ts
```diff
@@ -117,6 +117,7 @@ export const getBaseInputStyles = ({
         ? makeSpace(theme.spacing[3])
         : makeSpace(theme.spacing[4]),
     textAlign,
+    width: '100%',
```
chaitanyadeorukhkar, line 120: The individual inputs need to take up complete available width otherwise the layout messes up. Reference conversation: https://razorpay.slack.com/archives/G01B3LQ9H0W/p1663751935820919

packages/blade/src/components/Form/FormLabel.tsx
anuraghazra: Ideally, there should be another hidden input which will act as a proxy for the whole input. And the segmented OTP inputs will just be for UX/show. 

Basically

```ts
       <FormLabel htmlFor="otpinput-1" as="label" position={labelPosition}>
          {label}
        </FormLabel>
        <Box display="flex" flexDirection="row">
          {inputs}
        </Box>
        <HiddenInput id="otpinput-1" value={otpValue.join('')} />
```

packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+  LABEL_PROPS: 'Label Props',
+  VALIDATION_PROPS: 'Validation Props',
+  KEYBOARD_PROPS: 'Keyboard Props',
+};
+
+export default {
+  title: 'Components/Input/OTPInput',
+  component: OTPInputComponent,
+  args: {
+    placeholder: '••••••',
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Removed

packages/blade/src/components/Form/FormLabel.tsx
chaitanyadeorukhkar: Added a hidden input

packages/blade/src/components/Input/BaseInput/baseInputStyles.ts
chaitanyadeorukhkar: No change on the existing inputs as per chromatic. Did a manual check on existing stories for other inputs and things seem okay

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
```diff
@@ -0,0 +1,248 @@
...
+        flex={1}
+        paddingLeft={index == 0 ? 'spacing.0' : 'spacing.3'}
+        key={`${inputId}-${index}`}
+        maxWidth={100} // TODO: Get the correct value for this from design
+      >
+        <BaseInput
+          // eslint-disable-next-line jsx-a11y/no-autofocus
+          autoFocus={autoFocus && index === 0}
+          accessibilityLabel={`${index === 0 ? label : ''} character ${index + 1}`}
+          label=""
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Passed a valid label and added `hideLabelText` prop to hide the label

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
divyanshu013, line 65: Why do we need `onInput` along with `onChange` here 🤔 
Historically I think in React `onChange` works like `onInput`

chaitanyadeorukhkar: https://github.com/razorpay/blade/pull/731#discussion_r986712992

Because of this  ^

divyanshu013: Oh nice, missed this in diff view. We can perhaps add a comment in the code as well

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
```diff
@@ -0,0 +1,270 @@
...
+};
+
+const isReactNative = getPlatformType() === 'react-native';
+
+/**
+ *  Converts a string value of otp to array if passed otherwise returns an array of 6 empty strings
+ */
+const otpToArray = (code?: string): string[] => code?.split('') ?? Array(6).fill('');
+
+const OTPInput = ({
```
saurabhdaware, line 47: Maybe we can add some JSDoc with basic minimum usage example?

saurabhdaware (approved):
Looks good to me ✨ 

Added a tiny comment

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
chaitanyadeorukhkar: Good idea! Added!

PR #735, opened 2022-10-06
test: add tests for OTPInput

chaitanyadeorukhkar:
Adds web & native tests for OTPInput

Chained to PR: #731

### DIFF
--- a/packages/blade/src/components/Input/OTPInput/__tests__/OTPInput.web.test.tsx
+++ b/packages/blade/src/components/Input/OTPInput/__tests__/OTPInput.web.test.tsx
@@ -116,6 +116,7 @@ describe('<OTPInput />', () => {
       expect(input).toHaveValue(Array.from(otp)[index]);
     });
   });
+
   it('should change focus with arrow keys and tab key', async () => {
     const user = userEvent.setup();
     const label = 'Enter OTP';
@@ -135,6 +136,28 @@ describe('<OTPInput />', () => {
     expect(allInputs[1]).toHaveFocus();
     await user.keyboard('{Shift>1}{Tab}');
     expect(allInputs[0]).toHaveFocus();
+  });
+
+  it('should change focus to previous tab on backspace and delete keys', async () => {
+    const user = userEvent.setup();
+    const label = 'Enter OTP';
+    const otp = '123456';
+
+    const { getAllByLabelText } = renderWithTheme(
+      // eslint-disable-next-line jsx-a11y/no-autofocus
+      <OTPInput label={label} name="otp" />,
+    );
+
+    const allInputs = getAllByLabelText(/character/);
+    await user.type(allInputs[0], otp);
+    await user.click(allInputs[5]);
+    expect(allInputs[5]).toHaveFocus();
+    await user.keyboard('{Backspace}');
+    expect(allInputs[5]).toHaveValue('');
+    expect(allInputs[4]).toHaveFocus();
+    await user.keyboard('{Delete}');
+    expect(allInputs[4]).toHaveValue('');
+    expect(allInputs[3]).toHaveFocus();
   });
 
   it('should pass a11y', async () => {


packages/blade/src/components/Input/OTPInput/__tests__/OTPInput.web.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  it('should be focussed when autoFocus flag is passed', () => {
+    const label = 'Enter OTP';
+    // eslint-disable-next-line jsx-a11y/no-autofocus
+    const { getAllByLabelText } = renderWithTheme(<OTPInput label={label} autoFocus />);
+    const allInputs = getAllByLabelText(/character/);
+    expect(allInputs[0]).toHaveFocus();
+  });
+
+  it('should have hidden input for form submissions', () => {
```
divyanshu013, line 27: As a future enhancement we should add tests for form submissions too, will help us catch regressions https://github.com/razorpay/blade/issues/739

packages/blade/src/components/Input/OTPInput/__tests__/OTPInput.web.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+      expect(input).toBeValid();
+      expect(input).toBeEnabled();
+    });
+    // asserting inside a loop leads to an issue with multiple parallel runs ref: https://github.com/dequelabs/axe-core/issues/3426
+    await assertAccessible(allInputs[0]);
+    await assertAccessible(allInputs[1]);
+    await assertAccessible(allInputs[2]);
+    await assertAccessible(allInputs[3]);
+  });
+});
```
divyanshu013, line 143: Optional: you can add another test for testing the focus management of individual inputs with arrow / tab keys

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Input/OTPInput/__tests__/OTPInput.web.test.tsx
chaitanyadeorukhkar: Done! Thanks for the catch. Missed adding these 🙈 

Also added tests for backspace and delete key behaviours

saurabhdaware (approved):
Looks good to me 🌮

packages/blade/src/components/Input/OTPInput/__tests__/OTPInput.web.test.tsx
```diff
@@ -137,6 +138,28 @@ describe('<OTPInput />', () => {
...
+    const allInputs = getAllByLabelText(/character/);
+    await user.type(allInputs[0], otp);
+    await user.click(allInputs[5]);
+    expect(allInputs[5]).toHaveFocus();
+    await user.keyboard('{Backspace}');
+    expect(allInputs[5]).toHaveValue('');
+    expect(allInputs[4]).toHaveFocus();
+    await user.keyboard('{Delete}');
+    expect(allInputs[4]).toHaveValue('');
+    expect(allInputs[3]).toHaveFocus();
```
divyanshu013, line 160: Shouldn't <kbd>Del</kbd> key delete the character in front of cursor 🤔

chaitanyadeorukhkar: The input element only has 1 character at a time so deleting or backspacing should remove the character within that input itself, right?

divyanshu013: I think with <kbd>Del</kbd> it should delete the character in the next input box if cursor is after current number, will need to check some existing otp inputs in the wild. Can do a followup PR if needed.

chaitanyadeorukhkar: Yeah, lets discuss in the standup. The behaviour might be a bit weird if the next input is empty but the one after that has some value. Also how do we handle focus here? With backspace we move the focus.

PR #744, opened 2022-10-13
fix: use the correct maxWidth for OTPInput

chaitanyadeorukhkar:


packages/blade/src/components/Input/OTPInput/OTPInput.tsx
```diff
@@ -210,9 +212,9 @@ const OTPInput = ({
       inputs.push(
         <Box
           flex={1}
-          paddingLeft={index == 0 ? 'spacing.0' : 'spacing.3'}
+          marginLeft={index == 0 ? 'spacing.0' : 'spacing.3'}
           key={`${inputId}-${index}`}
-          maxWidth={100} // TODO: Get the correct value for this from design
+          maxWidth={platform === 'onDesktop' ? 36 : 40} // TODO: use size tokens
```
saurabhdaware, line 217: are we doing this TODO in followup PR?

chaitanyadeorukhkar: No, keeping it there cuz we need to add size tokens to Blade which we don't have right now. So across our components we've hardcoded sizes for now😢

divyanshu013: Can create an issue for tracking this.

divyanshu013 (approved):
Looks good to me.

saurabhdaware (approved):
🚢 🇮🇹

Issue #737, opened 2022-10-07
Type="button" prop not present for Link button variant


PranjalMaithani:
When using Link component with button variant, I cannot pass `type="button"` to it, unlike the Button component. This is causing unexpected behaviour when using Link button variant inside forms.

divyanshu013:
Can you share a sample code snippet with expected and current behavior please?

PranjalMaithani:
Sample snippet
```jsx
 const handleFormSubmit = (e) => {
    e.preventDefault();
    console.log('Handle Submit called');
  };

  const handleChangePhoneNumber = () => {
    console.log('change phone number called');
  };

  return (
    <form onSubmit={handleFormSubmit}>
      <Link variant="button" onClick={handleChangePhoneNumber}>
        Change phone number
      </Link>
      <div style={{ display: 'block' }}>
        <label for="otp">OTP</label>
        <input name="otp" />
      </div>
      <Button type="submit">SUBMIT</Button>
    </form>
  );
```

This has just console logs but in the actual scenario the Change Phone number CTA is taking the user to a different page

https://user-images.githubusercontent.com/75153839/195780073-33aadc3a-cef0-4cb7-bb4d-bbe61020af89.mov

Have raised a [PR here](https://github.com/razorpay/blade/pull/746)

PranjalMaithani:
Closing this as the PR for this has been merged

PR #746, opened 2022-10-13
fix(link): add type prop for button variant

PranjalMaithani:
Fixes this issue: https://github.com/razorpay/blade/issues/737

Currently the button variant for the `Link` component does not have the `type` prop. This is causing unexpected behavior in forms where buttons not intended to submit the form should have `type='button'`

Screenshot 2022-10-13 at 7 01 17 PM


packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -55,6 +55,7 @@ type BaseLinkAnchorVariantProps = BaseLinkPropsWithOrWithoutIcon & {
   target?: string;
   rel?: string;
   isDisabled?: undefined;
+  type?: never;
```
kamleshchandnani, line 58: I don't think we shall expose this prop. Since Link buttons ideally will only be used as href or `type=button` we can just pass it directly when we are rendering the component inside Link.

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -255,6 +258,7 @@ const BaseLink = ({
       rel={rel ?? defaultRel}
       onClick={handleOnClick}
       disabled={disabled}
+      type={type}
```
kamleshchandnani, line 261: probably we can check if it's a link or a button and basis that just set this internally here?

PranjalMaithani: Do you mean checking if it's a link and not passing it in that case? 
A developer can pass these possible values, we can't infer the value in that case.
```ts
```

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
PranjalMaithani: A developer might want to use any of these values for type though
```ts
```

Have kept it as `never` for the anchor variant since type shouldn't be passed to anchor tag

chaitanyadeorukhkar: +1 to @kamleshchandnani. 

@PranjalMaithani where do you forsee a Link that is semantically a button to be used as `reset` or `submit`? Imo this exposes us to a risk of improper usage of the Link component

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
chaitanyadeorukhkar: Lets check if the consumer is intending to use it as a `button` variant and add `type='button'`. This will protect us from consumers passing `reset` or `submit`?

```tsx
      type={variant==='button'? 'button' : undefined}
```

PranjalMaithani: If we want to omit `reset` and `submit` for Link button variant, should I go ahead with making these changes to set `type='button'` based on variant?

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
PranjalMaithani: Keeping it similar to Button props. Can omit `reset` and `submit` if they shouldn't be an option

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
kamleshchandnani: yes

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -143,6 +144,7 @@ const getProps = ({
   isVisited: boolean;
   target: BaseLinkProps['target'];
   hasIcon: boolean;
+  type?: 'button';
```
chaitanyadeorukhkar, line 147: is this needed? Don't see it being used anywhere @PranjalMaithani

chaitanyadeorukhkar (approved):
Thanks for the contribution @PranjalMaithani  🚀

PR #757, opened 2022-10-18
feat: Release blade on NPM registry

saurabhdaware:
## Description

Releasing blade on NPM registry so that it can be used with codesandbox, stackblitz, etc and installation gets easier for people outside of Razorpay.

## Approach

- Letting the changeset job take care of releasing and publishing to github package registry
- Created a `publish-to-npm.js` script that 
  - creates `.npmrc` with npm registry config
  - Runs `npm publish`
  - Deletes the temporary `.npmrc`

### Alternate Approach I had considered

Create new workflow file with `on: release` trigger and run `npm publish` from there. 

Dropped this because it might get triggered unnecessarily on release of other packages and fail. The current approach seems more straightforward to me

### Why not use `changeset publish` to publish over NPM as well?

Changeset in itself doesn't give option to publish over 2 registries. It just looks for existing `.npmrc` and publishes according to that.

We can run `changeset publish`, Create new temporary `.npmrc`, run `changeset publish` again. But changeset publish does a lot more than npm publish. It creates tags, releases, etc. It also tries to publish all the packages.

So there are chances of certain edge cases like-
- Will it re-create the tags and releases?
- We might end up publishing some package to NPM which we're not supposed to publish. E.g. in current setup it could've published blade-old as well.

We can move to using just `changeset publish` when we deprecate github package registry and use only NPM. But for now it seems a bit too complicated to me.

### Why not use `--registry` CLI option instead of creating temporary `.npmrc`

The `--registry` option in `npm publish` CLI doesn't give us any way to authenticate user with token. It only works for already logged in users with `npm login` which is not the case with CIs. 

We can alternatively use `npm set` and set token from CIs command-line but that too would require us to create some script because we can't directly set token over the entire command as `yarn release` initially publishes over github package registry.

TLDR: it's just hacky in all the directions

### Why not drop GitHub package registry and move to NPM

Answered here - https://github.com/razorpay/blade/pull/757#issuecomment-1284896633

## Testing

I had set to release a version on `--next` tag on push of this PR and I am able to see those releases on NPM.

Release job: https://github.com/razorpay/blade/actions/runs/3286795819/jobs/5415279896

This confirms the flow, tokens, publish script, etc are working fine.

We can merge this PR with changeset and check if it's able to make a release.

image


package.json
```diff
@@ -20,6 +20,7 @@
     "tsc:blade": "yarn workspace @razorpay/blade typecheck",
     "lint": "eslint --ext js,jsx,ts,tsx packages",
     "release": "changeset publish",
+    "release-npm": "lerna run --scope @razorpay/blade npm-publish",
```
kamleshchandnani, line 23: can't we use changeset publish to publish to npm? or that is the iteration you'll do after testing?

packages/blade/package.json
```diff
@@ -66,7 +66,8 @@
     "start:native": "run-p react-native:storybook:start react-native:storybook:android react-native:storybook:ios",
     "start:web": "yarn react",
     "start:all": "run-p start:native start:web",
-    "chromatic": "npx chromatic"
+    "chromatic": "npx chromatic",
+    "npm-publish": "npm publish --tag next --\"@razorpay:registry=https://registry.npmjs.org/\""
```
kamleshchandnani, line 70: okay since you're explicitly mentioning this here then the `repository` field from package.json is not respected

package.json
saurabhdaware: Can try it once after testing but `changeset publish` command is very limited so it doesn't support specifying registry with flag, dry runs, etc. So we might have to create a new npmrc and rename/copy-paste it which might get a bit tricky. Eventually I can make this command part of `release` command itself so we won't be calling this separately

saurabhdaware:
> we essentially don't need that. we can remove publishing to GitHub package registry once we start publishing to global npm registry.

Yupp so this, we can't do. Unless we move all the packages inside `@razorpay` scope to NPM registry including Universe.

When people install packages, they have this config in their `.npmrc`
```
@razorpay:registry=https://npm.pkg.github.com/
//npm.pkg.github.com/:always-auth=true
//npm.pkg.github.com/:_authToken=${GITHUB_ACCESS_TOKEN}
```

This just tries to fetch all `@razorpay/*` packages from github package registry including Universe, blade, etc.

NPM doesn't allow us to define 2 registries for one scope. Which means we can't do something like this - 
```
@razorpay:registry=https://npm.pkg.github.com/
@razorpay/blade:registry=https://registry.npmjs.org/

//npm.pkg.github.com/:always-auth=true
//npm.pkg.github.com/:_authToken=${GITHUB_ACCESS_TOKEN}
//registry.npmjs.org/:always-auth=true
//registry.npmjs.org/:_authToken=${NPM_TOKEN}
```

This doesn't work and there is no other way to do this.

kamleshchandnani:
> > we essentially don't need that. we can remove publishing to GitHub package registry once we start publishing to global npm registry.
> 
> Yupp so this, we can't do. Unless we move all the packages inside `@razorpay` scope to NPM registry including Universe.
> 
> When people install packages, they have this config in their `.npmrc`
> 
> ```
> @razorpay:registry=https://npm.pkg.github.com/
> //npm.pkg.github.com/:always-auth=true
> //npm.pkg.github.com/:_authToken=${GITHUB_ACCESS_TOKEN}
> ```
> 
> This just tries to fetch all `@razorpay/*` packages from github package registry including Universe, blade, etc.
> 
> NPM doesn't allow us to define 2 registries for one scope. Which means we can't do something like this -
> 
> ```
> @razorpay:registry=https://npm.pkg.github.com/
> @razorpay/blade:registry=https://registry.npmjs.org/
> 
> //npm.pkg.github.com/:always-auth=true
> //npm.pkg.github.com/:_authToken=${GITHUB_ACCESS_TOKEN}
> //registry.npmjs.org/:always-auth=true
> //registry.npmjs.org/:_authToken=${NPM_TOKEN}
> ```
> 
> This doesn't work and there is no other way to do this.
> 
> So internal users will have to continue using package from GitHub package registry atleast until we move out to NPM completely (which we should at some point)

Oh damn I completely forgot about that multi registry nuance. and now i recall when we had started building the new blade version while setting the release workflow this was the exact reason i had to park it to move ahead 😭

kamleshchandnani (approved):

packages/blade/scripts/publish-to-npm.js
```diff
@@ -0,0 +1,24 @@
+const fs = require('fs');
```
divyanshu013, line 1: Suggestion: you could copy some of the background from PR description to this file for future context.

divyanshu013 (approved):
Looks good to me. Thanks for the detailed PR. Now all we need next is experimental releases from PRs 🙏

divyanshu013 (approved):
:shipit:

saurabhdaware:
Lets test on prod :shipit:

Issue #603, opened 2022-08-03
Remove `gap` prop from `Box` component and its usage across components & stories


chaitanyadeorukhkar:
Since there is no support for `gap` on React Native yet, adding support for `gap` on `Box` would lead to UI breakage on native components. We have already used `gap` with `Box` at multiple places and we need to remove those instances and replace them with other appropriate spacings.

chaitanyadeorukhkar:
@anuraghazra do you want to take this up since it could have some impact on Checkbox?

anuraghazra:
Yes I'll fix this.

anuraghazra:
Fixed in #758

PR #758, opened 2022-10-18
refactor(blade): Checkbox/Radio changes for X

anuraghazra:
This PR includes: 

- Checkbox
   - adds "small" size
   - spacing changes
   - svg icon changes
   - hover interaction changes
 
- Radio
   - adds "small" size
   - spacing changes
   - hover interaction changes
   

Radio small variant: 
image

Checkbox small variant: 
image


packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
```diff
@@ -8,44 +8,108 @@ import { useTheme } from '~components/BladeProvider';
...
+        strokeWidth="0.5"
+        strokeLinecap="round"
+        strokeLinejoin="round"
       />
     </Svg>
   );
 };
 
-const IndeterminateIcon = ({ color }: { color: string }) => {
+const CheckedIconMedium = ({ color }: { color: string }) => {
```
divyanshu013, line 28: Why do we need different icons for different sizes? Won't the same svg with different height / width work 🤔

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
```diff
@@ -8,44 +8,108 @@ import { useTheme } from '~components/BladeProvider';
...
   return (
-    <Svg width="16px" height="16px" viewBox="0 0 16 16" fill="none">
+    <Svg width="12" height="12" viewBox="0 0 12 12" fill="none">
```
divyanshu013, line 30: We're removing `px` now in height / width?

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
```diff
@@ -45,27 +13,30 @@ const getCheckboxIconWrapperStyles = ({
...
+    height: makeSpace(checkboxSizes.icon[size].height),
+    borderWidth: makeBorderSize(theme.border.width.thick),
     borderStyle: 'solid',
-    margin: makeSpace(theme.spacing[0]),
+    margin: makeSpace(theme.spacing[1]),
     borderRadius: makeSize(theme.border.radius.small),
-    marginRight: makeSpace(theme.spacing[3]),
-    backgroundColor: getIn(theme, backgroundColor),
-    borderColor: getIn(theme, borderColor),
+    // marginRight: makeSpace(theme.spacing[3]),
```
divyanshu013, line 37:
```suggestion
```

packages/blade/src/tokens/global/border.ts
```diff
@@ -37,6 +39,7 @@ const border: Border = {
   width: {
     none: 0,
     thin: 1,
+    thick: 1.5,
```
divyanshu013, line 42: I'm wondering what will we call > 1.5 😹

divyanshu013 (commented):
Some comments.
- [x] Add changeset

packages/blade/src/tokens/global/border.ts
anuraghazra: >1.5 = "thicc"
>2.0 = "thiccc"
>3.0 = "thicccc"

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
anuraghazra: Good catch! I'll add it.

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
anuraghazra: Yeah i could, in fact I did that earlier. 
But with the width/height we also need to change the definition of the path, `<Path d="" />`. 
So i just took the simplified approach.

If this looks good, I'm fine with it: 

```ts
const IndeterminateIcon = ({ color, size }: { color: string; size: 'small' | 'medium' }) => {
  const path = {
    medium:
      'M2 6C2 5.72386 2.22386 5.5 2.5 5.5H9.5C9.77614 5.5 10 5.72386 10 6C10 6.27614 9.77614 6.5 9.5 6.5H2.5C2.22386 6.5 2 6.27614 2 6Z',
    small:
      'M1.3335 3.99984C1.3335 3.81574 1.48273 3.6665 1.66683 3.6665H6.3335C6.51759 3.6665 6.66683 3.81574 6.66683 3.99984C6.66683 4.18393 6.51759 4.33317 6.3335 4.33317H1.66683C1.48273 4.33317 1.3335 4.18393 1.3335 3.99984Z',
  };
  const dims = {
    small: {
      width: 8,
      height: 8,
      viewBox: '0 0 8 8',
    },
    medium: {
      width: 12,
      height: 12,
      viewBox: '0 0 12 12',
    },
  };

  const width = `${dims[size].width}px`;
  const height = `${dims[size].height}px`;
  const viewBox = dims[size].viewBox;
  const d = path[size];

```

divyanshu013: I see, I thought the same path `d` could work for all sizes but I guess not? In that case both the current implementation and the one you shared above seem fine to me.

anuraghazra: Yes. 

but considering that there will be another variant "large" later on, which approach do you think would be better? 🤔

divyanshu013: I guess we would end up anyway bundling all these icons so the one you shared here would be more preferred imo.

anuraghazra: Done.

divyanshu013 (approved):
Looks good to me, needs changeset.

packages/blade/src/components/Checkbox/Checkbox.tsx
```diff
@@ -89,6 +89,12 @@ type CheckboxProps = {
    * and `invalid` attribute will be added
    */
   validationState?: 'error' | 'none';
+  /**
+   * Size of the radios
```
chaitanyadeorukhkar, line 93:
```suggestion
   * Size of the checkbox
```

packages/blade/src/components/Checkbox/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -69,6 +72,12 @@ type CheckboxGroupProps = {
    * (Useful for form submission).
    */
   name?: string;
+  /**
+   * Size of the radios
```
chaitanyadeorukhkar, line 76:
```suggestion
   * Size of the checkbox
```

packages/blade/src/components/Checkbox/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -112,7 +126,7 @@ const CheckboxGroup = ({
           {label}
         </FormLabel>
         <Box>
-          <Box display="flex" flexDirection="column" gap={2}>
+          <Box display="flex" flexDirection="column" gap={gap}>
```
chaitanyadeorukhkar, line 129: `gap` doesn't work on react native. Can we remove gap's usage while we're working on this PR? I'm guessing this renders incorrectly on react native because of this @anuraghazra

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
```diff
@@ -8,27 +8,71 @@ import { useTheme } from '~components/BladeProvider';
 import Svg, { Path } from '~components/Icons/_Svg';
 import { getIn } from '~utils';
 
-const CheckedIcon = ({ color }: { color: string }) => {
+const svgSize = {
```
chaitanyadeorukhkar, line 11: **nit (but also lets do this 🙈 ):** Imo it would be better and more readable if we have `CheckedIconSmall`, `CheckedIconMedium`, etc. Having this logic for width, height, viewBox, d makes it seem too complicated. Lets treat it as an icon for small checkbox and an icon for medium checkbox?

packages/blade/src/components/Form/Selector/SelectorTitle.tsx
```diff
@@ -1,24 +1,31 @@
...
-        isDisabled ? 'surface.text.placeholder.lowContrast' : 'surface.text.subtle.lowContrast'
-      }
-      fontSize={100}
-    >
-      {children}
-    </BaseText>
+    <>
+      <Box marginLeft="spacing.2" />
+      <BaseText
+        lineHeight={isSmall ? 's' : 'l'}
```
chaitanyadeorukhkar, line 18: We're also going to have `large` size now. Better to maintain a size to lineHeight, fontSize mapping here?

packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -51,6 +63,8 @@ const Radio = ({ value, children, helpText, isDisabled }: RadioProps): React.Rea
   const name = groupProps?.name;
   const showHelpText = !hasError && helpText;
   const isReactNative = getPlatformType() === 'react-native';
+  const _size = groupProps.size ?? size;
+  const isSmall = _size === 'small';
```
chaitanyadeorukhkar, line 67: We're going to have `large` as well now. This logic will need to change and a boolean won't be enough :(

Upto you if you want to prepare for large size in this PR or the following PR for large size

packages/blade/src/components/Radio/RadioGroup/RadioGroup.tsx
```diff
@@ -94,12 +103,16 @@ const RadioGroup = ({
...
+    size,
   });
 
+  const { theme } = useTheme();
+  const { matchedDeviceType } = useBreakpoint({ breakpoints: theme.breakpoints });
   const showError = validationState === 'error' && errorText;
   const showHelpText = !showError && helpText;
   const accessibilityText = `,${showError ? errorText : ''} ${showHelpText ? helpText : ''}`;
   const isReactNative = getPlatformType() === 'react-native';
+  const gap = radioSizes.group.gap[size][matchedDeviceType];
```
chaitanyadeorukhkar, line 115: Lets remove `gap` please while we're working on this PR otherwise we won't revisit this and it will remain broken on react native

packages/blade/src/components/Radio/RadioIcon/RadioIcon.tsx
```diff
@@ -1,55 +1,55 @@
...
 
 export type RadioIconProps = {
   isDisabled?: boolean;
   isNegative?: boolean;
   isChecked?: boolean;
-};
+} & Required<Pick<RadioProps, 'size'>>;
+
+const CheckedIcon = ({ color, size }: { color: string; size: RadioIconProps['size'] }) => {
+  const width = radioSizes.icon[size].width;
```
chaitanyadeorukhkar, line 20: This logic seems too complex imo. Easier to just have multiple sets of icons as different components for different sizes

chaitanyadeorukhkar (changes requested):

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
chaitanyadeorukhkar: I think the d changes if you try to copy individual svg from figma for each size

Can we just use the same `d` and only change width & height. This is how svg solves such use-cases anyway. Viewbox will remain same but width & height will change. @anuraghazra can you try this once? It will simplify this logic a lot especially because we have large as well now

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
chaitanyadeorukhkar: Re: Lets do this instead

https://github.com/razorpay/blade/pull/758#discussion_r1004392364

packages/blade/src/components/Radio/RadioIcon/RadioIcon.tsx
chaitanyadeorukhkar: Lets do this instead:

https://github.com/razorpay/blade/pull/758#discussion_r1004392364

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
anuraghazra: We were discussing this. 

https://github.com/razorpay/blade/pull/758#discussion_r1004031973

We did the same way earlier.

packages/blade/src/components/Radio/Radio.tsx
anuraghazra: I'll do it on the next PR. and refactor it accordingly.

packages/blade/src/components/Radio/RadioGroup/RadioGroup.tsx
anuraghazra: What would be the equivalent of the gap here? 

I mean how should we approach it so that in react-native the gaps can be in between. Does the owl selector work in RN 🤔

packages/blade/src/components/Radio/RadioIcon/RadioIcon.tsx
anuraghazra: Doing it this way because the `getRadioIconWrapperStyles`'s width & height also needs to match the radio's icon sizing. 

Screenshot 2022-10-25 at 5 35 57 PM

Thus colocated, it's easier to just make the change in one place instead of jumping files to find where I need to make the sizing changes.

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
divyanshu013: This is exactly what I meant in the OP, using the same `d` 😆

anuraghazra: Yeah so i was also changing the "viewbox" to "0 0 8 8" so thats why needed to change the "d" also

packages/blade/src/components/Form/Selector/SelectorTitle.tsx
anuraghazra: Done.

packages/blade/src/components/Radio/RadioGroup/RadioGroup.tsx
chaitanyadeorukhkar: No, we'll have to stick to using margins :(

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
```diff
@@ -8,27 +8,52 @@ import { useTheme } from '~components/BladeProvider';
...
+  },
+};
+
+const CheckedIcon = ({ color, size }: { color: string; size: 'small' | 'medium' }) => {
+  const width = `${svgSize[size].width}px`;
+  const height = `${svgSize[size].height}px`;
+
   return (
-    <Svg width="16px" height="16px" viewBox="0 0 16 16" fill="none">
+    <Svg width={width} height={height} viewBox="0 0 8 8" fill="none">
```
chaitanyadeorukhkar, line 27: **nit**: why not use `viewBox="0 0 24 24"`as the base since that's what we're doing for rest of the icons on blade too?

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
```diff
@@ -8,27 +8,52 @@ import { useTheme } from '~components/BladeProvider';
...
+    height: 8,
+  },
+  medium: {
+    width: 12,
+    height: 12,
+  },
+};
+
+const CheckedIcon = ({ color, size }: { color: string; size: 'small' | 'medium' }) => {
+  const width = `${svgSize[size].width}px`;
```
chaitanyadeorukhkar, line 23:
```suggestion
  const width =  makeSize(svgSize[size].width);
```

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
```diff
@@ -8,27 +8,52 @@ import { useTheme } from '~components/BladeProvider';
...
+  const height = `${svgSize[size].height}px`;
```
chaitanyadeorukhkar, line 24:
```suggestion
  const height = makeSize(svgSize[size].height);
```

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIcon.tsx
anuraghazra: I'll confirm first if the Check icon is same as the svg in checkbox. 

I remember saurav changed the svgs

anuraghazra: [Confirmed](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1666863800967479?thread_ts=1666856828.448019&cid=G01B3LQ9H0W) with saurav. 

> You are not suppose to use icons inside checkboxes. You should export the SVG directly from the [Base DSL file](https://www.figma.com/file/LV9kjgzhkfsktvbRgTuGJ3/Blade---Base-DSL?node-id=13703%3A67331).

chaitanyadeorukhkar (approved):

chaitanyadeorukhkar:
Can we mark https://github.com/razorpay/blade/issues/603 as fixed after this @anuraghazra ?

PR #769, opened 2022-10-21
docs: update the api decisions for `BaseInput`

### LINKED ISSUE #736: Support for `keyboardType = 'numeric'` on `BaseInput` & `OTPInput`
The `BaseInput` [decision doc mentions `numeric`](https://github.com/razorpay/blade/blame/master/packages/blade/src/components/Input/BaseInput/_decisions/_decisions.md#L49) as a valid value for `keyboardType` but the `BaseInput` [implementation doesn't support `numeric`](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Input/BaseInput/BaseInput.tsx#L124) as a valid value. Is this intentional? Can we add support for `numeric`? Due to this constraint, we've added default `keyboardType` as `decimal` instead of `numeric` for `OTPInput`. After adding support for `numeric`, we should also change the default for `OTPInput` as `numeric`

kamleshchandnani:
fixes #736


chaitanyadeorukhkar (approved):

divyanshu013 (approved):
Looks good to me. Suggestion: we can add in the API doc why we don't recommend `numeric` for any future confusions.

kamleshchandnani:
> Looks good to me. Suggestion: we can add in the API doc why we don't recommend `numeric` for any future confusions.

added in [a5d0308](https://github.com/razorpay/blade/pull/769/commits/a5d0308d56e1c740b367cdb417ad27433c50e631)

PR #773, opened 2022-10-21
test(Path): fix tests

### LINKED ISSUE #693: Fix `Path` tests
Steps to replicate:
1. Run `yarn test:react --watch`

image

shivam1646:
resolves #693

### DIFF
--- a/packages/blade/src/components/Icons/_Svg/Path/__tests__/Path.web.test.tsx
+++ b/packages/blade/src/components/Icons/_Svg/Path/__tests__/Path.web.test.tsx
@@ -17,7 +17,6 @@ describe('<Path />', () => {
           strokeLinecap="square"
           strokeLinejoin="bevel"
           strokeWidth="2px"
-          is="x3d"
         />
       </Svg>,
     );
@@ -25,8 +24,12 @@ describe('<Path />', () => {
   });
 
   it('should not accept dashed props', () => {
-    // @ts-expect-error no dashed props are accepted
-    const { container } = renderWithTheme(<Path fill-opacity={0} d="" is="x3d" />);
+    const { container } = renderWithTheme(
+      <Svg viewBox="0 0 6 6" height="40px" width="40px" fill="none">
+        {/* @ts-expect-error no dashed props are accepted */}
+        <Path fill-opacity={0} d="" />
+      </Svg>,
+    );
     expect(container).toMatchSnapshot();
   });
 });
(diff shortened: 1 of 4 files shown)


packages/blade/src/components/Icons/_Svg/Path/__tests__/Path.web.test.tsx
```diff
@@ -25,7 +26,7 @@ describe('<Path />', () => {
 
   it('should not accept dashed props', () => {
     // @ts-expect-error no dashed props are accepted
-    const { container } = renderWithTheme(<Path fill-opacity={0} d="" />);
+    const { container } = renderWithTheme(<Path fill-opacity={0} d="" is="x3d" />);
```
divyanshu013, line 29: How does this resolve the issue?

packages/blade/src/components/Icons/_Svg/Path/__tests__/__snapshots__/Path.web.test.tsx.snap
```diff
@@ -18,16 +19,17 @@ exports[`<Path /> should render html path component 1`] = `
...
+      fillOpacity="0.5"
+      fillRule="nonzero"
+      is="x3d"
       stroke="#1212"
-      stroke-linecap="square"
-      stroke-linejoin="bevel"
-      stroke-width="2px"
+      strokeLinecap="square"
+      strokeLinejoin="bevel"
+      strokeWidth="2px"
```
divyanshu013, line 32: This change would actually break the svgs because camel cased props work only in React, not in the actual browser.

divyanshu013 (changes requested):
I think this change would end up breaking things instead because of how it affects the props. The issue is probably something related to how jest is understanding `path`.

packages/blade/src/components/Icons/_Svg/Path/__tests__/Path.web.test.tsx
shivam1646: So in the tests, the error indicates that browser does not recognise `path` as a valid dom tag. Using `is="x3d"` in the tests, tells browser to recognise this element as a valid SVG tag.

packages/blade/src/components/Icons/_Svg/Path/__tests__/__snapshots__/Path.web.test.tsx.snap
shivam1646: This change affects only the test file - `Path.web.test.tsx` where the error is thrown.

shivam1646:
@divyanshu013 
I misunderstood the issue earlier.

Actually, one of the test cases `'should not accept dashed props'` in `Path.web.test.tsx` uses `path` tag without a parent `svg` tag which leads to browser not able to recongnize the `path` tag.

Added the fix accordingly and reverted the previous changes. Sorry for the confusion.

divyanshu013:
Awesome, this seems like the correct fix 🎉

divyanshu013 (approved):
Looks good to me.

chaitanyadeorukhkar (approved):

PR #777, opened 2022-10-22
fix: Icon prop should always use valid IconComponent

### LINKED ISSUE #680: Add strict check for ensure `icon` prop is not being misused
Add strict checks to ensure `icon` props in components are being used as intended. 

✅ **Intended usage:**
```jsx
import { Button, CreditCardIcon } from '@razorpay/components';

<Button icon={CreditCardIcon} />
```

❌ **Misuse:**
```jsx
import { Button, CreditCardIcon } from '@razorpay/components';

<Button icon={() => <CreditCardIcon size="small" color="feedback.positive.icon.lowContrast"/>} />
```

shivam1646:
resolves #680

### DIFF
--- a/packages/blade/src/components/Button/BaseButton/BaseButton.tsx
+++ b/packages/blade/src/components/Button/BaseButton/BaseButton.tsx
@@ -28,6 +28,7 @@ import { BaseSpinner } from '~components/Spinner/BaseSpinner';
 import Box from '~components/Box';
 import type { DotNotationSpacingStringToken } from '~src/_helpers/types';
 import type { BaseTextProps } from '~components/Typography/BaseText/types';
+import iconMap from '~components/Icons/iconMap';
 
 type BaseButtonCommonProps = {
   size?: 'xsmall' | 'small' | 'medium' | 'large';
@@ -284,6 +285,12 @@ const BaseButton = ({
   if (!Icon && !children?.trim()) {
     throw new Error(
       `[Blade: BaseButton]: At least one of icon or text is required to render a button.`,
+    );
+  }
+
+  if (Icon && !Object.values(iconMap).includes(Icon)) {
+    throw new Error(
+      `[Blade: BaseButton]: Invalid usage of icon prop. Please use one of the icon components provided by Blade.`,
     );
   }
 
--- a/packages/blade/src/components/Button/BaseButton/BaseButton.tsx
+++ b/packages/blade/src/components/Button/BaseButton/BaseButton.tsx
@@ -28,7 +28,7 @@ import { BaseSpinner } from '~components/Spinner/BaseSpinner';
 import Box from '~components/Box';
 import type { DotNotationSpacingStringToken } from '~src/_helpers/types';
 import type { BaseTextProps } from '~components/Typography/BaseText/types';
-import iconMap from '~components/Icons/iconMap';
+import { iconMapKeys } from '~components/Icons/iconMap';
 
 type BaseButtonCommonProps = {
   size?: 'xsmall' | 'small' | 'medium' | 'large';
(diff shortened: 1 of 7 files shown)


packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -28,6 +28,7 @@ import { BaseSpinner } from '~components/Spinner/BaseSpinner';
 import Box from '~components/Box';
 import type { DotNotationSpacingStringToken } from '~src/_helpers/types';
 import type { BaseTextProps } from '~components/Typography/BaseText/types';
+import iconMap from '~components/Icons/iconMap';
```
divyanshu013, line 31: Although this might make the check work it isn't an ideal solution because it would bloat the bundle size with all the imported icons in `iconMap.ts` module.

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -15,6 +15,7 @@ import { makeAccessible, getIn } from '~utils';
 import type { LinkActionStates } from '~tokens/theme/theme';
 import type { DurationString, EasingString } from '~tokens/global/motion';
 import type { BaseTextProps } from '~components/Typography/BaseText/types';
+import iconMap from '~components/Icons/iconMap';
```
divyanshu013, line 18: Similarly here.

divyanshu013 (changes requested):
This solution will bloat bundle size unfortunately

shivam1646:
I have updated the logic such that it needs only the icon keys and not the icon component. This may help keep the bundle size in check while providing similar validation.

@divyanshu013 Could you please check once?

packages/blade/src/components/Icons/iconMap.ts
```diff
@@ -74,4 +74,6 @@ const iconMap: Record<string, IconComponent> = {
   TrendingDownIcon: TrendingDownIconComponent,
 };
 
+export const iconMapKeys = Object.keys(iconMap);
```
anuraghazra, line 77: Unfortunately this will still effect the bundle. :( 
Since we are importing all the icons here. 

Possible solution would be to maintain an array of all the icon names as strings only.

divyanshu013: Right, that would be hard to keep in sync though.

divyanshu013 (commented):
We might need to explore a different way of doing this 🤔

anuraghazra:
@divyanshu013 @Chaitanya-019 is there a requirement for runtime checking? 

If not, we can prevent from TS side too. 

```tsx
export type IconComponent = React.ComponentType<IconProps> & { displayName: `BladeIcon_${string}` };

// in icon file
ArrowUpRightIcon.displayName = 'BladeIcon_ArrowUpRightIcon';

```

image

divyanshu013:
I think TS check is a good start but the error message should be apt. Also I feel explicitly maintaining `displayName` for all might be an overhead but open to other opinions.

anuraghazra:
Wait.. actually the same approach can be done in runtime too. Can't believe I'm solving in type level first 😆 

```ts
// icon file
ArrowDownIcon.displayName = 'BladeIcon_ArrowDownIcon';


// runtime check
  if (icon && (typeof icon !== 'function' || !icon.displayName?.startsWith('BladeIcon_'))) {
    throw new Error(
      `[Blade: IconButton]: Invalid usage of icon prop. Please use one of the icon components provided by Blade.`,
    );
  }
```

anuraghazra:
> Also I feel explicitly maintaining displayName for all might be an overhead but open to other opinions.

Yeah i guess. but i think the icon names wont change often and whenever someone adds icon we just have to make sure the display name is there. So an okay tradeoff IMO.

divyanshu013:
That's true but seems like an overhead to me, extra maintenance for little gains 😅

shivam1646:
@divyanshu013 @anuraghazra 
Right now each icon component follows the naming convention of having `Icon` as a suffix. Can this information be used for validation?

divyanshu013:
Not really, people can pass any component which has an `Icon` suffix and that may break the validation

chaitanyadeorukhkar:
Should we close this PR and continue this conversation in an issue?

@divyanshu013 @shivam1646

shivam1646:
Sure, this can be closed.

divyanshu013:
Closing this till we can identify a better solution.

PR #780, opened 2022-10-25
fix: show internal stories in docs

### LINKED ISSUE #626: Internal components story doesn't render in docs section
[Example](https://master--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/docs/components-button-basebutton-internal--base-button) for BaseButton story

shivam1646:
resolves #626 

- Updates the stylesheet for `internal components` in the preview `iframe`( `preview.js` is used to render stories ) as well based on the global variable `showInternalComponents` value.

### DIFF
--- a/packages/blade/.storybook/react/preview.js
+++ b/packages/blade/.storybook/react/preview.js
@@ -1,4 +1,4 @@
-import { theme } from './manager';
+import { theme, hiddenStoryStyle } from './manager';
 import { global } from '@storybook/design-system';
 import { BladeProvider } from '../../src/components/BladeProvider';
 import { paymentTheme, bankingTheme } from '../../src/tokens/theme';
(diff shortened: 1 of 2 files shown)


chaitanyadeorukhkar:
@anuraghazra can you take up L1?

anuraghazra (approved):
Seems to be working fine. LGTM 

Thanks @shivam1646

packages/blade/.storybook/react/preview.js
```diff
@@ -43,6 +43,7 @@ export const parameters = {
 
 export const decorators = [
   (Story, context) => {
+    hiddenStoryStyle.disabled = context.globals.showInternalComponents
```
divyanshu013, line 46: Instead of mutating this imported module here my recommendation would be to instead expose a function from `manager` that toggles the flag for you. Keeps the logic contained within the `manager` module.

shivam1646: @divyanshu013 makes sense. Updated PR accordingly.

PR #781, opened 2022-10-25
fix: npm releases with changeset

anuraghazra:
Currently we are publishing to public npm registry on every commit to master. 
We only should publish it when we merge the builder version PRs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,3 +34,7 @@ jobs:
           publish: yarn release
           commit: 'build(blade): update version'
           title: 'build(blade): update version'
+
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -34,3 +34,7 @@ jobs:
           publish: yarn release
           commit: 'build(blade): update version'
           title: 'build(blade): update version'
+
```
chaitanyadeorukhkar, line 37: nit: remove new line?
```suggestion
```

chaitanyadeorukhkar (approved):

PR #788, opened 2022-10-27
docs(Indicator): add api decisions

divyanshu013:
Adds API design for `Indicator`
- [Formatted markdown](https://github.com/razorpay/blade/blob/div/indicator-api/packages/blade/src/components/Indicator/_decisions/decisions.md)


packages/blade/src/components/Indicator/_decisions/decisions.md
```diff
@@ -0,0 +1,111 @@
...
+Sample usecase:
+
+<img src="./with-label.png" width="380" alt="" />
+
+Sample usage:
+
+```jsx
+<Indicator
+  intent="positive"
+  label="Active | xyz@email.com"
```
anuraghazra, line 24: What was the conclusion on this label? Will it be just dot or label too?

packages/blade/src/components/Indicator/_decisions/decisions.md
```diff
@@ -0,0 +1,111 @@
...
+| ------------------ | --------------------------------------------------------- | ----------- | ---------------------------------------------- | -------- |
+| accessibilityLabel | `string`                                                  | `undefined` | Sets the aria-label for screen readers         | ♿️      |
+| label              | `string`                                                  | `undefined` | Shows a text label beside the circle indicator |          |
+| intent             | `positive`,`negative`, `notice`, `information`, `neutral` | `neutral`   | Sets the color tone of circle indicator        |          |
+| size               | `xsmall`,`small`, `medium`                                | `medium`    | Sets the size of circle indicator              |          |
+
+> **Note**
+>
+> ♿ `accessibilityLabel` is **required** when a `label` isn't provided and helps in setting `aria-label` for people relying on screen readers. Some people with color blindness can't distinguish between colors easily so passing a `label` / `accessibilityLabel` is crucial.
```
anuraghazra, line 42: 💯

packages/blade/src/components/Indicator/_decisions/decisions.md
```diff
@@ -0,0 +1,111 @@
...
+> **Note**
+>
+> At the moment we don't provide an out of box API for such composition. See the alternatives section for more info.
+
+## a11y
+
+Web:
+
+- The component accepts an `accessibilityLabel` prop to let users pass an `aria-label` used by screen readers. This is helpful in building UIs where an `Indicator` is used without a visible `label`.
+- If `label` is provided without any `accessibilityLabel` we use an `aria-labelledby` internally to let the screen readers use `label` for a11y purposes.
```
anuraghazra, line 70: > note: depending on the html structure we might not need to set arialabelledby if the content is a immediate children.

divyanshu013: I think it'll need to be sibling (flex siblings) but will keep this in mind

packages/blade/src/components/Indicator/_decisions/decisions.md
divyanshu013: Label would be needed for certain cases so we need both the dot and label

anuraghazra (approved):

packages/blade/src/components/Indicator/_decisions/decisions.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+### Indicator
+
+We'll expose an `Indicator` component with the following API:
+
+| Prop               | Type                                                      | Default     | Description                                    | Required |
+| ------------------ | --------------------------------------------------------- | ----------- | ---------------------------------------------- | -------- |
+| accessibilityLabel | `string`                                                  | `undefined` | Sets the aria-label for screen readers         | ♿️      |
+| label              | `string`                                                  | `undefined` | Shows a text label beside the circle indicator |          |
```
chaitanyadeorukhkar, line 36: Why use `label` as a prop and not `children` for rendering the label instead? That's what we're doing on Badge, Radio, Checkbox, etc. Label associated with the component is taken as `children`.

chaitanyadeorukhkar (commented):
Mostly lgtm. Just some doubts about what should the label prop be and if its font size changes as per indicator size

packages/blade/src/components/Indicator/_decisions/decisions.md
divyanshu013: Was thinking of using children for the composition API (in alternatives section) earlier. But we can use `children` too to match other components

divyanshu013: Updated in https://github.com/razorpay/blade/pull/788/commits/dcd11dae93bf81e674aa49356a68f824cb34b2c1

kamleshchandnani (approved):
LGTM 🚢

anshulsahni:
Looks Good

chaitanyadeorukhkar (approved):

PR #794, opened 2022-10-28
build: fix npm publish

chaitanyadeorukhkar:
We should check for a string `'true'` instead of a boolean `true`

Ref: https://github.com/changesets/action#:~:text=if%3A%20steps.changesets.outputs.published%20%3D%3D%20%27true%27


saurabhdaware (approved):

saurabhdaware (approved):
~Did you mean to add changeset?~

PR #801, opened 2022-10-31
chore: move pre commit checks to CI

chaitanyadeorukhkar:
**Changes:**
- Remove eslint, stylelint & tsc checks from pre-commit hook
- Remove conventional commit check from pre-commit hook
- Keep prettier --write in pre-commit hook
- Add eslint, stylelint, tsc checks to CI
- Run tsc:web & tsc:native in parallel on CI
- Run test:web & test:native in parallel on CI
- Run typescript checks before running tests on CI
- Add `pr-title-check` workflow that fails the PR check and adds a comment if the PR title doesn't follow conventional commit format. This PR check will pass and the comment will be automatically updated once the PR title is corrected.

**When PR title is incorrect:**
Screenshot 2022-10-31 at 4 30 43 PM

**When PR title is correct:**
Screenshot 2022-10-31 at 5 28 53 PM

**Note:**
- I tried deleting the PR Title check comment for when the title is correct using `delete: true` with https://github.com/marocchino/sticky-pull-request-comment but there seems to be some issue with the action. Cannot delete or hide a created comment. For time being, instead of deleting the comment we'll just update the comment to convey that the PR title is correct


.github/workflows/blade-pr-title-check.yml
```diff
@@ -0,0 +1,40 @@
+name: Blade PR Title Check
+
+on: 
+  pull_request:
+     types: [open, edited, synchronize]
```
chaitanyadeorukhkar, line 5: Need a separate workflow for this because the trigger types are different. If we keep the same workflow as `blade-validate`, rest of the validate jobs will get triggered even if the PR title is changed

.github/workflows/blade-pr-title-check.yml
```diff
@@ -0,0 +1,40 @@
...
+            
+            ```
+            ${{ steps.pr_title_check.outputs.error_message }}
+            ```
+      # Delete a previous comment when the issue has been resolved
+      - if: ${{ steps.pr_title_check.outputs.error_message == null }}
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:   
+          header: pr-title-comment
+          message: '✅ PR Title follows [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).'
```
chaitanyadeorukhkar, line 39: I'd prefer just doing `delete: true` but for some reason that's not working. Tried debugging for a while but using this as a workaround for now

.github/workflows/blade-pr-title-check.yml
```diff
@@ -0,0 +1,40 @@
...
+     types: [open, edited, synchronize]
+
+concurrency: # cancel previously running workflows when a new workflow is re-run
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  pr-title-check:
+    name: Check
+    runs-on: ubuntu-latest # nosemgrep: non-self-hosted-runner
```
divyanshu013, line 14: We're not running this on self hosted?

chaitanyadeorukhkar: No, I saw rest of our workflows aren't either. I can't recall why we made this decision though. @kamleshchandnani do you recall?

kamleshchandnani: nupe because this is public repo. I think we can't run self-hosted on public repo or they need to be configured. regardless i don't see advantages of why shall we run this on self-hosted

.github/workflows/blade-pr-title-check.yml
```diff
@@ -0,0 +1,40 @@
...
+    env:
+      GITHUB_TOKEN: ${{ secrets.CI_BOT_TOKEN }}
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        id: pr_title_check
+      - uses: marocchino/sticky-pull-request-comment@v2
+        # When the previous steps fails, the workflow would stop. By adding this
+        # condition you can continue the execution with the populated error message.
+        if: always() && (steps.pr_title_check.outputs.error_message != null)
+        with:
```
kamleshchandnani, line 24: can we also pass `recreate: true` so that if the PR title checks fails multiple times we don't end up with multiple comments?

.github/workflows/blade-pr-title-check.yml
```diff
@@ -0,0 +1,40 @@
...
+          header: pr-title-comment
+          message: |            
+            We follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/) and it looks like your PR Title doesn't fit the specification.
```
kamleshchandnani, line 27: add emoji?
```suggestion
            ❌ PR title doesn't follow [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).
```

.github/workflows/blade-pr-title-check.yml
kamleshchandnani, line 39:
```suggestion
          message: '✅ PR title follows [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/).'
```

.github/workflows/blade-pr-title-check.yml
```diff
@@ -0,0 +1,40 @@
+name: Blade PR Title Check
```
kamleshchandnani, line 1:
```suggestion
name: Blade PR Title Lint
```

.github/workflows/blade-pr-title-check.yml
```diff
@@ -0,0 +1,40 @@
...
+
```
kamleshchandnani, line 2: why not add this check inside `validate` itself? do we really need a separate workflow?

package.json
```diff
@@ -18,7 +18,7 @@
   },
   "scripts": {
     "tsc:blade": "yarn workspace @razorpay/blade typecheck",
-    "lint": "eslint --ext js,jsx,ts,tsx packages",
+    "lint": "eslint --ext js,jsx,ts,tsx packages && stylelint '**/*.{ts,jsx,ts,tsx}'",
```
kamleshchandnani, line 21:
```suggestion
    "lint": "eslint --ext js,jsx,ts,tsx packages && stylelint '**/*.{js,jsx,ts,tsx}'",
```

packages/blade/package.json
```diff
@@ -41,7 +41,8 @@
     }
   },
   "scripts": {
-    "typecheck": "run-s types:typecheck:*",
+    "typecheck": "run-p types:typecheck:*",
+    "test": "run-p test:*",
```
kamleshchandnani, line 45: nice 👌

divyanshu013 (approved):
Looks good to me.

.github/workflows/blade-pr-title-check.yml
chaitanyadeorukhkar: Need a separate workflow for this because the trigger types are different. If we keep the same workflow as blade-validate, rest of the validate jobs will get triggered even if the PR title is changed

kamleshchandnani: curious - open, edited, synchronize are the types that get triggered for validate too since we don't specify the types for it. which to me seems like these types are subset of validate workflow already.

chaitanyadeorukhkar: I think it doesn't get triggered with `edited` which is what we need for checking the PR title. Initially I started with keeping it in the same workflow but had to add these types for it to detect title changes

kamleshchandnani (approved):

PR #803, opened 2022-11-02
feat(Sandbox): add CodeSandbox integration to Storybook

saurabhdaware:
Integrate all the things

## Description

Adds CodeSandbox integration to storybook using [Sandpack](https://sandpack.codesandbox.io/) (Also used in new https://beta.reactjs.org)

You can use new `Sandbox` component to add the code editor to your storybook
```tsx
<Sandbox>{`<-- Your App.js code here -->`}</Sandbox>
```

> **Note**
>
> Unfortunately, Sandpack doesn't support TS autocomplete yet. They have an open issue to add support but might take time. We could've used CodeSandbox within iframe but that doesn't allow changing themes and inlined code. So it can get difficult to maintain for all the components.

Check out `Badge.stories` diff in this PR

Try it here - https://61c19ee8d3d282003ac1d81c-qbpcwspyho.chromatic.com/?path=/docs/components-badge--badge&globals=measureEnabled:false

https://user-images.githubusercontent.com/30949385/199402681-aeec8f7e-03a7-43de-837f-bd3c258f02b4.mov

**TODO**
- [x] ~Integrate this component with other stories~ (will do it in follow up PRs once this implementation is approved)
- [x] Figma Integration #804


packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
```diff
@@ -0,0 +1,94 @@
...
+            );
+            `,
+          [`/App.${language}`]: dedent(children),
+        }}
+        customSetup={{
+          dependencies: {
+            react: '18.0.0',
+            'react-dom': '18.0.0',
+            'react-scripts': '4.0.3',
+            '@razorpay/blade': '3.4.1',
```
saurabhdaware, line 77:
```suggestion
            '@razorpay/blade': '*',
```

I was wondering if we should put `*` here to avoid updating version everytime? (But if someone opens this in the codesandbox, it will show up as `*`. Hopefully people won't copy that from there 😅 )

divyanshu013: Should be fine I think, for now.

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
```diff
@@ -0,0 +1,94 @@
...
+    globals: { themeTokenName, colorScheme },
+  } = React.useContext(DocsContext);
+
+  return (
+    <Box>
+      <br />
+      <Sandpack
+        template="react-ts"
+        files={{
+          '/index.tsx': dedent`import { StrictMode } from "react";
```
divyanshu013, line 24: Btw does this support picking a file instead of typing the entire code here? That would be good quality of life improvement.

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
```diff
@@ -0,0 +1,94 @@
...
+            );
+            `,
+          [`/App.${language}`]: dedent(children),
+        }}
+        customSetup={{
+          dependencies: {
+            react: '18.0.0',
+            'react-dom': '18.0.0',
+            'react-scripts': '4.0.3',
+            '@razorpay/blade': '3.4.1',
```
divyanshu013, line 77: Will we have to bump the version here everytime?

divyanshu013 (approved):
Looks good overall, some queries.

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
saurabhdaware: Nope it doesn't :(. Unless we do `fs.readFileSync` sort of hack but seems too hacky. Anyway this file shouldn't change that often.

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
saurabhdaware: https://github.com/razorpay/blade/pull/803#discussion_r1011197348 yup need suggestions on that.

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
saurabhdaware: Yup. Changed it to `*`. We can change back to absolute version if it starts breaking a lot in future.

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
saurabhdaware: Changed it to `*`

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.native.tsx
```diff
@@ -0,0 +1,10 @@
+import dedent from 'dedent';
+import type { SandboxProps } from './Sandbox.web';
+import { Text } from '~components/Typography';
+
+// In React Native, the codesandbox doesn't work. So replacing it with normal text display for native
+function Sandbox({ children }: SandboxProps): JSX.Element {
+  return <Text>{dedent(children)}</Text>;
```
kamleshchandnani, line 7: can you add a screenshot of how it looks like on native? what if we replace it with `Code` instead of `Text`? or since we don't have multiline code variant then probably use BaseText here and maybe use the `code` font family instead of text so that the code looks like code?

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
```diff
@@ -0,0 +1,93 @@
...
+                  </BackgroundBox>
+                </BladeProvider>
+              </StrictMode>
+            );
+            `,
+          [`/App.${language}`]: dedent(children),
+        }}
+        customSetup={{
+          dependencies: {
+            react: '18.0.0',
```
kamleshchandnani, line 74: what if we keep all the deps same as package.json? we are on React 17. even though 18 is backward compatible not sure if we might come across a bug it'll be difficult to mitigate that it's the version issue

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import { Sandpack } from '@codesandbox/sandpack-react';
+import { DocsContext } from '@storybook/addon-docs';
+import dedent from 'dedent';
+import Box from '~components/Box';
+
+export type SandboxProps = {
+  children: string;
+  language?: 'js' | 'ts' | 'jsx' | 'tsx';
```
kamleshchandnani, line 9: do we need this type explicitly? we don't give the option to consumers in the playground right?

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
```diff
@@ -0,0 +1,93 @@
...
+            );
+
+            const GlobalStyles = createGlobalStyle\`
+              body {
+                margin: 0;
+                padding: 0;
+              }
+            \`;
+            
+            const rootElement = document.getElementById("root");
```
kamleshchandnani, line 51: does sandpack creates this root or storybook?

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
```diff
@@ -0,0 +1,93 @@
...
+            }
+            
+            const root = ReactDOMClient.createRoot(rootElement);
+            
+            root.render(
+              <StrictMode>
+                <BladeProvider themeTokens={${themeTokenName}} colorScheme="${colorScheme}">
+                  <GlobalStyles />
+                  <BackgroundBox>
+                    <App />
```
kamleshchandnani, line 64: what if directly do `dedent(children),` directly here inline? it doesn't work?

packages/blade/src/components/Badge/Badge.stories.tsx
```diff
@@ -50,7 +51,21 @@ const Page = (): ReactElement => {
       <br />
       <br />
       <Title>Usage</Title>
-      <Highlight language="tsx">{`import { Badge } from '@razorpay/blade/components'; \nimport type { BadgeProps } from '@razorpay/blade/components';`}</Highlight>
+      <Sandbox>
+        {`
+        import { Badge, InfoIcon } from '@razorpay/blade/components';
+        
+        function App() {
```
kamleshchandnani, line 58: I miss TS here 😭

packages/blade/src/components/Badge/Badge.stories.tsx
```diff
@@ -50,7 +51,21 @@ const Page = (): ReactElement => {
       <br />
       <br />
       <Title>Usage</Title>
-      <Highlight language="tsx">{`import { Badge } from '@razorpay/blade/components'; \nimport type { BadgeProps } from '@razorpay/blade/components';`}</Highlight>
+      <Sandbox>
```
kamleshchandnani, line 54: we need to add this change rest of the stories too right? are you gonna add it in this PR only?

saurabhdaware: Nope will add it in followup PRs after Figma's PR. Both sandbox and figma requires changes in other stories so will do it together.

packages/blade/src/components/Badge/Badge.stories.tsx
saurabhdaware: Yesss 😢 . We can add TS but autocomplete won't come in editor :(

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
saurabhdaware: it does but then users have to edit this cluttered file to make any changes. With this separated out in `App` component, users get this separate simple `App.tsx`  
image

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
saurabhdaware: Sandpack does. Inside an iframe.

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
saurabhdaware: I am setting `tsx` by default. If someone wants to add `ts` file, they can add it using this prop.

saurabhdaware: By users here, I mean the developers integrating this reusable component.

So I can do-
```tsx
<Sandpack language="ts">
{`
  const a: number = 3;
  const b: string = 'hello';
  // ts code here
`}
</Sandpack>
```

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.native.tsx
saurabhdaware: With `Text`
image

With `BaseText` and `code` fontFamily
image

saurabhdaware: I think we can go with `BaseText` and `code` font 🤔 Later migrate to codeblock component when we implement it.

Anyway currently the Sandbox is only used in docs pages which are not available on React Native. So added this more from the perspective of if someone adds the sandbox in story

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
saurabhdaware: makes sense. I'll just hardcode 17 here for now since we don't upgrade react that often. We can update it here as well when we update react.

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.native.tsx
saurabhdaware: Added BaseText

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
saurabhdaware: done

kamleshchandnani: > ~makes sense. I'll just hardcode 17 here for now since we don't upgrade react that often. We can update it here as well when we update react.~
> 
> Importing package.json and reading from the peerDependencies. This way we'll know if we genuinely break something

yep this is the ideal way 👌

packages/blade/src/_helpers/storybook/Sandbox/Sandbox.web.tsx
kamleshchandnani: cool we can remove js and jsx in that case? because I don't see a need for that to be added in examples/

saurabhdaware: yup makes sense

saurabhdaware: done

PR #814, opened 2022-11-03
feat(Indicator): add indicator component

divyanshu013:
This adds an `Indicator` component and some stories for composing with other components.

In next PRs:
- Tests
- Changeset and final export

## Screenshots

**Composition with other components**

Screen Shot 2022-11-04 at 14 34 57

**With label**

Screen Shot 2022-11-04 at 14 34 29

## References

- API Design #788 
- [Figma](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=8224%3A1)


anuraghazra:
Why is it looking like it's not aligned? 🤔 

image

packages/blade/src/components/Indicator/Indicator.tsx
```diff
@@ -0,0 +1,102 @@
...
+  accessibilityLabel,
+  children,
+  size = 'medium',
+  intent = 'neutral',
+}: IndicatorProps): ReactElement => {
+  const { theme } = useTheme();
+
+  const fillColor = theme.colors.feedback.background[intent].highContrast;
+  const strokeColor = theme.colors.brand.gray[200];
+  const getDimension = useCallback((): Dimensions => {
```
anuraghazra, line 69: Does this need to be a function? 

Can we use React.useMemo instead and inline the logic?

divyanshu013: Yes, both memo / callback works

packages/blade/src/components/Indicator/Indicator.tsx
```diff
@@ -65,7 +65,7 @@ const Indicator = ({
   const { theme } = useTheme();
 
   const fillColor = theme.colors.feedback.background[intent].highContrast;
-  const strokeColor = theme.colors.brand.gray[200];
+  const strokeColor = theme.colors.brand.gray[200].lowContrast;
```
anuraghazra, line 68: Ohh... was this the TS error? 

I wonder why ts couldn’t resolve this before.

divyanshu013: Yes, this was the issue

anuraghazra (approved):
LGTM

divyanshu013:
The issue here is on figma the alignment is not including the white stroke, will confirm with design

image

packages/blade/src/components/Indicator/Indicator.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+  return <IndicatorComponent {...args} />;
+};
+
+export const Default = IndicatorTemplate.bind({});
+
+export const WithoutLabel: ComponentStory<typeof IndicatorComponent> = ({ ...args }) => {
+  return <IndicatorComponent {...args} />;
+};
+WithoutLabel.args = {
+  children: undefined,
```
kamleshchandnani, line 80: children by default will be undefined right? do we need to explicitly pass it as undefined?

packages/blade/src/components/Indicator/Indicator.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+        '`Indicator` can be used without a label by skipping `children`. **Note**: in this case you should always pass `accessibilityLabel` for screen readers a11y',
+    },
+  },
+};
+
+export const Composition: ComponentStory<typeof IndicatorComponent> = ({ ...args }) => {
+  const isReactNative = getPlatformType() === 'react-native';
+  return (
+    <Box position="relative" display={isReactNative ? 'flex' : 'inline-flex'} alignSelf="center">
+      <Box position="absolute" top={isReactNative ? -8 : -4} right={-8} zIndex={10}>
```
kamleshchandnani, line 96: maybe put top in `%` because on RN when someone might open the story in a diff resolution emulator then this might not work. % might help there?

packages/blade/src/components/Indicator/Indicator.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const fillColor = theme.colors.feedback.background[intent].highContrast;
+  const strokeColor = theme.colors.brand.gray[200].lowContrast;
+  const getDimension = useCallback((): Dimensions => {
+    // the actual svg size would get reduced by 2 because of stroke width
+    switch (size) {
+      case 'small':
+        return { svgSize: '10', textSize: 'small' };
+      case 'xsmall':
+        return { svgSize: '8', textSize: 'small' };
```
kamleshchandnani, line 75: the text doesn't has to be xsmall here?

packages/blade/src/components/Indicator/_decisions/decisions.md
```diff
@@ -65,7 +65,7 @@ Sample usage:
 Web:
 
 - The component accepts an `accessibilityLabel` prop to let users pass an `aria-label` used by screen readers. This is helpful in building UIs where an `Indicator` is used without a visible `label`.
-- If `label` is provided without any `accessibilityLabel` we use an `aria-labelledby` internally to let the screen readers use `label` for a11y purposes.
+- If `label` is provided without any `accessibilityLabel` we use an `aria-label` internally to let the screen readers use `label` for a11y purposes.
```
kamleshchandnani, line 68: shouldn't this be `children` instead of `label` since that is the prop we pass? 

maybe something like

If `children` i.e label is provided without any `accessibilityLabel` we use an `aria-label` internally to let the screen readers use that for a11y purposes.

packages/blade/src/components/Indicator/Indicator.stories.tsx
divyanshu013: It should be but I'd specified the default value of `children` above in args 😅

packages/blade/src/components/Indicator/_decisions/decisions.md
divyanshu013: Right, missed during update

divyanshu013: Fixed in https://github.com/razorpay/blade/pull/814/commits/69498410036c299fc6f507af70adecddd4228364

packages/blade/src/components/Indicator/Indicator.stories.tsx
kamleshchandnani: did you test it on different device resolution emulators? iPhone5, 11 then google pixel and maybe Motorolla?

packages/blade/src/components/Indicator/Indicator.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+import type { ReactElement } from 'react';
+import { Highlight, Link } from '@storybook/design-system';
+
+import type { IndicatorProps } from './Indicator';
+import { Indicator as IndicatorComponent } from './Indicator';
+import useMakeFigmaURL from '~src/_helpers/storybook/useMakeFigmaURL';
+import Box from '~components/Box';
+import { Button } from '~components/Button';
+import { getPlatformType } from '~utils';
+
```
kamleshchandnani, line 12: you might need to incorporate the storywrapper and sandpack changes as Saurabh's PR is merged?

packages/blade/src/components/Indicator/Indicator.stories.tsx
divyanshu013: Verified on 6s and pixel 5, seems to be working fine. Unable to create older simulators currently. Think will need some digging to make older ones work. But I think it should translate correctly because the pixels are density independent.

image

packages/blade/src/components/Indicator/Indicator.stories.tsx
divyanshu013: Done in the refactor PR https://github.com/razorpay/blade/pull/830/commits/21df0fb87b7a8bfa523fc8fbfa81a72056d46efb

PR #815, opened 2022-11-03
test(Indicator): add web and native tests

divyanshu013:
Adds web and native tests.

In next PRs:
- Final export and changesets


anuraghazra (approved):
LGTM

packages/blade/src/components/Indicator/__tests__/Indicator.native.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
+
+describe('<Indicator />', () => {
+  it('should render', () => {
+    const { toJSON } = renderWithTheme(<Indicator accessibilityLabel="Success" />);
+
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render different variants for size and intent', () => {
```
kamleshchandnani, line 11: render all the size and intent perm and combination? i think it's ok for these kinds of tests to be verbose since they affect visual aspects of the component and also if one of the token changes we might not even discover if it affected indicator

packages/blade/src/components/Indicator/__tests__/Indicator.native.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+    );
+
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render text label', () => {
+    const label = 'Active';
+    const { getByText } = renderWithTheme(<Indicator>{label}</Indicator>);
+
+    getByText(label);
```
kamleshchandnani, line 23: shouldn't we assert here rather than just calling this function?

packages/blade/src/components/Indicator/__tests__/Indicator.web.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const { getByText } = renderWithTheme(<Indicator>{label}</Indicator>);
+
+    getByText(label);
+  });
+
+  it('should have role and a11y label when text label is passed', async () => {
+    const label = 'Active';
+    const { getByRole, getByLabelText } = renderWithTheme(<Indicator>{label}</Indicator>);
+
+    getByRole('status');
```
kamleshchandnani, line 31: if we are doing `assertAccessible` then do we need this `getByRole`? doesn't `assertAccessible` checks for role as well?

packages/blade/src/components/Indicator/__tests__/Indicator.native.test.tsx
divyanshu013: This automatically fails the test if the element isn't found, so we don't need to explicitly assert the element after `getBy*`

packages/blade/src/components/Indicator/__tests__/Indicator.web.test.tsx
divyanshu013: Here in `getByRole` we assert that the role should be `status`. `assertAccessible` will highlight issues where we can forget passing a role but it doesn't know that the `role` should be `status`.

kamleshchandnani:
@divyanshu013 tests failing?

packages/blade/src/components/Indicator/__tests__/Indicator.native.test.tsx
divyanshu013: Added in https://github.com/razorpay/blade/pull/815/commits/5b465905cd5269dd300c467e55a465d2c4537851

kamleshchandnani (approved):

PR #818, opened 2022-11-04
feat(storybook): add support for theming

saurabhdaware:
## Description

- Changes the canvas background to background color token. So the background of stories now changes when we change theme.
- Some aesthetic improvements

Try changing themes here - https://61c19ee8d3d282003ac1d81c-brglhsqsoq.chromatic.com/?path=/story/guides-intro--page&globals=measureEnabled:false

> **Note**
>
> This changes the canvas so the chromatic requires approvals for all components. I will do that once the PR is approved.

### Docs Changes

Things were too cluttered in some places. Improved the spacings a bit. Also changed the intro page badges and some font sizes for the love of aesthetics.

<table>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
</table>

### Theming

https://user-images.githubusercontent.com/30949385/199931619-674f2609-1961-49e5-a73c-d9cd1dda6e5f.mov

### DIFF
--- a/packages/blade/.storybook/react/manager.js
+++ b/packages/blade/.storybook/react/manager.js
@@ -23,7 +23,7 @@ export const theme = create({
   fontCode: 'monospace',
 
   // Text colors
-  textColor: '#333333',
+  textColor: 'rgb(19, 38, 68)',
   textInverseColor: '#FFFFFF',
   textMutedColor: '#666666',
 
--- a/packages/blade/.storybook/react/manager.js
+++ b/packages/blade/.storybook/react/manager.js
@@ -1,10 +1,11 @@
 // .storybook/manager.js
 import React from 'react';
+import { create } from '@storybook/theming';
 import { addons, types } from '@storybook/addons';
 import { useGlobals } from '@storybook/api';
 import { Icons, IconButton } from '@storybook/components';
 
-export const theme = {
+export const theme = create({
   base: 'light',
 
   colorPrimary: '#FF4785',
@@ -40,7 +41,7 @@ export const theme = {
   brandTitle: 'Blade',
   brandUrl: 'https://github.com/razorpay/blade',
   // brandImage: 'https://place-hold.it/350x150',
-};
+});
 
 const ADDON_ID = 'internal-components-addon';
 const TOOL_ID = 'internal-components-tool';
@@ -52,7 +53,7 @@ hiddenStoryStyle.textContent = `
 `;
 document.head.append(hiddenStoryStyle);
 
-export const toggleHiddenStoryStyle = isDisabled => hiddenStoryStyle.disabled = isDisabled
+export const toggleHiddenStoryStyle = (isDisabled) => (hiddenStoryStyle.disabled = isDisabled);
 
 const InternalStoryAddon = () => {
   const [{ showInternalComponents }, updateGlobals] = useGlobals();
@@ -61,7 +62,7 @@ const InternalStoryAddon = () => {
     updateGlobals({
       showInternalComponents: !showInternalComponents,
     });
-    toggleHiddenStoryStyle(!showInternalComponents)
+    toggleHiddenStoryStyle(!showInternalComponents);
   }, [showInternalComponents]);
 
   return (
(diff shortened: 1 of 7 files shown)


packages/blade/.storybook/react/manager.js
```diff
@@ -1,10 +1,11 @@
 // .storybook/manager.js
 import React from 'react';
+import { create } from '@storybook/theming';
 import { addons, types } from '@storybook/addons';
 import { useGlobals } from '@storybook/api';
 import { Icons, IconButton } from '@storybook/components';
 
-export const theme = {
+export const theme = create({
```
saurabhdaware, line 8: same thing but with TS types

packages/blade/.storybook/react/manager.js
```diff
@@ -22,7 +23,7 @@ export const theme = {
   fontCode: 'monospace',
 
   // Text colors
-  textColor: '#333333',
+  textColor: 'rgb(19, 38, 68)',
```
saurabhdaware, line 26: our actual default text color from blade

packages/blade/.storybook/react/preview.js
```diff
@@ -38,12 +40,63 @@ export const parameters = {
...
+        & :not(pre) > code {
+          line-height: 1;
+          margin: 0 2px;
+          padding: 3px 5px;
+          white-space: nowrap;
+          border-radius: 3px;
+          font-size: 13px;
+          border: 1px solid #eeeeee;
+          background-color: #f8f8f8;
+        }
```
saurabhdaware, line 65: It removed the styles of `code` component inside list so I copy-pasted the styles again in this.

In storybook overrides, at certain places like these you would think, "Hey why are we doing this?" and in most such cases the answer is "that is the only way to make it work"

saurabhdaware (commented):
Review Helpers

packages/blade/.storybook/react/preview.js
```diff
@@ -38,12 +39,60 @@ export const parameters = {
   },
   docs: {
     theme,
+    components: {
+      summary: styled.summary`
```
saurabhdaware, line 43: If we can override, can we not add Blade's components here?

We can. I did. It was painful. More context in description here - #817

packages/blade/.storybook/react/preview.js
kamleshchandnani: since it got overridden because you added these styles there might be other components whose style might have broken too and we might have to add those styles?

packages/blade/.storybook/react/preview.js
```diff
@@ -38,12 +39,60 @@ export const parameters = {
...
+          font-size: 13px;
+          border: 1px solid #eeeeee;
+          background-color: #f8f8f8;
+        }
+      `,
+    },
   },
 };
 
+const StoryCanvas = styled.div(
```
kamleshchandnani, line 70: curious: where did you find this thing? the docs of storybook itself is not good. finding something known itself is a challenge wondering how did you discover unknown 😄

saurabhdaware: I don't think so it is there in docs. I just saw `BladeProvider` and thought adding a component inside this will give access to themes and theme changes 🙈

packages/blade/.storybook/react/preview.js
saurabhdaware: Yupp!! So I checked `p` and it wasn't overridden. Definitely there might be something that is overriden but don't see anything broken in docs right now. We can fix it here if something looks strange in docs.

Also even with overriden styles, it doesn't completely go out of theme. It has some styles by default. `code` in particular was looking a bit strange.

divyanshu013 (approved):
Looks good to me. Would be nice if we could do a single click theme toggle rather than dark and then switch theme to banking, but this is nice quality of life improvement 💯

packages/blade/.storybook/react/manager.js
saurabhdaware: oh I tried but those `~` aliases from codebase are failing on storybook. I tried adding the aliases to storybook's webpack as well but its still failing to resolve certain files that way 🤔

kamleshchandnani (approved):
LGTM 🚢

PR #823, opened 2022-11-07
fix(Text): remove default text-align value

### LINKED ISSUE #785: Heading, Text components don't inherit center alignment
Reproduction: https://codesandbox.io/s/center-align-not-working-blade-l1sqi3?file=/src/App.tsx

I added the components inside center aligned div. It gets applied to some component but doesn't work on some components 🤔 

image

saurabhdaware:
Resolves #785 

Change: https://github.com/razorpay/blade/pull/823#discussion_r1015013494

Rest are snapshot updates


packages/blade/src/components/Typography/BaseText/getBaseTextStyles.ts
```diff
@@ -10,7 +10,7 @@ const getBaseTextStyles = ({
   fontStyle = 'normal',
   textDecorationLine = 'none',
   lineHeight = 'l',
-  textAlign = 'left',
+  textAlign,
```
saurabhdaware, line 13: This is the only change. Rest are snapshot updates

kamleshchandnani (approved):
LGTM 🚢

PR #830, opened 2022-11-09
refactor(Indicator): add figma embed and sandbox

divyanshu013:
Matches existing stories of other components.

> **Note**
>
> `Indicator` needs to be published before it starts showing in the sandbox


packages/blade/src/components/Indicator/Indicator.stories.tsx
```diff
@@ -35,11 +36,24 @@ const Page = (): ReactElement => {
         Indicators describe the condition of an entity. They can be used to convey semantic meaning,
         such as statuses and semantical-categories.
       </Subtitle>
```
kamleshchandnani, line 38: incorporate storybook wrapper changes as well?

divyanshu013: Done in https://github.com/razorpay/blade/pull/830/commits/21df0fb87b7a8bfa523fc8fbfa81a72056d46efb

kamleshchandnani (approved):
🚢 🚀

PR #837, opened 2022-11-11
chore: Badge design changes

chaitanyadeorukhkar:
- Adds a new `small` size
- Bumps existing `small` & `medium` to `medium` & `large` respectively
- Horizontal padding changes in the `large` size
- Collateral changes for `Counter` component since it's built on top of `Badge`. This will be rectified in a follow up PR.

Screenshot 2022-11-17 at 3 34 20 PM
Screenshot 2022-11-17 at 3 34 26 PM
Screenshot 2022-11-17 at 3 34 30 PM


packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -92,9 +89,7 @@ const getTextProps = <T extends { variant: TextVariant }>({
 const StyledText = styled(BaseText)(({ truncateAfterLines }) => {
   if (truncateAfterLines) {
     if (getPlatformType() === 'react-native') {
-      return {
-        numberOfLines: truncateAfterLines,
-      };
+      return null;
```
anuraghazra, line 92: Why are we returning null here instead?

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -12,19 +12,19 @@ type TextCommonProps = {
   contrast?: ColorContrastTypes;
   truncateAfterLines?: number;
   children: React.ReactNode;
+  weight?: keyof Theme['typography']['fonts']['weight'];
+  _color?: BaseTextProps['color'];
```
divyanshu013, line 16: So consumers will also be able to use `_color` on `Text` component right?

divyanshu013 (approved):
Looks good to me. Just a query around `_color` prop.

packages/blade/src/components/Typography/Text/Text.tsx
chaitanyadeorukhkar: since react native doesn't accept `numberOfLines` as a style property. Before this change, react native was throwing a warning for this. We're passing `numberOfLines` as a prop to react native's Text

anuraghazra (approved):
Looks good. Make sure to test on react-native while copy pasting the counter I saw there was a flex box issue where the counter was covering the whole screen.

PR #838, opened 2022-11-13
docs(Icons): add icon grid story

### LINKED ISSUE #669: Add icons grid story
Add a story to showcase all icons so it makes it easier to identify which already exist

harshcut:
**Description**

Closes #669. This is what I came up with. When `icon` is set to `undefined` it shows all available icons with their respective names. To see individual icon, select option can be used. More at https://github.com/razorpay/blade/issues/669#issuecomment-1312676483.

### DIFF
--- a/packages/blade/src/components/Icons/Icons.stories.tsx
+++ b/packages/blade/src/components/Icons/Icons.stories.tsx
@@ -5,7 +5,6 @@ import iconMap from './iconMap';
 import type { IconProps } from '.';
 import { CreditCardIcon } from '.';
 import Box from '~components/Box';
-import { Text } from '~components/Typography';
 import StoryPageWrapper from '~src/_helpers/storybook/StoryPageWrapper';
 import Sandbox from '~src/_helpers/storybook/Sandbox';
 
@@ -98,7 +97,16 @@ export const AllIcons: ComponentStory<ComponentType<IconProps>> = ({ ...args }) 
             key={key}
           >
             <IconComponent {...args} />
-            <Text size="small">{icon}</Text>
+            <small
+              style={{
+                width: '90%',
+                overflow: 'hidden',
+                textOverflow: 'ellipsis',
+                textAlign: 'center',
+              }}
+            >
+              {icon}
+            </small>
           </Box>
         );
       })}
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/Icons/Icons.stories.tsx
```diff
@@ -67,15 +69,38 @@ export default {
...
   icon,
   ...args
 }) => {
-  const IconComponent = iconMap[icon];
-  return <IconComponent {...args} />;
+  if (icon) {
+    const IconComponent = iconMap[icon];
+    return <IconComponent {...args} />;
+  }
+  return (
```
saurabhdaware, line 80: We don't have to add this in same template, We can create a new export where we loop over iconMap and render icons and we can let this icon template be as it is maybe?

You can check out how we do it in `Badge.stories.tsx`

saurabhdaware (commented):
Heya! Thanks for the PR! Added a small suggestion below -

packages/blade/src/components/Icons/Icons.stories.tsx
harshcut: What should be the name of the story(s) that includes all the icons?

saurabhdaware: How about `All Icons` 🤔 ?

harshcut: Yup, this is exactly what I had in mind. But none of the other components had a `All ~` story so I skipped over it.

divyanshu013 (commented):
Looks good. Added a comment on the related issue for styling fixes.

divyanshu013 (dismissed):
Looks good to me.

packages/blade/src/components/Icons/Icons.stories.tsx
```diff
@@ -79,3 +80,36 @@ export const Icon = IconTemplate.bind({});
...
+            height={95}
+            width={125}
+            display="inline-flex"
+            flexDirection="column"
+            alignItems="center"
+            gap="spacing.6"
+            key={key}
+          >
+            <IconComponent {...args} />
+            <small
```
saurabhdaware, line 100: I think the `small` tag here will crash the react native storybook. We can perhaps use `Box` here?

PR #845, opened 2022-11-16
feat(Sandbox): use csb preview blade version in sb previews

saurabhdaware:
Storybook Sandpack now uses the preview version deployed by codesandbox CI.

This will mean, if you add a new component that is not released yet. You will be able to preview that component as well in your storybook sandbox.

image

You can also run `GITHUB_SHA=<full commit sha of version of blade you want> yarn start:web` locally to use that version locally in your storybook.

### DIFF
--- a/.github/workflows/blade-chromatic.yml
+++ b/.github/workflows/blade-chromatic.yml
@@ -30,3 +30,5 @@ jobs:
           workingDir: packages/blade
           buildScriptName: react:storybook:build
           exitOnceUploaded: true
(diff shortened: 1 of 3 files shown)


kamleshchandnani:
overall lgtm 👍 
one question i had was when the codesandbox build is running and there are chances that storybook might get built before csb so during that time if i open the storybook the sandpack will fail to load right?

kamleshchandnani (approved):

PR #846, opened 2022-11-17
fix(Storybook): failing csb previews on master

saurabhdaware:
Sandpack is failing on master because we don't publish codesandbox ci preview version on master. This PR adds check that sets `*` version on master.

https://razorpay.slack.com/archives/G01B3LQ9H0W/p1668669020171609

### DIFF
--- a/.github/workflows/blade-chromatic.yml
+++ b/.github/workflows/blade-chromatic.yml
@@ -32,3 +32,4 @@ jobs:
           exitOnceUploaded: true
         env:
           GITHUB_SHA: ${{ github.sha }}
+          GITHUB_REF: ${{ github.ref }}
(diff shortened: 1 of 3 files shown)


.github/workflows/blade-chromatic.yml
```diff
@@ -32,3 +32,4 @@ jobs:
           exitOnceUploaded: true
         env:
           GITHUB_SHA: ${{ github.sha }}
```
kamleshchandnani, line 34: can we just do a conditional in actions itself? where if the ref is master then we set the GitHub Sha to empty or null instead of passing everything and then checking in the code? with this we won't even need to add more conditions to our code and it'll also fit your logic intent that if you sha we'll use that else just use `*`

saurabhdaware: Exactly what I tried first but tricky part is we can't do ternary operators in github actions so we'll have to do something like - 

```yaml
- name: Conditionally set env variable
  if: github.ref !== '/refs/heads/master'
  run: |
    echo "GITHUB_SHA=${{ github.sha }}" >> $GITHUB_ENV
```

seems hackier than just passing the ref forward and checking in code 🙈

PR #847, opened 2022-11-17
fix: remove aria-hidden for checkbox and radio

kamleshchandnani:
This PR removes the bug where the checkbox and radio buttons were not accessible with RTL when we do `getByRole` query

Now we can do `getByRole('checkbox',{ name:'label'})` where the name is the accessible label of the checkbox/radio controls and not the `name` attribute of these input elements.


anuraghazra (approved):
LGTM!

saurabhdaware (approved):

PR #853, opened 2022-11-18
feat(storybook): add dashboard recipe

saurabhdaware:
Stacked on top of #844 

Adds Dashboard Example to Storybook

Link to example: https://codesandbox.io/s/blade-dashboard-uyijy3?file=/src/Dashboard.tsx


kamleshchandnani (approved):
lgtm 🚀

kamleshchandnani (changes requested):
by mistake approved wrong PR 🤦🏻‍♂️

kamleshchandnani:
Some feedback:
* Can we change the toggle theme to select? because right now it feels that we are changing the dark/light modes and also not clear which theme is which one between banking and payment
* the `x` on text input has some gradient for some reason not sure why 🤔 
Screenshot 2022-11-21 at 11 01 57 AM

saurabhdaware:
> the x on text input has some gradient for some reason not sure why 🤔

I think it's macos default for `type="search"`. Let me create an issue and we can see if we can remove/change it.

> Can we change the toggle theme to select? because right now it feels that we are changing the dark/light modes and also not clear which theme is which one between banking and payment

Makes sense. Will change it.

kamleshchandnani:
> Oh we don't have select component yet so that in itself will look a bit out of theme or will have to override styles.

you can do this

```
<TextInput 
      label="Search"
      placeholder="Search" 
      icon={SearchIcon}
      showClearButton={true}
    />
```

saurabhdaware:
ops lol. Lemme create issue only 😂 

image

divyanshu013 (commented):
Looks good apart from the visual issues

kamleshchandnani:
did we change something? the current view just shows output window. previously it was showing code and output both. I was bit confused but then i dragged the dragbar and then it opened the code and output side by side.


Screenshot 2022-11-21 at 5 16 03 PM

packages/blade/src/storybook-recipes/SimpleDashboard.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { RecipeSandbox } from '~src/_helpers/storybook/Sandbox';
+import parameters from '~src/_helpers/storybook/recipeParameters';
+
+export function SimpleDashboard(): JSX.Element {
+  return (
+    <RecipeSandbox
+      title="Blade Dashboard"
+      codesandboxId="blade-dashboard-uyijy3"
+      activeFile="/src/Dashboard.tsx"
+      view="preview"
```
saurabhdaware, line 11: @kamleshchandnani I added view="preview" here because the dashboard is looking better in full screen mode. You can drag and bring code or also open it in codesandbox with "open in CodeSandbox" link there. 🤔

kamleshchandnani (approved):
LGTM 🚀

PR #854, opened 2022-11-20
chore: replaced `isNegative` prop with `hasError`

k-rajat19:
changed the prop name in `SelectorInput` component as  mentioned [here ](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Radio/Radio.tsx#L95)


anuraghazra (approved):
Thanks for the contribution Rajat. 

LGTM.

PR #858, opened 2022-11-21
fix: set input type='text' when type='search' passed

### LINKED ISSUE #857: Default Close Button on `type="search"` is visible
If we set `type="search"` in the TextInput, it adds default clear button with strange gradient.

```jsx
<TextInput
  type="search"
  name="search"
  label="Search"
  icon={SearchIcon}
/>
```

image

We also have `showClearButton={true}` that adds close button from our end which ends up adding 2 close buttons

```jsx
<TextInput
  type="search"
  name="search"
  label="Search"
  icon={SearchIcon}
  showClearButton={true}
/>
```
image

kamleshchandnani:
fixes #857

### DIFF
--- a/packages/blade/src/components/Input/TextInput/TextInput.tsx
+++ b/packages/blade/src/components/Input/TextInput/TextInput.tsx
@@ -134,6 +134,13 @@ const getKeyboardAndAutocompleteProps = ({
     keyboardAndAutocompleteProps.type = 'text';
   }
 
+  if (type === 'search') {
+    /* when input type:search is provided at that time browser adds a weird close button which collides with our clear button and then we have 2 clear buttons
+     * source: https://github.com/razorpay/blade/issues/857#issue-1457367160
+     */
+    keyboardAndAutocompleteProps.type = 'text';
+  }
+
   return keyboardAndAutocompleteProps;
 };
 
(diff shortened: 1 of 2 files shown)


saurabhdaware:
I feel we should remove it from css rather than changing the type cause there are certain behavioural things attached to it like - The search clears when you press ESC (doesn't happen with text type), or voice over will announce it as `search text field` and type text field will be announced as `text field` (not a bit deal if you have label but still a nice enhancement?) + the semantic meaning

kamleshchandnani:
> I feel we should remove it from css rather than changing the type cause there are certain behavioural things attached to it like - The search clears when you press ESC (doesn't happen with text type), or voice over will announce it as `search text field` and type text field will be announced as `text field` (not a bit deal if you have label but still a nice enhancement?) + the semantic meaning

so the main thing is keyboard autosuggest and autofill which we are anyways setting explicitly using the attributes. we do similar thing for numeric input as well so i think it shouldn't be a problem? clearing the `esc` thing is nice to have but not sure if we just take into account that? also on react-native the behavior might be different too.

divyanshu013 (approved):
Looks good to me. Does this change anything on native?

divyanshu013:
> I feel we should remove it from css

Is it possible to remove this using CSS 🤔 might be a good alternative

saurabhdaware (approved):
> Is it possible to remove this using CSS 🤔 might be a good alternative

I did find some solutions on stackoverflow - https://stackoverflow.com/questions/57011736/how-to-get-rid-of-the-x-on-the-search-bar. Should be possible.

I think the current implementation would also be fine for now but we should evaluate and replace these solutions with css solutions and try to keep HTML semantic. We might miss out on some nice browser-specific features with this.

divyanshu013:
Damn that SO question is so downvoted 😹 

Might need testing on various ios, android browsers

PR #860, opened 2022-11-21
fix: show `Spinner` on `TextInput` when `isLoading=true`

kamleshchandnani:
This PR does 2 things
* Adds spinner when `isLoading: true` is passed to `TextInput`. This was a long pending TODO
* Update Spinner sizes after the design was updated
	*  This doesn't need any code mod since there are 9 instances of spinner being used with default variant i.e medium


divyanshu013 (approved):
Looks good to me. Checklist:
- [x] Add changeset

packages/blade/src/components/Spinner/BaseSpinner/SpinningBox.web.tsx
```diff
@@ -14,6 +14,7 @@ const rotate = keyframes`
 
 const SpinningBox = styled(Box)(({ theme }) => {
   return css`
+    padding: 1px;
```
chaitanyadeorukhkar, line 17: Why is this needed?

chaitanyadeorukhkar:
@kamleshchandnani something looks off in the storybook. The small size seems too big. Can you check once?
Screenshot 2022-11-22 at 12 04 34 PM

divyanshu013:
I think because this PR removes `small` size so the stories might need to be updated.

kamleshchandnani:
> @kamleshchandnani something looks off in the storybook. The small size seems too big. Can you check once? Screenshot 2022-11-22 at 12 04 34 PM

fixed in [`f782168` (#860)](https://github.com/razorpay/blade/pull/860/commits/f782168034ecaba1856a70914f65ec1a58151914)

packages/blade/src/components/Spinner/BaseSpinner/SpinningBox.web.tsx
chaitanyadeorukhkar: @kamleshchandnani ^

kamleshchandnani: on design we added this to make it visually aligned with text previously without this the spinners were looking weird when placed alongside textinput or menu items

chaitanyadeorukhkar: But this also means that when Spinners are used standalone for example in center of a page, it will be 1px off 🤔 

@saurav12  is that okay?

saurav12: @chaitanyadeorukhkar Yes, that is intentional. And yes while it is rendered standalone on a page, it is okay to render it with `1px` padding.

chaitanyadeorukhkar (approved):

PR #861, opened 2022-11-21
chore(blade): added blade component data attributes

anuraghazra:
This PR adds data attributes in blade to help collect metric: [JIRA Ticket](https://razorpay.atlassian.net/browse/FC-1302?atlOrigin=eyJpIjoiZDlhNWI1YjdhYmM3NGUzM2EwNzMwN2I2MDgzNTViNjgiLCJwIjoiaiJ9)

- a new `dataAttribute` function which returns data attribute & value pair which we can spread into the html
- Added `data-blade-component={name}` to all components 

Check this [commit](https://github.com/razorpay/blade/pull/861/commits/b8929dd25575ef61b7cc553b0fd48c4f29f55961) to avoid snapshots while reviewing


packages/blade/src/components/Input/OTPInput/OTPInput.tsx
```diff
@@ -251,7 +251,7 @@ const OTPInput = ({
   };
 
   return (
-    <>
+    <Box {...dataAttribute('blade-component', 'otp-input')}>
```
anuraghazra, line 254: Added this Box wrapper around the OTPInput because previously the jsx structure was like so: 

```
<>
  <Box /> // main input
  <Box />  // hint content
</>
```
This is problematic since if users tries to wrap the OTP component with their own Flex components the hints & otp's content will be treated as separate flex child.

packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -225,6 +225,7 @@ const Alert = ({
       isFullWidth={isFullWidth}
       isBorderless={isBorderless}
       {...a11yProps}
+      {...dataAttribute('blade-component', 'alert')}
```
anuraghazra, line 228: Yes could've just done `is-blade-component=true` but decided to anyways add names to component in case we ever need to know exactly which components are on page.

kamleshchandnani: yup this is ideal 👌

packages/blade/src/utils/dataAttribute/dataAttribute.web.ts
```diff
@@ -0,0 +1,6 @@
+const dataAttribute = (name: string, value: string | undefined): Record<string, string> => {
```
kamleshchandnani, line 1: we should be able to add multiple data attributes. because we also need to add `data-test-id` as well. so then we need to call this function for each prop?

kamleshchandnani, line 1: maybe let's rename this file to `automationAttributes` or `metaAttributes` or something generic since there's no such thing as `data` attributes on RN

packages/blade/src/utils/dataAttribute/index.native.ts
```diff
@@ -0,0 +1 @@
+export * from './dataAttribute.native';
```
kamleshchandnani, line 1: why 2 index files? is that how we write in module suffixes world?

packages/blade/src/utils/dataAttribute/dataAttribute.web.ts
anuraghazra: I see, we could do that, but then for react-native we will need to add logic to filter everything out and only pick testID prop. 🤔 

Plus we are not adding `data-testid` at the same dom nodes where we are adding the `data-blade-component` attribute, so any which ways we will need to call this multiple times depending on where we need the attributes. 

So i just thought it will be better to separate out `dataAttribute` and `testID` function which we already have.

anuraghazra: `metaAttribute` sounds good.

packages/blade/src/utils/dataAttribute/index.native.ts
anuraghazra: Yes, 1 for web, 1 for native. Notice how we don't need another `dataAttribute.d.ts` file and then reexport the function again from there, much simpler & less confusing. 

You can check out the "Module Resolution Example" section at the end of [TypeScript RFC](https://github.com/razorpay/blade/blob/master/rfcs/writing-cross-platform-typescript.md)

image

packages/blade/src/utils/dataAttribute/dataAttribute.web.ts
anuraghazra: Renamed 👍

kamleshchandnani (approved):

packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -255,6 +255,7 @@ const Alert = ({
       isFullWidth={isFullWidth}
       isBorderless={isBorderless}
       {...a11yProps}
+      {...metaAttribute('blade-component', 'alert')}
```
chaitanyadeorukhkar, line 258: Lets store these as constants somewhere? Very easy to misspell any of these and mess up our data
```suggestion
      {...metaAttribute(MetaConstants.Component, MetaConstants.Alert)}
```

anuraghazra: Done.

PR #872, opened 2022-11-22
feat: add new icons (lock, settings, file-text, users, slash)

akkhil7:
## Users
Screenshot 2022-11-22 at 2 27 06 PM

## File Text
Screenshot 2022-11-22 at 2 27 28 PM

## Lock
Screenshot 2022-11-22 at 2 27 43 PM

## Slash
Screenshot 2022-11-22 at 2 27 50 PM

## Settings
Screenshot 2022-11-22 at 2 28 05 PM


.changeset/tough-ears-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+feat: add new icons
```
snitin315, line 5: I feel it would be good to mention which icons have been added.

akkhil7: Makes sense, added to the PR title and desc.

packages/blade/src/components/Icons/SlashIcon/SlashIcon.tsx
```diff
@@ -0,0 +1,17 @@
+import type { IconComponent } from '..';
+import useIconProps from '../useIconProps';
+import { Path, Svg } from '../_Svg';
+
+const SlashIcon: IconComponent = ({ size, color }) => {
+  const { height, width, iconColor } = useIconProps({ size, color });
+  return (
+    <Svg height={height} width={width} viewBox="0 0 22 22" fill="none">
```
anuraghazra, line 8: Are these icons taken from the blade figma files? If so ideally they should all have same viewBox of `0 0 24 24` right? Make sure you are copying the correct layer. 

cc @saurav12 @Chaitanya-019

akkhil7: I see, I copied the existing LinkIcon, and updated paths and variable names. That also has a viewBox os 0 0 22 22 apparently. Let me update it to `0 0 24 24`.

chaitanyadeorukhkar: @akkhil7 while copying the svg, you have to select the root icon element, in this case the parent "lock" element and not the ones within. All of these root svgs have a viewbox of `0 0 24 24`. After copying if the viewbox isn't `0 0 24 24` then there might be some issue while copying. Can you check this for all the icons added in this PR once?
Screenshot 2022-11-22 at 3 05 04 PM

chaitanyadeorukhkar (approved):

PR #873, opened 2022-11-22
fix(Alert): responsive design alignment

### LINKED ISSUE #864: Alert: discrepancy in title and icon alignment on mobile and native
We'd a design discrepancy in the alignment between title and the icon happening particularly on mobile displays (and react native). This was replicated on code as well. Based on [discussions](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1669030447849819) with design this should be handled responsively with more padding / margin.

## Screenshots

| Desktop | Mobile |
| - | - |
|  image | image |

divyanshu013:
Fixes icon alignment for mobile and react native. Was slightly off earlier. Resolves #864 

## Screenshots

Screen Shot 2022-11-22 at 14 21 36
Screen Shot 2022-11-22 at 14 25 25
Screen Shot 2022-11-22 at 14 25 59
Screen Shot 2022-11-22 at 14 26 32
Screen Shot 2022-11-22 at 14 27 00


packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -139,14 +141,37 @@ const Alert = ({
...
+
   const [isVisible, setIsVisible] = useState(true);
   const contrastType = `${contrast}Contrast` as const;
   const iconSize = isBorderless ? 'large' : 'medium';
   const textSize = isBorderless ? 'medium' : 'small';
 
   const Icon = intentIconMap[intent];
+  let iconOffset: DotNotationSpacingStringToken = 'spacing.1';
+
+  // certain special cases below needs special care for near perfect alignment
```
kamleshchandnani, line 156: 😭

packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -139,14 +141,37 @@ const Alert = ({
...
+    if (!isBorderless && title) {
+      iconOffset = 'spacing.2';
+    } else if (isBorderless && !title) {
+      iconOffset = 'spacing.2';
+    }
+  }
+
   const icon = (
-    <Box marginTop="spacing.1" display="flex">
+    <Box marginTop={iconOffset} display="flex">
```
kamleshchandnani, line 174: why is `spacing.1` behaving weirdly in diff scenarios? will it be solved if we use `alingItems:center` or something?

divyanshu013: Various cases:
- the bordered version uses a text as title whereas the borderless one uses heading
- the font size increases slightly for mobile
- we also need to handle description only alerts which uses a different sized text (plus web and mobile layouts)
- react native isn't mapping 1:1 with web (I think platform differences)

Regarding center `alignItems` that would make the icon off center because we need to align it to top

chaitanyadeorukhkar: Damn. The logic is quite specific 😢 

Have you tried running on both Android & iOS to make sure the logic stands true?

chaitanyadeorukhkar (approved):
Looks okay to me. Since its aligned to top and not center, don't see any other way to fix these issues. Can you just make sure it works okay on Android as well as iOS?

packages/blade/src/components/Alert/Alert.tsx
divyanshu013: Yes, please check the screenshots in PR description, took quite a while 😅

packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -139,14 +141,37 @@ const Alert = ({
...
+    } else if (!isBorderless && title) {
+      iconOffset = 'spacing.2';
+    }
+  } else if (isMobile) {
+    if (!isBorderless && title) {
+      iconOffset = 'spacing.2';
+    } else if (isBorderless && !title) {
+      iconOffset = 'spacing.2';
+    }
+  }
```
anuraghazra, line 171: Woaahh.. NOW lets see if GPT3 can explain this code!!

anuraghazra:
everything is unchanged. 
image

Idea: Would actually be nice if we could do visual testing for different screen sizes.

anuraghazra (approved):
Looks good to me. tho unfortunate there's no simple solution :(

divyanshu013:
@anuraghazra to quote your blog "Interoperability isn’t free" 🥲

packages/blade/src/components/Alert/Alert.tsx
divyanshu013: 😹

packages/blade/src/components/Alert/Alert.tsx
divyanshu013: I feel uneasy introducing these checks in the code 😅

PR #878, opened 2022-11-23
feat(IconButton.stories): show console by default

saurabhdaware:
Since we're console logging on clicks it's a nice quality of life improvement to show console by default


divyanshu013 (approved):
Looks good to me.

PR #879, opened 2022-11-23
docs: add usage guidelines

### LINKED ISSUE #865: Docs: add dev guidelines for using blade
There should be a dev guidelines doc with instructions on ideal usage of Blade. For example:
- Don't hardcode tokens in your style files
- Don't use DOM manipulation on components to morph the components into your specific usecases, prefer composition
- etc.

kamleshchandnani:
Fixes #865 

Added basic common things people get confused with while using Blade on Figma and Code.

Checkout the guide on [this link](https://61c19ee8d3d282003ac1d81c-maalculvjq.chromatic.com/?path=/docs/guides-how-to-use--page)


packages/blade/docs/guides/Usage.stories.mdx
```diff
@@ -1,72 +1,99 @@
...
+Blade offers 2 pre-built themes out of the box:
+
+1. `paymentTheme` - To be used in Payment(PG) products.
+2. `bankingTheme` - To be used in RazorpayX products.
+
+The benefits of using `BladeProvider`?
+
+- Your components will automatically become responsive on different breakpoints.
+- Capability to switch between light mode and dark mode.
+- Units for typography and other layouts(where to use `px`, `rem` etc. will automatically be handled).
```
divyanshu013, line 29: Minor
```suggestion
- Units for typography and other layouts (where to use `px`, `rem` etc. will automatically be handled).
```

packages/blade/docs/guides/Usage.stories.mdx
```diff
@@ -1,72 +1,99 @@
...
+import { paymentTheme } from '@razorpay/blade/tokens';
+import App from from './App';
+
+function AppWrapper(): JSX.Element {
+  return (
+    <React.Fragment>
+      <BladeProvider themeTokens={paymentTheme} colorScheme="light">
+        <App />
+      </BladeProvider>
+    </React.Fragment>
```
divyanshu013, line 43: I think we can skip Fragment here.

packages/blade/docs/guides/Usage.stories.mdx
```diff
@@ -1,72 +1,99 @@
...
+<Button type="secondary" size="medium" icon={DownloadIcon} iconPosition="left">
+  Export
+</Button>;
+```
+
+## Mapping Tokens from Figma to Blade in your code
+
+> Tokens are an integral part of design system and are used to store design decisions behind a token(variable) name. That also make things easier to update without you making any change. For this reason **never hardcode** token values in your code but rather use the token name.
+
+As mentioned earlier everything that you see on Figma built using Blade you'll always find that in code as well.
```
divyanshu013, line 77: Optional: think we don't need to stress on the point too much 😸 

```suggestion
Everything that you see on Figma built using Blade you'll always find that in code as well.
```

packages/blade/docs/guides/Usage.stories.mdx
```diff
@@ -1,72 +1,99 @@
...
+
+Let's see how you can spot a token name from figma and write it in your code
+
+<img src={figmaToken} alt="Figma mockup showing how to pick token" />
+
+Whenever you get a UI mockup handed over by your designer
+
+1. Select the component whose color token you want to use in code up
+2. Click on the inspect panel on the left and click on the token name `Surface/Background/Level3/lowContrast`.
+   > Remember, don't select the hardcoded `hsla` color value.
```
divyanshu013, line 87: We can make it more noticeable

```suggestion
   > **Note**
   >
   > Remember, don't select the hardcoded `hsla` color value.
```

divyanshu013 (approved):
Looks good to me.

packages/blade/docs/guides/Usage.stories.mdx
kamleshchandnani: this won't work because this is mdx of storybook and not GitHub's md 😞

divyanshu013: Oh, I thought it also parsed markdown so atleast it would bold the note

PR #885, opened 2022-11-25
feat(Dropdown): add Dropdown and ActionList API Decisions

saurabhdaware:
### [✨ Formatted Document ✨](https://github.com/razorpay/blade/blob/feat/select/api-decisions/packages/blade/src/components/Dropdown/_decisions/decisions.md)

TLDR:
image
Constraints*


packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
+# Dropdown & ActionList
+
+<br/>
+
+**Top-View Skeleton**
+
+```jsx
+<Dropdown>
+  <Dropdown.SelectInput label="Item Selection" />
```
kamleshchandnani, line 9: what happens if we don't provide `<Dropdown.SelectInput/>` and directly provide `<Dropdown.Overlay/>`?

saurabhdaware: We should throw an error ideally.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+};
+```
+
+### `ActionList.Section`
+
+Just an API that wraps section inside ActionList.
+
+This entire component is optional. So it may not be added in the ActionList that doesn't have any sections.
+
+It should also add divider after all items.
```
kamleshchandnani, line 355: do you mean add a divider once the section ends? basically after last item in the section?

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+```jsx
+<ActionList.Header title="Recent Searches" icon={HistoryIcon} />
+```
+
+Props
+
+```ts
+type ActionListHeaderProps = {
+  title: string;
+  icon: IconComponent;
```
kamleshchandnani, line 418: can we ask design if only these are 2 things that will ever exist in header?

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+
+- Dropdown is supposed to open when you focus on select and press `ENTER`, `SPACE` or any of the arrow keys. Arrow Up and Down can be pressed to move to next and previous items in the action list. Should close on pressing `ESC`.
+
+<br/><br/>
+
+## Open Questions
+
+### Dev
+
+- Vote: `variant` or `type` attribute on `Dropdown`?
```
kamleshchandnani, line 610: what if we have `selectionType: single | multiple` or `itemSelection: single | multiple`

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+
+  `<Dropdown variant="multiple" />` vs `<Dropdown type="multiple" />`
+
+- Should we call description/subtitle as `subtitle` everywhere, `description` everywhere, or change based on usage (E.g. See `ActionList.Item` component above).
```
kamleshchandnani, line 614: will vote for `subtitle` as we do similar at other places for eg: Cards.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+
+- `ActionList.Footer` `trailing` prop should allow all JSX components or limit to existing usecases (check Alternate API in [`ActionList.Footer` section](#actionlistfooter))
+
+- Should we give control to users to open/close dropdown? can't think of any usecase right now 🤔. We can do that using `useDropdown` API
```
kamleshchandnani, line 618: future enhancement 😅

kamleshchandnani (commented):
apart from other comments, some general comments

- do we allow typing things on select when it's focused? https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html
- if this is only desktop component then we shall think about adding `.desktop.tsx` to this

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+      leading={<StarIcon />}
+      trailing={<Text>⌘ + Space</Text>}
+    />
+  </ActionList.Section>
+  <ActionList.Item
+    title="Delete"
+    desciption="Delete this value"
+    value="delete"
+    intent="negative"
+    leading={<TrashIcon />}
```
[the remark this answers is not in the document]
saurabhdaware: There are usecases where we don't have any icon so we can't put TrashIcon by default 🤔

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+    trailing={<Text>⌘ + Space + D</Text>}
+  />
+  <ActionList.Footer
+    title="Footer Title"
+    description="Footer Description"
+    leading={<DocIcon />}
```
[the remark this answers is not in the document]
saurabhdaware: The reason I am calling it just `leading` is, `leadingVisual` sort of conveys that it's a "visual" which I would include icons, assets, images, etc. But in trailing for example we have a keyboard binding text which is a text which might not be just a visual.

Also in footer, we have `trailing` prop where we can pass an entire JSX of action buttons so that is also something I wouldn't consider as just "visual". Calling it `leading` and `trailing` leaves more space for us to generically use it.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+
+```jsx
+<ActionList>{/* Child components */}</ActionList>
+```
+
+Props
+
+```ts
+type ActionListProps = {
+  children: React.ReactNode;
```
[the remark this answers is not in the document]
saurabhdaware: didn't get 🤔? you mean on runtime checks?

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+   *
+   * @default neutral
+   */
+  intent?: 'negative' | 'neutral';
+
+  /** Click event on item */
+  onClick?: () => void;
+
+  /** URL. Should render item as `a` when this is passed */
+  href?: string;
```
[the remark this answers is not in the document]
saurabhdaware: ActionList can be used in navbars also so it can take users to some different links. Or there could be some action which takes people to different links.

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: yes

saurabhdaware:
> Is there a 1:1 mapping between web and native UI / UX?

By Native do you mean React Native? I'll have to wait for Dropdown Mobile Designs to figure out that but so far it seems like it will be very different component in itself. The `ActionList` might be reusable but Dropdown we will have to see. Ideally the same component should work.

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: I was actually favouring `description` over `subtitle` because that is what we use in `Alerts` and `subtitle` also conveys that it's a small title but in some cases it can be a 2-3 line information. So `descriptions` covers more usecases

saurabhdaware:
> do we allow typing things on select when it's focused? https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html

No. This will covered in AutoComplete component

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+- [Open Questions](#open-questions)
+- [Referrences](#referrences)
+
+---
+
+<br/>
+
+## Dropdown
+
+`Dropdown` will be a wrapper component which will handle the logic of opening and closing of the dropdown. It won't separately render anything in UI apart from the first children. (E.g. `Dropdown.SelectInput` in above example)
```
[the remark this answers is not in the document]
saurabhdaware: discussed in the office

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+  // ... FormLabelProps like `label`, `labelPosition`
+  // ... FormValidationProps like `validationState`, `helpText`, `errorText`, etc.
+  // ... State Props like `isRequired`, `isDisabled`,
+  icon: IconComponent;
+};
+```
+
+#### A11y
+
+Would be ideal if we can render it as `button` with `aria-expanded` prop to handle state of dropdown open and close
```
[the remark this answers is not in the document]
saurabhdaware: Added link to doc.

I haven't seen `aria-expanded` on any other component so far. Every dropdown and actionmenu I have referred to renders the top trigger as button only.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+
+### `Dropdown.Overlay`
+
+An Overlay that positions dropdown correctly.
+
+The animations for dropdown and Aria attributes like `aria-multiselectable="true"` `role="listbox"` like things can be added on Dropdown overlay here based on variant passed to `Dropdown` component.
```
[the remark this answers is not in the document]
saurabhdaware: ops it was outdated. removing it.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+Usage
+
+```jsx
+<ActionList.Item
+  title="Accept API"
+  value="accept"
+  desciption="Accept this API"
+  leading={<StarIcon />}
+  intent="negative"
+  trailing={<Text>⌘ + Space + A</Text>}
```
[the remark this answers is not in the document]
saurabhdaware: updated it to be similar like Cards

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: added it to open-questions

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: I think I'll prefer `type` in that case instead of introducing new prop. `type` also conveys the similar thing imo 🤔

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: As discussed, we can go with both of them depending on the usecase.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,672 @@
...
+#### API
+
+```jsx
+<Dropdown>
+  <Dropdown.SelectInput label="Click For Greetings" />
+  <Dropdown.Overlay>Hi</Dropdown.Overlay>
+</Dropdown>
+```
+
+It won't accept any props apart from children.
```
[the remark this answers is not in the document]
saurabhdaware: Added props heading with text _No Props Added_ to keep it consistent with rest of the doc

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,697 @@
...
+
+  description: string;
+
+  /** Value that we use in JS form. Similar to `value` in `Checkbox` */
+  value: string;
+
+  /**
+   * Leading Icon or assets. Accepts JSX Component
+   *
+   * Should not be allowed when ActionList variant is multiple select
```
kamleshchandnani, line 323: should be allowed na what is the harm in that?

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,697 @@
...
+   * Trailing Assets. Should ideally be ActionListItemText component for now
+   */
+  trailing?: React.ReactNode;
+
+  /**
+   * Changes item color to red
+   *
+   * @default neutral
+   */
+  intent?: 'negative' | 'neutral';
```
kamleshchandnani, line 337: is our neutral color same as feedback neutral? for eg: negative is red in tokens but is neutral color here same as feedback.neutra.text?

packages/blade/src/components/Dropdown/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: how will this behave when `href` and `onClick` both are passed?

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,697 @@
...
+   * Should not be allowed when intent is set to negative
+   */
+  isDefaultSelected?: boolean;
+
+  /** URL. Should render item as `a` when this is passed */
+  href?: string;
+};
+```
+
+### `ActionListSection`
```
kamleshchandnani, line 354: i hope to make sure all these subcomponents will be used under the context of allowed parents we will be adding context?

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,697 @@
...
+>
+> API yet to be finalized
+
+Footer of the ActionList
+
+![](2022-11-29-17-40-17.png)
+
+#### Design
+
+- [Figma - ActionListItem / Footer](https://www.figma.com/file/sAdplk2uYnI2ILnDKUxycW/Blade---Banking-Dark?node-id=13136%3A378703&t=RoEboUS3qWdStKK6-0)
```
kamleshchandnani, line 439: the action seems to be full width in leading example but with trailing it seems to be a non-full width variant. is that intentional or there's some criteria to switch between full width vs non-fullwidth?

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,697 @@
...
+
+**Props**
+
+```ts
+type ActionListFooter = {
+  title?: string;
+
+  description?: string;
+
+  /** Leading Asset. Ideally Icon but can be any JSX component */
```
kamleshchandnani, line 481:
```suggestion
  /** Leading Asset. Ideally ActionListFooterIcon */
```

??

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,697 @@
...
+
+- Dropdown is supposed to open when you focus on select and press `ENTER`, `SPACE` or any of the arrow keys. Arrow Up and Down can be pressed to move to next and previous items in the action list. Should close on pressing `ESC`.
+
+<br/><br/>
+
+## Open Questions
+
+### Dev
+
+- Vote: `variant` or `type` attribute on `Dropdown`? `<Dropdown variant="multiple" />` vs `<Dropdown type="multiple" />`
```
kamleshchandnani, line 623: how about `selectionType= single | multiple`?

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,697 @@
...
+
+<br/>
+
+**Top-View Skeleton**
+
+```jsx
+<Dropdown>
+  <SelectInput label="Item Selection" />
+  <DropdownOverlay>
+    <ActionList>
```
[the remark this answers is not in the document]
saurabhdaware: yup can move it to separate directory after review. It was easier to write this way.

packages/blade/src/components/Dropdown/_decisions/decisions.md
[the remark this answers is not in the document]
saurabhdaware: yes right we can block it on runtime

packages/blade/src/components/Dropdown/_decisions/decisions.md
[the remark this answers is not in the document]
saurabhdaware: like an anchor tag with onClick event on it. `href` is the deciding factor. If it is present then its `a` or its `buttont`

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,697 @@
...
+```ts
+type ActionListProps = {
+  children: React.ReactNode;
+
+  /**
+   * Renders simple menu or checkboxes depending on the variant
+   *
+   * @default single
+   */
+  variant?: 'single' | 'multiple';
```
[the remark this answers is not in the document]
saurabhdaware: https://github.com/razorpay/blade/pull/885#discussion_r1044277239 what do you think about this?

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: In multiple select the checkbox appears in leading. Having an icon after the checkbox just pushes actual text to right which looks strange. So this was a design call.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,697 @@
...
+    value="delete"
+    intent="negative"
+    leading={<ActionListItemIcon icon={TrashIcon} />}
+    trailing={<ActionListItemText>⌘ + Space + D</ActionListItemText>}
+  />
+  <ActionListFooter
+    title="Footer Title"
+    description="Footer Description"
+    leading={<ActionListFooterIcon icon={DocIcon} />}
+    trailing={
```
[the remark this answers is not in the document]
saurabhdaware: So it can't be similar to Cards because in Cards we just have primary and secondary actions wheras in this footer, The footer in dropdown has too many variations. 

image

Supporting these variations with stricter API like that just makes the API really really complicated. I have added what the API would look like with props in Alternate Footer API in ActionListFooter section.

image

packages/blade/src/components/Dropdown/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: yeah true that and i also saw a use case yesterday where the design was using sign out as negative intent and the icon was shown of signout

packages/blade/src/components/Dropdown/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: i think we shall confirm with design once for this because they haven't created a link actionlist item which might look different than our regular link. so if they haven't then maybe we shall add this prop later once they add

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,697 @@
...
+<ActionList
+  surfaceLevel={2}
+  onChange={(e) => {
+    console.log({ name: e.name, values: e.values });
+  }}
+>
+  <ActionListItem
+    title="Accept API"
+    value="accept"
+    desciption="Accept this API"
```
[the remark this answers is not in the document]
kamleshchandnani: this is pending 🙈

packages/blade/src/components/Dropdown/_decisions/decisions.md
kamleshchandnani: ?

kamleshchandnani (approved):
lgtm ✅

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: changed it to selectionType since that is what we use in design as well

packages/blade/src/components/Dropdown/_decisions/decisions.md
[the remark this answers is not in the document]
saurabhdaware: how did I typo at 7 places lol. changing it.

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: changed it to selectionType since that is what we use design as well and makes sense.

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: yup we'll put some validations on these components

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: It becomes full width when there is not footer title, description or other things.

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: ohhh nope its surface.text.normal. Lemme discuss this with design and see if we should some use some other prop instead of intent here then 🤔

saurabhdaware: Actually it seems valid to have surface.text.normal in normal so I'll change it here.

saurabhdaware: TLDR: so I had added `<ActionListItem intent="negative" />` which uses feedback negative token and we have the default variant which uses `surface.text.normal`

we don't exactly have intent="normal" anywhere else so I changed it to variant="negative | normal"

we had [intent vs variant discussion](https://github.com/razorpay/blade/blob/89cc8babfe8d37a29354f7f3bc7689a556b1bcc7/packages/blade/src/components/Badge/_decisions/decisions.md#variant-prop) during badge API decisions so referring to that for this where we decided to use intent for semantic values and variant for everything else.

kamleshchandnani: what if we just have `intent?: negative` so if it's undefined we set it to normal by default and if someone passes intent then the only value we support right now is negative.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,702 @@
...
+  // ... FormLabelProps like `label`, `labelPosition`
+  // ... FormValidationProps like `validationState`, `helpText`, `errorText`, etc.
+  // ... State Props like `isRequired`, `isDisabled`,
+  icon: IconComponent;
+};
+```
+
+#### A11y
+
+Would be ideal if we can render it as `button` with `aria-expanded` prop to handle state of dropdown open and close.
```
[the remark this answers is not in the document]
saurabhdaware: I think the `AutoComplete` component would qualify as combobox. In this particular case, we can't really type inside the input. It's just a button that opens and closes the menu.

image

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role

What we have here seems closer to [menu role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role) which seems to recommend button in every example I saw so far.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,702 @@
...
+
+### API
+
+#### Simplest Usage Possible
+
+```jsx
+<ActionList
+  surfaceLevel={2}
+  onChange={(e) => {
+    console.log({ name: e.name, values: e.values });
```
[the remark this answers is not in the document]
saurabhdaware: So the reason I kept it as `values` is because `ActionList` has multiple select variant, in that case it becomes closer to `CheckboxGroup` 

In `CheckboxGroup` we use `values` because we get array of all selected values.

I didn't really want to change it from `values` to `value` for single variant because it can break the UI if someone just changes the variant.

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: hmm yes that would also make sense. Lets go with that only then.

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,703 @@
...
+    description="Accept this API"
+    leading={<ActionListItemIcon icon={StarIcon} />}
+    trailing={<ActionListItemText>⌘ + Space + A</ActionListItemText>}
+    onClick={() => {}}
+  />
+  <ActionListItem
+    title="Delete"
+    description="Delete this value"
+    value="delete"
+    intent="negative"
```
nashcheez, line 152: Can we have an `ActionListItem` with `intent="negative"` when ActionListProps.selectionType is `multiple`? If yes, we need to check with design as this would be a bad practice.

nashcheez (commented):
@saurabhdaware Great stuff, this API decision doc is 🤯

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: Yes the last `ActionListSection` item in all scenarios should not have a divider. We'll have to take care of that.

packages/blade/src/components/Dropdown/_decisions/decisions.md
```diff
@@ -0,0 +1,703 @@
...
+
+### Custom Implementation instead of Library
+
+> **Note**
+>
+> Didn't get too much into implementation details during API decisions. Would recommend evaluating this once again while implementing the Dropdown.
+
+TLDR; Felt like our Dropdown was a bit more complicated than other dropdowns we've seen so we might have to fight against existing opinions from libraries if we go with libraries.
+
+Some libraries we evaluated
```
anuraghazra, line 590: Another candidate: https://ariakit.org/components/select

We explored ariakit on a11y RFC as well. 

Otherwise downshiftjs is a good option too, but might require extra effort to integrate because downshiftjs starts with a data based approach to rendering the dropdowns where in other hand our component API is purely declarative.

packages/blade/src/components/Dropdown/_decisions/decisions.md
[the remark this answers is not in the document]
saurabhdaware: Aha got it. I was confused because the voiceover asks you to "type" words when you're focussing on combobox. 

image

But w3 link mentions this - 

> However, like an HTML <select>, users can type characters to select matching options.

So I guess they mean typing words to jump on to that option and not exactly typing it inside the input.

I think some other libraries were using menu and not doing this ^^. We can probably do it though. Will update the API decisions

> Would be ideal if we don't handle it ourselves. 

I am really really hoping some library works out lol. So far none are properly working 😢

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: Ariakit Select doesn't seem ready yet. And yup you're right downshift is a bit complex to integrate because of data approach.

anuraghazra: Yeah ariakit is in alpha although we could do a POC. We did a POC while we wrote the a11y RFC about ariakit (that time it didn't even had docs website :p): 

https://github.com/razorpay/blade/blob/master/rfcs/2022-04-09-accessibility.md#platform-specific-implementation--1

packages/blade/src/components/Dropdown/_decisions/decisions.md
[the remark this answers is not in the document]
saurabhdaware: updated it to use combobox. Thanks!

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: Oh yup we can refer to it for a11y practices I guess. For now I have added links to respective w3.org since that covers most of the things.

packages/blade/src/components/Dropdown/_decisions/decisions.md
saurabhdaware: confirmed from designer that it we don't have usecase for negative intent in multi select. I have updated the doc

PR #887, opened 2022-11-25
fix(Checkbox): allows Checkbox to accept `childern` prop of type `React.ReactNode`

### LINKED ISSUE #787: Checkbox doesn't accept Link component as children
We can have usecases where consumers would want to add Link to checkbox. E.g. terms and conditions checkbox. 

Our `Checkbox` component doesn't accept anything but string right now so it throws a typescript error.

image

k-rajat19:
Currently, Checkbox only accepts `childern` of type `string`
but we have use cases where we may want to pass multiple elements, not just only string, as discussed in #787 

Resolves #787

### DIFF
--- a/packages/blade/src/components/Checkbox/__tests__/Checkbox.web.test.tsx
+++ b/packages/blade/src/components/Checkbox/__tests__/Checkbox.web.test.tsx
@@ -3,16 +3,20 @@
 import userEvents from '@testing-library/user-event';
 import React from 'react';
 import { Checkbox } from '../Checkbox';
+import { Link } from '~components/Link';
 import renderWithTheme from '~src/_helpers/testing/renderWithTheme.web';
 
 describe('<Checkbox />', () => {
   it('should render checkbox with label', () => {
-    const labelText = 'Remember password';
-    const { container, getByText, getByRole } = renderWithTheme(<Checkbox>{labelText}</Checkbox>);
+    const label = (
+      <>
+        I accept <Link>Terms and Conditions</Link>
+      </>
+    );
+
+    const { container, getByRole } = renderWithTheme(<Checkbox>{label}</Checkbox>);
     expect(container).toMatchSnapshot();
-    // the name attribute here is the accessibility label name and not the name of the input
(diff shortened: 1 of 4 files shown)


saurabhdaware:
So we intentionally try to limit the type to bare minimum to be a bit restrictive on what people can pass. We're yet to discuss on whether we have genuine usecases of needing nested components inside checkbox or not. 

As the issue is still in discussion I think we can hold this off until that discussion is complete

saurabhdaware:
We discussed this internally and we can have usecases of having `Link` in `Checkbox` so we can continue with PR reviews 🎉 

Thanks for the PR!!

saurabhdaware (approved):
Looks good to me 👯

packages/blade/src/components/Checkbox/__tests__/Checkbox.web.test.tsx
```diff
@@ -3,16 +3,20 @@
...
+import { Link } from '~components/Link';
 import renderWithTheme from '~src/_helpers/testing/renderWithTheme.web';
 
 describe('<Checkbox />', () => {
   it('should render checkbox with label', () => {
-    const labelText = 'Remember password';
-    const { container, getByText, getByRole } = renderWithTheme(<Checkbox>{labelText}</Checkbox>);
+    const label = (
+      <>
+        I accept <Link>Terms and Conditions</Link>
```
anuraghazra, line 13: Would be better if we add the href in here.

anuraghazra (changes requested):
Mostly LGTM, One small change.

packages/blade/src/components/Checkbox/__tests__/Checkbox.web.test.tsx
k-rajat19: done :)

k-rajat19:
ups...missed updating the snapshot

anuraghazra (approved):

saurabhdaware (approved):

saurabhdaware:
Thanks for the contributions 💃🏼

PR #888, opened 2022-11-28
fix(blade-pr-title-check): use github token instead of ci token

saurabhdaware:
It was failing in #887 because CI_TOKEN is not accessible to people outside of Razorpay

### DIFF
--- a/.github/workflows/blade-pr-title-check.yml
+++ b/.github/workflows/blade-pr-title-check.yml
@@ -1,8 +1,8 @@
 name: Blade PR Title Check
 
-on: 
+on:
   pull_request:
-     types: [open, edited, synchronize]
+    types: [open, edited, synchronize]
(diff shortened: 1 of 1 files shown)


.github/workflows/blade-pr-title-check.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     name: Check
     runs-on: ubuntu-latest # nosemgrep: non-self-hosted-runner
     env:
-      GITHUB_TOKEN: ${{ secrets.CI_BOT_TOKEN }}
+      GITHUB_TOKEN: ${{ github.token }}
```
saurabhdaware, line 16: Only this is changed. Rest are formatting changes.

chaitanyadeorukhkar (approved):

PR #889, opened 2022-11-28
feat(examples): add with-styled-components example

divyanshu013:
Code was scaffolded with `create-react-app`, mostly auto generated code.

This also acts as a repro for the issue with nested themes

image


chaitanyadeorukhkar:
Since this is a issue reproduction example. Can we name it more explicitly that this is an example for an issue? `merged-theme-style-components-bug`?

divyanshu013:
Was thinking this example can live here as it is for reference, though temporary usage is as bug repro. Later we can add different other examples like `with-nextjs`, etc.

saurabhdaware:
Should we call this `with-multiple-theme-providers`? because `basic` example is also using `styled-components`. Essentially all consumers will be using styled-components so it can be bit confusing.

divyanshu013:
Will update this example to use a theme key

examples/with-styled-components/src/App.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { ThemeProvider } from 'styled-components';
+import '@fontsource/lato/400.css';
+import '@fontsource/lato/400-italic.css';
+import '@fontsource/lato/700.css';
+import '@fontsource/lato/700-italic.css';
+
+import './app.css';
+
+const customTheme = {
+  colors: {
```
kamleshchandnani, line 12: aren't we showing how to avoid these conflicts? this is actually the bug wala scenario right?

divyanshu013: Yup, earlier we were thinking of adding theme key to blade new. Now this repro isn't needed since we're fixing on blade-old. This example needs to be updated to provide a theme key into this `customTheme` object.

divyanshu013: Added in https://github.com/razorpay/blade/pull/889/commits/75177f0affe2fd32564bd678e2532571c57f3303

kamleshchandnani:
> Should we call this `with-multiple-theme-providers`? because `basic` example is also using `styled-components`. Essentially all consumers will be using styled-components so it can be bit confusing.

+1 to this

divyanshu013:
Updated to `with-multiple-theme-providers`

examples/with-multiple-theme-providers/README.md
```diff
@@ -0,0 +1,7 @@
+# Blade with `styled-components`
+
+When using Blade along with multiple theme providers (for example with `blade-old` or your own custom theming) you should namespace your own theme so it doesn't conflict with blade's theme tokens.
```
chaitanyadeorukhkar, line 3: Should we add some more details here like what issue it causes and how it merges both themes and how adding a namespace would solve the issue? Maybe attach a link to some `styled-components` issue that talks about the same?>

examples/with-multiple-theme-providers/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@types/react-dom": "^18.0.9",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-scripts": "5.0.1",
+    "typescript": "^4.9.3",
+    "web-vitals": "^2.1.4",
+    "@razorpay/blade": "../../packages/blade/build/",
+    "@fontsource/lato": "^4.5.10",
+    "styled-components": "^5.3.6",
+    "@types/styled-components": "5.1.25"
```
chaitanyadeorukhkar, line 21: nit: use exact package versions

chaitanyadeorukhkar (approved):

examples/with-multiple-theme-providers/package.json
divyanshu013: Done in https://github.com/razorpay/blade/pull/889/commits/ed10d115f7b9898ba8e3a5cb236ef463a1496cb5

examples/with-multiple-theme-providers/README.md
divyanshu013: Added in https://github.com/razorpay/blade/pull/889/commits/2652f7bbf35f8192b5f786b8ba8d5e8a468c34e7

kamleshchandnani:
aren't we rendering this on storybook? like under recipes?

kamleshchandnani:
> Not yet, as an enhancement, we can perhaps add codesandbox examples of these directly from github path since these were built from CRA.

i would suggest to put this as recipe because almost all the examples that we put shall go as part of recipes. imagine as a consumer its very difficult for me to navigate the storybook for few things and then come to github repo for other things.

divyanshu013:
We can add in recipes but currently recipes are usage examples. This is more of an integration example, we can expand this by adding more integration examples with popular libraries and frameworks - example with nextjs, with express, with remix, etc.

kamleshchandnani:
i see. but the main concern is discoverability we need it somehow to be present on storybook. maybe we add sandpack things like we do it for other stories? or maybe just a `mdx` file like we have for other guides?

divyanshu013:
Sounds good. I think we can add an integration examples page which links to these examples

divyanshu013:
@kamleshchandnani added a docs page in https://github.com/razorpay/blade/pull/889/commits/2d3e816540626fc19e2353ea88b6ef82eeedfbf9

packages/blade/docs/guides/IntegrationExamples.stories.mdx
```diff
@@ -0,0 +1,10 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Guides/Integration examples" />
+
+# Integration examples
+
+You may checkout some examples on integrating `blade` in your project:
```
kamleshchandnani, line 7: @divyanshu013 i was mostly saying on the lines of adding a live playground or probably a codesandbox embed 😞 . For now we can merge this but for other upcoming integrations we shall create a directory on codesandbox too for such integrations

kamleshchandnani (approved):

divyanshu013:
I was trying to map our examples directory to codesandbox. Screenshot:

image

But something seems wrong, codesandbox gets stuck at installing dependencies step. Also, we'll need to update our examples to use the correct `@razorpay/blade` dependency. Right now it's mapped to pick blade locally from `build`. Creating a separate issue to track this.

Note: we can't have codesandbox for all examples (like nextjs) so those would need to remain without embeds anyway.

PR #890, opened 2022-11-28
feat: use separate eslint config for example

saurabhdaware:


saurabhdaware:
Closing this in favour of #891

PR #891, opened 2022-11-28
fix(examples/basic): use CRA tsconfig and eslint

saurabhdaware:
We were trying to use common eslint from monorepo earlier but it has some conflicts with eslint and tsconfig of CRA


divyanshu013 (approved):
Looks good to me.

PR #893, opened 2022-11-28
fix: allows `OTPInput` to accept only numeric values

### LINKED ISSUE #786: OTP Field is accepting text charcters
Repro: https://codesandbox.io/s/otp-accepts-text-blade-o9xp48?file=/src/App.tsx

OTPInput is also accepting text characters as input. Ideally we should set `type="number"` on the input fields to not allow these characters.

image

k-rajat19:
Resolves #786

### DIFF
--- a/packages/blade/src/components/Input/OTPInput/OTPInput.tsx
+++ b/packages/blade/src/components/Input/OTPInput/OTPInput.tsx
@@ -243,6 +243,7 @@ const OTPInput = ({
             errorText={errorText}
             helpText={helpText}
             hideFormHint={true}
+            type="numeric"
           />
         </Box>,
       );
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/Input/OTPInput/OTPInput.tsx
```diff
@@ -243,6 +243,7 @@ const OTPInput = ({
             errorText={errorText}
             helpText={helpText}
             hideFormHint={true}
+            type="numeric"
```
divyanshu013, line 246: How does this affect the UI? Can you post some screenshots please? Does it also show the number increase number decrease controls?

k-rajat19:
closed in favor of #898

PR #894, opened 2022-11-29
docs: Cards API Decision

anuraghazra:
## [See Formatted Document](https://github.com/razorpay/blade/blob/anu/card-api/packages/blade/src/components/Card/_decisions/decisions.md)


packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,265 @@
...
+With the above constraints we can propose the Card API as follows: 
+
+Sample usage:
+
+```jsx
+<Card backgroundLevel={1}>
+  <Card.Header 
+    title="Payments Links" 
+    subtitle="Share payment link via an email, SMS, messenger, chatbot etc." 
+    titlePrefix={<DollarIcon size="xl" />}
```
anuraghazra, line 106: One thing I notice while implementing the Card is that in the titlePrefix icon the user won't just need to pass `size` but also `color`

```jsx
      <CardComponent.Header
        title="Card Header"
        subtitle="Subtitle"
        titlePrefix={<AlertOctagonIcon color="surface.text.muted.lowContrast" size="xlarge" />}
        titleSuffix={<Counter value={12} />}
      />
```

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,265 @@
...
+| `subtitle`       | `string`          | `undefined` | Subtitle of the Card                                     |          |
+| `titlePrefix`    | `React.ReactNode` | `undefined` | Prefix element placed before title text (eg: Icon)       |          |
+| `titleSuffix`    | `React.ReactNode` | `undefined` | Suffix element placed after title text (eg: Counter)     |          |
+| `trailingVisual` | `React.ReactNode` | `undefined` | Trailing visual element placed on right side of the card |          |
+
+### `Card.Footer` API
+
+| Prop       | Type                                     | Default     | Description                                | Required |
+| ---------- | ---------------------------------------- | ----------- | ------------------------------------------ | -------- |
+| `title`    | `string`                                 | `undefined` | Title of the Card                          | ✅        |
```
chaitanyadeorukhkar, line 148: We have use-cases where there are only action buttons in the footer. Let's make title optional?

Screenshot 2022-11-30 at 3 54 33 PM

```suggestion
| `title`    | `string`                                 | `undefined` | Title of the Card                          |        |
```

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,265 @@
...
+| `subtitle` | `string`                                 | `undefined` | Subtitle of the Card                       |          |
+| `actions`  | `{ primary: Action, secondary: Action }` | `undefined` | Renders a primary/secondary action buttons |          |
+
+```ts
+type Action = {
+  onClick: () => void;
+  text: string;
```
chaitanyadeorukhkar, line 155: We should expose the following props too:

1. `variant` (Unsure about this but there is a use-case for tertiary button in the design)
2. `type` ( Could be needed for reset or submit buttons in a card)
3.  `accessibilityLabel`
4. `isLoading`
5. `isDisabled`
6. `icon` & `iconPosition`?

chaitanyadeorukhkar (commented):
Few comments, otherwise Lgtm 👍

packages/blade/src/components/Card/_decisions/decisions.md
anuraghazra: Good catch!

packages/blade/src/components/Card/_decisions/decisions.md
anuraghazra: 🤔 We can just expose variant, isLoading, type for now? 
Because not sure if we will even have a icon only button usecase in the footer actions.

packages/blade/src/components/Card/_decisions/decisions.md
kamleshchandnani: we have a restriction to use only the large icon but this is not easily achievable with the current API we are just relying on consumers

kamleshchandnani (commented):
overall i don't feel very confident about the API. It's too restrictive for some use cases and too  flexible for some use case which will leads to huge learning curve for consumers. let's discuss this

divyanshu013 (commented):
Think we need to update this based on latest discussions.

packages/blade/src/components/Card/_decisions/decisions.md
anuraghazra: Resolved with the new API.

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,265 @@
...
+### `Card.Header` API
+
+
+| Prop             | Type              | Default     | Description                                              | Required |
+| ---------------- | ----------------- | ----------- | -------------------------------------------------------- | -------- |
+| `title`          | `string`          | `undefined` | Title of the Card                                        | ✅        |
+| `subtitle`       | `string`          | `undefined` | Subtitle of the Card                                     |          |
+| `titlePrefix`    | `React.ReactNode` | `undefined` | Prefix element placed before title text (eg: Icon)       |          |
+| `titleSuffix`    | `React.ReactNode` | `undefined` | Suffix element placed after title text (eg: Counter)     |          |
+| `trailingVisual` | `React.ReactNode` | `undefined` | Trailing visual element placed on right side of the card |          |
```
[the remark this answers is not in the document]
anuraghazra: Resolved with the new API

packages/blade/src/components/Card/_decisions/decisions.md
anuraghazra: Done!

packages/blade/src/components/Card/_decisions/decisions.md
anuraghazra: Resolved with the new API.

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+- They can also use either variants of buttons `primary` or `secondary`
+
+## API
+
+With the above constraints we can propose the Card API as follows: 
+
+Sample usage:
+
+```jsx
+<Card backgroundLevel={1}>
```
divyanshu013, line 111: Do you mean

```suggestion
<Card surfaceLevel={1}>
```

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+  - If someone just uses the Card even if they are not using any of the other components like Button, IconButton, Badge their app will bloat with the components
+- We also discussed few approaches on how we can validate the components and throw a runtime error if an unallowed component is passed to the Card. 
+  - We cannot use displayName of React components since it will be stripped of on production
+  - We also cannot do `comp.type === CardHeaderCounter` because that way we anyway have to import the component, resulting in bloated bundle
+
+
+
+**Action Items:** 
+
+- Try adding a static method to the Card components to determine if it’s an allowed component or not. We can’t use displayName since it will be stripped off.
```
divyanshu013, line 504: This needs to be updated I think with the latest approach.

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+| ---------- | ---------------------------------------- | ----------- | ------------------------------------------ | -------- |
+| `title`    | `string`                                 | `undefined` | Title of the Card                          | ✅        |
+| `subtitle` | `string`                                 | `undefined` | Subtitle of the Card                       |          |
+
+### `Card.FooterTrailing` API
+
+| Prop       | Type                                     | Default     | Description                                | Required |
+| ---------- | ---------------------------------------- | ----------- | ------------------------------------------ | -------- |
+| `actions`  | `{ primary: Action, secondary: Action }` | `undefined` | Renders a primary/secondary action buttons |          |
+
```
divyanshu013, line 178: What about `Card.Header`, `Card.Footer` and `Card.Body`?

divyanshu013: Do these also accept any JSX as children? For customizing header, footer etc.?

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+| `surfaceLevel` | `2, 3` | `2`     | Surface level of the card background color, use this based on where the card is placed |          |
+
+
+### `Card.HeaderLeading` API
+
+
+| Prop             | Type              | Default     | Description                                              | Required |
+| ---------------- | ----------------- | ----------- | -------------------------------------------------------- | -------- |
+| `title`          | `string`          | `undefined` | Title of the Card                                        | ✅        |
+| `subtitle`       | `string`          | `undefined` | Subtitle of the Card                                     |          |
```
divyanshu013, line 156: In Alert we're using `description`. Should we follow similar convention here?

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+| `prefix`    | `React.ReactNode` | `undefined` | Prefix element placed before title text (restricted to: Icon)       |          |
+| `suffix`    | `React.ReactNode` | `undefined` | Suffix element placed after title text (restricted to: Counter)     |          |
+
+### `Card.HeaderTrailing` API
+
+| Prop             | Type              | Default     | Description                                              | Required |
+| ---------------- | ----------------- | ----------- | -------------------------------------------------------- | -------- |
+| `visual` | `React.ReactNode` | `undefined` | Trailing visual element placed on right side of the card (restricted to: Badge, Text, Link, IconButton) |          |
```
divyanshu013, line 164: `visual` seems bit odd terminology here (prefix, suffix have been used earlier).

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+    prefixIcon={CloseIcon}
+    prefixAvatar={}
+    counter={12}
+    counterIntent=""
+    trailingVisual={<Badge />}
+  />
+</Card>
+```
+
+**API 7: Final API**
```
kamleshchandnani, line 456: now API 8? where we move away from the static attributes name checking? 😄

anuraghazra: 😆 haha yeah!

packages/blade/src/components/Card/_decisions/decisions.md
divyanshu013: True that, how about using `children` here?

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+
+*Cons:*
+
+- Runtime enforcement is very hard to do.
+- API is verbose, and the learning curve is too large
+- The consumer would need to learn where and how to put the JSX structure to build the card
+- The API is too flexible, we want some control over the API to reduce the footguns.
+- We need a few runtime hacks and checks on the `Card.headerLeading` part to make sure we are putting everything on the right place.
+
+```jsx
```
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+- Users can totally omit action buttons
+- They can also use either variants of buttons `primary` or `secondary`
+
+## API
+
+With the above constraints we can propose the Card API as follows: 
+
+Sample usage:
+
+```jsx
```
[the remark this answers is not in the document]
anuraghazra: Right! Yes i've added the discussion notes at the end on why we switched from dot notation to component imports.

packages/blade/src/components/Card/_decisions/decisions.md
anuraghazra: Discussed about this with Rastogi. 

Conclusion: 

subtitle & description conveys different context based on different components. 

In Alert description the the body text which can be long form content that's why contextually the naming `description` makes sense
But in CardHeader `subtitle` makes more sense since it will not have any long form content more than 1 or 2 lines.

packages/blade/src/components/Card/_decisions/decisions.md
anuraghazra: Sorry could not understand, can you elaborate?

packages/blade/src/components/Card/_decisions/decisions.md
anuraghazra: Yeah i've updated the API, but kept this here for archiving the discussions we had while developing the API.

packages/blade/src/components/Card/_decisions/decisions.md
anuraghazra: Fixed.

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+  - Size & text variant cannot be changed
+- `Icon`
+  - Size & color cannot be changed
+- `Counter`
+  - Size can be changed but not recommended
+  - Variant can be changed
+
+**Header Trailing Section**
+
+Anything can be changed in these components, but whats given out of the box are the recommended:
```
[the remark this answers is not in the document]
anuraghazra: Removed.

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+## Card API Constraints 
+
+**Header Leading Section**
+
+- `Title` & `Subtitle`
+  - Size & text variant cannot be changed
+- `Icon`
+  - Size & color cannot be changed
+- `Counter`
+  - Size can be changed but not recommended
```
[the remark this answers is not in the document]
anuraghazra: Changed it to `Size cannot be changed`

packages/blade/src/components/Card/_decisions/decisions.md
divyanshu013: I mean would users be able to pass any JSX inside the subcomponents? Example:

```tsx
<Card.Header>
  <MyJSX />
</Card.Header>
```

anuraghazra: Yeah they can. 
Technically we can enforce this on the runtime, but don't think it's necessary cuz that's on them if they want to opt out of the header and add their own stuff.

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,512 @@
...
+
+## API
+
+With the above constraints we can propose the Card API as follows: 
+
+Sample usage:
+
+```jsx
+<Card backgroundLevel={1}>
+  <Card.Header>
```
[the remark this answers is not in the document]
anuraghazra: Yeah we can do that.

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -0,0 +1,590 @@
...
+</Card>
+```
+
+
+### `Card` API
+
+
+| Prop           | Type   | Default | Description                                                                            | Required |
+| -------------- | ------ | ------- | -------------------------------------------------------------------------------------- | -------- |
+| `surfaceLevel` | `2, 3` | `2`     | Surface level of the card background color, use this based on where the card is placed |          |
```
kamleshchandnani, line 159: why can't we have cards on surface level 1? can you confirm with design once? usually the cards will start on top of surface level 1

kamleshchandnani (approved):
just one tiny comment. rest lgtm

packages/blade/src/components/Card/_decisions/decisions.md
anuraghazra: Yes saurav confirmed this earlier, cards will have surface level 2 and 3 only. Since most backgrounds where cards will be added will be starting from 1 only. So if the background is level1 card will be level2.

Issue #701, opened 2022-09-08
plugin dts: syntax not yet supported error in build


kamleshchandnani:
**Issue:** aliased imports from `~components/Form` breaks the type generation

**Workaround(not ideal but unblocking release for now):** replace alias imports with relative imports `~components/Form` ➡️ `../../Form`

**Ideal solution:** the alias imports shall work too

Temporary workaround in https://github.com/razorpay/blade/pull/700

divyanshu013:
For now I think we should add a linting rule to disallow the problematic import /cc @anuraghazra

anuraghazra:
Faced this bug while working on BottomSheet too. 

I think I got the pattern to fix this bug. 

While building rollup will tell you have some random component syntax is broken, that means you are importing that component's type via the aliased import.

anuraghazra:
rollup-plugin-dts is failing in weird places & throwing syntax errors, wastes a lot of time trying to pin point the error. 

For example: 

This throws syntax error:
```jsx
<GestureHandlerRootView style={{ flex: 1 }}>
```
But this doesnt:
```jsx
<GestureHandlerRootView style={gestureHandlerStyle}>
```

**We should consider alternatives like [rollup-plugin-ts](https://github.com/wessberg/rollup-plugin-ts)**

saurabhdaware:
Also faced this while refactoring docs. Went on my own exploration unaware of this issue lol. We should definitely pick this soon 😭

divyanshu013:
No. of hours wasted on this issue += 2

It also throws error when you try to do `import { X as Y }` and then trying to export the same module

chaitanyadeorukhkar:
Number of hours wasted on this issue += 2 😭 

It also throws error when you try to import from the `src/components/index.js` file which re-exports all our components. Most of the times, this import is done unintentionally by vscode IntelliSense. Ref commit: https://github.com/razorpay/blade/pull/1258/commits/083d4c13da44e3bc6bfe260a4f74bb10bdfa294d

<img width="1384" alt="Screenshot 2023-06-27 at 12 34 44 PM" src="https://github.com/razorpay/blade/assets/24487274/7971f0c3-2a96-47fa-a5e8-6a33892a6dc7">

anuraghazra:
Number of hours wasted += 2

I had this code which was breaking: 

```jsx
const _BaseBox = React.forwardRef((props, ref) => {
  const { theme } = useTheme();
[...]

PR #896, opened 2022-11-30
feat: added Card component

anuraghazra:
Screenshots: 

Desktop Web: 

image

Mobile Web: 

image

React Native: 

image


packages/blade/ios/blade.xcodeproj/project.pbxproj
```diff
@@ -15,8 +15,8 @@
 		7EF68E3733C33B6898317E18 /* libPods-blade-bladeTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = ABFE59519B596E51CEFDCCC0 /* libPods-blade-bladeTests.a */; };
 		81AB9BB82411601600AC10FF /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 81AB9BB72411601600AC10FF /* LaunchScreen.storyboard */; };
 		BE94BDCFE4A742E89E116690 /* Lato-Regular.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 4C0A80DCDFF54BF28ED33285 /* Lato-Regular.ttf */; };
-		F31095E142B84AD0A9E345EB /* Lato-Bold.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 856B36C0CBEA4725AD05BF92 /* Lato-Bold.ttf */; };
 		DD50086558394F9D8B0F07DA /* Lato-Italic.ttf in Resources */ = {isa = PBXBuildFile; fileRef = A2D9B3E1EEE34EC9AD42AA9C /* Lato-Italic.ttf */; };
+		F31095E142B84AD0A9E345EB /* Lato-Bold.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 856B36C0CBEA4725AD05BF92 /* Lato-Bold.ttf */; };
```
anuraghazra, line 19: XCode auto updated these.

packages/blade/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,260 @@
...
+    },
+  },
+  parameters: {
+    docs: {
+      page: Page,
+    },
+  },
+} as Meta<StoryControlProps>;
+
+const CardTemplate = ({ ...args }: StoryControlProps): React.ReactElement => {
```
anuraghazra, line 232: Only added 1 story template, since there are multiple combinations of header, footer props don't think it will be feasible to create separate story examples for each combination. 

Thus created a customisable story.

packages/blade/src/components/Card/CardHeader.tsx
```diff
@@ -0,0 +1,194 @@
...
+      </Box>
+    </Box>
+  );
+};
+
+type CardHeaderTrailingProps = {
+  visual?: React.ReactNode;
+};
+
+const headerTrailingAllowedComponents = [
```
anuraghazra, line 163: This array holds the `componentId (string)` of the allowed components, not the components itself. Then we check it via `getComponentId()`

packages/blade/src/components/Card/CardSurface.web.tsx
```diff
@@ -0,0 +1,20 @@
+import styled from 'styled-components';
+import Box from '~components/Box';
+
+const CardSurface = styled(Box)<{ surfaceLevel: 2 | 3 }>(({ surfaceLevel, theme }) => {
+  const offsetX = theme.shadows.offsetX.level[1];
+  const offsetY = theme.shadows.offsetY.level[1];
+  const blur = theme.shadows.blurRadius.level[1];
+  const shadowColor = theme.shadows.color.level[1];
+
+  const shadow1 = `${offsetX}px ${offsetY}px ${blur}px 0px ${shadowColor}`;
```
anuraghazra, line 10: Might be better to have makeShadow() function? or composite shadow tokens?

packages/blade/src/components/Card/CardHeader.tsx
```diff
@@ -0,0 +1,194 @@
...
+import type { BadgeProps } from '~components/Badge';
+import { Badge } from '~components/Badge';
+import type { LinkProps } from '~components/Link';
+import { Link } from '~components/Link';
+import type { ButtonProps } from '~components/Button';
+import { Button } from '~components/Button';
+import { minHeight } from '~components/Button/BaseButton/buttonTokens';
+import { makeSpace } from '~utils';
+
+type CardHeaderProps = {
```
kamleshchandnani, line 20: nit: move this type close to the component like other types? i was scrolling up and down to locate for it :D

packages/blade/src/components/Card/CardHeader.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  return (
+    <Box marginBottom="spacing.7">
+      <Box
+        marginBottom="spacing.7"
+        display="flex"
+        flexDirection="row"
+        justifyContent="space-between"
+      >
+        {children}
```
kamleshchandnani, line 106: how are we preventing other jsx to be passed by the consumer to be rendered here since this children accepts React.ReactNode

packages/blade/src/components/Card/CardHeader.tsx
```diff
@@ -0,0 +1,194 @@
...
+const CardHeaderIconButton = (props: CardHeaderIconButtonProps): React.ReactElement => {
+  useInsideCard('CardHeaderIconButton');
+
+  return (
+    <Box width={makeSpace(minHeight.xsmall)}>
+      <Button {...props} variant="tertiary" size="xsmall" iconPosition="left" isFullWidth />
+    </Box>
+  );
+};
+CardHeaderIconButton.componentId = Components.CardHeaderIconButton;
```
kamleshchandnani, line 84: how is TS not screaming here because this attribute is not defined on this component?

packages/blade/src/components/Card/CardHeader.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+const getComponentId = (comp: React.ReactNode): string | null => {
+  if (!React.isValidElement(comp)) return null;
+  return (comp.type as any)?.componentId;
+};
+
+const isOfComponentId = (comp: React.ReactNode, id: string): boolean => {
```
kamleshchandnani, line 91:
```suggestion
const isValidAllowedChildren = (component: React.ReactNode, id: string): boolean => {
```

kamleshchandnani: also move this to shared utils because @saurabhdaware is gonna need it for dropdown too?

packages/blade/src/components/Card/CardHeader.tsx
```diff
@@ -0,0 +1,194 @@
...
+  Components.CardHeaderIconButton,
+  Components.CardHeaderBadge,
+];
+
+const CardHeaderTrailing = ({ visual }: CardHeaderTrailingProps): React.ReactElement => {
+  useInsideCard('CardHeaderTrailing');
+
+  if (visual && !headerTrailingAllowedComponents.includes(getComponentId(visual)!)) {
+    throw new Error(
+      `[Blade CardHeaderTrailing]: Only \`${headerTrailingAllowedComponents.join(
```
kamleshchandnani, line 175:
```suggestion
      `[Blade CardHeaderTrailing]: Only one of \`${headerTrailingAllowedComponents.join(
```

packages/blade/src/components/Card/CardSurface.native.tsx
```diff
@@ -0,0 +1,42 @@
...
+import styled from 'styled-components/native';
+import { useTheme } from '~components/BladeProvider';
+import Box from '~components/Box';
+import type { BoxProps } from '~components/Box/types';
+
+const CardSurfaceStyled = styled(Box)<{ elevation: number; surfaceLevel: 2 | 3 }>(
+  ({ surfaceLevel, theme }) => {
+    const backgroundColor = theme.colors.surface.background[`level${surfaceLevel}`].lowContrast;
+    return {
+      shadowOpacity: '1',
```
kamleshchandnani, line 11: let's use this from token itself? its commented out there but we shall uncomment it there now

packages/blade/src/components/Card/CardSurface.native.tsx
```diff
@@ -0,0 +1,42 @@
...
+  children,
+  surfaceLevel,
+  ...props
+}: CardSurfaceProps): React.ReactElement => {
+  const { theme } = useTheme();
+  return (
+    <CardSurfaceStyled
+      {...props}
+      surfaceLevel={surfaceLevel}
+      elevation={theme.shadows.androidElevation.level[1]}
```
kamleshchandnani, line 35: define it on the top only under styled since it doesn't depend on any consumer prop?

packages/blade/src/components/Card/Divider.native.tsx
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components/native';
+import { makeSize } from '~utils';
+
+const Divider = styled.View(({ theme }) => {
```
kamleshchandnani, line 4: can we make divider a generic component and move it out?

kamleshchandnani, line 4: use box here? and then you might then need just one divider file?

packages/blade/ios/Podfile.lock
```diff
@@ -1,14 +1,14 @@
 PODS:
   - boost (1.76.0)
   - DoubleConversion (1.1.6)
-  - FBLazyVector (0.66.4)
-  - FBReactNativeSpec (0.66.4):
+  - FBLazyVector (0.66.5)
```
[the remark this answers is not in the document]
anuraghazra: Not card, I think there were some changes related to font in master branch but we didn't run pod install after that. So it created few metadata in here.

packages/blade/src/components/Box/types.ts
```diff
@@ -25,6 +26,7 @@ export type BoxProps = {
   marginBottom?: DotNotationSpacingStringToken | 'auto' | number;
   marginLeft?: DotNotationSpacingStringToken | 'auto' | number;
   marginRight?: DotNotationSpacingStringToken | 'auto' | number;
+  borderRadius?: keyof Border['radius'] | number;
```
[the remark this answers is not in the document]
anuraghazra: Right, I don't think I need to allow number in here. Will remove that.

packages/blade/src/components/Card/__tests__/Card.native.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+  it('should throw error if any sub card components are used outside of Card', () => {
+    expect(() => renderWithTheme(<CardHeader />)).toThrow(
+      '[Blade Card]: CardHeader cannot be used outside of Card component',
+    );
+    expect(() => renderWithTheme(<CardHeaderLeading title="" />)).toThrow(
+      '[Blade Card]: CardHeaderLeading cannot be used outside of Card component',
+    );
+    expect(() => renderWithTheme(<CardHeaderTrailing />)).toThrow(
+      '[Blade Card]: CardHeaderTrailing cannot be used outside of Card component',
+    );
```
kamleshchandnani, line 117: nit: ocd triggering - swap leading and trailing order? 🙈

packages/blade/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,260 @@
...
+                  title="Card Header"
+                  subtitle="Subtitle"
+                  prefix={<CardHeaderIcon icon={InfoIcon} />}
+                  suffix={<CardHeaderCounter value={12} />}
+                />
+                <CardHeaderTrailing visual={<CardHeaderBadge variant="positive">NEW</CardHeaderBadge>} />
+              </CardHeader>
+              <CardBody>
+                <Text>
+                  Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has
```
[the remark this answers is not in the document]
anuraghazra: Cool, added [this](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=21248%3A456316&t=c0OQHL1s3WzLEt7a-0) example

packages/blade/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,260 @@
...
+  title: 'Components/Card',
+  component: Card,
+  args: {
+    surfaceLevel: 2,
+    headerTitle: 'Payment Links',
+    headerSubtitle: 'Share payment link via an email, SMS, messenger, chatbot etc.',
+    footerTitle: 'Built for Developers',
+    footerSubtitle: 'By Developers.',
+    body:
+      "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.",
```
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,260 @@
...
+                  Lorem Ipsum passages, and more recently with desktop publishing software like Aldus
+                  PageMaker including versions of Lorem Ipsum.
+                </Text>
+              </CardBody>
+              <CardFooter>
+                <CardFooterLeading title="Card footer title" subtitle="Subtitle" />
+                <CardFooterTrailing
+                  actions={{
+                    primary: {
+                      onClick: () => console.log(1),
```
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,63 @@
...
+  CardHeader,
+  CardHeaderIcon,
+  CardHeaderCounter,
+  CardHeaderLeading,
+  CardHeaderTrailing,
+  CardHeaderBadge,
+  CardHeaderIconButton,
+  CardHeaderLink,
+  CardHeaderText,
+} from './CardHeader';
```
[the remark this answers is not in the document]
anuraghazra: No it won't cause any issues, bundlers are smart enough to deduce what components are actually being used and do the code elimination. 

Tho i'll move it to index.ts because there's no point on keeping it here anymore since we moved away from the compound components.

packages/blade/src/components/Card/__tests__/Card.native.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
+import { InfoIcon } from '~components/Icons';
+import { Text } from '~components/Typography';
+import { Counter } from '~components/Counter';
+import { Badge } from '~components/Badge';
+
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
+afterAll(() => jest.restoreAllMocks());
+
+describe('<Card />', () => {
```
kamleshchandnani, line 25: add test where someone tries to render any jsx inside Card

packages/blade/src/components/Card/__tests__/Card.native.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+          </CardHeader>
+        </Card>,
+      ),
+    ).toThrow(
+      '[Blade CardHeaderTrailing]: Only `CardHeaderLink, CardHeaderText, CardHeaderIconButton, CardHeaderBadge` component is accepted in visual',
+    );
+  });
+
+  it('should throw error if any sub card components are used outside of Card', () => {
+    expect(() => renderWithTheme(<CardHeader />)).toThrow(
```
kamleshchandnani, line 109: add test for CardBody being used outside Card?

packages/blade/src/components/Card/__tests__/Card.native.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+      <Card surfaceLevel={2}>
+        <CardBody>
+          <Text>Plain Card</Text>
+        </CardBody>
+      </Card>,
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render a Card with Header', () => {
```
kamleshchandnani, line 37: same test suite for footer?

packages/blade/src/components/Card/__tests__/Card.web.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+  it('should throw error if any sub card components are used outside of Card', () => {
+    expect(() => renderWithTheme(<CardHeader />)).toThrow(
+      '[Blade Card]: CardHeader cannot be used outside of Card component',
+    );
+    expect(() => renderWithTheme(<CardHeaderLeading title="" />)).toThrow(
+      '[Blade Card]: CardHeaderLeading cannot be used outside of Card component',
+    );
+    expect(() => renderWithTheme(<CardHeaderTrailing />)).toThrow(
+      '[Blade Card]: CardHeaderTrailing cannot be used outside of Card component',
+    );
```
kamleshchandnani, line 117: swap 🙈

packages/blade/src/components/Card/CardContext.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+type CardContextType = true | null;
```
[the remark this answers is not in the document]
anuraghazra: We do need `null | true`

Say we want to check if `CardHeaderCounter` is inside `Card` we will call the `useInsideCard` hook in `CardHeaderCounter`. 

Now if the provider value is not initialised or null we will throw the error. But if we pass `true` and we get back the context's value that means we are inside card. 


```tsx
const useInsideCard = (componentName: string): CardContextType => {
  const context = React.useContext(CardContext);
  // ^ If this hook is called inside Card then it will return `true` 
  // But if we put the context's value as null only then this will return null and we will throw error even tho it's okay
  if (!context) {
    throw new Error(`[Blade Card]: ${componentName} cannot be used outside of Card component`);
  }
  return context;
};
```

packages/blade/src/components/Card/CardFooter.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { Button } from '~components/Button';
+import { Text } from '~components/Typography';
+import { useBreakpoint } from '~utils';
+import { useTheme } from '~components/BladeProvider';
+
+export type CardFooterAction = Pick<
+  ButtonProps,
+  'type' | 'accessibilityLabel' | 'isLoading' | 'isDisabled' | 'icon' | 'iconPosition' | 'onClick'
+> & {
+  text: string;
```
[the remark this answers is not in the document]
anuraghazra: Sure.

packages/blade/src/components/Card/CardContext.tsx
```diff
@@ -0,0 +1,19 @@
...
+const CardContext = React.createContext<CardContextType>(null);
+
+const useInsideCard = (componentName: string): CardContextType => {
```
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/Card/CardFooter.tsx
```diff
@@ -0,0 +1,105 @@
...
+};
+
+type CardFooterProps = {
+  children?: React.ReactNode;
```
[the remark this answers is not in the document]
anuraghazra: We are not preventing anything on the top level components be it Footer, Card, CardHeader etc. We are only preventing components on the props like prefix, suffix, etc. 

IMO it's not worth the effort to be that restrictive. If they add something random going out of their way to add things on Header they might as well just use CardBody at that point.

packages/blade/src/components/Card/CardFooter.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+type CardFooterTrailingProps = {
+  actions?: {
+    primary?: CardFooterAction;
+    secondary?: CardFooterAction;
+  };
+};
+const CardFooterTrailing = ({ actions }: CardFooterTrailingProps): React.ReactElement => {
+  const isMobile = useIsMobile();
+  useInsideCard('CardFooterTrailing');
```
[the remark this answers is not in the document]
anuraghazra: To get proper error message on which component they used wrongly.

packages/blade/src/components/Card/CardFooter.tsx
```diff
@@ -0,0 +1,105 @@
...
+    <Box
+      display="flex"
+      flexDirection="row"
+      alignSelf={isMobile ? 'auto' : 'center'}
+      marginTop={isMobile ? 'spacing.5' : 'spacing.0'}
+      marginLeft={isMobile ? 'spacing.0' : 'spacing.5'}
+    >
+      <Box flexGrow={1}>
+        {actions?.secondary ? (
+          <Button isFullWidth size="medium" variant="secondary" {...actions.secondary}>
```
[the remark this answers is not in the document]
anuraghazra: Need it so that the buttons span over the entire footer. 

image

This change was done as part of the responsive issues which I was discussion with Rastogi.

packages/blade/src/components/Card/CardHeader.tsx
```diff
@@ -0,0 +1,194 @@
...
+  'variant' | 'size' | 'iconPosition' | 'isFullWidth' | 'children'
+> & {
+  icon: IconComponent;
+};
+
+const CardHeaderIconButton = (props: CardHeaderIconButtonProps): React.ReactElement => {
+  useInsideCard('CardHeaderIconButton');
+
+  return (
+    <Box width={makeSpace(minHeight.xsmall)}>
```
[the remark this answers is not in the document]
anuraghazra: To control the width of the button & make the size 1:1. 

Otherwise the width get wider: 

image

With box: 

image

packages/blade/src/components/Card/CardHeader.tsx
anuraghazra: Yeah sure, I added that closer but then added other components and pushed it up xD

packages/blade/src/components/Card/CardHeader.tsx
anuraghazra: We are not: https://github.com/razorpay/blade/pull/896#discussion_r1050723602

packages/blade/src/components/Card/CardHeader.tsx
anuraghazra: It infers it. Maybe I should make it explicit.

packages/blade/src/components/Card/CardSurface.native.tsx
anuraghazra: There is no shadowOpacity token I believe. It's just 1. 🤔

packages/blade/src/components/Card/CardSurface.native.tsx
anuraghazra: How do I pass that inline while defining the styled() tho? 🤔 was there any syntax around it? I forgot.

packages/blade/src/components/Card/Divider.native.tsx
anuraghazra: Let's wait for more usecase? Does action list have dividers? Maybe @saurabhdaware can extract this out in that case.

anuraghazra: Yeah I tried that earlier, it was making the Box component quite complex. Complex in terms of adding other border support like borderTop, borderBottom, borderLeft. so just added this here only.

packages/blade/src/components/Card/CardContext.tsx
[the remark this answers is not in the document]
kamleshchandnani: correct that's waht i'm saying if context exists then just return `true` because we don't really use `context` value anywhere right?

packages/blade/src/components/Card/CardFooter.tsx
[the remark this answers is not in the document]
kamleshchandnani: but this should maybe apply in case of just mobile and not desktop? because on desktop it might look very big and weird? the below looks ideal IMO?
Screenshot 2022-12-19 at 1 24 24 PM

packages/blade/src/components/Card/CardHeader.tsx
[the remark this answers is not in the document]
kamleshchandnani: which is fine? because that is what generally happens with our regular icon buttons?

packages/blade/src/components/Card/CardSurface.native.tsx
kamleshchandnani: yes but we have that value in our tokens. [check this](https://github.com/razorpay/blade/blob/master/packages/blade/src/tokens/theme/paymentTheme.ts#L2202)

packages/blade/src/components/Card/Divider.native.tsx
kamleshchandnani: yes action list does have divider. sure, we can scope it here only but lets raise this to design to build this as a component and we can extract this out then

kamleshchandnani: can you post a sample snippet of complexity it is creating?

packages/blade/src/components/Card/CardContext.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type CardContextType = true | null;
+const CardContext = React.createContext<CardContextType>(null);
+
+const useVerifyInsideCard = (componentName: string): CardContextType => {
+  const context = React.useContext(CardContext);
+  if (!context) {
+    throw new Error(`[Blade Card]: ${componentName} cannot be used outside of Card component`);
+  }
+  return context;
```
kamleshchandnani, line 11: over here we can just return `true`?

packages/blade/src/components/Card/CardFooter.tsx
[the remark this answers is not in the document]
kamleshchandnani: we do prevent in header right? using that isOfComponentId function? can't we replicate that in all the card sub-components?

packages/blade/src/components/Card/__tests__/Card.web.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+    expect(() => renderWithTheme(<CardFooter />)).toThrow(
+      '[Blade Card]: CardFooter cannot be used outside of Card component',
+    );
+    expect(() => renderWithTheme(<CardFooterTrailing />)).toThrow(
+      '[Blade Card]: CardFooterTrailing cannot be used outside of Card component',
+    );
+    expect(() => renderWithTheme(<CardFooterLeading title="" />)).toThrow(
+      '[Blade Card]: CardFooterLeading cannot be used outside of Card component',
+    );
```
kamleshchandnani, line 126: leading then trailing?

packages/blade/src/components/Card/CardSurface.web.tsx
anuraghazra: Added

packages/blade/src/components/Card/CardContext.tsx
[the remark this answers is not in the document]
anuraghazra: Oh you mean here? 

```
  return context;
```

Yeah right we can do that.

packages/blade/src/components/Card/CardFooter.tsx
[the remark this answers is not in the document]
anuraghazra: > we do prevent in header right

Yeah but only on the props. like prefix, suffix, visual etc.

> can't we replicate that in all the card sub-components?

We can, but that's what I'm saying is that it's not worth IMO. It will introduce runtime complexities too. For example in Footer we have to loop over the childrens and do the checks on every child to ensure they are the components we expect them to be.

packages/blade/src/components/Card/CardHeader.tsx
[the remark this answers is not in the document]
anuraghazra: But in design the card's icon button width:height is 1:1 which we need to follow I think, cc @saurav12 ^

packages/blade/src/components/Card/CardSurface.native.tsx
anuraghazra: Ahh i see. but It's commented should we expose it?

packages/blade/src/components/Card/Divider.native.tsx
anuraghazra: will do.

packages/blade/src/components/Card/CardFooter.tsx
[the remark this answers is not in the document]
anuraghazra: In desktop it will look as expected. 

image

If you notice the Button is wrapped in Box which has `flexGrow={1}` 
The reason I added isFullWidth to Button so that it grows and shrinks with the responsiveness of the Flex items. and on desktop I've added `isMobile ? 'column' : 'row'` thus it won't span the whole card as you've mentioned on desktop.

packages/blade/src/components/Card/CardHeader.tsx
anuraghazra: Done.

packages/blade/src/components/Card/CardHeader.tsx
anuraghazra: Will do.

anuraghazra: Done.

packages/blade/src/components/Card/__tests__/Card.native.test.tsx
anuraghazra: Done.

packages/blade/src/components/Card/__tests__/Card.native.test.tsx
anuraghazra: There is `should render a Card with Footer` let me move it up for easy scanning.

packages/blade/src/components/Card/__tests__/Card.web.test.tsx
anuraghazra: done.

packages/blade/src/components/Card/CardContext.tsx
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/Card/Divider.native.tsx
anuraghazra: The main factor is rn. 

rn doesn't support shorthand form of the border properties so have to use the `borderBottomWidth`, I can add it in just for bottom but later on if we try to add more there will be a lot of checks needed like this: 

```
  ...(borderBottomWidth && { borderBottomStyle: 'solid', borderBottomWidth: 1 }),
  ...(borderTopWidth && { borderTopStyle: 'solid', borderTopWidth: 1 }),
```

Don't want to implement a half backed API in the Box, we are already using it everywhere on the codebase.

packages/blade/src/components/Card/CardFooter.tsx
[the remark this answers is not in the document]
kamleshchandnani: what i'm afraid of is that we have put the restriction in the props but this is an easy escape hatch where people will just put things they want by passing anything to Footer. Yes while can do that in Body as well and not use footer at all but if they use footer we can't do anything and people 100% will end up doing that based on the recent instances 😢

packages/blade/src/components/Card/CardHeader.tsx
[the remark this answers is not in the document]
kamleshchandnani: then we shall add this to our icon button base component directly since it should look same across

packages/blade/ios/blade.xcodeproj/project.pbxproj
divyanshu013: Saw this on Chaitanya's PR too 🤔

divyanshu013 (commented):
- [ ] Add changeset
- [ ] Add screenshot on Android (elevation sometimes looks differently on ios and android)
- [ ] Storybook small mobile UI is breaking

anuraghazra:
@divyanshu013 added changeset, fixed the layout bug. 

Here's the shadow. 
It does look different on ios vs android even though I'm using the same tokens and `elevation` defined in tokens. 

image

divyanshu013:
Exactly, will need to debug, check for box shadow issues on react native. On android I think react native supports some different elevation props while ios supports shadow based props.

packages/blade/ios/blade.xcodeproj/project.pbxproj
anuraghazra: Yeah right, he will need to remove those changes in his PR.

divyanshu013 (approved):
Looks good to me.

anuraghazra:
Forgot to reexport from index, :p need an eslint rule for this too

anuraghazra:
blade is failing to build, investigating the issue.

anuraghazra:
Fixed the problem, similar to this https://github.com/razorpay/blade/issues/701

packages/blade/src/components/Box/types.ts
```diff
@@ -25,6 +26,7 @@ export type BoxProps = {
   marginBottom?: DotNotationSpacingStringToken | 'auto' | number;
   marginLeft?: DotNotationSpacingStringToken | 'auto' | number;
   marginRight?: DotNotationSpacingStringToken | 'auto' | number;
+  borderRadius?: keyof Border['radius'];
```
chaitanyadeorukhkar, line 29: Shouldn't this be
```suggestion
  borderRadius?: DotNotationBorderRadiusStringToken
```

anuraghazra: No since border radius is a plain object, no dotnotation needed. 

{
    none: 0;
    small: 2;
    medium: 4;
    large: 8;
    max: 9999;
    round: "50%";
}

anuraghazra: And the API will be like: 

<Box borderRaidus="small" /> instead of <Box borderRadius="radius.small" />

packages/blade/src/components/Card/CardSurface.native.tsx
```diff
@@ -0,0 +1,54 @@
...
+import styled from 'styled-components/native';
+import { useTheme } from '~components/BladeProvider';
+import Box from '~components/Box';
+import type { BoxProps } from '~components/Box/types';
+
+const isAndroid = Platform.OS === 'android';
+// TODO: Temporary workaround to make android shadows look as close as iOS
+const androidShadow = {
+  color: undefined,
+  elevation: 2,
```
chaitanyadeorukhkar, line 12: Lets use our elevation tokens here?

https://github.com/razorpay/blade/blob/master/packages/blade/src/tokens/theme/paymentTheme.ts#L2192

anuraghazra: Need to do this for android to make it look close to iOS shadow. Can't use token here since in the value is 5 there. :(

https://razorpay.slack.com/archives/G01B3LQ9H0W/p1671601580827619?thread_ts=1671538782.461409&cid=G01B3LQ9H0W

packages/blade/src/components/Box/getBoxStyles.ts
```diff
@@ -91,6 +92,7 @@ const getBoxStyles = ({
   transform,
   background,
   zIndex,
+  borderRadius: getIn(theme, `border.radius.${borderRadius}`),
```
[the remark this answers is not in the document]
anuraghazra: Will be the same like "getIn(theme.border, `radius.${borderRadius}`)"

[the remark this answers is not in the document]
anuraghazra: will add the makeSize tho

packages/blade/src/components/Card/CardSurface.native.tsx
chaitanyadeorukhkar: We need to relook at our elevation tokens then. Lets hardcode it now but can you raise an issue to track this?

anuraghazra: Yes raised issue already #922

chaitanyadeorukhkar (approved):

PR #897, opened 2022-12-01
feat: allow masked otp input

chaitanyadeorukhkar:
Adds support for masked input fields in `OTPInput`

**Demo:** 

https://user-images.githubusercontent.com/24487274/204982587-e1400d43-7f01-4bc5-8b85-6072cdb473a9.mov

### DIFF
--- a/packages/blade/src/components/Input/OTPInput/OTPInput.tsx
+++ b/packages/blade/src/components/Input/OTPInput/OTPInput.tsx
@@ -248,7 +248,7 @@ const OTPInput = ({
             errorText={errorText}
             helpText={helpText}
             hideFormHint={true}
-            type={isMasked ? 'password' : 'numeric'}
+            type={isMasked ? 'password' : undefined}
           />
         </Box>,
       );
--- a/packages/blade/src/components/Input/OTPInput/OTPInput.tsx
+++ b/packages/blade/src/components/Input/OTPInput/OTPInput.tsx
@@ -36,6 +36,10 @@ export type OTPInputProps = Pick<
    * The callback function to be invoked when all the values of the OTPInput are filled
    */
   onOTPFilled?: FormInputOnEvent;
+  /**
+   * Masks input characters in all the fields
+   */
+  isMasked?: boolean;
 };
 
 const isReactNative = getPlatformType() === 'react-native';
@@ -76,6 +80,7 @@ const OTPInput = ({
   successText,
   validationState,
   value: inputValue,
+  isMasked,
 }: OTPInputProps): React.ReactElement => {
   const inputRefs: React.RefObject<HTMLInputElement>[] = [];
   const [otpValue, setOtpValue] = useState<string[]>(otpToArray(inputValue));
@@ -243,6 +248,7 @@ const OTPInput = ({
             errorText={errorText}
             helpText={helpText}
             hideFormHint={true}
+            type={isMasked ? 'password' : 'numeric'}
           />
         </Box>,
       );
--- a/packages/blade/src/components/Input/OTPInput/__tests__/__snapshots__/OTPInput.web.test.tsx.snap
+++ b/packages/blade/src/components/Input/OTPInput/__tests__/__snapshots__/OTPInput.web.test.tsx.snap
@@ -332,7 +332,7 @@ exports[`<OTPInput /> should render 1`] = `
                   inputmode="decimal"
                   placeholder=""
                   required=""
-                  type="text"
+                  type="numeric"
                   value=""
                 />
               </div>
@@ -369,7 +369,7 @@ exports[`<OTPInput /> should render 1`] = `
                   inputmode="decimal"
                   placeholder=""
                   required=""
-                  type="text"
+                  type="numeric"
                   value=""
                 />
               </div>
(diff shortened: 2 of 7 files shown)


packages/blade/src/components/Input/OTPInput/__tests__/__snapshots__/OTPInput.web.test.tsx.snap
```diff
@@ -332,7 +332,7 @@ exports[`<OTPInput /> should render 1`] = `
                   inputmode="decimal"
                   placeholder=""
                   required=""
-                  type="text"
+                  type="numeric"
```
anuraghazra, line 335: Numeric?

chaitanyadeorukhkar: Keyboard type is decimal, wanted to make the type similar too. Should've done thing before only

anuraghazra (approved):
Mostly LGTM

packages/blade/src/components/Input/OTPInput/__tests__/__snapshots__/OTPInput.web.test.tsx.snap
anuraghazra: Related to this right? https://github.com/razorpay/blade/issues/786 
Hopefully no one is depending on OTP being alphanumeric.

chaitanyadeorukhkar: Probably not. We can revisit this if we see someone facing any issues. Don't think we need to work on this right now

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
```diff
@@ -243,6 +248,7 @@ const OTPInput = ({
             errorText={errorText}
             helpText={helpText}
             hideFormHint={true}
+            type={isMasked ? 'password' : 'numeric'}
```
saurabhdaware, line 251: Shouldn't this be `type="text"` and `inputmode="numeric"`? couldn't find any mdn documentation of `numeric` type as such https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#using_number_inputs

packages/blade/src/components/Input/OTPInput/__tests__/__snapshots__/OTPInput.web.test.tsx.snap
saurabhdaware: So `numeric` won't really stop that. We might have to add `pattern="\d*"` for this.
image

chaitanyadeorukhkar: Yeah, it won't stop it. Just made it consistent with the keyboardType. Whether we want to stop it or not can be tackled separately. Lets scope this PR to just addition of masked input. I added `numeric` to make it consistent with they keyboardType but I can revert it if its an issue 😅

saurabhdaware: Oh yes yes makes sense. Only this concern then https://github.com/razorpay/blade/pull/897#discussion_r1036766905. Rest looks good to me.

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
kamleshchandnani: @chaitanyadeorukhkar - https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Input/BaseInput/_decisions/_decisions.md#why-we-are-not-having-keyboardtypenumeric

packages/blade/src/components/Input/OTPInput/__tests__/__snapshots__/OTPInput.web.test.tsx.snap
anuraghazra: AHH makes sense. So numeric will change the keyboard behaviour not the input's validation.

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
chaitanyadeorukhkar: Yeah, seems like type `numeric` doesn't exist. I used `BaseInput`'s props only. We have specific `numeric` as an acceptable value for type. Let me remove this change. For now I'll just keep it as is and raise an issue for it.

chaitanyadeorukhkar: @kamleshchandnani the link you specified is for `keyboardType` attribute not for `type`. `type=numeric` doesn't exist but we accept that value. We should ideally make it `type=number`?

chaitanyadeorukhkar: Went through [this](https://css-tricks.com/better-form-inputs-for-better-mobile-user-experiences/)

So we don't want to pass `type=number` instead we want to pass `inputmode=decimal`. inputmode is `keyboardType` for us which we're already passing. `type=numeric` seems to be an incorrect value we're accepting and rather it should be `type=number`. I'll make that change in a different PR

Screenshot 2022-12-01 at 4 34 34 PM

chaitanyadeorukhkar: Raised a PR for that change

PR #898, opened 2022-12-01
fix: change BaseInput `type` from  `numeric` to `number`

chaitanyadeorukhkar:
Ref: https://github.com/razorpay/blade/pull/897#discussion_r1036766905

**EDIT:** Looks like there is already some logic on `TextInput` that handles figuring out `type` attribute. [Ref](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Input/TextInput/TextInput.tsx#L76)

Changing it from `numeric` to `number` breaks that logic. This means we will have to :
1. Change that logic to move the checks from `numeric` to `number`
2. Add this logic to BaseInput instead of TextInput so that anyone using `BaseInput` can leverage this logic automatically


.changeset/sour-chefs-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
```
chaitanyadeorukhkar, line 2: Making it a patch change since we have no consumers using this prop yet

chaitanyadeorukhkar:
@kamleshchandnani need your input here. I see there is some logic to handle `numeric` on TextInput. Why is it limited to TextInput? Shall we move it to `BaseInput`? Can we change the acceptable type as `number` instead of `numeric`?

kamleshchandnani:
> @kamleshchandnani need your input here. I see there is some logic to handle `numeric` on TextInput. Why is it limited to TextInput? Shall we move it to `BaseInput`? Can we change the acceptable type as `number` instead of `numeric`?

we can't move that logic from TextInput to BaseInput since BaseInput is a like a config thing every other component might not need that for example TextArea, password input, OTP input hence keeping it contextual to where it belongs is the right thing IMO. later what we can do is if this being used at a lot of places we can extract this and move it up under `Input` as common util

saurabhdaware (approved):
Looks good! should we also add `pattern` for `type=number`?

It will block form submission http://jsfiddle.net/bHWcu/1/ 🤔

kamleshchandnani:
> Looks good! should we also add `pattern` for `type=number`?
> 
> It will block form submission [http://jsfiddle.net/bHWcu/1](http://jsfiddle.net/bHWcu/1/) 🤔

we anyways don't rely on default form submission and do it via javascript so maybe we can live with this for now

PR #899, opened 2022-12-01
fix(theme): add bladeOld as theme key for blade-old themeing

divyanshu013:
This is to prevent theme collision when blade is used within another styled-components theme provider.

- We'll release a hotfix from the target branch once this is merged.
- Validated web storybook, seems to be working fine
- No diff in snapshots so theme should be getting passed and accessed correctly

- [ ] Needs sanity test on mobile

## Migration steps for consumers

This is a hotfix, meant only for subset of consumers.

Steps for migration:
1. Wherever you're using a token from `theme` in blade-old, replace that to be now accessed from a theme key `theme.bladeOld`.

Example:

```diff
- theme.colors.something
+ theme.bladeOld.colors.something
```


packages/blade-old/src/tokens/theme.web.js
```diff
@@ -1,206 +1,216 @@
 import { colors, fonts, spacings } from '../tokens';
 
+/**
+ * Styled components merges passed themes when multiple theme providers are nested.
```
kamleshchandnani, line 4: let's elaborate this a bit that it merges when then top level keys are same. update this in the PR description too

kamleshchandnani (changes requested):
Add the version number to the `package.json` too?

kamleshchandnani:
shouldn't this go the `blade-old` codebase? 🤔  we don't have a git history there is it?

divyanshu013:
Closing in favor of https://github.com/razorpay/blade-old/pull/9

PR #903, opened 2022-12-05
feat: add `ProgressBar` component

chaitanyadeorukhkar:
Adds the ProgressBar component.
- #833 

### This PR contains the following:
- Determinate Progress Bar
- Meter variant (static)
- Label & Percentage
- Contrast
- Intent
- Motion (fill + pulse animations)
- Accessibility (for `progressbar` as well as `meter`)

### Pending items (in a follow-up PR):
- Indeterminate Progress Bar
- Stories
- Tests

### Demo

**Web:** 

https://user-images.githubusercontent.com/24487274/206695341-4dcc1114-896d-4ecd-a25b-94e8119600cc.mov

**iOS:**

https://user-images.githubusercontent.com/24487274/206695381-c4b82f8b-0c03-4734-bcbe-f686f43c48c4.mov

**Android:**

https://user-images.githubusercontent.com/24487274/206695398-00d8bf17-f756-47be-b4b5-1617938c9396.mov


packages/blade/ios/blade.xcodeproj/project.pbxproj
```diff
@@ -15,8 +15,8 @@
 		7EF68E3733C33B6898317E18 /* libPods-blade-bladeTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = ABFE59519B596E51CEFDCCC0 /* libPods-blade-bladeTests.a */; };
 		81AB9BB82411601600AC10FF /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 81AB9BB72411601600AC10FF /* LaunchScreen.storyboard */; };
 		BE94BDCFE4A742E89E116690 /* Lato-Regular.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 4C0A80DCDFF54BF28ED33285 /* Lato-Regular.ttf */; };
-		F31095E142B84AD0A9E345EB /* Lato-Bold.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 856B36C0CBEA4725AD05BF92 /* Lato-Bold.ttf */; };
 		DD50086558394F9D8B0F07DA /* Lato-Italic.ttf in Resources */ = {isa = PBXBuildFile; fileRef = A2D9B3E1EEE34EC9AD42AA9C /* Lato-Italic.ttf */; };
+		F31095E142B84AD0A9E345EB /* Lato-Bold.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 856B36C0CBEA4725AD05BF92 /* Lato-Bold.ttf */; };
```
chaitanyadeorukhkar, line 19: Xcode did this

packages/blade/ios/blade.xcodeproj/project.pbxproj
```diff
@@ -45,11 +45,11 @@
 		6C97AB639B58BBB4B15BBE30 /* Pods-blade-bladeTests.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-blade-bladeTests.debug.xcconfig"; path = "Target Support Files/Pods-blade-bladeTests/Pods-blade-bladeTests.debug.xcconfig"; sourceTree = "<group>"; };
 		81AB9BB72411601600AC10FF /* LaunchScreen.storyboard */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = file.storyboard; name = LaunchScreen.storyboard; path = blade/LaunchScreen.storyboard; sourceTree = "<group>"; };
 		856B36C0CBEA4725AD05BF92 /* Lato-Bold.ttf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = 9; includeInIndex = 0; lastKnownFileType = unknown; name = "Lato-Bold.ttf"; path = "../public/fonts/Lato-Bold.ttf"; sourceTree = "<group>"; };
+		A2D9B3E1EEE34EC9AD42AA9C /* Lato-Italic.ttf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = undefined; includeInIndex = 0; lastKnownFileType = unknown; name = "Lato-Italic.ttf"; path = "../public/fonts/Lato-Italic.ttf"; sourceTree = "<group>"; };
```
chaitanyadeorukhkar, line 48: Xcode did this

packages/blade/src/components/Box/types.ts
```diff
@@ -36,4 +36,5 @@ export type BoxProps = {
   transform?: CSSObject['transform'];
   background?: CSSObject['background'];
   zIndex?: CSSObject['zIndex'];
+  id?: string;
```
chaitanyadeorukhkar, line 39: Need to add this explicitly. Box as div on web has this by default but native does not and leads to failure in TS checks on native

packages/blade/src/components/Checkbox/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -127,6 +128,8 @@ const CheckboxGroup = ({
           position={labelPosition}
           id={ids.labelId}
           accessibilityText={accessibilityText}
+          spacingRight={labelPosition === 'left' ? 'spacing.5' : 'spacing.0'}
+          spacingBottom={labelPosition === 'top' || isReactNative ? 'spacing.2' : 'spacing.0'}
```
chaitanyadeorukhkar, line 132: Explicitly adding margins for labels on individual components instead of relying on the Form label to add these margins. Better to keep the re-usable component free of margin styling so that individual components can place them with appropriate styles

packages/blade/src/components/Checkbox/__tests__/__snapshots__/CheckboxGroup.native.test.tsx.snap
```diff
@@ -16,12 +16,12 @@ exports[`<CheckboxGroup /> should render with label 1`] = `
...
   <View
     marginBottom="spacing.2"
-    marginRight="spacing.5"
+    marginRight="spacing.0"
     style={
       Array [
         Object {
           "marginBottom": 4,
-          "marginRight": 16,
+          "marginRight": 0,
```
chaitanyadeorukhkar, line 24: Label is always on top on native but for some reason native had a marginRight of 16px. Removed that now.

packages/blade/src/components/Checkbox/__tests__/__snapshots__/CheckboxGroup.web.test.tsx.snap
```diff
@@ -209,7 +209,7 @@ exports[`<CheckboxGroup /> should render with help text 1`] = `
   >
     <span
       id="checkbox-group-49-label-54"
-      style="width: auto; flex-shrink: 0; margin-right: 16px; word-break: break-all;"
+      style="width: auto; flex-shrink: 0; word-break: break-all;"
```
chaitanyadeorukhkar, line 212: When the label is on top on web, we used to add a margin-bottom as well as margin-right. We do not need a right margin when the label is on top. Removed the unnecessary margin which led to a snapshot change.

packages/blade/src/components/Form/FormLabel.tsx
```diff
@@ -50,6 +61,9 @@ const FormLabel = ({
   children,
   id,
   htmlFor,
+  contrast = 'low',
```
chaitanyadeorukhkar, line 64: We need the label to also follow contrast of the component where it's being used

packages/blade/ios/blade.xcodeproj/project.pbxproj
anuraghazra: Yeah on Card PR too. We might have a conflict  then.

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+  variant?: Extract<ProgressBarVariant, 'progress'>;
+  isIndeterminate?: boolean;
+};
+
+type ProgressBarMeterProps = ProgressBarCommonProps & {
+  variant?: Extract<ProgressBarVariant, 'meter'>;
+  isIndeterminate?: undefined;
+};
+
+type ProgressBarProps<T> = T extends {
```
anuraghazra, line 34: Do we need conditional type here? 

```ts
type ProgressBarProps = ProgressBarProgressProps | ProgressBarMeterProps;

```

should work too right?

chaitanyadeorukhkar: It should. I did this initially but then changed it. Can't recall why. Anyway made the change now

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+            >{`${progressValue}%`}</Text>
+          </Box>
+        ) : null}
+      </Box>
+      <Box
+        id={id}
+        {...makeAccessible({
+          role: variant === 'meter' ? 'meter' : 'progressbar',
+          label: accessibilityLabel ?? label,
+          busy: !isMeter,
```
anuraghazra, line 108: aria-busy should not be applied to the progressbar, aria-busy should be applied to the region which is being loaded. 

For example if a infinite scroll feed is loading aria-busy will be added to the whole infinite scrolling div.

packages/blade/src/components/ProgressBar/ProgressBarFilled.web.tsx
```diff
@@ -0,0 +1,81 @@
...
+    'pulseMotionDuration' | 'pulseMotionDelay' | 'motionEasing' | 'variant'
+  >
+>(({ theme, pulseMotionDuration, pulseMotionDelay, motionEasing, variant }) =>
+  variant === 'progress'
+    ? css`
+        height: 100%;
+        width: 100%;
+        opacity: 0;
+        background-color: white;
+        animation: ${pulse} ${castWebType(makeMotionTime(getIn(theme.motion, pulseMotionDuration)))}
```
anuraghazra, line 47: Can we extract the variables out? Finding it super hard to reason about which interpolation is which. 

```ts
  const duration = castWebType(makeMotionTime(getIn(theme.motion, pulseMotionDuration)));
  const easing = castWebType(getIn(theme.motion, motionEasing));
  const delay = castWebType(makeMotionTime(getIn(theme.motion, pulseMotionDelay)));

  return variant === 'progress'
    ? css`
        height: 100%;
        width: 100%;
        opacity: 0;
        background-color: white;
        animation: ${pulse} ${duration} ${easing} infinite;
        animation-delay: ${delay};
      `
    : '';
```

packages/blade/src/components/ProgressBar/ProgressBar.tsx
anuraghazra: Looks good now, I think we did conditional type on the Text because we wanted to intelisense to properly show the fontWeight=regular|bold but here we don't need this.

packages/blade/src/components/Box/types.ts
anuraghazra: Better to do this explicitly anyways

packages/blade/src/components/ProgressBar/ProgressBar.tsx
chaitanyadeorukhkar: Fair. The mdn docs [say the same](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role#:~:text=set%20the%20aria%2Dbusy%20attribute%20to%20true%20on%20the%20region%20until%20it%20is%20finished%20loading.).  Although I looked at a few other design systems around and some of them had aria-busy set to true. Better to not add it here and let the consumers add it in the region then.

anuraghazra (approved):
Looks good!

packages/blade/ios/Podfile.lock
```diff
@@ -1,14 +1,14 @@
 PODS:
   - boost (1.76.0)
   - DoubleConversion (1.1.6)
-  - FBLazyVector (0.66.4)
-  - FBReactNativeSpec (0.66.4):
+  - FBLazyVector (0.66.5)
```
divyanshu013, line 4: Pods updating intentional?

packages/blade/src/components/ProgressBar/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </Sandbox>
+    </StoryPageWrapper>
+  );
+};
+
+export default {
+  title: 'Components/ProgressBar',
+  component: ProgressBarComponent,
+  args: {
+    children: 'Learn More',
```
divyanshu013, line 44: Seems like mistype?
```suggestion
```

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * @default 'small'
+   */
+  value?: number;
+};
+
+type ProgressBarVariant = 'progress' | 'meter';
+
+type ProgressBarProgressProps = ProgressBarCommonProps & {
+  /**
+   * Sets the variant to be rendered for thr progress bar.
```
divyanshu013, line 46:
```suggestion
   * Sets the variant to be rendered for the progress bar.
```

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * @default 'progress'
+   */
+  variant?: Extract<ProgressBarVariant, 'progress'>;
```
divyanshu013, line 49: You could just use `progress` here instead of extracting from the variant? 🤔

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,185 @@
...
+  hidePercentage?: undefined;
+};
+
+type ProgressBarProps = ProgressBarProgressProps | ProgressBarMeterProps;
+
+type ProgressBarUnfilledProps = {
+  backgroundColor: string;
+};
+
+const ProgressBarUnfilled = styled(Box)<ProgressBarUnfilledProps>(({ backgroundColor }) => ({
```
divyanshu013, line 86: Optional: we could add / use `backgroundColor` prop on `Box` for this?

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,185 @@
...
+  backgroundColor,
+}));
+
+const getProgress = (value: number): number => {
+  return Math.floor(Math.min(100, Math.max(0, value)));
```
divyanshu013, line 91: Optional: logic is bit complex to understand here, might be more readable to split this and replace with variables or add some comments

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,185 @@
...
+  contrast = 'low',
+  intent,
+  isIndeterminate,
+  label,
+  hidePercentage = false,
+  size = 'small',
+  value = 0,
+  variant = 'progress',
+}: ProgressBarProps): ReactElement => {
+  console.log(isIndeterminate); // will be in follow up PR
```
divyanshu013: Todo: cleanup log

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,185 @@
...
+    console.warn(`[Blade: ProgressBar]: Cannot set 'isIndeterminate' when 'variant' is 'meter'.`);
+  }
+
+  if (variant === 'meter' && hidePercentage) {
+    console.warn(
+      `[Blade: ProgressBar]: Cannot set 'hidePercentage' when 'variant' is 'meter'. Percentage is always hidden for meter.`,
+    );
+  }
+
+  const { theme } = useTheme();
```
divyanshu013, line 120: Suggestion: Ideally we should be invoking all hooks before any conditionals. If someone adds a conditional throw in either of the conditions above it could lead to weird rendering issues because hooks are meant to be invoked unconditionally. Not sure if our linting rules catch that right now.

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,185 @@
...
+          </FormLabel>
+        ) : null}
+        {hidePercentage || isMeter ? null : (
+          <Box marginBottom="spacing.2">
+            <Text
+              type="subdued"
+              variant="body"
+              contrast={contrast}
+              size="small"
+            >{`${progressValue}%`}</Text>
```
divyanshu013, line 149: Just confirming, do we want to render the UI even if it's `0%`?

packages/blade/src/utils/makeAccessible/accessibilityMapNative.ts
```diff
@@ -97,4 +97,5 @@ export const accessibilityRoleMap = {
   keyboardkey: 'keyboardkey',
   label: 'label',
   text: 'text',
+  meter: 'progressbar',
```
divyanshu013, line 100: Should we map the `meter` role to progressbar on native? Not sure if every static meter should be treated as a progressbar 🤔

packages/blade/src/components/ProgressBar/ProgressBar.tsx
anuraghazra: Or also be better if we rename it to something like `clamp` or `clampValue` since it's clamping the value from 0 to 100

packages/blade/src/components/ProgressBar/ProgressBar.stories.tsx
chaitanyadeorukhkar: Removed in #911

packages/blade/src/components/ProgressBar/ProgressBar.tsx
chaitanyadeorukhkar: Done

packages/blade/src/components/ProgressBar/ProgressBar.tsx
chaitanyadeorukhkar: Yup!

packages/blade/src/utils/makeAccessible/accessibilityMapNative.ts
chaitanyadeorukhkar: There's no other similar role available on native. I tried looking up what we could do but no luck. Progress bar is the closest to the use-case here

packages/blade/src/components/ProgressBar/ProgressBar.tsx
chaitanyadeorukhkar: Lodash has clamp and its documented well. Used that instead. Removed the `getProgress` function

packages/blade/src/utils/makeAccessible/accessibilityMapNative.ts
divyanshu013: Not much expertise on native a11y, perhaps we don't add a role for this? Just works as a regular UI.

divyanshu013 (approved):
Looks good to me.

PR #909, opened 2022-12-09
feat(Alert): update `isFullWidth` to make inline borderless alerts on desktop

divyanshu013:
## Details

Now `isFullWidth` prop is rewired to cause these effects:
- Makes the alert borderless (same UI as earlier with `isBorderless` on mobile and native)
- Makes the actions inline (appear horizontal) with the content besides the above change (only on desktop). Note that actions always appear vertically on mobile and react native.

## Screenshots

Tested various combinations, some screenshots:

Screen Shot 2022-12-09 at 18 06 37

Screen Shot 2022-12-09 at 18 07 32

Screen Shot 2022-12-09 at 18 08 10

Screen Shot 2022-12-09 at 18 09 56

Screen Shot 2022-12-09 at 18 10 50

Screen Shot 2022-12-09 at 18 11 10

And... oh well, here's a meme 😅 

image

### DIFF
--- a/packages/blade/src/components/Alert/Alert.tsx
+++ b/packages/blade/src/components/Alert/Alert.tsx
@@ -81,7 +81,8 @@ type AlertProps = {
   contrast?: ColorContrastTypes;
 
   /**
-   * Makes the Alert span the entire container width, instead of the default max width of `584px`
+   * Makes the Alert span the entire container width, instead of the default max width of `584px`.
+   * This also makes the alert borderless, useful for creating full bleed layouts.
    *
    * @default false
    */
@@ -93,13 +94,6 @@ type AlertProps = {
    * @default neutral
    */
   intent?: Feedback;
-
-  /**
-   * Removes border and border radii, useful for creating full bleed layouts. Automatically sets `isFullWidth` to `true` when enabled.
-   *
-   * @default false
-   */
-  isBorderless?: boolean;
 
   /**
    * Renders a primary action button and a secondary action link button
@@ -139,7 +133,6 @@ const Alert = ({
   contrast = 'low',
   isFullWidth = false,
   intent = 'neutral',
-  isBorderless = false,
   actions,
 }: AlertProps): ReactElement | null => {
   if (!actions?.primary && actions?.secondary) {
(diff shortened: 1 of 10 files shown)


packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -235,15 +230,26 @@ const Alert = ({
     secondaryActionParams.rel = actions.secondary.rel;
   }
   const secondaryAction = actions?.secondary ? (
-    <SecondaryActionWrapper>
+    <Box marginRight="spacing.4" display={isReactNative ? 'flex' : 'inline-flex'}>
```
anuraghazra, line 233: Where does this spacing.4 come from?

divyanshu013: Need it for spacing on right of the link button

image

packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -235,15 +230,26 @@ const Alert = ({
...
+    </Box>
   ) : null;
 
-  const _actions =
-    primaryAction || secondaryAction ? (
+  // For certain cases we wish to render actions inline with text content
+  const showActionsHorizontal = isFullWidth && isDesktop;
+
+  const actionsHorizontal =
+    showActionsHorizontal && (primaryAction || secondaryAction) ? (
```
kamleshchandnani, line 244:
```suggestion
    showActionsHorizontal && (primaryAction || secondaryAction || isDismissable) ? (
```

shall check for dismissable as well?

kamleshchandnani (commented):
just one tiny comment rest LGTM but little upset that too many conditions. almost feels like business logic 😭

packages/blade/src/components/Alert/Alert.tsx
divyanshu013: Dismissible check isn't added because we don't want to render the container of actions at all if both primary and secondary actions aren't present. Say we add this in the OR conditional and none of primary and secondary action are added, it would still render an empty `Box` (container for actions)

divyanshu013:
> almost feels like business logic 😭
Yup, we've too many variations now 😿

kamleshchandnani (approved):
LGTM 🚀

chaitanyadeorukhkar (approved):
:shipit:

PR #910, opened 2022-12-12
fix(Alert)!: fix typo in prop `isDismissible`

### LINKED ISSUE #901: Typo in Alert's `isDismissable` prop
We need to change it from `isDismissable` to `isDismissible`. Since this is a breaking change for a few consumers, we will make this change along with some other breaking changes we make with blade. We are keeping this issue open till then. 

https://github.com/razorpay/blade/blob/756f4b46c376a897e2c6324b5ad68d932ab4aef1/packages/blade/src/components/Alert/Alert.tsx#L69

divyanshu013:
Update `isDismissable` to `isDissmissible`. Fixes #901 

> **Warning**
>
> Breaking change. To be released in the next major release.


.changeset/mighty-glasses-rhyme.md
```diff
@@ -0,0 +1,20 @@
+---
+"@razorpay/blade": major
+---
+
+fix(Alert)!: fix typo in prop `isDismissible`
+
+> **Warning**
+>
+> Breaking change. Update prop `isDismissable` to `isDismissible`.
```
kamleshchandnani, line 9: are we sure to put it in the major release? what does the amplitude data says on the usage of this prop?

kamleshchandnani: let's club it with the other alert changes?

divyanshu013: There are [few occurrences](https://cs.github.com/?scope=org%3Arazorpay&scopeName=razorpay&q=isdismissable) across projects (total 9 occurrences). I think it might be better to push this in the next major release because we don't need this urgently.

kamleshchandnani: I checked the files there are like 8 occurrences. and most of them are explicitly setting this prop to false which by default is false only right? still unsure if this requires a major breaking release from political standpoint 🙊

divyanshu013: Default value is `true` 😅

anuraghazra (approved):
LGTM.

chaitanyadeorukhkar (approved):

snitin315 (approved):
:shipit:

kamleshchandnani (commented):
:shipit:

Issue #835, opened 2022-11-11
Typecheck in validate workflow isn't working properly for test files


divyanshu013:
## Steps to replicate

1. Go to a test file (example `Indicator.web.test.tsx`) and make a type error on a prop (I chose passing `size` as `xsmall` which is incorrect).
2. Run `yarn tsc` - you should be able to see type error
3. Run `yarn typecheck` (which runs on CI) - you won't be able to see the type error (therefore you can ship code which is failing typecheck to prod)

## Screenshots

**With `yarn tsc`**

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/6682655/201288949-bef8943d-aabf-42c3-95bb-bd8cfe6ed3c6.png">

**With `yarn typecheck`**

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/6682655/201288985-20d239a1-65f7-4a5f-8540-ba03f9cb4e29.png">

PR #916, opened 2022-12-16
chore: typecheck test files on CI

anuraghazra:
fixes https://github.com/razorpay/blade/issues/835

Removed exclusion of test files from tsconfig config.

### DIFF
--- a/packages/blade/src/components/Typography/Text/__tests__/Text.web.test.tsx
+++ b/packages/blade/src/components/Typography/Text/__tests__/Text.web.test.tsx
@@ -74,7 +74,6 @@ describe('<Text />', () => {
     try {
       const displayText = 'Displaying some text';
       renderWithTheme(
-        // @ts-expect-error testing failure case when weight='bold' is passed with variant='caption'
         <Text type="normal" variant="caption" weight="bold">
           {displayText}
         </Text>,
--- a/packages/blade/src/components/Typography/Text/__tests__/Text.web.test.tsx
+++ b/packages/blade/src/components/Typography/Text/__tests__/Text.web.test.tsx
@@ -53,37 +53,14 @@ describe('<Text />', () => {
   });
 
   it('should throw error when variant is "caption" and size "small" is passed', () => {
-    try {
-      const displayText = 'Displaying some text';
+    const displayText = 'Displaying some text';
+    expect(() =>
       renderWithTheme(
         // @ts-expect-error testing failure case when size='small' is passed with variant='caption'
         <Text type="normal" variant="caption" truncateAfterLines={3} size="small">
           {displayText}
         </Text>,
-      );
-    } catch (error: unknown) {
-      if (error instanceof Error) {
-        expect(error.message).toEqual(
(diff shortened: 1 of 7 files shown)


packages/blade/tsconfig-typecheck.web.json
```diff
@@ -1,5 +1,11 @@
 {
   "extends": "./tsconfig-generate-types.web.json",
+  "exclude": [
```
divyanshu013, line 3: How does this work? Won't it exclude these sub extensions from typechecking 🤔

anuraghazra: Yes it will exclude all the files with these extensions `.stories.tsx`, `.native.ts` etc

packages/blade/src/components/Typography/Text/__tests__/Text.web.test.tsx
```diff
@@ -53,38 +53,14 @@ describe('<Text />', () => {
...
-  it('should throw error when variant is "caption" but weight "bold" is passed', () => {
-    try {
-      const displayText = 'Displaying some text';
-      renderWithTheme(
-        // @ts-expect-error testing failure case when weight='bold' is passed with variant='caption'
-        <Text type="normal" variant="caption" weight="bold">
-          {displayText}
-        </Text>,
-      );
-    } catch (error: unknown) {
```
anuraghazra, line 82: Should probably add this rule https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-expect.md

packages/blade/tsconfig-typecheck.web.json
divyanshu013: I didn't understand, it looks as if it would skip typechecking of these files, is that how it works? We want to typecheck everything

anuraghazra: Oh i see where you got confused.

So previously the `**.test.ts` files were being "excluded" from typechecking. 
The exclusions were coming from [tsconfig-generate-types.web.json](https://github.com/razorpay/blade/blob/master/packages/blade/tsconfig-generate-types.web.json#L4-L5) file. 

In this PR, I've redefined the "exclude" in `tsconfig-typecheck.web.json` and removed the `**test.ts` files.

divyanshu013: Lots of tsconfig files now, can we add some comments to these tsconfig files? I think tsconfig json supports comments:
- One comment at the top about the purpose of file
- Inline comment around exclude

packages/blade/src/components/Typography/Text/__tests__/Text.web.test.tsx
divyanshu013: Let's add

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Typography/Text/__tests__/Text.web.test.tsx
anuraghazra: Will do as a separate PR.

Issue #828, opened 2022-11-09
Default variant not loading on docs page of Internal Stories


saurabhdaware:
Ideally a default variant should be loaded to preview the changes from controllers but that is not happening in internal stories.

https://master--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/docs/components-button-iconbutton-internal--icon-button

<img width="1176" alt="image" src="https://user-images.githubusercontent.com/30949385/200747036-b11f3807-3587-42c1-965e-c0fc92a4d283.png">

https://master--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/docs/components-button-basebutton-internal--base-button

<img width="1191" alt="image" src="https://user-images.githubusercontent.com/30949385/200747111-0a6cba91-ca64-4792-8e6c-23a6011b6d04.png">

anuraghazra:
It's a race condition, since unless you enable the internal story toggle a consumer won't even be able to explore the component from the sidebar. 

But say you do this: 

1. go to storybook
2. enable internal stories
3. Click on BaseButton internal story
4. disable internal stories
5. then share the link 

Then the edge case will happen. 

Otherwise you will only share this link: 
https://master--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/docs/components-button-basebutton-internal--base-button&globals=showInternalComponents:true

anuraghazra:
Oh wait I got the bug now. 

Even when in the url the `showInternalComponents:true` is present, storybook did not enable that on initial load. Will check.

PR #917, opened 2022-12-16
fix: storybook internal component addon

anuraghazra:
fixes https://github.com/razorpay/blade/issues/828

Previously storybook wasn't able to load the global variable `showInternalComponents` on initial load of the addon. Thus even when URL was shared with the global parameter storybook wasn't rendering the story.


divyanshu013 (approved):
Looks good to me.

kamleshchandnani (approved):

Issue #272, opened 2021-05-08
[Blade/New]: consider runtime checks and showing logs only in dev mode


kamleshchandnani:
We need create a way to strip off development logs while building the library. This will also save us tons on bundlesize

divyanshu013:
Now that we're also adding lot of runtime checks, this can be quite useful

divyanshu013:
We're seeing quite some occurrences ☝️, should we prioritize this sometime soon @kamleshchandnani @nashcheez ?

nashcheez:
As the library becomes mature, this is pretty important. Let's take a task next sprint.

divyanshu013:
Another usecase of making these dev only. Sometimes users dynamically determine the props based on some API responses, etc. Say if they forget to test out a particular combination on development some end users might end up running into those in production 💥

chaitanyadeorukhkar:
Have we already done this in some of the mentioned PRs @divyanshu013 ?

divyanshu013:
No, we haven't yet.

PR #926, opened 2022-12-23
fix(blade): restrict childrens in Card component

anuraghazra:
This PR adds: 

- Restricted childrens components in Card to be one of CardHeader, CardBody, CardFooter
- Restricted childrens components in CardHeader to be one of CardHeaderLeading, CardHeaderTrailing
- Restricted childrens components in CardFooter to be one of CardFooterLeading, CardFooterTrailing

### DIFF
--- a/packages/blade/src/components/Card/Card.tsx
+++ b/packages/blade/src/components/Card/Card.tsx
@@ -40,7 +40,7 @@ export type CardProps = {
    * - Docs: https://blade.razorpay.com/?path=/docs/tokens-colors--page#-theme-tokens
    * - Figma: https://shorturl.at/fsvwK
    */
-  surfaceLevel: 2 | 3;
+  surfaceLevel?: 2 | 3;
 };
 
 const Card = ({ children, surfaceLevel = 3 }: CardProps): React.ReactElement => {
--- a/packages/blade/src/components/Card/Card.tsx
+++ b/packages/blade/src/components/Card/Card.tsx
@@ -1,8 +1,25 @@
 import React from 'react';
 import { CardSurface } from './CardSurface';
-import { CardProvider, useVerifyInsideCard } from './CardContext';
+import { CardProvider, useVerifyInsideCard, useVerifyOnlyAllowedComponents } from './CardContext';
 import Box from '~components/Box';
+import type { WithComponentId } from '~utils';
 import { metaAttribute, MetaConstants } from '~utils';
+
+export const ComponentIds = {
+  CardHeader: 'CardHeader',
+  CardHeaderTrailing: 'CardHeaderTrailing',
+  CardHeaderLeading: 'CardHeaderLeading',
+  CardFooter: 'CardFooter',
+  CardFooterTrailing: 'CardFooterTrailing',
+  CardFooterLeading: 'CardFooterLeading',
+  CardBody: 'CardBody',
+  CardHeaderIcon: 'CardHeaderIcon',
+  CardHeaderCounter: 'CardHeaderCounter',
+  CardHeaderBadge: 'CardHeaderBadge',
+  CardHeaderText: 'CardHeaderText',
+  CardHeaderLink: 'CardHeaderLink',
+  CardHeaderIconButton: 'CardHeaderIconButton',
+};
 
 export type CardProps = {
   /**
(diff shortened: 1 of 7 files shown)


packages/blade/src/components/Card/Card.tsx
```diff
@@ -23,10 +40,16 @@ export type CardProps = {
    * - Docs: https://blade.razorpay.com/?path=/docs/tokens-colors--page#-theme-tokens
    * - Figma: https://shorturl.at/fsvwK
    */
-  surfaceLevel: 2 | 3;
+  surfaceLevel?: 2 | 3;
```
anuraghazra, line 43: Set a default value but forgot to make this optional 🤦

packages/blade/src/components/Card/Card.tsx
```diff
@@ -1,9 +1,26 @@
 import React from 'react';
 import { CardSurface } from './CardSurface';
-import { CardProvider, useVerifyInsideCard } from './CardContext';
+import { CardProvider, useVerifyInsideCard, useVerifyOnlyAllowedComponents } from './CardContext';
```
kamleshchandnani, line 3: can we rename `useVerifyOnlyAllowedComponents` ➡️ `useVerifyAllowedComponents`?

packages/blade/src/components/Card/CardContext.tsx
```diff
@@ -11,9 +12,29 @@ const useVerifyInsideCard = (componentName: string): CardContextType => {
...
+  componentName: string,
+  allowedComponents: string[],
+): void => {
+  React.Children.forEach(children, (child) => {
+    const isValidChild = child && allowedComponents.includes(getComponentId(child)!);
+    if (!isValidChild) {
+      throw new Error(
+        `[Blade Card]: Only one of \`${allowedComponents.join(
+          ', ',
+        )}\` component is accepted in ${componentName} children`,
```
kamleshchandnani, line 29:
```suggestion
        )}\` component is accepted as ${componentName} children`,
```

packages/blade/src/components/Card/__tests__/Card.native.test.tsx
```diff
@@ -171,4 +171,55 @@ describe('<Card />', () => {
...
+    );
+  });
+
+  it('should restrict childrens in CardHeader', () => {
+    expect(() =>
+      renderWithTheme(
+        <Card>
+          <CardHeader>
+            <CardHeaderLeading title="" />
+            <div>some random children</div>
```
kamleshchandnani, line 196: there's never any div in Native so it will regardless throw error. can use Box instead for both web and native

kamleshchandnani (approved):
minor comments. rest LGTM

packages/blade/src/components/Card/__tests__/Card.native.test.tsx
anuraghazra: Right!

packages/blade/src/components/Card/Card.tsx
```diff
@@ -48,10 +71,11 @@ type CardBodyProps = {
   children: React.ReactNode;
 };
 
-const CardBody = ({ children }: CardBodyProps): React.ReactElement => {
+const CardBody: WithComponentId<CardBodyProps> = ({ children }) => {
```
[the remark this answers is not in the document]
anuraghazra: I made this WithComponentId<> generic to automatically attach the `componentId` static property to the React component. Previous approach didn't explicitly defined that static property on the component. 

```ts
type CounterFunction = (props: { value: number }) => React.ReactElement;

const Counter: CounterFunction = (props) => {
  return <></>
};

// With static property `componentId`

type CounterFunctionWithId = CounterFunction & { componentId: string };

const CounterWithId: CounterFunctionWithId = (props) => {
  return <></>;
};
CounterWithId.componentId = 'nice';


// Generic 

type WithComponentId<Props> = ((props: Props) => React.ReactElement) & { componentId: string; };
```

packages/blade/src/components/Card/Card.tsx
anuraghazra: Will do.

packages/blade/src/components/Card/CardContext.tsx
anuraghazra: Done.

packages/blade/src/components/Card/CardContext.tsx
```diff
@@ -11,9 +12,29 @@ const useVerifyInsideCard = (componentName: string): CardContextType => {
...
 
+/**
+ * Verify if the passed childrens are only of allowedComponents list
+ */
+const useVerifyAllowedComponents = (
+  children: React.ReactNode,
+  componentName: string,
+  allowedComponents: string[],
+): void => {
+  React.Children.forEach(children, (child) => {
```
divyanshu013, line 23: We're entering into lot of runtime checks territory. Similar to https://github.com/razorpay/blade/issues/272 I think we should start considering moving all runtime checks to development mode only /cc @kamleshchandnani

divyanshu013 (approved):
:shipit:

PR #928, opened 2022-12-26
test: add ssr tests for all the components

### LINKED ISSUE #851: infra for verifying server side rendering of components
We need an infrastructure to ensure that all the components we build work with server side rendering out of the box. As of today, we write unit tests that use `jsdom` so essentially it tests the browser environment only, on the other side we use storybook to render our components built on top of Create React App which again is client-side rendering only.

kamleshchandnani:
Add SSR tests for all the components

Fixes #851


packages/blade/src/components/Radio/__tests__/RadioGroup.ssr.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+    const { container } = renderWithSSR(
+      <RadioGroup helpText={helpText} label={labelText}>
+        <Radio value="apple">Apple</Radio>
+        <Radio value="mango">Mango</Radio>
+        <Radio value="orange">Orange</Radio>
+      </RadioGroup>,
+    );
+    // expect(getByRole('group', { name: labelText })).toBeInTheDocument();
+    // @TODO: help and label are creating issues since its read as one
+    // expect(getByText(helpText)).toBeInTheDocument();
```
anuraghazra, line 19: Let's try `expect(getByText(/select fruit/i)).toBeInTheDocument();`

packages/blade/src/components/Badge/__tests__/__snapshots__/Badge.ssr.test.tsx.snap
```diff
@@ -0,0 +1,61 @@
...
...<div display=\\"flex\\" class=\\"Boxweb__Box-xvyj85-0 iUvyUJ\\"><svg aria-hidden=\\"true\\" data-blade-component=\\"icon\\" height=\\"12px\\" viewBox=\\"0 0 24 24\\" width=\\"12px\\" fill=\\"none\\"><path d=\\"M12 11C12.5523 11 13 11.4477 13 12V16C13 16.5523 12.5523 17 12 17C11.4477 17 11 16.5523 11 16V12C11 11.4477 11.4477 11 12 11Z\\" fill=\\"hsla(216, 27%, 36%, 1)\\"></path><path d=\\"M12 7C11.4477 7 11 7.44772 11 8C11 8.55228 11.4477 9 12 9H12.01C12.5623 9 13.01 8.55228 13.01 8C13.01 7.44772 12.5623 7 12.01 7H12Z\\" fill=\\"hsla(216, 27%, 36%, 1)\\"></path><path d=\\"M1 12C1 5.92487 5.92487 1 12 1C18.0751 1 23 5.92487 23 12C23 18.0751 18.0751 23 12 23C5.92487 23 1 18.0751 1 12ZM12 3C7.02944 3 3 7.02944 3 12C3 16.9706 7.02944 21 12 21C16.9706 21 21 16.9706 21 12C21 7.02944 16.9706 3 12 3Z\\" clip-rule=\\"evenodd\\" fill=\\"hsla(216, 27%, 36%, 1)\\" fill-rule=\\"evenodd\\"></path></svg></div><p color=\\"feedback.text.neutral.lowContrast\\" font-family=\\"text\\" font-size=\\"75\\" font-weight=\\"bold\\" font-style=\\"normal\\" class=\\"BaseTextweb__StyledBaseText-sc-1ob51ew-0 kPyLsB Text__StyledText-wkh56f-0 hkbuMZ\\" data-blade-component=\\"text\\">Label</p></div></div></div>"`;
```
anuraghazra, line 3: Why there are two snapshots, 1 looks minified and 1 normal in the same test file?

packages/blade/src/components/Alert/__tests__/Alert.ssr.test.tsx
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable @typescript-eslint/no-empty-function */
+import { Alert } from '../';
+import renderWithSSR from '~src/_helpers/testing/renderWithSSR.web';
+
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
+afterAll(() => jest.restoreAllMocks());
```
anuraghazra, line 6: We should not mock the console tho 🤔 let us atleast see if there are errors on console. 

Saying this because there are certain warning and errors only happens when you do SSR. 
For example using useLayoutEffect on SSR react will throw an warning, thus we need `useIsomorphicLayoutEffect` etc.

anuraghazra (approved):
Mostly looks good, only minor comments.

packages/blade/src/components/Badge/__tests__/__snapshots__/Badge.ssr.test.tsx.snap
kamleshchandnani: because one snapshot is when we do renderToString on server so that's minified and then we render the same markup on client so that is the normal one and we do the snapshot of that as well

packages/blade/src/components/Alert/__tests__/Alert.ssr.test.tsx
kamleshchandnani: we need to do a cleanup for everything for that. Also, we can remove these from the top level and move to the test where we test the negative test cases otherwise the whole console gets filled with a negative test where we check `toThrow` things. Created #930 to track this

PR #931, opened 2022-12-30
feat(Code): use new alpha color in background

saurabhdaware:
changes the background color from gray.400 to gray.a50 

(we earlier went with `gray.400` because alpha shades were unavailable.


kamleshchandnani (approved):
lgtm 🚀

PR #932, opened 2023-01-02
feat(SelectInput): add SelectInput

saurabhdaware:
**Stack** 
- [x] #932 (This PR)
- [x] #948 
- [x] #966  
- [x] Cleanup, Storybook, Tests

**Description**

Add SelectInput as part of Dropdown components

Figma: https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=13590%3A171090&t=s10tapr4f2bsAScZ-0

Tested for-
- Form Submissions
  image
- Accessibility (Tried focussing on input with voiceover and it announces about the popup)
- Validation States of our inputs

TODO:
- [ ] Update storybook recipes (will do this at the end after all Dropdown implementations are reviewed)
- [x] ~Changeset~ (we don't need this as a version so will create one final release when dropdown is ready)


packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
```diff
@@ -68,19 +96,47 @@ export const StyledBaseInput = React.forwardRef<HTMLInputElement, StyledBaseInpu
...
   ) => {
-    return (
+    return props.as === 'button' ? (
+      <StyledBaseNativeButton
+        // @ts-expect-error: TS doesnt understand that this will always be `button`
+        ref={ref}
+        name={name}
+        type="button"
+        disabled={isDisabled}
+        required={isRequired}
```
anuraghazra, line 119: Don't think we will need to add aria-required in the button.

packages/blade/src/components/Input/SelectInput/HiddenInput.web.tsx
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components';
+
+const HiddenInput = styled.input`
```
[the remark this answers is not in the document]
saurabhdaware: so I wanted to make it absolute positioned on top right of relative box so that the error message of form validation shows up on the correct position
image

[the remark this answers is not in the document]
saurabhdaware: do you see any issues with adding `top: 0; right: 0` in VisuallyHidden? I think we can use that one itself then

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+   * @TODO handle expanded state with Dropdown
+   */
+  const isPopupExpanded = false;
+
+  const { icon, ...baseInputProps } = props;
+
+  return (
+    <Box position="relative">
+      {/* @TODO Use this for form submissions */}
+      <HiddenInput
```
anuraghazra, line 38: It'll also need tabIndex={-1} so that keyboard doesn't get focus to the hidden input.

anuraghazra:
Bug: if we don't specify placeholder the styles messes up 
image

packages/blade/src/components/Input/SelectInput/HiddenInput.web.tsx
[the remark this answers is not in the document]
anuraghazra: Didn't saw anything different it's still at the right place with VisuallyHidden 

image

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -0,0 +1,69 @@
...
+        required={props.isRequired}
+        value=""
+        // Accessibility is covered in the select input itself so we hide this field from a11y tree
+        aria-hidden={true}
```
anuraghazra, line 42: It'll be better to just directly add this on the HiddenInput.web.ts file instead of here, since this file will execute for both native & web.

saurabhdaware: I removed that file. Using `VisuallyHidden` now with if else check

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: ops yes good catch

packages/blade/src/components/Input/SelectInput/HiddenInput.web.tsx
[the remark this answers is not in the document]
saurabhdaware: oh right. the input is at the top so top left is default position right now. should work then. changed it to use visuallyhidden

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: fixed it in new commit

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -53,6 +53,8 @@ const SelectInput = (props: SelectInputProps): JSX.Element => {
         {...baseInputProps}
         as="button"
         textAlign="left"
+        value={value}
+        defaultValue="Select Option"
```
anuraghazra, line 57: Let's do this for now, but once you start working on the Dropdown, make sure to double check this. 

Since in the native HTML `select` you'll need to pass and empty string "" to signify that nothing is selected or selected value is empty.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -64,35 +71,47 @@ const KeyboardTypeToNativeValuesMap = {
...
+      prefix: props.prefix,
+      interactionElement: props.interactionElement,
+      suffix: props.suffix,
+      trailingIcon: props.trailingIcon,
+    }),
+    lineHeight: undefined,
+    textAlignVertical: 'top',
+    height: props.isTextArea
+      ? `${props.theme.typography.lineHeights.xl * (props.numberOfLines ?? 0)}px`
+      : '36px',
```
anuraghazra, line 104: Might actually be better to have a TODO comment here to fix this magic number later when we have size tokens.

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: Oh so that hidden input you see here will basically help in doing the form validation (I've kept the value blank there)

The reason being, 
- we don't have access to all options while we're rendering select so it is a bit complex to render a select there. 
- We anyway cover accessibility of combobox on the button and dropdown itself so the select will be aria-hidden
- `input` field is enough for setting the value for us. I've also seen some libraries do this. (react-select does this, radix-ui weirdly doesn't care about form submissions at all lol)

anuraghazra: Hmm 🤔  okay. 

radix does render native [select & options](https://github.com/radix-ui/primitives/blob/81b25f4b40c54f72aeb106ca0e64e1e09655153e/packages/react/select/src/Select.tsx#L1112) though so it should work as long as we pass the props like isRequired etc. 

react-select on other hand takes an autocomplete based approach where you can also type on the input thus uses input. 

Anyways let's complete the implementation by adding the dropdown, it's hard to do a proper behaviour & accessibility audit without the full implementation ready. After that we can check for bugs & edge cases.

anuraghazra (approved):

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: > react-select on other hand takes an autocomplete based approach where you can also type on the input thus uses input.

Yup but they also have the select-only ones where also they use input only
image


> radix does render native [select & options](https://github.com/radix-ui/primitives/blob/81b25f4b40c54f72aeb106ca0e64e1e09655153e/packages/react/select/src/Select.tsx#L1112) though so it should work as long as we pass the props like isRequired etc.

Strange I tried to search for `<select` on their demo page's inspect and couldn't find any 🤔 https://www.radix-ui.com/docs/primitives/components/select

saurabhdaware: > Anyways let's complete the implementation by adding the dropdown, it's hard to do a proper behaviour & accessibility audit without the full implementation ready. After that we can check for bugs & edge cases.

Yup makes sense

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -32,7 +32,7 @@ export type BaseInputProps = FormInputLabelProps &
     /**
      * Determines if it needs to be rendered as input or textarea
      */
-    as?: 'input' | 'textarea';
+    as?: 'input' | 'textarea' | 'button';
```
divyanshu013, line 35: The JSDoc comment might need an update now.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -113,11 +133,32 @@ export const StyledBaseInput = React.forwardRef<TextInput, StyledBaseInputProps>
...
     },
     ref,
   ) => {
-    return (
+    return hasPopup ? (
+      <StyledNativeBaseButton
+        // the types of styled-components for react-native is creating a mess, so there's no other option but to type `ref` as any
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ref={ref as any}
+        // @ts-expect-error: not able to type onClick for both web and native
```
divyanshu013, line 146: In this file it should be typed only for native right?

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
```diff
@@ -1,41 +1,65 @@
...
   Omit<StyledBaseInputProps, 'accessibilityProps' | 'setCurrentInteraction' | 'type'>
+>(getWebInputStyles);
+
+const StyledBaseNativeButton = styled.button<
+  Omit<StyledBaseInputProps, 'accessibilityProps' | 'setCurrentInteraction' | 'type'>
 >((props) => ({
-  ...getBaseInputStyles({
-    isDisabled: props.disabled,
+  ...getWebInputStyles(props),
+  ...getTextStyles({
```
kamleshchandnani, line 54: getTextStyles is already being called within `getBaseInputStyles` right? so couldn't understand why do we need to call it again here? 🤔

packages/blade/src/components/Input/BaseInput/types.ts
```diff
@@ -10,6 +10,7 @@ export type StyledBaseInputProps = {
   handleOnBlur?: FormInputHandleOnEvent;
   handleOnKeyDown?: FormInputHandleOnKeyDownEvent;
   handleOnInput?: FormInputHandleOnEvent;
+  onClick?: MouseEventHandler<HTMLInputElement>;
```
kamleshchandnani, line 13: are we sending the entire click event object to consumers? for other even handlers like onChange, onFocus we pass only the selected args like name, value, etc.

another thing to notice here - the reason you might not be able to type this and you have to use `@ts-expect-error` is because the event handler is diff on native and web. what we can do is we can handle this as we have done in Button

```ts
onClick?: Platform.Select<{
    web: (event: React.MouseEvent<HTMLButtonElement>) => void;
  }>;
```

or we can also just pick this from the types of button

packages/blade/src/components/Input/SelectInput/SelectInput.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import type { Meta } from '@storybook/react';
+import SelectInputComponent from './SelectInput';
+
+const SelectInputStoryMeta: Meta = {
```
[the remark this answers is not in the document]
kamleshchandnani: probably, that's because the dropdown is not implemented yet.

also, one thing to notice here is that maybe we need to focus the select input when the icon is clicked? right now it might be confusing because there's no feedback and hence probably why divyanshu is also confused

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+  | 'validationState'
+  | 'helpText'
+  | 'errorText'
+  | 'successText'
+  | 'name'
+  | 'isDisabled'
+  | 'isRequired'
+  | 'prefix'
+  | 'suffix'
+  | 'autoFocus'
```
kamleshchandnani, line 22: placeholder?

kamleshchandnani: can't find any of the click handlers? onClick, onFocus, etc?

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const platform = getPlatformType();
+
+  return (
+    <Box position="relative">
+      {/* @TODO Use this for form submissions */}
+      {platform !== 'react-native' ? (
+        <VisuallyHidden>
+          <input
+            tabIndex={-1}
+            required={props.isRequired}
```
kamleshchandnani, line 45: that native bubble looks weird since we set this 😢. probably we shall handle this with our validation state?

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+            value={value}
+            // Accessibility is covered in the select input itself so we hide this field from a11y tree
+            aria-hidden={true}
+          />
+        </VisuallyHidden>
+      ) : null}
+      <BaseInput
+        {...baseInputProps}
+        as="button"
+        textAlign="left"
```
kamleshchandnani, line 55: do we always default this to left? because on `TextInput` we let consumers decide the alignment so i think we shall follow the same behavior? otherwise in a form where all the input fields have content aligned other than left then this will look weird.

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+        value={value}
+        defaultValue="Select Option"
+        /**
+         * @TODO
+         * this will come from Dropdown component
+         */
+        id="input-123"
+        leadingIcon={icon}
+        hasPopup
```
kamleshchandnani, line 64: why do we need this prop?

kamleshchandnani (commented):
overall lgtm. some minor comments. can you also post a screenshot of select on how it renders on RN?

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
saurabhdaware: So we can't type `onClick` because it is coming from common prop so if we type it for native here, the parent component from where we're passing down the prop starts failing for web.


We can type `onPress` to have web's `onClick` but I think it is better to ignore the error than incorrectly overriding the onPress

saurabhdaware: Anyway we're not writing much of a code inside of that prop so don't see much value in having correct type there

packages/blade/src/components/Input/SelectInput/SelectInput.stories.tsx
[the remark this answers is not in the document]
saurabhdaware: yes I think we should ignore pointer events on trailing icons from BaseInput. It's not working even after implementing dropdown

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
saurabhdaware: done

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -200,6 +204,18 @@ export type BaseInputProps = FormInputLabelProps &
      * for internal metric collection purposes
      */
     componentName?: string;
+    /**
+     * Input has a popup
```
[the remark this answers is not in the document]
saurabhdaware: done

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
```diff
@@ -68,19 +95,46 @@ export const StyledBaseInput = React.forwardRef<HTMLInputElement, StyledBaseInpu
...
+          handleOnFocus?.({ name, value: event });
+        }}
+        onKeyDown={(event: React.KeyboardEvent<HTMLInputElement>) => {
+          handleOnKeyDown?.({ name, key: event.key, code: event.code, event });
+        }}
+        {...props}
+        {...accessibilityProps}
+      >
+        {props.value ? props.value : props.defaultValue}
+      </StyledBaseNativeButton>
```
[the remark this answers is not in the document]
saurabhdaware: done

packages/blade/src/components/Input/SelectInput/SelectInput.stories.tsx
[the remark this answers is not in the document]
saurabhdaware: it should work now. Treating it as an interactiveElement instead of trailingIcon.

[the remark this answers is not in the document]
saurabhdaware: https://github.com/razorpay/blade/pull/932/commits/d592ea33d6fe21f83da579994c56a6567f00899d

packages/blade/src/components/Input/SelectInput/SelectInput.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+  title: 'Components/Input/SelectInput',
```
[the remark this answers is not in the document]
saurabhdaware: We have a predefined text of "Select Option" in SelectInput. It changes to the actual selected items after selection.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -65,6 +65,10 @@ export type BaseInputProps = FormInputLabelProps &
      * The callback function to be invoked when the value of the input field changes
      */
     onChange?: FormInputOnEvent;
+    /**
+     * The callback function to be invoked when input is clicked
+     */
+    onClick?: MouseEventHandler<HTMLInputElement>;
```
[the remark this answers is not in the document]
saurabhdaware: It is anyway undefined by default so should be fine I guess? In future if we have any usecases to allow onClick we can just start passing it to BaseInput.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -113,11 +133,32 @@ export const StyledBaseInput = React.forwardRef<TextInput, StyledBaseInputProps>
       type,
       numberOfLines,
       isTextArea,
+      hasPopup,
```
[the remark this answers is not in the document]
saurabhdaware: It is custom (based on this we render a button or a input element). The user can pass it from BaseInput.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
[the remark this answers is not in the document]
saurabhdaware: Lets say I type onClick for React Native, then we start getting typescript error on web on it's parent elements and if we type it for web, we get errors on RN.

I didn't spend much time here because there isn't any code that we write here and it is anyway typed for web on `BaseInput` so we still get autocompletion outside.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
```diff
@@ -68,19 +95,46 @@ export const StyledBaseInput = React.forwardRef<HTMLInputElement, StyledBaseInpu
...
+          setCurrentInteraction('default');
+          handleOnBlur?.({ name, value: event });
+        }}
+        onFocus={(event: React.ChangeEvent<HTMLInputElement>): void => {
+          setCurrentInteraction('active');
+          handleOnFocus?.({ name, value: event });
+        }}
+        onKeyDown={(event: React.KeyboardEvent<HTMLInputElement>) => {
+          handleOnKeyDown?.({ name, key: event.key, code: event.code, event });
+        }}
```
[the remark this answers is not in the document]
saurabhdaware: yup done that now

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
saurabhdaware: So the same styles are applied on `::placeholder` as well as innerText. which is why I abstracted it out into this function and now I am calling it for `::placeholder` inside `getWebInputStyles` and adding it for innerText on Button's styles

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: I think we should ideally support native bubble. Consumers can always preventDefault during form submission (This sort of protects from scenario where someone forgets to write validations)

We're also supporting this in other inputs.

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: It decides whether to render button or input. It also adds `listbox` role and decides other props required for popups (expanded, controls, etc)

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: I don't see us using `textAlign` on any input other than `TextInput`. What was the reason to support it in TextInput? wouldn't center aligned select look a bit strange?

packages/blade/src/components/Input/BaseInput/types.ts
saurabhdaware: > are we sending the entire click event object to consumers? for other even handlers like onChange, onFocus we pass only the selected args like name, value, etc.

I just realised I don't need any other value from event property in Dropdown (earlier while building Select I was not sure so I was using entire event object) so I changed it to be similar to onFocus and onBlur events where we only pass name and value.

This also fixed TS issues since I just used the existing type helpers from forms

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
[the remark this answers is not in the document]
saurabhdaware: Update: Fixed - https://github.com/razorpay/blade/pull/932#discussion_r1085081773

saurabhdaware: Update: Changed the implementation a bit to make it similar to our onFocus and onBlur events. That fixed the TS issues - https://github.com/razorpay/blade/pull/932#discussion_r1085081773

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: Added `onClick` `onFocus` and `onBlur`.


`onChange` is added in the ActionList PR since that needs a bit special handling.

saurabhdaware: placeholder we're not supporting right now. It is always `Select Option` in the start and then changes into an actual value after selections

saurabhdaware: Also in this case, even if we support it, it will be handled by `defaultValue` prop instead of `placeholder` because text you see inside Select is more of a value and not placeholder.

divyanshu013 (approved):
Looks good to me.

PR #942, opened 2023-01-04
fix(blade): TextInput clear button state

anuraghazra:
Fixes a bug where TextInput shows clear button even when showClearButton is false with initialValue set. 

image

Fixes: https://app.asana.com/0/1200620439276878/1203628642944573/f

### DIFF
--- a/packages/blade/src/components/Input/TextArea/TextArea.tsx
+++ b/packages/blade/src/components/Input/TextArea/TextArea.tsx
@@ -70,9 +70,11 @@ const TextArea = ({
 }: TextAreaProps): React.ReactElement => {
   const inputRef = React.useRef<HTMLInputElement | TextInputReactNative>(null);
 
-  const [shouldShowClearButton, setShouldShowClearButton] = React.useState<boolean>(
-    Boolean(showClearButton && (value?.length || defaultValue?.length)),
-  );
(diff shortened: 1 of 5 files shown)


packages/blade/src/components/Input/TextArea/TextArea.tsx
```diff
@@ -70,9 +70,11 @@ const TextArea = ({
 }: TextAreaProps): React.ReactElement => {
   const inputRef = React.useRef<HTMLInputElement | TextInputReactNative>(null);
 
-  const [shouldShowClearButton, setShouldShowClearButton] = React.useState<boolean>(
-    Boolean(showClearButton && (value?.length || defaultValue?.length)),
-  );
+  const [shouldShowClearButton, setShouldShowClearButton] = React.useState(false);
+
+  React.useEffect(() => {
```
anuraghazra, line 75: Could've just put this on the useState() but IMO it's better to keep this on useEffect to avoid any stale state. 

If we don't use useEffect, The input won't rerender and it will behave strangely, this will happen if while typing someone tries to change the showClearButton. 

Can be shown in storybook: 

https://user-images.githubusercontent.com/35374649/210552166-919964d6-e94a-42a8-b350-f39b2f0ac644.mov

kamleshchandnani (approved):
LGTM :shipit:

PR #946, opened 2023-01-06
feat(icon): added transactions icon

abhieshekumar:
- Transactions Icon

image


packages/blade/src/components/Icons/TransactionIcon/TransactionIcon.tsx
```diff
@@ -0,0 +1,18 @@
+import { Svg, Path } from '../_Svg';
+import type { IconComponent } from '..';
+import useIconProps from '../useIconProps';
+
+const TransactionIcon: IconComponent = ({ size, color }) => {
```
anuraghazra, line 5: In figma and products I think we call it `Transactions` (with s). Guess will be better to name it that.

abhieshekumar: changed the name. please check.

divyanshu013 (approved):
Looks good to me.

PR #949, opened 2023-01-11
feat: added ref support for input components

anuraghazra:


packages/blade/src/hooks/useExposeInputRef.ts
```diff
@@ -0,0 +1,85 @@
...
+  targetRef: React.ForwardedRef<NativeHTMLInputRefProps>,
+  inputRef: React.RefObject<TextInputReactNative | HTMLInputElement>,
+) => {
+  React.useImperativeHandle(
+    targetRef,
+    (): NativeHTMLInputRefProps => {
+      const input = inputRef.current as HTMLInputElement & {
+        selectedOptions: HTMLSelectElement['selectedOptions'];
+      };
+      return {
```
kamleshchandnani, line 24: let's limit this to only the ones we know we need it? as an when requirement comes in we can add other options?

packages/blade/src/components/Input/PasswordInput/PasswordInput.stories.tsx
```diff
@@ -208,4 +211,49 @@ ControlledInput.parameters = {
   },
 };
 
+const PasswordInputHookFormTemplate: ComponentStory<typeof PasswordInput> = () => {
+  const { control, handleSubmit } = useForm<{ password: string }>({
+    defaultValues: { password: '' },
+  });
+
+  const { field, fieldState } = useController({
```
kamleshchandnani, line 219: this we should not add to stories but add to integrations like @divyanshu013 had added. `withReactHookForm`

packages/blade/src/hooks/useExposeInputRef.ts
anuraghazra: Yeah technically for the usecases of the PG, we need to expose focus only. 

I checked the codebase, and it seems that if you want to use uncontrolled input via the `register()` method of hook form, then you need to expose the other input related props. We can remove other props too.

packages/blade/src/components/Input/PasswordInput/PasswordInput.stories.tsx
anuraghazra: you mean added these to /examples folder?

packages/blade/src/components/Input/PasswordInput/PasswordInput.tsx
```diff
@@ -67,87 +68,98 @@ type PasswordInputProps = Pick<
...
+        // eslint-disable-next-line jsx-a11y/no-autofocus
+        autoFocus={autoFocus}
+        autoCompleteSuggestionType={autoCompleteSuggestionType}
+        keyboardReturnKeyType={keyboardReturnKeyType}
+      />
+    );
+  },
+);
+
+PasswordInput.displayName = 'PasswordInput';
```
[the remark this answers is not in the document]
anuraghazra: 🤦

packages/blade/src/hooks/useExposeInputRef.ts
```diff
@@ -0,0 +1,29 @@
...
+type NativeHTMLInputRefProps = Pick<HTMLElement, 'focus' | 'scrollIntoView'>;
+
+/**
+ * A hook which only exposes the properties of html input element via imparative hook
+ *
+ * It avoids exposing other native properties of HTMLElement
+ * like `style` `classList` to avoid unintended usage of refs.
+ */
+const useExposeInputRef = (
+  targetRef: React.ForwardedRef<NativeHTMLInputRefProps>,
```
kamleshchandnani, line 13: why do we need 2 ref objects here as args?

packages/blade/src/hooks/useExposeInputRef.ts
```diff
@@ -0,0 +1,29 @@
...
+
+type NativeHTMLInputRefProps = Pick<HTMLElement, 'focus' | 'scrollIntoView'>;
+
+/**
+ * A hook which only exposes the properties of html input element via imparative hook
+ *
+ * It avoids exposing other native properties of HTMLElement
+ * like `style` `classList` to avoid unintended usage of refs.
+ */
+const useExposeInputRef = (
```
kamleshchandnani, line 12: its not limited to input ref. it might be used on button, checkbox, radio, etc. so maybe we can rename it to `useMaskedRef` or `useSafeRef` or if you can think about better name

packages/blade/src/hooks/useExposeInputRef.ts
```diff
@@ -0,0 +1,29 @@
...
+      return {
+        focus: (opts) => input.focus(opts),
+        scrollIntoView: (opts) => input.scrollIntoView(opts),
+      };
+    },
+    [inputRef],
+  );
+};
+
+export { useExposeInputRef, NativeHTMLInputRefProps };
```
kamleshchandnani, line 29: do we need to expose the type as well?

packages/blade/src/hooks/useExposeInputRef.ts
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import type { TextInput as TextInputReactNative } from 'react-native';
+
+type NativeHTMLInputRefProps = Pick<HTMLElement, 'focus' | 'scrollIntoView'>;
```
kamleshchandnani, line 4:
```suggestion
type BladElementRefProps = Pick<HTMLElement, 'focus' | 'scrollIntoView'>;
```

packages/blade/src/components/Input/TextInput/TextInput.tsx
```diff
@@ -152,128 +154,137 @@ const isReactNative = (_textInputRef: any): _textInputRef is TextInputReactNativ
...
-      defaultValue={defaultValue}
-      value={value}
-      name={name}
-      maxCharacters={maxCharacters}
-      onChange={({ name, value }) => {
-        if (showClearButton && value?.length) {
-          // show the clear button when the user starts typing in
-          setShouldShowClearButton(true);
-        }
+    useExposeInputRef(ref, textInputRef);
```
kamleshchandnani, line 231: why do you need this here?

packages/blade/src/components/Input/TextInput/TextInput.tsx
```diff
@@ -152,128 +154,137 @@ const isReactNative = (_textInputRef: any): _textInputRef is TextInputReactNativ
...
+      onClearButtonClick,
+      isLoading,
+      suffix,
+      autoFocus,
+      keyboardReturnKeyType,
+      autoCompleteSuggestionType,
+    },
+    ref,
+  ): ReactElement => {
+    const textInputRef = React.useRef<HTMLInputElement | TextInputReactNative>(null);
```
kamleshchandnani, line 190: we can move this directly in our custom use ref hook? so all you then need to do is 

```
const textInputRef = useMaskedRef(ref)
```
and then `useMaskedRef` will internally call `React.useRef` because otherwise we'll write this at every place and also call our custom ref hook too then pass the ref to it

packages/blade/src/components/Input/TextInput/TextInput.tsx
```diff
@@ -152,128 +154,137 @@ const isReactNative = (_textInputRef: any): _textInputRef is TextInputReactNativ
...
-  onClearButtonClick,
-  isLoading,
-  suffix,
-  autoFocus,
-  keyboardReturnKeyType,
-  autoCompleteSuggestionType,
-}: TextInputProps): ReactElement => {
-  const textInputRef = React.useRef<HTMLInputElement | TextInputReactNative>(null);
-  const [shouldShowClearButton, setShouldShowClearButton] = useState(false);
+export const TextInput = React.forwardRef<NativeHTMLInputRefProps, TextInputProps>(
```
kamleshchandnani, line 157: nit: just for the sake of readability can we keep this declaration of component to how it was before? and just call forwardRef at the end? like below

```
const _TextInput = ({props},ref) => ()

export const TextInput = React.forwardRef(_TextInput,ref)
```

packages/blade/src/components/Input/TextInput/TextInput.tsx
```diff
@@ -152,128 +154,137 @@ const isReactNative = (_textInputRef: any): _textInputRef is TextInputReactNativ
...
+          type,
+          keyboardReturnKeyType,
+          autoCompleteSuggestionType,
+        })}
+      />
+    );
+  },
+);
+
+TextInput.displayName = 'TextInput';
```
kamleshchandnani, line 290: did we add this now?

packages/blade/src/components/Input/PasswordInput/PasswordInput.tsx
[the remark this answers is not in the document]
kamleshchandnani: why are we adding display name now?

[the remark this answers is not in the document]
anuraghazra: Displayname is required because ForwardRef will make it so that components will become anonymous functions.

packages/blade/src/hooks/useExposeInputRef.ts
anuraghazra: targetRef is the ref which will get exposed to the user. 

and inputRef is the one which has the actual internal ref to the html input element. If we just want to directly expose the ref we won't need to do this but since we are using useImparativeHandle hook we need it.

packages/blade/src/hooks/useExposeInputRef.ts
anuraghazra: 🤔 but it's only for inputs tho, checkbox/radios are just basically <inputs \/> and button will not be exposed like that with the ref right?

packages/blade/src/hooks/useExposeInputRef.ts
anuraghazra: Yes since we need to type the ForwardRef's ref.

packages/blade/src/components/Input/TextInput/TextInput.tsx
anuraghazra: Okay. makes sense.

packages/blade/src/hooks/useExposeInputRef.ts
kamleshchandnani: why not?

anuraghazra: wait are we exposing refs for Button component too?

packages/blade/src/components/Input/PasswordInput/PasswordInput.stories.tsx
anuraghazra: done.

packages/blade/src/hooks/useExposeInputRef.ts
anuraghazra: no need to two args now since we create & return the inputRef.

packages/blade/src/hooks/useExposeInputRef.ts
kamleshchandnani: yes we have to.

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -36,22 +39,28 @@ type ButtonWithIconProps = ButtonCommonProps & {
...
+    isLoading = false,
+    onClick,
+    size = 'medium',
+    type = 'button',
+    variant = 'primary',
+    accessibilityLabel,
+  },
+  ref,
+) => {
+  const buttonRef = useBladeInnerRef(ref);
```
kamleshchandnani, line 58: shouldn't this be part of `BaseButton` and not the Button component itself?

packages/blade/src/components/Checkbox/Checkbox.tsx
```diff
@@ -97,21 +99,25 @@ type CheckboxProps = {
...
+    onChange,
+    value,
+    children,
+    helpText,
+    errorText,
+    size = 'medium',
+  },
+  ref,
+) => {
+  const inputRef = useBladeInnerRef(ref);
```
kamleshchandnani, line 120: what if we move this to SelectorInput then all we'll have to do is forward the ref arg and then handle it inside SelectorInput so that way even the radio will get it?

packages/blade/src/components/Input/PasswordInput/PasswordInput.tsx
```diff
@@ -5,8 +5,8 @@ import { EyeIcon, EyeOffIcon } from '~components/Icons';
 import Box from '~components/Box';
 import { CharacterCounter } from '~components/Form/CharacterCounter';
 import { IconButton } from '~components/Button/IconButton';
-import type { NativeHTMLInputRefProps } from '~src/hooks/useExposeInputRef';
-import { useExposeInputRef } from '~src/hooks/useExposeInputRef';
+import type { BladeRefElement } from '~src/hooks/useBladeInnerRef';
```
kamleshchandnani, line 8:
```suggestion
import type { BladeElementRef } from '~src/hooks/useBladeInnerRef';
```
goes well with `useBladeInnerRef` where the first part defines the name of the variable and then the suffix defines the type of it which in this case is `Ref`

kamleshchandnani:
@anuraghazra lets add test cases for testing refs to each component

packages/blade/src/components/Button/Button/Button.tsx
anuraghazra: Both are React components so we have to wrap React.ForwardRef on both Button & BaseButton. 

In Button (public facing) i've used useBladeInnerRef,
and for BaseButton i've directly exposed the native ref without any constraints.

packages/blade/src/components/Checkbox/Checkbox.tsx
anuraghazra: I'll try it out and see if it could work.

packages/blade/src/components/Input/PasswordInput/PasswordInput.tsx
anuraghazra: Makes sense!

packages/blade/src/components/Button/Button/Button.tsx
kamleshchandnani: okay what i mean is in public facing API we have to use forwardref that is fine but `const buttonRef = useBladeInnerRef(ref);` this part can remain inside BaseButton only? and in this public facing component we can just take the ref and pass it to BaseButton just forward it as it is because BaseButton anyways is the place where the actual ref gets assigned to the native HTML button element

packages/blade/src/components/Checkbox/Checkbox.tsx
anuraghazra: Works, though it doesn't simplify the code, just moves the useBladeInnerRef to SelectorInput. 

I'm fine with either of them. will push the change.

packages/blade/src/components/Button/Button/Button.tsx
anuraghazra: I see, yeah that works. Same as https://github.com/razorpay/blade/pull/949#discussion_r1070920063 

Though just wanted to keep the BaseButton's ref more flexible and expose the ref as is since it's for internal use. 
Not a big deal tho, we can change later if requirements comes.

.changeset/seven-comics-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
```
kamleshchandnani, line 2: make it minor since this is an enhancement

packages/blade/src/components/Button/Button/__tests__/Button.web.test.tsx
```diff
@@ -117,4 +118,34 @@ describe('<Button />', () => {
...
+            }}
+          >
+            Focus
+          </Button>
+        </>
+      );
+    };
+    const { getByRole } = renderWithTheme(<Example />);
+
+    const button = getByRole('button', { name: label });
```
kamleshchandnani, line 143: either have label as a const for everything or use inline for everything 😅.

packages/blade/src/components/Checkbox/__tests__/Checkbox.web.test.tsx
```diff
@@ -160,4 +161,35 @@ describe('<Checkbox />', () => {
       (getByRole('checkbox', { name: labelText }) as HTMLInputElement).indeterminate,
     ).toBeTruthy();
   });
+
+  it(`should expose native element methods via ref`, async () => {
+    const label = 'Accept';
```
kamleshchandnani, line 166: same as above. either use inline labels or extract them as consts

packages/blade/src/components/Radio/__tests__/Radio.web.test.tsx
```diff
@@ -64,4 +65,39 @@ describe('<Radio />', () => {
...
+
+  it(`should expose native element methods via ref`, async () => {
+    const label = 'Accept';
+
+    const Example = (): React.ReactElement => {
+      const ref = React.useRef<HTMLInputElement>(null);
+
+      return (
+        <>
+          <RadioGroup label="focus test">
```
kamleshchandnani, line 77: how does focus work in the case of the Radio and Checkbox group? does the 1st child gets the focus?

kamleshchandnani (commented):
We need 1 more test cases for the `useBladeInnerRef` hook where we test that only allowed methods are accessible and the ones which are not allowed throws error

packages/blade/src/components/Radio/__tests__/Radio.web.test.tsx
anuraghazra: RadioGroup is not focusable. Radio only has the input. 

Users will add the ref to the Radio component and handle the focus if they need to. 

In this test I've used RadioGroup just because Radio cannot be used outside of RadioGroup

anuraghazra:
>  1 more test cases for the useBladeInnerRef hook where we test that only allowed methods are accessible and the ones which are not allowed throws error

Added test for the hook separately and tested for the focus/scrollIntoView. 

Though not really feasible to throw a custom error for each and every method which are present on the HTMLElement interface, there are quite a lot of them.

packages/blade/src/components/Radio/__tests__/Radio.native.test.tsx
```diff
@@ -60,4 +61,29 @@ describe('<Radio />', () => {
...
+  it('should expose native element methods via ref', () => {
+    let refValue = null;
+    const Example = (): React.ReactElement => {
+      const ref = React.useRef<HTMLInputElement>(null);
+      return (
+        <RadioGroup label="Hello">
+          <Radio
+            value="apple"
+            ref={(value) => {
+              console.log(value);
```
kamleshchandnani, line 74: remove log?

packages/blade/src/components/Radio/__tests__/Radio.native.test.tsx
```diff
@@ -60,4 +61,29 @@ describe('<Radio />', () => {
...
+            }}
+          >
+            Apple
+          </Radio>
+        </RadioGroup>
+      );
+    };
+
+    renderWithTheme(<Example />);
+    expect(refValue).toBeNull();
```
kamleshchandnani, line 87: why null? there's no focus on native for radio?

packages/blade/src/components/Radio/__tests__/Radio.native.test.tsx
```diff
@@ -60,4 +61,29 @@ describe('<Radio />', () => {
...
+      const ref = React.useRef<HTMLInputElement>(null);
+      return (
+        <RadioGroup label="Hello">
+          <Radio
+            value="apple"
+            ref={(value) => {
+              console.log(value);
+              // @ts-expect-error
+              ref.current = value;
+              refValue = value;
```
kamleshchandnani, line 77: why this?

packages/blade/src/hooks/__tests__/useBladeInnerRef.web.test.tsx
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable react/display-name */
+import React from 'react';
+import type { BladeElementRef } from '../useBladeInnerRef';
+import { useBladeInnerRef } from '../useBladeInnerRef';
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.web';
+
+describe('useBladeInnerRef', () => {
+  it('should only expose focus and scrollIntoView', () => {
+    let refValue = null;
```
kamleshchandnani, line 9: why not 
```suggestion
    let ref = React.useRef(null);
```

packages/blade/src/hooks/useBladeInnerRef.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A hook which only exposes the properties of html input element via imparative hook
+ *
+ * It avoids exposing other native properties of HTMLElement
+ * like `style` `classList` to avoid unintended usage of refs.
+ */
+const useBladeInnerRef = (
+  targetRef: React.ForwardedRef<BladeElementRef>,
+): React.RefObject<BladeElementRef> => {
+  const textInputRef = React.useRef<HTMLInputElement | TextInputReactNative>(null);
```
kamleshchandnani, line 15:
```suggestion
  const innerRef = React.useRef<HTMLInputElement | TextInputReactNative>(null);
```

packages/blade/src/hooks/useBladeInnerRef.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  React.useImperativeHandle(
+    targetRef,
+    (): BladeElementRef => {
+      const input = textInputRef.current;
```
kamleshchandnani, line 20:
```suggestion
      const element = textInputRef.current;
```

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
```diff
@@ -57,20 +60,30 @@ const StyledInput = styled.input<HoverProps>(({ theme, isChecked, isDisabled, ha
...
     <StyledInput
       isChecked={isChecked}
       isDisabled={isDisabled}
       hasError={hasError}
       {...inputProps}
+      // merging both refs because inputProps.ref needs to have access to indeterminate state
+      // to be able to set the mixed value via setMixed() function
+      // TODO: replace with a generic `mergeRefs()` util if we do this in other places
+      ref={(value) => {
+        inputProps.ref.current = value;
```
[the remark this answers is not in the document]
anuraghazra: > where is this inputprops coming from? 

`inputProps` is coming form the `useCheckboxHook` which sets/handles all the core logic for the checkbox DOM element. 


image

> why are there 2 refs now?

We need the `inputProps.ref` because we need to set the indeterminate state of the input element which can only be set via the DOM element attribute. `inputRef.indeterminate = true;`

There are two refs:

1. the inputProps.ref which we need to handle the indeterminate state
2. the one which gets exposed to the public (with useBladeInnerRef)

packages/blade/src/components/Radio/__tests__/Radio.native.test.tsx
anuraghazra: There is no input primitives for react-native radio/checkbox, there's only textual inputs on rn. so no ref.

packages/blade/src/hooks/__tests__/useBladeInnerRef.native.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+    let refValue = null;
+    const CustomButton = React.forwardRef<BladeElementRef>((_props, ref) => {
+      const innerRef = useBladeInnerRef(ref);
+
+      return <View ref={innerRef as React.RefObject<View>}>click</View>;
+    });
+    const Example = (): React.ReactElement => {
+      return (
+        <CustomButton
+          ref={(value) => {
```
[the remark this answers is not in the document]
anuraghazra: Not for that reason. I just want to extract the value of the ref and save it in a local variable so that I can assert & check it later.

packages/blade/src/hooks/__tests__/useBladeInnerRef.web.test.tsx
anuraghazra: If you notice, it's a temporary variable outside the React component. I'm using this variable to hold the value of the ref so that I can assert it later. `expect(refValue).toHaveProperty('focus');`

packages/blade/src/components/Radio/__tests__/Radio.native.test.tsx
anuraghazra: https://github.com/razorpay/blade/pull/949#discussion_r1071866113

packages/blade/src/components/Input/PasswordInput/PasswordInput.tsx
```diff
@@ -67,30 +68,34 @@ type PasswordInputProps = Pick<
...
+    onFocus,
+    onBlur,
+    name,
+    autoFocus = false,
+    keyboardReturnKeyType = 'done',
+    autoCompleteSuggestionType,
+  },
+  ref,
+) => {
+  const inputRef = useBladeInnerRef(ref);
```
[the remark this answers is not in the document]
anuraghazra: Yeah we can do that. at this point we are doing it on all the other components.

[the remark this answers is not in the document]
anuraghazra: Okay wait it's actually simpler to do this in here only. 

Because, for example take TextInput: 

We need the ref access in the TextInput itself to handle `textInputRef.current.clear()` & `textInputRef.current.focus()` for the clear button, plus we also need to pass down ref to the BaseInput to be able to populate the ref with the DOM element. 

If we remove useBladeInnerRef() here and move it to BaseInput, then in TextInput we need to merge the incoming ref & the textInputRef: 

````ts
const _TextInput = (
  props,
  incomingRef,
) => {
  const textInputRef = React.useRef<HTMLInputElement | TextInputReactNative>(null);
  ...
  ...
  const renderInteractionElement = () => {
    // here we need access to textInputRef
    // textInputRef.current.focus() / clear()
  }

  return <BaseInput ref={mergeRefs([incomingRef, textInputRef]} /> // need to merge ref here
}
````

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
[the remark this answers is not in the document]
kamleshchandnani: ah ok makes sense

packages/blade/src/components/Input/PasswordInput/PasswordInput.tsx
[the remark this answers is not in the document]
kamleshchandnani: alright this makes sense now. lets not change it then

kamleshchandnani (commented):
this lgtm now :shipit:

kamleshchandnani (approved):
phew 😅

PR #950, opened 2023-01-13
docs: rfc for layout

divyanshu013:
Layout primitives and components RFC.

- [View formatted markdown](https://github.com/razorpay/blade/blob/div/layout-rfc/rfcs/2023-01-06-layout.md)


rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+There can be multiple ways to implement this in practice, for example:
+
+1. Instead of using `style`, `className` could be used by consumers preferring to keep CSS separate.
+1. React Native consumers might be using `View` instead of `div` with [`StyleSheet`](https://reactnative.dev/docs/stylesheet) API.
+1. Some may be using `styled-components` and instead of writing styles inline creating [wrapper components](https://styled-components.com/docs/basics#getting-started) like `FlexContainer`, `ButtonContainer` with `styled.div` API and applying the style attributes there.
+1. Some may have a custom `Flex` component that accepts props such as `flexDirection`, `alignItems` or a custom `Box` component for applying spacing with a `marginTop` prop.
+
+Considering the variety of ways this can be implemented, few things are quite cumbersome:
+
+1. Need for a lot of wrapper components _(wrapocalypse)_ for spacing and alignment, which quickly grows out of hand as UI evolves. One can't add styling for layout directly on a Blade component because we don't support `className` or `style` prop, which makes wrappers a necessity even for small things.
```
chaitanyadeorukhkar, line 73: > wrapocalypse

Masterpiece 👏

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+- Built-in support for responsive styles on theme aware properties by either accepting an array / object format for values at different breakpoints
+- Finally, some design systems also support a `style` or `className` prop on components for overrides and layouts for maximum flexibility
+
+A brief study of some layout primitives is compiled in the [references](#references) section below.
+
+### Layout primitives and components for Blade
+
+In order to provide a good balance of flexibility and constraints within blade we should support the theme aware components and primitives listed below.
+
+#### `Box` component
```
chaitanyadeorukhkar, line 136: Do we really need a generic `Box` if we are offering `Flex`, `Grid` & `Stack`?

Is there any use-case we see where `Flex`, `Grid` & `Stack` won't be enough and a developer would need to use a `Box` instead?

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+
+> **Note**
+>
+> Not all CSS grid properties are covered in `styled-system` spec. These should be referenced from MDN during implementation. Here's a nice reference from [CSS Tricks](https://css-tricks.com/snippets/css/complete-guide-grid).
+
+**Spacing values**
+
+- All spacing based properties should accept spacing tokens i.e. `spacing.1`, `spacing.2`, etc.
+- TypeScript auto suggestion hints should also show the `px` value associated with a spacing token, this helps in translating design from Figma to code since designs would have these in `px`.
+- They should also accept `number` or `string` values eg. `1`, `'1px'`, `'100%'`, `'auto'`, `'1fr'` etc. Numeric literals imply the pixel value (as in vanilla React), if someone wants to use a spacing token it must be prefixed like `spacing.*`.
```
chaitanyadeorukhkar, line 175: imo we shouldn't accept unitless values like `1`. If the consumers are choosing to not rely on our tokens and pass their custom value, it would be better if they also specify the unit along with it instead of us appending a unit to it later on. 

The only case where we should do any logical computation here is when a consumer passes a spacing token, otherwise take the value as is

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+
+**Negative values**
+
+There are some usecases where you may want to pass negative values (eg. negative margin values to create full-bleed layouts that break out of main content). In order to support this we would need spacing values to support negative tokens as well. Instead of negative spacing tokens like `-spacing.1`, `-spacing.2`, etc. we'll let users pass negative literals or pixel values directly eg. `-4`, `-4px`, etc.
+
+**Few points to note for React (web)**
+
+- We should avoid supporting shorthands to save from confusions _(should I use `mb` or `marginBottom`)_.
+- Flex gap has ~91% global [coverage](https://caniuse.com/?search=flex%20gap) (support added in some browsers mid 2020). Although the support looks decent, this might be problematic for some products supporting these browsers. There is a [post CSS plugin](https://github.com/gavinmcfarland/flex-gap-polyfill) which can polyfill this _(with some known limitations)_. Alternative is to internally translate gap properties to utilize CSS selectors such as [lobotomized owl](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/) `* + *`.
```
chaitanyadeorukhkar, line 185: Let's let the consumers take responsibility for the polyfill. If they're using flex gap (or any other property for that matter), they should be aware of its browser compatibility and use it accordingly. We should worry about this property only if we're also using it internally for a blade component since that would reduce it's coverage and impact our consumers using the component.

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+- Going forward with native `gap` supports seems like a fine choice here in favor of keeping things simple and native. Tradeoff being, older browsers will be able to render the layout with a degraded experience _(without spacing)_.
+- CSS grid has ~96% global [coverage](https://caniuse.com/?search=css%20grid). Although we can only natively support it on web, it's worth the investment since it can simplify building grid based layouts.
```
chaitanyadeorukhkar, line 187: Most of the times the layout logic will be reused on desktop web & mweb. It would be extremely rare where layout logic is shared between web & react native for us. We can take this tradeoff for supporting grid on web only 👍

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+
+**Few points to note for React Native**
+
+- React Native [recently added](https://reactnative.dev/blog/2023/01/12/version-071) native support for `gap` specific properties in `v0.71`.
```
chaitanyadeorukhkar, line 191: We would need to add to our docs that Blade works only with RN `v0.71+` once we add layout components. We'll put this out as a minimum requirement for our mobile apps while onboarding.

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+- Polyfilling this on native for older releases seems bit hacky _(with cloning and wrapping child elements)_ and unneeded because unlike browsers, all users can receive an updated React Native v0.71 app, tradeoff being we will only be supporting `gap` based props in React Native going forward from `0.71`, so consumer apps should update their React Native version in order to consume these APIs.
+- No support for grid props since it isn't supported natively.
+- The TypeScript typings and autosuggestions should appropriately be adjusted for web and native to accommodate the differences, eg. grid properties, `inline-flex`, `block` display properties won't be supported. This can be done with [`moduleSuffixes`](https://www.typescriptlang.org/tsconfig#moduleSuffixes) detailed in the [related RFC](./writing-cross-platform-typescript.md).
+
+We have an [internal Box component](../packages/blade/src/components/Box/) but it exposes lot more properties than the ones associated with layout. It seems like a good idea to have a different public `Box` component with the restrictive but fully covered layout properties from above table.
```
chaitanyadeorukhkar, line 196: Mentioned this above too. We should evaluate if we really need to expose a Box component as long as we have other layout components in place.

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+</Box>
+```
+
+Supported properties:
+
+| Prop     | Description                        |
+| -------- | ---------------------------------- |
+| `size`   | Sets both height and width equally |
+| `height` | Sets height                        |
+| `width`  | Sets width                         |
```
chaitanyadeorukhkar, line 233: Why do we need to add width & height support on spacer component? Shouldn't it instead expose margin & padding properties?

```jsx
<Spacer 
    margin={{ top: 'spacing.1', right: '3px', bottom: '5%', left: '2rem' }} 
    padding={{ top: 'spacing.1', right: '3px', bottom: '5%', left: '2rem' }}
```

Probably also along with responsive support

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+All properties are optional and accepts only spacing tokens. If no prop is provided the spacer is set to `flex: 1`, useful for building layouts where you might want a `Spacer` to fill all available space in between siblings.
+
+#### Responsive styles
+
+In order to ease up building responsive UIs all the style prop values will also support an alternate **object based format** that accepts breakpoints specific values, keeping mobile-first approach (`@media(min-width)`). Sample usage:
+
+```tsx
+<Box
+  display="flex"
+  flexDirection={{ base: 'column', l: 'row' }}
```
chaitanyadeorukhkar, line 244: Should we avoid using shorthands here for `l`, `m`?

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+- Separate utility component for hiding (same can be achieved with responsive properties on `display` set to `none`, `flex`, `block`, etc.)
+- Page level layout components (nav, header, sidebar, footer, etc.)
+
+### Roadmap
+
+Since layout components have a huge scope we should target to launch these progressively. This also gives us room to iterate based on feedback from consumers.
+
+Here's a proposed timeline:
+
+1. `Box`, `Flex`, `Grid` component
```
chaitanyadeorukhkar, line 335: We could skip `Box` & deliver `Spacer` instead

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+1. `Spacer` component
+1. Responsive styles
```
chaitanyadeorukhkar, line 337: Shouldn't responsive styles be supported from day 1 for all layout components? Why release this iteratively? Even though this is a frequently asked component, its a non-blocker since folks are already using workarounds. We should instead build this out properly in the first go to avoid API changes in iterations.

I think we should do
1. `Flex` & `Grid`(web only) components with responsive styles support
2. `Spacer` component with responsive styles support
3. `sx` prop with responsive styles support

`useSx`, we can take it up later if valid use-cases come in

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+- Since we're also accepting numeric values it can lead to some confusions _(does this `2` imply `2px` or `spacing.2`)_. Rationale is to keep the current behavior closer to vanilla React where `2` gets translated to `2px`, users have to specific when using spacing tokens with `spacing.2`, etc.
+- We've constrained the supported properties both on layout components and `sx` prop in order to avoid unexpected usage, look and feel of components. This might feel too restrictive to some users compared to other general purpose design systems. The idea is to iterate and evolve what all properties should be supported based on feedback and usecases we see as users start using these primitives.
+- Minor caveat of `sx` and object based responsive properties, users will pass object literals so it would re-render all the time, however not a concern for now
+
+As a contrasting philosophy, Primer has an interesting take on supporting all properties _which means you can actually style a button to look something very different from a button_:
+
+> Assume that people will break the rules, provide safe ways for them to do so
```
chaitanyadeorukhkar, line 352: May the force be with you 🙏

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+>
+> - [Primer](https://primer.style/react/philosophy#assume-that-people-will-break-the-rules-provide-safe-ways-for-them-to-do-so)
+
+## Alternatives
+
+- Not doing this and let users manage layout on their own (non ideal because of reasons mentioned in [motivation](#motivation)).
```
chaitanyadeorukhkar, line 358: Adding this to every component API going forward 🤣

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+- Instead of supporting `sx` prop we could consider adding style props directly on components like we're doing for `Box` component. This primarily helps with separating layout properties in a separate prop which is focussed purely for this.
```
chaitanyadeorukhkar, line 359: That exposes the component to a lot of misuse and alterations

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+- Supporting all possible CSS properties, this gives maximum flexibility but also offers least constraints and therefore can lead to inconsistent UIs.
+- Having a different API for accepting arbitrary values in layout props or keys like `unsafeSpacing`, `unsafeMarginTop` etc. However this would mean there would be multiple ways to do the same thing which could be used together like `marginTop`, `unsafeMarginBottom` which seems confusing. Also the friction here doesn't help in ensuring consistency. There can be genuine usecases where people might want to build custom components et al.
```
chaitanyadeorukhkar, line 361: Yeah. Better to keep the API clean instead.

chaitanyadeorukhkar (commented):
The RFC is quite thorough and I loved the narrative explaining the problems, solutions and alternatives 👏

rfcs/2023-01-06-layout.md
divyanshu013: For web, `Flex` might not always be a good fit. `Box` can be helpful for creating `block`, `inline-block`, `none` layouts.

On native, everything would follow a `flex` layout but not for web, so wrapping any element with `Flex` may result in unwanted changes. For example:

```tsx
<Box marginTop="spacing.2">
  {/* here using a Flex would change the child to flex instead of block */}
  <MyCustomComponent />
</Box>
```

rfcs/2023-01-06-layout.md
divyanshu013: React [automatically transforms](https://reactjs.org/docs/dom-elements.html#style) unitless values to pixel values, so I thought it might be good to keep this closer to that. A related downside is we'll have to add runtime checks to throw in case of numeric values (for JS projects). But yes, it does have the downside which I also added in drawbacks - leading to confusions _does this 1 mean 1 px or `spacing.1`_.

rfcs/2023-01-06-layout.md
divyanshu013: The polyfill has some known limitation plus integration efforts for post css, so it won't work for all cases. Most foolproof alternative I find is internally handling the `gap` if we've to with `* + *` margins but its spacing algorithm would be different from the one provided by flexbox.

Chakra handles it in a way where they have a `Flex` component which uses the native `gap` but their `Stack` component uses `* + *`. Along similar lines I feel if we're providing a `Flex` component it should use the native gap functionality. If a product has legit usecases for supporting older browsers without compromising on the UI they can use margin properties on siblings.

rfcs/2023-01-06-layout.md
divyanshu013: Yup, these props should all support responsive properties. Regarding the prop names I feel `height` and `width` seems closer to the purpose of component.

Margin and padding seems like an overkill here because we're not rendering anything inside the spacer (just space).
- Should I use margin or padding for adding this space?
- If I need to add `spacing.2` space in between should I use `marginTop`, `marginBottom`, `paddingTop`, `paddingBottom` etc. Since I know the space I should be able to directly specify that with `size` or `height`
- Margin and padding are applied twice in a direction (top, bottom when vertically and left, right when horizontally) which will add another area to think about if applying spacing with margin padding properties

rfcs/2023-01-06-layout.md
divyanshu013: These replicate our breakpoints tokens so I feel we should use the same convention.

rfcs/2023-01-06-layout.md
divyanshu013: This can definitely work, we can club responsive properties in first iteration.

rfcs/2023-01-06-layout.md
kamleshchandnani: +1

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+- Polyfilling this on native for older releases seems bit hacky _(with cloning and wrapping child elements)_ and unneeded because unlike browsers, all users can receive an updated React Native v0.71 app, tradeoff being we will only be supporting `gap` based props in React Native going forward from `0.71`, so consumer apps should update their React Native version in order to consume these APIs.
+- No support for grid props since it isn't supported natively.
+- The TypeScript typings and autosuggestions should appropriately be adjusted for web and native to accommodate the differences, eg. grid properties, `inline-flex`, `block` display properties won't be supported. This can be done with [`moduleSuffixes`](https://www.typescriptlang.org/tsconfig#moduleSuffixes) detailed in the [related RFC](./writing-cross-platform-typescript.md).
+
+We have an [internal Box component](../packages/blade/src/components/Box/) but it exposes lot more properties than the ones associated with layout. It seems like a good idea to have a different public `Box` component with the restrictive but fully covered layout properties from above table.
+
+#### `Flex` component
```
kamleshchandnani, line 198: I think we shall expose one of Flex and Grid depending on what is supported on both web and native platforms. exposing too many things for similar purpose might confuse people.

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+
+A helper component over the `Box` component with `display` set as `Flex`. Arguably, this component seems unneeded without flex specific shorthand properties because the behavior can be achieved with a `display="flex"` prop on `Box` but it improves code readability. Supports same properties.
+
+#### `Grid` component
+
+A helper **web only** component over the `Box` component with `display` set as `grid`. Similar to `Flex` component, it may be argued this component is unneeded with `display="grid"` set on `Box` but it improves code readability. Supports same properties.
+
+#### `Spacer` component
```
kamleshchandnani, line 206: if we plan to expose layout props(like `sx`) on all of our elements then do we see a need for this component?

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+  paddingX="spacing.1"
+  gap={{ base: 'spacing.1', m: 'spacing.2', l: 'spacing.3' }}
+>
+  <Button>Button 1</Button>
+  <Button>Button 2</Button>
+  <Button>Button 3</Button>
+</Box>
+```
+
+The keys are derived from **breakpoints tokens**. An extra `base` key can be added as a helper to define values for breakpoints below the ones provided, for example `{ base: 'column', l: 'row' }` implies the value is `column` for all breakpoints less than `l`. All wider viewports will inherit largest provided breakpoint value because of `min-width`.
```
kamleshchandnani, line 257: should we make it more simpler like - we provide `default` as the key which will be applied to every breakpoint and then anything that needs to be overridden can be specified explicitly by the consumers?  I've always found this approach confusing even on other external systems.

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+| Spacing             | `margin*`                                                                      | `marginTop`, `marginLeft`, `marginRight`, `marginBottom`, `marginX`, `marginY` (React Native `marginHorizontal`, `marginVertical` mapped to these)                  |
+| Flexbox             | `alignSelf`, `justifySelf`, `order`                                            | Only restricted set of properties applicable on flex children. Note: `justifySelf` doesn't work for flex layout but it can be used for block layout (web only).     |
+| Position            | `position`, `zIndex`, `top`, etc.                                              | [List of props](https://github.com/styled-system/styled-system/blob/master/docs/table.md#position) referenced from styled-system.                                   |
+| Grid **(web only)** | `gridColumn*`, `gridRow*`, `gridArea`, `justifySelf`, `alignSelf`, `placeSelf` | Only restricted set of properties applicable on [grid children](https://css-tricks.com/snippets/css/complete-guide-grid/#aa-properties-for-the-childrengrid-items). |
```
kamleshchandnani, line 289: do these props make sense on presentational components? usually you'll use this prop mostly for spacing and for everything else we have other primitive components right?

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,461 @@
...
+| Layout              | None                                                                           |                                                                                                                                                                     |
+| Gap                 | None                                                                           |                                                                                                                                                                     |
+
+#### `useSx` hook
+
+A `useSx` hook, inspired from [Dripsy](https://www.dripsy.xyz/apis/useSx#recommendations) can be helpful for targeting some minor usecases where users may want to style custom components or components which aren't directly accessible to be styled (for eg. `contentContainerStyle` on `FlatList`). I don't suspect a lot of usecases of this prop currently so this can be taken up later as en enhancement. Sample usage:
```
kamleshchandnani, line 295: yup i think we can skip this for now

rfcs/2023-01-06-layout.md
kamleshchandnani: I'm still wondering why do we need these many components? Majorly we need
- Flex/Box(anyone) with responsive capabilities
- spacing props for presentational components eg: Button, Text, etc.

Even for Flex I would propose we can have a general purpose name, something like View as reshaped/spectrume uses it. and then use it for Flex things as well has other props from box like position props, background color, etc.

rfcs/2023-01-06-layout.md
divyanshu013: Updated spec to disallow `number` values in https://github.com/razorpay/blade/pull/950/commits/f4e3b6cebb79e8d1a3b72ed54eb53ed67c55dec4

rfcs/2023-01-06-layout.md
divyanshu013: Updated in https://github.com/razorpay/blade/pull/950/commits/f4e3b6cebb79e8d1a3b72ed54eb53ed67c55dec4 to clarify this will be a native `gap` property with an alternative to rely on margin properties otherwise

rfcs/2023-01-06-layout.md
divyanshu013: Updated in https://github.com/razorpay/blade/pull/950/commits/f4e3b6cebb79e8d1a3b72ed54eb53ed67c55dec4

rfcs/2023-01-06-layout.md
divyanshu013: Closing this, let's continue discussion over there.

rfcs/2023-01-06-layout.md
divyanshu013: Updated https://github.com/razorpay/blade/pull/950/commits/f4e3b6cebb79e8d1a3b72ed54eb53ed67c55dec4 to not expose `Spacer` from our side but in RFC have kept it as a recommendation which users may build over `Box` as a local abstraction.

rfcs/2023-01-06-layout.md
divyanshu013: Updated roadmap in https://github.com/razorpay/blade/pull/950/commits/f4e3b6cebb79e8d1a3b72ed54eb53ed67c55dec4

rfcs/2023-01-06-layout.md
divyanshu013: Updated roadmap in https://github.com/razorpay/blade/pull/950/commits/f4e3b6cebb79e8d1a3b72ed54eb53ed67c55dec4

Current proposal removes these helper components from our side. Though it has slight side effect on readability, listed in drawbacks.

rfcs/2023-01-06-layout.md
divyanshu013: Updated in https://github.com/razorpay/blade/pull/950/commits/f4e3b6cebb79e8d1a3b72ed54eb53ed67c55dec4 with a note on readability in drawbacks

rfcs/2023-01-06-layout.md
divyanshu013: Gave this more thought. I feel these can cause more confusions compared to the min-width approach.

For example if a layout is supposed to appear horizontally across all sizes greater than `m` people will have to specify the prop for all breakpoints like `l`, `xl`, etc.

Some may also forget to pass the value for `xl` expecting it would keep the value from previous breakpoint but instead it would pick the `default` value and make the layout vertical at `xl`.

rfcs/2023-01-06-layout.md
divyanshu013: Yup, we would need the children specific props such as `gridColumnStart`, `gridRowEnd`, etc. [Ref](https://css-tricks.com/snippets/css/complete-guide-grid/#aa-properties-for-the-childrengrid-items)

kamleshchandnani (approved):
LGTM :shipit:

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,450 @@
...
+| Sounds quite generic                                                      | Sounds layout specific                                      | Sounds native only - related to React Native `View` |
+| Users might expect this to support all CSS properties as in other systems | Can be reasoned to be limited to layout specific properties | Similar to `Box`                                    |
+
+> There are only two hard problems in software engineering - cache invalidation, naming things and off by one error.
+
+Considering the scope of `Box` may be expected to increase over time I'm leaning towards going ahead with `Box`. This also has an added benefit of incremental migration _(new properties will be available on this component)_. Plus it's a relatively well known terminology in the community. `Layout` does address the issue of reasoning about restricted set of properties but if we come across the need for having other CSS properties _(as usecases grows)_ they won't fit within `Layout` and we may need other components to support them. So, just by calling this primitive `Layout` it won't necessarily contain the need for other properties in building UIs.
+
```
saurabhdaware, line 339: Should we add a short answer here based on the discussion?

saurabhdaware (approved):
Looks good to me. Everything is so well documented 👯 🔥

chaitanyadeorukhkar (approved):

anuraghazra (approved):

rfcs/2023-01-06-layout.md
divyanshu013: Done in https://github.com/razorpay/blade/pull/950/commits/a76d5aa03125cf9229ca3faa754711666374cc5f

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,450 @@
...
+This suffered from some related problems:
+
+1. Need for creating nested wrapper components for everything for any adjustments - `Space`, `Size` and `View`
+1. Readability wasn't as good because actual styles would get applied to the nested `View` which also made refactors harder
+1. Slightly higher learning curve
+1. No built-ins for responsive styles
+
+### Proposed solution
+
+With layout primitives in place, the above example could be implemented like:
```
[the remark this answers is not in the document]
divyanshu013: Done in https://github.com/razorpay/blade/pull/950/commits/a76d5aa03125cf9229ca3faa754711666374cc5f

anshulsahni (approved):
Looks very good

kkyusufk (approved):
Beautiful document, got ton of information on the layout philosophy in general!

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,460 @@
...
+| Flexbox             | `flex*`, `align*`, `justify*`, `order`      | [List of props](https://github.com/styled-system/styled-system/blob/master/docs/table.md#flexbox) referenced from styled-system.       |
+| Gap                 | `gap`, `rowGap`, `columnGap`                | Provides native flex functionality for spacing siblings.                                                                               |
+| Position            | `position`, `zIndex`, `top`, etc.           | [List of props](https://github.com/styled-system/styled-system/blob/master/docs/table.md#position) referenced from styled-system.      |
+| Grid **(web only)** | `grid*`                                     | [List of props](https://github.com/styled-system/styled-system/blob/master/docs/table.md#grid-layout)\* referenced from styled-system. |
+
+> **Note**
+>
+> Not all CSS grid properties are covered in `styled-system` spec. These should be referenced from MDN during implementation. Here's a nice reference from [CSS Tricks](https://css-tricks.com/snippets/css/complete-guide-grid).
```
nashcheez, line 177: The CSS grid properties are very extensive and for a reason `styled-system` supports only a few of them. We should limit the scope for the number of properties we should include otherwise the execution can be painstaking. I would vote for keeping it synchronous with what `styled-system` exposes.

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,460 @@
...
+
+**Few points to note for React (web)**
+
+- We should avoid supporting shorthands to save from confusions _(should I use `mb` or `marginBottom`)_.
+- Flex gap has ~91% global [coverage](https://caniuse.com/?search=flex%20gap) (support added in some browsers mid 2020). Although the support looks decent, this might be problematic for some products supporting these browsers. There is a [post CSS plugin](https://github.com/gavinmcfarland/flex-gap-polyfill) which can polyfill this _(with some known limitations)_. Alternative is to internally translate gap properties to utilize CSS selectors such as [lobotomized owl](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/) `* + *`.
+- Going forward with native `gap` supports seems like a fine choice here in favor of keeping things simple and native _(keeping it close to CSS `gap` today)_. Tradeoff being, older browsers will be able to render the layout with a degraded experience _(without spacing)_. If users wish to target older browsers they can opt for not using native flex gap and instead use margins (supported with layout styling props).
```
nashcheez, line 195: How do we plan to let developers know or raise awareness about this? A landing page if created using `gap` may have significant traffic coming from older browsers and those users will see a broken UI if not prevented.

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,460 @@
...
+| Flexbox             | `alignSelf`, `justifySelf`, `order`                                            | Only restricted set of properties applicable on flex children. Note: `justifySelf` doesn't work for flex layout but it can be used for block layout (web only).     |
+| Position            | `position`, `zIndex`, `top`, etc.                                              | [List of props](https://github.com/styled-system/styled-system/blob/master/docs/table.md#position) referenced from styled-system.                                   |
+| Grid **(web only)** | `gridColumn*`, `gridRow*`, `gridArea`, `justifySelf`, `alignSelf`, `placeSelf` | Only restricted set of properties applicable on [grid children](https://css-tricks.com/snippets/css/complete-guide-grid/#aa-properties-for-the-childrengrid-items). |
+| Layout              | None                                                                           |                                                                                                                                                                     |
```
nashcheez, line 271: Seeing the [spec](https://github.com/styled-system/styled-system/blob/master/docs/table.md#layout) here, why no `display` or `verticalAlign`? 
- A switch b/w a `display: none` and a `display: block` on an event handler is a valid use case for component styling
- If I don't have a parent component or a `Box` that's a flexbox a dev will have to rely on knowing the height of the container and apply manual marginTop to centrally align, `verticalAlign` may come in handy here.

nashcheez, line 271: I would have also asked for float but I understand why you chose to omit it.

time-touse-float-ijustbroke-the-wholesite-everytime-i-deal-with-44245421

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,460 @@
...
+The following shouldn't be considered right now:
+
+- Separate components for `Flex`, `Grid`, `Spacer`, same can be achieved with `Box`
+- Stack components (same can be achieved with `Flex`, `Grid`)
+- `sx` prop, same can be achieved with layout based styling props on components
+- `useSx` hook _(see [reference](#usesx-hook))_
+- `className`, `style` props support _(see [drawbacks and constraints](#drawbacksconstraints))_
+- Grid systems
+- Bleed layout components
+- Separate utility component for hiding (same can be achieved with responsive properties on `display` set to `none`, `flex`, `block`, etc.)
```
nashcheez, line 285: I am assuming this would be on the parent container cause I don't see it in the component's layout based styling prop?

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,460 @@
...
+- Minor caveat of object based responsive properties, users will pass object literals so it would re-render all the time, however not a concern for now. This is applicable in general to React.
+- A drawback of not exposing the helper components such as `Flex`, `Grid`, `Spacer` from our side is it can lead to deviating implementations. Some users may directly use `Box` component everywhere which will make the markup slightly harder to read (compared to having more readable `Flex`, `Grid`, etc.) in the JSX. However, as a constraint this ensures we don't have multiple components to do the same thing, affecting learning curve and usage.
+
+As a contrasting philosophy, Primer has an interesting take on supporting all properties _which means you can actually style a button to look something very different from a button_:
+
+> Assume that people will break the rules, provide safe ways for them to do so
```
nashcheez, line 309: Reminded me of this 😄 

css

nashcheez (approved):
This RFC is 🔥

rfcs/2023-01-06-layout.md
```diff
@@ -0,0 +1,460 @@
...
+  <Button marginTop="spacing.1">Button 2</Button>
+  <Button marginTop="spacing.2">Button 3</Button>
+</Box>
+```
+
+Styling props will be the preferred choice for minor adjustments that can be done directly on the component without a need for creating extra `Box` wrapper _(which also keeps the DOM less nested and easier to read)_. Later on, we might want to expand the scope of these props to cover things beyond layout if need arises.
+
+As a tradeoff, styling props are pretty dangerous in terms of the flexibility they provide. They give power to customize any component to beyond recognition, for example you could customize the padding that comes on `Button`. In light of this capability we should support only a restricted set of properties to begin with _(which can't have adverse effect on look and feel of components)_.
+
+Styling props should be the preferred choice for minor adjustments only. You may note `padding*` is skipped since it can break existing padding of blade components. Supported restricted properties:
```
sidag95, line 263: **question:** ‏Can we elaborate a little on how padding on the Layout component breaks blade components? From my understanding, blade components will not support the styling props and only the layout components will 🤔 
An example/explanation would help.

Without padding, any non-blade component will not be able to use the Layout component at all.

sidag95 (commented):
Overall, it looks great, quite similar to what docs has done so migrating should be easy 🙌🏽 

Just leaving one question about padding.

rfcs/2023-01-06-layout.md
anuraghazra: New proposal: A Clearfix component in blade. Let's remove all the flexbox stuff. 

```ts
<Clearfix />
```

rfcs/2023-01-06-layout.md
divyanshu013: I found some useful properties (eg. `grid-row-start`, `grid-column-start`, etc.) weren't documented in the linked styled-system doc so added this as reference for confirming during implementation.

Implementation wise we would be mostly just exposing these as props and passing them as CSS properties, so I think it shouldn't affect execution.

rfcs/2023-01-06-layout.md
divyanshu013: For the style props on components I think we should skip these (`display` property will be supported on `Box` component). Reasoning:
- Having the ability to modify `display` property on presentational components can break styling (folks can pass `display` as `block` where the component's outer container might be `flex` styled and vice-versa)
- So for conditional rendering of presentational components users may render it as `condition ? <Alert /> : null`
- `vertical-align` would work only for table cells or `inline-*`, I don't think we've lot of presentational components that would fit this usage. For the table use case users may pass `vertical-align` to their `td` element 🤔

rfcs/2023-01-06-layout.md
divyanshu013: Yup, possibly a separate utility component such as [`<Hidden />`](https://reshaped.so/content/docs/utilities/hidden)

rfcs/2023-01-06-layout.md
divyanshu013: Added minor clarification in https://github.com/razorpay/blade/pull/950/commits/7633c183161d7d23854719d1259fff80b7f27617

`padding*` props will be available on the main layout component `Box` but intentionally skipped on presentational components such as `Button`, `Alert`, etc. where they may break component styling.

So non-blade components should still be able to use `Box` to build UIs.

rfcs/2023-01-06-layout.md
divyanshu013: Added a suggestion in https://github.com/razorpay/blade/pull/950/commits/7633c183161d7d23854719d1259fff80b7f27617 to add some JSDoc comments to reinforce that we're relying on native flex gap. We can also double clarify this in our layout usage guidelines docs later.

rfcs/2023-01-06-layout.md
sidag95: Works! 🙌🏽

sidag95 (approved):
LGTM 🙌🏽

rfcs/2023-01-06-layout.md
saurabhdaware: Oh I didn't understand the display part. We are supporting it in Box right?

divyanshu013: Yup, we would definitely need `display` on `Box`. The tables have properties listed for both `Box` and other presentational components.

PR #952, opened 2023-01-13
docs: add List decisions

chaitanyadeorukhkar:
### [View Formatted Decisions](https://github.com/razorpay/blade/blob/docs/list-decisions/packages/blade/src/components/List/_decisions/decisions.md)


packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+<img src="./ordered-list-filled.png" width="380" />
+
+#### Ordered List with inline Link in items
+
+<img src="./ordered-list-link.png" width="380" />
+
+
+## API
+
+### List Component
```
saurabhdaware, line 61: Alternative: We can have 2 components, `UnorderedList` and `OrderedList`. `OrderedList` can have `variant="filled"`.
```ts
import { UnorderedList, OrderedList } from '@razorpay/blade/components'
```

This will also keep things similar to how they are in HTML.

saurabhdaware: Having it closer to HTML will also help in MDX (which is where I believe lists will be used heavily. We can directly pass our blade components to MDX overrides https://github.com/razorpay/docs/blob/062498c779dacdc8c3c661f043378204e8540237/docs-website/src/shared/components/layouts/mdx/ProductPageLayout.tsx#L49-L54 without having to create wrappers)

saurabhdaware: Although I see some cons would be:
- Can create confusion with 2 component names
- Might not be intuitive to search for "UnorderedList" from components.

saurabhdaware (commented):
Yay nice on point RFC!! Added comments

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+#### Unordered List with inline Link in items
+
+<img src="./unordered-list-link.png" width="380" />
+
+### Ordered List
+#### Ordered List with 3 levels of nesting
+
+<img src="./ordered-list.png" width="380" />
+
+#### Ordered List with 3 levels of nesting and _filled background_ numbers
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah, added the title to ensure we mention the use-case. The 3 level nesting is not added on Figma yet but it will be. I'll update the image once its added

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+### List Component
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| children | `List.Item` | `undefined` | List items to render. | ✅ |
+| variant | `ordered`, `unordered` | `unordered` | Sets the list as ordered or unordered. |  |
+| isOrderedFilled | `boolean` | `false` | Sets whether the numbers in ordered list have a filled background.  |  |
+| icon | `BladeIcon` | `undefined` | Icon to be used for the items in the list. This can only be set when `variant` is `unordered`. |  |
+| size | `small`, `medium` | `medium` | Size of the list items. |  |
+
+### List.Item Component
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Makes sense. Better to keep the consistency. I'll make this change

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| children | `string`, `List.Item.Link`, `List` | `undefined` | String to render in the list item. You may also pass a `List.Item.Link` to render an inline link. You may also pass a `List` component which contains its own `List.Item` components to create a nested list.| ✅ |
+| icon | `BladeIcon` | `undefined` | Icon to be used for the items in the list. This can only be set when `variant` is `unordered`. |  |
+
+### List.Item.Link Component
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: That's mainly to control the size of the Link so that it remains consistent with the rest of the font sizes of the list item (we offer 2 font sizes for list item)

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+**Reasons to not use this API:**
+- Not very intuitive
+- Hard to decipher the nesting just by looking at the code
+- Could make looping through data (nested objects) dynamically to create a list component more difficult in this case
+
+## Accessibility
+- I will explore using `<ol>`,`<ul>` & `<li>` directly and see if it allows us to do all the customizations that we need for our use-cases instead of using `role` on a `<div>` for web since that's the recommended practice for lists.
+  - We will be able to hide default styling & icons using `list-style-type: none; padding: 0px, margin: 0px`
+  - [POC CodeSandbox Link](https://codesandbox.io/s/festive-sanne-2rb18j?file=/src/App.js)
+- Set `role` as `list` for the `List` component
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Kept it as an alternative. We won't add this if actual HTML list elements work out. (which they mostly seem to be)

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+  - This does not work well with screen readers. It doesn't announce the item number, remaining number of items in the list and the current nesting level of the list.
+
+## Open Questions
+- ~Discuss alternative APIs with the team~
+- ~Would we allow different sizes within the same list?~ Nobody would use this but we also shouldn't add extra measures to restrict it
+- ~onHover support for list items?~ No
+- ~Do we allow ordered list within unordered list and vice versa?~ Yes, thats a valid use-case.
+- ~Do we allow inline Code component?~ Yes
+- What should we call `isOrderedFilled` instead?
+- Should we call it `List.Item.Link` or `List.Link` or `List.ItemLink`?
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Need control over the size. 
We would also need to further add Icon & Code component support. So we would have `ListItemCode`, `ListItemIcon` components as well. Lets discuss with the wider team if we want to do this or just let users pass Link, Code & Icon components and tweak sizes on their own

Ref: https://github.com/razorpay/blade/pull/952#discussion_r1073119822

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+
+
+## API
+
+### List Component
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| children | `List.Item` | `undefined` | List items to render. | ✅ |
+| variant | `ordered`, `unordered` | `unordered` | Sets the list as ordered or unordered. |  |
+| isOrderedFilled | `boolean` | `false` | Sets whether the numbers in ordered list have a filled background.  |  |
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: I'd prefer keeping 
`variant='ordered' | 'unordered'`
and adding `type='fillled' | 'default'` where `filled` is only available for `variant='ordered'`

Thoughts?

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+| rel | `string` | `undefined` | The rel for the link component.<br>-When `target` prop is set to `_blank`, `rel` will be automatically set to `noopener noreferrer` which can be overridden by passing the `rel` prop. | |
+| accessibilityLabel | `string` | `undefined` | The `aria-label` (web) & `accessibilityLabel` (native) of the link component. | |
+
+
+### Sample Usage
+```jsx
+import { List, InfoIcon, EditIcon } from '@razorpay/components';
+
+<List variant='ordered' icon={InfoIcon} size='medium'>
+  <List.Item icon={EditIcon}>
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Example seems to be misleading, I will change it. Icon can only be passed for `unordered` list. If its an `ordered` list then passing an icon will have no effect

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+| target | `string` | `undefined` | The target for the link component. | |
+| rel | `string` | `undefined` | The rel for the link component.<br>-When `target` prop is set to `_blank`, `rel` will be automatically set to `noopener noreferrer` which can be overridden by passing the `rel` prop. | |
+| accessibilityLabel | `string` | `undefined` | The `aria-label` (web) & `accessibilityLabel` (native) of the link component. | |
+
+
+### Sample Usage
+```jsx
+import { List, InfoIcon, EditIcon } from '@razorpay/components';
+
+<List variant='ordered' icon={InfoIcon} size='medium'>
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: No, ordered list cannot have icons. Only unordered list can. I will update this example

packages/blade/src/components/List/_decisions/decisions.md
chaitanyadeorukhkar: The major con is discoverability. A single List component is much easier to understand and discover which is why I kept it as a single component. I see most other design systems doing the same except a few. We can still take a call on this with opinions from others.

packages/blade/src/components/List/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Now that I'm seeing this pattern a lot where prop A can only contain X when prop B is Y. 

Will it make more sense to have this API: 

```
<List variant="ordered | ordered-filled | unordered" />
```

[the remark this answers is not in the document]
chaitanyadeorukhkar: Hmm yeah, that would simplify things too. Better than having `type`

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,305 @@
...
+| icon | `BladeIcon` | `undefined` | Icon to be used for the items in the list. This can only be set when `variant` is `unordered`. |  |
+| size | `small`, `medium` | `medium` | Size of the list items. |  |
+
+### ListItem Component
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| children | `string`, `ListItemLink`, `List` | `undefined` | String to render in the list item. You may also pass a `ListItemLink` to render an inline link. You may also pass a `List` component which contains its own `ListItem` components to create a nested list.| ✅ |
+| icon | `BladeIcon` | `undefined` | Icon to be used for the items in the list. This can only be set when `variant` is `unordered`. |  |
+
+### ListItemLink Component
```
anuraghazra, line 78: Is it necessary to create a ListItemLink instead of using Link component directly? Are we restricting anything specific to List component?

anuraghazra (approved):
Mostly LGTM

packages/blade/src/components/List/_decisions/decisions.md
saurabhdaware: Makes sense. I like the approach Anurag suggested in above comment. Can go with that `order-filled` variant instead and avoid these components.

packages/blade/src/components/List/_decisions/decisions.md
saurabhdaware, line 78: I think this we can explore if we can inherit parent size in all typography and link components right?

saurabhdaware (commented):
Looks good to me. Just one comment - https://github.com/razorpay/blade/pull/952/files#r1082201175

packages/blade/src/components/List/_decisions/decisions.md
anuraghazra: We should be able to do this with Context, we are doing the same on CheckboxGroup/RadioGroup too.

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,305 @@
...
+<img src="./ordered-list-link.png" width="380" />
+
+
+## API
+
+### List Component
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| children | `ListItem` | `undefined` | List items to render. | ✅ |
+| variant | `ordered`, `ordered-filled`, `unordered` | `unordered` | Sets the list as ordered or unordered. |  |
```
divyanshu013, line 68: In nested ordered list, how would the user specify the indexing pattern (similar to the screenshot above)?
- 1, 2, 3
- a, b, c
- i, ii, iii

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,305 @@
...
+| size | `small`, `medium` | `medium` | Size of the list items. |  |
+
+### ListItem Component
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| children | `string`, `ListItemLink`, `List` | `undefined` | String to render in the list item. You may also pass a `ListItemLink` to render an inline link. You may also pass a `List` component which contains its own `ListItem` components to create a nested list.| ✅ |
+| icon | `BladeIcon` | `undefined` | Icon to be used for the items in the list. This can only be set when `variant` is `unordered`. |  |
+
+### ListItemLink Component
+| Prop | Type | Default | Description | Required |
```
divyanshu013, line 79: Are the props same as `Link`? We can probably remove the table and just say the props are inherited from `Link`

divyanshu013 (approved):
:shipit:

packages/blade/src/components/List/_decisions/decisions.md
chaitanyadeorukhkar: Doing it with context. Inheriting parent sizes is a bit tricky on react native in our set up

packages/blade/src/components/List/_decisions/decisions.md
chaitanyadeorukhkar: Level 1 will be 1,2,3
Level 2 will be a,b,c
Level 3 will be i,ii,iii

packages/blade/src/components/List/_decisions/decisions.md
chaitanyadeorukhkar: Yeah the props are same as Link, just the size prop is not available.

packages/blade/src/components/List/_decisions/decisions.md
divyanshu013: Maybe we can specify that here

chaitanyadeorukhkar: Done

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,311 @@
...
+- ~Would we allow different sizes within the same list?~ Nobody would use this but we also shouldn't add extra measures to restrict it
+- ~onHover support for list items?~ No
+- ~Do we allow ordered list within unordered list and vice versa?~ Yes, thats a valid use-case.
+- ~Do we allow inline Code component?~ Yes
+- ~Should we call it `ListItemLink` or `List.Link` or `ListItemLink`?~ We will call it `ListItemLink` to maintain consistency across other components where we made similar decisions and aid tree-shaking via this approach
+- ~Bold ordered numbers?~ No bold numbers
+- ~H1, H2, etc. for list items?~ No H1, H2 support from design
+- ~Use-cases for List of level 1 with item that also contains list of level 1?~ No such support. The discovered use-cases in our product were incorrect usage.
+- Ordered list with custom numbers?
+- What should we call `isOrderedFilled` instead?
```
kamleshchandnani, line 301: update: this is answered

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,311 @@
...
+- Supporting images, GIFs, videos, custom components, bold text, non-bulleted text, and icons within list items.
```
kamleshchandnani, line 302: update: this is answered

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,311 @@
...
+  <ListItem/>
+</List>
+```
+
+**Reasons to not use this API:**
+- Not very intuitive
+- Hard to decipher the nesting just by looking at the code
+- Could make looping through data (nested objects) dynamically to create a list component more difficult in this case
+
+## Accessibility
```
kamleshchandnani, line 281: what is the final approach on a11y?

kamleshchandnani (commented):
hardly any comment on the API spec. nicely written 👌

packages/blade/src/components/List/_decisions/decisions.md
chaitanyadeorukhkar: Updated

packages/blade/src/components/List/_decisions/decisions.md
chaitanyadeorukhkar: Yup we're restricting `size`

packages/blade/src/components/List/_decisions/decisions.md
```diff
@@ -0,0 +1,312 @@
...
+| icon | `BladeIcon` | `undefined` | Icon to be used for the items in the list. This can only be set when `variant` is `unordered`. |  |
+| size | `small`, `medium` | `medium` | Size of the list items. |  |
+
+### ListItem Component
+| Prop | Type | Default | Description | Required |
+|---|---|---|---|---|
+| children | `string`, `ListItemLink`, `List` | `undefined` | String to render in the list item. You may also pass a `ListItemLink` to render an inline link. You may also pass a `List` component which contains its own `ListItem` components to create a nested list.| ✅ |
+| icon | `BladeIcon` | `undefined` | Icon to be used for the items in the list. This can only be set when `variant` is `unordered`. |  |
+
+### ListItemLink Component
```
sidag95, line 84: **suggestion:** ‏From docs perspective, there are a few other use cases that we need to support
1. ListItem with inline link and text
2. ListItem with text and an image
3. ListItem with arbitrary components (we have a callout component that is used sometimes)

[Reference](https://razorpay.com/docs/payments/dashboard/support#registered-razorpay-users)

To support such outlier use cases (I would assume non-documentation products don't have such use-cases), should we still allow the consumer to provide custom children (using some override key if required) and let them handle the consistency of size?

The override key would deter the majority of projects from providing custom implementations while be flexible enough for niche use cases like docs. 

Thoughts?

sidag95 (commented):
Majority of the RFC looks great, the specified API is 👌🏽 , left an open question around children of ListItem

packages/blade/src/components/List/_decisions/decisions.md
chaitanyadeorukhkar: > ListItem with inline link and text
That is already supported in the current API. We will expose a component `ListItemLink` which can be used inline with text

> ListItem with text and an image
> ListItem with arbitrary components (we have a callout component that is used sometimes)
We have scoped this out of List component for now until design adds support for it. @saurav12 mentioned that we don't have enough use-cases for this and docs design will also be relooking at how List is to be used. @saurav12 can add more to this

`override` key might be a deviation from our API design. If we do add support for custom children then we should allow it for every project as a valid use-case imo. 

I'll discuss the support for text, image and arbitrary components again with @saurav12 to make sure we're aligned on the scope of the current List.

sidag95: > ListItemLink which can be used inline with text

For this, from the examples I can see, the entire list item would be a link. However, there is a use case in docs where the list item has some link and text.

sidag95: >override key might be a deviation from our API design

Makes sense. However, can we consider directly allowing any children and letting the users handle the styling deviations? This is similar to how any arbitrary value is allowed in spacing for the layout component. If a consumer wants to override the default behaviour then it should be their responsibility to adhere to the design spec.

chaitanyadeorukhkar: > For this, from the examples I can see, the entire list item would be a link. However, there is a use case in docs where the list item has some link and text.

We would support inline text + link. I'll add an example on the doc as well, that's a miss

Eg)
```jsx
<ListItem>
    For more information <ListItemLink href=''>Click here</ListItemLink>
</ListItem>
```

PR #953, opened 2023-01-14
Update README.md

utkarsh006:
- View Changes [HERE](https://github.com/utkarsh006/blade/blob/Readme_Updation/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 <h1 align="center">
   Blade
 </h1>
+
 <p align="center">
   <a href="https://github.com/styled-components/styled-components">
     <img src="https://img.shields.io/badge/style-%F0%9F%92%85%20styled--components-orange.svg?colorB=daa357&colorA=db748e" alt="Blade is styled with styled-components" />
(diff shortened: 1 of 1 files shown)


utkarsh006:
@kamleshchandnani Kindly Review.

cseas:
What value do these changes add? The only change added apart from formatting is a badge for Repo Size which we should avoid showing as it's a pointless (possibly misleading) stat for consumers of the npm package.

utkarsh006:
@cseas Nothing is pointless. I have also formatted some text.

README.md
```diff
@@ -14,25 +15,24 @@
 </p>
 
 <p align="center">
-   The Design System that powers Razorpay
+   <b>Design System that powers Razorpay⚡  </b>
 <p align="center">
 
 ### [`@razorpay/blade`](https://github.com/razorpay/blade/tree/master/packages/blade)
+![Github Repo Size](https://img.shields.io/github/repo-size/razorpay/blade?style=for-the-badge&color=aqua)
```
kamleshchandnani, line 22: While this might be great for some repos, in Blade's case, I don't see any advantage that we will get, and it might mislead people. Hence would request if you could remove this change.

README.md
```diff
@@ -14,25 +15,24 @@
...
 
----
 
-> **Note**
->
-> The blade-old package has been moved to it's own repo.
-> [github.com/razorpay/blade-old](https://github.com/razorpay/blade-old) (Only accessible to Razorpay employees)
->
-> It is under **maintenance** and it won't have any major releases.
+## Note
```
kamleshchandnani, line 32: this was on purpose because GitHub markdown highlights this content as below and we want this as information rather than a heading in its own. Can you revert this?
Screenshot 2023-01-16 at 10 45 25 AM

utkarsh006:
@kamleshchandnani I hope its better now

kamleshchandnani:
@utkarsh006 Thank you for your contribution 🎉 . While I appreciate the efforts you put in to suggest some changes. I'm afraid to say that the changes are something we would not be able to accept and the reasons are mentioned as comments for each change above. 

Having said that, this is not the end of your contributions to Blade. We will look forward to your contributions in the future that would add value and aligns with what we expect from Blade. 🙂

PR #940, opened 2023-01-04
chore(blade): added all icons from figma

anuraghazra:
**In this PR:** 

- Added all icons from figma library via a custom script (the script is not added in the repo)

**TODOs:**

1. Once manually verify few components. 

This script adds the icons by the name that they are defined in Figma, but in code side previously we already have added few icons with different name. 

For example: `RotateCounterClockWiseIcon` icon was already added manually but this script added `RotateCcwIcon` 

----

2. Verify if we can simplify and remove few extraneous svg elements from icons

For example, in `WifiOffIcon` there are `Defs`, `ClipPath` etc 

image

PR #964, opened 2023-01-17
feat(blade): add new icons

anuraghazra:
Added new icons

image


packages/blade/src/components/Icons/index.tsx
```diff
@@ -87,7 +87,7 @@ export { default as ExternalLinkIcon } from './ExternalLinkIcon';
 export { default as MailIcon } from './MailIcon';
 export { default as AlertCircleIcon } from './AlertCircleIcon';
 export { default as AlertOnlyIcon } from './AlertOnlyIcon';
-export { default as ScanIcon } from './ScanIcon';
+export { default as QRCodeIcon } from './QRCodeIcon';
```
divyanshu013, line 90: Would be a breaking change?

packages/blade/src/components/Icons/index.tsx
```diff
@@ -305,3 +305,12 @@ export { default as ShuffleIcon } from './ShuffleIcon';
...
 export { default as TransactionsIcon } from './TransactionsIcon';
+export { default as AnnouncementIcon } from './AnnouncementIcon';
+export { default as AppStoreIcon } from './AppStoreIcon';
+export { default as CoinsIcon } from './CoinsIcon';
+export { default as BillIcon } from './BillIcon';
+export { default as StampIcon } from './StampIcon';
+export { default as MenuDotsIcon } from './MenuDotsIcon';
+export { default as SendIcon } from './SendIcon';
+export { default as MailOpenIcon } from './MailOpenIcon';
+// # append_icon_export
```
divyanshu013, line 316: Stray comments

packages/blade/src/components/Icons/index.tsx
anuraghazra: Not a breaking change since ScanIcon was not present on any of the releases before. It was added only in https://github.com/razorpay/blade/pull/940/files#diff-ee401bacbc5df34a732325643f7e734ace96c88126b4f74ecc3b2200a81966daR1 (which isn't released)

anuraghazra: Plus this will be clubbed together with the v6 release anyways.

packages/blade/src/components/Icons/index.tsx
anuraghazra: fixed.

kamleshchandnani (approved):

PR #966, opened 2023-01-19
feat(ActionList): add ActionList styling

saurabhdaware:
**Stack** 
- [x] #932 
- [x] #948 
- [x] #966  (This PR)
- [x] #981 

**Changes** 

- ActionList styling
- Refactors
- Dropdown Animations
- React Native Support

Preview: https://61c19ee8d3d282003ac1d81c-ftuhxfqnrm.chromatic.com/?path=/story/components-dropdown--with-select&globals=measureEnabled:false

Usage: https://github.com/razorpay/blade/blob/saurabh/feat/actionlist/packages/blade/src/components/Dropdown/Dropdown.stories.tsx

**Screenshots**

image

image image

**TODO**
- [x] Dropdown Tests (Will create separate PR for tests, documentation and minor cleanups)
- [x] Documentation


packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,115 +1,136 @@
...
-const ActionListItem = (props: ActionListItemProps): JSX.Element => {
+
+const StyledActionList = styled(Box)<{ surfaceLevel: ActionListProps['surfaceLevel'] }>(
+  ({ theme, surfaceLevel = 2 }) => {
+    const offsetX = theme.shadows.offsetX.level[1];
+    const offsetY = theme.shadows.offsetY.level[2];
+    const blur = theme.shadows.blurRadius.level[2];
+    const shadowColor = theme.shadows.color.level[1];
+
+    const elevation200 = `${makeSize(offsetX)} ${makeSize(offsetY)} ${makeSize(
```
anuraghazra, line 27: In android we need additional logic to make them look similar to web, otherwise the shadows will look vastly different, check the CardSurface.native.tsx file to see how I did it.

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,115 +1,136 @@
...
     }
-
     return child;
   });
 
   React.useEffect(() => {
     setOptions(actionListOptions);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [setOptions, children]);
+  }, [optionsRecalculateToggle]);
```
anuraghazra, line 112: What does this `optionsRecalculateToggle` do?

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }, [setHasFooterAction, props.trailing]);
+
+  const isOnlyActionButton = !props.title && !props.leading && props.trailing;
+
+  return (
+    <StyledActionListFooter
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      ref={footerRef as any}
+      // eslint-disable-next-line
```
anuraghazra, line 51: Let's disable this eslint rule for the file. 

Plus it might just be better to define the type on the StyledActionListFooter itself, even if we are not using it on native. 

```ts
const StyledActionListFooter = styled(Box)<{ onMouseDown: ..., onKeyDown: ..., onBlur: ... }>((props) => {
```

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,108 @@
...
+        if ((nativeEvent.key === ' ' || nativeEvent.key === 'Enter') && activeIndex < 0) {
+          return;
+        }
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        onSelectKeydown?.({ event: e.nativeEvent } as any);
+      }}
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error
+      // @ts-ignore: Ignoring because the TS fails for React Native and works for web
+      onBlur={(e) => {
+        const nextItem = e.nativeEvent.relatedTarget;
```
anuraghazra, line 70: Why use `e.nativeEvent.relatedTarget` instead of `e.relatedTarget` ?? That will solve the `getAttribute` not present issue too.

packages/blade/src/components/ActionList/ActionListHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+    )}`,
+  };
+});
+
+const ActionListHeader = (props: { title: string; leading?: React.ReactNode }): JSX.Element => {
+  return (
+    <StyledActionListHeader>
+      <Box>{props.leading}</Box>
+      <Box paddingLeft="spacing.3" paddingRight="spacing.3">
+        <BaseText color="surface.text.subdued.lowContrast" fontStyle="italic" fontSize={50}>
```
anuraghazra, line 27: Possible to use the Text component instead of BaseText?

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,235 @@
...
+import { isReactNative, makeAccessible, makeSize } from '~utils';
+import { BaseText } from '~components/Typography/BaseText';
+import { Checkbox } from '~components/Checkbox';
+
+const ActionListItemContext = React.createContext<{
+  intent?: ActionListItemProps['intent'];
+}>({});
+
+const SectionDivider = styled(Box)((props) => ({
+  height: makeSize(1),
```
anuraghazra, line 19: Let's add a TODO comment here, if we even add a 1px token we can replace this.

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,235 @@
...
+        {...makeAccessible({
+          selected: isSelected,
+          role: getActionListItemRole(props.href),
+        })}
+        {...makeActionListItemClickable((e: React.MouseEvent<HTMLButtonElement>): void => {
+          if (typeof props.index === 'number') {
+            onOptionClick(e, props.index);
+          }
+          props.onClick?.(e);
+        })}
```
anuraghazra, line 160: Is the order matters here? 

If not, can we do this interpolated props at the end of all other props? (bit cleaner to look at)

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,235 @@
...
+        id={`${dropdownBaseId}-${props.index}`}
+        tabIndex={-1}
+        href={props.href}
+        className={activeIndex === props.index ? 'active-focus' : ''}
+        {...makeAccessible({
+          selected: isSelected,
+          role: getActionListItemRole(props.href),
+        })}
+        {...makeActionListItemClickable((e: React.MouseEvent<HTMLButtonElement>): void => {
+          if (typeof props.index === 'number') {
```
anuraghazra, line 156: isNumber() ?

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,235 @@
...
+      >
+        <Box display="flex" marginTop={props.description ? 'spacing.2' : undefined}>
+          {selectionType === 'multiple' ? (
+            // Adding aria-hidden because the listbox item in multiselect in itself explains the behaviour so announcing checkbox is unneccesary and just a nice UI tweak for us
+            <Checkbox
+              isChecked={isSelected}
+              tabIndex={-1}
+              {...makeAccessible({
+                hidden: true,
+              })}
```
anuraghazra, line 186: Pretty sure this makeAccessible does nothing since Checkbox doesn't accept arbitrary props.

packages/blade/src/components/ActionList/StyledActionListItem.web.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return {
+    ...getBaseActionListItemStyles(props),
+    // Web-specific styles
+    '&:hover': {
+      backgroundColor:
+        props.intent === 'negative'
+          ? props.theme.colors.feedback.background.negative.lowContrast
+          : props.theme.colors.brand.gray.a50.lowContrast,
+    },
+    '&.active-focus': {
```
anuraghazra, line 16: Where do get or set this .active-focus class?

packages/blade/src/components/ActionList/getBaseActionListItemStyles.ts
```diff
@@ -0,0 +1,37 @@
...
+  intent?: ActionListItemProps['intent'];
+  onPress?: (e: React.TouchEvent<TouchableOpacity>) => void;
+  isSelected?: boolean;
+};
+
+const getBaseActionListItemStyles = (
+  props: StyledActionListItemProps & { theme: Theme },
+): CSSObject => {
+  return {
+    borderWidth: makeSize(3),
```
anuraghazra, line 21: Add todo for size token.

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -0,0 +1,94 @@
...
+        position="absolute"
+        transition={isOpen ? fadeIn : fadeOut}
+        onAnimationEnd={() => {
+          if (isOpen) {
+            setDisplay('block');
+          }
+
+          if (!isOpen) {
+            setDisplay('none');
+          }
```
anuraghazra, line 84:
```ts
          if (isOpen) {
            setDisplay('block');
          } else {
            setDisplay('none');
          }
          
```

or

```
setDisplay(isOpen ? 'block' : 'none');
```

😅

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,115 +1,136 @@
...
-    selectInputRef,
+    setSelectedIndices,
+    optionsRecalculateToggle,
   } = useDropdown();
+  const actionListOptions: OptionsType = [];
 
-  const platformType = getPlatformType();
-  const renderOnWebAs = props.href ? 'a' : 'button';
-  const isReactNative = platformType === 'react-native';
+  const defaultSelectedIndices: number[] = [];
```
anuraghazra, line 54: Do we not need to keep these two values in React state?

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
saurabhdaware: lol yes will do. Was probably 🍻 while writing this

packages/blade/src/components/ActionList/ActionList.tsx
saurabhdaware: It is a toggle flag which I change to re-run this useEffect. So if there are any errors in loading options, it just retriggers this function.

packages/blade/src/components/ActionList/ActionList.tsx
saurabhdaware: done in upcoming commit

packages/blade/src/components/ActionList/ActionListHeader.tsx
saurabhdaware: Ohhh rightt! I forgot about caption variant. I was trying to find italic text there 🙈 changed it in upcoming commit

packages/blade/src/components/ActionList/ActionListFooter.tsx
saurabhdaware: oh yes right!! changed it in upcoming commit

packages/blade/src/components/ActionList/ActionListFooter.tsx
saurabhdaware: Disabled the ban-ts rule for this file. I tried adding types on footer but that too was failing with some big overload errors. Didn't spend much time on it.

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: done in upcoming commit

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: I have kept the custom properties on styled component at the end and native HTML attributes at the start. So that its easier to debug for custom properties.

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: You mean `Number(props.index)`? `isNumber` doesn't exist natively (comes from lodash). There is also `Number.isInteger`

Anyway `typeof props.index` seems a bit easier to understand to me. Not introducing any new methods.

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: ops lol yes. I added a Box around it and added hidden on that. It works now. Strangely the voiceover didn't announce it even without this. probably because the focus was disabled.

packages/blade/src/components/ActionList/StyledActionListItem.web.tsx
saurabhdaware: https://github.com/razorpay/blade/blob/736d33458bdd9a703651d18df2c7b73a61089b68/packages/blade/src/components/ActionList/ActionListItem.tsx#L155

Here

packages/blade/src/components/ActionList/ActionList.tsx
saurabhdaware: We have `selectedIndiced` and `options` state on top level in the context. We just add these local variables to that context so don't need to maintain separate state for these in particular

kamleshchandnani:
When the delete is selected the active state should be red color or blue? right now its blue but hover is red which seems right

Screenshot 2023-01-25 at 11 59 32 AM

saurabhdaware:
@kamleshchandnani discussing that with Abhishek here - https://razorpay.slack.com/archives/G01B3LQ9H0W/p1674627922011559?thread_ts=1674536439.113129&cid=G01B3LQ9H0W

packages/blade/src/components/ActionList/ActionListItem.tsx
anuraghazra: Okay not a big deal.

packages/blade/src/components/ActionList/ActionListItem.tsx
anuraghazra: Yes should be fine anyways since tabIndex={-1} is set

packages/blade/src/components/ActionList/ActionList.tsx
anuraghazra: Should be fine, just I hope there will be no stale values :p

packages/blade/src/components/ActionList/ActionList.tsx
anuraghazra: >  So if there are any errors in loading options, it just retriggers this function.

What kind of errors?

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
saurabhdaware: done

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+      {children}
+    </Text>
+  );
+};
+
+const ActionListCheckboxWrapper = styled(Box)((_props) => ({
+  pointerEvents: 'none',
+}));
+
+const getActionListItemRole = (href?: string): 'link' | 'menuitem' | 'option' => {
```
kamleshchandnani, line 79: the way we have `accessibilityMap` shall we also create a11y roles so we can put platform dependent things at one place and then we don't need to put checks here and it can also act as constants?

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+/**
+ *
+ * ActionListItem
+ *
+ *
+ */
+type ActionListItemProps = {
+  title: string;
+  description?: string;
+  onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void;
```
kamleshchandnani, line 100: is passing the entire event necessary? or just few args like name and value as we do in input fields will suffice?

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: I feel it might not be consistent. Just because here we're using `menuitem` instead of `option`, does not mean every `option` role can be replaced with `menuitem`. It might also be a `listitem` in some scenarios. So I think we should handle this on case-by-case.

packages/blade/src/components/ActionList/StyledActionListItem.web.tsx
```diff
@@ -0,0 +1,28 @@
+import styled from 'styled-components';
+import type { StyledActionListItemProps } from './getBaseActionListItemStyles';
+import { getBaseActionListItemStyles } from './getBaseActionListItemStyles';
+import Box from '~components/Box';
+
+const StyledActionListItem = styled(Box)<StyledActionListItemProps>((props) => {
+  return {
+    ...getBaseActionListItemStyles(props),
```
kamleshchandnani, line 8: so since we already have a cross platform Box component can't we just have platform dependent styles instead of a styled component all together? so your `StyledActionListItem.web.tsx` can become `getActionListItemStyles.web.ts` and `getActionListItemStyles.native.ts` and you can put the `styled(Box)` directly under `ActionListItem.tsx` file?

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -40,8 +37,14 @@ function Dropdown({ children, selectionType = 'single' }: DropdownProps): JSX.El
   const [shouldIgnoreBlur, setShouldIgnoreBlur] = React.useState(false);
   const selectInputRef = React.useRef<HTMLButtonElement>(null);
   const actionListRef = React.useRef<HTMLDivElement>(null);
-  const dropdownBaseId = useId('dropdown');
+  const [hasFooterAction, setHasFooterAction] = React.useState(false);
+  const [optionsRecalculateToggle, setOptionsRecalculateToggle] = React.useState(false);
```
kamleshchandnani, line 41: didn't understand what is this used for?

packages/blade/src/components/Dropdown/DropdownOverlay.native.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import styled from 'styled-components/native';
+import { componentIds } from './dropdownUtils';
+import { useDropdown } from './useDropdown';
+import Box from '~components/Box';
+import { makeSize } from '~utils';
+
+const StyledDropdownOverlay = styled(Box)((props) => ({
+  transform: `translateY(${makeSize(props.theme.spacing[3])})`,
```
kamleshchandnani, line 9: does this works on ios and android both since its static?

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { useTheme } from '~components/BladeProvider';
+
+const dropdownFadeIn = keyframes`
+from {
+  transform: translateY(0px);
+  opacity: 0;
+}
+
+to {
+  transform: translateY(8px);
```
kamleshchandnani, line 17: token?

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -0,0 +1,95 @@
...
+  opacity: 1;
```
kamleshchandnani, line 18: token?

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (isOpen) {
+      setDisplay('block');
+    }
+  }, [isOpen]);
+
+  return (
+    <Box position="relative">
+      <StyledDropdownOverlay
+        style={{ opacity: isOpen ? 1 : 0 }}
+        display={display}
```
kamleshchandnani, line 75: shouldn't display be also part of your CSS animation just like opacity? so you save on one re-render and also flicker since there are too many setstates and most also affecting the animations which might make it more flickery the more complex this component becomes?

packages/blade/src/components/Dropdown/useDropdown.ts
```diff
@@ -177,6 +190,12 @@ const useDropdown = (): UseDropdownReturnValue => {
    * Blur handler on combobox. Also handles the selection logic when user moves focus
    */
   const onSelectBlur = (): void => {
+    if (rest.hasFooterAction) {
+      // When Footer has action buttons, we ignore the blur (by setting shouldIgnoreBlur to true in onSelectKeyDown)
```
kamleshchandnani, line 194: ignore blur on dropdown or select?

packages/blade/src/components/Dropdown/useDropdown.ts
```diff
@@ -258,15 +279,32 @@ const useDropdown = (): UseDropdownReturnValue => {
...
         selectCurrentOption: () => {
           selectOption(activeIndex);
+          if (rest.hasFooterAction) {
+            rest.selectInputRef.current?.focus();
+          }
+
+          const anchorLink = options[activeIndex]?.href;
+          if (anchorLink) {
+            window.location.href = anchorLink;
+            if (window.top) {
```
kamleshchandnani, line 304: why do we do this as well in addition to location.href?

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
```diff
@@ -62,15 +62,16 @@ const StyledInput = styled.input<HoverProps>(({ theme, isChecked, isDisabled, ha
 
 const _SelectorInput: React.ForwardRefRenderFunction<
   BladeElementRef,
-  HoverProps & { inputProps: any }
-> = ({ inputProps, isChecked, isDisabled, hasError }, ref) => {
+  HoverProps & { inputProps: any; tabIndex?: number }
```
kamleshchandnani, line 65: since this is optional and if not set then it'll be undefined means it'll set to 0 which is the default behavior I'm assuming?

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -103,6 +111,8 @@ const SelectInput = (props: SelectInputProps): JSX.Element => {
   );
 };
 
+const SelectInput = React.forwardRef(_SelectInput);
+// @ts-expect-error: componentId is our custom attribute
 SelectInput.componentId = 'SelectInput';
```
kamleshchandnani, line 116: oh you can actually use `WithComponentId` from utils

packages/blade/src/utils/platform/isAndroid.native.ts
```diff
@@ -0,0 +1,3 @@
+import { Platform } from 'react-native';
+
+export const isAndroid = (): boolean => Platform.OS === 'android';
```
kamleshchandnani, line 3: can probably do it like below?
```
getNativePlatform() {
  if(Platform.OS === 'android') {
    return 'android'
  }
  if(Platform.OS === 'iOS') {
    return 'iOS';
  }
}
```

packages/blade/src/components/ActionList/ActionList.tsx
saurabhdaware: yes. This is mostly try catch equivalent. I haven't found any case where this happens.

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,119 +1,149 @@
...
-      as={!isReactNative ? renderOnWebAs : undefined}
-      id={`${dropdownBaseId}-${props._index}`}
-      role="option"
-      data-value={props.value}
-      data-index={props._index}
-      aria-selected={
-        typeof props._index === 'number' ? selectedIndices.includes(props._index) : undefined
+  const defaultSelectedIndices: number[] = [];
+
+  const getActionListItemChildWithId = (child: React.ReactNode): React.ReactNode => {
```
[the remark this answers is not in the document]
saurabhdaware: Changed it to `getActionListItemWithId` in upcoming commit

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,119 +1,149 @@
...
-  }[] = [];
+      const clonedChild = React.cloneElement(child, {
+        // @ts-expect-error: TS doesn't understand the child's props
+        _index: currentIndex,
+      });
+      return clonedChild;
+    }
 
-  console.count('ActionList');
+    return child;
```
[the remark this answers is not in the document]
saurabhdaware: Assuming React itself will throw error. The isValid check is mostly to make typescript happy. Returning child seems more of a default behaviour no? 🤔

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,119 +1,149 @@
...
+    <StyledActionList
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       ref={actionListRef as any}
       {...makeAccessible({
-        role: 'listbox',
+        role: isReactNative() ? 'menu' : 'listbox',
         multiSelectable: selectionType === 'multiple',
         labelledBy: `${dropdownBaseId}-label`,
       })}
       id={`${dropdownBaseId}-listbox`}
```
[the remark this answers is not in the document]
saurabhdaware: I moved the a11y things into separate `getA11yRoles.ts` file for now so that we can control roles from outside of all this code.

By generic if you mean generic for menus, then it is right now generic for menus. 

But by generic if you mean, usable as a plain list somewhere then the accessibility completely changes. 90% of the code in ActionList won't be used in that scenario. We can go several ways there
- Have a separate Navigation component (this would be ideal for bundle size, and managing code accessibility and behaviour for us)
- Add if else checks to not run some of the code (click events, selection calls, keydown events, the runtime logic to add index to items, etc)

Lets discuss this after standup today.

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { isReactNative, makeSize } from '~utils';
+import { Text } from '~components/Typography';
+
+type ActionListFooterProps = {
+  title?: string;
+  leading?: React.ReactNode;
+  trailing?: React.ReactNode;
+};
+
+const StyledActionListFooter = styled(Box)((props) => {
```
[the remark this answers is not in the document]
saurabhdaware: So I am using styled-components in parts where there are too many event handlers or other props in the component. Also the padding and margin shorthands are not supported on Box right now so it will increase the props.

So with this the 'styling' is separate from the 'logic'. Don't have strong opinion so either will work.

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,106 @@
...
+  React.useEffect(() => {
+    if (!isReactNative() && footerRef.current?.querySelector('button, a')) {
+      setHasFooterAction(true);
+    }
+  }, [setHasFooterAction, props.trailing]);
+
+  const isOnlyActionButton = !props.title && !props.leading && props.trailing;
+
+  return (
+    <StyledActionListFooter
```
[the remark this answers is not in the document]
saurabhdaware: yes. that focus on select and apply button represents actual focus whereas actionlist items are virtual focus. 🤔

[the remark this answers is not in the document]
saurabhdaware: https://github.com/razorpay/blade/pull/966#discussion_r1090143637

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,119 +1,149 @@
...
     return child;
   });
 
   React.useEffect(() => {
     setOptions(actionListOptions);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [setOptions, children]);
+  }, [optionsRecalculateToggle]);
+
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
saurabhdaware: oh I am assuming these things to work after hydration only.

Even with isomorphic hook, it won't run on server. We use `useEffect` on server and `useLayoutEffect` on browser which might speed things up a bit. But in React docs they mostly suggested to use `useEffect` wherever possible. And the speed at which the options and selections load isn't that big of a deal right now. Its helpful when you want to load something quick on first fold of the page but in our case, we're only loading the options in the state.

Overall its not possible to run it on server because we're setting it into the state.

image

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,119 +1,149 @@
...
+        _index: currentIndex,
+      });
+      return clonedChild;
+    }
 
-  console.count('ActionList');
+    return child;
+  };
+
+  const childCompIdArray = React.Children.toArray(children).map((child) => getComponentId(child));
```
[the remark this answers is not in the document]
saurabhdaware: oh yes right. missed it 😢

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,119 +1,149 @@
...
+  }, []);
 
   return (
-    <Box
+    <StyledActionList
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       ref={actionListRef as any}
       {...makeAccessible({
-        role: 'listbox',
+        role: isReactNative() ? 'menu' : 'listbox',
```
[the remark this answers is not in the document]
saurabhdaware: so I had tried that earlier but screenreader announces role="dialog" explicitly as dialog which I though would throw users away since dialogs are mostly modals.

So now, I am conditionally checking if footer has buttons. If it has buttons, then I switch to `dialog` role but otherwise I stay on `listbox` only. https://github.com/razorpay/blade/pull/966/commits/f1560106cb47c0d7576f345327f3d38174f7717a

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,106 @@
...
+      ref={footerRef as any}
+      // @ts-ignore: Ignoring because the TS fails for React Native and works for web
+      onMouseDown={() => {
+        setShouldIgnoreBlur(true);
+      }}
+      // @ts-ignore: Ignoring because the TS fails for React Native and works for web
+      onKeyDown={(e) => {
+        const nativeEvent = e.nativeEvent;
+        // We ignore the selection keydowns on footer to let users click on items
+        if ((nativeEvent.key === ' ' || nativeEvent.key === 'Enter') && activeIndex < 0) {
```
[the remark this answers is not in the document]
saurabhdaware: yes. Should the list close in that scenario? I was thinking it shouldn't close so I added this. Lemme start a thread on slack we can discuss this with designer.

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,119 +1,149 @@
...
+  const childCompIdArray = React.Children.toArray(children).map((child) => getComponentId(child));
+  const lastActionListSectionIndex = childCompIdArray.lastIndexOf(componentIds.ActionListSection);
+
+  const isActionListItemPresentAfterSection = childCompIdArray
+    .slice(lastActionListSectionIndex)
+    .includes(componentIds.ActionListItem);
 
   // Looping through ActionListItems to add index to them and get an options array for moving focus between items
-  const childrenWithId = React.Children.map(children, (child) => {
+  const childrenWithId = React.Children.map(children, (child, index) => {
```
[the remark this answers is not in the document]
saurabhdaware: kind of yes. It needs to re-render on any changes on children. Most examples I saw on internet for looping over children were looping over it outside any hook.

To be safe, I also put `console.count` to see how many times it is being re-rendered and it wasn't rendered too many times so should be fine?

packages/blade/src/components/ActionList/ActionListFooter.tsx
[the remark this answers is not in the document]
saurabhdaware: https://razorpay.slack.com/archives/G01B3LQ9H0W/p1675054935464709

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }, [setHasFooterAction, props.trailing]);
+
+  const isOnlyActionButton = !props.title && !props.leading && props.trailing;
+
+  return (
+    <StyledActionListFooter
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      ref={footerRef as any}
+      // @ts-ignore: Ignoring because the TS fails for React Native and works for web
+      onMouseDown={() => {
```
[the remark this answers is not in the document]
saurabhdaware: If you mean that flicker of animation, it is happening because the blur of input gets triggered and we focus on it again. 

This blur functionality in particular doesn't really change the `BaseInput`'s blur behaviour. It only changes the blur functionality that dropdown controls.

To fix the flicker, BaseInput will need some refactor to support ignoring the animation for the input blur. Lemme look into once again.

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,106 @@
...
+          <Text variant="caption" color="surface.text.subdued.lowContrast">
+            {props.title}
+          </Text>
+        </Box>
+      ) : null}
+      {props.trailing ? (
+        <Box
+          display="flex"
+          alignItems="center"
+          marginLeft={isOnlyActionButton ? undefined : 'auto'}
```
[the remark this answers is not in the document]
saurabhdaware: This is for that only button in footer scenario. When there is no other item but only a button, we do want to change the layout a bit. Normally I have to reserve some space for the text.

image

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,106 @@
...
+          width={isOnlyActionButton ? '100%' : undefined}
+        >
+          {props.trailing}
+        </Box>
+      ) : null}
+    </StyledActionListFooter>
+  );
+};
+
+ActionListFooter.componentId = componentIds.ActionListFooter;
```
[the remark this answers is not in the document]
saurabhdaware: duplicate of https://github.com/razorpay/blade/pull/966#discussion_r1086280192

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: Missed that one. Using that now for rest of the components. Although here its not possible to use that utility because I am also using ref and it is messing up with types when we're forwarding ref.

Tried few things but didn't quiet work so disabling the ts error for now here.

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
saurabhdaware: It won't set to 0, It will be undefined no? it won't set `tabindex` prop in that case

packages/blade/src/components/Dropdown/useDropdown.ts
saurabhdaware: It is for scenarios where we're inside an iframe. E.g. storybook is inside iframe so it won't really work with storybook.

packages/blade/src/components/Dropdown/useDropdown.ts
saurabhdaware: on select.

Which is why it also has shouldIgnoreBlur. We ignore the blur when click is on dropdown item. This function itself is called on select blur.

saurabhdaware: Offtopic: I am renaming things from `onSelectBlur` to `onTriggerBlur` since there can be confusion that it is attached to select input (it is not)

packages/blade/src/components/ActionList/ActionList.tsx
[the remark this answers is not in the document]
saurabhdaware: refactored this entire file to make more readable and also added comments

packages/blade/src/components/ActionList/ActionList.tsx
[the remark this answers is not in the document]
saurabhdaware: Oh so figured out a way to not run it on every render. I added it inside useMemo and it only runs on children change. that did reduce few re-renders now.

packages/blade/src/components/ActionList/ActionListFooter.tsx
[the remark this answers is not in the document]
saurabhdaware: TODO:

packages/blade/src/components/ActionList/ActionListFooter.tsx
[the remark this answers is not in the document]
saurabhdaware: Don't know how it was working lol. added WithComponentId utility anyway.

packages/blade/src/components/ActionList/ActionListHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+    alignItems: 'center',
+    padding: `${makeSize(props.theme.spacing[3])} ${makeSize(props.theme.spacing[5])}`,
+    backgroundColor: props.theme.colors.brand.gray.a50.lowContrast,
+    margin: `-${makeSize(props.theme.spacing[3])} -${makeSize(props.theme.spacing[3])} ${makeSize(
+      props.theme.spacing[3],
+    )}`,
+  };
+});
+
+const ActionListHeader = (props: { title: string; leading?: React.ReactNode }): JSX.Element => {
```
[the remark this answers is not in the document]
saurabhdaware: We can do this without a breaking change -

```jsx
<ActionListHeader>
  <TextInput />
</ActionListHeader>
```

Or a bit uglier version - 

```jsx
<ActionListHeader leading={<TextInput />} />
```

It can also just be default based on a `variant="filterable"` on Dropdown maybe? 

It seems rare that we will need a breaking change to support input.

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+import { componentIds } from './componentIds';
+import type { StyledActionListItemProps } from './getBaseActionListItemStyles';
+import Box from '~components/Box';
+import type { IconComponent } from '~components/Icons';
+import { useDropdown } from '~components/Dropdown/useDropdown';
+import { Text } from '~components/Typography';
+import { isReactNative, makeAccessible, makeSize } from '~utils';
+import { Checkbox } from '~components/Checkbox';
+
+const ActionListItemContext = React.createContext<{
```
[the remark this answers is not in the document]
saurabhdaware: haha yes done in coming commit

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+}));
+
+const ActionListSection = ({
+  title,
+  children,
+  hideDivider,
+}: {
+  title: string;
+  children: React.ReactNode[] | React.ReactNode;
+  hideDivider?: boolean;
```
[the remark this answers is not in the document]
saurabhdaware: nope. renamed it to `_hideDivider` and marked `@private` in jsdoc.

Also removed this entire logic from web and relying on descendant styling of css there.

Keeping it in react native because there isnt any other simpler way.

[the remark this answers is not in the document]
saurabhdaware: Oh more context: I am looping over children and checking if ActionListSection is the final section and  we hide the divider there.

https://github.com/razorpay/blade/blob/f32c66695951d71312aa96c27485cf5c3e072cd2/packages/blade/src/components/ActionList/actionListUtils.ts#L119-L120

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+}>({});
+
+const SectionDivider = styled(Box)((props) => ({
+  // @TODO: replace this with token value if we add 1px token
+  height: makeSize(1),
+  backgroundColor: props.theme.colors.surface.border.normal.lowContrast,
+  margin: `${makeSize(props.theme.spacing[1])} ${makeSize(props.theme.spacing[3])}`,
+}));
+
+const StyledActionListSectionTitle = styled(Box)((props) => ({
```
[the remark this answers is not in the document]
saurabhdaware: yess no padding shorthand so we need 4 props to do this thing 😢 

Adding a TODO here to replace it with new layout component once I implement that.

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: don't see it necessary as of now. Changed it to `{ name: string; value: boolean }` (name is the name of option and value is the `isSelected` boolean)

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+}: {
+  title: string;
+  children: React.ReactNode[] | React.ReactNode;
+  hideDivider?: boolean;
+}): JSX.Element => {
+  return (
+    <Box>
+      <StyledActionListSectionTitle>
+        <Text color="surface.text.muted.lowContrast" size="small" weight="bold">
+          {title}
```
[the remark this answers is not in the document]
saurabhdaware: I added `role=group` here so that should announce the title. It does have a bit strange behaviour in how it counts the options but I checked few dropdowns from github and they seem to announce it in the similar way so it is consistent with that.

For checking the title, I think we can rely on TS. If the designer has given them a section with title, then that is the only way to add that title in code so don't think so anyone will miss it. Even if they do, it will look strange and will be caught in reviews.

Just wanted to avoid runtime checks as much as possible because we already have too many to make the dropdown work.

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+  } = useDropdown();
+
+  const renderOnWebAs = props.href ? 'a' : 'button';
+  const isSelected =
+    typeof props._index === 'number'
+      ? selectedIndices.includes(props._index)
+      : props.isDefaultSelected;
+
+  return (
+    <ActionListItemContext.Provider value={{ intent: props.intent }}>
```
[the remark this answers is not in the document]
saurabhdaware: done

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+          }
+        }}
+        onMouseDown={() => {
+          setShouldIgnoreBlur(true);
+        }}
+        data-value={props.value}
+        data-index={props._index}
+        // Custom props for changes in styles
+        selectionType={selectionType}
+        hasDescription={!!props.description}
```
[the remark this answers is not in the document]
saurabhdaware: done 🙈

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+        intent={props.intent}
+        isSelected={isSelected}
+      >
+        <Box display="flex" marginTop={props.description ? 'spacing.2' : undefined}>
```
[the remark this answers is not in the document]
saurabhdaware: So first item here is with description and next item is without description,

When we don't have description I can just vertically align the option and it works perfectly.

But when we have description, I have to align it on top baseline and then add a bit of margin on top of icon to not make it go completely on top and align it with title properly
image

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: removed it. will explain over call

packages/blade/src/components/ActionList/StyledActionListItem.web.tsx
saurabhdaware: The element also changes. Here we have `styled(Box)` and on native we have `styled(TouchableOpacity)`

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+        </Box>
+      </StyledActionListItem>
+    </ActionListItemContext.Provider>
+  );
+};
+
+ActionListItem.componentId = componentIds.ActionListItem;
+
+export {
+  ActionListItem,
```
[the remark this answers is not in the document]
saurabhdaware: wasn't sure if we should have that many runtime checks. I added it inside `useEffect` now so that should be fine I guess

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
saurabhdaware: done

packages/blade/src/components/Dropdown/DropdownOverlay.native.tsx
saurabhdaware: yup seems to work

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+            selectInputRef.current?.focus();
+          }
+        }}
+        onMouseDown={() => {
+          setShouldIgnoreBlur(true);
+        }}
+        data-value={props.value}
+        data-index={props._index}
+        // Custom props for changes in styles
+        selectionType={selectionType}
```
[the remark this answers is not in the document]
saurabhdaware: The backgroundColor on selection is only there in single select.
image

In multiselect, it is represented by a checkbox and doesn't need that backgroundColor. Passing this from here to decide that in styled components.

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+  /**
+   * Internally passed from ActionList. No need to pass it explicitly
+   *
+   * @private
+   */
+  _index?: number;
+  leading?: React.ReactNode;
+  trailing?: React.ReactNode;
+  isDefaultSelected?: boolean;
+  intent?: 'negative';
```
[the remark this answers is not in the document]
saurabhdaware: done

packages/blade/src/components/ActionList/ActionListHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <StyledActionListHeader>
+      <Box>{props.leading}</Box>
```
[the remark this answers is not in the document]
saurabhdaware: Added

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,119 +1,163 @@
...
+  shadowOpacity?: '1';
+  shadowRadius?: number;
+  shadowColor?: string;
+  shadowOffset?: string;
+} => {
+  if (isReactNative()) {
+    return {
+      shadowOpacity: '1',
+      shadowRadius: isReactNative() ? undefined : blur,
+      shadowColor: isAndroid() ? undefined : shadowColor,
```
anuraghazra, line 42: I hope you double checked the shadow to ensure it looks as close as possible in all three platforms.

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,119 +1,163 @@
...
+      shadowColor,
+      blur: theme.shadows.blurRadius.level[1],
+    }),
+  };
+});
+
+const StyledListBoxWrapper = styled(Box)((_props) => {
+  if (!isReactNative()) {
+    return {
+      [`& [role=group]:last-child > [role=separator]:last-child`]: {
```
anuraghazra, line 88: ouuu.. very complex, can we add a comment about what this selector does?

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,142 @@
...
+        }
+      }}
+      // @ts-ignore: Ignoring because the TS fails for React Native and works for web
+      onKeyDown={(e) => {
+        const nativeEvent = e.nativeEvent;
+        const shouldIgnoreDropdownKeydown =
+          (nativeEvent.key === ' ' || nativeEvent.key === 'Enter') && activeIndex < 0;
+        // We ignore the selection keydowns on footer to let users click on items on the footer
+        if (shouldIgnoreDropdownKeydown) {
+          if (selectionType === 'single') {
```
anuraghazra, line 88: Can be done with 1 if statement? 
shouldIgnoreDropdownKeydown && selectionType == single

packages/blade/src/components/ActionList/ActionListFooter.tsx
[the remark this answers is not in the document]
anuraghazra: > Don't know how it was working lol

TypeScript can infer/introspect static methods, though it's not recommended to do so, since once you generate the `.d.ts` files the typedefs won't have this.

packages/blade/src/utils/platform/getNativePlatform.native.ts
```diff
@@ -0,0 +1,7 @@
+import { Platform } from 'react-native';
+
+export const getNativePlatform = (): typeof Platform.OS => {
```
anuraghazra, line 3: This naming is quite confusing. 

getNativePlatform?? Native means react-Native or Native to the platform...!? 

Can't we use `getPlatformType` util?

packages/blade/src/components/ActionList/ActionList.tsx
saurabhdaware: Yup I did. Looks a bit strange on Android but don't see much options there anyway ;__;. Also this component on React Native will be replaced with BottomSheet so it's temporary so didn't spend much time to make it too accurate.

ios:
image

android: 
image

packages/blade/src/utils/platform/getNativePlatform.native.ts
saurabhdaware: It's Native to the platform.

`getPlatformType` returns `react-native` but at some places we also want to differentiate between `web`, `ios`, and `android` so that's what it does.

Agree that it is a bit confusing name. Couldn't think of anything else 🤔

packages/blade/src/components/ActionList/ActionList.tsx
saurabhdaware: haha yup. I've been on my css god days for last few weeks because of dropdown. lemme add comment.

packages/blade/src/components/ActionList/ActionListFooter.tsx
saurabhdaware: Nope its not the same right?

we always want to run that `return` when `shouldIgnoreDropdownKeydown` is true. But we just also want to run `setIsOpen(false` when `selectionType === single` is true.

With `&&` it will only run that block when both of it is true.

packages/blade/src/components/ActionList/ActionList.tsx
saurabhdaware: done

packages/blade/src/utils/platform/getNativePlatform.native.ts
saurabhdaware: Oh renamed it to `getOS`. I think that makes more sense. on web it will return `web` but I think that should be fine since that is what react-native-web also seems to do for `Platform.OS`.

anuraghazra: Sounds good.

anuraghazra (approved):

PR #967, opened 2023-01-19
chore: fix dependabot alerts

anuraghazra:
ran https://www.npmjs.com/package/yarn-audit-fix since yarn doesn't support --fix

### DIFF
--- a/.changeset/forty-rabbits-melt.md
+++ b/.changeset/forty-rabbits-melt.md
@@ -1,0 +1,5 @@
+---
+'@razorpay/blade': patch
+---
+
+chore(blade): fix dependabot alerts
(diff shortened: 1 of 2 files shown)


divyanshu013 (commented):
Would also need a patch release.

kamleshchandnani:
@divyanshu013 why would this need a patch release? this just has changed things at this project's level right?

divyanshu013:
Wouldn't these also show as transitive deps for blade consumers 🤔

anuraghazra:
Adding the patch changelog then, Yes I think it's safer, If something breaks atleast people won't scratch their head trying to figure out "what changed???"

divyanshu013 (approved):
Looks good to me.

PR #968, opened 2023-01-19
chore: move examples inside packages

kamleshchandnani:
This PR does 2 things
* Removes CardWithTokens from recipes as its no longer required since we have better recipes now
* move examples directory under packages for easy discovery and if we want to link with our current version of Blade locally then we can add it to workspace and it'll help us with that


anuraghazra (approved):

kamleshchandnani:
@divyanshu013 didn't do that now. because again the hoisting and everything will kick in and might mess up things and will restrict our ability to do experimentation with examples

divyanshu013 (approved):
Looks good to me.

anuraghazra:
What's pending in this? 🤔

divyanshu013:
Looks good to merge @kamleshchandnani

cseas (commented):
I didn't understand the motivation behind this. How does moving examples to a subdirectory help discovery? It actually hurts discovery.

All popular I'm familiar with that provide examples have them at the root level:
- Next.js: https://github.com/vercel/next.js
- Redux: https://github.com/reduxjs/redux

Adding it under `packages/` makes it even more confusing to read the project structure because the general assumption is that every directory under `packages/` is an actual individual package.

cseas (commented):
> move examples directory under packages for easy discovery

I didn't understand the motivation behind this. How does moving examples to a subdirectory help discovery? It actually hurts discovery.

All popular projects I'm familiar with that provide examples have them at the root level:
- Next.js: https://github.com/vercel/next.js
- Redux: https://github.com/reduxjs/redux

Adding it under `packages/` makes it even more confusing to read the project structure because the general assumption is that every directory under `packages/` is supposed to be an individually published package.

PR #970, opened 2023-01-21
fix: package types exports for ESM projects

### LINKED ISSUE #960: TypeScript cannot find declaration files with module nodenext
## Problem

Blade ships with `d.ts` declaration files. However, since the bundled `package.json` doesn't contain a `"types"` field, the consumer project is unable to find these declaration files and as a result ignores the types for components that Blade ships.

VSCode can't find the declaration files either as shown by the below warning:

Screenshot 2023-01-17 at 1 10 50 PM

This means that IntelliSense in the editor doesn't work for types/props. Also, passing wrong props to a component works. For example, the below component compiles without errors.

```tsx
<Button hello="world">Pay Now</Button>
```

## Solution

The `package.json` should contain a ["types"][1] field that points to the declaration file so the bundler and IDE can find it.

Refer ESM guide for types in `exports` field: https://www.typescriptlang.org/docs/handbook/esm-node.html

Note that the *fallback* types field in the above link is required even for ESM bundles right now because proper support for `exports` field is a part of TypeScript 5 milestone: https://github.com/microsoft/TypeScript/issues/50794#event-8030311249

[1]: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package

cseas:
Fixes #960 

## Description

TypeScript requires types to be declared in the `"types"` property under `"exports"` section of `package.json` of npm packages to detect declaration files in a project setup that outputs ESM. 
(combination of `type:module` in project's `package.json` and `module:NodeNext` in `tsconfig.json`)

I've added the missing fields. These fields are only used by TypeScript in an ESM setup so backwards compatibility shouldn't be affected.

**Note:** An alternate shorter solution is to rename the generated types files for web from `index.d.ts` to `index.web.d.ts`. While that would work for the above config, it will break types for `moduleResolution:node`. Since the automatic detection of TypeScript varies based on the moduleResolution used, it's better to tell TypeScript explicitly where the types are for each export instead of relying on automatic detection. Renaming the generated file will also cause a backwards compatibility issue as pointed out by this code comment:

https://github.com/razorpay/blade/blob/master/packages/blade/rollup.config.js#L127

## Testing

Tested the below combinations in a web project.

|typescript|module|moduleResolution|
|-|-|-|
| 4.9.4 | NodeNext | NodeNext |
| 4.9.4 | es6 | node |
| 4.6.4 | es6 | node |

**Note:** Haven't tested this solution for a react-native project.

## Screenshots

|Before|
|:-:|
| Screenshot 2023-01-21 at 2 48 07 PM |
| After |
| Screenshot 2023-01-21 at 3 40 06 PM |

## References

- https://www.typescriptlang.org/docs/handbook/esm-node.html
- https://webpack.js.org/guides/package-exports/

### DIFF
--- a/.changeset/pretty-swans-perform.md
+++ b/.changeset/pretty-swans-perform.md
@@ -1,0 +1,5 @@
+---
+'@razorpay/blade': major
+---
+
+fix: add types field to package exports for ESM TypeScript projects
--- a/.changeset/pretty-swans-perform.md
+++ b/.changeset/pretty-swans-perform.md
@@ -1,5 +1,5 @@
 ---
-'@razorpay/blade': major
+'@razorpay/blade': patch
 ---
 
 fix: add types field to package exports for ESM TypeScript projects
--- a/packages/blade/package.json
+++ b/packages/blade/package.json
@@ -28,8 +28,14 @@
   ],
   "exports": {
     "./components": {
-      "react-native": "./build/components/index.native.js",
-      "default": "./build/components/index.web.js"
+      "react-native": {
+        "import": "./build/components/index.native.js",
+        "types": "./build/components/index.native.d.ts"
+      },
+      "default": {
+        "import": "./build/components/index.web.js",
+        "types": "./build/components/index.d.ts"
+      }
     },
     "./tokens": {
       "react-native": "./build/tokens/index.native.js",
--- a/packages/blade/package.json
+++ b/packages/blade/package.json
@@ -38,12 +38,24 @@
       }
     },
     "./tokens": {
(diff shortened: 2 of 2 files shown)


kamleshchandnani:
can you test this on react native project and confirm? AFAIK the exports field has no effect on RN projects yet.

kamleshchandnani (approved):
LGTM ✅ 

Just check if things are not breaking on React Native

cseas:
Ack. Will test on React Native soon as I get some time.

PR #973, opened 2023-01-24
chore(OTPInput): add `autoCompleteSuggestionType` and disable password manager with `isMasked` 

chaitanyadeorukhkar:
We want to disable the password managers for OTPInput when `isMasked` is set. The straightforward way to do this is set `autocomplete='off'` (i.e autoCompleteSuggestionType='none'). The issue with `autocomplete` is that its not an enforcement but a suggestion to the browser to follow. Safari mostly adheres to it but Chrome and Firefox ignores it and shows the password managers anyway. @saurabhdaware suggested a workaround to unset `type` on first render and set it to `password` once a value is entered. This way the password managers won't make any suggestions when the user is on an empty OTP input field.


divyanshu013 (approved):
Looks good to me.

divyanshu013 (commented):
- [x] Add changeset

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
```diff
@@ -214,6 +249,9 @@ const OTPInput = ({
     for (let index = 0; index < otpLength; index++) {
       const currentValue = inputValue ? otpToArray(inputValue)[index] || '' : otpValue[index] || '';
       const ref = React.createRef<HTMLInputElement>();
+      // if an inputValue is passed (controlled) and isMasked is set, inputType will always be password
+      const currentInputType =
+        inputValue && isMasked ? 'password' : isMasked ? inputType[index] : undefined;
```
kamleshchandnani, line 254: why undefined and not text?

chaitanyadeorukhkar: It was `undefined` before this too. I think we didn't set it as text explicitly since we weren't sure if our use-cases would be numeric or alphanumeric.

saurabhdaware (approved):
Looks good to me 

image

.changeset/beige-steaks-hug.md
```diff
@@ -0,0 +1,7 @@
+---
+"@razorpay/blade": patch
+---
+
+chore(OTPInput): add `autoCompleteSuggestionType` prop and disable password manager with isMasked
```
divyanshu013, line 5:
```suggestion
chore(OTPInput): add `autoCompleteSuggestionType` prop and disable password manager with `isMasked`
```

.changeset/beige-steaks-hug.md
```diff
@@ -0,0 +1,7 @@
...
+
+We wanted to disable the password managers for OTPInput when isMasked is set. The straightforward way to do this is set autocomplete='off' (i.e autoCompleteSuggestionType='none'). The issue with autocomplete is that its not an enforcement but a suggestion to the browser to follow. Safari mostly adheres to it but Chrome and Firefox ignores it and shows the password managers anyway. We decided on a workaround to unset `type` on first render and set it to `password` once a value is entered. This way the password managers won't make any suggestions when the user is on an empty OTP input field.
```
divyanshu013, line 7:
```suggestion
We wanted to disable the password managers for OTPInput when `isMasked` is set. The straightforward way to do this is set autocomplete='off' (i.e autoCompleteSuggestionType='none'). The issue with autocomplete is that its not an enforcement but a suggestion to the browser to follow. Safari mostly adheres to it but Chrome and Firefox ignores it and shows the password managers anyway. We decided on a workaround to unset `type` on first render and set it to `password` once a value is entered. This way the password managers won't make any suggestions when the user is on an empty OTP input field.
```

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
```diff
@@ -99,6 +118,23 @@ const OTPInput = ({
     }
   }, [otpValue, otpLength, name, inputValue, onOTPFilled]);
 
+  useEffect(() => {
```
divyanshu013, line 121: I think this might not be the best approach, prone to breaking. What happens during the initial render during SSR? Would it correctly mask the values for controlled / uncontrolled?

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
divyanshu013, line 254: Suggestion: the nested ternaries is making this condition harder to read

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
chaitanyadeorukhkar: I ran a test with controlled input on ssr tests and it does create a snapshot with an input of type password. Is there any other way we can test this?

Screenshot 2023-02-06 at 3 22 08 PM

divyanshu013: This one seems fine. How about uncontrolled one?

Overall though, this logic still feels complex 😅

divyanshu013, line 121: Suggestion: add a comment to explain why we need this `useEffect`

divyanshu013: Let's skip for now since it's hard to validate right now.

divyanshu013 (approved):
Looks good to me. A couple of minor comments above.

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
chaitanyadeorukhkar: Done

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
chaitanyadeorukhkar: Simplified it

PR #974, opened 2023-01-24
fix: mark react-native dependencies as optional

### LINKED ISSUE #957: Mark react-native as an optional dependency
## Problem

This is a follow-up for #954 

After the fix deployed from #955, the dependency resolution for react v18 in npm v8 is working correctly. 🎉 

However, the npm install still fails because `react-native` is marked as a peerDependency which in turn depends on `react@"17.0.2"`. Please see the below log for the full error:

```sh
npm i @razorpay/blade@5.5.1
```

> npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: undefined@undefined
npm ERR! Found: react@18.2.0
npm ERR! node_modules/react
npm ERR!   react@"^18.2.0" from the root project
npm ERR!   peer react@">=17" from @razorpay/blade@5.5.1
npm ERR!   node_modules/@razorpay/blade
npm ERR!     @razorpay/blade@"5.5.1" from the root project
npm ERR!   1 more (react-dom)
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"17.0.2" from react-native@0.66.5
npm ERR! node_modules/react-native
npm ERR!   peer react-native@"^0.66.5" from @razorpay/blade@5.5.1
npm ERR!   node_modules/@razorpay/blade
npm ERR!     @razorpay/blade@"5.5.1" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.

## Solution

I feel this is largely a problem created by the react-native team since they specify exact version of react in peerDependencies instead of a range. And the error can be fixed upstream from react-native.

However, on blade's end, requiring a consumer project (that's only building for web) to install react-native isn't right either. We can fix this error from blade too by marking the react-native dependency as optional in [peerDependenciesMeta][1]. This would make sure a consumer project installs the right version of react-native if they want to, but the install would be optional.

Happy to send a PR if this makes sense.

[1]: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

cseas:
[Jira][1] | Fixes #957 

Users of Blade are forced to use the `--legacy-peer-deps` flag for installing packages because starting from v7, npm install fails if all peer dependencies specified by a package are not installed by the host project. This is a problem because this means Blade's peerDependencies config will force all web projects to install react-native unless they use the above flag.

Screenshot from the dashboard project:

Screenshot 2023-01-19 at 2 12 50 PM

The `peerDependenciesMeta` field of `package.json` is used for solving exactly this problem. This will inform npm that it needs to check the version of react-native peerDependencies of Blade only for projects that already have react-native installed on the host. 

Refer: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

[1]: https://razorpay.atlassian.net/browse/OM-16

### DIFF
--- a/.changeset/chilled-radios-greet.md
+++ b/.changeset/chilled-radios-greet.md
@@ -1,0 +1,5 @@
+---
+'@razorpay/blade': patch
+---
+
+fix: mark react-native peerDependencies as optional
--- a/packages/blade/package.json
+++ b/packages/blade/package.json
@@ -175,6 +175,14 @@
     "react-native-svg": "^12.3.0",
     "styled-components": "^5"
   },
+  "peerDependenciesMeta": {
+    "react-native": {
+      "optional": true
+    },
+    "react-native-reanimated": {
+      "optional": true
+    }
+  },
   "resolutions": {
     "@storybook/**/react-dom": "17.0.2",
     "@storybook/**/react": "17.0.2",


kamleshchandnani (approved):
Thanks for this contribution 🎉

PR #977, opened 2023-01-25
feat: add `List` component

chaitanyadeorukhkar:
- Adds `List` component
- RFC For List API: #952


packages/blade/src/components/Icons/_Svg/Rect/Rect.native.tsx
```diff
@@ -2,8 +2,8 @@ import type { ReactElement } from 'react';
 import { Rect as RectNative } from 'react-native-svg';
 import type { RectProps } from './types';
 
-const Rect = ({ height, width, rx, ry, x, y }: RectProps): ReactElement => {
-  return <RectNative height={height} width={width} rx={rx} ry={ry} x={x} y={y} />;
+const Rect = ({ height, width, fill, rx, ry, x, y }: RectProps): ReactElement => {
```
anuraghazra, line 5: So because of this, is there a chance that some icon colors were broken?

chaitanyadeorukhkar: Yup 😓

packages/blade/src/components/Box/getBoxStyles.ts
```diff
@@ -22,13 +22,29 @@ const getSpacingValue = <SpacingValue extends string | number | undefined>({
...
+const getBorderRadiusValue = <RadiusValue extends string | undefined>({
+  value,
+  theme,
+}: {
+  value: RadiusValue;
+  theme: Theme;
+}): string | undefined => {
+  if (value) {
+    const radius = getIn(theme, `border.radius.${value}`);
+    if (radius.toString().includes('%')) return radius;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: We have a radius token of `50%` that should be used as is. Rest of the radius tokens are numbers that need to be converted to px with `makeSpace`

packages/blade/src/components/List/listTokens.ts
```diff
@@ -0,0 +1,142 @@
...
+  >
+> = {
+  unordered: {
+    onDesktop: {
+      small: 6,
+      medium: 'spacing.3',
+    },
+    onMobile: {
+      small: 6,
+      medium: 10,
```
anuraghazra, line 55: For these tokens in some places we are using `number` but in some places `{number}px`, let's use only 1 format for hardcoded tokens?

packages/blade/src/components/List/getOrderedListItemBullet.ts
```diff
@@ -0,0 +1,74 @@
...
+  const firstAlphabets = indexedNumber >= 26 ? alphabetize(Math.floor(indexedNumber / 26) - 1) : '';
+  const lastAlphabet = alphabets[indexedNumber % 26];
+  return `${firstAlphabets}${lastAlphabet}`;
+};
+
+/**
+ * Get the list item bullet number/roman number/alphabet for the passed level and list item number
+ *
+ * @param {{itemNumber: number, level: number}} { itemNumber, level }
+ * @returns {string} OrderedList bullet
```
anuraghazra, line 52: Let's remove the jsdoc typings, since we already have types

anuraghazra:
Bug in storybook "Ordered - Sizes"

image

packages/blade/src/components/List/ListContext.ts
```diff
@@ -0,0 +1,24 @@
...
+const ListContext = React.createContext<ListContextType>({
+  level: undefined,
+  size: 'medium',
+  variant: 'unordered',
+});
+const ListProvider = ListContext.Provider;
+
+const useListContext = (): ListContextType => {
+  const context = React.useContext(ListContext);
+  return context;
```
anuraghazra, line 21: Let's throw an error here if `ListItem` is used outside of <List \/>?

if(!context) throw error

packages/blade/src/components/List/ListItem.tsx
```diff
@@ -0,0 +1,146 @@
...
+                ? getIn(theme, 'colors.brand.gray.a100.lowContrast')
+                : undefined
+            }
+          >
+            {variant === 'ordered' ? (
+              <Text variant="body" type="subtle" size={size}>
+                {`${getOrderedListItemBullet({
+                  itemNumber: _itemNumber ?? 1,
+                  level: level ?? 1,
+                })}.`}
```
anuraghazra, line 122: Why is there a `.` in here but not in the else branch?

anuraghazra (approved):
Minor Comments, LGTM

packages/blade/src/components/Box/getBoxStyles.ts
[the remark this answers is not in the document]
anuraghazra: Oh I see, got a bit, lets add the comment on code?

packages/blade/src/components/List/listTokens.ts
chaitanyadeorukhkar: So wherever the spacing value matched our tokens, I opted to use our tokens. Its only the cases where we can't use our tokens we'll hard code those values.

packages/blade/src/components/List/ListItem.tsx
chaitanyadeorukhkar: Thats as per design. I've cleaned it up a bit now to make it more readable. `ordered` variant has a `.` but `ordered-filled` variant doesn't have a `.`

kamleshchandnani:
Some UI feedback:
* So if we have nesting after the 3rd level we show all of them at the same level? if that is expected can we throw an error and ask users to clean up and add things at the 3rd-level flat? right now I can go on nesting but the rendering will stop nesting after 3rd level.
* The a11y contrast ratio is failing when we have links. This seems a core problem with links itself not with this component but let's ask Saurav and fix it. Here's the [link to the story](https://61c19ee8d3d282003ac1d81c-yppubiruph.chromatic.com/?path=/story/components-list--list-with-link-and-icon&globals=measureEnabled:false) where you can check this
Screenshot 2023-02-06 at 1 01 39 PM

* Few of the stories where the knobs are not required but still are shown can we remove them? examples - ordered sizes, unordered sizes, and unordered-filled sizes.

* The character `a` is not vertically center aligned when nesting with `ordered-filled`
Screenshot 2023-02-06 at 1 16 37 PM

chaitanyadeorukhkar:
>So if we have nesting after the 3rd level we show all of them at the same level? if that is expected can we throw an error and ask users to clean up and add things at the 3rd-level flat? right now I can go on nesting but the rendering will stop nesting after 3rd level.

Added an error

> The a11y contrast ratio is failing when we have links. This seems a core problem with links itself not with this component but let's ask Saurav and fix it. Here's the [link to the story](https://61c19ee8d3d282003ac1d81c-yppubiruph.chromatic.com/?path=/story/components-list--list-with-link-and-icon&globals=measureEnabled:false) where you can check this

Will follow up

>Few of the stories where the knobs are not required but still are shown can we remove them? examples - ordered sizes, unordered sizes, and unordered-filled sizes.

Done

> The character a is not vertically center aligned when nesting with ordered-filled

That would be because the spacing we've defined on top was made only for numerical values. The `ordered-filled` variant will have only 1 level of nesting in design. I didn't add this as an extra restriction since this is already a specific use-case variant and it already has a restriction on design. So the screenshot you've shared will never really be used. I'm not sure if its worth investing time in fixing this or explicitly adding restrictions on this specific variant. 

// @kamleshchandnani

packages/blade/src/components/List/ListItem.tsx
```diff
@@ -0,0 +1,141 @@
...
+  /**
+   * Icon to be rendered for a ListItem's bullet.
+   *
+   */
+  icon?: IconComponent;
+  /**
+   * This is a private prop to be used only for internal logic purposes.
+   *
+   */
+  _itemNumber?: undefined;
```
kamleshchandnani, line 36: i'm afraid people might abuse it 😢 . can't we set this prop as context value or something so that can be read just internally?

packages/blade/src/components/List/ListItem.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const ItemIcon = Icon ?? ListContextIcon;
+
+  if (level && level > 3) {
+    throw new Error('[Blade List]: List Nesting is allowed only upto 3 levels.');
+  }
+
+  const childrenArray = React.Children.toArray(children);
+
+  const childItem = childrenArray.filter((child) =>
+    isValidAllowedChildren(child, 'List') ? null : child,
```
kamleshchandnani, line 68: what if i pass something else than List and text? that will also be considered as a valid list item based on this logic but we shall prevent that IMO

packages/blade/src/components/List/ListItemCode.tsx
```diff
@@ -0,0 +1,14 @@
+import type { CodeProps } from '../Typography';
+import { Code } from '../Typography';
+import { useListContext } from './ListContext';
+import { listItemCodeSize } from './listTokens';
+
+type ListItemCodeProps = Exclude<CodeProps, 'size' | 'variant' | 'isDisabled'>;
```
kamleshchandnani, line 6: why exclude? we shall pass all the necessary props for code right?

packages/blade/src/components/List/__tests__/List.native.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+      <List variant="ordered-filled" size="small">
+        <ListItem>Level 1</ListItem>
+        <ListItem>Level 2</ListItem>
+        <ListItem>Level 3</ListItem>
+      </List>,
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render medium ordered-filled List', () => {
```
kamleshchandnani, line 171: add tests for error cases?
* What happens when you pass some other element apart from ListItem to List?

packages/blade/src/components/List/__tests__/List.web.test.tsx
```diff
@@ -0,0 +1,229 @@
...
+                      <ListItem>Level 4</ListItem>
+                    </List>
+                  </ListItem>
+                </List>
+              </ListItem>
+            </List>
+          </ListItem>
+        </List>,
+      ),
+    ).toThrow('[Blade List]: List Nesting is allowed only upto 3 levels.');
```
kamleshchandnani, line 208: didn't see this test case in native?

packages/blade/src/components/List/getOrderedListItemBullet.ts
```diff
@@ -0,0 +1,68 @@
...
+    ix: 9,
+    v: 5,
+    iv: 4,
+    i: 1,
+  };
+  let result = '';
+
+  Object.keys(romanKeys).forEach((key) => {
+    result += key.repeat(Math.floor(number / romanKeys[key]));
+    number %= romanKeys[key];
```
kamleshchandnani, line 25: can't this just work with a map from 1-i, 2-ii and so on? why does it needs this logic?

packages/blade/src/components/List/listTokens.ts
```diff
@@ -0,0 +1,152 @@
...
+import type { TypographyPlatforms } from '~tokens/global/typography';
+import type { DotNotationSpacingStringToken } from '~src/_helpers/types';
+
+const listItemPaddingLeft: Record<
+  NonNullable<ListProps['variant'] | 'unorderedWithIcon'>,
+  Record<number, DotNotationSpacingStringToken>
+> = {
+  unordered: {
+    1: 'spacing.2',
+    2: 'spacing.5',
```
kamleshchandnani, line 12: why do level 2 and level 3 have same padding? shouldn't 3 have more padding 🤔

packages/blade/src/components/List/listTokens.ts
```diff
@@ -0,0 +1,152 @@
+import type { CodeProps } from '../Typography';
+import type { ListProps } from './List';
+import type { TypographyPlatforms } from '~tokens/global/typography';
+import type { DotNotationSpacingStringToken } from '~src/_helpers/types';
+
+const listItemPaddingLeft: Record<
```
kamleshchandnani, line 6: do we really need to have padding? can we not use margins here?

packages/blade/src/components/List/listTokens.ts
kamleshchandnani: wherever we are not using our tokens why not plainly use hardcoded px value directly? for eg: `small: '6px',`?

packages/blade/src/components/Box/getBoxStyles.ts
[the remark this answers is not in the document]
chaitanyadeorukhkar: We already had a `makeBorderSize` function. Idk why I didn't use that in the first place. Removed all this now

packages/blade/src/components/List/listTokens.ts
chaitanyadeorukhkar: Because our `Box` component does that for us. Box accepts a `number` or a `DotNotationSpacingValue`. If we pass a number, it'll run `makeSpace` on it

packages/blade/src/components/Checkbox/__tests__/__snapshots__/Checkbox.web.test.tsx.snap
```diff
@@ -82,7 +82,7 @@ exports[`<Checkbox /> should render checkbox with label 1`] = `
   font-style: normal;
   -webkit-text-decoration-line: none;
   text-decoration-line: none;
-  line-height: 1.25rem;
+  line-height: 1rem;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Because this test uses Link component 🤦  Nothing to worry about

```jsx
it('should render checkbox with label', () => {
    const label = (
      <>
        I accept <Link href="#">Terms and Conditions</Link>
      </>
    );
```

packages/blade/src/components/List/List.tsx
```diff
@@ -0,0 +1,127 @@
...
+  MetaConstants,
+} from '~utils';
+import type { DotNotationSpacingStringToken } from '~src/_helpers/types';
+
+type ListCommonProps = {
+  /**
+   * ListItem to be rendered for the List.
+   *
+   */
+  children: React.ReactElement<ListItemProps> | React.ReactElement<ListItemProps>[];
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: No, because when we only pass a single ListItem component, it will throw an error demanding an array of it 😓

packages/blade/src/components/List/List.tsx
```diff
@@ -0,0 +1,127 @@
...
+  variant?: 'ordered';
+  icon?: undefined;
+};
+
+type ListOrderedFilledProps = ListCommonProps & {
+  variant?: 'ordered-filled';
+  icon?: undefined;
+};
+
+type ListProps = ListUnorderedProps | ListOrderedProps | ListOrderedFilledProps;
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yes thats a simpler type def

packages/blade/src/components/List/ListContext.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export type ListContextType = {
+  level?: number;
+  size: NonNullable<ListProps['size']>;
+  icon?: IconComponent;
+  variant: NonNullable<ListProps['variant']>;
+};
+
+const ListContext = React.createContext<ListContextType>({
+  level: undefined,
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: So List will be used within another List which means that the ListContext will exist within another ListContext. So by default, the level of any ListContext is unknown (undefined) until its evaluated by checking if another ListContext exists. So only on level 1 will we have the default `undefined` level but as we move deeper the level will be defined. So for the convenience of doing `level = listContext.level ? listContext.level +1 : 1` we set it to be undefined before any evaluation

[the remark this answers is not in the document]
kamleshchandnani: then make this as 0th level instead of undefined? at least it will be predictable?

packages/blade/src/components/List/List.tsx
```diff
@@ -0,0 +1,127 @@
...
+      size: size ?? listContextSize,
+      icon,
+      variant,
+    }),
+    [icon, level, listContextSize, size, variant],
+  );
+
+  const childrenArray = React.Children.toArray(children);
+  const childListItems = childrenArray.filter((child) =>
+    isValidAllowedChildren(child, 'ListItem') ? child : null,
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: I see how this could be misleading. I've cleaned up the logic and added comments. This also throws an error on invalid children now

packages/blade/src/components/List/List.tsx
```diff
@@ -0,0 +1,127 @@
...
+      <ListElement
+        marginTop={!level ? 'spacing.3' : undefined}
+        {...metaAttribute(MetaConstants.Component, MetaConstants.List)}
+        {...makeAccessible({ role: 'list' })} // Role needed for react-native
+      >
+        {variant === 'unordered'
+          ? childListItems
+          : childListItems.map(
+              (child, index) =>
+                React.cloneElement(child as React.ReactElement, { _itemNumber: ++index }), // adds _itemNumber for rendering ordered list bullets
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: No strong reason. Just preference 😅 Anyway I'll make it index+1 for readability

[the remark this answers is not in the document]
chaitanyadeorukhkar: Ordered list also has a bullet its just a number/roman number/alphabet. Couldn't find a better common name other than bullet to address it 😅

packages/blade/src/components/List/ListItem.tsx
chaitanyadeorukhkar: Couldn't really see a cleaner way to do this than setting a private prop. Context is a bit messy because there will be 0 to 3 levels of ListContext. This makes the logic much simpler and easy to understand

packages/blade/src/components/List/ListItem.tsx
chaitanyadeorukhkar: Yes, prevented that now. It will throw an error

packages/blade/src/components/List/ListItemCode.tsx
chaitanyadeorukhkar: Copy paste issue. Code doesn't have variant or isDisabled. Removed it

packages/blade/src/components/List/__tests__/List.native.test.tsx
chaitanyadeorukhkar: Added

chaitanyadeorukhkar: Added

packages/blade/src/components/List/__tests__/List.web.test.tsx
chaitanyadeorukhkar: Added

packages/blade/src/components/List/getOrderedListItemBullet.ts
chaitanyadeorukhkar: This just makes it scalable and the logic won't fail even if we add 100+ items.  Same reasoning for a,b,c... level

packages/blade/src/components/List/listTokens.ts
chaitanyadeorukhkar: Since level 2 items are rendered inside level 1 items, there is already a padding added by level 1 and this level 2 padding is space in addition to the level 1 space. So if level 1 has 4px padding and we specify 8px padding to level 2 then level 2 will be 4+8=12px from the leftmost point

packages/blade/src/components/List/listTokens.ts
chaitanyadeorukhkar: Made it margins. I wanted to do this earlier but lost track of it

packages/blade/src/components/List/ListContext.ts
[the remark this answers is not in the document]
chaitanyadeorukhkar: Its not sitting well with the logic. List can be of level 1, 2, 3 or it can be unknown. It will never be a level 0. Making it 0 will mean that I will have to do something like thi
```diff
- level: level ? level + 1 : 1
+ level: level +1 
```

Now looking at this it would make me think that level is _always_ incremented even when there is just 1 list which looks like incorrect logic. Whereas doing a check if the level is unknown then make it 1 otherwise increment it seems more logical.

kamleshchandnani (commented):
overall lgtm 🚀 
minor comments

packages/blade/src/components/List/ListItem.tsx
```diff
@@ -64,11 +70,26 @@ const ListItem = ({ children, icon: Icon, _itemNumber }: ListItemProps): React.R
 
   const childrenArray = React.Children.toArray(children);
 
+  // Get children that are not a List component
   const childItem = childrenArray.filter((child) =>
-    isValidAllowedChildren(child, 'List') ? null : child,
+    getComponentId(child) !== 'List' ? child : null,
   );
+  const validChildItem = childItem.filter((child) => {
+    if (
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yes yes

packages/blade/src/components/List/ListItemCode.tsx
```diff
@@ -3,12 +3,14 @@ import { Code } from '../Typography';
...
 
 const ListItemCode = ({ children }: ListItemCodeProps): React.ReactElement => {
   const { size } = useListContext();
 
   return <Code size={listItemCodeSize[size]} children={children} />;
 };
 
+ListItemCode.componentId = 'ListItemCode';
+
 export { ListItemCode, ListItemCodeProps };
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: ListItemCode is not an internal component, people will consume this to add code inline to list item

kamleshchandnani (approved):

PR #981, opened 2023-01-30
test(Dropdown): add tests, documentation, and cleanup

saurabhdaware:
Add tests, storybook documentation, and jsdoc.

**Storybook:** https://61c19ee8d3d282003ac1d81c-brgjegznoq.chromatic.com/?path=/docs/components-dropdown-with-select--with-single-select&globals=measureEnabled:false

**JSDoc changes:**
image


packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -85,7 +83,7 @@ const _SelectInput = (
         {...baseInputProps}
         as="button"
         componentName={MetaConstants.SelectInput}
-        ref={triggererRef as React.MutableRefObject<HTMLInputElement>}
+        ref={!isReactNative() ? (triggererRef as React.MutableRefObject<HTMLInputElement>) : null}
```
saurabhdaware, line 86: React Native Testing library doesn't support refs https://github.com/callstack/react-native-testing-library/issues/1069#issuecomment-1221105858

There are some workarounds but we're anyway not using this ref in React native so removed it from here

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -9,12 +9,17 @@ import { isValidAllowedChildren } from '~utils';
 type DropdownProps = {
   selectionType?: 'single' | 'multiple';
   children: React.ReactNode[];
-};
+}; //
```
anuraghazra, line 12:
```suggestion
};
```

packages/blade/src/components/Dropdown/DropdownOverlay.native.tsx
```diff
@@ -8,42 +8,44 @@ import Box from '~components/Box';
...
 
   return (
     <Box position="relative">
-      <StyledClosableArea
+      <StyledCloseableArea
         display={isOpen ? 'flex' : 'none'}
         onPress={() => {
           setIsOpen(false);
         }}
+        testID="closeable-area"
```
anuraghazra, line 37: possible to use `testID()` function from utils? 

Plus is the data-testid set on web side?

saurabhdaware: nope. On web I can use `getByRole` with a name option on listbox/menu but in the `@testing-library/react-native` version that we're using, it doesn't seem to be supported.

saurabhdaware: Do I use `testID` util if I only want it in react-native? don't see any harm in adding `data-testid` as well just that I wouldn't recommend using that for web tests 🤔

packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
```diff
@@ -0,0 +1,347 @@
...
+    // Move to first item
+    await user.keyboard('{ArrowDown}');
+    expect(getActiveDescendant()).toBe('Apple');
+
+    // Move to second item
+    await user.keyboard('{ArrowDown}');
+    expect(getActiveDescendant()).toBe('Mango');
+
+    // Select option
+    await user.keyboard('[Space]');
```
anuraghazra, line 182: Let's also test PageUp/PageDown, moves focus to first/last element

packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
```diff
@@ -0,0 +1,347 @@
...
+      </Dropdown>,
+    );
+
+    const selectInput = getByRole('combobox', { name: 'Fruits' });
+
+    expect(selectInput).toBeInTheDocument();
+    expect(selectInput.textContent).toBe('Select Option');
+    expect(container.querySelector('[role=listbox]')).not.toBeVisible();
+
+    const getActiveDescendant = (): string | null | undefined => {
```
anuraghazra, line 207: Move this outside maybe? We are redeclaring again.

packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+    expect(selectInput.textContent).toBe('Mango');
+
+    // Ensure menu did not close
+    expect(getByRole('listbox', { name: 'Fruits' })).toBeVisible();
+
+    // Move to third item and select
+    await user.keyboard('{ArrowDown}');
+    expect(getActiveDescendant()).toBe('Orange');
+    await user.keyboard('[Space]');
```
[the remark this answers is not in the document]
saurabhdaware: `{` is for `event.key` and `[` is for `event.code`. 

In code we're using `event.key` so I kept that similar here.

For spacebar, `event.key` is ` ` (empty space) so here it becomes `{ }` which looks a bit strange 😅 so using `[Space]` for spacebar

anuraghazra (approved):
Mostly lgtm.

packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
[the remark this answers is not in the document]
anuraghazra: Got it, yeah I used [] only on other test, got confused with this {}

packages/blade/src/components/Dropdown/DropdownOverlay.native.tsx
anuraghazra: > Do I use testID util if I only want it in react-native? don't see any harm in adding data-testid as well just that I wouldn't recommend using that for web tests 🤔

Sure, if we are not doing any testing with the testId let's not add it on the web.

packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
saurabhdaware: done

packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
saurabhdaware: Added cases for Home, End, PageUp, and PageDown in focus move test 👯

PR #982, opened 2023-01-30
fix(blade): checkbox icon wrapper position

anuraghazra:
fixes https://razorpay.slack.com/archives/C01H13RTF8V/p1675077649413379?thread_ts=1675077600.453539&cid=C01H13RTF8V

### DIFF
--- a/packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
+++ b/packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
@@ -25,6 +25,7 @@ const getCheckboxIconWrapperStyles = ({
   const borderColor = border === 'transparent' ? border : getIn(theme, border);
 
   return {
+    position: 'relative',
     display: 'flex',
     alignItems: 'center',
     justifyContent: 'center',
--- a/packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
+++ b/packages/blade/src/components/Radio/RadioIcon/RadioIconWrapperStyles.ts
@@ -32,6 +32,7 @@ const getRadioIconWrapperStyles = ({
   const borderColor = border === 'transparent' ? border : getIn(theme, border);
 
   return {
+    position: 'relative',
     display: 'flex',
     alignItems: 'center',
     justifyContent: 'center',


packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
```diff
@@ -25,6 +25,7 @@ const getCheckboxIconWrapperStyles = ({
   const borderColor = border === 'transparent' ? border : getIn(theme, border);
 
   return {
+    position: 'relative',
```
kamleshchandnani, line 28: can you add a story as well for a lot of checkboxes(a good one not lorem ipsum types 😄)?

anuraghazra: > lot of checkboxes(a good one not lorem ipsum types 😄)?

If there are lot of checkboxes will be better to suggest people to use MultiSelect in that case. (You'll need like 40 or 50 checkboxes to reproduce this bug on page level 😅)

Also, I don't think we should add stories for fixed bugs. Maybe can add it in tests but then again this bug won't be caught in there.

.changeset/short-ligers-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix(blade): checkbox icon wrapper position
```
divyanshu013, line 5: Changelog should also say what bug it fixes. Makes it easier to understand from reader's perspective imo.

anuraghazra: Done!

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Checkbox/CheckboxIcon/CheckboxIconWrapperStyles.ts
kamleshchandnani: > I don't think we should add stories for fixed bugs.
we shall because to prevent future regressions, ideally should be part of tests but since this can't be part of tests we need to put it in stories.

anuraghazra: > maybe add a block of say 30-40px height and add 6-7 checkboxes

Sure, can do that. but fyi still won't catch the regression since you need to scroll the overflow container to see this bug. 

(Good usecase for the storybook play functions though :p)

anuraghazra: Added in https://github.com/razorpay/blade/pull/982/commits/ab15b812c998937f4ea9e49f7b63d1aa741eee9c

PR #988, opened 2023-02-03
feat(Dropdown): SelectInput, ActionList, Dropdown Integration Branch

saurabhdaware:
> **Warning**
>
> Integration branch of stacked PRs. If you're planning to review this directly, maybe don't? checkout the PR stack below

Storybook: https://saurabh-feat-dropdown-integration--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/docs/components-dropdown-with-select--with-single-select

## Stack

- [x] #932 
- [x] #948 
- [x] #966 
- [x] #981 
- [x] Add disabled state (d8d2f3cc8fb6b022737252ff35a8a8971cba5681 63c6fdcf44f7811f3474ce48bb64bab3d3b12766 - Added it as commit in this PR)
- [x] #994 

## Changes

- Adds `Dropdown`, `SelectInput`, and `ActionList` components 
- Adds tests
- Adds documentation
- Minor changes in utils at some places
- Some changes in BaseInput for SelectInput

image

<details>
<summary>Mobile Screenshots</summary>

#### android and ios
image image

</details>


anuraghazra:
Can you add "MultiSelect Form Submission" story? 

image

anuraghazra:
Add storybook props and other controls: 

image

I think you can group them how Kamlesh did it for TextInput story

packages/blade/src/components/Dropdown/useDropdown.ts
```diff
@@ -0,0 +1,359 @@
...
+    else {
+      window.clearTimeout(searchTimeout);
+      searchString = '';
+    }
+  };
+
+  /**
+   * Keydown event of combobox. Handles most of the keyboard accessibility of dropdown
+   */
+  const onTriggerKeydown = (e: { event: React.KeyboardEvent<HTMLInputElement> }): void => {
```
anuraghazra, line 300: Why is this argument like `e.event` can we just pass in `event` directly?

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -64,35 +72,48 @@ const KeyboardTypeToNativeValuesMap = {
...
+const getRNInputStyles = (
+  props: StyledComponentInputProps &
+    ThemeProps<DefaultTheme> &
+    (TextInputProps | TouchableHighlightProps),
+): CSSObject => {
+  return {
+    ...getBaseInputStyles({
+      theme: props.theme,
+      isFocused: props.isFocused,
+      isDisabled: 'editable' in props ? !props.editable : undefined,
```
anuraghazra, line 94: 🤔 Whats happening here? why editable is not in `props`? 

It should just be `!props.editable` 
And add editable in here: 

```
type StyledComponentInputProps = {
  editable: TextInputProps['editable'];
}
```

kamleshchandnani:
some UI interactions feedback:

1. There are a few stories that don't get affected with single/multiple controls but the knobs are still shown. maybe we shall hide when it's not applicable
Screenshot 2023-02-03 at 4 12 47 PM

2. Sometimes the group is announced but sometimes they are not when you have a header and footer
Screenshot 2023-02-03 at 4 04 20 PM

Screenshot 2023-02-03 at 4 04 46 PM

Steps to repro above:
i. first Tab
ii. focus comes on the select input - expected
iii. row over the items the screen reader has some issues
	  - when at "Home" it doesn't says 3 groups
	  - when selecting Download it sometimes says its part of the group but sometimes it doesn't
	  - when at "Pricing" it doesn't says 3 groups

3. When you press the command key after that the focus ring never goes off the selected list item

https://user-images.githubusercontent.com/11384858/216589165-27866c96-895f-469a-9377-333fe275c148.mov

4. On every item selection change the input loses focus and gets back its visible as the blue underline starts animating 😞 

https://user-images.githubusercontent.com/11384858/216589372-0a28401e-2d67-43b5-8b7f-030dc085ff49.mov

5. The HTML required field is it required? Shouldn't we mark it as required and then show our validation state?
Screenshot 2023-02-03 at 3 54 26 PM

6. the icon on the right looks weird, did the design team confirm that this is expected?
Screenshot 2023-02-03 at 3 05 39 PM

7. what is the use case for the footer leading part? right now we have a storybook example with "Search" but is there any real-life use case for that?
8. not now, but we are showing keyboard shortcuts in stories that might mislead because we don't listen to keyboard events yet. so we can fix stories for now by removing the keyboard shortcuts suffix and creating an issue to add the `onkeychange` event
9. Add story with Ref example
10. Add a story with validation states?

saurabhdaware:
> Sometimes the group is announced but sometimes they are not when you have a header and footer

This is expected. `Pricing` is not part of that group. You can see divider there.

kamleshchandnani:
> > Sometimes the group is announced but sometimes they are not when you have a header and footer
> 
> This is expected. `Pricing` is not part of that group. You can see divider there.

but pricing is a group in itself right? like there are 3 option groups? and then 2nd has more options in it?

saurabhdaware:
> but pricing is a group in itself right? like there are 3 option groups? and then 2nd has more options in it?

Nope. There are 3 items (groups and individual items)

Home and Pricing are individual top level items unless we put them inside `ActionListItemSection` and give a title (which becomes a group label)

saurabhdaware:
> The HTML required field is it required? Shouldn't we mark it as required and then show our validation state?

I feel we should support HTML required. I am not sure if we should be the ones handling form submits. 

And if we do that then we might have to go handle validation everywhere no? For example right now this will happen even if you put a wrong email id

saurabhdaware:
> When you press the command key after that the focus ring never goes off the selected list item

This is expected right? we enable the focus when a key is pressed assuming the user wants to navigate using the keyboard https://razorpay.slack.com/archives/G01B3LQ9H0W/p1675411109401079?thread_ts=1675404254.262649&cid=G01B3LQ9H0W

> The HTML required field is it required? Shouldn't we mark it as required and then show our validation state?

I feel we should support HTML required. I am not sure if we should be the ones handling form submits.

And if we do that then we might have to go handle validation everywhere no? For example right now this will happen even if you put a wrong email id

> the icon on the right looks weird, did the design team confirm that this is expected?

By weird you mean the alignment? like the gap between icon and title? Abhishek did mention that the leading part is vertically center aligned with the title https://razorpay.slack.com/archives/G01B3LQ9H0W/p1675319721039799?thread_ts=1674536439.113129&cid=G01B3LQ9H0W

> not now, but we are showing keyboard shortcuts in stories that might mislead because we don't listen to keyboard events yet. so we can fix stories for now by removing the keyboard shortcuts suffix and creating an issue to add the onkeychange event

Can you elaborate this? that won't be handled by us no? E.g. if someone shows cmd + S in suffix then they are supposed to handle opening of settings when those keys are pressed right?

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
saurabhdaware: ops right yes fixed it

packages/blade/src/components/Dropdown/useDropdown.ts
saurabhdaware: So on Select we're using `onKeyDown={onTriggerKeydown}` because our `BaseInput` by default passes `{ event }` object.

I am keeping things in that format only so that we can pass events like we need. E.g. in normal HTML button it will become `onKeyDown={(e) => onTriggerKeydown({ event: e.nativeEvent })}`

anuraghazra (approved):

kamleshchandnani:
> > When you press the command key after that the focus ring never goes off the selected list item
> 
> This is expected right? we enable the focus when a key is pressed assuming the user wants to navigate using the keyboard [razorpay.slack.com/archives/G01B3LQ9H0W/p1675411109401079?thread_ts=1675404254.262649&cid=G01B3LQ9H0W](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1675411109401079?thread_ts=1675404254.262649&cid=G01B3LQ9H0W)

Umm, for the focus ring, we shall enable it when you tab into it right(or you press a key for an item name)? pressing command key has no co-relation with focus ring being enabled? 🤔 

> > The HTML required field is it required? Shouldn't we mark it as needed and then show our validation state?
> 
> I feel we should support HTML required. I am not sure if we should be the ones handling form submits.
> 
> And if we do that then we might have to go handle validation everywhere no? For example, right now this will happen even if you put in a wrong email id

we can rethink this for all our fields. Ideally, we shall control how these browser fields are shown beautifully

> > the icon on the right looks weird, did the design team confirm that this is expected?
> 
> By weird you mean the alignment? like the gap between icon and title? Abhishek did mention that the leading part is vertically center aligned with the title [razorpay.slack.com/archives/G01B3LQ9H0W/p1675319721039799?thread_ts=1674536439.113129&cid=G01B3LQ9H0W](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1675319721039799?thread_ts=1674536439.113129&cid=G01B3LQ9H0W)

yes for leading it makes sense, I've asked for trailing icons on the same thread.

> > not now, but we are showing keyboard shortcuts in stories that might mislead because we don't listen to keyboard events yet. so we can fix stories for now by removing the keyboard shortcuts suffix and creating an issue to add the onkeychange event
> 
> Can you elaborate this? that won't be handled by us no? E.g. if someone shows cmd + S in suffix then they are supposed to handle opening of settings when those keys are pressed right?

yes but do we have places for them? for example in the current story we have cmd+option+space so if lets say the dropdown is opened and the consumer adds an event listener then do we have the corresponding item selection function exposes to them because in that case we'll have to show the action item being selected and then dropdown being closed. are we able to support that currently?

divyanshu013 (approved):
Looks good apart from the reported ios issues. Can be taken up as enhancement imo.

divyanshu013 (approved):
:shipit:

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,173 @@
+/* eslint-disable @typescript-eslint/prefer-ts-expect-error */
```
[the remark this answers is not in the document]
saurabhdaware: there's several `ts-ignore`s that we have to write in this file to ignore the errors on react native due to some event handlers. And we have 2 rules. `ban-ts-ignore` and `prefer-ts-expect-error` and both we have to disable for every `@ts-ignore` so we just decided to ignore it on file level

packages/blade/src/components/ActionList/ActionListFooter.tsx
```diff
@@ -0,0 +1,173 @@
...
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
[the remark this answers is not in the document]
saurabhdaware: we do have it on project level already I think. I have added comments where I've ts-ignored.

In this particular case, `@ts-expect-error` was not resolving the errors

packages/blade/src/hooks/useBladeInnerRef.ts
```diff
@@ -11,6 +11,9 @@ type BladeElementRef = Pick<HTMLElement, 'focus' | 'scrollIntoView'> | Pick<View
  */
 const useBladeInnerRef = (
   targetRef: React.ForwardedRef<BladeElementRef>,
+  handlers?: {
```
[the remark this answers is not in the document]
saurabhdaware: done

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -73,8 +73,10 @@ const DropdownOverlay: WithComponentId<DropdownOverlayProps> = ({ children }): J
 
   React.useEffect(() => {
     if (isOpen) {
+      triggererRef.current?.focus();
```
kamleshchandnani, line 76: how was the focus working on chrome family browsers then?

saurabhdaware: It's a bug specifically on safari browsers https://razorpay.slack.com/archives/G01B3LQ9H0W/p1676028810932499?thread_ts=1676005222.027229&cid=G01B3LQ9H0W

on chrome, the focus works as expected by default. This on chrome doesn't do anything now because it will already have focus.

kamleshchandnani (approved):

PR #992, opened 2023-02-03
fix(checkbox): fixed screen reader class preventing layout jumps

anuraghazra:
Fixes layout jumps on a edge case where if we have lots of checkboxes in a small overflowed container the browser is trying to jump to the hidden inputs which is causing unexpected jumps in the scroll. 

Bug Video: 

https://user-images.githubusercontent.com/35374649/216601410-794ba464-ca45-4f96-a887-70218b400efc.mov

Slack thread where I dived into why it's happening: https://razorpay.slack.com/archives/C01H13RTF8V/p1675426161654839?thread_ts=1675077600.453539&cid=C01H13RTF8V


packages/blade/src/components/VisuallyHidden/ScreenReaderStyles.ts
```diff
@@ -10,6 +10,7 @@ const screenReaderStyles: CSSObject = {
   padding: 0,
   position: 'absolute',
   width: `1px`,
+  left: '-10000px',
```
divyanshu013, line 13: Seems like a hacky fix, is this the only way 👀

divyanshu013: Perhaps add a comment in the code for future reference

anuraghazra: Yeah definitely "looks" hacky, will add the link to https://webaim.org/techniques/css/invisiblecontent/#techniques which explains the properties being used.

divyanshu013: If there is an ultra wide screen display, would this end up being visible to the user?

anuraghazra: > end up being visible

No it will still be invisible, the other properties will ensure that. 

Though it's quite unlikely, zooming out to 25% on my mac only goes to 5760px. 😅 the designs will most likely break first at those extreme screen sizes.

divyanshu013 (approved):
Looks good to me. Do validate on native too.

kamleshchandnani (approved):

PR #993, opened 2023-02-05
docs: add Amount decisions

Priyanshu-C:
[VIEW FORMATTED DOC](https://github.com/razorpay/blade/blob/1c56c5cb4a630ffc6e5ac16d97f74c19351928f5/packages/blade/src/components/Amount/_decisions/decisions.md)


packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,88 @@
+# Amount <!-- omit in toc -->
+
+A amount component is a UI element that displays and formal various currency values in a specified format.
```
divyanshu013, line 3:
```suggestion
Amount component is a UI element that displays and formats various currency values to a specified format.
```

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,88 @@
...
+
+<img  src="./amount-thumbnail.png"  width="380" />
+
+## Design
+
+- [Figma - Amount](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=28012%3A580578&t=3peAz8A2n2Gw4WMl-1)
+
+## API
+
+| Prop | Type | Default | Description | Required |
```
divyanshu013, line 13: Table formatting is broken, please validate in markdown preview.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,88 @@
...
+
+|---|---|---|---|---|
+
+| children | `string` | `undefined` | The label to be rendered within the badge. | ✅ |
+
+| intent | `default`, `positive`,`negative`, `notice`, `information`, `blue` | `default` | The variant of the badge to be rendered. | |
+
+| contrast | `low`, `high` | `low` The contrast of the badge to be rendered. | |
```
divyanshu013, line 21: Seems stale, mentions "badge"

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,88 @@
...
+
+- [Backbase](https://designsystem.backbase.com/v1/components/amount/web)
+
+- [Mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)
+
+- [Salesforce](https://developer.salesforce.com/docs/component-library/bundle/ui:inputCurrency)
+
+- [Dribbble](https://dribbble.com/tags/money_components)
+
+- [Pega](https://design.pega.com/design/currency/)
```
divyanshu013, line 87: Formatting issue, remove line breaks.

divyanshu013 (commented):
Please also include a link to the formatted markdown in PR description.

There are also formatting and stale copy paste issues. Could you please validate the formatted document once and re-request review?

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: Fixed.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: Fixed.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: Fixed.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+| intent         | `default`, `positive`, `negative`, `notice`, `information`, `blue` | `default`   | The variant of the amount to be rendered                                       |          |
+| contrast       | `low`, `high`                                                      | `low`       | The contrast of the amount to be rendered                                      |          |
+| fontWeight     | `bold`, `regular`                                                  | `bold`      | The weight of the amount to be rendered                                        |          |
+| hasDecimals    | `true`, `false`                                                    | `true`      | Indicates whether the amount has or can have decimals                          |          |
+| hasTextPostfix | `true`, `false`                                                    | `true`      | Indicates whether a text postfix should be used                                |          |
+| size           | `3xl`, `2xl`, `xl`, `large`, `medium`, `small`                     | `3xl`       | The size of the amount to be rendered                                          |          |
```
chaitanyadeorukhkar, line 21: We should prefer not to use shortforms in the API. Also, do we need so many sizes?
```suggestion
| size           | `3xlarge`, `2xlarge`, `xlarge`, `large`, `medium`, `small`                     | `3xlarge`       | The size of the amount to be rendered                                          |          |
```

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+| isStyled       | `true`, `false`                                                    | `true`      | Highlight the main amount by making the prefix symbol and decimal digits small |          |
+| screenSize     | `desktop`, `mobile`                                                | `desktop`   | Provides the size of the component based on the platform                       |          |
```
chaitanyadeorukhkar, line 23: This is for design only. On code we automatically handle this based on our platform specific tokens. So if you choose `font[100]` it will reflect to a different size on web & mobile. 
```suggestion
```

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Sample Usage
+
+```jsx
+import { Amount } from '@razorpay/components';
+
+<Amount
+  size="2XL"
+  isStyled="true"
```
chaitanyadeorukhkar, line 32: This should be boolean
```suggestion
  isStyled={true}
```

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+  1000.27
+</Amount>;
+```
+
+## Open Questions
+
+#### Dev
+
+- Should we call it `variant` or `intent`, `isStyled` or something more apt?
+- Should we separate the amount with commas based on the Indian currency format?
```
chaitanyadeorukhkar, line 48: Are we only supporting 1 currency format (INR)? Or should we allow the consumer to decide what format it should be?

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+- What should be the default `intent`? Kept it as `default` for now, but `neutral` seems more apt.
+- Scope of A11y?
+
+#### Design
+
+- Do we need to support more currency types?
```
chaitanyadeorukhkar, line 54: +1

@saurav12

divyanshu013:
Please also include a link to the formatted markdown in PR description.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Design
+
+- [Figma - Amount](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=28012%3A580578&t=3peAz8A2n2Gw4WMl-1)
+
+## API
+
+| Prop           | Type                                                               | Default     | Description                                                                    | Required |
+| -------------- | ------------------------------------------------------------------ | ----------- | ------------------------------------------------------------------------------ | -------- |
+| children       | `string`                                                           | `undefined` | The value to be rendered within the component                                  | ✅       |
```
divyanshu013, line 15: With amount, we're dealing with numbers. I think the value being passed should be a number otherwise it'll involve internal parsing of string to number. Instead we can have a `value` prop for accepting the numeric literal.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+| intent         | `default`, `positive`, `negative`, `notice`, `information`, `blue` | `default`   | The variant of the amount to be rendered                                       |          |
+| contrast       | `low`, `high`                                                      | `low`       | The contrast of the amount to be rendered                                      |          |
+| fontWeight     | `bold`, `regular`                                                  | `bold`      | The weight of the amount to be rendered                                        |          |
+| hasDecimals    | `true`, `false`                                                    | `true`      | Indicates whether the amount has or can have decimals                          |          |
+| hasTextPostfix | `true`, `false`                                                    | `true`      | Indicates whether a text postfix should be used                                |          |
```
divyanshu013, line 20: We're generally following `show*` terminology across other components. Think this should also follow similar naming.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+| -------------- | ------------------------------------------------------------------ | ----------- | ------------------------------------------------------------------------------ | -------- |
+| children       | `string`                                                           | `undefined` | The value to be rendered within the component                                  | ✅       |
+| intent         | `default`, `positive`, `negative`, `notice`, `information`, `blue` | `default`   | The variant of the amount to be rendered                                       |          |
+| contrast       | `low`, `high`                                                      | `low`       | The contrast of the amount to be rendered                                      |          |
+| fontWeight     | `bold`, `regular`                                                  | `bold`      | The weight of the amount to be rendered                                        |          |
+| hasDecimals    | `true`, `false`                                                    | `true`      | Indicates whether the amount has or can have decimals                          |          |
```
divyanshu013, line 19: Is there a usecase for showing no. of decimals? For example decimal upto 1 place / 2 place etc.? Should be confirmed with design.

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013, line 20: Can there be both text postfix and decimals? If not I think both these props should be clubbed in a way that doesn't allow impossible combinations.

divyanshu013, line 20: text postfix seems like a very vague terminology. As a dev I'm not able to visualize what this would do? I think we should use a clearer terminology. Same can be communicated to design and updated in figma.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+| hasTextPostfix | `true`, `false`                                                    | `true`      | Indicates whether a text postfix should be used                                |          |
+| size           | `3xl`, `2xl`, `xl`, `large`, `medium`, `small`                     | `3xl`       | The size of the amount to be rendered                                          |          |
+| isStyled       | `true`, `false`                                                    | `true`      | Highlight the main amount by making the prefix symbol and decimal digits small |          |
```
divyanshu013, line 22: Similarly, this also reads as a vague terminology. styled doesn't actually convey the intent of the prop.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+  size="2XL"
+  isStyled="true"
+  fontWeight="regular"
+  hasDecimals="true"
+  hasTextPostfix="true"
+  intent="information"
+  contrast="high"
+>
+  1000.27
+</Amount>;
```
divyanshu013, line 40: Should be updated to use booleans as `{false}` etc.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+## Open Questions
+
+#### Dev
+
+- Should we call it `variant` or `intent`, `isStyled` or something more apt?
```
divyanshu013, line 47: In other components we're using `intent` where they represent a feedback. In others such as badge where there are non semantic naming like "blue" we're using variant.

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013, line 48: On the designs I see this is being done to all values. In this case we should do this as a default. If there are usecases for values without commas could you confirm with design if they're needed?

saurav12: For now, we will be supporting only INR. Going forward we will add more currencies. One of the major problems right now is we use ₹ and INR interchangeably. But in Malaysia we only have RM and there is no symbol associated with it. Hence we are staying away from this. Will solve this once the use-case arrives.

packages/blade/src/components/Amount/_decisions/decisions.md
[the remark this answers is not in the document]
chaitanyadeorukhkar: Yup it should be `showCurrencyPostfix`

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: @saurav12 What is the required spec here? and what should be the logic of rounding here?

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: It is not impossible to have both postfix and decimal, we can have 21.20 l as a value.

Priyanshu-C: changed it to **showCurrencyPostfix**

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: @chaitanyadeorukhkar @divyanshu013 Can you folks suggest a better name here.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: changed it to **variant**

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: @saurav12 Can you check this?

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: Is this case documented on Figma 🤔

divyanshu013: Sounds good, can you please communicate and get that updated on figma too?

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,71 @@
...
+
+## API
+
+| Prop                | Type                                                       | Default     | Description                                                                    | Required |
+| ------------------- | ---------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------ | -------- |
+| children            | `number`                                                   | `undefined` | The value to be rendered within the component                                  | ✅       |
+| variant             | `positive`, `negative`, `notice`, `information`, `blue`    | `undefined` | The variant of the amount to be rendered                                       |          |
+| contrast            | `low`, `high`                                              | `low`       | The contrast of the amount to be rendered                                      |          |
+| fontWeight          | `bold`, `regular`                                          | `bold`      | The weight of the amount to be rendered                                        |          |
```
anuraghazra, line 18: 🤔 should we add the `font` prefix here? In Text we are naming it just `weight` 

/cc @saurav12

divyanshu013: `fontWeight` definitely is clearer to understand as a prop

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: Please also add comments on figma at the occurence for design

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: Please also confirm on figma

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: I couldn't find this particular case on figma, where it shows both decimals and a suffix like `100.00 k`

divyanshu013: Wondering if we can consolidate decimals and suffix usecases in this case.

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: Maybe `isAmountEnlarged`, `isAmountHighlighted` though a bit verbose prop

Priyanshu-C: `isHighlighted` then?

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: Done.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,71 @@
...
+| hasDecimals         | `true`, `false`                                            | `true`      | Indicates whether the amount has or can have decimals                          |          |
+| showCurrencyPostfix | `true`, `false`                                            | `false`     | Indicates whether a text postfix should be used                                |          |
```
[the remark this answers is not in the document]
Priyanshu-C: Done.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,69 @@
...
+| ------------------------- | ---------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------ | -------- |
+| children                  | `number`                                                   | `undefined` | The value to be rendered within the component                                  | ✅       |
+| intent                    | `positive`, `negative`, `notice`, `information`, `blue`    | `undefined` | The variant of the amount to be rendered                                       |          |
+| fontWeight                | `bold`, `regular`                                          | `bold`      | The weight of the amount to be rendered                                        |          |
+| hasDecimals               | `true`, `false`                                            | `true`      | Indicates whether the amount has or can have decimals                          |          |
+| suffix                    | `true`, `false`                                            | `false`     | Indicates whether a text suffix should be used                                 |          |
```
chaitanyadeorukhkar, line 19:
```suggestion
| hasSuffix                    | `true`, `false`                                            | `false`     | Indicates whether a text suffix should be used                                 |          |
```

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,69 @@
...
+| size                      | `3xlarge`, `2xlarge`, `xlarge`, `large`, `medium`, `small` | `3xlarge`   | The size of the amount to be rendered                                          |          |
+| isSuffixPrefixHighlighted | `true`, `false`                                            | `true`      | Highlight the main amount by making the prefix symbol and decimal digits small |          |
```
chaitanyadeorukhkar, line 21: Where is the prefix?

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+## Open Questions
+
+#### Dev
+
+- Should we call it `variant` or `intent`, `isStyled` or something more apt?
+- Should we separate the amount with commas based on the Indian currency format?
+- What should be the default `intent`? Kept it as `default` for now, but `neutral` seems more apt.
+- Scope of A11y?
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Any update on this @Priyanshu-C ^

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Should we separate the amount with commas based on the Indian currency format?
+
+- For now, we will be supporting only INR. Going forward we will add more currencies. One of the major problems right now is we use ₹ and INR interchangeably. But in Malaysia we only have RM and there is no symbol associated with it. Hence we are staying away from this. Will solve this once the use-case arrives.
+
+## References
+
+- [Ant](https://ant.design/components/input)
+- [Primeng](https://primeng.org/inputnumber)
+- [Blueprint](https://blueprintjs.com/docs/#core/components/file-input)
```
chaitanyadeorukhkar, line 64: These seem unrelated?

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: They are referring to the Rupee symbol here.

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: Can't think of a better prop name right now 🤔 

Definitely seems better than `isStyled`.

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: Should we just make this `isHighlighted`, seems too verbose right now and actually we're not highlighting them, instead we're shrinking them 😅

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+## Design
+
+- [Figma - Amount](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=28012%3A580578&t=3peAz8A2n2Gw4WMl-1)
+
+## API
+
+| Prop                      | Type                                                       | Default     | Description                                                                    | Required |
+| ------------------------- | ---------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------ | -------- |
+| children                  | `number`                                                   | `undefined` | The value to be rendered within the component                                  | ✅       |
+| intent                    | `positive`, `negative`, `notice`, `information`, `blue`    | `undefined` | The variant of the amount to be rendered                                       |          |
```
divyanshu013, line 16: Is this `intent` or `variant`? We also have `blue`

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+| fontWeight                | `bold`, `regular`                                          | `bold`      | The weight of the amount to be rendered                                        |          |
+| suffix                    | `true`, `false`                                            | `false`     | Indicates whether a suffix should be used                                      |          |
```
divyanshu013, line 18: What about the `showCurrencySuffix` from earlier? Just `suffix` sounds vague.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+import { Amount } from '@razorpay/components';
+
+<Amount
+  size="2xlarge"
+  isSuffixPrefixHighlighted={true}
+  fontWeight="regular"
+  suffix={false}
+  intent="information"
+>
+  1000.27
```
divyanshu013, line 34: Will have to be `{1000.27}` with proposed API. Added a comment to consider a `value` prop.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+| children                  | `number`                                                   | `undefined` | The value to be rendered within the component                                  | ✅       |
+| intent                    | `positive`, `negative`, `notice`, `information`, `blue`    | `undefined` | The variant of the amount to be rendered                                       |          |
+| fontWeight                | `bold`, `regular`                                          | `bold`      | The weight of the amount to be rendered                                        |          |
+| suffix                    | `true`, `false`                                            | `false`     | Indicates whether a suffix should be used                                      |          |
+| size                      | `3xlarge`, `2xlarge`, `xlarge`, `large`, `medium`, `small` | `3xlarge`   | The size of the amount to be rendered                                          |          |
+| isSuffixPrefixHighlighted | `true`, `false`                                            | `true`      | Highlight the main amount by making the prefix symbol and decimal digits small |          |
+
```
divyanshu013, line 21: Don't see the APIs for decimals 🤔

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: @divyanshu013 I think we are settling at using `hasSuffix` as the only prop for both decimals and humanize. Looking at design we will not be supporting both decimal and suffix at the same time.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: @divyanshu013 @kamleshchandnani suggested to use `isSuffixPrefixHighlighted` instead.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: That does makes sense and it `isHighlighted` is not too verbose, @kamleshchandnani @chaitanyadeorukhkar @divyanshu013 final call.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,67 @@
...
+## Design
+
+- [Figma - Amount](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=28012%3A580578&t=3peAz8A2n2Gw4WMl-1)
+
+## API
+
+| Prop                      | Type                                                       | Default     | Description                                                                    | Required |
+| ------------------------- | ---------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------ | -------- |
+| children                  | `number`                                                   | `undefined` | The value to be rendered within the component                                  | ✅       |
```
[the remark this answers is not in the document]
Priyanshu-C: I was thinking of following a similar approach as other components and taking the value as a string in children and then manipulating it accordingly. We can simply check it if the string is a valid number.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: changed it to `hasSuffix`.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: I think we should use the same approach as other components.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: `hasSuffix` will have three states `Decimals`, `Humanize` and `None`. Please check figma once.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: `variant` is more apt here.

packages/blade/src/components/Amount/_decisions/decisions.md
kkyusufk: Just leaving my thoughts here. 
Dashboard is currently preparing for launching in Malaysia. I am not sure about the timelines but I do see a big chunk of code already being merged to prepare for it, so the use-case might arrive sooner. Perhaps we can also support Malaysia along with INR for now.

for reference this is one of the PR's which was merged couple of months back
https://github.com/razorpay/dashboard/pull/11454/files

packages/blade/src/components/Amount/_decisions/decisions.md
chaitanyadeorukhkar: How about `isAffixHighlighted`?

Prefixes and Suffixes are types of affixes. This way we're not being too vague about what it means too.

packages/blade/src/components/Amount/_decisions/decisions.md
chaitanyadeorukhkar: Agreed. We should start thinking about currency support from now. We can limit it to just INR for now and add more later

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: Sure, this works.

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: Can we use `showSuffix`? Generally we're following `show*` terminology in other components.

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: But in API decision table I only see `true` and `false`

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: Linking to my reasonings [above](https://github.com/razorpay/blade/pull/993#discussion_r1099986273). I feel the same approach wouldn't work here as well because this component is specifically designed to work with numbers and currencies.

packages/blade/src/components/Amount/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: Added a comment [below](https://github.com/razorpay/blade/pull/993#discussion_r1108115833)

packages/blade/src/components/Amount/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: Let's do some exploration on this too please.

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: Added a comment [below](https://github.com/razorpay/blade/pull/993#discussion_r1108096596)

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: Opps, my bad. This got updated later, I did not update the prop values. It's updated now.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: Sure, let's do that.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: cool, done.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,73 @@
...
+| variant                   | `positive`, `negative`, `notice`, `information`, `blue`    | `undefined` | The variant of the amount to be rendered                                                                                                                                                          |          |
+| fontWeight                | `bold`, `regular`                                          | `bold`      | The weight of the amount to be rendered                                                                                                                                                           |          |
+| showSuffix                | `none`, `decimals` , `humanise`                            | `decimals`  | The type of affix to be displayed with the currency value, (none, decimals, and humanise) that respectively indicate no affix, the number of decimal places, and the scale of the currency value. |          |
```
divyanshu013, line 18: Since this isn't a boolean anymore we can avoid the `show` prefix and just make the prop `suffix` perhaps:

```ts
suffix: 'none' | 'decimals' | 'humanise'
```

One thing to confirm whether to opt for `humanise` (british) or `humanize` (american) 😸

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,73 @@
...
+| size                      | `3xlarge`, `2xlarge`, `xlarge`, `large`, `medium`, `small` | `3xlarge`   | The size of the amount to be rendered                                                                                                                                                             |          |
+| isSuffixPrefixHighlighted | `true`, `false`                                            | `true`      | Highlight the main amount by making the prefix symbol and decimal digits small                                                                                                                    |          |
```
divyanshu013, line 20: Are we not opting for affix? /cc @chaitanyadeorukhkar

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Sample Usage
+
+```jsx
+import { Amount } from '@razorpay/components';
+
+<Amount
+  showSuffix="decimals"
```
divyanshu013, line 28: To be updated based on prop name change.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,73 @@
...
+  size="2xlarge"
+  isSuffixPrefixHighlighted={true}
+  fontWeight="regular"
+  variant="information"
+  value={1000}
+/>;
+```
+
+## Open Questions
```
divyanshu013, line 37: Let's also add the final resolutions for the open questions now.

divyanshu013 (commented):
Minor comments. Let's also post on Slack for wider review.

packages/blade/src/components/Amount/_decisions/decisions.md
chaitanyadeorukhkar: +1

I'd vote for `humanize`

packages/blade/src/components/Amount/_decisions/decisions.md
chaitanyadeorukhkar, line 20: We need to changes this to:
```suggestion
| isAffixHighlighted | `true`, `false`                                            | `true`      | Highlight the main amount by making the prefix symbol and decimal digits small                                                                                                                    |          |
```

Priyanshu-C: This is still open for discussion, not sure what we are finalizing.

Priyanshu-C: @chaitanyadeorukhkar when this value will be true, the affix won't be highlighted, it is the value which will get highlighted.

Priyanshu-C: Yes, it is incorrect and open to be replaced with a better name.

Priyanshu-C: **isHighlighted=true** 
image
**isHighlighted=false** 
image

chaitanyadeorukhkar: Can it be like this:

`isAffixHighlighted=true`


`isAffixHighlighted=false`

divyanshu013: This is really taking a test of my English grammar 😅 

How about `isAffixDeemphasized`?

chaitanyadeorukhkar: Why go in the negation though? `isAffixHighlighted` would still work

divyanshu013: Because it isn't actually "highlighted" (in terms of the language), it matches the other content (the number part - both have same size, weight etc.) 😅

chaitanyadeorukhkar: Yeah, Makes sense. How about `isAffixSubtle`? Seems a bit simpler to write than `isAffixDeemphasized`

saurabhdaware: Should we call it `isCurrencyHighlighted` instead? I understand that it also changes emphasis on decimal number but maybe just to keep the name simple?

TIL `Affix` 🙈

saurabhdaware (commented):
Small comment. Rest looks good to me. Great work 👯

packages/blade/src/components/Amount/_decisions/decisions.md
divyanshu013: Actually currency isn't "highlighted" per se, just the other things are shrinked and faded 😅 

So far these two seems like best options to me:
- `isAffixDeemphasized`
- `isAffixSubtle`

Priyanshu-C: let's go with this `isAffixSubtle` ?

saurabhdaware: +1 to `isAffixSubtle`

divyanshu013:
@Priyanshu-C right now the view formatted doc link in original comment is pointing to a commit, would be good to instead point to the branch so it updates for latest commits.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,75 @@
...
+
+Amount component is a UI element that displays and formats various currency values in a specified format.
+
+<img  src="./amount-thumbnail.png"  width="380" />
+
+## Design
+
+- [Figma - Amount](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=28012%3A580578&t=3peAz8A2n2Gw4WMl-1)
+
+## API
```
divyanshu013, line 11: Let's also get the props synced to figma. Could you please followup with @saurav12 and ensure these are updated on the component figma?

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,75 @@
...
+  suffix="decimals"
+  size="2xlarge"
+  isAffixSubtle={true}
+  fontWeight="regular"
+  variant="information"
+  value={1000}
+/>;
+```
+
+## Open Questions
```
divyanshu013, line 37: Let's also add resolutions to the open questions alongside.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,75 @@
...
+
+- Should we call `isSuffixPrefixHighlighted` or `isHighlighted` or `isAffixHighligted`, considering when this value is true the value will be highligted and not the affix.
+
+#### Dev
+
+- Should we call it `variant` or `intent`, `isStyled` or something more apt?
+- Should we separate the amount with commas based on the Indian currency format?
+- What should be the default `intent`? Kept it as `default` for now, but `neutral` seems more apt.
+- We currently use amount value in paise in backend, so how should we convert it to rupee here?
```
divyanshu013, line 46: How are these values generally sent? I thought it might be in rupees for example `0.01` INR?

divyanshu013 (commented):

packages/blade/src/components/Amount/_decisions/decisions.md
chaitanyadeorukhkar: Lets fix `isAffixSubtle` then 👍

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,75 @@
...
+| ------------- | ---------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| value         | `number`                                                   | `undefined` | The value to be rendered within the component                                                                                                                                                     | ✅       |
+| variant       | `positive`, `negative`, `notice`, `information`, `blue`    | `undefined` | The variant of the amount to be rendered                                                                                                                                                          |          |
```
kamleshchandnani, line 16: do we have blue color for Amount? can you check on designer once? have mentioned you in a comment on Figma

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,75 @@
...
+| fontWeight    | `bold`, `regular`                                          | `bold`      | The weight of the amount to be rendered                                                                                                                                                           |          |
```
kamleshchandnani, line 17:
```suggestion
| weight    | `bold`, `regular`                                          | `bold`      | The weight of the amount to be rendered                                                                                                                                                           |          |
```

we have been using weight across Blade for fontWeight

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,75 @@
...
+| suffix        | `none`, `decimals` , `humanize`                            | `decimals`  | The type of affix to be displayed with the currency value, (none, decimals, and humanize) that respectively indicate no affix, the number of decimal places, and the scale of the currency value. |          |
```
kamleshchandnani, line 18:
```suggestion
| suffix        | `none`, `decimals`, `humanize`                            | `decimals`  | The type of suffix to be displayed with the currency value, (none, decimals, and humanize) that respectively indicate no suffix, the number of decimal places, and the scale of the currency value. |          |
```
affix includes suffix and prefix both

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,75 @@
...
+| size          | `3xlarge`, `2xlarge`, `xlarge`, `large`, `medium`, `small` | `3xlarge`   | The size of the amount to be rendered                                                                                                                                                             |          |
+| isAffixSubtle | `true`, `false`                                            | `true`      | Highlight the main amount by making the prefix symbol and decimal digits small and faded                                                                                                          |          |
```
kamleshchandnani, line 20:
```suggestion
| isAffixSubtle | `true`, `false`                                            | `true`      | Makes the prefix symbol and decimal digits small and faded                                                                                                          |          |
```

Highlighted and faded sounds contradicting

kamleshchandnani: also, suffix is not changed when the suffix type is humanise the color remains the same

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,75 @@
...
+  isAffixSubtle={true}
+  fontWeight="regular"
+  variant="information"
+  value={1000}
+/>;
+```
+
+## Open Questions
+
+- Should we call `isSuffixPrefixHighlighted` or `isHighlighted` or `isAffixHighligted`, considering when this value is true the value will be highligted and not the affix.
```
kamleshchandnani, line 39: `isAffixSubtle` is what we agreed upon so lets add that conclusion here?

packages/blade/src/components/Amount/_decisions/decisions.md
kamleshchandnani: is this fixed on design @Priyanshu-C? what is the conclusion?

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,75 @@
...
+
+#### Dev
```
kamleshchandnani, line 41: @Priyanshu-C lets add all the answers to the questions based on the discussions inline so we can then focus on closing the remaining ones?

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: Currently in X, they use it in rupee and the value transformation is handled by the consumer.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: I am not sure about the requirement, I thought we will add it the same way we did for Badge.

Priyanshu-C:
@kamleshchandnani @chaitanyadeorukhkar After discussion with @saurav12, there were few changes - 

- Instead of having a single prop as suffix, we will have two props `hasDecimals` and `Humanize`. As there will be a case where values in humanize will require to be rendered with decimals. eg - 1234.56 will be rendered as ₹1.2k in case of humanize being true. 
- We are removing `blue` as a variant as it is not required, but will keep the prop name as variant as more variants will be added based on requirement.

divyanshu013:
Please update the props table too for this change

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,92 @@
...
+| ------------- | ---------------------------------------------------------------------------------------------- | ----------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| value         | `number`                                                                                       | `undefined`       | The value to be rendered within the component                                                                                                                                                       | ✅       |
+| intent        | `positive`, `negative`, `notice`, `information`                                                | `undefined`       | The intent of the amount to be rendered                                                                                                                                                             |
```
chaitanyadeorukhkar, line 16: There's also `neutral` right?

chaitanyadeorukhkar (approved):

chaitanyadeorukhkar (dismissed):

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,92 @@
...
+| intent        | `positive`, `negative`, `notice`, `information` ,`neutral`                                                                                                                        | `undefined`       | The intent of the amount to be rendered                                                                                                                                                             |
+| prefix        | `currency-symbol`, `currency-code`                                                                                                                                                | `currency-symbol` | The type of prefix to be shown                                                                                                                                                                      |          |
```
kamleshchandnani, line 17: the values shall be title case? `currencyCode`, `currencySymbol`?

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,92 @@
...
+| value         | `number`                                                                                                                                                                          | `undefined`       | The value to be rendered within the component                                                                                                                                                       | ✅       |
+| intent        | `positive`, `negative`, `notice`, `information` ,`neutral`                                                                                                                        | `undefined`       | The intent of the amount to be rendered                                                                                                                                                             |
```
kamleshchandnani, line 16: i don't see `neutral` on figma. but i see `none` . can you once confirm @chaitanyadeorukhkar @Priyanshu-C what is finalised?

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: I check List decisions, we have used 'ordered-filled'. I was thinking to use the same kebab case.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,92 @@
...
+
+<img  src="./amount-thumbnail.png"  width="380" />
+
+## Design
+
+- [Figma - Amount](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=28012%3A580578&t=3peAz8A2n2Gw4WMl-1)
+
+## API
+
+| Prop          | Type                                                                                                                                                                              | Default           | Description                                                                                                                                                                                         | Required |
```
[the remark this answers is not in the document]
Priyanshu-C: Currency, i am hiding this prop, although implementation for both INR and MYR have been done. We can enable it later.

packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,92 @@
...
+- We initially considered using two different props, but then realized that only one of them would be shown at a time. So in that case, we decided to use a single prop called 'suffix'. We ultimately settled on using a prop with three possible values: 'none', 'decimals', and 'humanize'.
+
+### Should we call `weight` or `fontWeight`?
+
+- We decided to call it `weight` as this represents the weight of the amount component with the affix changing accordingly. This also follows the same convention as other components.
+- Eventually we removed this prop as size prop will give weight too
+
+### Should we separate the amount with commas based on the Indian currency format?
+
+- For now, we will be supporting only INR. Going forward we will add more currencies. One of the major problems right now is we use ₹ and INR interchangeably. But in Malaysia we only have RM and there is no symbol associated with it. Hence we are staying away from this. Will solve this once the use-case arrives.
```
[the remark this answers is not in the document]
Priyanshu-C: Using the same, discussion was not updated.

packages/blade/src/components/Amount/_decisions/decisions.md
Priyanshu-C: Removed.

packages/blade/src/components/Amount/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: we can expose the prop with default value of INR?

kamleshchandnani (approved):

PR #994, opened 2023-02-07
fix(Dropdown): ignore flicker in Select when clicking items

### LINKED ISSUE #989: Flicker on Dropdown Selection
**Reproduction**

- When you select an item in SelectInput, the focus animation re-runs which looks a bit strange.

(SelectInput and Dropdown are not yet merged. You can check storybook preview of #988 until then)

saurabhdaware:
fixes #989 

- I passed `shouldIgnoreBlur` and `setShouldIgnoreBlur` to `BaseInput` so that we can ignore the blur event when we don't really want to call the blur out animation.
- ~I also removed the `triggererRef` from the Dropdown context because we only needed it to reput the focus on select which is not needed anymore as we just ignore the blur.~ We're only ignoring blur animation and not the blur itself so we still need to manage focus.

### DIFF
--- a/packages/blade/src/components/ActionList/ActionListItem.tsx
+++ b/packages/blade/src/components/ActionList/ActionListItem.tsx
@@ -206,6 +206,7 @@ const ActionListItem: WithComponentId<ActionListItemProps> = (props): JSX.Elemen
     onOptionClick,
     selectedIndices,
     setShouldIgnoreBlur,
+    setShouldIgnoreBlurAnimation,
     selectionType,
     dropdownTriggerer,
     isKeydownPressed,
@@ -258,6 +259,10 @@ const ActionListItem: WithComponentId<ActionListItemProps> = (props): JSX.Elemen
         {...metaAttribute(MetaConstants.Component, MetaConstants.ActionListItem)}
         onMouseDown={() => {
           setShouldIgnoreBlur(true);
+          setShouldIgnoreBlurAnimation(true);
+        }}
+        onMouseUp={() => {
+          setShouldIgnoreBlurAnimation(false);
         }}
         data-value={props.value}
         data-index={props._index}
--- a/packages/blade/src/components/ActionList/ActionListItem.tsx
+++ b/packages/blade/src/components/ActionList/ActionListItem.tsx
@@ -259,9 +259,12 @@ const ActionListItem: WithComponentId<ActionListItemProps> = (props): JSX.Elemen
         {...metaAttribute(MetaConstants.Component, MetaConstants.ActionListItem)}
         onMouseDown={() => {
           setShouldIgnoreBlur(true);
+          // We want to keep focus on Dropdown's trigger while option is being clicked
+          // So We set this flag that ignores the blur animation to avoid the flicker between focus out + focus in
           setShouldIgnoreBlurAnimation(true);
         }}
         onMouseUp={() => {
+          // (Contd from above comment...) We set this flag back to false since blur of SelectInput is done calling by this time
           setShouldIgnoreBlurAnimation(false);
         }}
         data-value={props.value}
--- a/packages/blade/src/components/ActionList/ActionListItem.tsx
+++ b/packages/blade/src/components/ActionList/ActionListItem.tsx
@@ -207,7 +207,6 @@ const ActionListItem: WithComponentId<ActionListItemProps> = (props): JSX.Elemen
     selectedIndices,
     setShouldIgnoreBlur,
     selectionType,
-    triggererRef,
     dropdownTriggerer,
     isKeydownPressed,
   } = useDropdown();
@@ -257,12 +256,6 @@ const ActionListItem: WithComponentId<ActionListItemProps> = (props): JSX.Elemen
           props.onClick?.({ name: props.value, value: isSelected });
         })}
         {...metaAttribute(MetaConstants.Component, MetaConstants.ActionListItem)}
-        onFocus={() => {
-          // We don't want to keep the browser's focus on option item. We move it to selectInput
-          if (!isReactNative()) {
-            triggererRef.current?.focus();
-          }
-        }}
         onMouseDown={() => {
           setShouldIgnoreBlur(true);
         }}
--- a/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
+++ b/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
@@ -88,11 +88,11 @@ export type BaseInputProps = FormInputLabelProps &
      */
     onBlur?: FormInputOnEvent;
     /**
-     * Ignores the next blur event
+     * Ignores the blur event animation (Used in Select to ignore blur animation when item in option is clicked)
      */
     shouldIgnoreBlurAnimation?: boolean;
     /**
-     * Used to set the shouldIgnoreBlurAnimation flag back to false after blur is ignored
+     * Used to set the shouldIgnoreBlurAnimation flag back to false after blur is ignored so we only ignore 1 event
      */
     setShouldIgnoreBlurAnimation?: (shouldIgnoreBlurAnimation: boolean) => void;
     /**
--- a/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
+++ b/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
@@ -88,6 +88,14 @@ export type BaseInputProps = FormInputLabelProps &
      */
     onBlur?: FormInputOnEvent;
     /**
+     * Ignores the next blur event
+     */
+    shouldIgnoreBlur?: boolean;
+    /**
+     * Used to set the shouldIgnoreBlur flag back to false after blur is ignored
+     */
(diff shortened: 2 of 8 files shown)


packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -181,6 +187,10 @@ export const StyledBaseInput = React.forwardRef<
           setCurrentInteraction('active');
         }}
         onBlur={(): void => {
```
kamleshchandnani, line 189: can we move the common props out in an object? its very difficult when you add/remove things from here as someone need to be very cautious to add/remove  from styled button as well as styled input

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
```diff
@@ -104,12 +104,18 @@ export const StyledBaseInput = React.forwardRef<
...
       hasPopup,
+      shouldIgnoreBlur,
+      setShouldIgnoreBlur,
       ...props
     },
     ref,
   ) => {
     const commonProps = {
       onBlur: (event: React.ChangeEvent<HTMLInputElement>): void => {
+        if (shouldIgnoreBlur && setShouldIgnoreBlur) {
```
kamleshchandnani, line 115: when do we set this as true?

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
saurabhdaware: I did that for web file. In native, there are just 2 props which are same. Rest need a bit of separate handling because one is input where we read `nativeEvent.text` for value and one is button where we read the inner text content for the value.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
saurabhdaware: In actionlistitem when someone clicks on item.

This is the same flag that we handle to ignore the dropdown's blur when click in somewhere inside the dropdown itself

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
saurabhdaware: moved these 2 props that are same

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -137,11 +137,23 @@ export const StyledBaseInput = React.forwardRef<
...
     const buttonValue = props.value ? props.value : props.defaultValue;
+    const commonProps = {
+      onBlur: (): void => {
+        if (shouldIgnoreBlurAnimation && setShouldIgnoreBlurAnimation) {
+          setShouldIgnoreBlurAnimation(false);
+          return;
+        }
+        setCurrentInteraction('default');
+      },
+      isFocused: currentInteraction === 'active',
```
divyanshu013, line 155: Suggestion: add some comments to explain the "why"

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -87,6 +87,14 @@ export type BaseInputProps = FormInputLabelProps &
      * For React Native this will call `onEndEditing` event since we want to get the last value of the input field
      */
     onBlur?: FormInputOnEvent;
+    /**
+     * Ignores the next blur event
+     */
+    shouldIgnoreBlurAnimation?: boolean;
```
divyanshu013, line 93: Is it only the "next" blur event? Perhaps the prop should be named appropriately.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -87,6 +87,14 @@ export type BaseInputProps = FormInputLabelProps &
...
+    /**
+     * Used to set the shouldIgnoreBlurAnimation flag back to false after blur is ignored
+     */
+    setShouldIgnoreBlurAnimation?: (shouldIgnoreBlurAnimation: boolean) => void;
```
divyanshu013, line 97: Can we contain the setter logic within `BaseInput` to ignore the blur animations based on a flag? Right now we're passing this flag from outside and also a setter for it.

divyanshu013 (commented):
Some doubts.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
saurabhdaware: so the prop itself doesn't handle next blur. We're passing `setShouldIgnoreBlurAnimation` and we set it to false so that makes it only ignore next blur. Lemme rephrase the jsdoc comment instead.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
saurabhdaware: we actually have to handle it on the outside because the flag setting logic happens when ActionListItem is clicked.

We're only passing prop here so that we can toggle the flag once one event is ignored.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
saurabhdaware: done

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
saurabhdaware: done

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -137,11 +137,25 @@ export const StyledBaseInput = React.forwardRef<
...
       ...props
     },
     ref,
   ) => {
     const buttonValue = props.value ? props.value : props.defaultValue;
+    const commonProps = {
+      onBlur: (): void => {
+        // In certain cases like SelectInput, we want to ignore the blur animation when option item is clicked.
+        // The selectinput should always look like it is in focus otherwise it triggers blur + focus again which can cause flicker
+        if (shouldIgnoreBlurAnimation && setShouldIgnoreBlurAnimation) {
```
kamleshchandnani, line 151: so if i pass `shouldIgnoreBlurAnimation={true}` then this will set it to false isn't weird? or am i not reading correctly?

saurabhdaware: yup it will. It will ignore the first blur animation and set it to false.

We would only want to ignore the animation once right? otherwise the animation will never happen

saurabhdaware: Actually you'll have to pass the state variable with it's set function. A bit hacky but alternative I can think of is to bring all the interaction logic outside BaseInput which won't make sense. 

We want to ignore the blur animation specifically when actionlistitem is clicked so that logic is outside of BaseInput. 

It's similar to how you would pass state control to child in React. You pass it's `value` and it's `setValue`.

kamleshchandnani: so wouldn't it be simpler if i just pass setter to the child components or just give the state to the child and then handle the setter internally? I'm just wondering that giving too many things might confuse the developer building on top of blade

saurabhdaware: Can you check this now f76bfbf09676e6c6d1fb9f1066b37a23a9a6d35b

So handling the setters inside BaseInput was complex because setting is primarily dependent on the ActionListItem clicks. One way was to have `shouldIgnoreNextBlur` prop and then ignore that prop internally if it is already ignored once but that too included lots of complex conditions. Also things like what do we mean by "next" cause if it is always true then all events become next.

I am now handling it outside BaseInput in different state and passing the state to BaseInput.

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -257,14 +257,12 @@ const ActionListItem: WithComponentId<ActionListItemProps> = (props): JSX.Elemen
...
-          if (!isReactNative()) {
-            triggererRef.current?.focus();
-          }
-        }}
         onMouseDown={() => {
           setShouldIgnoreBlur(true);
+          setShouldIgnoreBlurAnimation(true);
+        }}
+        onMouseUp={() => {
+          setShouldIgnoreBlurAnimation(false);
```
divyanshu013, line 265: We can add some comments on why to ignore blur animation on mouse down but not on mouse up

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: done

kamleshchandnani (approved):

PR #997, opened 2023-02-08
refactor(Box): rename Box to BaseBox

saurabhdaware:
As part of introducing Layout props, we'll be adding a new Box component that we expose to consumers.

To enable that change, I rename earlier Box to BaseBox.

**Changes**
- Move `Box` component inside `Box/BaseBox` directory
- Change imports
- Rename `Box` variable to `BaseBox` everywhere
- Update snapshots 

**Find and Replace queries I performed**
- Find: `<Box`, Replace `<BaseBox`
- Find: `</Box>`, Replace `</BaseBox>`
- Import URLs were auto changed by VSCode after changing directory
- Find: `import Box`, Replace `import BaseBox`
- Find: `styled(Box)`, Replace `styled(BaseBox)`
- Find: `BoxProps`, Replace `BaseBoxProps`
- Did `yarn typecheck` and fixed some tiny messups
- Update snapshots

**TODO**
- [x] Pull Dropdown changes 
- [x] Pull List changes


divyanshu013 (dismissed):
Looks good to me. Can you please also validate the react native stories locally?

chaitanyadeorukhkar (approved):
LGTM, but had a few questions

- Can we uses the new Box (Layout) component that you will create across our codebase?
- What would be the differences between BaseBox and new Box?
- What strategy can we take to migrate all BaseBox instances to the new Box?

saurabhdaware:
> Can we uses the new Box (Layout) component that you will create across our codebase?

We can in some places. I am not planning to do that right now. We can do this once everything is properly shipped in layout

> What would be the differences between BaseBox and new Box?

I have tried to minimize the differences (like I am removing number type from spacing in all, Box and BaseBox and moving all spacing values to string in #1009).

Eventually, few props that exist in BaseBox may not exist in Box. E.g. `backgroundColor`, `borderRadius`, etc. In those cases we will have to continue using BaseBox.

> What strategy can we take to migrate all BaseBox instances to the new Box?

We can find and replace BaseBox with Box and then run `yarn typecheck`. Then we can revert back to BaseBox wherever it is failing.

saurabhdaware:
> Looks good to me. Can you please also validate the react native stories locally?

Did. They seem to work. (some ref stories are breaking in master branch too in React Native. Will create an issue for them)

divyanshu013:
I think Box can be a lightweight wrapper over BaseBox from where we can control which all properties are exposed to public.

PR #1002, opened 2023-02-10
chore: added bottom sheet api decisions

anuraghazra:
### [View Formatted](https://github.com/razorpay/blade/blob/anu/bottomsheet-api/packages/blade/src/components/BottomSheet/_decisions/decisions.md)


packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+### `BottomSheet`
+
+We'll expose an `BottomSheeet` component with the following API:
+
+| Prop            | Type        | Default                 | Description                                                                                     | Required |
+| --------------- | ----------- | ----------------------- | ----------------------------------------------------------------------------------------------- | -------- |
+| open            | `boolean`   | `false`                 | toggles bottom sheet state content                                                              |          |
+| snapPoints      | `string[]`  | `['25%', '50%', '85%']` | snappoints in which the bottom sheeet will rest on                                              |          |
+| onDismiss       | `Callback`  | `undefined`             | called when the bottom sheet is closed, either by user state, hitting `esc` or tapping backdrop |          |
+| initialFocusRef | `React.Ref` | `undefined`             | ref element you want to get keyboard focus when opening the sheet                               |          |
```
divyanshu013, line 72: What would this be used for?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+
```
divyanshu013, line 73: Is there any controls for the close button?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+| Prop       | Type              | Default     | Description                 | Required |
+| ---------- | ----------------- | ----------- | --------------------------- | -------- |
+| `children` | `React.ReactNode` | `undefined` | Contents of the BottomSheet | ✅       |
+
+### `BottomSheetHeaderLeading` API
+
+| Prop     | Type            | Default     | Description                          | Required |
+| -------- | --------------- | ----------- | ------------------------------------ | -------- |
+| `title`  | `string`        | `undefined` | Title of the Card                    | ✅       |
```
divyanshu013, line 84: Card copy pasta 😸

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+| Prop     | Type            | Default     | Description                          | Required |
+| -------- | --------------- | ----------- | ------------------------------------ | -------- |
+| `title`  | `string`        | `undefined` | Title of the BottomSheet footer      | ✅       |
+| `prefix` | `IconComponent` | `undefined` | Prefix icon placed before title text |          |
+
+### `BottomSheetFooterTrailing` API
+
+| Prop      | Type                                     | Default     | Description                                | Required |
+| --------- | ---------------------------------------- | ----------- | ------------------------------------------ | -------- |
+| `actions` | `{ primary: Action, secondary: Action }` | `undefined` | Renders a primary/secondary action buttons |          |
```
divyanshu013, line 104: Are both primary and secondary required?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+
+```ts
+type Action = {
+  onClick: () => void;
+  text: string;
+  type?: 'button' | 'reset' | 'submit';
```
divyanshu013, line 110: What would be purpose of reset and submit?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+      <ActionListItem leading={<ActionListItemIcon icon={InfoIcon} />} title="Info" value="info" />
+    </>
+  );
+};
+
+const App = () => {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <Dropdown selectionType={selectionType}>
```
divyanshu013, line 173: What is `selectionType`

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+
+- Easy implementation for user's end
+- User's don't have to think about breakpoints, conditional rendering etc
+- From blade side, we will have greater control over the pattern
+
+Cons:
+
+- Bundle size will be impacted, even if users are desktop they will get the bundle of BottomSheet (vice versa)
+
+2. We expose BottomSheet indepandantly and let users lazy load the component as needed
```
divyanshu013, line 137: Why can't we lazy load the component on Blade side? Let's add that.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+- Not trivial to implement from user's end, they will have to compose the BottomSheet & Select as per their needs.
+
+
+Considering the bundle size downside to approach 1, we decided to go ahead with approach 2.
+
+### Composition Example
+
+```jsx
+const BottomSheet = React.lazy();
+const DropdownOverlay = React.lazy();
```
divyanshu013, line 154: Since these are being lazy loaded, would we need a `Suspense` boundary in `Dropdown`?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+  );
+};
+
+const App = () => {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <Dropdown selectionType={selectionType}>
+      <SelectInput label="Select Action" onClick={() => setOpen(true)} />
+      {isMobile ? (
```
divyanshu013, line 175: Where is the `isMobile` coming from?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+
+- Tab, Shift+Tab should cycle the focused elements
+- The focus should be trapped inside the bottom sheet
+- Escape closes the bottom sheet, after closing the focus will go to the triggered element
+- When opening the bottom sheet it should either focus on the first focusable element on the sheet or element provided by the user via `initialFocusRef` prop
+
+## Open questions
+
+1. What is `action` in the header trailing visual?
+2. In BottomSheetHeaderLeaing & BottomSheetFooterLeading will the `prefix` only support Icon component?
```
divyanshu013, line 219: Are these resolved? Let's add the resolutions.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+  type?: 'button' | 'reset' | 'submit';
+  accessibilityLabel: string;
+  isLoading: boolean;
+  isDisabled: boolean;
+  icon: React.ReactNode;
+  iconPosition: 'left' | 'right';
+};
+```
+
+## Composition with DropDown
```
saurabhdaware, line 119:
```suggestion
## Composition with Dropdown
```

🙈

packages/blade/src/components/BottomSheet/_decisions/decisions.md
saurabhdaware: selectionType="single" or selectionType="multiple". It decides whether its single select or multi select.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+          <BottomSheetBody>
+            <SelectContent />
+          </BottomSheetBody>
+          <BottomSheet>
+            <BottomSheetLeading title="Footer Title" prefix={icon} />
+            <BottomSheetTrailing
+              actions={{
+                primaryAction: { text: 'Confirm' },
+                secondaryAction: { text: 'Close' },
+              }}
```
saurabhdaware, line 189: Can we do something like this?

```jsx
<ResponsiveMenuOverlay>
  <BottomSheetHeader /> // we hide this desktop
  <ActionList>
    <ActionListItem />
  </ActionList>
  <BottomSheetFooter /> // hide this in desktop
</ResponsiveMenuOverlay>
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+    <Dropdown selectionType={selectionType}>
+      <SelectInput label="Select Action" onClick={() => setOpen(true)} />
+      {isMobile ? (
+        <BottomSheet open={open} onDismiss={() => setOpen(false)}>
+          <BottomSheetHeader>
+            <BottomSheetHeaderLeading title="Payments Links" prefix={PayIcon} />
+          </BottomSheetHeader>
+          <BottomSheetBody>
+            <SelectContent />
+          </BottomSheetBody>
```
saurabhdaware, line 182: Can we use ActionList component instead of BottomSheetBody?

```jsx
<ActionList>
  <SelectContent />
</ActionList>
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+Sample usage:
+
+```jsx
+import { BottomSheet } from '@razorpay/blade';
+
+<BottomSheet open={boolean} snapPoints={[]}>
+  <BottomSheetHeader>
+    <BottomSheetHeaderLeading title="Payments Links" prefix={BladeIcon} />
+    <BottomSheetHeaderTrailing visual={Link | Action | Icon} />
+  </BottomSheetHeader>
```
saurabhdaware, line 24: Can we make this similar to ActionListHeader?
```jsx
<BottomSheetHeader 
  title="Payments Links" 
  leading={<BottomSheetHeaderIcon icon={BladeIcon}} 
  trailing={<BottomSheetHeaderIcon icon={Icon} /> | <BottomSheetHeaderLink /> | <BottomSheetHeaderButton />}
/>
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+  <BottomSheetBody>Body Content</BottomSheetBody>
+  <BottomSheet>
+    <BottomSheetLeading title="Footer Title" prefix={icon} />
+    <BottomSheetTrailing
+      actions={{
+        primaryAction: { text: 'Know more', onClick: () => {} },
+        secondaryAction: { text: 'Read Docs', onClick: () => {} },
```
saurabhdaware, line 31: Same for the footer. Can we make it similar to `<ActionListHeader />`'s API?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+    </>
+  );
+};
+
+const App = () => {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <Dropdown selectionType={selectionType}>
+      <SelectInput label="Select Action" onClick={() => setOpen(true)} />
```
saurabhdaware, line 174: this `onClick` and handling isOpen on consumer end might seem a bit inconsistent with Dropdown no?

Also consumers will have to set something like `aria-expanded` right?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
saurabhdaware: And basically somehow do lazy-loading on our end 🤔 ?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: We need to keep these props on the bottomsheet since it should be an independent component on it's own, the integration with the ActionList we need to hack around a bit, we can internally handle these props while wrapped inside Dropdown. 

I'll need to figure out the integration with the Dropdown functionality, might also need to restructure the Dropdown a bit since right now it's tied to the ActionList & SelectInput logic.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Yes, but functionality might [not](https://github.com/razorpay/blade/pull/1002#discussion_r1112791644) work. I need to check this. Since it's tied to the dropdown a bit. 

Also the <ActionList \/>'s root styling also not needed in the bottomsheet. (it has some padding around it so I might need to create a BaseActionList for functionality)

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: from userland, users can assert on our breakpoints. or we can also expose a util for isMobile.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+  - Trailing
+    - Primary action button
+    - Secondary action button
+
+- The BottomSheetFooter will behave similar to Card's footer.
+- The Header, Footer can be individually omitted
+
+### `BottomSheet`
+
+We'll expose an `BottomSheeet` component with the following API:
```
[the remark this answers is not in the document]
anuraghazra: Sheeeesshhh!

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+
+[APG Guidelines for Modal](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)
+
+Behaviours:
+
+- Tab, Shift+Tab should cycle the focused elements
+- The focus should be trapped inside the bottom sheet
+- Escape closes the bottom sheet, after closing the focus will go to the triggered element
+- When opening the bottom sheet it should either focus on the first focusable element on the sheet or element provided by the user via `initialFocusRef` prop
+
```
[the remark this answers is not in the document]
anuraghazra: Good catch, I'll double check this.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
divyanshu013: Let's add in the code sample too

packages/blade/src/components/BottomSheet/_decisions/decisions.md
divyanshu013: Let's add a sample usage in code example, right now it seems to come magically in between 😸

packages/blade/src/components/BottomSheet/_decisions/decisions.md
saurabhdaware: Yup. Let me know if you need any help understanding that code. We can pair up and figure out.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+      actions={{
+        primaryAction: { text: 'Know more', onClick: () => {} },
+        secondaryAction: { text: 'Read Docs', onClick: () => {} },
+      }}
+    />
+  </BottomSheet>
+</BottomSheet>;
+```
+
+**Design Constraints**
```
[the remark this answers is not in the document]
anuraghazra: will do.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+- The Header, Footer can be individually omitted
+
+### `BottomSheet`
+
+We'll expose an `BottomSheeet` component with the following API:
+
+| Prop            | Type        | Default                 | Description                                                                                     | Required |
+| --------------- | ----------- | ----------------------- | ----------------------------------------------------------------------------------------------- | -------- |
+| open            | `boolean`   | `false`                 | toggles bottom sheet state content                                                              |          |
+| snapPoints      | `string[]`  | `['25%', '50%', '85%']` | snappoints in which the bottom sheeet will rest on                                              |          |
```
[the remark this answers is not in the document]
anuraghazra: Yes we will limit it to max 3. will update this.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: BottomSheet's a11y is like a modal, we need to provide users with an option to focus on an element when they open the modal, for example they might choose to focus on a primary button first. 

Internally we can also automatically focus on the first "tabbable" dom element inside the BottomSheetBody (need to check the feasibility with ActionList)

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: No not sure it's needed because the "close" button will only have 1 action "closing the sheet", so instead if users want to react to or know the state of the sheet from outside they can use the onDismiss prop (which will get triggered when we click the close button)

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Nope, will update this. 

The actions will behave similar to the Card's footer actions. 

- Only two action buttons can be present primary or secondary
- Users can totally omit action buttons
- They can also use either variants of buttons primary or secondary
- If 1 action is present the button will become fullwidth

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Just accepting the same props as the Button. 

Tho might be useful if someone uses the bottomsheet as form submission modal.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: > Let's add in the code sample too

Tho It's the same there will be no change in the BottomSheet's API because of it, i think selectionType is well documented in the Dropdown's API Decision.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Will do. In stories I'll also add a full fledged example.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: > And basically somehow do lazy-loading on our end 🤔 ?

Not feasible to do lazy loading our end as mentioned in the composition section.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Wanted to keep the API similar to Card since the BottomSheet is basically a "fancy" modal. and the anatomy of it is very similar, eg: the footer/footerActions are 1:1 with the API & functionality with Card.

But we can discuss this with kamlesh once.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Updated.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Updated the examples with Suspense boundary.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Added that.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,225 @@
...
+- The BottomSheetFooter will behave similar to Card's footer.
+- The Header, Footer can be individually omitted
+
+### `BottomSheet`
+
+We'll expose an `BottomSheeet` component with the following API:
+
+| Prop            | Type        | Default                 | Description                                                                                     | Required |
+| --------------- | ----------- | ----------------------- | ----------------------------------------------------------------------------------------------- | -------- |
+| open            | `boolean`   | `false`                 | toggles bottom sheet state content                                                              |          |
```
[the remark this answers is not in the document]
anuraghazra: Yes that `open` prop is a boolean that you pass `<BottomSheet open={true} />`

Do you mean to rename it to `isOpen`?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Checked this in implementation, doesn't overlap.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Added

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: @saurav12 can you check these? 

1. What is `action` in the header trailing visual?

image

2. In BottomSheetHeaderLeading & BottomSheetFooterLeading will the `prefix` only support Icon component?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
chaitanyadeorukhkar: Why is ActionListHeader and CardHeader different in terms of leading and trailing? Do we have a strong rationale?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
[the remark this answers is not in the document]
chaitanyadeorukhkar: +1

boolean values lets append them with `is`

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,276 @@
...
+| snapPoints      | `[number, number, number]`  | `[0.35, 0.5, 0.85]` | Snappoints in which the bottom sheeet will rest on, this accepts a number between 0 & 1 which maps to the total view height of the screen. 0.5 means 50% of screen height.                                              |          |
+| onDismiss       | `Callback`  | `undefined`             | called when the bottom sheet is closed, either by user state, hitting `esc` or tapping backdrop |          |
+| initialFocusRef | `React.Ref` | `undefined`             | ref element you want to get keyboard focus when opening the sheet                               |          |
```
chaitanyadeorukhkar, line 104: Does this mean that if I pass a ref of a TextInput, it will get keyboard focus and I can type directly? Would it also mean that the ref we receive from the user needs to be focusable?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,276 @@
...
+- No uneccesary bundle size impact for any of the platforms
+
+Cons:
+
+- Not trivial to implement from user's end, they will have to compose the BottomSheet & Select as per their needs.
+
+
+Considering the bundle size downside to approach 1, we decided to go ahead with approach 2.
+
+### Composition Example
```
chaitanyadeorukhkar, line 182: This seems very tricky. The DX with approach 2 is quite confusing. How bad of an impact to the bundle size will we have if we do add bottom sheet in it as well?  If the impact isn't that heavy, we should really consider a better DX

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,276 @@
...
+
+
+1. It's generally not feasible to lazy load components from library side, We can't possibly know where & how a blade component can be used or imported in the user's codebase. Even if we know that it will also require changes in how we bundle blade, since right now blade distributes the bundle in single index.js chunks, introducing lazy loading will create more fragmented chunks. 
+
+2. There are multiple ways to lazy load react components and it depends on the app's architecture. For example, React.lazy only does client side lazy loading while loadable support SSR too, thus the descision needs to be handed over to the consumers to import & integrate the components as they fit on their stack.
+
+
+## Accessibility
+
+The bottom sheet will follow the accessibility charechtaristics of a Modal.
```
chaitanyadeorukhkar, line 255:
```suggestion
The bottom sheet will follow the accessibility characteristics of a Modal.
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: Right, and it should be required prop I guess?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
divyanshu013: Do we need to expose this functionality 🤔 

What could be a possible usecase?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
divyanshu013: So, will we always show the close button?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,276 @@
...
+const App = () => {
+  const { theme } = useTheme();
+  const { matchedDeviceType } = useBreakpoint({ breakpoints: theme.breakpoints });
+  const isMobile = matchedDeviceType === 'mobile';
+
+  return (
+    <Dropdown selectionType={selectionType}>
+      <SelectInput label="Select Action" />
+      {/* We can either put a fallback spinner or show skeleton loaders */}
+      <React.Suspense fallback={<Spinner />}>
```
divyanshu013, line 199: This will look pretty awkward if the bundle takes time to load 🥲

packages/blade/src/components/BottomSheet/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: > Right, and it should be required prop I guess?

Can't make it required since the bottom sheet needs work as both "independent" & a "composed" component. 

when I say Independent i mean: 

```jsx
const [userPassedIsOpen, setIsOpen] = React.useState(false);

<BottomSheet isOpen={userPassedIsOpen}>
  <BottomSheetBody>some random stuff</BottomSheetBody>
</BottomSheet>

```

and also composable with Dropdown: 

```jsx
<Dropdown>
  <SelectInput />

  <BottomSheet> // <--- notice how the isOpen state is managed by the dropdown
    <BottomSheetBody>some random stuff</BottomSheetBody>
  </BottomSheet>
</Dropdown>
  
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Yes if the bottomsheet renders the header, it will always have the close button.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: > Does this mean that if I pass a ref of a TextInput, it will get keyboard focus and I can type directly

Yes initially it will be focused, check this [ariakit example](https://ariakit.org/examples/dialog-react-router) the textarea is focused initially when the modal opens. 

> Would it also mean that the ref we receive from the user needs to be focusable?

Yes the element needs to be "any [tababble](https://allyjs.io/data-tables/focusable.html) element"

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Need to test it out, at minimum it will have the bundle of the Dropdown 11kb + whatever the bottomsheet adds on top. 

Now considering the bundle impact and the [UX consideration](https://github.com/razorpay/blade/pull/1002#discussion_r1118614517) we can decide whether it's worth going for approach 2 or not.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: > What could be a possible usecase?

check this [ariakit example](https://ariakit.org/examples/dialog-react-router) the textarea is focused initially when the modal opens.

This pattern is common in most modals too, for example go to the x dashboard and click "Add Contact" 
image

You'll see that when the modal opens the "Contact Name" field is focused. 

> Do we need to expose this functionality 🤔

Probably we can avoid exposing this and just take the first focusable element as "initial focus" but there are also certain UX rules which users won't be able to do which is outlined in the [wai-aria notes](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/): 
image

packages/blade/src/components/BottomSheet/_decisions/decisions.md
saurabhdaware: oh yup saw your comments on slack. Makes sense.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,276 @@
...
+        {isMobile ? (
+          <BottomSheet>
+            <BottomSheetHeader>
+              <BottomSheetHeaderLeading title="Payments Links" prefix={PayIcon} />
```
saurabhdaware, line 203: 1. **Bundle size doubts**
In this API, we're using `<Dropdown />` so it will add the Dropdown's bundle right? 🤔 The only component I see we're code-splitting is `<DropdownOverlay />` which should be less than 2kb since it is just that modal that opens. It doesn't really have any major logic.\
\
All the critical Dropdown logic is in Dropdown component inside `useDropdown` hook and utilities.


2. **Should we have it controlled?**
  Dropdown is uncontrolled and open-close logic of Dropdown is handled by Dropdown itself. So the `open={}` API seems a bit different to me compared to Dropdown. It's like the component on Desktop handles state internally but the similar component of mobile handles it externally.\
\
  Apart from that, you might have to add `aria-expanded` or `aria-controls` to the trigger right? if we give the state management to consumer I would assume the consumer has to handle that?

I wanted to suggest something like this so that we take care of above issues- 
```jsx
<BottomSheet>
  <SelectInput label="Select Action" />
  <BottomSheetContainer>
     <BottomSheetHeader />
     <ActionList />
     <BottomSheetFooter />
  </BottomSheetContainer>
</BottomSheet>
```

Then we'll also be able to do
```jsx
const ResponsiveDropdown = isMobile ? BottomSheet : Dropdown;
const ResponsiveDropdownOverlay = isMobile ? BottomSheetContainer : DropdownOverlay;
const ResponsiveBottomSheetHeader = isMobile ? BottomSheetHeader : React.Fragament;
const ResponsiveBottomSheetFooter = isMobile ? BottomSheetFooter : React.Fragment;

// ...

<ResponsiveDropdown>
   <SelectInput />
   <ResponsiveDropdownOverlay>
      <ResponsiveBottomSheetHeader />
      <ActionList />
      <ResponsiveBottomSheetFooter />
    </ResponsiveDropdownOverlay>
</ResponsiveDropdown>



```

what do you think?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
saurabhdaware: as far I remember we went with that API is CardHeader because even inside leading, it had prefix and suffix of it's own. This part -
image

So we had to go with `leading` as a prop, then there wasn't any way to introduce this prefix and suffix. right @anuraghazra 🤔 ?

BottomSheetHeader doesn't have that usecase no?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
chaitanyadeorukhkar: Got it. Do we have a way to enforce consumers to only pass tabbable element? Or will we handle that via documentation only?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
saurabhdaware: https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Dropdown/_decisions/decisions.md#components-as-jsx-props-instead-of-compound-components this was the reason to go with props in ActionList

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: > Or will we handle that via documentation only?

This should be fine. 

Tho even if they pass a non-tabbable element, internally we will call `focus` on the element, so if the focus method is not found or it's not a focusable element it will be a noop.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+
+### `BottomSheet`
+
+We'll expose a `BottomSheet` component with the following API:
+
+| Prop            | Type        | Default                 | Description                                                                                     | Required |
+| --------------- | ----------- | ----------------------- | ----------------------------------------------------------------------------------------------- | -------- |
+| isOpen            | `boolean`   | `false`                 | Toggles bottom sheet state                                                              |          |
+| snapPoints      | `[number, number, number]`  | `[0.35, 0.5, 0.85]` | Snappoints in which the bottom sheeet will rest on, this accepts a number between 0 & 1 which maps to the total view height of the screen. 0.5 means 50% of screen height.                                              |          |
+| onDismiss       | `Callback`  | `undefined`             | called when the bottom sheet is closed, either by user state, hitting `esc` or tapping backdrop |          |
```
kamleshchandnani, line 91: why not `onClose`? i'm coming from the perspective of reduced learning curve and naming common interactions in a similar fashion

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+- The Header, Footer can be individually omitted
+
+### `BottomSheet`
+
+We'll expose a `BottomSheet` component with the following API:
+
+| Prop            | Type        | Default                 | Description                                                                                     | Required |
+| --------------- | ----------- | ----------------------- | ----------------------------------------------------------------------------------------------- | -------- |
+| isOpen            | `boolean`   | `false`                 | Toggles bottom sheet state                                                              |          |
+| snapPoints      | `[number, number, number]`  | `[0.35, 0.5, 0.85]` | Snappoints in which the bottom sheeet will rest on, this accepts a number between 0 & 1 which maps to the total view height of the screen. 0.5 means 50% of screen height.                                              |          |
```
kamleshchandnani, line 90: when snappoints is not provided what is the behavior? what will be the height of the bottomsheet? also do i need to pass all the snap points? what if I only pass one snappoint? will the initial height of the bottomsheet will be equal to this number?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+
+| Prop       | Type              | Default     | Description                 | Required |
+| ---------- | ----------------- | ----------- | --------------------------- | -------- |
+| `children` | `React.ReactNode` | `undefined` | Contents of the BottomSheet | ✅       |
+
+### `BottomSheetHeader` API
+
+| Prop     | Type            | Default     | Description                          | Required |
+| -------- | --------------- | ----------- | ------------------------------------ | -------- |
+| `title`  | `string`        | `undefined` | Title of the Header                    | ✅       |
```
kamleshchandnani, line 104: are we sure we don't have a use case of `subtitle`?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+| `leading` | `IconComponent` | `undefined` | leading icon placed before title text |          |
```
kamleshchandnani, line 105: leading icon is enough? we don't have a use case for badge or anything else? did we check with Merchant Dashboard and SME team? Can you check with Abhishek?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+
+### `BottomSheetFooter` API
```
kamleshchandnani, line 107: just initial thoughts:

- seems unnecessarily constraint-based to me? 
- the fact that I need to use a trailing prop to render 2 full-width action buttons has a steep learning curve and is confusing because it doesn't match the layout visually. I wouldn't have assumed that to be in trailing
- next, the fact that leading is always an icon component is not very justifiable. it feels like its meant to be used with the title

what if we make footer flexible to any jsx element? this might even come up when we are generalising Header and Footer components across our components. Lets brainstorm this with design team

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+| `leading` | `IconComponent` | `undefined` | leading icon placed before title text |          |
+| `trailing` | `React.ReactNode` | `undefined` | Renders trailing content |          |
+
+## Composition with Dropdown
+
+We will export `BottomSheet` component separately as an independant component but generally the pattern will be to use it with SelectInput, where in mobile devices the Select's dropdown will be replaced by the BottomSheet.
+
+There are two approaches to doing it:
+
+#### 1. We coupled the BottomSheet & SelectInput tightly and internally conditionally switch the components
```
kamleshchandnani, line 121:
```suggestion
#### 1. We couple the BottomSheet & SelectInput tightly and internally conditionally switch the components  
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+
+Pros:
+
+- Easy implementation for user's end
+- User's don't have to think about breakpoints, conditional rendering etc
+- From blade side, we will have greater control over the pattern
+
+Cons:
+
+- Bundle size will be impacted, even if users are desktop they will get the bundle of BottomSheet (vice versa)
```
kamleshchandnani, line 131:
```suggestion
- Bundle size will be impacted. Products where users are desktop only or majority are on desktop they will get the bundle of BottomSheet (vice versa)
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+- BottomSheet's state will get coupled with SelectDropdown
```
kamleshchandnani, line 132: add examples like snappoints which are unrelated with dropdown

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+
+Pros:
+
+- No uneccesary bundle size impact for any of the platforms
+- BottomSheet is decoupled from the state and internal logic of SelectDropdown & can also be used independently
+- Reduced implementation complexity & [interdependency](https://www.notion.so/BottomSheet-API-Discussion-e2aa79cd45274ef280fdb998efa6b98f?pvs=4#630d5e93897b4ca0bf54f475f0cd7d6e) of state
+
+Cons:
+
+- Not trivial to implement from user's end, they will have to compose the BottomSheet & Select as per their needs.
```
kamleshchandnani, line 148:
```suggestion
- Not trivial to implement from user's end, they will have to compose the BottomSheet & Dropdown as per their needs.
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+
+
+Considering the bundle size downside to approach 1, we decided to go ahead with approach 2.
+
+
+> **API Rabbit Hole**
```
kamleshchandnani, line 154: can you just copy paste the contents in this doc itself maybe at the bottom so we don't have to move back and forth between GH and Notion

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+**Q.** Why can't we lazy load from blade side? 
+
+There are two major reasons: 
+
+
+1. It's generally not feasible to lazy load components from library side, We can't possibly know where & how a blade component can be used or imported in the user's codebase. Even if we know that it will also require changes in how we bundle blade, since right now blade distributes the bundle in single index.js chunks, introducing lazy loading will create more fragmented chunks. 
+
+2. There are multiple ways to lazy load react components and it depends on the app's architecture. For example, React.lazy only does client side lazy loading while loadable support SSR too, thus the descision needs to be handed over to the consumers to import & integrate the components as they fit on their stack.
```
kamleshchandnani, line 215: can you add a sample with both the ways to import lazily? put it under a detail(collapsible section) for people to check it out if required

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+- Tab, Shift+Tab should cycle the focused elements
+- The focus should be trapped inside the bottom sheet
+- Escape closes the bottom sheet, after closing the focus will go to the triggered element
+- When opening the bottom sheet it should either focus on the first focusable element on the sheet or element provided by the user via `initialFocusRef` prop
+
+## Open questions
+
+1. What is `action` in the header trailing visual?
+2. In BottomSheetHeaderLeaing & BottomSheetFooterLeading will the `prefix` only support Icon component?
+3. In BottomSheetBody will we only have ActionList or users can add any of their own components too?
```
kamleshchandnani, line 235: this we can't restrict for cases like where people expect input elements for OTP?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
kamleshchandnani: bumping this

packages/blade/src/components/Card/_decisions/decisions.md
```diff
@@ -587,4 +587,4 @@ import {
 ## Open Questions
 
 - In footer action buttons should `secondary` action button only exist if `primary` action button is present? - 
-  - **ANS:** Secondary, Primary can independently work
\ No newline at end of file
+  - **ANS:** Secondary, Primary can independently work
```
kamleshchandnani, line 590: why is this showing a diff?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
kamleshchandnani: imagine use cases for checkout

where they'll have one action button saying "Pay Now" towards the right/trailing and towards leading they might show things like "Pay Using XX32" card number(like we see in zomato)

packages/blade/src/components/BottomSheet/_decisions/bottom-sheet-thumbnail.png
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/BottomSheet/_decisions/bottomsheet-anatomy.png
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+
+Components:
+
+- BottomSheet
+  - BottomSheetHeader
+  - BottomSheetBody
+  - BottomSheetFooter
+
+## API
+
```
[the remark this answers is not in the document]
anuraghazra: We need a small discussion around this, wether to keep isOpen state controlled or uncontrolled, let's discuss this tommorow on standup along with the Saurabh's discussion about the Dropdown.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+Sample usage:
```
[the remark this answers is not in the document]
anuraghazra: Sorry which POC? I did the library exploration only & then the implementation which I demoed and you folks did user-testing.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+  />
+  <BottomSheetBody>
+    <Text>Body Content<Text>
+  </BottomSheetBody>
+  <BottomSheetFooter
+    title="Footer Title"
+    leading={BladeIcon}
+    trailing={
+      <>
+        BladeIcon | IconButton | Button
```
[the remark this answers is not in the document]
anuraghazra: tailing will basically have any JSX elements similar to ActionListFooter so Buttons, Icons etc. 

image

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+      </>
+    }
+  />
+</BottomSheet>;
+```
+
+**Design Constraints**
+
+- BottomSheetHeader:
+  - title
```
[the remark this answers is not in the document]
anuraghazra: Design contraint is just showing the contraints for each prop, like in leading: we will have BladeIcon only etc. 

> that title is required prop?
That information is on the prop table.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Fine with either one, though leaning towards dismiss because in Alert we have isDismissable & onDismiss.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Yeah design mein to nehi hain. We can get same feedback from @saurav12

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: > what if we make footer flexible to any jsx element?

Flexible hi to hain? the trailing can take any JSX element, the API is 1:1 with ActionListFooter. 

Are you saying make the whole footer accept any jsx?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: The lazy loading code will just be similar only. 

```
const BottomSheet = React.lazy(() => import('@razorpay/blade/components/bottomsheet'))

vs

const BottomSheet = loadable(() => import('@razorpay/blade/components/bottomsheet'))
```

So nothing much change in terms of code, but there will be changes in architecture of the app. But can add a small example with we have in the notion doc.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: This was old open question on the first draft, removed. we can have anything.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: :( just waaaaay to much content to copy paste and put it here, bohot zayada indirection ho jayga. So just put only few important final decisions.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+    - BladeIcon
+
+- BottomSheetFooter:
+  - title
+  - leading:
+    - BladeIcon
+  - trailing
+    - any jsx element
+
+- The BottomSheetFooter will behave similar to ActionListFooter.
```
chaitanyadeorukhkar, line 80: Can you elaborate on what we mean by "behave" here? Someone shouldn't need to understand ActionListFooter to understand the constraint here

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,245 @@
...
+- The Header, Footer can be individually omitted
+
+### `BottomSheet`
+
+We'll expose a `BottomSheet` component with the following API:
+
+| Prop            | Type        | Default                 | Description                                                                                     | Required |
+| --------------- | ----------- | ----------------------- | ----------------------------------------------------------------------------------------------- | -------- |
+| isOpen            | `boolean`   | `false`                 | Toggles bottom sheet state                                                              |          |
+| snapPoints      | `[number, number, number]`  | `[0.35, 0.5, 0.85]` | Snappoints in which the bottom sheeet will rest on, this accepts a number between 0 & 1 which maps to the total view height of the screen. 0.5 means 50% of screen height.                                              |          |
```
chaitanyadeorukhkar, line 90: Since we have multiple snapPoints do we need to give consumers the ability to decide which snapPoint will the initial snap point?

anuraghazra: That might need discussion and design inputs. In third party libraries i explored they provide defaultSnapPoint option.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Will remove this line, seems redundant.

anuraghazra: So basically BottomSheetFooter has similar API surface as ActionListFooter

packages/blade/src/components/BottomSheet/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: So right now it behaves like controlled. 

> What if someone passed isOpen as true and then drags it down to shut it
it will trigger the onDismiss function which then user will set the setIsOpen(false)

So the usage will be like: 

```
const [isOpen, setOpen] = React.useState(false);

<BottomSheet isOpen={isOpen} onDismiss={() => setIsOpen(false)} />
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Added

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,298 @@
...
+| `title`  | `string`        | `undefined` | Title of the Header                    | ✅       |
+| `leading` | `IconComponent` | `undefined` | leading icon placed before title text |          |
+
+### `BottomSheetFooter` API
+
+| Prop     | Type            | Default     | Description                          | Required |
+| -------- | --------------- | ----------- | ------------------------------------ | -------- |
+| `title`  | `string`        | `undefined` | Title of the BottomSheet footer      | ✅       |
+| `leading` | `IconComponent` | `undefined` | leading icon placed before title text |          |
+| `trailing` | `React.ReactNode` | `undefined` | Renders trailing content |          |
```
divyanshu013, line 112: How would we render such cases?

image

I suspect people might use trailing for rendering everything 😆

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Okay agree on that, maybe better to keep everything on our end only. I will copy paste the content into a different md file and link it.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,298 @@
...
+- The Header, Footer can be individually omitted
+
+### `BottomSheet`
+
+We'll expose a `BottomSheet` component with the following API:
+
+| Prop            | Type        | Default                 | Description                                                                                     | Required |
+| --------------- | ----------- | ----------------------- | ----------------------------------------------------------------------------------------------- | -------- |
+| isOpen            | `boolean`   | `false`                 | Toggles bottom sheet state                                                              |          |
+| snapPoints      | `[number, number, number]`  | `[0.35, 0.5, 0.85]` | Snappoints in which the bottom sheeet will rest on, this accepts a number between 0 & 1 which maps to the total view height of the screen. 0.5 means 50% of screen height.                                              |          |
```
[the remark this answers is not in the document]
anuraghazra: Yes has to be 3 snapping points. 😅 as showing in the types as tuple `[number, number, number]` will add a note about it.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: This is what is pending from design side, we need to think of a better unified API for all the header footers.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,298 @@
...
+  - leading:
+    - BladeIcon
+  - trailing
+    - any jsx element
+
+- The Header, Footer can be individually omitted
+
+### `BottomSheet`
+
+We'll expose a `BottomSheet` component with the following API:
```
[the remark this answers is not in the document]
anuraghazra: Nope, design removed it. We will only close it via the backdrop / dragging down. 

For screen reader users we will have a hidden button so users can focus on that and close.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: As discussed in the notion doc, from our end we won't do any code splitting if user wants it they can do it, though we won't recommend doing it because bundle size savings either way isn't that huge. Like in storybook examples and docs we will only recommend the way without lazy loading.

divyanshu013 (commented):
Can we also add the considerations for building a custom solution rather than using a custom library in this doc?

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Resolved.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Resolved.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,298 @@
...
+- BottomSheet
+  - BottomSheetHeader
+  - BottomSheetBody
+  - BottomSheetFooter
+
+## API
+
+Sample usage:
+
+```jsx
```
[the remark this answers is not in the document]
anuraghazra: Done.

anuraghazra:
> Can we also add the considerations for building a custom solution rather than using a custom library in this doc?

Done.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,296 @@
...
+| `children` | `React.ReactNode` | `undefined` | Contents of the BottomSheet | ✅       |
+
+### `BottomSheetHeader` API
+
+| Prop      | Type            | Default     | Description                           | Required |
+| --------- | --------------- | ----------- | ------------------------------------- | -------- |
+| `title`   | `string`        | `undefined` | Title of the Header                   |       |
+| `subtitle`   | `string`        | `undefined` | Subtitle of the Header                   |       |
+| `leading` | `React.ReactNode` | `undefined` | leading asset or icon to be placed at the left most side of the BottomSheetHeader |          |
+| `trailing` | `Badge, Link, Text, IconButton`, `undefined` | trailing component to be placed at the right most side of the BottomSheetHeader |          |
```
burhanuday, line 107:
```suggestion
| `trailing` | `Badge, Link, Text, IconButton`, `undefined` | undefined  |     trailing component to be placed at the right most side of the BottomSheetHeader     |
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,296 @@
...
+
+```jsx
+import { BottomSheet } from '@razorpay/blade';
+
+const [isOpen, setIsOpen] = React.useState(false);
+
+<BottomSheetComponent
+  isOpen={isOpen}
+  onDismiss={() => {
+    setFirstOpen(false);
```
divyanshu013, line 50:
```suggestion
    setIsOpen(false);
```

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,296 @@
...
+### `BottomSheetHeader` API
+
+| Prop      | Type            | Default     | Description                           | Required |
+| --------- | --------------- | ----------- | ------------------------------------- | -------- |
+| `title`   | `string`        | `undefined` | Title of the Header                   |       |
+| `subtitle`   | `string`        | `undefined` | Subtitle of the Header                   |       |
+| `leading` | `React.ReactNode` | `undefined` | leading asset or icon to be placed at the left most side of the BottomSheetHeader |          |
+| `trailing` | `Badge, Link, Text, IconButton`, `undefined` | undefined  |     trailing component to be placed at the right most side of the BottomSheetHeader     |    |
+| `titleSuffix` | `Counter` | `undefined` | A component to be placed adjacent to the title text |          |
+| `hideDivider` | `boolean` | `false` | Show or hide the divider |          |
```
divyanshu013, line 109: Instead I think we should have `showDivider` which is by default `true` and users can make it `showDivider={false}`. `hideDivider={false}` involves a double negation which feels harder to parse the boolean, also makes it consistent with other `show*` props we've (eg. `showBackButton` in the same component.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,296 @@
...
+| `showBackButton` | `boolean` | `false` | Show or hide back button |          |
+| `onBackButtonClick` | `boolean` | `false` | Event handler for the back button |          |
```
divyanshu013, line 111: How about `onClickBackButton`, we've been following the convention `onClick*` at other places (eg. alert)

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,296 @@
...
+
+### `BottomSheetFooter` API
+
+
+| Prop       | Type              | Default     | Description                 | Required |
+| ---------- | ----------------- | ----------- | --------------------------- | -------- |
+| `children` | `React.ReactNode` | `undefined` | Contents of the BottomFooter | ✅       |
```
divyanshu013, line 118: Missing props - trailing, primary, secondary actions, etc. Let's also add some usage examples around same

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,296 @@
...
+
+## Composition with Dropdown
+
+We will export `BottomSheet` component separately as an independant component but generally the pattern will be to use it with SelectInput, where in mobile devices the Select's dropdown will be replaced by the BottomSheet.
```
divyanshu013, line 122:
```suggestion
We will export `BottomSheet` component separately as an independent component but generally the pattern will be to use it with SelectInput, where in mobile devices the Select's dropdown will be replaced by the BottomSheet.
```

divyanshu013: Also a screenshot will help to visualize the usecase.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,296 @@
...
+- Reduced implementation complexity & [interdependency](https://www.notion.so/BottomSheet-API-Discussion-e2aa79cd45274ef280fdb998efa6b98f?pvs=4#630d5e93897b4ca0bf54f475f0cd7d6e) of state
+
+Cons:
+
+- Not trivial to implement from user's end, they will have to compose the BottomSheet & Dropdown as per their needs.
+
+Considering the bundle size downside to approach 1, we decided to go ahead with approach 2.
+
+> **API Rabbit Hole**  
+> If you want to go deep into the API Rabbit Hole, We also had explored various other APIs and discussed internally about the pros and cons of each which we **[documented in this Notion doc](https://www.notion.so/BottomSheet-API-Discussion-e2aa79cd45274ef280fdb998efa6b98f?pvs=4#db6bd9826bc048bd97d24a50269fc45d)**
```
divyanshu013, line 156: Let's add this here in a separate md file please, this link may not work in the future.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
divyanshu013: > because bundle size savings either way isn't that huge

This seems contrasting with the recommendation above favoring approach (2) over (1) because of bundle size.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,296 @@
...
+- Escape closes the bottom sheet, after closing the focus will go to the triggered element
+- When opening the bottom sheet it should either focus on the first focusable element on the sheet or element provided by the user via `initialFocusRef` prop
+
+## Open questions
+
+- ~~What is `action` in the header trailing visual?~~
+- ~~In BottomSheetHeaderLeaing & BottomSheetFooterLeading will the `prefix` only support Icon component?~~
+
+- Consideration for using custom solution rather than third party libraries like [react-spring-bottomsheet](https://github.com/stipsan/react-spring-bottom-shee) in web.
+   - **ANS**: We did a initial research for various third-party libraries documented in [here](https://docs.google.com/document/d/1lRIBo5nkp1R9_DC9nRiFFwBEH3oVhMZjLbV7jxFrYBA/edit), most of them did not fit out use case, react-spring-bottomsheet had lot of features covered but comes with a bit of bundle size baggage like usegesture/xstate/xstate-react/react-spring libraries which in mobile users will have to carry.
```
divyanshu013, line 283: Move to a md file here please, I'm unable to access this doc.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
```diff
@@ -0,0 +1,296 @@
...
+   - Plus we had few requirements like stacked bottomsheets, special logic for where the bottomsheet will open initially which either way had to be implemented.
+
+## Alternative APIs
+
+- If you want to go deep into the API Rabbit Hole, We also had explored various other APIs and discussed internally about the pros and cons of each which we **[documented in this Notion doc.](https://www.notion.so/BottomSheet-API-Discussion-e2aa79cd45274ef280fdb998efa6b98f?pvs=4#db6bd9826bc048bd97d24a50269fc45d)**
+
+- Header/Footer [Unified APIs Discussion Doc](https://docs.google.com/document/d/1gzWRv5RyTgwJiJkzxyvCwV1XVo8QMmoevoakbDD5qsQ/edit)
```
divyanshu013, line 290: This as well, please copy paste as md in a separate file here.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Footer only accepts JSX as children, no other props.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
anuraghazra: Should be fine anyway we are not recommending this, just a contrived example that it "could be possible" but at the end of the day what we want is to keep dropdown & bottomsheet separate component which we've done.

packages/blade/src/components/BottomSheet/_decisions/decisions.md
divyanshu013: Oh, might have checked an old story on a different PR 🥵

PR #1007, opened 2023-02-13
fix(Dropdown): change footer title to avoid confusion

saurabhdaware:
Search with search icon on footer makes it feel like its input with search. Changing the footer title and icon as Abhishek suggested
https://razorpay.slack.com/archives/C01H13RTF8V/p1676280664087239?thread_ts=1676278793.897139&cid=C01H13RTF8V

### DIFF
--- a/packages/blade/src/components/Dropdown/Dropdown.stories.tsx
+++ b/packages/blade/src/components/Dropdown/Dropdown.stories.tsx
@@ -22,7 +22,7 @@ import {
   InfoIcon,
   ArrowRightIcon,
   HistoryIcon,
-  SearchIcon,
+  FileTextIcon,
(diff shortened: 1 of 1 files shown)


packages/blade/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -72,7 +72,7 @@ const Page = (): ReactElement => {
       }}
     >
       <Title>Usage</Title>
-      <Sandbox editorHeight={600}>
+      <Sandbox showConsole editorHeight={600}>
```
saurabhdaware, line 75: showing console by default since we're printing onChange values on console

kamleshchandnani (approved):

PR #1008, opened 2023-02-13
docs(Dropdown): add note on Safari bug

divyanshu013:
Missed to post in review. Added a note so we do not have to scratch our heads while reading the code in future


saurabhdaware (approved):
yay thanks

kamleshchandnani (approved):

PR #1010, opened 2023-02-14
feat(amount): add amount component

Priyanshu-C:
feat: Amount Component

Screenshots - 
image
image
image


packages/blade/src/components/Amount/_decisions/decisions.md
```diff
@@ -0,0 +1,81 @@
+# Amount <!-- omit in toc -->
```
divyanshu013, line 1: The API decisions is already a part of #993, we can remove from this PR.

packages/blade/src/components/Amount/BaseAmount.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const affixFontWeight = getDecimalFontWeight(isAffixSubtle);
+    const affixFontSize = getSuffixPrefixFontSize(isAffixSubtle, size);
+    return (
+      <>
+        <BaseText fontSize={amountTextSizes[size]} fontWeight={weight} color={textColor}>
+          {integer}.
+        </BaseText>
+        <BaseText fontWeight={affixFontWeight} fontSize={affixFontSize} color={prefixSuffixColor}>
+          {decimal}
+        </BaseText>
```
divyanshu013, line 49: How is text wrapping working in this case? We don't want the decimals to text wrap without the integer amount ideally.

divyanshu013 (commented):
Could you fix the storybook and add some PR description with details / screenshots please? The chromatic link is not generating. Also please add screenshots of native stories on ios / android too.

packages/blade/src/components/Amount/BaseAmount.tsx
```diff
@@ -0,0 +1,61 @@
+import type { ReactElement } from 'react';
```
chaitanyadeorukhkar, line 1: Why do we need a `BaseAmount` component?

packages/blade/src/components/Amount/BaseAmount.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const affixFontWeight = getDecimalFontWeight(isAffixSubtle);
+    const affixFontSize = getSuffixPrefixFontSize(isAffixSubtle, size);
+    return (
+      <>
+        <BaseText fontSize={amountTextSizes[size]} fontWeight={weight} color={textColor}>
+          {integer}.
+        </BaseText>
+        <BaseText fontWeight={affixFontWeight} fontSize={affixFontSize} color={prefixSuffixColor}>
+          {decimal}
+        </BaseText>
```
chaitanyadeorukhkar, line 50: Lets nest the Text elements together?

```tsx
<BaseText>
  {integer}
  <BaseText>
    {decimal}
  </BaseText>
</BaseText>
```

packages/blade/src/components/Amount/StyledAmount.web.tsx
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components';
+import type { StyledAmountProps } from './types';
+
+const StyledAmount = styled.div<StyledAmountProps>(() => ({
+  width: 'fit-content',
```
chaitanyadeorukhkar, line 5: Why do we need to make it `fit-content`?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,177 @@
...
+  }
+  const renderedValue = formatAmountWithSuffix(suffix, value);
+  const { textColor, prefixSuffixColor } = getColorProps({
+    variant,
+  });
+  const rupeeweight = getRupeeweight(isAffixSubtle, weight);
+  const rupeeFontSize = getSuffixPrefixFontSize(isAffixSubtle, size);
+
+  return (
+    <StyledAmount {...metaAttribute(MetaConstants.Component, MetaConstants.Amount)}>
```
chaitanyadeorukhkar, line 148: Why do we need a `StyledAmount`? Can't we use `Box` component directly?

packages/blade/src/components/Amount/_decisions/decisions.md
chaitanyadeorukhkar: +1

packages/blade/src/components/Amount/types.ts
```diff
@@ -0,0 +1,5 @@
+import type { Theme } from '~components/BladeProvider';
+
+export type StyledAmountProps = {
+  theme: Theme;
```
chaitanyadeorukhkar, line 4: Why do we need to pass down theme? any component can get this directly by using `useTheme()`

divyanshu013 (commented):
Will review after finalizing the size discussion. Let's remove API decision from this PR, it's already scoped in #993

chaitanyadeorukhkar:
@Priyanshu-C We should add `.00` in the end instead of just showing a `.` when `decimals` is the suffix and there is no decimal value
Screenshot 2023-03-13 at 1 13 49 PM

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+Amount.args = {
+  value: 122345.678,
+  intent: 'neutral',
+  size: 'title-medium',
+  suffix: 'decimals',
+  isAffixSubtle: true,
+};
+Amount.storyName = 'Default';
+
+const AmountSizesTemplate: ComponentStory<typeof AmountComponent> = ({ ...args }) => {
```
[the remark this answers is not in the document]
Priyanshu-C: There was a minor issue here, but the design has these conditions for prefix to be bold 
- if isAffixSubtle is true, then bold
- if contains bold then, then bold. 
In anyother case it will be regular.

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+NoSuffix.storyName = 'No Suffix';
+
+export const HumanizeSuffix = AmountTemplate.bind({});
+HumanizeSuffix.args = {
+  ...defaultArgs,
+  suffix: 'humanize',
+};
+HumanizeSuffix.storyName = 'Humanize Suffix';
+
+export const AffixSubtleOff = AmountTemplate.bind({});
```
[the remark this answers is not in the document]
Priyanshu-C: It is a comma and not a dot. Weird, let's connect to debug this. 

image

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -0,0 +1,184 @@
+import type { ComponentStory, Meta } from '@storybook/react';
```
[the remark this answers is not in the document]
Priyanshu-C: Done.

chaitanyadeorukhkar:
@Priyanshu-C when the suffix is none, the amount is not formatted with commas anymore. Can you check this?
Screenshot 2023-03-14 at 5 40 13 PM

chaitanyadeorukhkar:
@Priyanshu-C the story seems a bit difficult to understand visually when the heading (size name) and the output is one left and right. Can we move them one below the other like we've done on List component stories.
Ref:

Amount: 
Screenshot 2023-03-14 at 5 42 34 PM

List:
Screenshot 2023-03-14 at 5 42 34 PM

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+  suffix: 'none',
+};
+NoSuffix.storyName = 'No Suffix';
+
+export const HumanizeSuffix = AmountDefaultTemplate.bind({});
+HumanizeSuffix.args = {
+  ...defaultArgs,
+  suffix: 'humanize',
+};
+HumanizeSuffix.storyName = 'Humanize Suffix';
```
chaitanyadeorukhkar, line 166: Suggestion: Add more variety of amounts here like 1L, 2Cr, 3k

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+    | `title-medium`
+    | `title-small`
+    | `heading-large`
+    | `heading-large-bold`
+    | `heading-small`
+    | `heading-small-bold`
+    | `body-medium`
+    | `body-medium-bold`
+    | `body-small`
+    | `body-small-bold`;
```
anuraghazra, line 52: Seems like we are missing few sizes, body-xsmall, heading-medium etc 🤔 

We should compose this union from the title/heading/body props, otherwise they might get out of sync.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+const getColorProps = ({ intent }: { intent: AmountProps['intent'] }): ColorProps => {
+  const props: ColorProps = {
+    textColor: 'surface.text.normal.lowContrast',
+    prefixSuffixColor: 'surface.text.muted.lowContrast',
+  };
+  if (!intent) return props;
+  props.textColor = `feedback.text.${intent}.lowContrast`;
+  props.prefixSuffixColor = `feedback.text.${intent}.lowContrast`;
+  if (intent === 'neutral') {
+    props.prefixSuffixColor = `surface.text.muted.lowContrast`;
```
chaitanyadeorukhkar, line 87: Why can't we use `feedback.text.neutral.lowContrast` here? Why are we using surface color for a feedback intent here?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+  }
+  return props;
+};
+
+const getFlooredFixed = (value: number, precision: number): number => {
+  const factor = 10 ** precision;
+  const roundedValue = Math.floor(value * factor) / factor;
```
chaitanyadeorukhkar, line 94: I'm bad at math. Why are we doing this? 😅 

Math.floor won't be enough? Also add comments explaining what's happening here for math genuises like me

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+  return Number(roundedValue.toFixed(precision));
+};
+
+const addCommas = (num: number, currency: currencyType): string => {
+  const locale = currencylocaleMapping[currency];
```
chaitanyadeorukhkar, line 99:
```suggestion
  const locale = currencyLocaleMapping[currency];
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+  return props;
+};
+
+const getFlooredFixed = (value: number, precision: number): number => {
+  const factor = 10 ** precision;
+  const roundedValue = Math.floor(value * factor) / factor;
+  return Number(roundedValue.toFixed(precision));
+};
+
+const addCommas = (num: number, currency: currencyType): string => {
```
chaitanyadeorukhkar, line 98: Lets try to keep the variables more self explanatory
```suggestion
const addCommas = (amountValue: number, currency: currencyType): string => {
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const locale = currencylocaleMapping[currency];
+  return num.toLocaleString(locale);
```
chaitanyadeorukhkar, line 100: Does this also support MYR locale?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+};
+
+const getFormattedAmountWithSuffixSymbol = (num: number, currency: currencyType): string => {
```
chaitanyadeorukhkar, line 103: Seems more contextual
```suggestion
const getHumanizedAmount = (amountValue: number, currency: Currency): string => {
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+import type { TestID } from '~src/_helpers/types';
+import { metaAttribute, MetaConstants } from '~utils';
+
+export const suffixTypes = {
+  HUMANIZE: 'humanize',
+  DECIMALS: 'decimals',
+  NONE: 'none',
+};
+
+type currencyType = 'INR' | 'MYR';
```
chaitanyadeorukhkar, line 23: Lets make it official! We'll use this as the type for when we add a new `currency: Currency` prop
```suggestion
type Currency = 'INR' | 'MYR';
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+  return Number(roundedValue.toFixed(precision));
+};
+
+const addCommas = (num: number, currency: currencyType): string => {
+  const locale = currencylocaleMapping[currency];
+  return num.toLocaleString(locale);
+};
+
+const getFormattedAmountWithSuffixSymbol = (num: number, currency: currencyType): string => {
+  const abbreviations: { value: number; symbol: string }[] = currencyAbbreviationsMapping[currency];
```
chaitanyadeorukhkar, line 104: Why do we need to explicitly type this? currencyAbbreviationsMapping should take care of the type here right?
```suggestion
  const abbreviations = currencyAbbreviationsMapping[currency];
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  const abbreviation = abbreviations.find((abbr) => num >= abbr.value);
+  if (abbreviation) {
+    num = num / abbreviation.value;
+    const formattedNum = Number(getFlooredFixed(num, 2));
```
chaitanyadeorukhkar, line 109: Verbose variable name. Also, why do we need to wrap it in `Number`? Doesn't getFloorFixed always return number?
```suggestion
    const formattedAmountValue = getFlooredFixed(num, 2);
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+        alignItems="baseline"
+        paddingLeft="spacing.2"
+        paddingRight="spacing.2"
+      >
+        <BaseBox paddingRight="spacing.1">
+          <BaseText fontWeight={currencyWeight} fontSize={currencyFontSize} color={currencyColor}>
+            {currencyPrefix}
+          </BaseText>
+        </BaseBox>
+        <AmountValue
```
chaitanyadeorukhkar, line 180: Why do we need a separate file and component for `AmountValue`? Its not re-used and neither does it have native/web specific things. Lets render everything in `Amount.tsx` directly only and remove `AmountValue` component?

packages/blade/src/components/Amount/AmountValue.tsx
```diff
@@ -0,0 +1,71 @@
+import type { ReactElement } from 'react';
```
chaitanyadeorukhkar, line 1: We shouldn't have this as a separate file. Any strong reason for it?

packages/blade/src/components/Amount/AmountValue.tsx
```diff
@@ -0,0 +1,71 @@
...
+import BaseBox from '~components/Box/BaseBox';
+import { BaseText } from '~components/Typography/BaseText';
+
+const getAffixFontWeight = (isAffixSubtle: true | false): 'regular' | 'bold' => {
+  if (isAffixSubtle) return 'regular';
+  return 'bold';
+};
+
+const getMainValuetWeight = (size: NonNullable<AmountProps['size']>): 'regular' | 'bold' => {
+  return size.includes('bold') || size.includes('title') ? 'bold' : 'regular';
```
chaitanyadeorukhkar, line 17: use startsWith for more accuracy?
```suggestion
  return size.includes('bold') || size.startsWith('title') ? 'bold' : 'regular';
```

packages/blade/src/components/Amount/__tests__/Amount.native.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+    const { toJSON } = renderWithTheme(<Amount size="heading-small-bold" value={1000} />);
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render medium size Amount', () => {
+    const { toJSON } = renderWithTheme(<Amount size="title-medium" value={1000} />);
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render medium size Amount', () => {
```
chaitanyadeorukhkar, line 74: Incorrect test descriptions

packages/blade/src/components/Amount/__tests__/Amount.native.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+  });
+
+  it('should render amount with Humanize value', () => {
+    const { toJSON } = renderWithTheme(
+      <Amount size="heading-small" suffix="humanize" value={1000.22} />,
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render positive variant Amount ', () => {
```
chaitanyadeorukhkar, line 93:
```suggestion
  it('should render positive intent Amount ', () => {
```

packages/blade/src/components/Amount/__tests__/Amount.native.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+    const { toJSON } = renderWithTheme(<Amount intent="negative" value={1000} />);
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render neutral variant Amount ', () => {
+    const { toJSON } = renderWithTheme(<Amount intent="neutral" value={1000} />);
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render positive variant Amount ', () => {
```
chaitanyadeorukhkar, line 113: incorrect description

packages/blade/src/components/Amount/__tests__/Amount.native.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+    const { toJSON } = renderWithTheme(<Amount intent="notice" value={1000} />);
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render positive variant Amount with bold and isAffixSubtle false', () => {
```
chaitanyadeorukhkar, line 118: Why is this combination needed to be tested? Anything different here?

packages/blade/src/components/Amount/__tests__/Amount.native.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+    const { toJSON } = renderWithTheme(
+      <Amount isAffixSubtle={false} intent="positive" value={1000} />,
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render information variant Amount ', () => {
```
chaitanyadeorukhkar, line 125: incorrect description for this and the following few test cases where you're checking with intext and affix subtle as false

packages/blade/src/components/Amount/__tests__/Amount.native.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+    const { toJSON } = renderWithTheme(
+      <Amount isAffixSubtle={false} intent="information" value={1000} />,
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render negative variant Amount ', () => {
```
chaitanyadeorukhkar, line 132: I think lets just add test for 1 type of intent and affix as false instead of all of them, snapshots are way too much otherwise and if one is correct, others would be correct too since we're using feedback colors directly

packages/blade/src/components/Amount/__tests__/Amount.ssr.test.tsx
```diff
@@ -0,0 +1,10 @@
+import { Amount } from '..';
+import renderWithSSR from '~src/_helpers/testing/renderWithSSR.web';
+
+describe('<Badge />', () => {
+  it('should render Badge on server', () => {
+    const { container } = renderWithSSR(<Amount value={10000} />);
```
chaitanyadeorukhkar, line 6: copy pasta. Needs to say Amount

packages/blade/src/components/Amount/__tests__/Amount.web.test.tsx
```diff
@@ -0,0 +1,146 @@
+import { Amount } from '../Amount';
```
chaitanyadeorukhkar, line 1: Same issues as the ones highlighted in the native test. Description errors, etc.

packages/blade/src/components/Amount/_decisions/decisions.md
chaitanyadeorukhkar: @Priyanshu-C lets not keep this here. Chain this PR to Amount decisions so that this PR only contains the implementation.

packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -0,0 +1,62 @@
+const prefixSuffixTextSizes = {
```
chaitanyadeorukhkar, line 1:
```suggestion
const affixFontSizes = {
```

chaitanyadeorukhkar, line 1: Add proper types
```suggestion
const affixFontSizes: Record<NonNullable<AmountProps['size']>, keyof FontSize> = {
```

packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -0,0 +1,62 @@
...
+  'body-medium-bold': 75,
+  'heading-small': 75,
+  'heading-small-bold': 75,
+  'heading-large': 100,
+  'heading-large-bold': 100,
+  'title-small': 300,
+  'title-medium': 400,
+} as const;
+
+const amountTextSizes = {
```
chaitanyadeorukhkar, line 14: Add proper type

packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -0,0 +1,62 @@
...
+  'body-medium-bold': 100,
+  'heading-small': 200,
+  'heading-small-bold': 200,
+  'heading-large': 400,
+  'heading-large-bold': 400,
+  'title-small': 600,
+  'title-medium': 700,
+} as const;
+
+const currencyPrefixMapping = {
```
chaitanyadeorukhkar, line 27: Add proper type

packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -0,0 +1,62 @@
...
+    'currency-symbol': '₹',
+    'currency-code': 'INR',
+  },
+  MYR: {
+    'currency-symbol': 'RM',
+    'currency-code': 'MYR',
+  },
+};
+
+const currencylocaleMapping = {
```
chaitanyadeorukhkar, line 38: Add proper type

packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -0,0 +1,62 @@
...
+  INR: 'en-IN',
+  MYR: 'en-MY',
+};
+
+const currencyAbbreviationsMapping = {
```
chaitanyadeorukhkar, line 43: Add proper type

packages/blade/src/components/Amount/Amount.tsx
Priyanshu-C: Yes

packages/blade/src/components/Amount/__tests__/Amount.ssr.test.tsx
Priyanshu-C: hehe, sorry :P

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const { textColor, prefixSuffixColor } = getColorProps({
+    intent,
+  });
+
+  const currencyColor = isAffixSubtle ? prefixSuffixColor : textColor;
+  const currencyFontSize = getAffixFontSize(isAffixSubtle, size);
+  const currencyWeight = getCurrencyWeight(isAffixSubtle, size);
+
+  return (
+    <StyledAmount {...metaAttribute({ name: MetaConstants.Amount, testID })}>
```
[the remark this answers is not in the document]
Priyanshu-C: Gotcha, I did not know BaseBox as View as native.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+  amountValueColor: BaseTextProps['color'];
+  affixColor: BaseTextProps['color'];
+};
+
+const getTextColorProps = ({ intent }: { intent: AmountProps['intent'] }): ColorProps => {
+  const props: ColorProps = {
+    amountValueColor: 'surface.text.normal.lowContrast',
+    affixColor: 'surface.text.muted.lowContrast',
+  };
+  if (!intent) return props;
```
chaitanyadeorukhkar, line 77: We should also add a check for `neutral` here. We will throw TS error but non-js projects can still pass `neutral` and this function will return `feedback.text.neutral.lowContrast` and will render it too

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+  props.amountValueColor = `feedback.text.${intent}.lowContrast`;
+  props.affixColor = `feedback.text.${intent}.lowContrast`;
+  return props;
+};
+
+const getAffixFontWeight = (isAffixSubtle: true | false): 'regular' | 'bold' => {
```
chaitanyadeorukhkar, line 83: Do we really need a separate function for this? Just do this?

```tsx
  const affixFontWeight = isAffixSubtle ? 'regular' : 'bold'
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+  if (isAffixSubtle) return 'regular';
+  return 'bold';
+};
+
+const getMainValueWeight = (
```
chaitanyadeorukhkar, line 88: Make the name more contextual?
```suggestion
const getAmountValueWeight = (
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+  value,
+  size = 'body-medium',
+  amountValueColor,
+  isAffixSubtle = true,
+  suffix = 'decimals',
+  affixColor,
+}: AmountValue): ReactElement => {
+  const affixFontWeight = getAffixFontWeight(isAffixSubtle);
+  const affixFontSize = getAffixFontSize(isAffixSubtle, size);
+  const valueForWeight = getMainValueWeight(size);
```
chaitanyadeorukhkar, line 118: These can be direct inline checks since there's nothing complex happening here

```suggestion
  const affixFontWeight = isAffixSubtle ? 'regular' : 'bold';
  const affixFontSize = isAffixSubtle ? affixFontSizes[size] : amountTextSizes[size];;
  const valueForWeight = size.includes('bold') || size.startsWith('title') ? 'bold' : 'regular';
```

packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -0,0 +1,85 @@
...
+  'body-medium-bold': 75,
+  'heading-small': 75,
+  'heading-small-bold': 75,
+  'heading-large': 100,
+  'heading-large-bold': 100,
+  'title-small': 300,
+  'title-medium': 400,
+} as const;
+
+const amountTextSizes: Record<NonNullable<AmountProps['size']>, keyof FontSize> = {
```
chaitanyadeorukhkar, line 17: Lets keep the naming consistent. We're calling it `affixFontSizes` above, so lets call it FontSize here also
```suggestion
const amountFontSizes: Record<NonNullable<AmountProps['size']>, keyof FontSize> = {
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+  return (
+    <BaseBox
+      paddingLeft="spacing.2"
+      paddingRight="spacing.2"
+      {...metaAttribute({ name: MetaConstants.Amount, testID })}
+    >
+      <BaseBox
+        display="flex"
+        alignItems="baseline"
```
chaitanyadeorukhkar, line 249: Do we need to explicitly pass `baseline`? Is it not aligned properly otherwise?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+  });
+
+  const currencyColor = isAffixSubtle ? affixColor : amountValueColor;
+  const currencyFontSize = getAffixFontSize(isAffixSubtle, size);
+  const currencyWeight = getCurrencyWeight(isAffixSubtle, size);
+
+  return (
+    <BaseBox
+      paddingLeft="spacing.2"
+      paddingRight="spacing.2"
```
chaitanyadeorukhkar, line 244: Why is there a padding on left and right for Amount? Don't see this on figma

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+      {...metaAttribute({ name: MetaConstants.Amount, testID })}
+    >
+      <BaseBox
+        display="flex"
+        alignItems="baseline"
+        paddingLeft="spacing.2"
+        paddingRight="spacing.2"
```
chaitanyadeorukhkar, line 251: Why is this padding here? that's a padding of 4px on each side, which is not on figma either.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+  const { amountValueColor, affixColor } = getTextColorProps({
+    intent,
+  });
+
+  const currencyColor = isAffixSubtle ? affixColor : amountValueColor;
+  const currencyFontSize = getAffixFontSize(isAffixSubtle, size);
+  const currencyWeight = getCurrencyWeight(isAffixSubtle, size);
+
+  return (
+    <BaseBox
```
chaitanyadeorukhkar, line 242: Something seems off. The amount component is taking up quite a lot of space here. 

Screenshot 2023-03-17 at 12 42 17 PM

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -0,0 +1,155 @@
+import type { ComponentStory, Meta } from '@storybook/react';
+import { Title } from '@storybook/addon-docs';
+import type { AmountProps } from './Amount';
+import { Amount as AmountComponent } from './Amount';
+import BaseBox from '~components/Box/BaseBox';
+import { Sandbox } from '~src/_helpers/storybook/Sandbox';
+import { Text } from '~components/Typography';
+import StoryPageWrapper from '~src/_helpers/storybook/StoryPageWrapper';
+
+const Page = (): React.ReactElement => {
```
chaitanyadeorukhkar, line 10: Can you add a story where Amount is used inline with Title, Heading and Text components?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+      paddingRight="spacing.2"
+      {...metaAttribute({ name: MetaConstants.Amount, testID })}
+    >
+      <BaseBox
+        display="flex"
+        alignItems="baseline"
+        paddingLeft="spacing.2"
+        paddingRight="spacing.2"
+      >
+        <BaseBox paddingRight="spacing.1">
```
chaitanyadeorukhkar, line 253: We don't need to add a basebox wrapper here. You can directly pass padding to BaseText now. Take a pull from master.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+  amountValueColor: BaseTextProps['color'];
+  value: string;
+}
+
+const AmountValue = ({
+  value,
+  size = 'body-medium',
+  amountValueColor,
+  isAffixSubtle = true,
+  suffix = 'decimals',
```
chaitanyadeorukhkar, line 113: Why are we assigning defaults here? This could lead to issues in the future. For example, if I change the default on `Amount`, it could interfere with the default set here. Lets keep all the default assignments on the Amount component only and ensure that `AmountValue` props will get guaranteed values and never be undefined, that way we don't need it to have a default value

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+  const affixFontWeight = getAffixFontWeight(isAffixSubtle);
+  const affixFontSize = getAffixFontSize(isAffixSubtle, size);
+  const valueForWeight = getMainValueWeight(size);
+  if (suffix === 'decimals' && isAffixSubtle) {
+    const integer = value.split('.')[0];
+    const decimal = value.split('.')[1];
+
+    return (
+      <>
+        <BaseBox paddingRight="spacing.1">
```
chaitanyadeorukhkar, line 125: Lets take a pull from master and add spacing directly to BaseText. Will make your implementation a lot simpler

packages/blade/src/components/Amount/Amount.tsx
chaitanyadeorukhkar: Still don't understand it. For my sake can you give an example where it would fail without using factor calculations? Won't `Math.toFixed(value, digits)` be enough for all use-cases?

packages/blade/src/components/Amount/getAmountBadgeStyles.ts
```diff
@@ -0,0 +1,10 @@
+import type { CSSObject } from 'styled-components';
```
chaitanyadeorukhkar, line 1: Do we need this file anymore?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+  /**
+   * Sets the size of the amount.
+   *
+   * @default 'heading'
+   */
+  size?:
+    | `title-${NonNullable<Exclude<TitleProps['size'], 'large'>>}`
+    | `heading-${NonNullable<Exclude<HeadingProps<{ variant: 'regular' }>['size'], 'medium'>>}`
+    | `heading-${NonNullable<
+        Exclude<HeadingProps<{ variant: 'regular' }>['size'], 'medium'>
```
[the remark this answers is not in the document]
Priyanshu-C: because heading-medium is not part of the design. 
https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=28012-580578&t=jI2b5KYW6UPc0bcD-0

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,271 @@
...
+      >}-${NonNullable<Exclude<TextProps<{ variant: 'body' }>['weight'], 'regular'>>}`
+    | `body-${NonNullable<Exclude<TextProps<{ variant: 'body' }>['size'], 'xsmall'>>}`
```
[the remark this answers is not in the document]
Priyanshu-C: Same reason as above.

packages/blade/src/components/Amount/Amount.tsx
Priyanshu-C: Yes, this was required for the affix.

Priyanshu-C:
@chaitanyadeorukhkar isFixed rounds off the numbers - 1.909.isFixed(2) will be equal to 1.91.

chaitanyadeorukhkar:
@Priyanshu-C there's still some spacing issue. 2px margin is _after_ the decimal point and not _before. Attaching screenshots of Figma & Storybook for reference

**Figma:**
Screenshot 2023-03-20 at 12 38 48 PM

**Storybook:**
Screenshot 2023-03-20 at 12 38 41 PM

packages/blade/src/components/Amount/Amount.stories.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: This is resolved now? bold and regular works as it should, right?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,245 @@
...
+        Exclude<TextProps<{ variant: 'body' }>['weight'], 'regular'>
+      >}`;
+  /**
+   * Indicates what the suffix of amount should be
+   *
+   * @default 'decimals'
+   */
+  suffix?: 'decimals' | 'none' | 'humanize';
+  /**
+   * in amount by making the prefix symbol and decimal digits s
```
chaitanyadeorukhkar, line 53: This doesn't seem to be a legible sentence. Looks like something messed up in merges.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,245 @@
...
+   * @default true
+   */
+  isAffixSubtle?: true | false;
+  /**
+   * Prefix to be shown before the amount value. The prefix can be either a currency symbol or a currency code.
+   *
+   * @default 'currency-symbol'
+   */
+  prefix?: 'currency-symbol' | 'currency-code';
+} & TestID;
```
chaitanyadeorukhkar, line 64: We also need to add support for `StyledPropsBlade`. Take a look at the implementation for `Heading` component on master. This was recently added but we can't ship amount without this. Should be a quick addition. Let me know if you need any help on adding this.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,245 @@
...
+  amountValueColor: BaseTextProps['color'];
+  affixColor: BaseTextProps['color'];
+};
+
+const getTextColorProps = ({ intent }: { intent: AmountProps['intent'] }): ColorProps => {
+  const props: ColorProps = {
+    amountValueColor: 'surface.text.normal.lowContrast',
+    affixColor: 'surface.text.muted.lowContrast',
+  };
+  if (!intent && intent !== 'neutral') return props;
```
chaitanyadeorukhkar, line 76: Can we also throw an error when someone passes `neutral` as an intent? Otherwise people might pass neutral in JS projects and raise a bug that it doesn't work.
```tsx
if(intent==='neutral')
{
  throw new Error("[Blade Amount]: `neutral` intent is not support.")
}
```

packages/blade/src/components/Amount/Amount.tsx
chaitanyadeorukhkar: @Priyanshu-C can you check this?

packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -0,0 +1,85 @@
...
+  MYR: 'en-MY',
+};
+
+type CurrencyAbbreviation = {
+  value: number;
+  symbol: string;
+};
+
+type CurrencyAbbreviationsMapping = {
+  [key: string]: CurrencyAbbreviation[];
```
chaitanyadeorukhkar, line 63: Shouldn't the key be of type `Currency`?

chaitanyadeorukhkar:
@Priyanshu-C amount alignment is broken on react native now
Simulator Screen Shot - iPhone 14 Pro - 2023-03-20 at 14 20 46

chaitanyadeorukhkar:
This story is still broken @Priyanshu-C 
Screenshot 2023-03-21 at 3 07 40 PM

chaitanyadeorukhkar:
Can we add some margin below amount for this story?
Screenshot 2023-03-21 at 3 08 44 PM

Priyanshu-C:
image
Fixed this issue along with spacing issue before the decimal point.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,270 @@
...
+}: AmountProps): ReactElement => {
+  if (isNaN(value)) {
+    throw new Error('[Blade: Amount]: `value` prop must be of type `number` for Amount.');
+  }
+  // @ts-expect-error netural intent should throw error
+  if (intent === 'neutral') {
+    throw new Error('[Blade Amount]: `neutral` intent is not support.');
+  }
+
+  // This will be added to prop and can be switched to a currency
```
chaitanyadeorukhkar, line 229: remove this?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,270 @@
...
+  intent,
+  prefix = 'currency-symbol',
+  testID,
+  currency = 'INR',
+  ...styledProps
+}: AmountProps): ReactElement => {
+  if (isNaN(value)) {
+    throw new Error('[Blade: Amount]: `value` prop must be of type `number` for Amount.');
+  }
+  // @ts-expect-error netural intent should throw error
```
chaitanyadeorukhkar, line 224:
```suggestion
  // @ts-expect-error neutral intent should throw error
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,270 @@
...
+}: AmountValue): ReactElement => {
+  const affixFontWeight = isAffixSubtle ? 'regular' : 'bold';
+  const affixFontSize = isAffixSubtle ? affixFontSizes[size] : amountFontSizes[size];
+  const valueForWeight = size.includes('bold') || size.startsWith('title') ? 'bold' : 'regular';
+  const isNative = useIsNative();
+  if (suffix === 'decimals' && isAffixSubtle) {
+    const integer = value.split('.')[0];
+    const decimal = value.split('.')[1];
+
+    // Becasue Basetext doesn't support display flex we need to wrap it in a fragment for web
```
chaitanyadeorukhkar, line 119: Instead of this can you mention why we need to use BaseText on native?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,270 @@
...
+    const AmountWrapper = isNative ? BaseText : React.Fragment;
```
chaitanyadeorukhkar, line 120: Why do we need a variable that calls a function that calls another function to do just one check? Use it directly?
```suggestion
    const AmountWrapper = getPlatformType()==='react-native' ? BaseText : React.Fragment;
```

chaitanyadeorukhkar:
Minor comments otherwise LGTM 🚀 

We'll to an L2 review and then merge this!

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,264 @@
...
+   *
+   * @default undefined
+   */
+  intent?: Exclude<Feedback, 'neutral'>;
+  /**
+   * Sets the size of the amount.
+   *
+   * @default 'body-medium'
+   */
+  size?:
```
divyanshu013, line 39: Optional: I feel we can maintain a union separately since there are limited options, seems too complex to read. eg. `title-small | title-medium | ...`

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,264 @@
...
+  amountValueColor: BaseTextProps['color'];
+  affixColor: BaseTextProps['color'];
+};
+
+const getTextColorProps = ({ intent }: { intent: AmountProps['intent'] }): ColorProps => {
+  const props: ColorProps = {
+    amountValueColor: 'surface.text.normal.lowContrast',
+    affixColor: 'surface.text.muted.lowContrast',
+  };
+  if (!intent && intent !== 'neutral') return props;
```
divyanshu013, line 86: The conditional seems confusing - `!intent && intent !== 'neutral'`. The first part sounds like it should only execute further if `intent` is a falsy value so the second part seems confusing where we're saying it shouldn't be `neutral` 😕

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,264 @@
...
+          {integer}.
+        </BaseText>
+        <BaseText
+          marginLeft="spacing.1"
+          fontWeight={affixFontWeight}
+          fontSize={affixFontSize}
+          color={affixColor}
+        >
+          {decimal || '00'}
+        </BaseText>
```
divyanshu013, line 133: Why is everything rendering as `div`?

image

divyanshu013: You perhaps need to specify `as`

divyanshu013, line 133: Could you please check if this doesn't cause unexpected wrapping? For example where the decimal might wrap to a different line in case integer part is large.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,264 @@
...
+  const roundedValue = Math.floor(value * factor) / factor;
+  return Number(roundedValue.toFixed(decimalPlaces));
+};
+
+const addCommas = (amountValue: number, currency: Currency): string => {
+  const locale = currencyLocaleMapping[currency];
+  return amountValue.toLocaleString(locale);
+};
+
+const getHumanizedAmount = (amountValue: number, currency: Currency): string => {
```
divyanshu013, line 157: Some comment to explain what do we mean by "humanized"

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  const abbreviation = abbreviations.find((abbr) => amountValue >= abbr.value);
+  if (abbreviation) {
+    amountValue = amountValue / abbreviation.value;
+    const formattedAmountValue = getFlooredFixed(amountValue, 2);
+    return addCommas(formattedAmountValue, currency) + abbreviation.symbol;
+  } else {
+    return amountValue.toString();
+  }
+};
```
divyanshu013, line 168: We should add some unit tests for these utils

packages/blade/src/components/Amount/__tests__/Amount.native.test.tsx
```diff
@@ -0,0 +1,109 @@
+import { Amount } from '../Amount';
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
+
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
+afterAll(() => jest.restoreAllMocks());
```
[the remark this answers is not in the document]
Priyanshu-C: removing, not getting any.

packages/blade/src/components/Amount/Amount.tsx
Priyanshu-C: While taking snapshots, we are conversing this.

packages/blade/src/components/Amount/Amount.tsx
Priyanshu-C: No, it doesn't.

Priyanshu-C:
@divyanshu013 Please check, I have made the required changes.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -120,6 +123,7 @@ const AmountValue = ({
           fontSize={amountFontSizes[size]}
           fontWeight={valueForWeight}
           color={amountValueColor}
+          as={isReactNative ? undefined : 'span'}
```
divyanshu013, line 126: `AmountWrapper` is rendering a `div`, should it be `p` tag instead?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -19,6 +19,17 @@ import type { StyledPropsBlade } from '~components/Box/styledProps';
 
 type Currency = 'INR' | 'MYR';
 
+type SizeType =
```
divyanshu013, line 22: What I meant was can we define the types directly rather than interpolating as template literal types? Because the sizes are limited it can simplify readability, i.e. `type SizeType = 'title-small' | 'title-medium' | ...`

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -154,6 +159,7 @@ const addCommas = (amountValue: number, currency: Currency): string => {
   return amountValue.toLocaleString(locale);
 };
 
+// This function returns the humanized amount ie: human readble currency format
```
divyanshu013, line 162: An example might be more helpful on what we mean by "human readable". Also you can use the following comment style so it shows in vscode hints:

```js
/**
 * This function ...
 *
 */
```

divyanshu013:
@Priyanshu-C could you please also review the chromatic UI tests? I see some unrelated changes on `Link`

packages/blade/src/components/Amount/Amount.tsx
divyanshu013: Yup, but having few tests for these utils will be helpful for future maintenance. Just 2 - 3 tests per utils should do to cover possible scenarios.

packages/blade/src/components/Amount/Amount.tsx
Priyanshu-C: https://github.com/razorpay/blade/pull/1010#discussion_r1136922692

packages/blade/src/components/Amount/Amount.tsx
Priyanshu-C: We need that for alignment in native, have left a comment.

divyanshu013: I meant for web it's rendering a `div` container

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -213,12 +219,12 @@ const Amount = ({
   currency = 'INR',
   ...styledProps
 }: AmountProps): ReactElement => {
-  if (isNaN(value)) {
+  if (Number.isNaN(value) || typeof value !== 'number') {
```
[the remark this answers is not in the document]
Priyanshu-C: image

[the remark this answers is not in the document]
divyanshu013: Ofcourse people won't be passing `NaN` in `value` 😄

Priyanshu-C:
@divyanshu013 Let me know if you want the size types to be reverted because we discussed and added them like that. //cc @chaitanyadeorukhkar. Also, I have not made any change in any base component so chromatic is showing invalid change.

packages/blade/src/components/Amount/Amount.tsx
divyanshu013: Okay, let's go ahead for now as it's.

divyanshu013:
Let's go ahead as it's for now. Could you please check the other comments too?

packages/blade/src/components/Amount/Amount.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: just thought a bit more about this. If we add new sizes in title or text, it will automatically update the acceptable size on Amount but we actually won’t have that support on Amount and that would not render as a consumer would expect. Rather its better if we add a new size to Title and then get an issue raised that Amount doesn’t support that size.
Seems its safer to keep literal strings for size.

divyanshu013:
@Priyanshu-C discussed this with @chaitanyadeorukhkar 

> Chaitanya: just thought a bit more about this. If we add new sizes in title or text, it will automatically update the acceptable size on Amount but we actually won’t have that support on Amount and that would not render as a consumer would expect. Rather its better if we add a new size to Title and then get an issue raised that Amount doesn’t support that size.
>
> Seems its safer to keep literal strings for size

So we can revert to the sizes as string literals like it was earlier

Priyanshu-C:
@divyanshu013 Done.

divyanshu013:
Thanks, could you please also take a look at the unresolved comments from last couple of reviews?

divyanshu013 (dismissed):
Looks good overall, few minor comments above.

packages/blade/src/components/Amount/Amount.tsx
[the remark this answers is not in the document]
Priyanshu-C: Okay, done!

packages/blade/src/components/Amount/Amount.tsx
Priyanshu-C: image

Only the external most container is a div.

Priyanshu-C: AmountWrapper is rendering a fragment.

packages/blade/src/components/Amount/__tests__/Amount.ssr.test.tsx
```diff
@@ -0,0 +1,10 @@
+import { Amount } from '..';
+import renderWithSSR from '~src/_helpers/testing/renderWithSSR.web';
+
+describe('<Badge />', () => {
```
chaitanyadeorukhkar, line 4:
```suggestion
describe('<Amount />', () => {
```

packages/blade/src/components/Amount/index.tsx
```diff
@@ -0,0 +1 @@
+export * from './Amount';
```
chaitanyadeorukhkar, line 1: Lets have some control over what we're exposing. We're also doing export `*` in the root index which means we won't have any control if we do * everywhere
```suggestion
export { Amount } from './Amount';
export type { AmountProps } from './Amount';
```

chaitanyadeorukhkar (dismissed):

chaitanyadeorukhkar (approved):

PR #1009, opened 2023-02-14
feat(BaseBox): Add new layout features to BaseBox

saurabhdaware:
## Description

Base for our Box component. In addition to all props that we plan to support in Box, BaseBox also supports few more things like `backgroundColor`, `borderRadius`, etc

> **Note**
>
> ~I have kept snapshot updates in separate PR #1011. Will merge that PR separately in this one once it is approved.~
> Moved snapshots to this same PR now 

## What Changed?

Things that have changed in BaseBox component compared to earlier version-
- ⚠️ `background` prop is removed (we already have `backgroundColor` and `background` was never used anywhere
- ⚠️ `padding*`, `margin*`, `height*`, `width*` props cannot have number values now (this is what we're following for `Box` so extended this to `BaseBox` as well to minimize the checks. I changed the instances in code where we were using number values since they were not too many)
- `padding` and `margin` props won't be forwarded to underlying element in React Native (they still exist in style={}) object so we can still snapshot test on them. Currently they were being repeated and after new changes they were failing on React Native because react native doesn't understand the responsive object in itself.
- They now support responsive props 🎉
  ```jsx
  <BaseBox padding={{ base: 'spacing.3', m: 'spacing.1' }} />
  ``` 
  In React Native, it picks up props in this priority -> `s > xs > base`. So in above example, React Native will add "spacing.3"
- They now support padding and margin shorthands 🕺🏻 
  ```jsx
  <BaseBox padding={["spacing.3", "10px", "spacing.6", "spacing.10"]} />
  ``` 
- BaseBox now gives autocomplete for `backgroundColor` prop
- Fixed a bug where `flex="1"` was working on web but throwing error on React Native because of its type. Now it works everywhere.

## Testing

- The chromatic snapshots did not fail for any component. 
- I did one topview check on storybook of both web and native.
- Added tests for utilities

PR #1011, opened 2023-02-15
fix(BaseBox): snapshots

saurabhdaware:
Snapshots updates from #1009. Keeping it separate since its too much 😅


packages/blade/src/components/Alert/__tests__/__snapshots__/Alert.native.test.tsx.snap
```diff
@@ -29,12 +29,10 @@ exports[`<Alert /> should render 1`] = `
 >
   <View
     display="flex"
-    marginTop="spacing.0"
```
anuraghazra, line 32: Why are these getting removed tho? we just renamed to BaseBox right?

saurabhdaware: nope. Rename PR is merged. This is snapshots updates for #1009 

I removed the margin and padding props top-level. They usually go and fail on react native when we pass `{ base: 'spacing.0', xs: '21px' }` something like this. They have a certain meaning when you put them on top-level and expect it to be string with `px` value only and can't take objects.

Anyway we only need them in styled-components to create style props so not needed on actual DOM

PR #1017, opened 2023-02-21
fix(Dropdown): infinite render onChange, positioning in flex

saurabhdaware:
Fixes 2 bugs
- [Inifinite render](https://razorpay.slack.com/archives/CMQ3RBHEU/p1676970701130639?thread_ts=1676869625.828389&cid=CMQ3RBHEU)
- [positioning in display flex container](https://razorpay.slack.com/archives/CMQ3RBHEU/p1676871892270009?thread_ts=1676869625.828389&cid=CMQ3RBHEU)

### DIFF
--- a/packages/blade/src/components/Dropdown/Dropdown.stories.tsx
+++ b/packages/blade/src/components/Dropdown/Dropdown.stories.tsx
@@ -734,7 +734,7 @@ export const WithMultipleDropdowns = (args: AllDropdownProps): JSX.Element => {
   } = args;
 
   return (
-    <BaseBox display="flex" flexDirection="row" minHeight={300}>
+    <BaseBox display="flex" flexDirection="row" minHeight={300} gap="spacing.2">
       <BaseBox flex={1}>
         <Dropdown selectionType={selectionType}>
           <SelectInput label="Top 2 design systems" {...selectInputArgs} />
--- a/packages/blade/src/components/Dropdown/Dropdown.stories.tsx
+++ b/packages/blade/src/components/Dropdown/Dropdown.stories.tsx
@@ -722,4 +722,52 @@ WithRefUsage.args = {
   name: 'design-system',
 };
 
+export const WithMultipleDropdowns = (args: AllDropdownProps): JSX.Element => {
+  const {
+    selectionType,
+    surfaceLevel,
+    title = '',
+    description,
+    value = '',
+    actionListItemIcon,
+    ...selectInputArgs
+  } = args;
+
+  return (
+    <BaseBox display="flex" flexDirection="row" minHeight={300}>
+      <BaseBox flex={1}>
+        <Dropdown selectionType={selectionType}>
+          <SelectInput label="Top 2 design systems" {...selectInputArgs} />
+          <DropdownOverlay>
+            <ActionList surfaceLevel={surfaceLevel}>
+              <ActionListItem title={title} value={value} />
+              <ActionListItem title="Primer" value="primer" />
+              <ActionListItem title="Geist" description="by Vercel" value="geist" />
+              <ActionListItem title="Airbnb Design" value="airbnb" />
+            </ActionList>
+          </DropdownOverlay>
+        </Dropdown>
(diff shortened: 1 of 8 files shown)


packages/blade/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -510,13 +510,15 @@ export const WithValueDisplay = (args: AllDropdownProps): JSX.Element => {
...
     <BaseBox minHeight={300}>
-      <Text>Selected Values: {dropdownValues}</Text>
+      <Text>Selected Values: {JSON.stringify(dropdownValues)}</Text>
       <BaseBox marginTop="spacing.5" />
       <Dropdown selectionType={selectionType}>
         <SelectInput
           label="Select Action"
-          onChange={({ values }) => {
-            setDropdownValues(values.join(', '));
+          onChange={({ name, values }) => {
```
saurabhdaware, line 518: it was working when state was string but it was breaking when state was object. Changed it in the story as well so we catch these issues.

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -66,7 +66,8 @@ const _SelectInput = (
 
   React.useEffect(() => {
     onChange?.({ name: props.name, values: value.split(', ') });
-  }, [value, onChange, props.name]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [value, props.name]);
```
saurabhdaware, line 70: Having `onChange` in this array was making the function trigger on parent re-renders which was causing infinite re-renders.

sudhakar-selva: [useEvent](https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md) hook might be useful here

saurabhdaware: yup

packages/blade/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -720,4 +722,52 @@ WithRefUsage.args = {
...
+            <ActionList surfaceLevel={surfaceLevel}>
+              <ActionListItem title={title} value={value} />
+              <ActionListItem title="Primer" value="primer" />
+              <ActionListItem title="Geist" description="by Vercel" value="geist" />
+              <ActionListItem title="Airbnb Design" value="airbnb" />
+            </ActionList>
+          </DropdownOverlay>
+        </Dropdown>
+      </BaseBox>
+      <BaseBox flex={1}>
```
divyanshu013, line 751: Minor: can perhaps add some spacing in between

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Dropdown/Dropdown.stories.tsx
saurabhdaware: done

saurabhdaware:
@kamleshchandnani can we merge this?

kamleshchandnani (commented):
overall LGTM, one thing I would recommend is that
1. Can we explicitly add stories for controlled vs uncontrolled?
2. What are the remaining props of dropdown/select that we aren't part of the storybook we shall add all of them.

saurabhdaware:
> Can we explicitly add stories for controlled vs uncontrolled?

We don't exactly have fully uncontrolled dropdown right now. For that we'll need `useDropdown` hook that can give functions like `selectOption` (this can be used for that keyboard bindings usecase we had discussed). 

We do have `isDefaultSelected` prop that sets default selection (I have added it in #1020. It was missing)

> What are the remaining props of dropdown/select that we aren't part of the storybook we shall add all of them.

Added it here #1020 (this is a bit of temporary solution though. not fully satisfied with docs right now. I'll fix it all when I am adding Button trigger and other things to docs)

PR #1016, opened 2023-02-21
feat(styled-props): add styled props to components

saurabhdaware:
## Description

This PR adds styled-props to all our components

### TODO
- [x] Add Styled Props to Dropdown (blocked on #1017 since that adds the BaseBox around Dropdown)
- [x] Add Styled Props to Icons #1021
- [x] Validate on React Native

## How to add Style Props to component 

_Also added this to jsdoc of `getStyledProps` function_

- Extend component's props with `StylePropsType`
```ts
import type { StyledProps } from '~utils';

type AlertProps = { /* your props */ } & StyledProps 
```
- Wrap your component in `BaseBox` or if it already has wrapper, extend with the `styled(BaseBox)` instead of div
- Import `getStyleProps` utility and pass its return values in wrapper's props.
```tsx
import { getStyledProps } from '~utils';

// .. your code
<BaseBox {...getStyledProps(props)}>{ /* Code */ }</BaseBox>

// OR if you already have a wrapper component

const StyledAlertWrapper = styled(BaseBox)` styles `;
<StyledAlertWrapper {...getStyledProps(props)}>{/* Code */}</StyledAlertWrapper>

// Or if you don't want to add a new wrapper
import { useStyledProps } from '~utils';

const StyledTextWrapper = styled.p((props) => {
  const styledPropsCSSObject = useStyledProps(props);
  return styledPropsCSSObject;
});
```

PR #1018, opened 2023-02-23
feat(styled-props): update snapshots

saurabhdaware:
The gods of snapshots have spoken 🤲🏼 

**Stack**

- #1016 
- #1021 
- #1018 (this)


packages/blade/src/components/Badge/__tests__/__snapshots__/Badge.native.test.tsx.snap
```diff
@@ -49,75 +49,83 @@ exports[`<Badge /> should render Badge with Icon 1`] = `
...
-        bbHeight="12px"
-        bbWidth="12px"
-        fill="none"
-        focusable={false}
-        height="12px"
-        importantForAccessibility="no-hide-descendants"
-        meetOrSlice={0}
-        minX={0}
-        minY={0}
+      <View
```
chaitanyadeorukhkar, line 52: I think we should avoid adding a wrapper to every icon. Can't we inject styling directly into the SVG element?

saurabhdaware: Made the change in other PR

PR #1019, opened 2023-02-23
refactor: migrate to size tokens

divyanshu013:
Migrate hard coded height and width sizes to size tokens.

- Using static tokens because right now there isn't much co-relation between mobile and desktop values
- Created the scale based on my understanding, keeping it similar to the style followed for colors so we can add new values easily without breaking changes later. See [packages/blade/src/tokens/global/sizes.ts](https://github.com/razorpay/blade/pull/1019/files#diff-90190416748a6d8e7a7ec95d01226fa4fea177a4593df5a9f249a46876579bbd).
- Not exposed as part of theme for consumers (this will only be internal at the moment)


packages/blade/src/tokens/global/sizes.ts
```diff
@@ -0,0 +1,79 @@
+export type Sizes = Readonly<{
+  /** 0: 0px */
+  0: 0;
+  /** 25: 1px */
+  25: 1;
+  /** 50: 2px */
+  50: 2;
+  /** 75: 3px */
+  75: 3;
```
saurav12, line 9: @divyanshu013 Can we know where are we using `3px` as a size token?

divyanshu013: In radio icon [dot radius](https://github.com/razorpay/blade/pull/1019/files#diff-6486bcbed9d9e376d3529b6ff2454df2dffc7d6e00f67dac8d2f03895a7dbd49R27)

packages/blade/src/tokens/global/sizes.ts
```diff
@@ -0,0 +1,79 @@
...
+  /** 1000: 40px */
+  1000: 40;
+  /** 1200: 48px */
+  1200: 48;
+  /** 2000: 100px */
+  2000: 100;
+  /** 2050: 120px */
+  2050: 120;
+  /** 4000: 300px */
+  4000: 300;
```
saurav12, line 45: @divyanshu013 Where is `300` used?

divyanshu013: Only at [one place](https://github.com/razorpay/blade/pull/1019/files#diff-8ef63ee40c1047f4d5138dd1288573ae858afdbcd5b8bc44b6aa8d3d97a82dc0R8), for list wrapper max height.

saurav12: Got it! Good to go it seems.

packages/blade/src/tokens/global/sizes.ts
```diff
@@ -0,0 +1,79 @@
...
+  /** 6100: 584px */
+  6100: 584;
+}>;
+
+/**
+ * Size tokens are currently not exposed for users (therefore not available in theme)
+ */
+const sizes: Sizes = {
```
chaitanyadeorukhkar, line 53: Should we call it `size` (singular) similar to `spacing`, `border`?

chaitanyadeorukhkar (commented):
Idk how scalable these tokens would be and whether we even should tokenize it. Do we allow people to override this in theme? Won't that expose us to major breakage in consistency?

Anyway, the implementation lgtm. Just want to make sure we want to do this

divyanshu013:
Right now we're not exposing as theme tokens (no usecase right now). In terms of scalability, primary intent was to get away from hard coded values across components. I think the scale will have to evolve further though, right now this exercise was done on dev side only; there aren't any such tokens on Figma.

packages/blade/src/tokens/global/sizes.ts
divyanshu013: Either works, right now we've a mix of things like `colors`,`breakpoints` are plural

packages/blade/src/tokens/global/sizes.ts
```diff
@@ -0,0 +1,79 @@
...
+  0: 0,
```
kamleshchandnani, line 54: okay now that i look at it,
* we don't have a logical scale
* we aren't exposing it to consumers

so what if we just use the same key:value pair for scale and then when we have logical scale we'll modify at that time and then expose to consumers? because without logic the internal DX is also going to get hampered that when to use what scale?

divyanshu013: Done in https://github.com/razorpay/blade/pull/1019/commits/9153397891c6d98bf242e36fe58a9eb464af3902 and https://github.com/razorpay/blade/pull/1019/commits/0cbfce49ee2bf630ee7251b3a815ce56daae1a03

packages/blade/src/tokens/global/sizes.ts
divyanshu013: Done in https://github.com/razorpay/blade/pull/1019/commits/9153397891c6d98bf242e36fe58a9eb464af3902 and https://github.com/razorpay/blade/pull/1019/commits/0cbfce49ee2bf630ee7251b3a815ce56daae1a03

kamleshchandnani (approved):
lgtm. lets close on the naming and then merge

chaitanyadeorukhkar (approved):

PR #1020, opened 2023-02-23
docs(Dropdown): add missing props from Select and ActionListItem

saurabhdaware:
- Adds missing properties for Select and ActionListItem to Dropdown stories

The documentation isn't very satisfactory to me right now. Let me fix it all when I am adding documentation for Button triggers and individual documentation of Select and ActionList.


divyanshu013 (commented):
`ActionListItem[0]` is looking bit weird in the props table

kamleshchandnani:
> * Adds missing properties for Select and ActionListItem to Dropdown stories
> 
> The documentation isn't very satisfactory to me right now. Let me fix it all when I am adding documentation for Button triggers and individual documentation of Select and ActionList.

I think you already know that this is not that great so that's a bigger thing 😄

kamleshchandnani (approved):

PR #1021, opened 2023-02-23
feat: add styled props to icons

saurabhdaware:
Same as #1016 but for icons.

This is just some find an replace query combined with some regex 🙈


divyanshu013 (approved):
Find and replace goes brrr

kamleshchandnani:
lets merge this?

Issue #811, opened 2022-11-03
`Img` component inside `Link` throw error


saurabhdaware:
Link is typed to accept only `string` and even without typescript it uses some string functions internally so breaks on runtime

Sandbox: https://codesandbox.io/s/image-in-link-bug-36puxz

```jsx
<Link href="/#">
  <img src="./logo.png" alt="example logo" />
</Link>
```

saurabhdaware:
We also can't use variables inside because it considers it as multiple strings rather than one string
```ts
<Link href="/#">Open {props.title} Recipe in CodeSandbox</Link>
```

anuraghazra:
Yeah similar problems with Button too: 

https://razorpay.slack.com/archives/G01B3LQ9H0W/p1669028285672599

Kushagra-ag:
Any updates here folks? Facing the need of this feature in website footer. [Ref](https://github.com/razorpay/blade/issues/1253)

divyanshu013:
Initiating a discussion for this

saurabhdaware:
Although not a blocker, I think we should do this-
1. I don't see much harm in opening it up. Can't think of any way it can cause inconsistency in the experience
2. Image that links to somewhere seems common usecase (E.g. clicking on logo, takes you to `/` page on most sites)


@Kushagra-ag Offtopic but for the social media icons usecase that you mentioned, maybe we can explore using our Icons instead of logo images? Can you check with designers and see what they feel about it?
<img width="238" alt="image" src="https://github.com/razorpay/blade/assets/30949385/0fa5bd2c-f500-4a80-8583-2510c3bc2b9e">

Kushagra-ag:
@saurabhdaware 
1. I can initiate discussion. Just a heads up - the icon list does not contain the linkedIn icon (But this is a low effort task I assume)
2. These type of icons would still require this functionality - 
![image](https://github.com/razorpay/blade/assets/31826645/f11baf26-756c-49be-968b-e08103098666)

saurabhdaware:
Adding slack thread about social media icons for reference - https://razorpay.slack.com/archives/CTM086NSF/p1685956826605019

Can rest of you add your opinions as well? I am more in favour of supporting this because I think atleast `children` prop should be more flexible as we move forward in design system. Its intuitive to add image or any other thing inside the children of `<Link />` (or even other components but we can discuss that later).

Kushagra-ag:
Any updates here, folks?

[...]

PR #1024, opened 2023-02-24
feat: support string array in children

saurabhdaware:
The following code now works with this PR-

```jsx
<Button>{someVariable} hello</Button>
```

Fixes the issue from this comment https://github.com/razorpay/blade/issues/811#issuecomment-1318289632

Slack Thread: https://razorpay.slack.com/archives/G01B3LQ9H0W/p1669028285672599

~Just changing it to `string | string[]` instead of `ReactText | ReactText[]` since we have several checks for strings which I didn't want to break.~ Changed it to `ReactText | ReactText[]`

### DIFF
--- a/packages/blade/src/components/Button/BaseButton/BaseButton.tsx
+++ b/packages/blade/src/components/Button/BaseButton/BaseButton.tsx
@@ -35,10 +35,11 @@ import { announce } from '~components/LiveAnnouncer';
 import type { BaseSpinnerProps } from '~components/Spinner/BaseSpinner';
 import { BaseSpinner } from '~components/Spinner/BaseSpinner';
 import BaseBox from '~components/Box/BaseBox';
-import type { DotNotationSpacingStringToken } from '~src/_helpers/types';
+import type { DotNotationSpacingStringToken, StringChildrenType } from '~src/_helpers/types';
 import type { BaseTextProps } from '~components/Typography/BaseText/types';
 import type { BladeElementRef } from '~src/hooks/useBladeInnerRef';
 import { useBladeInnerRef } from '~src/hooks/useBladeInnerRef';
+import { getStringChildrenFromArray } from '~src/utils/getStringChildren';
 
 type BaseButtonCommonProps = {
   size?: 'xsmall' | 'small' | 'medium' | 'large';
@@ -62,7 +63,7 @@ Mandatory children prop when icon is not provided
 */
 type BaseButtonWithoutIconProps = BaseButtonCommonProps & {
   icon?: undefined;
-  children: string;
+  children: StringChildrenType;
 };
 
 /*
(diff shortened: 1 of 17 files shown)


.changeset/shy-peaches-design.md
```diff
@@ -0,0 +1,10 @@
+---
+"@razorpay/blade": patch
+---
+
+feat: support string array in children
+
+Users can now use dynamic variables inside children and don't have to wrap it around with string literals
+```jsx
+<Text>{someVariable} hello</Text>
```
saurabhdaware, line 9:
```suggestion
<Button onClick={}>{someVariable} hello</Button>
```

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -62,15 +63,15 @@ Mandatory children prop when icon is not provided
 */
 type BaseButtonWithoutIconProps = BaseButtonCommonProps & {
   icon?: undefined;
-  children: string;
+  children: StringChildrenType;
```
anuraghazra, line 66: Forgot to change the type in the actual public Button.

anuraghazra:
Shouldn't we allow this? Seems like a valid usecase & could be an annoyance for cases where we want to interpolate numbers. 

image

anuraghazra:
TODO: 
- Test on RN

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
saurabhdaware: ops yes. Added.

saurabhdaware:
Working in React Native.

Also changed from `string | string[]` to `React.ReactText | React.ReactText[]`

image

packages/blade/src/utils/getStringChildren/getStringChildren.ts
```diff
@@ -0,0 +1,17 @@
+import type { StringChildrenType } from '~src/_helpers/types';
+
+const getStringFromReactText = (
+  childReactText: StringChildrenType | undefined,
+): string | undefined => {
+  if (!childReactText) {
```
anuraghazra, line 6: Let's check for actual undefined. 

Or else, This will fail
```
<BadgeComponent>{0}</BadgeComponent>
```

anuraghazra (dismissed):

packages/blade/src/utils/getStringChildren/getStringChildren.ts
saurabhdaware: Oh right. good catch 👯

saurabhdaware: done

chaitanyadeorukhkar (approved):

PR #1028, opened 2023-02-28
fix: change import to default in package exports

cseas:
The changes made in #970 caused Jest tests to [break][1] for Blade consumers.
This is because Jest [does not support][2] the "import" condition in package exports.

Proper support for the exports field for Jest is still an open issue:
https://github.com/browserify/resolve/issues/222

But we can fix this just by providing a "default" field which Jest does recognise. Since we're not exporting a dual package build which requires explicit `import`/`require` conditions, the default field is sufficient for all tools consuming Blade.

I've also moved the "types" to be defined first as is recommended by TypeScript. It still works right now in `master` because of a [bug in TypeScript][3]. Refer: https://arethetypeswrong.github.io/

[1]: https://razorpay.slack.com/archives/C02S42DDKU3/p1677557736749599
[2]: https://github.com/okta/okta-auth-js/issues/1268
[3]: https://github.com/microsoft/TypeScript/issues/50762

### DIFF
--- a/packages/blade/package.json
+++ b/packages/blade/package.json
@@ -28,16 +28,34 @@
   ],
   "exports": {
     "./components": {
-      "react-native": "./build/components/index.native.js",
-      "default": "./build/components/index.web.js"
+      "react-native": {
+        "types": "./build/components/index.native.d.ts",
+        "default": "./build/components/index.native.js"
+      },
+      "default": {
+        "types": "./build/components/index.d.ts",
+        "default": "./build/components/index.web.js"
+      }
     },
     "./tokens": {
-      "react-native": "./build/tokens/index.native.js",
-      "default": "./build/tokens/index.web.js"
+      "react-native": {
+        "types": "./build/tokens/index.native.d.ts",
+        "default": "./build/tokens/index.native.js"
+      },
+      "default": {
+        "types": "./build/tokens/index.d.ts",
+        "default": "./build/tokens/index.web.js"
+      }
     },
     "./utils": {
-      "react-native": "./build/utils/index.native.js",
-      "default": "./build/utils/index.web.js"
+      "react-native": {
+        "types": "./build/utils/index.native.d.ts",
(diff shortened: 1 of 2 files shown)


kamleshchandnani:
have you tested with all the P&C as you did before? Want to be sure that this doesn't break what was fixed as part of #970

packages/blade/package.json
```diff
@@ -29,32 +29,32 @@
   "exports": {
     "./components": {
       "react-native": {
-        "import": "./build/components/index.native.js",
-        "types": "./build/components/index.native.d.ts"
+        "types": "./build/components/index.native.d.ts",
+        "default": "./build/components/index.native.js"
```
kamleshchandnani, line 33: can you link what is the diff between using `default` and `import` here?

cseas: Whenever Blade is used with `import` (e.g. `import { Button } from "@razorpay/blade/components";`), the tool processing this import (bundler/IDE/test) will look for the "import" condition in Blade's `exports`. This is true for all tools that support properly resolving the "exports" field.

Jest only supports "browser", "require" or "default" as mentioned in https://github.com/okta/okta-auth-js/issues/1268#issue-1331894269, I couldn't find a docs link explaining this. But Jest's support for the "exports" field is still an open issue: https://github.com/browserify/resolve/issues/222

The difference is that `import` will only work if the tool processing it supports it, jest doesn't. The default field instead works as "if nothing else resolves before this, finally resolve to this". If a "default" is not provided, the resolution of the package will fail if the tool doesn't support any of the other defined keys.

kamleshchandnani: found it [here](https://nodejs.org/api/packages.html#subpath-exports). Now i understand why jest fails to import because jest still doesn't support importing esm things so we could have done `require` and `import` so jest could use require `export` path and webpack can do `import` export path. but its fine for now since we don't have cjs and esm builds

kamleshchandnani (approved):

PR #1031, opened 2023-03-01
feat(theme): add name property to watch over theme changes

saurabhdaware:
> **Note**
>
> Will merge this PR separately after Layouts. Not that important right now and it can wait. You can review the PR though.

Working Sandbox: https://codesandbox.io/s/blade-theme-change-r736xm

Imagine a usecase where you want to watch when theme is changed. 

This is how you will currently do it.
```ts
const { theme } = useTheme();
React.useMemo(() => {/* something when theme is switched */}, [theme]);
```

This will likely not work as expected in most of the usecases because theme is a deep object.

The proposed solution is this -
```ts
const { theme } = useTheme();
React.useMemo(() => {/* something when theme is switched */}, [theme.name]);
```

So now you only watch when theme itself is switched from state. 

Also helpful for doing any conditional checks in the code like-
```ts
const { theme } = useTheme();
let assets = paymentAssets;
if (theme.name === 'bankingTheme') {
  assets = bankingAssets;
}
```

Although we don't have an actual usecase of switching themes right now. This is something we've been using in our Dashboard recipe in storybook. And the theme switch might not work as expected with something like useMemo that we use in Box. We can add theme object directly to useMemo but the JSON.stringify can get expensive with that.

### DIFF
--- a/packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.web.ts
+++ b/packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.web.ts
@@ -2,6 +2,7 @@ import type { Theme } from '~components/BladeProvider';
 import { paymentTheme } from '~tokens/theme';
 
 const paymentLightTheme: Theme = {
+  name: 'paymentTheme',
   colors: { ...paymentTheme.colors.onLight },
   border: { ...paymentTheme.border },
   motion: { ...paymentTheme.motion },
--- a/packages/blade/src/components/Box/BaseBox/useMemoizedStyles.web.ts
+++ b/packages/blade/src/components/Box/BaseBox/useMemoizedStyles.web.ts
@@ -12,7 +12,7 @@ const getMemoDependency = (props: BaseBoxProps & { theme?: Theme }): string | Ba
     // I know this looks illegal but Dan approves it - https://twitter.com/dan_abramov/status/1104414272753487872?s=20
     // A lot of times passing object to dependency prop isn't efficient and mostly won't work
     // Hence we JSON.strinfigy and pass the string as dependency prop. React handles this way better.
-    dependencyPropString = JSON.stringify(rest);
+    dependencyPropString = `${JSON.stringify(rest)}-${theme?.name}`;
   } catch (err: unknown) {
     console.warn(
       '[Blade - BaseBox]: stringification of props failed in BaseBox so falling back to re-calculations on all changes\n\n If you see this warning, please create issue on https://github.com/razorpay/blade as this could degrade runtime styling performance',
--- a/packages/blade/src/components/Box/BaseBox/useMemoizedStyles.web.ts
+++ b/packages/blade/src/components/Box/BaseBox/useMemoizedStyles.web.ts
@@ -4,16 +4,19 @@ import { getBaseBoxStyles } from './baseBoxStyles';
 import type { BaseBoxProps } from './types';
 import type { Theme } from '~components/BladeProvider';
 import { useTheme } from '~components/BladeProvider';
+import type { ColorSchemeNames } from '~tokens/theme';
 
-const getMemoDependency = (props: BaseBoxProps & { theme?: Theme }): string | BaseBoxProps => {
+const getMemoDependency = (
+  props: BaseBoxProps & { theme?: Theme } & { colorScheme: ColorSchemeNames },
+): string | BaseBoxProps => {
   // These are the props that change nothing in the getBaseBoxStyles calculations
-  const { theme, children, className, id, ...rest } = props;
+  const { theme, colorScheme, children, className, id, ...rest } = props;
   let dependencyPropString: string | BaseBoxProps = '';
   try {
     // I know this looks illegal but Dan approves it - https://twitter.com/dan_abramov/status/1104414272753487872?s=20
     // A lot of times passing object to dependency prop isn't efficient and mostly won't work
     // Hence we JSON.strinfigy and pass the string as dependency prop. React handles this way better.
-    dependencyPropString = `${JSON.stringify(rest)}-${theme?.name}`;
+    dependencyPropString = `${JSON.stringify(rest)}-${theme?.name}-${colorScheme}`;
   } catch (err: unknown) {
     console.warn(
       '[Blade - BaseBox]: stringification of props failed in BaseBox so falling back to re-calculations on all changes\n\n If you see this warning, please create issue on https://github.com/razorpay/blade as this could degrade runtime styling performance',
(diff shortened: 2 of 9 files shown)


packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.web.ts
```diff
@@ -2,6 +2,7 @@ import type { Theme } from '~components/BladeProvider';
 import { paymentTheme } from '~tokens/theme';
 
 const paymentLightTheme: Theme = {
+  name: 'paymentTheme',
```
divyanshu013, line 5: Should we also include the color scheme just like in file name?

packages/blade/src/tokens/theme/theme.d.ts
```diff
@@ -155,6 +155,7 @@ export type Colors = {
 export type ColorsWithModes = Record<ColorSchemeModes, Colors>;
 
 export type ThemeTokens = {
+  name: 'paymentTheme' | 'bankingTheme' | (string & Record<never, never>); // Can be used to watch over state changes between theme without watching over entire theme object
```
divyanshu013, line 158: Interesting, what does `| (string & Record<never, never>)` do?

anuraghazra: This makes the union accept "any" string but also shows intellisense. 

Say you have `createTheme()` function which takes a name arg as `string` you want it to be "paymentTheme | bankingTheme | string"

```ts
type Name = "paymentTheme" | "bankingTheme" | string;

function createTheme(name: Name) {}

createTheme("bankingTheme")
createTheme("or-any-random-name");
```

but now users won't get the autocomplete for the `bankingTheme | paymentTheme` in the createTheme(), this happens because `string` is a super type of both  literal string `bankingTheme` & `paymentTheme` so typescript compiler simplifies the type to just `string` 

image


But `adding (string & {})` makes it so that the union can accept any string but also tricks the typescript compiler to show the intellisense 

image

[TSPlayground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAcghgW2gXigIjHESB2wAqAFhEmlAD7oBGcOA1gJY4DmRJEZlAFAM7ABOTZlABkUAN4BfAJQBuAFDyAZgFccAY2AMA9jijr+EOMAhskXHIggAuWFekTJi+QHoXUEBBABCKHBXA2uraCGAANhAm8gZGJmYQXGhocq7uUOkZGQB6APxQmNgQePEUUDT0QvGKMcamxOZoVAEAtLQgzfy0ACYhzZakKUA)


`(string & {})` works because when you intersect a the type `string` with the empty object `{}` typescript won't be able to simplify the the type to just `string`

divyanshu013: TIL 🤯

packages/blade/src/components/Box/BaseBox/useMemoizedStyles.web.ts
```diff
@@ -9,7 +9,7 @@ const getDependencyProps = (props: BaseBoxProps & { theme?: Theme }): string | B
   const { theme, children, className, id, ...rest } = props;
   let dependencyPropString: string | BaseBoxProps = '';
   try {
-    dependencyPropString = JSON.stringify(rest);
+    dependencyPropString = `${JSON.stringify(rest)}-${theme?.name}`;
```
chaitanyadeorukhkar, line 12: So we need to update the memoization of our Box component whenever something is changed in the theme for example spacing tokens, right?

packages/blade/src/tokens/theme/theme.d.ts
chaitanyadeorukhkar, line 158: Should we add a comment addressing why we've added `Record<never, never>`? Would be useful for future contributors when they come across this

chaitanyadeorukhkar: So cool!

cseas: It's an open issue in TypeScript so it'd be great to link to this in a code comment for clarity:
https://github.com/microsoft/TypeScript/issues/29729

packages/blade/src/components/Box/BaseBox/useMemoizedStyles.web.ts
saurabhdaware: With this approach, This won't exactly get triggered if someone takes the object and changes something manually in between in the state (I hope no one does that 🙈). It will only work when someone switches the themeToken itself.

packages/blade/src/components/BladeProvider/__tests__/paymentLightTheme/paymentLightTheme.web.ts
saurabhdaware: colorScheme is not a static value. It can be accessed inside react component only using `useTheme`.

Token is a static object which has things like `onLight`, `onDark`. Now while we are defining the token object, we can't tell what is going to be the colorScheme.

If someone wants to watch over colorScheme, they can do `const { colorScheme } = useTheme()` and watch over that variable since it is just string.

divyanshu013: So the file is named `paymentLightTheme` but we're naming the theme `paymentTheme`. I feel both should be same. Say for `paymentDarkTheme`, would we use the same theme name? In terms of look and feel the visuals would be different 🤔

saurabhdaware: Oh this is limited for testing. Its basically a testing utility that returns paymentTheme with all the `onLight` tokens manually.
image

So we can use this utility when we want to pass theme object to some function (e.g. in a lot of helper functions of Box I am passing theme object so in testing, I use this file)

But the actual theming tokens send both `onLight` and `onDark` tokens and then decide which one to use based on `colorScheme` later.

saurabhdaware: We might need some restructuring there though. I plan to explore this in my performance debugging task. We currently have 2x the theme bundle than it ideally should be 🤔

packages/blade/src/tokens/theme/theme.d.ts
divyanshu013: Right now since we only support these two themes, should we remove accepting any other string value?

divyanshu013 (approved):
Looks good to me.

packages/blade/src/tokens/theme/theme.d.ts
saurabhdaware: I kept it for `overrideTheme` So if someone is overriding the theme, they can name it something different.

kamleshchandnani (approved):

packages/blade/src/tokens/theme/theme.d.ts
saurabhdaware: done. Also created a type utility that can be reused.

chaitanyadeorukhkar (dismissed):

chaitanyadeorukhkar (approved):

kamleshchandnani (approved):

kamleshchandnani (approved):

PR #1034, opened 2023-03-02
chore: add `testID` to all components  for react native

chaitanyadeorukhkar:
- Adds `testID` to all components (adds `testID` for native and `data-testid` for web.
- Added some missing meta attributes to `VisuallyHidden` & `Radio` components


packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -55,7 +56,8 @@ const getTextProps = <T extends { variant: TextVariant }>({
   weight,
   size,
   contrast,
-}: Pick<TextProps<T>, 'type' | 'variant' | 'weight' | 'size' | 'contrast'>): Omit<
+  textAlign,
+}: Pick<TextProps<T>, 'type' | 'variant' | 'weight' | 'size' | 'contrast' | 'textAlign'>): Omit<
```
anuraghazra, line 60: textAlign added?

packages/blade/src/utils/metaAttribute/metaAttribute.web.ts
```diff
@@ -1,6 +1,9 @@
-const metaAttribute = (name: string, value: string | undefined): Record<string, string> => {
-  if (!value) return {};
-  return { [`data-${name}`]: value };
+import { MetaConstants } from './metaConstants';
+
+const metaAttribute = ({ name }: { name?: string; testID?: string }): Record<string, string> => {
```
divyanshu013, line 3: I think we should add for web too, in favor of parity. Otherwise this would show in TS suggestions but not be actually applied (which sounds more confusing to me)

anuraghazra (approved):
LGTM

packages/blade/src/utils/metaAttribute/metaAttribute.web.ts
chaitanyadeorukhkar: Hmm that's fair too. I'll add it as `data-tesid`? ref: https://testing-library.com/docs/queries/bytestid/

.changeset/odd-jars-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
```
divyanshu013, line 2: Should be minor imo (new feature)?
```suggestion
"@razorpay/blade": minor
```

.changeset/odd-jars-brake.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+chore: add `testID` prop to all components for react native
```
divyanshu013, line 5:
```suggestion
feat: add `testID` prop to all components
```

packages/blade/src/utils/metaAttribute/metaAttribute.native.ts
```diff
@@ -1,5 +1,7 @@
-const metaAttribute = (_name: string, _value: string): Record<string, string> => {
-  return {};
+const metaAttribute = ({ testID }: { name?: string; testID?: string }): Record<string, string> => {
```
divyanshu013, line 1: Could you also add a test for this please?

packages/blade/src/utils/metaAttribute/metaAttribute.web.ts
```diff
@@ -1,6 +1,16 @@
...
+const metaAttribute = ({
+  name,
+  testID,
+}: {
+  name?: string;
+  testID?: string;
+}): Record<string, string> => {
+  return {
+    ...(name ? { [`data-${MetaConstants.Component}`]: name } : {}),
+    ...(testID ? { [`data-testid`]: testID } : {}),
```
divyanshu013, line 12: A test for this as well please, asserting test id is added if provided and otherwise not existing.

divyanshu013 (commented):
Looks good overall. Suggestion for adding some tests.

packages/blade/src/components/Typography/Text/Text.tsx
chaitanyadeorukhkar: Good catch, was for a different PR. Removed

packages/blade/src/utils/metaAttribute/metaAttribute.native.ts
chaitanyadeorukhkar: Yup, adding tests. While adding tests to some components I realised that just spreading it as a meta attribute might not be enough since some of these components internally have their own .web and .native components and I would need to pass it further down. For example on Button we have StyledButton.native where I need to pass it down. Adding tests to all components for this

divyanshu013: Nice find

packages/blade/src/components/Checkbox/Checkbox.tsx
```diff
@@ -165,6 +167,7 @@ const _Checkbox: React.ForwardRefRenderFunction<BladeElementRef, CheckboxProps>
   const _hasError = _validationState === 'error';
   const _isDisabled = isDisabled ?? groupProps?.isDisabled;
   const _name = name ?? groupProps?.name;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
chaitanyadeorukhkar, line 170: Not sure why this started failing suddenly. I've ignored it instead of making it the following to avoid breaking changes and unintended behaviour changes

```diff
-  const _isChecked = isChecked ?? groupProps?.state?.isChecked(value!);
+  const _isChecked = isChecked ?? groupProps?.state?.isChecked(value);
```

packages/blade/src/components/Checkbox/CheckboxGroup/useCheckboxGroup.ts
```diff
@@ -45,6 +45,7 @@ const useCheckboxGroup = ({
   const [checkedValues, setValue] = useControllableState({
     value,
     defaultValue: defaultValue || [],
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
chaitanyadeorukhkar, line 48: Not sure why this started failing suddenly. I've ignored it instead of making it the following to avoid breaking changes and unintended behaviour changes

```diff
- onChange: (values: string[]) => onChange?.({ values, name: name! }),
+ onChange: (values: string[]) => onChange?.({ values, name: name }),
```

packages/blade/src/components/Checkbox/Checkbox.tsx
anuraghazra: Weird, should be fine to ignore. maybe some config or package got updated and tslint started throwing this.

anuraghazra (dismissed):

divyanshu013 (approved):
Looks good to me.

PR #1038, opened 2023-03-06
experiment: styled-props approach - handling case-by-case

saurabhdaware:


saurabhdaware:
We decided to give styled-props more priority. Made the changes in #1016

PR #1039, opened 2023-03-06
chore: fix wrong link in integration docs

kkyusufk:
Since we moved examples folder inside packages, the integration guide page on story book was pointing to the wrong path which was resulting in a broken experience.


saurabhdaware (approved):
Yay thanks!

PR #1040, opened 2023-03-06
feat: add `autoCapitalize` support to `BaseInput`, `TextInput` & `PasswordInput`

chaitanyadeorukhkar:
We need `autoCapitalize` as an explicit prop since native doesn't automatically turn off autocapitalize when type of the input is email

- Adds `autoCapitalize` prop to BaseInput & TextInput
- Sets `autoCapitalize` as `none` whenever `type` of input is `email` or `url` 
-  Sets `autoCapitalize` as `none` for `PasswordInput`


saurabhdaware:
Shouldn't we set autoCapitalize as none for everything by default? what is the usecase where you would expect autoCapitalize? 

Also what was the behaviour so far? did it not capitalize anywhere?

chaitanyadeorukhkar:
> Shouldn't we set autoCapitalize as none for everything by default? what is the usecase where you would expect autoCapitalize?
> 
> Also what was the behaviour so far? did it not capitalize anywhere?

Most text inputs on native we keep it as `undefined` which defaults to `sentences` for native. If we make it `none` then it would have unintended consequences for native

divyanshu013 (dismissed):
Looks good to me.

saurabhdaware (approved):
Looks good to me

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -229,6 +229,10 @@ export type BaseInputProps = FormInputLabelProps &
      * true if popup is in expanded state
      */
     isPopupExpanded?: boolean;
+    /**
+     * sets the autocapitalize behavior for the input
+     */
+    autoCapitalize?: 'none' | 'sentences' | 'words' | 'characters';
```
anuraghazra, line 235: Is there a difference between setting "none" & undefined (not present) ?

anuraghazra (approved):

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
chaitanyadeorukhkar: Yup, none will turn off autoCapitalize. `undefined` will reset it to its default behaviour. On mweb, when the type is email it will turn off autocapitalize but on native it will leave it on. passing it explicitly will turn  it off for all platforms to maintain parity

PR #1042, opened 2023-03-07
feat(Input): add onSubmit prop for native

chaitanyadeorukhkar:
- Adds `onSubmit` prop only for react-native. For web onSubmit actions are handled via a form.
- We need this on react-native to handle use-cases like logging in a user once they hit done on the keyboard and sending analytics events


anuraghazra (dismissed):
LGTM

packages/blade/tsconfig-typecheck.web.json
```diff
@@ -6,7 +6,9 @@
     "**/**.stories.tsx",
     "**/**.stories.internal.tsx",
     "**/**.native.ts",
-    "**/**.native.tsx"
+    "**/**.native.tsx",
+    "**/**.native.test.ts",
```
anuraghazra, line 10: you were facing type errors on the test files before this 🤔

anuraghazra (approved):

packages/blade/tsconfig-typecheck.web.json
chaitanyadeorukhkar: Yeah. Only when we check types with `.native` as module suffix

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -233,7 +239,21 @@ export type BaseInputProps = FormInputLabelProps &
...
+      /**
+       * The callback function to be invoked when the value of the input field is submitted.
+       */
+      onSubmit?: FormInputOnEvent;
+    };
+    web: {
+      /**
+       * This is a react-native only prop and has no effect on web.
+       */
+      onSubmit?: undefined;
```
saurabhdaware, line 254: Is there any way to hide this from autocomplete itself 🤔 ?

packages/blade/src/components/Input/TextArea/TextArea.tsx
```diff
@@ -26,6 +26,7 @@ type TextAreaProps = Pick<
   | 'onChange'
   | 'onFocus'
   | 'onBlur'
+  | 'onSubmit'
```
saurabhdaware, line 29: Do we not need this on OTPInput, SelectInput?

.changeset/healthy-ties-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": minor
+---
+
+feat(Input): add onSubmit prop on BaseInput, TextInput, TextArea, OTPInput & PasswordInput for react-native
```
saurabhdaware, line 5: I don't see any change on OTPInput files? 🤔

packages/blade/src/components/Input/BaseInput/__tests__/BaseInput.native.test.tsx
```diff
@@ -149,6 +149,28 @@ describe('<BaseInput />', () => {
     expect(onBlur).toHaveBeenCalledWith({ name: 'name', value: userName });
   });
 
+  it('should handle onSubmit', () => {
+    const placeholder = 'First Last';
+    const onSubmit = jest.fn();
+    const userName = 'Divyanshu';
```
saurabhdaware, line 155: Divyanshu is our john doe 😆

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
chaitanyadeorukhkar: That would be quite chaotic since we reuse BaseInput in TextInput and other components where we do pass onSubmit, there its value is a function or undefined, if we completely remove it as a prop from web then it will not even accept undefined as a valid value

packages/blade/src/components/Input/TextArea/TextArea.tsx
chaitanyadeorukhkar: Not that I can think of. OTPInput has `onOTPFilled` which is called when all input fields are filled and it won't need onSubmit. SelectInput, I can't think of a use-case where onSubmit would be needed. We can always add it when use-case comes up

.changeset/healthy-ties-change.md
chaitanyadeorukhkar: I'll remove it

saurabhdaware (approved):
looks good to me

PR #1044, opened 2023-03-09
fix: lodash bundle-size

saurabhdaware:
This import was adding entire lodash bundle. There are some issues with tree-shaking which I am figuring out but this was a quickfix

### Before - 47kb

image

### After - 23kb

image

### DIFF
--- a/.changeset/fifty-penguins-tickle.md
+++ b/.changeset/fifty-penguins-tickle.md
@@ -1,5 +1,5 @@
 ---
-"@razorpay/blade": patch
+'@razorpay/blade': patch
 ---
 
-fix: lodash tree shaking. Reduce lodash bundle size from 47kb -> 23kb
+fix: lodash tree shaking to reduce effective bundle-size.
--- a/packages/blade/src/utils/getStringChildren/getStringChildren.ts
+++ b/packages/blade/src/utils/getStringChildren/getStringChildren.ts
@@ -1,5 +1,5 @@
-import { isEmpty } from 'lodash';
 import type { StringChildrenType } from '~src/_helpers/types';
+import { isEmpty } from '~utils';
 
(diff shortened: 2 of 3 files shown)


packages/blade/src/utils/getStringChildren/getStringChildren.ts
```diff
@@ -1,5 +1,5 @@
-import { isEmpty } from 'lodash';
```
kamleshchandnani, line 1: 😢

.changeset/fifty-penguins-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix: lodash tree shaking. Reduce lodash bundle size from 47kb -> 23kb
```
divyanshu013, line 5: Let's instead reword - reduce effective bundle size ... :)

packages/blade/src/utils/getStringChildren/getStringChildren.ts
divyanshu013, line 1: We probably need lodash esm or webpack lodash plugin if we don't want to do `'lodash/isEmpty'`

divyanshu013 (dismissed):
Looks good to me.

packages/blade/src/utils/getStringChildren/getStringChildren.ts
saurabhdaware: yup. I think we're anyway re-exporting it. auto-import imported it and I didn't notice earlier ;__;

We can have some eslint rules to disable these imports.

kamleshchandnani (dismissed):

anuraghazra (approved):

kamleshchandnani (approved):

PR #1045, opened 2023-03-10
fix: tree-shaking in blade components

### LINKED ISSUE #959: Performance Issue: Doesn't support tree shaking
## Problem

Blade ships all components into one flat file (`@razorpay/blade/build/components/index.web.js`) which bundlers in consumer projects cannot tree shake. To support tree shaking, Blade needs to ship with the original tree structure of the project.

For example, if a user imports one component from Blade (say Button), they will get **all** the components & utils from Blade in their final build. This hurts performance of the applications built with Blade since the bundles delivered to the browsers/apps are much bigger than they need to be.

See example [main.bundle.js][1] built with Blade by using one Button.

## Suggestion

- Use preserveModules option from Rollup to retain original tree in the bundled package.
- Add sideEffects field to package.json to signal to bundlers to use tree shaking.
- Fix the output format of the bundle. While it does have `import/export` statements, I see it's also full of `var` declarations so we need to update the output format.

Screenshot 2023-01-17 at 9 49 46 AM

[1]: https://github.com/cseas/repro/blob/main/build/main.bundle.js

saurabhdaware:
## Description

Reduces bundle-size from 44kb to 13kb (for an App that is only using `Button` component)

Fixes #959

### Before

image

### After

image

### 😇  This creates no side-effects

```ts
const MyComponent = () => {};
MyComponent.displayName = 'MyComponent';
export { MyComponent }; 
```

### 😇  This creates no side-effects

```ts
const _MyComponent = () => {};
const MyComponent = /*#__PURE__*/ React.forwardRef(_MyComponent); // In this case, PURE comment is added by babel so we don't have to explicitly add it
// MyComponent.displayName = 'MyComponent';
export { MyComponent }; 
```

### 👿 This CREATES side-effects 

Here, it considers the `.displayName` assignment on ref as a side-effect and bundles the entire file

```ts
const _MyComponent = () => {};
const MyComponent = /*#__PURE__*/ React.forwardRef(_MyComponent)
MyComponent.displayName = 'MyComponent';
export { MyComponent }; 
```

### 😇  Solution - This creates no side-effects

```ts
const _MyComponent = () => {};
const MyComponentWithRef = /*#__PURE__*/ React.forwardRef(_MyComponent)
const MyComponent = /*#__PURE__*/ Object.assign(MyComponentWithRef, {
  componentId: 'MyComponent',
});;
export { MyComponent }; 
```

Also, in ActionList's case, having multiple components using each other in same file and defining `.componentId` was creating side-effects as well. Doing the same Object.assign with PURE comment fixed that as well.


packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -130,9 +129,9 @@ const ActionListSection: WithComponentId<ActionListSectionProps> = ({
   );
 };
 
-ActionListSection.componentId = componentIds.ActionListSection;
+/*#__PURE__*/ Object.assign(ActionListSection, { componentId: componentIds.ActionListSection });
```
anuraghazra, line 132: 😅 this looks scary

saurabhdaware: lol check it out now. It got scarier 😆

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -443,7 +443,8 @@ const _BaseButton: React.ForwardRefRenderFunction<BladeElementRef, BaseButtonPro
   );
 };
 
-const BaseButton = React.forwardRef(_BaseButton);
-BaseButton.displayName = 'BaseButton';
+const BaseButton = /*#__PURE__*/ Object.assign(React.forwardRef(_BaseButton), {
```
divyanshu013, line 446: We can also add the explanation as comment to one particular component (probably button)

divyanshu013: We might need a lint rule to enforce this, easy to miss out. How did you even figure this out lol 😸

divyanshu013: Btw is there a babel plugin that automatically does this transformation? Seems like a common issue libraries might be running into 🤔

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -101,7 +101,8 @@ const _StyledBaseButton: React.ForwardRefRenderFunction<BladeElementRef, StyledB
   );
 };
 
-const StyledBaseButton = React.forwardRef(_StyledBaseButton);
-StyledBaseButton.displayName = 'StyledBaseButton';
+const StyledBaseButton = /*#__PURE__*/ Object.assign(React.forwardRef(_StyledBaseButton), {
```
anuraghazra, line 104: So here, what is the main factor? 

The PURE comment or the Object.assign? which one does the magic?

anuraghazra:
So here: 

Why is the Input also got bundled when we are only importing Button? 

image

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
saurabhdaware: There is this babel plugin - https://github.com/mbrowne/babel-plugin-pure-static-props. I tried this first but it didn't work. I think it works when props are on react component and not when they are on refs. Even for react components it didn't work at lots of places for some reason. Couldn't find any other plugin either.

saurabhdaware: > We might need a lint rule to enforce this, easy to miss out. 

Yes definitely! Exploring if it can be eslint rule or babel plugin.

> How did you even figure this out lol 😸

Lol it was tricky

saurabhdaware:
> Why is the Input also got bundled when we are only importing Button?

Input is getting bundled because we have this autocompleteSuggestionType array as separate variable. This might need some change in logic (maybe move this array inside component render or try something else) so I didn't get much into this right now. I'll create separate issue for input treeshaking.

image

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
saurabhdaware: The PURE comment.

You can also define component inside a function and add prop inside that and return the component with imediate function call and this would probably work.
```ts
const StyledBaseButton = /*#__PURE__*/ (() => {
  const MyComponent = () => {};
  MyComponent.displayName = 'MyComponent';
  return MyComponent;
})();
```

The PURE comment only works on function calls so `Object.assign` just seems cleaner way compared to any other option.

burhanuday:
Not sure why we are using the Pure pragma and not bundler level config? @saurabhdaware 
Since most files will be side effects free, couldnt we use the `sideEffects` property in package json? I am sure webpack has this feature though not sure about Rollup. There is an issue referencing it here - https://github.com/rollup/rollup/issues/2593#issuecomment-616208761

saurabhdaware:
sideEffects property goes on consumer's package.json usually. You can see in the issue you mentioned, rollup doesn't have sideEffects property support yet. I also tried adding sideEffects option to consumer with current latest blade but that didn't strip off Blade's bundle either 🤔 

Also I am not sure if I would prefer that because that can cause unexpected results in production unless you have a good eslint setup that stops you from adding side-effects (which I couldn't find yet). 

In my opinion, Ideal way to maintain treeshaking might be having a bundle-size check in PRs, having eslint rules to stop us from adding side-effects and a babel plugin to take care of common patterns 🤔

anuraghazra:
> Input is getting bundled because we have this autocompleteSuggestionType array as separate variable

I feel like something is wrong with our rollup setup or maybe with how the consumer end's config, because why is a variable which is defined as "const" treated as a side effect? 

If we did something like "window.someVar = []" or "var someVar = []" then I could understand that. 

Although we do transpile through babel which converts the const to var. so I think something can be done on the consumer's end. 

@saurabhdaware can you push your testing setup on the /packages/example folder?

saurabhdaware:
> I feel like something is wrong with our rollup setup or maybe with how the consumer end's config, because why is a variable which is defined as "const" treated as a side effect?

Let me check this actually. I didn't try to fix it yet and assumed it is because of that variable (don't see anything else as of now)

packages/blade/scripts/manualPureFunctions.js
```diff
@@ -0,0 +1,19 @@
+const pureFunctions = ['assignWithoutSideEffects'];
+
+/**
+ * This plugin marks defined function name as PURE. This is temporary implementation of manualPureFunctions.
+ *
+ * Once we upgrade to rollup 3.5.0, we can use treeshake.manualPureFunctions config from rollup instead of this plugin.
+ * https://rollupjs.org/configuration-options/#treeshake-manualpurefunctions
+ */
+const manualPureFunctions = () => ({
```
anuraghazra, line 9: :p why is this "manual", shouldn't this be "auto"?

anuraghazra (dismissed):
LGTM, check bundling the code and do a sanity on the examples package.

saurabhdaware:
> LGTM, check bundling the code and do a sanity on the examples package.

Yup did. Working 👯

packages/blade/scripts/manualPureFunctions.js
chaitanyadeorukhkar: Can't we upgrade our rollup version and use this?

chaitanyadeorukhkar:
I think its high time we add bundle size check on our PRs now 😅

saurabhdaware:
> I think its high time we add bundle size check on our PRs now 😅

Yes!! Definitely required!

chaitanyadeorukhkar (dismissed):

packages/blade/src/components/Card/CardHeader.tsx
```diff
@@ -25,41 +24,52 @@ import {
   makeSpace,
 } from '~utils';
 import type { TestID } from '~src/_helpers/types';
+import { assignWithoutSideEffects } from '~src/utils/assignWithoutSideEffects';
 
-const CardHeaderIcon: WithComponentId<{ icon: IconComponent }> = ({ icon: Icon }) => {
+const _CardHeaderIcon = ({ icon: Icon }: { icon: IconComponent }): JSX.Element => {
```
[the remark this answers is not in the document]
saurabhdaware: That is what we use generally for return type of React components so used the same

packages/blade/scripts/manualPureFunctions.js
saurabhdaware: we can. It's a major version bump though so didn't spend much time right now. can pick it up next quarter

chaitanyadeorukhkar (dismissed):

divyanshu013 (commented):
Looks good overall, one comment to make the return types same as earlier.

packages/blade/src/components/Card/CardHeader.tsx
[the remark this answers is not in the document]
saurabhdaware: I've been adding `JSX.Element` on all components I built 🙈 and there are few other instances of it in code. 

I reverted the change here to how it was to not create breaking changes.

Created this issue to later change all JSX.Element instances to React.ReactElement https://github.com/razorpay/blade/issues/1089

divyanshu013 (approved):
:shipit:

PR #1049, opened 2023-03-14
feat(blade): typography component streamline

anuraghazra:
Changes: 

- Added `xlarge` size for Title
  - `side-effect:` large size got moved to xlarge
- Heading `small` & `subheading` size for mobile screen would become 16px because of global font scale changes
- Added `large` size for `<Text variant="body" />`
- ~Changed body text to `semibold`~

- Updated typography lineHeight scale to use numbered values (100, 200, 300 etc)


packages/blade/src/components/Typography/Text/getTextStyles.ts
```diff
@@ -11,8 +11,9 @@ const getTextStyles = <T extends { variant: TextVariant }>({
   size,
   contrast,
   theme,
-}: Pick<TextProps<T>, 'type' | 'variant' | 'weight' | 'size' | 'contrast'> & {
+}: Pick<TextProps<T>, 'type' | 'variant' | 'size' | 'contrast'> & {
```
kamleshchandnani, line 14: why remove from here? weight is there na on textprops?

packages/blade/src/tokens/global/typography.ts
```diff
@@ -4,6 +4,7 @@ import type { FontFamily } from './fontFamily/types';
 type FontWeight = {
   regular: 400;
   bold: 700;
+  semibold: 600;
```
kamleshchandnani, line 7: is this the correct weight for semibold?

kamleshchandnani (commented):
do you want to open Link large variant in this PR only?

anuraghazra:
> do you want to open Link large variant in this PR only?

No i'll keep it separate. Let's resolve all the typography stuff first, large link will then be easier.

packages/blade/src/tokens/global/typography.ts
anuraghazra: Should be, although because we didn't found the semibold font on google fonts, not sure how should we proceed now.

anuraghazra: Even on https://fontsource.org/fonts/lato there's only 400,700,900. 

So in this case we might have to treat the black (900) as "bold" and 700 as semibold. 

@saurav12

packages/blade/src/components/Typography/Text/getTextStyles.ts
anuraghazra: Yes removed from here because in the TextProps the weight is `bold | regular` but getBaseTextStyles's types accept `bold | regular | semibold` since we added `semibold` in the global font weight token.

kamleshchandnani (approved):
LGTM overall 🎉 

Checklist before merging
- [ ] Verify the snapshots on chromatic/visually test components with the help of Design
- [ ] Finalise about semibold

packages/blade/src/components/Alert/__tests__/__snapshots__/Alert.native.test.tsx.snap
```diff
@@ -148,7 +148,7 @@ exports[`<Alert /> should render 1`] = `
               "fontSize": 12,
               "fontStyle": "normal",
               "fontWeight": "400",
-              "lineHeight": 20,
+              "lineHeight": 16,
```
chaitanyadeorukhkar, line 151: Breaking changes in native lineHeight? can you verify if all the components look okay on native once (along with all supported sizes/variants)? Would be quite hectic to check everything but I think we should check at least the components whose snapshots have been updated

anuraghazra: I've done the sanity check of all the components nothing seems to be breaking, just changes on the text sizes in few places.

chaitanyadeorukhkar (approved):
Approving this to merge with the chained PR so that we can check the UI diff

PR #1060, opened 2023-03-15
feat(Box): add support for `as` prop

saurabhdaware:
## Changes

- Add `as` property
- Add tests
- Add documentation example

### DIFF
--- a/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
+++ b/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
@@ -100,6 +100,20 @@ type BackgroundColorString<T extends ColorObjects> = `${T}.background.${DotNotat
   Theme['colors'][T]['background']
 >}`;
 
+// Created this as an array so I can reuse it for runtime validation
+const validBoxAsValues = [
+  'div',
+  'section',
+  'footer',
+  'header',
+  'main',
+  'aside',
+  'nav',
+  'span',
+] as const;
+
+type BoxAsType = typeof validBoxAsValues[number];
+
 type BaseBoxVisualProps = MakeObjectResponsive<
   {
     borderRadius: keyof Border['radius'];
@@ -117,7 +131,11 @@ type BaseBoxVisualProps = MakeObjectResponsive<
 
 type BoxVisualProps = MakeObjectResponsive<{
   backgroundColor: BackgroundColorString<'surface'>;
-}>;
+}> & {
+  // Intentionally keeping this outside of MakeObjectResponsive since we only want as to be string and not responsive object
+  // styled-components do not support passing `as` prop as an object
+  as: BoxAsType;
+};
 
 type StyledPropsBlade = Partial<
   MarginProps &
@@ -156,4 +174,4 @@ type BaseBoxProps = Omit<BoxProps, keyof BoxVisualProps> &
     }
   >;
 
-export { BaseBoxProps, BoxProps, StyledPropsBlade };
+export { BaseBoxProps, BoxProps, StyledPropsBlade, validBoxAsValues };
--- a/packages/blade/src/components/Box/Box.tsx
+++ b/packages/blade/src/components/Box/Box.tsx
@@ -108,7 +108,7 @@ const makeBoxProps = (props: BoxProps): KeysRequired<Omit<BoxProps, 'testID'>> =
     // Visual
     backgroundColor: props.backgroundColor,
     children: props.children,
-    as: props.as,
+    as: isReactNative() ? undefined : props.as, // as is not supported on react-native
   };
 };
 
--- a/packages/blade/src/components/Box/Box.tsx
+++ b/packages/blade/src/components/Box/Box.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import BaseBox from './BaseBox';
 import type { BoxProps, MakeValueResponsive } from './BaseBox/types';
+import { validBoxAsValues } from './BaseBox/types/propsTypes';
 import type { KeysRequired } from '~src/_helpers/types';
-import { metaAttribute, MetaConstants } from '~utils';
+import { isReactNative, metaAttribute, MetaConstants } from '~utils';
 
 const validateBackgroundString = (stringBackgroundColorValue: string): void => {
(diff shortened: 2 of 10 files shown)


packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -117,7 +131,11 @@ type BaseBoxVisualProps = MakeObjectResponsive<
 
 type BoxVisualProps = MakeObjectResponsive<{
   backgroundColor: BackgroundColorString<'surface'>;
-}>;
+}> & {
+  // Intentionally keeping this outside of MakeObjectResponsive since we only want as to be string and not responsive object
+  // styled-components do not support passing `as` prop as an object
+  as: BoxAsType;
```
saurabhdaware, line 137: We can't do `as={{ base: 'section', m: 'footer' }}`. Styled Components is failing. The `as` property is not handled anywhere by us. Styled Component takes care of it.

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -100,6 +100,20 @@ type BackgroundColorString<T extends ColorObjects> = `${T}.background.${DotNotat
   Theme['colors'][T]['background']
 >}`;
 
+// Created this as an array so I can reuse it for runtime validation
+const validBoxAsValues = [
```
kamleshchandnani, line 104: 🤯

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
kamleshchandnani: ye to anyways nai hona chaiye lol 🤣 

seems like an antipattern

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -100,6 +100,20 @@ type BackgroundColorString<T extends ColorObjects> = `${T}.background.${DotNotat
...
+  'section',
+  'footer',
+  'header',
+  'main',
+  'aside',
+  'nav',
+  'span',
+] as const;
+
+type BoxAsType = typeof validBoxAsValues[number];
```
kamleshchandnani, line 115: TIL this will actually give us union 🤯  

```
BoxAsType = 'div' | 'section'
```

packages/blade/src/components/Box/Box.tsx
```diff
@@ -150,6 +152,22 @@ const Box = (props: BoxProps): JSX.Element => {
     validateBackgroundProp(props.backgroundColor);
   }, [props.backgroundColor]);
 
+  React.useEffect(() => {
```
kamleshchandnani, line 155: how till this work on SSR 🤔 

can you add this on SSR test?

packages/blade/src/components/Box/Box.tsx
```diff
@@ -150,6 +152,22 @@ const Box = (props: BoxProps): JSX.Element => {
...
+    if (props.as && typeof props.as === 'string') {
```
kamleshchandnani, line 156: do we need to add `string` typeof check? we shall just do if `props.as` is not empty? maybe just this `!isEmpty(props.as)`

packages/blade/src/components/Box/__tests__/Box.web.test.tsx
```diff
@@ -34,6 +34,29 @@ describe('<Box />', () => {
     `);
   });
 
+  it('should render Box as footer tag', () => {
```
kamleshchandnani, line 37: add ssr test?

packages/blade/src/components/Box/Box.tsx
saurabhdaware: Oh it won't. It will work during hydration part of SSR rather than renderToString part. Should be fine right? it will still fail in dev-server.

packages/blade/src/components/Box/Box.tsx
saurabhdaware: yup right we didn't need that typeof check. don't think so we need isEmpty check either. Just `if (props.as)` should be enough.

packages/blade/src/components/Box/__tests__/Box.web.test.tsx
saurabhdaware: done

packages/blade/src/components/Box/Box.tsx
saurabhdaware: added test for errors in web, ssr, and native

packages/blade/src/components/Box/__tests__/__snapshots__/Box.ssr.test.tsx.snap
```diff
@@ -1,5 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`<Box /> should render Box as footer tag 1`] = `"<div id=\\"root\\" data-reactroot=\\"\\"><footer data-blade-component=\\"Box\\" display=\\"block\\" class=\\"BaseBoxweb__BaseBox-sc-1icfu8j-0 gxRbfF\\">Footer test!</footer></div>"`;
```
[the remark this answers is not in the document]
saurabhdaware: didn't get 🤔 Its not different no?  That outer div is just this div that we use to wrap all tests in `renderWithSSR`

image

saurabhdaware:
> but that will lead to complete rehydration right because react will see this as tree is different on server and client

The `useEffect` runs after the initial hydration. In this sequence -

- Server Render
- Client Hydration
- UseEffect Executes

So it doesn't break the hydration itself.

packages/blade/src/components/Box/__tests__/__snapshots__/Box.ssr.test.tsx.snap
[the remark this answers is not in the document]
kamleshchandnani: cool my concern is the error will be thrown on client if the values are not valid `as` props. ideally server pe error hona chaiye. but for now I'll merge this you can think about this.

also wondering why are we adding this in useEffect and not without it?

kamleshchandnani (approved):
lessss gooo. last PR 🎉

PR #1061, opened 2023-03-15
feat(OTPInput): add `onFocus` & `onBlur` props

chaitanyadeorukhkar:
- Adds onFocus & onBlur props.
- Product teams need this to track analytics events


packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
```diff
@@ -25,6 +25,12 @@ export default {
     onChange: ({ name, value }): void => {
       console.log(`input field ${name} content changed to ${value}`);
     },
+    onBlur: ({ name, value, inputIndex }): void => {
```
saurabhdaware, line 28: image

It is showing up outside of category.


Also I am not able to see console log when I focus on OTPInput in storybook 🤔

packages/blade/src/components/Input/OTPInput/__tests__/OTPInput.web.test.tsx
```diff
@@ -95,6 +95,48 @@ describe('<OTPInput />', () => {
...
+    const otp = '123456';
+
+    const { getAllByLabelText } = renderWithTheme(
+      <OTPInput label={label} name="otp" onFocus={onFocus} />,
+    );
+    const allInputs = getAllByLabelText(/character/);
+    await user.type(allInputs[0], otp);
+
+    // should be called for each keystroke
+    expect(onFocus).toHaveBeenCalledTimes(6);
```
saurabhdaware, line 111: Why do they want it to be called on every block focus? Shouldn't we call it once? 🤔

saurabhdaware (commented):
Mostly looks good to me. Some comments on storybook and some doubts

packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
chaitanyadeorukhkar: You have to see it in the action tab
Screenshot 2023-03-16 at 12 06 47 PM

packages/blade/src/components/Input/OTPInput/__tests__/OTPInput.web.test.tsx
chaitanyadeorukhkar: Each block is a focusable input on its own, if we call it once, which one do we call it for? What if the user focuses on some other top field (like 3rd or 4th) than the one we added the event to? That's why I'm doing the callback for every input and also passing the index of the field that's focused or blurred

packages/blade/src/components/Input/OTPInput/__tests__/OTPInput.web.test.tsx
```diff
@@ -95,6 +95,48 @@ describe('<OTPInput />', () => {
...
+    const otp = '123456';
+
+    const { getAllByLabelText } = renderWithTheme(
+      <OTPInput label={label} name="otp" onBlur={onBlur} />,
+    );
+    const allInputs = getAllByLabelText(/character/);
+    await user.type(allInputs[0], otp);
+
+    // should be called for each keystroke
+    expect(onBlur).toHaveBeenCalledTimes(5); // 5 because the last input will remain focused
```
anuraghazra, line 132: Is this expected behaviour that the user would want? 🤔 Shoudn't we treat the OTPInput as a "single input"

Since the OTPInput is considered as a "composite" of 6 smaller inputs, should the user care about all the blur events of every 6 element or should this just be called once?

chaitanyadeorukhkar: Once when? As a dev if you see 6 input fields, regardless of whether they're logically together or not, would you not expect them to have their own events?

chaitanyadeorukhkar: For instance, onChange is called for every field update, why would this logic change for onBlur and onFocus events?

Although unlikely, a user could enter the otp starting midway. Won't we still want to trigger the analytics call? 😅

packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
saurabhdaware: ooo TIL about actions tab lol. I was checking in browser console

chaitanyadeorukhkar: storybook redirects all `on*` console logs to the actions tab 😅

packages/blade/src/components/Input/OTPInput/__tests__/OTPInput.web.test.tsx
saurabhdaware: I was thinking can we put focus handlers on outer View? so if any of the fields are focussed, then we say the OTPInput is focussed. If someone clicks out, then its blur 🤔

chaitanyadeorukhkar: React Native might not behave as expected for that. View doesn't have `onBlur` or `onFocus` events, only TextInput does. Ref: https://reactnative.dev/docs/view

anuraghazra: I see, okay then keep it as is. Hopefully woudn’t cause any issues.

anuraghazra (approved):

saurabhdaware (approved):

PR #1069, opened 2023-03-16
feat: add box-sizing border-box for alert

saurabhdaware:
We added `width: 100%` for isFullWidth with position="fixed" on Alert. But we'll also need boxSizing border-box to make sure Alert stays inside the outer box. https://razorpay.slack.com/archives/G01B3LQ9H0W/p1677822733203969

Normally apps are adding `* { box-sizing: border-box }` to their global styles. This is for apps that forget to add that global style.

### DIFF
--- a/.changeset/red-badgers-tease.md
+++ b/.changeset/red-badgers-tease.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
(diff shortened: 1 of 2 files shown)


saurabhdaware:
Currently it's broken for Card and other components as well. Might be worth just recommending `* { box-sizing: border-box }` to consumers. That's what most other styling libraries do

PR #1075, opened 2023-03-17
perf(blade): improve dropdown/actionlist performance

anuraghazra:
Changes: 

- Extracted out ActionListBody and memoized it to improve rerenders

**Before doing the optimization:**

- Average time to render: **36**
- ActionListItem always rerenders

image

**After doing this optimization:** 

- Average time to render: **15**
- ActionListItemBody doesn’t rerender throughout selection

image

More in [the Notion Doc](https://www.notion.so/Dropdown-ActionList-Performance-Improvements-4e35bb4673494164a19c17881fe315b9)


packages/blade/.storybook/react-native/Storybook.tsx
```diff
@@ -8,6 +8,7 @@ import { name as appName } from '../../app.json';
 
 const App = (): React.ReactElement => {
   const Storybook = getStorybookUI({
+    shouldPersistSelection: true,
```
anuraghazra, line 11: Improve RN DX by 10x. 🚀 no more storybook changing the story to first one on hot refresh.

chaitanyadeorukhkar: 10x seems less. Should be 100x 😆

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -126,4 +126,7 @@ const ActionList = ({ children, surfaceLevel = 2, testID }: ActionListProps): JS
   );
 };
 
+// TODO: Add treeshake wrapper
+const ActionList = React.memo(_ActionList);
```
chaitanyadeorukhkar, line 130: I'm a little skeptical about this. Can we very if dynamic updation of list items doesn't cause any issues?

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -46,7 +46,7 @@ const Dropdown: WithComponentId<DropdownProps> = ({
   selectionType = 'single',
   ...styledProps
 }): JSX.Element => {
-  const [isOpen, setIsOpen] = React.useState(false);
+  const [isOpen, setIsOpen] = React.useState(true);
```
chaitanyadeorukhkar, line 49: Need to revert this?

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -64,22 +64,24 @@ const Dropdown: WithComponentId<DropdownProps> = ({
...
-    if (React.isValidElement(child)) {
-      if (
-        !isValidAllowedChildren(child, 'SelectInput') &&
-        !isValidAllowedChildren(child, componentIds.DropdownOverlay)
-      ) {
-        throw new Error(
-          `[Dropdown]: Dropdown can only have \`SelectInput\` and \`DropdownOverlay\` as children\n\n Check out: https://blade.razorpay.com/?path=/story/components-dropdown`,
-        );
-      }
+  React.useEffect(() => {
```
chaitanyadeorukhkar, line 67: Won't this be an issue on SSR?

packages/blade/src/components/Dropdown/Dropdown.tsx
anuraghazra: Done.

packages/blade/src/components/ActionList/ActionList.tsx
anuraghazra: Sorry couldn't understand your concern here, can you expand on this?

chaitanyadeorukhkar (approved):
small comment. otherwise lgtm

packages/blade/src/components/ActionList/ActionList.tsx
anuraghazra: Yes I've done that.

chaitanyadeorukhkar: Toh why is there still a `// TODO: Add treeshake wrapper`?

anuraghazra: Oh wait, I think we don't need the treeshake wrapper here, because we are not assigning any static properties to the component eg: `displayName` or `componentId`

anuraghazra: Yep don't need it, rollup already adds the PURE comment: 

image

packages/blade/src/components/Dropdown/Dropdown.tsx
anuraghazra: Okay reverted this, didn't cause SSR issue but it had a reason to be outside useEffect, it needs to synchronously run on component's first mount so that we can save the `dropdownTriggerer` in a variable and set it to `SelectInput` by looping over the children, if we do this in an useEffect the dependant props will cause a flashing effect.

packages/blade/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -805,4 +805,81 @@ WithMultipleDropdowns.args = {
...
+    'Longan',
+    'Lychee',
+    'Loquat',
+    'Mango',
+    'Mandarin',
+    'Orange',
+    'Mulberry',
+  ];
+
+  console.log('total items', fruits.length);
```
divyanshu013, line 862: Stray log?

```suggestion
```

packages/blade/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -805,4 +805,81 @@ WithMultipleDropdowns.args = {
...
+  const [clickCount, setClickCount] = React.useState(0);
```
divyanshu013, line 863: Do we need this in the story?

divyanshu013 (approved):
Looks good to me. Minor comments. Also:

- [ ] Validate chromatic checks

packages/blade/src/components/Dropdown/Dropdown.stories.tsx
anuraghazra: Let's keep it, I'll need to pull in changes in the other PR. I'll remove it then.

PR #1078, opened 2023-03-17
feat/link hit slop react native

chaitanyadeorukhkar:
- feat(Link): add support for hitSlop on native
- add hitSlop to Link and move platform check to Link

### DIFF
--- a/packages/blade/src/components/Link/BaseLink/BaseLink.tsx
+++ b/packages/blade/src/components/Link/BaseLink/BaseLink.tsx
@@ -21,6 +21,7 @@ import type { DurationString, EasingString } from '~tokens/global/motion';
 import type { BaseTextProps } from '~components/Typography/BaseText/types';
 import { getStringFromReactText } from '~src/utils/getStringChildren';
 import type { StyledPropsBlade } from '~components/Box/styledProps';
+import type { Platform } from '~utils';
 
 type BaseLinkCommonProps = {
   intent?: 'positive' | 'negative' | 'notice' | 'information' | 'neutral';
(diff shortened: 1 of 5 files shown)


packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -36,6 +36,14 @@ type BaseLinkCommonProps = {
    * @default medium
    */
   size?: 'small' | 'medium';
+  hitSlop?:
```
kamleshchandnani, line 39: add jsdoc that this is native only prop?

kamleshchandnani (approved):

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
chaitanyadeorukhkar: Added

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -36,6 +36,17 @@ type BaseLinkCommonProps = {
    * @default medium
    */
   size?: 'small' | 'medium';
+  /**
+   * Defines how far your touch can start away from the link. This is a react-native only prop and has no effect on web.
+   */
+  hitSlop?:
+    | {
+        top?: number;
```
saurabhdaware, line 44: do we have to take it as a prop for consumer? can we standardize the click-area internally without taking it from consumers?

chaitanyadeorukhkar: Ideally we should be standardising this but we haven't put any design effort on this. On the PG Mobile have they have different hitSlops for different use-cases. We'll have to take up hitSlop as a wider charter. Right now this is a blocker only for Links, hence added this but we would need to explore hitSlop support for all components in the future. Unblocking the PG team with this

kamleshchandnani (approved):

PR #1081, opened 2023-03-21
feat(Box): add different types for display on react native

saurabhdaware:
We will now get TS error during react native typecheck when we write `display="block"` in common code in our repo.

Also better autocomplete for consumers on react native.

### DIFF
--- a/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
+++ b/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
@@ -18,7 +18,18 @@ type LayoutProps = MakeObjectResponsive<
     width: SpacingValueType;
     minWidth: SpacingValueType;
     maxWidth: SpacingValueType;
-  } & Pick<CSSObject, 'display' | 'overflow' | 'overflowX' | 'overflowY'>
+    /**
+     * The **`display`** CSS property sets whether an element is treated as a block or inline element and the layout used for its children, such as flow layout, grid or flex.
+     *
+     * On React Native, `display` supports `none` and `flex`
+     *
+     * @see https://developer.mozilla.org/docs/Web/CSS/display
+     */
+    display: Platform.Select<{
+      web: CSSObject['display'];
+      native: 'none' | 'flex';
+    }>;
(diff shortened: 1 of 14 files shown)


packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -18,7 +18,28 @@ type LayoutProps = MakeObjectResponsive<
...
+      };
+      native: {
+        /**
+         *
+         *
+         * On React Native, **`display`** property sets whether an element can be `flex` or `none`
+         *
+         * @see https://reactnative.dev/docs/layout-props.html#display
+         */
+        display: 'none' | 'flex';
```
chaitanyadeorukhkar, line 40: Do we need to mention what the default value would be? On Native its flex by default

chaitanyadeorukhkar (dismissed):
🚀

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
saurabhdaware: added for react native

saurabhdaware: added for react native. On web, it might depend on default value of what we pass from `as`

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -18,7 +18,30 @@ type LayoutProps = MakeObjectResponsive<
...
+      };
+      native: {
+        /**
+         *
+         *
+         * On React Native, **`display`** property sets whether an element can be `flex` or `none`
+         *
+         * @see https://reactnative.dev/docs/layout-props.html#display
+         *
+         * @defaultValue flex
```
chaitanyadeorukhkar, line 40: We use `@default` for jsdoc right? 🤔 
```suggestion
         * @default 'flex'
```

saurabhdaware: Oh yup. done.

PR #1083, opened 2023-03-21
fix(Alert, Card): box-sizing issue

saurabhdaware:
adds `box-sizing: border-box` for Card and Alert to fix spacings.

Currently storybook by default adds `* { box-sizing: border-box }` in its global styles so you won't see this issue on storybook.

In Sandbox, you can remove the global styles and there these issues can show up. 

## Before

image

image

## After

image

image


chaitanyadeorukhkar (approved):

divyanshu013 (commented):
Nice find. I think we should add this to all components (maybe a common util) because we're building all components assuming `border-box` box sizing

PR #1088, opened 2023-03-23
fix: ListItemLink alignment

chaitanyadeorukhkar:
React native has a vertical alignment bug that causes nested text which also has its own view inside to cause misalignment. This PR wraps all the children of list item (text and view) in a flex and aligns them. 

Ref: https://github.com/facebook/react-native/issues/31955

**Before this PR:**
Simulator Screen Shot - iPhone 14 Pro - 2023-03-28 at 11 07 59

**After this PR:**
Simulator Screen Shot - iPhone 14 Pro - 2023-03-28 at 11 09 24

### DIFF
--- a/packages/blade/src/components/Link/BaseLink/BaseLink.tsx
+++ b/packages/blade/src/components/Link/BaseLink/BaseLink.tsx
@@ -186,7 +186,7 @@ const getProps = ({
       isVisited,
     }) as IconProps['color'],
     fontSize: size === 'medium' ? 100 : 75,
-    lineHeight: size === 'medium' ? 'm' : 's',
+    lineHeight: size === 'medium' ? 'l' : 's',
     iconSize: size,
     iconPadding: children?.trim() ? 'spacing.2' : 'spacing.0',
     textColor: getColorToken({
--- a/packages/blade/src/components/List/ListItem.tsx
+++ b/packages/blade/src/components/List/ListItem.tsx
@@ -66,7 +66,7 @@ const ListItemContentChildren = ({
   children,
   size,
 }: {
-  children: (React.ReactChild | React.ReactFragment | React.ReactPortal)[];
+  children: React.ReactNode[];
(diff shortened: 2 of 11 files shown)


packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -186,7 +186,7 @@ const getProps = ({
       isVisited,
     }) as IconProps['color'],
     fontSize: size === 'medium' ? 100 : 75,
-    lineHeight: size === 'medium' ? 'm' : 's',
+    lineHeight: size === 'medium' ? 'l' : 's',
```
chaitanyadeorukhkar, line 189: This was an incorrect implementation 🙈. It is `l` instead of `m` on Figma as well.

packages/blade/src/components/List/ListItem.tsx
```diff
@@ -60,6 +61,36 @@ const StyledListItem = styled(ListItemElement)<{
     : 0,
 }));
 
+const ListItemContentChildren = ({
+  children,
+  size,
+}: {
+  children: (React.ReactChild | React.ReactFragment | React.ReactPortal)[];
```
anuraghazra, line 68: Any reason we kept this so specific? eg: Fragment, ReactPortal. 

Can we use React.ReactNode instead?

kamleshchandnani (approved):

packages/blade/src/components/List/ListItem.tsx
chaitanyadeorukhkar: Had done it to suppress a warning. Made it ReactNode now. Draft PR hai, have mercy 😢

kamleshchandnani (approved):

packages/blade/src/components/List/ListItem.tsx
anuraghazra: lol sorry, was reviewing this at airport :p

PR #1095, opened 2023-03-28
perf(blade): Integrate SectionList in ActionList

anuraghazra:
// TBD


packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -98,6 +101,24 @@ const _ActionList = ({ children, surfaceLevel = 2, testID }: ActionListProps): J
...
+  const renderActionListSectionHeader = React.useCallback(({ section: { title } }) => {
+    if (!title) return null;
+    return <ActionListSection title={title} _hideDivider={true} children={undefined} />;
+  }, []);
+
+  const renderActionListSectionDivider = React.useCallback(
+    ({ section: { title, hideDivider } }) => {
+      if (!title) return null;
+      if (hideDivider) return null;
+      return <ActionListSectionDivider />;
```
anuraghazra, line 117: While in the web we wrap multiple ActionListItems inside ActionListSection, in here we can't do that so the 'section header' & 'section footer' components are rendered separately. 

Basically it's a flat structure.

packages/blade/src/components/ActionList/actionListUtils.ts
```diff
@@ -45,24 +45,36 @@ const actionListAllowedChildren = [
...
 ): {
+  sectionData: SectionData;
   childrenWithId?: React.ReactNode[] | null;
   actionListOptions: OptionsType;
   defaultSelectedIndices: number[];
   actionListHeaderChild: React.ReactElement | null;
   actionListFooterChild: React.ReactElement | null;
 } => {
+  const sectionData: SectionData = [];
+  let currentSection: string | null = null;
```
anuraghazra, line 68: If currentSection is null, we don't render SectionListHeader

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -112,13 +133,25 @@ const _ActionList = ({ children, surfaceLevel = 2, testID }: ActionListProps): J
     >
       {actionListHeaderChild}
       <StyledListBoxWrapper
+        sections={sectionData}
+        windowSize={5}
+        initialNumToRender={5}
+        bouncesZoom={false}
+        bounces={false}
```
chaitanyadeorukhkar, line 140: Lets keep bounces as true? That's the expected behaviour on iOS

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -112,13 +133,25 @@ const _ActionList = ({ children, surfaceLevel = 2, testID }: ActionListProps): J
     >
       {actionListHeaderChild}
       <StyledListBoxWrapper
+        sections={sectionData}
+        windowSize={5}
+        initialNumToRender={5}
```
chaitanyadeorukhkar, line 138: Lets keep it as default values that RN has instead of us passing it?

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -112,13 +133,25 @@ const _ActionList = ({ children, surfaceLevel = 2, testID }: ActionListProps): J
...
+        bouncesZoom={false}
+        bounces={false}
+        keyExtractor={(item: any) => {
+          return item.value;
+        }}
```
chaitanyadeorukhkar, line 143: Not the best practice to use index as key especially for use-cases where item's indexes keep changing. Unlikely in our case but not impossible (would change during typeahead). Do we know what the value of `item.value` is going to be? I think it expects a number/string only
```suggestion
        keyExtractor={(item: any, index: number) => {
          return index;
        }}
```

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -1,8 +1,10 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
chaitanyadeorukhkar, line 1: The entire implementation of ActionList seems too messy at the moment. We're handling RN as well as web specific things here. I think its better if we create `ActionList.native.tsx` & `ActionList.web.tsx`. Both of them have very different implementations anyway, better to keep them isolated. We can share certain things like styling via common functions. What do you think?

packages/blade/src/components/ActionList/actionListUtils.ts
```diff
@@ -71,6 +83,27 @@ const getActionListProperties = (
       });
       const currentIndex = actionListOptions.length - 1;
 
+      const foundSection = sectionData.find((v) => v.title === currentSection);
+      // push the item in the appropriate bucket
+      if (foundSection) {
+        foundSection?.data.push({
```
chaitanyadeorukhkar, line 89: We're pushing data  into `foundSection` but where are we returning it to? I can see in the else branch, we're pushing data into `sectionData` and returning `sectionData`  to the caller. What we we doing with `foundSection`?

packages/blade/src/components/ActionList/ActionList.tsx
anuraghazra: Bounces messes up with the Section list headers in iOS

https://user-images.githubusercontent.com/35374649/231438994-2456b83e-c1f4-4628-b903-00107b37c3e2.mov

packages/blade/src/components/ActionList/ActionList.tsx
anuraghazra: 🤔 not using the index, I'm using item.value only which is the value which user passes to the ActionListItem

packages/blade/src/components/ActionList/actionListUtils.ts
anuraghazra: foundSection is not the actual data we represent, it's a reference to a single item from `sectionData` which we find and push the data into to build a data structure like this: 

```json
[
  { // <-------- this whole object is what "foundSection" references to, 
                 we find it via the `title` & push more items into the `data` array
    "title": "Recent 1",
    "hideDivider": false,
    "data": [
      {
        "title": "Avocados",
        "value": "Avocados",
        "_index": 4
      },
      {
        "title": "Bananas",
        "value": "Bananas",
        "_index": 5
      },
      {
        "title": "Blueberries",
        "value": "Blueberries",
        "_index": 6
      }
    ]
  },
  {
    "title": "Recent 2",
    "hideDivider": true,
    "data": [
      {
        "title": "Cherries",
        "value": "Cherries",
        "_index": 7
      },
      {
        "title": "Crab apples",
        "value": "Crab apples",
        "_index": 8
      },
      {
        "title": "Jambolan",
        "value": "Jambolan",
        "_index": 9
      }
    ]
  }
]
```

packages/blade/src/components/ActionList/ActionList.tsx
anuraghazra: Yes right, lets refactor this.

packages/blade/src/components/ActionList/styles/StyledListBoxWrapper.native.tsx
```diff
@@ -1,9 +1,12 @@
-import { ScrollView } from 'react-native';
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable react/display-name */
+import { SectionList } from 'react-native';
 import styled from 'styled-components/native';
 import { getBaseListBoxWrapperStyles } from './getBaseListBoxWrapperStyles';
 
 // It's not really styled here but have to keep name same since we have to style this with overflowY and maxWidth in web
-const StyledListBoxWrapper = styled(ScrollView)((props) => ({
+const StyledListBoxWrapper = styled(SectionList)((props) => ({
```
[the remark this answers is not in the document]
saurabhdaware: Like what if we render `SectionList` here instead of BaseBox? https://github.com/razorpay/blade/blob/b79b2ac3f04c4f6d1ff5dd5503f6114d4c40c0ab/packages/blade/src/components/ActionList/ActionListItem.tsx#L109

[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah, it is styled right?

packages/blade/src/components/ActionList/ActionListBox.native.tsx
```diff
@@ -0,0 +1,64 @@
...
+        bouncesZoom={false}
+        bounces={false}
+        keyExtractor={(item: any) => {
+          return item.value;
+        }}
+        renderSectionHeader={renderActionListSectionHeader}
+        renderSectionFooter={renderActionListSectionDivider}
+        renderItem={renderActionListItem}
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ref={ref as any}
```
chaitanyadeorukhkar, line 52: Why is this any?

packages/blade/src/components/ActionList/actionListUtils.ts
chaitanyadeorukhkar: I still don't understand where we're using foundSection?

packages/blade/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -805,19 +807,46 @@ WithMultipleDropdowns.args = {
   name: 'design-system',
 };
 
-export const DropdownPerformance = (): React.ReactElement => {
+export const ASectionsPerf = (): React.ReactElement => {
```
chaitanyadeorukhkar, line 810: Renamed this? What is AsectionsPerf? DrowndownPerformance seemed more contextual

packages/blade/src/components/ActionList/ActionList.tsx
chaitanyadeorukhkar: Wtf 😆

chaitanyadeorukhkar: I think I meant, should we use index here because value's value could be unpredictable. Even though using index is not the best practice it might be safer than value. This "value" is what the consumer would pass as options, right?

packages/blade/src/components/ActionList/actionListUtils.ts
anuraghazra: foundSection is a temporary variable we use to find and mutate the sectionData. 

Minimal example: 

image

packages/blade/src/components/ActionList/ActionList.tsx
anuraghazra: > This "value" is what the consumer would pass as options, right?

Yes that should also be fine, value should be stable across renders. 

Either way users should not change the value of any <ActionListItem /> in the runtime, doesn't make sense to do so either way. 

Imagine: 
```
<select>
  <option value={Math.random() > 0.5 ? "what" : "why?"}>hello</option>
  <option value="world">world</option>
</select>
```

packages/blade/src/components/ActionList/ActionList.tsx
anuraghazra: Okay added bg white looks okay. There was still a bit of a gap even after adding bg color because of padding in the container, had to use a bit of negative margin hacks to fix it.

packages/blade/src/components/ActionList/ActionListBox.native.tsx
anuraghazra: Most of our refs are casted to any because react refs don't play along very nicely with multiple unions refs like (TextInput | HTMLInputElement), React ref types are [way too messed up](https://twitter.com/mattpocockuk/status/1559844928422830080) to even worth fixing them.

packages/blade/src/components/ActionList/styles/StyledListBoxWrapper.native.tsx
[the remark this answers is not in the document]
anuraghazra: > Like what if we render SectionList here instead of BaseBox?

We can't do that, SectionList needs the structured data of all the items to be rendered and then it does the virtualization on that data and then renders only the component which are in view. 

It won't be possible to use ActionListSection and somehow patch it to use SectionList because our ActionListSection only groups the data in a visual hierarchy, but in react-native side we only need 1 wrapper which we push all the data into. 

Right now what we do is we take the JSX and convert it to the data structure which feeds into the SectionList. 

Frame 1483

> And what is the advantage of using SectionList if we are wrapping all items in it?

So we are not wrapping all the things in SectionList, sectionList doesn't even accept any children prop, we just feed the data, and provide functions like `renderItem`, `renderSectionFooter`, `renderSectionHeader` which does the rendering.

packages/blade/src/components/Dropdown/Dropdown.stories.tsx
anuraghazra: Renamed to internal components so it doesn't show up on consumer end.

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -105,12 +106,19 @@ const _ActionListSection = ({
...
   return (
     <BaseBox
       {...makeAccessible({
         role: getActionListSectionRole(),
         label: title,
       })}
+      style={{
+        paddingTop: theme.spacing[3],
+        marginTop: -theme.spacing[3],
+      }}
```
saurabhdaware, line 120: Can use direct props from Box right?
```suggestion
      paddingTop="spacing.3"
      marginTop={`-${theme.spacing[3]}px`}
```

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -105,12 +106,19 @@ const _ActionListSection = ({
...
+      backgroundColor="surface.background.level2.lowContrast"
```
saurabhdaware, line 121: This will set static color for actionlist no? ideally it should take from `surfaceLevel` prop. We can set transparent or nothing here to let it read from parent. Currently it breaks when surfaceLevel is set to 3

image

packages/blade/src/components/ActionList/ActionListBox.native.tsx
saurabhdaware, line 1: Nicee 👯

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: Oh I think this is messing up with spacings of its upper element? The padding of element above section header gets cropped out.

image

image

packages/blade/src/components/ActionList/ActionListBox.native.tsx
```diff
@@ -0,0 +1,62 @@
...
+      [],
+    );
+
+    return (
+      <StyledListBoxWrapper
+        sections={sectionData}
+        windowSize={5}
+        initialNumToRender={5}
+        keyExtractor={(item: any) => {
+          return item.value;
```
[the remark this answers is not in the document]
anuraghazra: > predictability

Yes values will always be unique string, similar to how <select \/> works on html.

packages/blade/src/components/ActionList/ActionListItem.tsx
anuraghazra: Right good catch!

packages/blade/src/components/ActionList/ActionListItem.tsx
anuraghazra: Will check.

anuraghazra:
@chaitanyadeorukhkar video demo: 


https://user-images.githubusercontent.com/35374649/232464498-3ed8a346-da76-4859-bb26-149b7b167998.mov

chaitanyadeorukhkar (dismissed):

packages/blade/src/components/ActionList/ActionListItem.tsx
anuraghazra: Can't add transparent becuase of this https://github.com/razorpay/blade/pull/1095#discussion_r1166693555

packages/blade/src/components/ActionList/ActionListItem.tsx
anuraghazra: Fixed.

saurabhdaware (approved):
Looks so good on native now 👯

Issue #1013, opened 2023-02-15
Ref are breaking in React Native


saurabhdaware:
Ref stories on react native storybook are breaking. We might have to add some validation in ref hook to ignore in react native.

saurabhdaware:
Same as #1090. That one has more information on how to fix so closing this

Issue #1090, opened 2023-03-28
Referencing `HTMLElement` is causing a crash on React Native


chaitanyadeorukhkar:
This is causing the Button Ref Story to crash on native. Could also have
https://github.com/razorpay/blade/blob/1b9e3b037592a9e94f2f44daff3c7799e103f90e/packages/blade/src/hooks/useBladeInnerRef.ts#L4

saurabhdaware:
#1111

PR #1111, opened 2023-04-11
fix: ref breakage on react native

chaitanyadeorukhkar:
Fixes: #1013 #1090

`instanceof HTMLElement` tries to use HTMLElement as a value which is not available for react native. Added a platform check to ensure 
`instanceof HTMLElement` doesn't run.

### DIFF
--- a/packages/blade/src/hooks/useBladeInnerRef.ts
+++ b/packages/blade/src/hooks/useBladeInnerRef.ts
@@ -1,5 +1,6 @@
 import React from 'react';
 import type { TextInput as TextInputReactNative, View } from 'react-native';
+import { getPlatformType } from '~utils';
 
 type BladeElementRef = Pick<HTMLElement, 'focus' | 'scrollIntoView'> | Pick<View, 'focus'>;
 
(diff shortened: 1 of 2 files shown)


packages/blade/src/hooks/useBladeInnerRef.ts
```diff
@@ -23,7 +24,7 @@ const useBladeInnerRef = (
     targetRef,
     (): BladeElementRef => {
       const element = innerRef.current;
-      if (element instanceof HTMLElement) {
+      if (getPlatformType() !== 'react-native' && element instanceof HTMLElement) {
```
anuraghazra, line 27: Just check if on react-native the "focus" method is working as expected by the ref. (Try on the input focus stories)

anuraghazra (approved):

packages/blade/src/hooks/useBladeInnerRef.ts
chaitanyadeorukhkar: Works. Something else is breaking there in terms of story's styling (we've used alignItem='end' but it is not available on native) but works otherwise

PR #1113, opened 2023-04-12
fix(Box): react-native style types

### LINKED ISSUE #1112: BaseBox: Add platform-specific types for React Native
Few options like `alignItems="end"` are not supported on React Native right now. We should add platform-specific types in BaseBox and Box.

Slack: https://razorpay.slack.com/archives/G01B3LQ9H0W/p1681213185989129

saurabhdaware:
fixes #1112 and bunch of other types in Box and styled-props

## Changes

- Created `PickCSSByPlatform` type utility which picks from `CSSObject` type in web and `ViewStyle` in react native and automatically returns `never` when prop doesn't exist on react native.

We won't have to explicitly handle types anymore.

### DIFF
--- a/packages/blade/src/_helpers/types.ts
+++ b/packages/blade/src/_helpers/types.ts
@@ -87,14 +87,6 @@ type TestID = {
 };
 
 /**
- * Returns the object as it is in web and sets all properties of object to type `never` in react native
- */
-type MakeObjectWebOnly<T> = Platform.Select<{
-  web: T;
-  native: Record<keyof T, never>;
-}>;
-
-/**
  * Similar to `Pick` except this returns `never` when value doesn't exist (native `Pick` returns `unknown`).
  *
  * You might have to ts-ignore the non-existing type error while using this. This is done so that you can get jsdoc from actual type.
@@ -120,8 +112,8 @@ type PickIfExist<T, K extends keyof T> = {
  * // On Native --> This will be just `flex` and `none`
  * ```
  */
-type PickCSSByPlatform<T extends keyof CSSObject | keyof ViewStyle> = Platform.Select<{
-  web: Pick<CSSObject, T>;
+type PickCSSByPlatform<T extends keyof React.CSSProperties | keyof ViewStyle> = Platform.Select<{
+  web: PickIfExist<CSSObject, T>;
   // @ts-expect-error: T passed here may not neccessarily exist. We return `never` type when it doesn't
   native: PickIfExist<ViewStyle, T>;
 }>;
@@ -134,7 +126,6 @@ export {
   StringChildrenType,
   StringWithAutocomplete,
   TestID,
-  MakeObjectWebOnly,
   PickIfExist,
   PickCSSByPlatform,
 };
--- a/packages/blade/src/_helpers/types.ts
+++ b/packages/blade/src/_helpers/types.ts
@@ -1,6 +1,9 @@
 import type React from 'react';
+import type { ViewStyle } from 'react-native';
+import type { CSSObject } from 'styled-components';
 import type { Spacing } from '~tokens/global';
 import type { EasingFunctionFactory } from '~tokens/global/motion';
+import type { Platform } from '~utils';
 
 /**
  * @template TokenType token type generic
@@ -83,6 +86,46 @@ type TestID = {
   testID?: string;
 };
 
+/**
+ * Returns the object as it is in web and sets all properties of object to type `never` in react native
+ */
+type MakeObjectWebOnly<T> = Platform.Select<{
+  web: T;
+  native: Record<keyof T, never>;
+}>;
+
+/**
+ * Similar to `Pick` except this returns `never` when value doesn't exist (native `Pick` returns `unknown`).
+ *
+ * You might have to ts-ignore the non-existing type error while using this. This is done so that you can get jsdoc from actual type.
+ *
+ * E.g. This will pick from ViewStyle prop if value exists else returns undefined.
+ *
+ * ```ts
+ * // @ts-expect-error: T passed here may not neccessarily exist. We return `never` type when it doesn't
+ * native: PickIfExist<ViewStyle, T>;
+ * ```
+ */
+type PickIfExist<T, K extends keyof T> = {
+  [P in K]: P extends keyof T ? T[P] : never;
+};
+
+/**
+ * Picks the types based on the platform (web / native).
+ *
+ * E.g.
+ * ```ts
+ * type CSSObject = PickCSSByPlatform<'display'>
+ * // On Web --> This will be all possible web display properties like `block`, `flex`, `inline`, and more.
+ * // On Native --> This will be just `flex` and `none`
+ * ```
+ */
+type PickCSSByPlatform<T extends keyof CSSObject | keyof ViewStyle> = Platform.Select<{
+  web: Pick<CSSObject, T>;
+  // @ts-expect-error: T passed here may not neccessarily exist. We return `never` type when it doesn't
+  native: PickIfExist<ViewStyle, T>;
+}>;
+
 export {
   DotNotationColorStringToken,
   DotNotationMotionStringToken,
@@ -91,4 +134,7 @@ export {
   StringChildrenType,
   StringWithAutocomplete,
   TestID,
+  MakeObjectWebOnly,
+  PickIfExist,
+  PickCSSByPlatform,
 };
--- a/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
+++ b/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
@@ -215,7 +215,6 @@ const getAllProps = (
     // Visual props
     backgroundColor: getBackgroundValue(props.backgroundColor, props.theme, breakpoint),
     borderRadius: getBorderRadiusValue(props.borderRadius, props.theme, breakpoint),
-    transform: getResponsiveValue(props.transform, breakpoint),
     lineHeight: getSpacingValue(props.lineHeight, props.theme, breakpoint),
     border: getResponsiveValue(props.border, breakpoint),
     borderTop: getResponsiveValue(props.borderTop, breakpoint),
--- a/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
+++ b/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
@@ -1,14 +1,8 @@
-import type { CSSObject } from 'styled-components';
 import type { MarginProps, PaddingProps, SpacingValueType } from './spacingTypes';
 import type { MakeObjectResponsive } from './responsiveTypes';
 import type { Theme } from '~components/BladeProvider';
 import type { Border } from '~tokens/global';
-import type {
-  DotNotationColorStringToken,
-  MakeObjectWebOnly,
-  PickCSSByPlatform,
-  TestID,
-} from '~src/_helpers/types';
+import type { DotNotationColorStringToken, PickCSSByPlatform, TestID } from '~src/_helpers/types';
 
 type LayoutProps = MakeObjectResponsive<
   {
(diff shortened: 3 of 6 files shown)


packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
```diff
@@ -215,7 +215,6 @@ const getAllProps = (
     // Visual props
     backgroundColor: getBackgroundValue(props.backgroundColor, props.theme, breakpoint),
     borderRadius: getBorderRadiusValue(props.borderRadius, props.theme, breakpoint),
-    transform: getResponsiveValue(props.transform, breakpoint),
```
saurabhdaware, line 218: `transform` was only allowed in BaseBox and not exported to consumer. 

`transform` is very different on web vs native. On Native it takes a completely different object.

I went through our codebase and saw it was also never used. So don't see it's worth adding bunch of checks to make it work on web and native both when its never really used.

packages/blade/src/_helpers/types.ts
```diff
@@ -83,6 +86,46 @@ type TestID = {
   testID?: string;
 };
 
+/**
+ * Returns the object as it is in web and sets all properties of object to type `never` in react native
+ */
+type MakeObjectWebOnly<T> = Platform.Select<{
```
chaitanyadeorukhkar, line 92: You're the new TS god 🙌

packages/blade/src/_helpers/types.ts
```diff
@@ -83,6 +86,46 @@ type TestID = {
...
+ * Picks the types based on the platform (web / native).
+ *
+ * E.g.
+ * ```ts
+ * type CSSObject = PickCSSByPlatform<'display'>
+ * // On Web --> This will be all possible web display properties like `block`, `flex`, `inline`, and more.
+ * // On Native --> This will be just `flex` and `none`
+ * ```
+ */
+type PickCSSByPlatform<T extends keyof CSSObject | keyof ViewStyle> = Platform.Select<{
```
chaitanyadeorukhkar, line 123: Super nice!

packages/blade/src/_helpers/types.ts
```diff
@@ -83,6 +86,46 @@ type TestID = {
...
+  web: Pick<CSSObject, T>;
+  // @ts-expect-error: T passed here may not neccessarily exist. We return `never` type when it doesn't
```
chaitanyadeorukhkar, line 125: Any workaround so that we don't have to use `tsexpecterror`. Lord @anuraghazra ?

packages/blade/src/_helpers/types.ts
saurabhdaware: Learning from the OG TS god Anurag  🙌🏼

packages/blade/src/_helpers/types.ts
saurabhdaware: @anuraghazra This is what I am trying to do

https://user-images.githubusercontent.com/30949385/231411096-ba672ad9-c7b7-4291-95d8-8da11d64a3d4.mov

[TS Playground](https://www.typescriptlang.org/play?#code/PQKhAIGEHsAcEsCmATcAzATtAtuARgDYCGyi4IwAsAFCg3ngBCGRAdsgM7hHgAuAnrDLYiAa3isA5nwAWiXEQDGvbl1Y4JRAvQgABDojIzevWBwBcwYNhTwArtgB0inMF0BpRADcJjAOIYhqyIwBx2GD4+UgC0vHKxgogcihjwsLzRiC4c-By88tF4LOwSktFsyAlCsUSSkqXRAGyKaI2IAKwAjI3tAEwAzMg6VNSIAB6w0BgqAkJMxcgAPAAqiQA04AByRDbg4-nsXHmpUgB84AC84KtzAGTgAN7gAPrPRRWvAPzmWztkAL4AbhoNHGk2mfESv148C8iAA8hgAOqIPDMCqXeYVRZsfgbADkrCIMLh+PAAB9wPiAO6o-GnYHUGhEmwcWBKMgABWIvDQU1wD3oewmUxmUO2uyuhOJsMQ+MZ4HAoBAQoYACV4JJjOB1NSNnFhNBkHYCGRAhxoAQ7DDoKxwPAjogZtBwObLXDwAADRy0vCe9DwU0cNaqlXURUMZYyOxcbnEvkYJwAZUQpuUi3OszI1MDBFdTvCdoNXt9-qzoZGirBoshcxTad4i3h6Xgtq4+0Qh0e4F9PzsrFEutYgJ1MrhfYHQ-A-3OV3R7EWqvAzZtrA4AG0JYgALoh8MR-GwHkJ7DRX34oUMmj-EFMu-K8AAWQAmlB4QARACi5BGNAfUYdcAXFZdAsFwfFAiUDIiRJOVwHZRRRFqMgKBoLNwAANSQakkwEU1MUFfdFWVJdIxkQCACkk3faBFHADgZGgE1UHGB0VAkfMoNHWDSNQojwGQB0jyIfgfnxNBTTGMlKUJW05UZa9b3-ciuGApJQJwKk8n4U1KmAyZglYXgODJBCkMkMgfzQqFICTJN4TwAArLIVCuQiIyVMBeOuFTwComi6IYpiCBYsY2PtO1fV4ysI0EtliFEqlCFo0RpKpCTxjS2TgnlJcSP4sjAMA31oltAh+HgrAhGmCqMWQaAklYfEVFYvIIu42V6LwpJoqXSRUmQMSxn4AAvC9qBvO9rLmTl4EQgBJNBPzCvIVg2dxhQOThwFERB+GgNBrgpHa9oOqA7Ic5zlFnR4hXXTl2vcbcfgejsu12-bDuWcBPmue7t3AH5gjhDBGSBW9pq5ObRFspNGH4ONeX5FZNs7baPrO2HLpc46McOrDEBw7qbsRk9HHrFzFncntURe6HFixpyXI2ZYr33GDZTphalpWxsCaJnTEBZtmZwU6h0MYaAxlwwWuCuWbENh+HSeR-E4uE-gsv6+BkHpJSvPDCA-GgLRzB0cBonAJM7FgcEVAQRDSgE+A0DQRBAiM2t1NtZ3Xfdzt7ePflzctvwnT86jaIExA0AkFB2v5mX8IxRmrt4YZbxcNcVDwKWfkl6Xurl27+PVhKxIyqTryAA)

anuraghazra: In here, the `keyof CSSObject | keyof ViewStyle` generic contraint just becomes `string | number` because CSSObject is a index signature and doesn't specify the concrete CSS object with all the values. 

So right now intellisense either way isn't working for `PickCSSByPlatform` so might be better to do this: 

```ts
type PickCSSByPlatform<T extends string> = Platform.Select<{
  web: Pick<CSSObject, T>;
  native: PickIfExist<ViewStyle, T>;
}>;

type PickIfExist<T, K extends string> = {
  [P in K]: P extends keyof T ? T[P] : never;
};
```

packages/blade/src/_helpers/types.ts
anuraghazra: :godmode:

packages/blade/src/_helpers/types.ts
saurabhdaware: But then we loose the JSDoc that comes from actual types

image

Expected outcome would be this on react native where it also picks the jsdoc from ViewStyle -
image

anuraghazra: In that case then keep the PickIfExist as is. Adding ts-expect-error should not be an issue.

JSDoc vanishes because in TypeScript there is something called homomorphic types which compiler smartly figures out based on similar keys of the object, If we pass `T extends string` then the compiler isn't able to relate the original object to and just creates a "fresh" object property instead of reusing the original one (which had jsdoc)

```
type PickIfExist<T, K extends keyof T> = {
  [P in K]: P extends keyof T ? T[P] : never;
};
```

anuraghazra: > Basically just replacing CSSObject to React.CSSProperties. Then the intellisense also works for PickCSSByPlatform.

Yeah in `PickCSSByPlatform` the keep it as React.CSSProperties so that we get intellisese. 

```
type PickCSSByPlatform<T extends keyof ViewStyle | keyof React.CSSProperties> 
```

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -161,19 +130,22 @@ type BoxVisualProps = MakeObjectResponsive<{
...
+        GridProps,
+        | 'gridColumn'
+        | 'gridRow'
+        | 'gridRowStart'
+        | 'gridRowEnd'
+        | 'gridColumnStart'
+        | 'gridColumnEnd'
+        | 'gridArea'
+      >,
+    '__brand__'
```
anuraghazra, line 147: In here why do we need to omit __brand__?

anuraghazra: Oh nvm, got it. since it leaks into the consumer end.

packages/blade/src/_helpers/types.ts
kamleshchandnani: can't find its usage? where is this used?

packages/blade/src/_helpers/types.ts
```diff
@@ -83,6 +86,46 @@ type TestID = {
...
+ *
+ * E.g.
+ * ```ts
+ * type CSSObject = PickCSSByPlatform<'display'>
+ * // On Web --> This will be all possible web display properties like `block`, `flex`, `inline`, and more.
+ * // On Native --> This will be just `flex` and `none`
+ * ```
+ */
+type PickCSSByPlatform<T extends keyof CSSObject | keyof ViewStyle> = Platform.Select<{
+  web: Pick<CSSObject, T>;
```
kamleshchandnani, line 124: why not use PickIfExist here as well?

packages/blade/src/_helpers/types.ts
```diff
@@ -83,6 +86,46 @@ type TestID = {
...
+  // @ts-expect-error: T passed here may not neccessarily exist. We return `never` type when it doesn't
+  native: PickIfExist<ViewStyle, T>;
```
kamleshchandnani, line 126: what will this return?
so,

I say `PickIfExist<ViewStyle, 'display'>`, then it will call the utility type which will be called as
```
type PickIfExist<ViewStyle, 'display'> = {
  [display]: display extends keyof ViewStyle ? ViewStyle[display] : never;
}; 
```

so what will it return?

saurabhdaware: `display extends keyof ViewStyle` this condition will be true because react native's ViewStyle supports `display`. So it will return the value of `{ display: ViewStyle[display] }`

packages/blade/src/_helpers/types.ts
saurabhdaware: Yup I removed it's usage later because now we can just use `PickCSSByPlatform<'display' | 'non-existing-prop'>` and that in itself removes props that are not supported in react native. 
 
Removing the utility as well. Can add it when we need.

packages/blade/src/_helpers/types.ts
saurabhdaware: Because we don't have react-native specific props in Box so everything is supported in web. But yup changed it to PickIfExist so in case we add any react-native specific prop in future, it will be handled.

packages/blade/src/_helpers/types.ts
saurabhdaware: > Yeah in PickCSSByPlatform keep it as React.CSSProperties so that we get intellisese.

yup made this change. thanks!

packages/blade/src/components/Box/styledProps/getStyledProps.ts
```diff
@@ -31,7 +31,8 @@ const removeUndefinedStyledProps = (obj: StyledPropsInputType): StyledPropsInput
   const onlyDefinedStyledProps: StyledPropsBlade = {};
   for (const key in obj) {
     if (obj[key as keyof StyledPropsBlade] !== undefined) {
-      // @ts-expect-error: complex type error
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error
```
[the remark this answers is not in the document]
anuraghazra: why did we replaced tsexpecterror with tsignore tho?

anuraghazra (approved):

packages/blade/src/components/Box/styledProps/getStyledProps.ts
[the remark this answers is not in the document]
nashcheez: Better thicc than skinny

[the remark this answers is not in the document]
saurabhdaware: because ts-expect-error goes unused in react native typecheck but used in web typecheck. So on react-native, it throws unused error on cli.

image

PR #1122, opened 2023-04-14
docs(Dropdown): Controlled Dropdown API Decision

saurabhdaware:
### TODOs
- [x] Do a POC implementation to see how feasible proposed approach is


packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,192 @@
...
+## Open Questions
+
+### Are `<ActionListItem />` "selectable" components?
+
+**Context:** We support `isDefaultSelected` on ActionList and plan to support `isSelected` now so that means we are treating ActionList as selectable items. When we have `Button` as trigger, we turn ActionList into menu which is not "selectable" but rather a list of "clickable" items.
+
+- **Pros of keeping ActionList selectable**
+
+  - Selection logic is still applicable in `<AutoComplete />`, or any selectable trigger we plan to add in future
+  - Here we are only talking about "selection" logic on consumer-facing props but selection logic on our end will anyway always be there irrespective of this call as we have single component that acts as `<MenuOption />` and `<SelectOption />`, both. Normally I have seen libraries tackle this in separate components.
```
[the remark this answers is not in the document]
kamleshchandnani: also, how do we do selection logic currently for ActionListItem? based on SelectInput's context?

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,192 @@
...
+
+Currently there is no ([legal](https://codesandbox.io/s/state-fill-usecase-42u1ev?file=/App.tsx)) way to achieve this because we only support Uncontrolled Dropdown where you can set default selection on first render, but every other selection is handled internally only.
+
+### Problem Statement 2: Reset SelectInput value
+
+If you have a "Reset" button in your form, there isn't way to de-select an already selected item ([legally](https://razorpay.slack.com/archives/C01H13RTF8V/p1681287510398639?thread_ts=1681286843.013449&cid=C01H13RTF8V)). For such usecases as well we need controlled Dropdown
+
+## Solutions
+
+### `onChange` + `isSelected` ✅
```
chaitanyadeorukhkar, line 19: This does seem like the best approach but it will compromise the re-usability of ActionList in cases where it's a clickable.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,192 @@
...
+  setCurrentSelection(values[0]);
+};
+
+<Dropdown>
+  <SelectInput label="City" onChange={onSelectChange} />
+  <DropdownOverlay>
+    <ActionList>
+      <ActionListItem
+        // New Prop 👇🏼
+        isSelected={currentSelection === 'mumbai'}
```
chaitanyadeorukhkar, line 44: How do we plan to handle multiple `isSelected={true}` in a single select dropdown?

kamleshchandnani (commented):
This is so beautifully written. Crisp-Concise ❤️

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,192 @@
...
+| DX (Opinion)             | ✅ Similar to ``tag in HTML. Also we already have`isDefaultSelected`so might be intuitive to have`isSelected` as well                                                                                                                                       | Also intuitive because we support similar thing in other Inputs. My concerns was `value={['mango', 'apple']}` with multiselect. Passing array to `value` might seem a bit strange. Also this approach will only make sense if we overall [decide to make "ActionList" non-selectable items on consumer end](#are-actionlistitem--selectable-components) | This forces you to write selection logic outside of JSX which might not be as intuitive when you want to do something like fetching initial data from API and set selected values while rendering the ActionList. More info at [why not `useSelect`](#why-not-useselect) |
```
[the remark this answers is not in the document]
saurabhdaware: Yup. When `select` tag has `multiple` attribute, the only way to select option is setting `selected` attribute on `options`

```jsx
<select multiple>
  <option selected>mango</option>
  <option selected>apple</option>
  <option>banana</option>
</select>
```

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,192 @@
...
+| State Management         | ✅ Independent of how state is manged (consumers can store value of item in state as shown in example above, they can store selectedIndex, they might also have entire ActionList data as array somewhere and they can use `selected` prop from their data) | Consumer can store `currentSelectionValue` in state and handle on basis of that                                                                                                                                                                                                                                                                         | Handles state internally. Requires FE Core to export appropriate methods. E.g. If consumer is storing selected data in some object, they have to transform that data to our input format and then call `selectItem`                                                      |
```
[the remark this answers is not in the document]
saurabhdaware: So in `isSelected` approach, your selection logic is part of JSX so it can be written in rendering part which would be more intuitive when your data is stored in different ways.

Whereas with something like `selectItem()` method, you have to call that function when state is updating to select an item.

You can checkout example in this section - https://github.com/razorpay/blade/blob/saurabh/controlled-dropdown/api-decisions/packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md#why-not-useselect

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,192 @@
...
+| DX (Opinion)             | ✅ Similar to ``tag in HTML. Also we already have`isDefaultSelected`so might be intuitive to have`isSelected` as well                                                                                                                                       | Also intuitive because we support similar thing in other Inputs. My concerns was `value={['mango', 'apple']}` with multiselect. Passing array to `value` might seem a bit strange. Also this approach will only make sense if we overall [decide to make "ActionList" non-selectable items on consumer end](#are-actionlistitem--selectable-components) | This forces you to write selection logic outside of JSX which might not be as intuitive when you want to do something like fetching initial data from API and set selected values while rendering the ActionList. More info at [why not `useSelect`](#why-not-useselect) |
+
+#### Why not `useSelect`?
+
+Throughout initial discussions we mostly talked about creating this hook so noting down things here which made me change my opinion-
```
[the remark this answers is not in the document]
saurabhdaware: learning from the team 🤗

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,192 @@
...
+### Referrences
+
+- Primer
+  - [ActionMenu](https://primer.style/react/ActionMenu#with-selection) (Closest to our approach. Uses `selected` and `onSelect`)
+    ```jsx
+    <ActionList.Item
+      key={index}
+      selected={index === selectedIndex}
+      //
+      onSelect={() => setSelectedIndex(index)}
```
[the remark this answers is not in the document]
saurabhdaware: So their `<ActionList.Item />` is only "clickable" Its not used inside their Select. For them, `<SelectPanel />` is different component with completely different API - https://primer.style/react/SelectPanel

Here they are probably using "onSelect" as an alternative to "onClick" (probably because elements in dropdown can also be actioned by keyboard navigations and are not necessarily limited to clicks)

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,192 @@
...
+    ```
+- Radix [[CodeSandbox Example](https://codesandbox.io/p/sandbox/holy-cache-5kwcgm?selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A28%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A24%7D%5D&file=%2FApp.jsx)]
+- [Polaris](https://polaris.shopify.com/components/selection-and-input/autocomplete) (They have `selected` object with `onSelect` on top component itself)
+- [downshift-js](https://github.com/downshift-js/downshift/tree/master/src/hooks/useSelect#actions) (`useSelect` hook)
+
+## Open Questions
+
+### Are `<ActionListItem />` "selectable" components?
+
+**Context:** We support `isDefaultSelected` on ActionList and plan to support `isSelected` now so that means we are treating ActionList as selectable items. When we have `Button` as trigger, we turn ActionList into menu which is not "selectable" but rather a list of "clickable" items.
```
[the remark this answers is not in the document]
saurabhdaware: So few things are different between a `menu` and `select` -

In menu, options perform some action (going to different page, triggering some function, etc). E.g. "Go to Settings", "Log Out" kind of options.

In select, when option is clicked, it's selected and added to select's value. It's not right UX if we just redirect the consumer to some page when item is selected. E.g. SelectInputs which selects states and cities, Filters menu, etc.


Think of a Dropdown that has "Go to Settings", "Log Out" kind of option. Having "Log Out" option selected would mean nothing right? It's a one-off action.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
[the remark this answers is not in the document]
saurabhdaware: So inside the `onClick` of ActionListItem, we have logic to select that item. We basically use our internal `useDropdown` hook and select that particular item when it's clicked.

It's also more than `onClick`, we also have to select that item when keyboard keys are pressed, we have to set `aria-selected` on it, we have to set `aria-current` to show the current active element.

So overall there is lots of selection logic that will anyway be part of `<ActionListItem />` only. From implementation point, `<ActionListItem />` has to be selectable item because of a11y and functional requirements.


By libraries tackling this separately, I was referring to few libraries who have separate components like this - 

- GitHub
  - [SelectPanel](https://primer.style/react/SelectPanel)
  - [ActionMenu](https://primer.style/react/ActionMenu)
- Radix
  - [DropdownMenu](https://www.radix-ui.com/docs/primitives/components/dropdown-menu)
  - [Select](https://www.radix-ui.com/docs/primitives/components/select#select)

In this case, they can handle selection semantics in select components and clickable action items in dropdown / ActionMenu type of components.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,192 @@
...
+
+- **Pros of keeping ActionList selectable**
+
+  - Selection logic is still applicable in `<AutoComplete />`, or any selectable trigger we plan to add in future
+  - Here we are only talking about "selection" logic on consumer-facing props but selection logic on our end will anyway always be there irrespective of this call as we have single component that acts as `<MenuOption />` and `<SelectOption />`, both. Normally I have seen libraries tackle this in separate components.
+
+- **Cons**
+  - Not applicable in Button trigger so `isSelected`, `isDefaultSelected` would do nothing in that case
```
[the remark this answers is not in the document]
saurabhdaware: It is definitely a con but we would have to take this tradeoff either way if we only have single `<ActionListItem />` component. 

Think of this in `SelectInput` where `<ActionListItem />`, rather than just being selectable items are trying to do some more actions. As a UX, this is not expected from SelectInput. It's more of a "menu" thing. So it not only impacts `isDefaultSelected` type of props but also impacts `onClick`, `href` on ActionListItem.
```jsx
<Dropdown>
  <SelectInput />
  <DropdownOverlay>
    <ActionList>
       <ActionListItem href="/settings" title="Settings" />
      <ActionListItem onClick={() => logOut()} title="Log Out" />
    </ActionList>
   </DropdownOverlay>
</Dropdown>
```

If we are treating `<ActionListItem />` as both menu and selectable item, there will always be props that work in menu but don't work in select and that work in select but don't work in menu since they are both UX-wise different items. 

If want to remove these impossible scenarios, we can create `<SelectOptions>` and `<SelectOptionItem>` kind of component that replace ActionList in Select. Then the selection logic goes into these components and ActionListItem handles menu items.

Although I still feel our API is good since consumers get one component that can do both so they don't have to think about which type of menu to use. But yup, it comes with these confusing impossible states between menu and select.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,192 @@
...
+  - Will be a breaking change (We'll move to using `defaultValue` and `value` on `<SelectInput />` instead of `isDefaultSelected` and `isSelected` on `<ActionListItem />`)
```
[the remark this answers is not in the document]
saurabhdaware: I realised that just moving to `defaultValue` won't solve all problems. 

The core problems is there are props that releated to menu (E.g. `onClick`, `href` on `ActionListItem`) and there are props that are related to selection (E.g. `isSelected`, `isDefaultSelected` on `ActionListItem`). There will always be impossible or non-intended cases as long as we have one component that is handling two usecases.

Explained in details here - https://github.com/razorpay/blade/pull/1122#discussion_r1169894812

divyanshu013 (commented):
One point of inconvenience is that we've controls at three places now:
- selectionType at dropdown
- onChange at select input
- isSelected at actionlist item

One plus of `value` based approach I can think of is having both the value and change handler at the same place

saurabhdaware:
I have updated the doc with final conclusion we had in a meeting. You all can take a look once

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,342 @@
...
+          // This handles internal controlled state
+          setCurrentSelection('mumbai');
+        }}
+      />
+      <ActionListItem title="Bangalore" value="bangalore" />
+    </ActionList>
+  </DropdownOverlay>
+</Dropdown>;
+```
+
```
divyanshu013, line 205: I think a note on the functionality with selectionType is missing.

- what if selection type is single but user passes isSelected for multiple actionlistitems

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
saurabhdaware: Added this and another note about other impossible state in additional notes in second last section

divyanshu013 (dismissed):
Looks good to me.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,342 @@
...
+const onSomeAction = () => {
+  setCurrentSelection('mumbai');
+};
+
+<Dropdown>
+  <DropdownButton>Show Dropdown</DropdownButton>
+  <DropdownOverlay>
+    <ActionList>
+      <ActionListItem
```
kamleshchandnani, line 190: When we have dropdown button, wouldn't we need to update the text of the button to the selected item? 🤔

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,342 @@
...
+        title="Mumbai"
+        value="mumbai"
+        // New Props 👇🏼
+        isSelected={currentSelection === 'mumbai'}
+        onClick={() => {
```
kamleshchandnani, line 195: shouldn't the param of this function also give the value of this prop?

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,342 @@
...
+Lets compare 2 `onChange` prop-based approaches now-
+
+| `onChange` + `isSelected`                                                                    | `onChange` + `value`                                                       |
+| -------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------- |
+| Preferred Approach if we say `ActionListItem` is selectable component independent of trigger | Preferred Approach if only `SelectInput` trigger has selectable components |
+
+**Conclusion:** We decided that we would need a combination of both approaches. Because-
+
+- Based on the [selectable usecase of Menu](#menu-with-selected-item), the ActionListItem has to be "selectable" individually. Which means we definitely need something like `isSelected` on ActionListItem.
+- However the concerns raised with `onChange` + `isSelected` approach were that the controlled states end up on 2 different indepdent component which is confusing and they in a way become bound to each other.
```
kamleshchandnani, line 238: elaborate this line for the simplicity of readers as to what we mean by independent component. maybe take an example that `onChange` on select input and isSelected on `ActionListItem`

kamleshchandnani: there's another downside as well that we discussed and we shall add here, where if we have isSelected then consumers will have to always mention that when consuming this prop so poor DX.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,342 @@
...
+  <DropdownOverlay>
+    <ActionList>
+      <ActionListItem title="Mumbai" value="mumbai" />
+      <ActionListItem title="Bangalore" value="bangalore" />
+    </ActionList>
+  </DropdownOverlay>
+</Dropdown>;
+```
+
+### When Trigger is Button or other triggers
```
kamleshchandnani, line 174: add another example where ActionList is used outside of DropDown? For eg: BottomSheet

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,342 @@
...
+- [MUI](https://mui.com/material-ui/react-menu/#selected-menu)
+- [Primer](https://primer.style/react/ActionMenu#with-selection)
+
+</details>
+
+##### Conclusion
+
+Based on the usecases like Menu with Selected Items, we can conclude that `ActionListItem` is indeed a selectable component. However, In Menu, it can only be selected in a controlled state (When someone passes `isSelected`) and cannot be selected when item is clicked.
+
+Whereas in Select, it can be selected in a controlled state as well as when user clicks on item.
```
kamleshchandnani, line 320: this seems confusing to readers?
can we simply put this is as

>Whereas in Select, it can be selected in a controlled state as well as when user clicks on item.

👇 

Whereas in Select, it will be autoselected when the user clicks on item.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,342 @@
...
+
+#### Additional Notes and Action Items
+
+- We noticed that we've been mixing the "Menu" and "Select" so far. It's better to refactor and separate them out into individual internal component for better separation of concerns and UX.
+  - [ ] Refactor the internal code of `ActionListItem` into `SelectItem` and `MenuItem` components to handle 2 separate behaviours and semantics.
+- There are also "impossible" states like passing `isSelected` on ActionListItem when `SelectInput` is used, we decided to handle it validations end. Things like trying to select multiple items in single select will also be handled on validations end.
```
kamleshchandnani, line 326:
```suggestion
- There are also "impossible" states like passing `isSelected` on ActionListItem when `SelectInput` is used, we decided to handle it via validations. Things like trying to select multiple items in single select will also be handled via validations.
```

kamleshchandnani (commented):
overall looks good to me, minor suggestions for readers 🎉

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
```diff
@@ -0,0 +1,342 @@
...
+  <DropdownOverlay>
+    <ActionList>
+      <ActionListItem title="Mumbai" value="mumbai" />
+      <ActionListItem title="Bangalore" value="bangalore" />
+    </ActionList>
+  </DropdownOverlay>
+</Dropdown>;
+```
+
+Inspirations
```
AnshulNautiyal, line 128: Link? @saurabhdaware

saurabhdaware: ops I missed this heading here. I moved the references to the end of doc

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
saurabhdaware: nope. That should ideally be a SelectInput if you have that usecase. Although if someone wants to have that, they can still do that since anyway the selection is controlled here. Won't recommend though. Most of the times, SelectInput would be the right component for such usecases.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
saurabhdaware: It is based on trigger and not necessarily based on what it is being rendered inside. E.g. even in SelectInput BottomSheet, we do use `Dropdown` component so the same `onChange` + `value` will continue to work.

And in BottomSheet with Button, the same Button trigger rules apply so the same example is valid for that as well. 

It won't really change anything in the approach so didn't add example here to not deviate and confuse from the focus of this part.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
saurabhdaware: > Whereas in Select, it will be selected in controlled state and will be autoselected when the user clicks on item

this will be the correct one no? It won't be autoselected in Controlled state. There users have to handle which is selected item and which is not based on `onChange` and their logic.

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
saurabhdaware: done

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
saurabhdaware: yup it does give. updated this example to use that value instead

divyanshu013 (approved):

packages/blade/src/components/Dropdown/_decisions/controlled-dropdown.md
kamleshchandnani: see this is a new transactions module
Screenshot 2023-05-08 at 5 23 23 PM

PR #1131, opened 2023-04-19
test(blade): added bottomsheet test cases

anuraghazra:
BottomSheet PR Stack: 

- BottomSheet Test cases - #1131 - THIS PR
- BottomSheet Header Footer - #1151 
- BottomSheet Native - #1036


packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+    <ActionList>
+      {fruites.map((fruit) => {
+        return <ActionListItem key={fruit} title={fruit} value={fruit} />;
+      })}
+    </ActionList>
+  );
+};
+
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
+afterAll(() => jest.restoreAllMocks());
```
divyanshu013, line 58: Let's have this only at related test blocks

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+            <BottomSheetBody>
+              <Text>BottomSheet body</Text>
+            </BottomSheetBody>
+          </BottomSheet>
+        </>
+      );
+    };
+    const { getByText, queryByTestId } = renderWithTheme(<Example />);
+
+    expect(queryByTestId('bottomsheet-body')).not.toBeInTheDocument();
```
divyanshu013, line 108: Can we query on the bottomsheet content instead? That would be closer to user behavior.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.web';
+import { Text } from '~components/Typography';
+import { Button } from '~components/Button';
+import { Dropdown } from '~components/Dropdown';
+import { SelectInput } from '~components/Input/SelectInput';
+import { ActionList, ActionListItem } from '~components/ActionList';
+import { Badge } from '~components/Badge';
+import { Counter } from '~components/Counter';
+
+export const sleep = (delay = 10): Promise<number> =>
```
divyanshu013, line 17: Can we rely on testing library primitives for [waiting](https://testing-library.com/docs/guide-disappearance/#waiting-for-appearance)?

packages/blade/jest-setup.native.js
```diff
@@ -25,4 +26,6 @@ jest.mock('react-native-reanimated', () => ({
       return '';
     }
   },
+  makeMutable: (value) => ({ value }),
+  useWorkletCallback: require('react').useCallback,
```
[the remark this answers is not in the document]
anuraghazra: Added comment, they come from nowhere, it was throwing error so I added the mocks :D

packages/blade/tsconfig.json
```diff
@@ -11,6 +11,6 @@
     },
     // In Blade side we are targetting `.web` types
     // We can't add both `.native` & `.web` at once.
-    "moduleSuffixes": [".all", ".web", ""]
+    "moduleSuffixes": [".all", ".native", ""]
```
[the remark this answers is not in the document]
anuraghazra: reverted

packages/blade/src/utils/testId/testId.web.ts
```diff
@@ -1,3 +1,7 @@
-export const testID = (defaultKey: string): { 'data-testid': string } => ({
-  'data-testid': defaultKey,
-});
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export const testID = (defaultKey: string) => {
```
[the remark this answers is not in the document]
anuraghazra: I think I can use the metaAttribute func which chaitanya added, will remove then all together.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
anuraghazra: Right. updated the ones which have text, but kept the ActionList examples to use testID only because we keep rendering those element just set opacity: 0 to the bottomsheet, I'm unable to do `expect(getByRole('combobox', { name: 'Select Action' })).not.toBeVisible();` seems like jest doesn't take into consideration that the parent element is invisible while calculating opacity.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
anuraghazra: Tried those none of it worked actually, not exactly sure why. I have to wait for the animations to finish in realtime timers, cant also use jest.useFakeTimers()

packages/blade/src/components/BottomSheet/BottomSheetBackdrop.web.tsx
```diff
@@ -23,6 +23,7 @@ const BottomSheetBackdrop = ({ zIndex }: { zIndex: number }): React.ReactElement
 
   return (
     <StyledBottomSheetBackdrop
+      {...testID('bottomsheet-backdrop')}
```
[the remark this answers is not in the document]
anuraghazra: Should be fine for now, it will be way too much effort to figure out and drill down unique IDs into these internal BottomSheet components.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.native.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import React from 'react';
+import { BottomSheet, BottomSheetHeader, BottomSheetFooter } from '../BottomSheet';
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
+import { Button } from '~components/Button';
+import { Badge } from '~components/Badge';
+import { Counter } from '~components/Counter';
+
+beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
+afterAll(() => jest.restoreAllMocks());
```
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.native.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <BottomSheetHeader
+            title="Address Details"
+            subtitle="Saving addresses will improve your checkout experience"
+            trailing={<Badge variant="positive">Action Needed</Badge>}
+            titleSuffix={<Counter intent="positive" value={2} />}
+          />
+          {/* Can't render BottomSheetBody because https://github.com/gorhom/react-native-bottom-sheet/issues/11#issuecomment-1283588472 */}
+          {/* <BottomSheetBody>
+            <Text>BottomSheet body</Text>
+          </BottomSheetBody> */}
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Get rid of the commented out JSX as well? Seems confusing. Just keep the comment regarding the issue

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.native.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    };
+    renderWithTheme(<Example />);
+
+    expect(console.warn).toHaveBeenCalledWith(
+      expect.stringContaining(
+        '[Blade BottomSheetHeader]: Do not pass "size" to "Badge" while inside BottomSheetHeader trailing, because we override it.',
+      ),
+    );
+  });
+});
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: back handler we should leave up to the consumer, right? In case they want to keep the bottom sheet open and back does something else within the bottom sheet

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
chaitanyadeorukhkar: Why can't we use fakeTimers? For animations its best to use fakeTimers. This arbitrary sleep will eventually become  flaky and also add a delay in our test runs

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+            </BottomSheetBody>
+          </BottomSheet>
+        </>
+      );
+    };
+    const { getByText, queryByTestId } = renderWithTheme(<Example />);
+
+    expect(queryByTestId('bottomsheet-body')).not.toBeInTheDocument();
+    await user.click(getByText(/open/i));
+    await sleep(250);
```
chaitanyadeorukhkar, line 110: These will become flaky and add delay in our tests. Lets try to make fake timers work

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+    expect(queryByTestId('bottomsheet-body')).toBeInTheDocument();
```
chaitanyadeorukhkar, line 111: `query` returns an array of similar elements. Should we use `get` instead?

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+    await user.click(queryByTestId('bottomsheet-backdrop')!);
+    await sleep(250);
+    expect(queryByTestId('bottomsheet-body')).not.toBeInTheDocument();
+  });
+
+  it.skip('should close with close button', async () => {
```
chaitanyadeorukhkar, line 117: Why are we skipping this test?

packages/blade/tsconfig.json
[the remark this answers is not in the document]
chaitanyadeorukhkar: Doesn't seem reverted 🤔

packages/blade/src/utils/testId/testId.web.ts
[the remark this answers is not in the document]
chaitanyadeorukhkar: Lets use metaAttribute only

packages/blade/tsconfig.json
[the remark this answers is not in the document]
anuraghazra: Yeah forgot to push. done.

packages/blade/src/utils/testId/testId.web.ts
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.native.test.tsx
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
anuraghazra: Done.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
anuraghazra: Yeah I agree. but I've spent quite a while on this, nothing seemed to work not exactly sure why. 


This 250 is the duration of our moderate motion token, i can update it to reference that so that if we change the duration on bottomsheet it changes here also. 

Let's track this & I'll create a backlog issue for this. BottomSheet anyway needs better test with E2E

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
anuraghazra: queryAll returns array, query returns 1 element only.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
anuraghazra: Fixed, for some reason user.click didn't worked, had to use fireEvent.click. weird.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
anuraghazra: Tracked here: https://github.com/razorpay/blade/issues/1168

chaitanyadeorukhkar (approved):

PR #1132, opened 2023-04-19
feat: add bundle size checks

chaitanyadeorukhkar:


.github/workflows/release.yml
```diff
@@ -39,5 +39,12 @@ jobs:
         if: steps.changesets.outputs.published == 'true' || github.event_name == 'workflow_dispatch' # run when the package is published via changeset or if a release is triggered manually.
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-        run: yarn publish-npm
+      - name: Build Blade if not built by changeset
+        if: steps.changesets.outputs.published == 'false'
```
divyanshu013, line 43: When will this happen, wouldn't it always be built in release 🤔

divyanshu013 (approved):
Looks good to me. One query.

.github/workflows/release.yml
chaitanyadeorukhkar: Release workflow always runs but it only builds when blade is to be published that is when update version PR is merged otherwise it doesn't build.

PS: I've added this as my best understanding of the workflow, whether it will work as intended or not we'll know only when we merge it to master 🤡 

But it should work. I'll monitor it

divyanshu013 (approved):
Looks good to me.

.github/workflows/release.yml
saurabhdaware: yolo

packages/blade/package.json
```diff
@@ -218,5 +221,48 @@
     "react-dom": "17.0.2",
     "react": "17.0.2",
     "styled-components": "^5"
-  }
+  },
+  "bundlemon": {
+    "files": [
+      {
```
saurabhdaware, line 227: Do we want to add utilities as well?

saurabhdaware (dismissed):
Looks good to me 👯

packages/blade/package.json
chaitanyadeorukhkar: Damn. Thanks. Missed it. Added!

divyanshu013 (approved):

.github/workflows/release.yml
anuraghazra: Why do we need to build blade in release workflow?

saurabhdaware (approved):

snitin315 (approved):
:shipit:

PR #1138, opened 2023-04-25
doc(blade): added switch api decisions

anuraghazra:
### Formatter Document


packages/blade/src/components/Switch/_decisions/decisions.md
```diff
@@ -0,0 +1,69 @@
...
+| defaultChecked | `boolean` | `false` | Default state of the switch, if isChecked is not provided the switch will become [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html) |        |
+| onChange | `({ event, isChecked, value }) => void` | `undefined` | The function to be called when the switch state changes. |        |
+| isDisabled | `boolean` | `false` | Control whether the switch is disabled or not. |        |
+| accessibilityLabel | `string` | `undefined` | Provides accessible label for internal checkbox component that sets the aria-label prop for screen readers. |        |
+| size | `small, medium` | `medium` | Size of the switch component |        |
+| name (html) | `string` | `undefined` | The name of the input field in a switch, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name) |        |
+| value (html) | `string` | `undefined` | The value of the input field in a switch, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) |        |
+
```
divyanshu013, line 32: Would we be utilizing the checkbox utils? Seems quite similar

packages/blade/src/components/Switch/_decisions/decisions.md
```diff
@@ -0,0 +1,69 @@
...
+```js
+// basic
+<Switch value="dark-mode" />
+```
+
+#### Controlled, Uncontrolled
+
+```js
+const Controlled = () => {
+  const [chceked, setChecked] = React.useState(false);
```
divyanshu013, line 46:
```suggestion
  const [checked, setChecked] = React.useState(false);
```

packages/blade/src/components/Switch/_decisions/decisions.md
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <Switch isChecked={checked} onChange={({isChecked}) => setChecked(isChecked)} value="dark-mode" />
```
divyanshu013, line 49: Should also change the `value` to more accurately represent form value usecase :)

packages/blade/src/components/Switch/_decisions/decisions.md
divyanshu013, line 32: Do we need label (see some usecases in design)

image

 validation states, helper texts etc.? eg if being used in form fields (though don't see these on design side)

divyanshu013, line 32: Do we also need a control for the icon (in on state)? Is that non configurable and non toggle-able?

divyanshu013 (commented):
Looks good overall, few queries

packages/blade/src/components/Switch/_decisions/decisions.md
anuraghazra: <Switch value={true | false} /> won't really be intuitive enough, the form value is either "on" or "off". 

Since Switch is just a fancy Checkbox, we should make the API same to reduce learning curve.

packages/blade/src/components/Switch/_decisions/decisions.md
anuraghazra: Yeah these are not in the design. I've asked from design side if they are needed.

anuraghazra: No icon isn't configurable.

anuraghazra: Yes, we will be using useCheckbox & SelectorInput to do a11y and state management. Since it's just a checkbox.

anuraghazra: > Do we need label 

Discussing this internally.

packages/blade/src/components/Switch/_decisions/decisions.md
divyanshu013: Oh, just checked the spec for checkbox that value is only submitted if the checkbox is checked

packages/blade/src/components/Switch/_decisions/decisions.md
```diff
@@ -0,0 +1,85 @@
...
+  );
+};
+
+const Uncontrolled = () => {
+  return (
+    <Switch defaultChecked={true} value="dark-mode" />
+  );
+};
+```
+
```
divyanshu013, line 60: I think also add an example on form submissions enforcing the fact that switch will work analogous to checkboxes for forms

anshulsahni:
The component doesn't have the option to show the two possible values on sides, are we not considering switch as an alternative to choose from two possible values or we're expecting that in such use cases devs will have to add those labels by themselves

saurav12:
> The component doesn't have the option to show the two possible values on sides, are we not considering switch as an alternative to choose from two possible values or we're expecting that in such use cases, devs will have to add those labels by themselves

@anshulsahni Yes, this call has been taken intentionally. The switch is used in different places in different styles. Some use labels on the left, and some use them on the right. Not only this, but some also change the size & style of the label used based on the kind of UI they are trying to create. Therefore we took a call to remove the label from the component. And we also observed the same pattern being used by major other systems as well.

Mentioning a few use examples:
image
image

chaitanyadeorukhkar (approved):

chaitanyadeorukhkar (approved):

divyanshu013 (approved):
:shipit:

packages/blade/src/components/Switch/_decisions/decisions.md
```diff
@@ -0,0 +1,85 @@
...
+| onChange | `({ event, isChecked, value }) => void` | `undefined` | The function to be called when the switch state changes. |        |
+| isDisabled | `boolean` | `false` | Control whether the switch is disabled or not. |        |
+| accessibilityLabel | `string` | `undefined` | Provides accessible label for internal checkbox component that sets the aria-label prop for screen readers. |        |
+| size | `small, medium` | `medium` | Size of the switch component |        |
+| name (html) | `string` | `undefined` | The name of the input field in a switch, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name) |        |
+| value (html) | `string` | `undefined` | The value of the input field in a switch, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) |        |
+| id (html) | `string` | `undefined` | The id of the input field in a switch, useful for associating a label element with the input via `htmlFor` prop |        |
```
saurabhdaware, line 32: can you create an issue to see if this can be added to other inputs as well? seems useful. We can discuss over the issue

saurabhdaware (approved):

packages/blade/src/components/Switch/_decisions/decisions.md
anuraghazra: You are talking about the ID prop right? 

Actually it might just only make sense for Switch input only since this is the only component where we are not explicitly exposing label <Swtich label="Toggle darkmode" \/>, thats why we need the ID to associate with a htmlFor label. 

In other inputs we already have built-in labels, `<Checkbox>Toggle darkmode</Checkbox>`, `<TextInput label="" />` etc

divyanshu013: We haven't intentionally exposed `id` on any other components btw

saurabhdaware: oh got it. makes sense.

packages/blade/src/components/Switch/_decisions/decisions.md
anuraghazra: From design side we decided to not add labels, consumers can either use their own label or use accessibilityLabel when a visual label isn't appropriate or required.

packages/blade/src/components/Switch/_decisions/decisions.md
```diff
@@ -0,0 +1,85 @@
...
+
+## `Switch` API
+
+| Prop       | Type              | Default     | Description                 | Required |
+| ---------- | ----------------- | ----------- | --------------------------- | -------- |
+| isChecked | `boolean` | `false` | State of the switch, if explicitly set the switch will become [controlled component](https://reactjs.org/docs/forms.html#controlled-components) |        |
+| defaultChecked | `boolean` | `false` | Default state of the switch, if isChecked is not provided the switch will become [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html) |        |
+| onChange | `({ event, isChecked, value }) => void` | `undefined` | The function to be called when the switch state changes. |        |
+| isDisabled | `boolean` | `false` | Control whether the switch is disabled or not. |        |
+| accessibilityLabel | `string` | `undefined` | Provides accessible label for internal checkbox component that sets the aria-label prop for screen readers. |        |
```
kamleshchandnani, line 28: `accessibilityLabel` should be required right?

packages/blade/src/components/Switch/_decisions/decisions.md
kamleshchandnani: yes but there's an ask where we will have to make labels optional(bunch of places on dashboards) so I would say lets not expose the id yet until we figure out all the solution for all form elements? Anyways it'll be a broken functionality for form submissions but that's an edge case and we have ariaLabel for a11y so safe to take this tradeoff(yes it's broken functionality hence tradeoff)

packages/blade/src/components/Switch/_decisions/decisions.md
```diff
@@ -0,0 +1,85 @@
...
+  const [chceked, setChecked] = React.useState(false);
+
+  return (
+    <Switch isChecked={checked} onChange={({isChecked}) => setChecked(isChecked)} value="dark-mode" />
+  );
+};
+
+const Uncontrolled = () => {
+  return (
+    <Switch defaultChecked={true} value="dark-mode" />
```
kamleshchandnani, line 56: put the `onChange` here as well in demo? because even if its uncontrolled we'll expose the checked value to the consumers via `onChange` right?

packages/blade/src/components/Switch/_decisions/decisions.md
```diff
@@ -0,0 +1,85 @@
...
+</Box>
+
+// or
+
+<label>Dark mode <Switch /></label>
+```
+
+## Accessibility
+
+Switch will work similarly as Checkbox, where the hidden input field will follow the [Checkbox accessibility](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/) pattern. But we will also have a visual button which will follow the [Switch accessibility](https://www.w3.org/WAI/ARIA/apg/patterns/switch/) pattern.
```
kamleshchandnani, line 80: didn't quite understand this statement? 🤔

packages/blade/src/components/Switch/_decisions/decisions.md
anuraghazra: No we can't make it optional. 

If people bind the label to the Switch, accessibilitlyLabel prop becomes redundant. 

```jsx
<label htmlFor="dark-mode">
   Dark mode
</label>
<Switch id="dark-mode" accessibilityLabel="Dark Mode" /> // Redundant aria-label, no need because label is binded to the switch already
  
// or

<label>
  Dark mode 
  // Redundant aria-label, no need because label is binded to the switch already
  <Switch accessibilityLabel="Dark Mode" />
</label>
```

packages/blade/src/components/Switch/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/Switch/_decisions/decisions.md
anuraghazra: Sorry for the confusion, updated with better wording and example from WAI ARIA site.

packages/blade/src/components/Switch/_decisions/decisions.md
divyanshu013: If switches aren't meant to be used as form fields (hence from design side no labels support), perhaps we don't need `id` support right now

anuraghazra: > If switches aren't meant to be used as form fields 

Okay then let's take it as an enhancement later then.

packages/blade/src/components/Switch/_decisions/decisions.md
kamleshchandnani: i don't even think anyone will care to use this prop 😮‍💨

packages/blade/src/components/Switch/_decisions/decisions.md
kamleshchandnani: +1 to @divyanshu013. Anyways we can keep value and other prop but can probably remove the id prop. its more chaotic to remove it later than to add it later

packages/blade/src/components/Switch/_decisions/decisions.md
anuraghazra: Hmm okay now that we are saying the "primary" function of the Switch won't be to used inside a form, but rather as a standalone button, then we can enforce accessibilityLabel, and treat `<label><swtich/></label>` as a rare case.

divyanshu013 (approved):
Looks good to me.

PR #1139, opened 2023-04-27
feat(Dropdown): Controlled SelectInput Implementation

### LINKED ISSUE #1102: Dropdown onChange triggers on the initial mount
Dropdown's onChange fn is getting triggered on the initial mount with values as `[""]`. This value is not dependent on ActionList or ActionListItem.

**Code to reproduce the issue:**
```
      <Dropdown selectionType="single">
        <SelectInput
          label="Filter:"
          labelPosition="top"
          onChange={({ values }) => console.log(values)}
          placeholder="Select Option"
          validationState="none"
          // name="overviewFilter"
        />
        <DropdownOverlay>
          {/* <ActionList surfaceLevel={2}>
            {overviewConfigFilterOptions.map(({ label, value }, index) => (
              <ActionListItem isDefaultSelected={index === 0} title={label} value={value} />
            ))}
          </ActionList> */}
        </DropdownOverlay>
      </Dropdown>
```

saurabhdaware:
- API Decision #1122 
- Resolves #1102 

## 🚧 Expected Migration for Consumers 🚧 

> **Note**
>
> This is breaking change only for consumers who use `isDefaultSelected` on `ActionListItem` and / or `onChange` on SelectInput. If you're not using either of it, feel free to upgrade without any migration.

- Remove `isDefaultSelected` and use `defaultValue` on SelectInput. You can pass array of values to `defaultValue` in case of multiselect
  ```diff
  <Dropdown>
    <SelectInput 
      label="Select City"
  +   defaultValue="mumbai" 
    />
    <DropdownOverlay>
      <ActionListItem 
        title="Mumbai" 
        value="mumbai" 
  -     isDefaultSelected 
       />
      <ActionListItem title="Bangalore" value="bangalore" />
    </DropdownOverlay>
  </Dropdown>
  ```
- `onChange` now will not be triggered on initial mount. In your code if there is any logic that expects it to trigger on mount, make sure to move that logic into mount of your own component. 

## Changes

- `value` + `defaultValue` + `onChange` implementation on SelectInput
- 🚧 Remove `isDefaultSelected`, it won't be required anymore as in SelectInput you would use `defaultValue` and in Button trigger you would use `isSelected`.
- Tests
- Story for Controlled Dropdown

## TODOs in this PR

- [x] ~Add changeset with migration guide~ Will add combined changeset in integration branch

## TODO PRs

- [x] `isSelected` + `onClick` implementation on ActionListItem #1152 
- [x] Button Trigger #1152
- [x] ~Refactor to separate out Menu and Select logic~ Evaluated. Too complex to refactor since a lot of code is indeed shareable.
- [x] Docs Improvements #1161


packages/blade/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -914,4 +913,106 @@ export const InternalDropdownPerformance = (): React.ReactElement => {
...
+    title = '',
+    description,
+    value = '',
+    actionListItemIcon,
+    ...selectInputArgs
+  } = args;
+
+  return (
+    <>
+      <Button onClick={() => setCurrentSelection('bangalore')}>Select Bangalore</Button>
```
divyanshu013, line 931: Minor: add some space below this

packages/blade/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -914,4 +913,106 @@ export const InternalDropdownPerformance = (): React.ReactElement => {
...
+    <>
+      <Button
+        onClick={() => {
+          if (!currentSelection.includes('bangalore')) {
+            setCurrentSelection([...currentSelection, 'bangalore']);
+          }
+        }}
+      >
+        Select Bangalore
+      </Button>
```
divyanshu013, line 986: Minor: add some space below this.

packages/blade/src/components/Dropdown/__tests__/Dropdown.native.test.tsx
```diff
@@ -79,7 +79,7 @@ describe('<Dropdown />', () => {
     // Click on combobox
     fireEvent.press(getByRole('combobox'));
     expect(getByTestId('dropdown-overlay').props.display).toBe('flex');
-    expect(selectOnChange).toBeCalledWith({ name: 'fruits', values: [''] });
+    expect(selectOnChange).not.toBeCalled();
```
divyanshu013, line 82: Nice :)

packages/blade/src/components/Dropdown/__tests__/Dropdown.native.test.tsx
```diff
@@ -171,4 +171,99 @@ describe('<Dropdown />', () => {
...
+          </Dropdown>
+        </>
+      );
+    };
+
+    const { getByTestId, getByRole } = renderWithTheme(<ControlledDropdown />);
+
+    const selectInput = getByRole('combobox');
+    expect(selectInput).toHaveTextContent('Select Option');
+    fireEvent.press(getByTestId('select-blr-button'));
```
divyanshu013, line 208: No other way to get the button? Other than using test id i.e.

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -64,12 +82,89 @@ const _SelectInput = (
...
+      return;
+    }
+
+    if (props.onChange && !isFirstRenderRef.current) {
+      props.onChange({
+        name: props.name,
+        values: getValuesArrayFromIndices(),
+      });
+    }
+    // onChange?.({ name: props.name, values: value.split(', ') });
```
divyanshu013, line 165: Can we remove this comment?

```suggestion
```

divyanshu013 (commented):
Looks good overall, needs changeset with migration plan

packages/blade/src/components/Dropdown/__tests__/Dropdown.native.test.tsx
kkyusufk: FYI, `userEvent` is a better api to write these tests. 
This blog really does a good job to explain the difference very well
https://blog.mimacom.com/react-testing-library-fireevent-vs-userevent/

saurabhdaware: > No other way to get the button? Other than using test id i.e.

Not in react native unfortunately. The API is very limited (we can't use `name` property of getByRole in react native. Also the reason why we added `testID` to our components recently)

> FYI, userEvent is a better api to write these tests.

It is! We use it in web tests, on native its not supported.

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -111,11 +120,16 @@ const _Dropdown = ({
       hasLabelOnLeft,
       setHasLabelOnLeft,
       dropdownTriggerer: dropdownTriggerer.current,
+      changeCallbackTriggerer,
+      setChangeCallbackTriggerer,
+      isControlled,
+      setIsControlled,
     }),
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
kamleshchandnani, line 128: why do we need to disable this eslint warning or why do we not put all the deps in the dependency array?

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -35,6 +35,16 @@ type SelectInputProps = Pick<
...
 > & {
   icon?: IconComponent;
+  /**
+   * Controlled value of the Select. Use it in combination of `onChange`.
+   *
+   * Check out [Controlled Dropdown Documentation](https://blade.razorpay.com/?path=/story/components-dropdown-with-select--controlled-dropdown&globals=measureEnabled:false) for example.
+   */
+  value?: string | string[];
+  /**
+   * Set initial value of Select. Use `value` instead if you want to make Dropdown controlled.
```
kamleshchandnani, line 45: Probably below description? this is what we use in text input.
Used to set the default value of Select Input field when it's uncontrolled

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -35,6 +35,16 @@ type SelectInputProps = Pick<
   | 'testID'
 > & {
   icon?: IconComponent;
+  /**
+   * Controlled value of the Select. Use it in combination of `onChange`.
+   *
+   * Check out [Controlled Dropdown Documentation](https://blade.razorpay.com/?path=/story/components-dropdown-with-select--controlled-dropdown&globals=measureEnabled:false) for example.
+   */
+  value?: string | string[];
```
kamleshchandnani, line 43: should it be `value` or `values`? same thing for `defaultValue`

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -64,12 +82,88 @@ const _SelectInput = (
...
+      }
+
+      selectValues(props.value);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [props.value, options]);
+
+  React.useEffect(() => {
+    // Ignore calling onChange on mount
+    if (isFirstRenderRef.current) {
```
kamleshchandnani, line 154: why do we need to check this explicitly?

packages/blade/src/components/Dropdown/__tests__/Dropdown.native.test.tsx
divyanshu013: Doesn't `getByText`, `getByLabelText` work 🤔 used them on alert native tests

divyanshu013 (commented):
- [ ] Add changeset with migration guide
- [ ] Review and approve chromatic tests

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: Good question. This is why you're principle engineer 🤭🫡

I think it depends on what we do here - https://github.com/razorpay/blade/pull/1139#discussion_r1188142533

I am too feeling `values` might be better option now. The reason we had `value` was because this code seems a bit complex for single select usecases-
```jsx
const [currentSelection, setCurrentSelection] = React.useState()

return (
  <Dropdown>
    <SelectInput 
      values={[currentSelection]} 
      onChange={({values}) => setCurrentSelection(values[0])}
    />
    <DropdownOverlay>
      <ActionList />
    </DropdownOverlay>
  </Dropdown>
)
```

You can also store array in useState and directly set values and onChange values but then on consumer end when they are programatically selecting something, they will have to ensure that array only has one value.



But... since we already decided to have `values` in `onChange` for the similar reason, I feel we'll have to be consistent with that decision.

packages/blade/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -356,6 +354,7 @@ export const WithSingleSelect = DropdownTemplate.bind({});
 export const WithMultiSelect = DropdownTemplate.bind({});
 WithMultiSelect.args = {
   selectionType: 'multiple',
+  defaultValue: 'home',
```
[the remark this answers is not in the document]
saurabhdaware: https://github.com/razorpay/blade/pull/1139#discussion_r1189307586

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -61,6 +64,10 @@ const _Dropdown = ({
   const [hasFooterAction, setHasFooterAction] = React.useState(false);
   const [hasLabelOnLeft, setHasLabelOnLeft] = React.useState(false);
   const [isKeydownPressed, setIsKeydownPressed] = React.useState(false);
+  const [changeCallbackTriggerer, setChangeCallbackTriggerer] = React.useState<
```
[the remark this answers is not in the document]
saurabhdaware: It can be used in dependency array to know when the change has happened. The tricky part for us is that our `onChange` is on `SelectInput` component but that onChange is supposed to be triggered when ActionList item has some change. So this is the way to tell SelectInput that something has changed.

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: we haven't added the setter functions of state to dependency array. We don't need them no? in most cases adding something that is not needed in dependency array has ended me with infinite loops or some rendering issues lol so I prefer to disable this as long as some prop is not consciously needed.

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: Turning into `values` to be consistent

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: `useEffect` by default will run on mount. Which means the `onChange` call inside will run on mount which is not expected. We only want to trigger onChange after something has changed and not on mount. This is the part that fixes #1102

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
divyanshu013: Leaning towards using `value`. I think `value` should be fine whether it's a `string` or `string[]`. This would be analogous with `event.target.value` and also we're using `value` prop in input fields 🤔 

A slight inconvenience, otherwise for some components it would be `value` and for some `values`.

saurabhdaware: That is true. Also, A single select with plain values is a lot more common usecase than multiselect. So if we create API that is more multiselect-friendly, it would cause inconvenience to a lot more people.

packages/blade/src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
saurabhdaware: Added jsdoc description

packages/blade/src/components/Dropdown/useDropdown.ts
```diff
@@ -30,6 +30,11 @@ type DropdownContextType = {
    */
   selectedIndices: number[];
   setSelectedIndices: (value: number[]) => void;
+  /**
+   * contains the indexes of selected items during controlled selection
```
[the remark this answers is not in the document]
saurabhdaware: it is indexes. It is array of all selected item indexes.

packages/blade/src/components/Dropdown/useDropdown.ts
```diff
@@ -70,13 +75,21 @@ type DropdownContextType = {
    */
   hasLabelOnLeft: boolean;
   setHasLabelOnLeft: (value: boolean) => void;
+
+  changeCallbackTriggerer: number;
```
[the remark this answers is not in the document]
saurabhdaware: done

saurabhdaware:
Will add combined changeset to integration branch and release menu, docs refactor, and this together.

packages/blade/src/components/Dropdown/__tests__/Dropdown.native.test.tsx
saurabhdaware: It works when the node is just `<Text>Some Text</Text>` but doesn't work when you want to select outer element like `<Button><Text>Some Text</Text></Button>`

packages/blade/src/components/Dropdown/Dropdown.stories.tsx
saurabhdaware: done 🤣

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
kamleshchandnani: but onchange shouldn't trigger on mount regardless right?

packages/blade/src/components/Dropdown/useDropdown.ts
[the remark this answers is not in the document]
kamleshchandnani: is this to mimic the memoization to avoid triggering the onchange unnecessarily?

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
saurabhdaware: > but onchange shouldn't trigger on mount regardless right?

Ok so normally in other inputs, you can just call the `onChange` callback when `onChange` on the input is triggered. So when user types into TextInput, the onChange of text input works as expected. 

However in Select + Dropdown's case, we're changing values of Select programatically (based on state, when actionlistitem is clicked). So we cannot directly put `onChange` on `BaseInput` of Select because it will never get called (`onChange` only gets triggered when value is changed on user interaction with that input. Not when we programatically set that value).

So we kind of have to simulate the `onChange` behaviour here. So all this useEffect, and callbackTrigerer, first mount ignore is to simulate that behaviour.

packages/blade/src/components/Dropdown/useDropdown.ts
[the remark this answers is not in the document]
saurabhdaware: Not exactly. This is the only way I could think of. These comments might explain it-
- https://github.com/razorpay/blade/pull/1139#discussion_r1189308689
- https://github.com/razorpay/blade/pull/1139#discussion_r1192075425

packages/blade/src/components/Dropdown/__tests__/Dropdown.native.test.tsx
divyanshu013: Weird, I remember using same for testing the CTAs on alert

divyanshu013 (approved):
Looks good to me. Leaning towards opting for `value` terminology [here](https://github.com/razorpay/blade/pull/1139#discussion_r1190726884), we should close this. Rest :shipit:

divyanshu013:
Also please do check the chromatic snapshot updates

packages/blade/src/components/Dropdown/__tests__/Dropdown.native.test.tsx
saurabhdaware: ops yes righttt. `getByText` is working! It was some other visibility error earlier. Changed it here to `getByText`. In some scenarios where we have to select outer ActionList component or select Dropdown Overlay (which doesn't have any text inside it), I have still kept it `getByTestId`. But for ActionListItem and Button it's working fine now.

kamleshchandnani (approved):

Issue #1103, opened 2023-04-03
Dropdown Label when passed as empty string takes up margin space


joel-jaimon:
Label takes up space (margin) even when passed as an empty string. 

![Screenshot 2023-04-03 at 5 48 57 PM](https://user-images.githubusercontent.com/64752455/229507798-eee561fc-dbb8-474f-ba09-4ebd0feb9ca4.png)

divyanshu013:
Fixed in #1141

PR #1141, opened 2023-04-27
fix(blade): dropdown label when passed as empty string takes up margin space

Priyanshu-C:
Issue: https://github.com/razorpay/blade/issues/1103


divyanshu013 (approved):
Looks good, thanks for the fix!

PR #1146, opened 2023-05-01
relook and updated console mocks in test files

### LINKED ISSUE #930: Relook at all the console mocks in test files
Ref https://github.com/razorpay/blade/pull/928#discussion_r1058850148

Right now we have mocked console errors on top of almost every test file. While that prevents us from filling up our console unnecessarily where we test negative test cases using `toThrow`, the downside is that in doing so we might miss out on sometimes genuine essential errors. 

My proposal would be to add the console mocks to the test suite where required and remove them from the top of all the files

dev-cj:
resolves #930

### DIFF
--- a/packages/blade/src/components/Alert/__tests__/Alert.native.test.tsx
+++ b/packages/blade/src/components/Alert/__tests__/Alert.native.test.tsx
@@ -2,9 +2,6 @@ import { fireEvent } from '@testing-library/react-native';
 import { Alert } from '..';
 import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
 
-beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
-afterAll(() => jest.restoreAllMocks());
-
 describe('<Alert />', () => {
   it('should render', () => {
     const { toJSON } = renderWithTheme(
--- a/packages/blade/src/components/Alert/__tests__/Alert.web.test.tsx
+++ b/packages/blade/src/components/Alert/__tests__/Alert.web.test.tsx
@@ -4,9 +4,6 @@ import { Alert } from '..';
 import renderWithTheme from '~src/_helpers/testing/renderWithTheme.web';
 import assertAccessible from '~src/_helpers/testing/assertAccessible.web';
 
-beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
-afterAll(() => jest.restoreAllMocks());
-
 describe('<Alert />', () => {
   it('should render', () => {
     const { container } = renderWithTheme(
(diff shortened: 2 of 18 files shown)


divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Alert/__tests__/Alert.web.test.tsx
```diff
@@ -108,6 +106,7 @@ describe('<Alert />', () => {
         />,
       ),
     ).toThrow(`[Blade: Alert]: SecondaryAction is allowed only when PrimaryAction is defined.`);
+    jest.restoreAllMocks();
```
saurabhdaware, line 109: I think this would restore all the mocks no? even the ones that are not set inside this `it`.

I think so far we've been doing `afterAll` so should've been fine there. But if I add a `beforeAll` mock in this file now and write a test after this test block, I would imagine the mock would get restored before the next test runs which might not be expected 🤔 

Can we figure out a way to only restore `console.error` here?

packages/blade/src/components/Alert/__tests__/Alert.native.test.tsx
```diff
@@ -2,9 +2,6 @@ import { fireEvent } from '@testing-library/react-native';
 import { Alert } from '..';
 import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
 
-beforeAll(() => jest.spyOn(console, 'error').mockImplementation());
-afterAll(() => jest.restoreAllMocks());
-
```
saurabhdaware, line 7: no more unused mocks 👯 thanks for the PR 🥳

packages/blade/src/components/Alert/__tests__/Alert.web.test.tsx
saurabhdaware: I think something like `console.error.mockRestore()` might make more sense no? because if someone puts a mock in `beforeAll`, then that will only work for first test then.

divyanshu013: Sounds good. @dev-cj could you please incorporate the suggestion from this thread?

dev-cj: Yeah make sense, will implement it.

dev-cj: New changes added.

saurabhdaware (approved):
Yay looks good to me! thanks for the contributions 🎉

divyanshu013 (approved):
:shipit:

PR #1149, opened 2023-05-03
chore: add meta attribute to native

chaitanyadeorukhkar:
Adds meta attribute to native, similar to web. We need this for blade coverage tracking on native


packages/blade/src/utils/metaAttribute/metaAttribute.native.ts
```diff
@@ -1,5 +1,15 @@
...
+
+const metaAttribute = ({
+  testID,
+  name,
+}: {
+  name?: string;
+  testID?: string;
+}): Record<string, string> => {
   return {
+    ...(name ? { [`data-${MetaConstants.Component}`]: name } : {}),
```
chaitanyadeorukhkar, line 11: kept this to maintain parity with web but we won't be using this.

saurabhdaware:
Conflicts 🙈

packages/blade/src/utils/metaAttribute/metaAttribute.native.ts
```diff
@@ -1,5 +1,14 @@
-const metaAttribute = ({ testID }: { name?: string; testID?: string }): Record<string, string> => {
+import { MetaConstants } from './metaConstants';
+
+const metaAttribute = ({
```
divyanshu013, line 3: Minor: might be good to have some unit tests for this util

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Box/__tests__/Box.native.test.tsx
```diff
@@ -17,6 +17,7 @@ describe('<Box />', () => {
     );
     expect(toJSON()).toMatchInlineSnapshot(`
       <View
+        data-blade-component="box"
```
kamleshchandnani, line 20: does this works on native? IIRC from past, custom attributes were discarded by RN?

chaitanyadeorukhkar: It doesn't discard it during tests. So the react-test-renderer that's used during tests to generate these snapshots doesn't discard it.

divyanshu013 (approved):
:shipit:

Issue #979, opened 2023-01-27
Add as={ReactRouterLink} property for the Link component


snitin315:
> We should use the Link component from react-router-dom not Blade directly so the routing works smoothly on the client side after hydration.

<img width="358" alt="image" src="https://user-images.githubusercontent.com/46647141/215002116-97583e6e-5ba2-4842-ac73-c1f2e4f57ad9.png">

[Reported on slack](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1668515397799029?thread_ts=1668508148.107359&cid=G01B3LQ9H0W)

References:

- [Chakra UI Link](https://chakra-ui.com/docs/components/link#usage-with-routing-library)
- [Link usage in website](https://github.com/razorpay/frontend-website/blob/61843a74f4822608857889c0c9971489c18c2d58/src/components/Link/Link.tsx#L24)

saurabhdaware:
Few things to consider 
- How it will work with `<ActionListItem href="" />`
- How it will work with `<ListItemLink />`

saurabhdaware:
X's implementation - https://github.com/razorpay/x/blob/master/src/js/ui/Link/LinkV2.tsx#L34-L40

saurabhdaware:
React Router v6 Approach - https://reactrouter.com/en/main/upgrading/v5#remove-link-component-prop

chaitanyadeorukhkar:
@saurabhdaware considering the suggested approach by react router, can we not suggest the same usage to our consumers?

```jsx
import { Link as BladeLink } from "@razorpay/blade/components";
import {
  useHref,
  useLinkClickHandler,
} from "react-router-dom";

const Link = React.forwardRef(
  (
    {
      onClick,
      replace = false,
      state,
      target,
      to,
      ...rest
    },
    ref
  ) => {
    let href = useHref(to);
    let handleClick = useLinkClickHandler(to, {
      replace,
      state,
      target,
    });

    return (
      <BladeLink
        {...rest}
        href={href}
        onClick={(event) => {
          onClick?.(event);
          if (!event.defaultPrevented) {
            handleClick(event);
          }
        }}
        ref={ref}
        target={target}
      />
    );
  }
);
```


I thought we could handle this wrapping on our end but that would mean adding react-router as a dependency which I'm not in favour of since that is still a product decision

saurabhdaware:
[...]

PR #1152, opened 2023-05-03
feat(Dropdown): Add Menu (Button Trigger, `isSelected` on ActionListItem)

saurabhdaware:
Check out Dropdown -> With Button stories

**Changes**
- Add `DropdownButton` trigger for Dropdown
- Add `onClick` + `isSelected` in ActionList for Controlled Menu usecase
- Add `onDismiss` callback to tell consumers about open / close state

TODO:
- [x] Fixing story controls in #1161 
- [x] Merge conflict


packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -74,11 +76,26 @@ const _Dropdown = ({
...
+      onDismiss();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isOpen]);
 
   React.Children.map(children, (child) => {
     if (React.isValidElement(child)) {
       if (
         !isValidAllowedChildren(child, 'SelectInput') &&
+        !isValidAllowedChildren(child, 'DropdownButton') &&
```
anuraghazra, line 98: let's refactor out all these component ids into the componentIds constant.

packages/blade/src/components/Dropdown/DropdownButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+      })}
+      onClick={(e) => {
+        if (props.onClick) {
+          props.onClick(e as React.MouseEvent<HTMLButtonElement>);
+        }
+        onTriggerClick();
+      }}
+      onBlur={() => {
+        // With button trigger, there is no "value" as such. It's just clickable items
+        onTriggerBlur?.({ name: '', value: '' });
```
anuraghazra, line 40: maybe we should return null instead?

packages/blade/src/components/ActionList/actionListUtils.ts
```diff
@@ -77,7 +77,20 @@ const getActionListProperties = (
...
-        href: child.props.href,
+        onClickTrigger: (value) => {
+          const anchorLink = child.props.href;
+          child.props.onClick?.({
+            name: child.props.value,
+            value: child.props.isSelected ?? value,
+          });
+
+          if (anchorLink) {
+            window.location.href = anchorLink;
```
anuraghazra, line 88: Won't this do a full page reload? what if users wants to do client side navigation?

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -74,11 +76,26 @@ const _Dropdown = ({
...
+    if (isFirstRenderRef.current) {
+      isFirstRenderRef.current = false;
+      return;
+    }
+
+    if (!isOpen && onDismiss) {
+      onDismiss();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isOpen]);
```
anuraghazra, line 92: Instead of relying on isOpen state and syncing it to call the onDismiss. 

Won't it be better to actually call the onDismiss in places where we did setIsOpen(false). 
That would fix the firstRender ignore edge case too.

packages/blade/src/components/Dropdown/dropdownUtils.ts
```diff
@@ -212,7 +211,9 @@ type ActionsType = {
  */
 export const performAction = (
   action: SelectActionsType,
-  e: FormInputOnKeyDownEvent,
+  e: {
```
anuraghazra, line 214: nit: 

```suggestion
  payload: {
```

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -306,8 +312,20 @@ const _ActionListItem = (props: ActionListItemProps): JSX.Element => {
   const isMobile = platform === 'onMobile';
 
   const renderOnWebAs = props.href ? 'a' : 'button';
-  const isSelected =
-    typeof props._index === 'number' ? selectedIndices.includes(props._index) : undefined;
+
+  const getIsSelected = (): boolean | undefined => {
```
anuraghazra, line 316: Can you add some comments here what this is doing?

anuraghazra:
TODO: 

- [ ] Add tests
- [ ] Add changesets

packages/blade/src/components/ActionList/actionListUtils.ts
saurabhdaware: Kind of unsure what our client-side navigation strategy is right now also. Also, this only gets triggered when you select item with keyboard. Otherwise it will just work like a normal link.

I think once we we resolve https://github.com/razorpay/blade/issues/979, it will be clearer on what to do here and in other links. Kind of applies to `<ListItemLink />` kind of components as well since we can't use ReactRouter there because of validations. 

Added a comment here to keep track of this - https://github.com/razorpay/blade/issues/979#issuecomment-1550510357

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: yup wanted to do that only. Except to do that, we'll have to set the `onDismiss` callback in the DropdownContext. Wanted to avoid that because we already have so many things there.

packages/blade/src/components/Dropdown/DropdownButton.tsx
saurabhdaware: Changed it to undefined. That we already support in types as standard when values / name doesn't exist. 

If I change it to null here, we have to go and add checks at several places in other inputs to support null everywhere.

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: done

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: done

saurabhdaware:
Added tests. Changeset I'll add on integration branch after merging things. This is just stacked branch.

packages/blade/src/components/Dropdown/Dropdown.tsx
anuraghazra: You don't need to pass onDismiss into the DropdownContext. 

You could: 

```
const close = useCallback(() => {
  setIsOpen(false);
  onDismiss?.();
})
```

and just pass the close function downstream and whenever you setIsOpen(false) call `close`

packages/blade/src/components/ActionList/actionListUtils.ts
```diff
@@ -77,7 +77,21 @@ const getActionListProperties = (
...
+            name: child.props.value,
+            value: child.props.isSelected ?? value,
+          });
+
+          if (anchorLink) {
+            const target = child.props.target ?? '_self';
+            window.open(anchorLink, target);
+            if (window.top) {
+              window.top.open(anchorLink, target);
+            }
```
divyanshu013, line 92: Is this running on web only? `window` won't work on native

divyanshu013 (commented):
Looks good overall, few queries

packages/blade/src/components/Dropdown/__tests__/__snapshots__/Dropdown.native.test.tsx.snap
```diff
@@ -636,7 +636,6 @@ exports[`<Dropdown /> should render dropdown 1`] = `
               </View>
             </View>
             <RCTScrollView
-              accessibilityRole="menu"
```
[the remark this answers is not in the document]
saurabhdaware: yup 2 components were getting role menu. The outer container and inner container. Removed it from one.

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: > pass the close function downstream

How exactly 🤔 Because Dropdown is a parent component and we need to call close inside `ActionList`, `DropdownOverlay`, and from `useDropdown`. So we will have to set it in context right?

packages/blade/src/components/ActionList/actionListUtils.ts
saurabhdaware: yes it only runs when you click item with keyboard down which won't happen in mobile. I've added isReactNative check anyway to avoid breakage in case that happens in some edge case. But ideally `onKeyboardDown` callback itself doesn't exist on native so shouldn't happen.

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: oh wait I think we will have to set it in context but it will work with useCallback. Lemme try.

saurabhdaware: done. should look better now.

anuraghazra (approved):

PR #1156, opened 2023-05-04
fix(cardFooter): alignment issue

### LINKED ISSUE #1140: Bug: Issues in the Card component
**Problem**
While rendering card component, if we do not pass `CardFooterLeading` component and try to render `CardFooterTrailing` only, the buttons are rendered on the left rather than right.
I did not render the CardFooterLeading component, hence the buttons were rendered on the left
image (3)

I rendered CardFooterLeading but I did not pass any props, now the buttons do shift to the right
image (4)

Note: This behaviour only effects desktop variant.

**Expected Behaviour**
Ideally the buttons in the footer should be placed on the  bottom right.

sumitd94:
Resolves #1140 

When we do not provide either `footerTrailing` or `footerLeading` component to the `cardFooter` component, then it's always aligned to the right.

More information on this [issue](https://github.com/razorpay/blade/issues/1140)

Added a fix, so that the `footerTrailing` is alway aligned to the right.

> I did not provide the `footerLeading` component

Screenshot 2023-05-04 at 2 44 15 PM

### DIFF
--- a/packages/blade/src/components/Card/CardFooter.tsx
+++ b/packages/blade/src/components/Card/CardFooter.tsx
@@ -46,7 +46,7 @@ const _CardFooter = ({ children, testID }: CardFooterProps): React.ReactElement 
     footerChildrensArray.length && footerChildrensArray.length === 2
       ? 'space-between'
       : React.isValidElement(footerChildrensArray[0]) && footerChildrensArray[0]?.props?.actions // FooterTrailing component receives actions as props
-      ? 'end'
+      ? 'flex-end'
       : 'space-between';
 
   return (
--- a/packages/blade/src/components/Card/CardFooter.tsx
+++ b/packages/blade/src/components/Card/CardFooter.tsx
@@ -45,6 +45,10 @@ const _CardFooter = ({ children, testID }: CardFooterProps): React.ReactElement 
     throw new Error(`Invalid React Element ${footerChildrensArray}`);
   }
 
+  // the reason why we are checking for actions here is, because we want the footerTrailing
+  // to always be aligned to the right
+  // if we don't check for action here, and if we do not have footerTrailing and only footerLeading
+  // then the content of footerLeading will be justified to the end.
   const baseBoxJustifyContent =
     footerChildrensArray.length === 2 || !footerChildrensArray[0]?.props?.actions
       ? 'space-between'
--- a/packages/blade/src/components/Card/CardFooter.tsx
+++ b/packages/blade/src/components/Card/CardFooter.tsx
@@ -40,6 +40,15 @@ const _CardFooter = ({ children, testID }: CardFooterProps): React.ReactElement 
     ComponentIds.CardFooterTrailing,
   ]);
 
+  const footerChildrensArray = React.Children.toArray(children);
+
+  const baseBoxJustifyContent =
+    footerChildrensArray.length && footerChildrensArray.length === 2
+      ? 'space-between'
+      : React.isValidElement(footerChildrensArray[0]) && footerChildrensArray[0]?.props?.actions // FooterTrailing component receives actions as props
+      ? 'end'
+      : 'space-between';
+
   return (
     <BaseBox marginTop="auto" {...metaAttribute({ name: MetaConstants.CardFooter, testID })}>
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/Card/CardFooter.tsx
```diff
@@ -40,6 +40,15 @@ const _CardFooter = ({ children, testID }: CardFooterProps): React.ReactElement
     ComponentIds.CardFooterTrailing,
   ]);
 
+  const footerChildrensArray = React.Children.toArray(children);
+
+  const baseBoxJustifyContent =
+    footerChildrensArray.length && footerChildrensArray.length === 2
+      ? 'space-between'
+      : React.isValidElement(footerChildrensArray[0]) && footerChildrensArray[0]?.props?.actions // FooterTrailing component receives actions as props
```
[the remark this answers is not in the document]
sumitd94: That wont work, have given my explanation here - https://github.com/razorpay/blade/issues/1140#issuecomment-1532738114

packages/blade/src/components/Card/CardFooter.tsx
```diff
@@ -40,6 +40,15 @@ const _CardFooter = ({ children, testID }: CardFooterProps): React.ReactElement
     ComponentIds.CardFooterTrailing,
   ]);
 
+  const footerChildrensArray = React.Children.toArray(children);
+
+  const baseBoxJustifyContent =
+    footerChildrensArray.length && footerChildrensArray.length === 2
```
saurabhdaware, line 46:
```suggestion
    footerChildrensArray.length === 2
```

this should be enough here right?

saurabhdaware: === 2 itself will ignore any falsy values

packages/blade/src/components/Card/CardFooter.tsx
```diff
@@ -40,6 +40,15 @@ const _CardFooter = ({ children, testID }: CardFooterProps): React.ReactElement
     ComponentIds.CardFooterTrailing,
   ]);
 
+  const footerChildrensArray = React.Children.toArray(children);
+
```
saurabhdaware, line 44: Is this an equivalent condition?

```ts
  const footerChildrensArray = React.Children.toArray(children);
  if (!React.isValidElement(footerChildrensArray[0]) {
    throw new Error(`Invalid React Element ${footerChildrensArray}`);
  }

 const baseBoxJustifyContent = footerChildrensArray.length === 2 || !footerChildrensArray[0]?.props?.actions ? 'space-between' : 'end'
```

packages/blade/src/components/Card/CardFooter.tsx
sumitd94: makes sense, thanks

packages/blade/src/components/Card/CardFooter.tsx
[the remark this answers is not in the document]
sumitd94: @anuraghazra Refactored the condition for better readability.
But we still need the check for `props.action`

packages/blade/src/components/Card/CardFooter.tsx
```diff
@@ -40,6 +40,16 @@ const _CardFooter = ({ children, testID }: CardFooterProps): React.ReactElement
     ComponentIds.CardFooterTrailing,
   ]);
 
+  const footerChildrensArray = React.Children.toArray(children);
+  if (!React.isValidElement(footerChildrensArray[0])) {
+    throw new Error(`Invalid React Element ${footerChildrensArray}`);
+  }
+
+  const baseBoxJustifyContent =
+    footerChildrensArray.length === 2 || !footerChildrensArray[0]?.props?.actions
```
anuraghazra, line 49: Let's add a comment here explaining why we need this props.actions check.

anuraghazra (approved):
Update or add tests as necessary.

sumitd94:
> Update or add tests as necessary.

Have already checked, don't think we need any tests update here because it's an alignment issue mostly.

anuraghazra (approved):

saurabhdaware (approved):
Looks good to me. Thanks for the contribution 🎉

Issue #914, opened 2022-12-12
Fonts are not working in Sandpack and CodeSandbox examples


saurabhdaware:
The font in https://blade.razorpay.com/?path=/story/recipes-simple-dashboard--simple-dashboard&globals=measureEnabled:false and other codesandbox and sandpack examples is not Lato.

Looks like some font loading issues

<img width="696" alt="image" src="https://user-images.githubusercontent.com/30949385/207024957-09458807-faad-4f05-8729-76e95a10684a.png">

Reported by @chaitanyadeorukhkar on Slack: https://razorpay.slack.com/archives/G01B3LQ9H0W/p1670831023173849

ashutosh887:
Should I work on this @saurabhdaware 🙂😅

PR #1161, opened 2023-05-08
docs(Dropdown): ActionList, Dropdown docs refactor

saurabhdaware:


packages/blade/src/components/Dropdown/docs/stories.ts
divyanshu013, line 1: How easy is maintaining this file? All the code is inside template literal 😅

packages/blade/src/components/Dropdown/docs/DropdownDocs.tsx
```diff
@@ -0,0 +1,133 @@
...
+  Playground,
+  WithControlledMenuStory,
+  WithControlledSelectStory,
+  WithSimpleMenuStory,
+} from './stories';
+import StoryPageWrapper from '~src/_helpers/storybook/StoryPageWrapper';
+import { Code, Heading, Text, Title } from '~components/Typography';
+import { Sandbox, VerticalSandbox } from '~src/_helpers/storybook/Sandbox';
+import { Box } from '~components/Box';
+// import BaseBox from '~components/Box/BaseBox';
```
divyanshu013, line 13: Stray comment
```suggestion
```

divyanshu013 (commented):
Looks nice, though will add to maintenance 🥲 

- Also, please check the chromatic snapshots.
- Noticing font diff in the sandpack stories, is the font not loading properly?
- How is this rendering on native? Could you attach screenshots?
- Where are the `DropdownButton` props?

packages/blade/src/components/Dropdown/docs/stories.ts
saurabhdaware: Not easy tbh. We miss out on all the typescript. I don't see any other option though 😢 Most live editors will take code as string.

For chromatic and react native, I've added some internal stories that are plain stories without sandpack. So they should give some signs in case we break something.

packages/blade/src/components/ActionList/docs/propsTable.tsx
```diff
@@ -0,0 +1,62 @@
+import type { ActionListItemProps, ActionListSectionProps } from '../ActionListItem';
+import type { ActionListProps } from '../ActionList';
+import { Code } from '~components/Typography';
+import { ScrollLink } from '~src/_helpers/storybook/ScrollLink';
+
+type PropsTableType<T> = Record<
+  keyof Omit<T, 'testID' | '_hideDivider' | '_index'>,
+  string | JSX.Element | { note: string; type: string | JSX.Element }
+>;
+const actionListPropsTables: {
```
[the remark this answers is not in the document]
saurabhdaware: true but we anyway have to maintain props and their types separately in storybook file to support appropriate stories.

That was in fact a bit more difficult than this because we had to create fake props like "actionListItemLeadingIcon" to match it to `<ActionListItem leading={<ActionListItemIcon />} />` component. 

This will be learning curve for anyone working but still simpler to maintain that the storybook controls mapping

packages/blade/src/components/Dropdown/docs/stories.ts
saurabhdaware: Also we will have to keep code inside template inside literal irrespective of whether we go with this approach or not because the code we get from `show code` is just simply wrong and to override that in storybook, we anyway have to pass string literal with static code.

E.g. this is what we get in show code of controlled dropdown. This doesn't cover the useState code and messes up the callback functions. There's no way for someone to read this and figure out how to do state management for controlled dropdown - 
image


This will be comparatively better approach because in storybook PR preview Sandpacks at least the code will break if something is outdated or invalid.

saurabhdaware:
> Also, please check the chromatic snapshots.


> Noticing font diff in the sandpack stories, is the font not loading properly?

Yupp its broken #914 😢 

> How is this rendering on native? Could you attach screenshots?

My React Native setup is broken with that upgrade 😢 Ideally it should just write plain text code examples in sandpack examples and add the internal stories which will be the ones that work properly

> Where are the DropdownButton props?

We don't have any unique props there. I think I'll just add a link to Button props and mention its the same.

packages/blade/src/components/Dropdown/docs/stories.ts
divyanshu013: I think we can go ahead since you've already gone through the efforts for improving the docs, let's revisit if we start seeing this hampering maintenance. I was thinking if there was a way for storybook to read a TSX file as a file (serialized) we could've been able to pass that as a string, might need exploration though.

divyanshu013 (dismissed):
Looks good to me, please check the chromatic snapshots too.

kamleshchandnani (approved):
overall LGTM, can we just move the `Stories/WithSelect` and `Stories/WithButton` to top level itself? `WithSelect` and `WithButton` so its easily discoverable?

saurabhdaware:
Added link to one of the stories in react native to open it in web. Seems to work fine. Will add other links separately anyway the internal components work as it is in react native since they are not under sandpack.
image image

divyanshu013 (approved):

PR #1174, opened 2023-05-16
feat(blade): added bottomsheet component ids 

anuraghazra:


saurabhdaware (approved):
Looks good to me.

Offtopic: Do we want to support styled-props on BottomSheet? I think it can be misused right?

anuraghazra:
> Offtopic: Do we want to support styled-props on BottomSheet? I think it can be misused right?

Yes probably should not add them, Could mess with dragging behaviour if users pass margins/paddings in wrong places.

saurabhdaware (approved):

PR #1189, opened 2023-05-23
docs(Accordion): add api spec

divyanshu013:
See the [formatted markdown](https://github.com/razorpay/blade/blob/div/accordion-api/packages/blade/src/components/Accordion/_decisions/decisions.md).


packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+  </AccordionItem>
+  <AccordionItem title="How can I transfer money to customers?" description="Just use Razorpay" />
+</Accordion>;
+```
+
+### Accordion
+
+| Prop         | Type                   | Default     | Description                                                    | Required |
+| ------------ | ---------------------- | ----------- | -------------------------------------------------------------- | -------- |
+| variant      | `unordered`, `ordered` | `unordered` | `ordered` shows a numeric prefix to titles                     |          |
```
chaitanyadeorukhkar, line 43: Why do we need `unordered` here? I can only see number & icon on Figma. 

Also, do we need to call it ordered or do we call it `numeric`? or even `showNumberPrefix={true}`

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+> - `onChange` callback signature `({ index }) => {}`, `index` represents the expanded item's index
+
+### AccordionItem
+
+| Key         | Type            | Default     | Description                                                         | Required |
+| ----------- | --------------- | ----------- | ------------------------------------------------------------------- | -------- |
+| title       | `string`        | `undefined` | Title text content                                                  | ✅       |
+| description | `string`        | `undefined` | Body text content                                                   |          |
+| icon        | `IconComponent` | `undefined` | Renders a Blade icon as title prefix (requires `unordered` variant) |          |
+| children    | `JSX`           | `undefined` | Slot, renders any custom content                                    |          |
```
chaitanyadeorukhkar, line 63: Should we add some restriction to disallow using an accordion inside an accordion here?

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+> - For using accordion in controlled state, use `index` coupled with `onChange`
+> - `onChange` callback signature `({ index }) => {}`, `index` represents the expanded item's index
+
+### AccordionItem
+
+| Key         | Type            | Default     | Description                                                         | Required |
+| ----------- | --------------- | ----------- | ------------------------------------------------------------------- | -------- |
+| title       | `string`        | `undefined` | Title text content                                                  | ✅       |
+| description | `string`        | `undefined` | Body text content                                                   |          |
+| icon        | `IconComponent` | `undefined` | Renders a Blade icon as title prefix (requires `unordered` variant) |          |
```
chaitanyadeorukhkar, line 62: Oh I see, we're using `unordered` to indicate usage of Icon. 

Do we throw TS/JS errors when using `ordered` with `icon` or `unordered` without `icon`?

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+        description="Just use Razorpay"
+      />
+    </Accordion>
+  );
+};
+```
+
+## Alternatives
+
+- Instead of `index`, `defaultIndex` considered `value`, `defaultValue`, however it doesn't co-relate well with how these APIs work in other components where they're mostly being used for user inputs (eg. in forms) and therefore termed values. On the other hand, accordion is mostly a presentational component and shouldn't be used for user inputs _(what do we treat as value in accordion)_.
```
chaitanyadeorukhkar, line 124: `index` and `defaultIndex` makes sense to me.

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+- Instead of `index`, `defaultIndex` on the root `Accordion` component, considered an approach to instead put relatable props such as `isExpanded`, `isDefaultExpanded` on the child `AccordionItem` component. However, with this, controlled usecase becomes tricky since we would still ideally want a single callback listener for `onChange`:
```
chaitanyadeorukhkar, line 125: This would be too confusing imo

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+Accordion shares the toggle interaction with Collapsible. We can create abstractions that work for both plus other components such as tabs, switch, etc. (different from this public API spec):
+
+- Build accordion interaction on top of a base collapsible component
+- A disclosure based hook (references - [Chakra](https://chakra-ui.com/docs/hooks/use-disclosure), [Mantine](https://mantine.dev/hooks/use-disclosure/), [Paste](https://paste.twilio.design/components/disclosure))
+
+Find in page (automatically expanding the accordion if someone does <kbd>Cmd</kbd> + <kbd>F</kbd>) is a tricky feat, need to check feasibility and should be taken up as an enhancement (references - [blog by Anurag](https://dev.to/anuraghazra/thoughts-on-find-in-page-with-custom-accordion-elements-5573), [`hidden="until-found"`](https://github.com/WICG/display-locking/blob/main/explainers/hidden-content-explainer.md)).
+
+## Open questions
+
+**Q1.** Is the deviation in props and using index for controlled state a fine tradeoff? Please post alternatives with usecase and rationale otherwise.
```
chaitanyadeorukhkar, line 143: Seems good to me

packages/blade/src/components/Accordion/_decisions/decisions.md
chaitanyadeorukhkar, line 62: Are we controlling the size & color of the icon? Size I understand but won't we need to give some flexibility on Icon color?

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+## API
+
+Annotated component with props ([figma](https://www.figma.com/file/wzCTQ9kMjWWHKFu7kCzYGi/Accordion-breakdown?type=design&node-id=0-1&t=GfrkTz89y2M46wQN-0)):
+
+![Accordion breakdown](./accordion-breakdown.png)
+
+Sample usage, composes `Accordion` and `AccordionItem`:
+
+```jsx
+import { Accordion, AccordionItem } from '@razorpay/blade';
```
kamleshchandnani, line 26: On design I see we are using `AccordionGroup` and `AccordionItem` can we keep both in sync? my vote will be `Accordion` and `AccordionItem` itself

packages/blade/src/components/Accordion/_decisions/decisions.md
kamleshchandnani: +1

We shall keep it inline with design. `leading?: number` even icon will be directly applied on Item and we shall then also have a restriction when someone has passed leading value and then still passing icon in the accordion so both shouldn't be allowed at the same time

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+  <AccordionItem title="How can I transfer money to customers?" description="Just use Razorpay" />
+</Accordion>;
+```
+
+### Accordion
+
+| Prop         | Type                   | Default     | Description                                                    | Required |
+| ------------ | ---------------------- | ----------- | -------------------------------------------------------------- | -------- |
+| variant      | `unordered`, `ordered` | `unordered` | `ordered` shows a numeric prefix to titles                     |          |
+| defaultIndex | `number`               | `undefined` | Makes the passed item index expanded by default (uncontrolled) |          |
```
kamleshchandnani, line 44: do we have a use case for this?

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+| index        | `number`               | `undefined` | Expands the passed index (controlled)                          |          |
```
kamleshchandnani, line 45: probably we shall name this something explicit - `expandedItemIndex` or something like that?

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+> - `defaultIndex` can accept an accordion item index to render the accordion in an uncontrolled state
+> - For using accordion in controlled state, use `index` coupled with `onChange`
+> - `onChange` callback signature `({ index }) => {}`, `index` represents the expanded item's index
+
+### AccordionItem
+
+| Key         | Type            | Default     | Description                                                         | Required |
+| ----------- | --------------- | ----------- | ------------------------------------------------------------------- | -------- |
+| title       | `string`        | `undefined` | Title text content                                                  | ✅       |
+| description | `string`        | `undefined` | Body text content                                                   |          |
```
kamleshchandnani, line 61: shouldn't descript be required as well? without description we shall not allow using Accordion?

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
...
+| children    | `JSX`           | `undefined` | Slot, renders any custom content                                    |          |
+
+## a11y
+
+Web:
+
+Follows the standard practices (aria expanded, controls, etc.) [ARIA guide](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/), [example](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/#rps_label):
+
+- <kbd>Enter</kbd> or <kbd>Space</kbd> can toggle accordion
+- <kbd>Tab</kbd> key to cycle focus
```
kamleshchandnani, line 72: will the tab will trap focus? like you start at first item and then when you tab tab tab and after you reach the last item shall it go to next focussable element or go to first item in the group again?

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,159 @@
+# Accordion
+
+An accordion is used to allow users to toggle between different content sections in a compact vertical stack.
```
kamleshchandnani, line 3: lets also somehwere mention that behavior is similar to checkbox group?

packages/blade/src/components/Accordion/_decisions/decisions.md
divyanshu013: Made it `ordered` to keep it similar to `List` component (because we want the nos to be auto added in accordion items). Alternatively can add a boolean for controlling to the root accordion.

packages/blade/src/components/Accordion/_decisions/decisions.md
divyanshu013: On design side there isn't any restriction which means people can do crazy things like putting their entire app inside accordion 😸 

Think it should be fine to avoid checks for this one

packages/blade/src/components/Accordion/_decisions/decisions.md
divyanshu013: Yes, will throw a runtime error if someone is using icon on accordion item while making the root accordion show number prefixes

divyanshu013: No design usecase currently so the color would be fixed

packages/blade/src/components/Accordion/_decisions/decisions.md
divyanshu013: Yes, for an uncontrolled accordion where we want to show a particular item expanded by default (on loading the page). Stock behavior is all accordion items will appear collapsed (added details in note section)

packages/blade/src/components/Accordion/_decisions/decisions.md
divyanshu013: Yes, got the clarification from design that this shall be always required (even on design side), so someone can't just use a slot without providing any description.

packages/blade/src/components/Accordion/_decisions/decisions.md
divyanshu013: Added in https://github.com/razorpay/blade/pull/1189/commits/f6a8627de1aa246be547492accb9f7052dc8381e

packages/blade/src/components/Accordion/_decisions/decisions.md
divyanshu013: Added in https://github.com/razorpay/blade/pull/1189/commits/f6a8627de1aa246be547492accb9f7052dc8381e

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,169 @@
...
+> - For using accordion in controlled state, use `index` coupled with `onChange`
+> - `onChange` callback signature `({ expandedIndex }) => {}`, `expandedIndex` represents the expanded item's index
+
+### AccordionItem
+
+| Key         | Type            | Default     | Description                                                                | Required |
+| ----------- | --------------- | ----------- | -------------------------------------------------------------------------- | -------- |
+| title       | `string`        | `undefined` | Title text content                                                         | ✅       |
+| description | `string`        | `undefined` | Body text content                                                          | ✅       |
+| icon        | `IconComponent` | `undefined` | Renders a Blade icon as title prefix (requires `showNumberPrefix={false}`) |          |
```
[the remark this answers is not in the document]
divyanshu013: Future: might need to also think about a generic `Icon` component or `createIcon` util that can accept any SVG so we can utilize it for such needs.

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,169 @@
...
+</Accordion>;
+```
+
+### Accordion
+
+| Prop                 | Type       | Default     | Description                                                    | Required |
+| -------------------- | ---------- | ----------- | -------------------------------------------------------------- | -------- |
+| defaultExpandedIndex | `number`   | `undefined` | Makes the passed item index expanded by default (uncontrolled) |          |
+| expandedIndex        | `number`   | `undefined` | Expands the passed index (controlled)                          |          |
+| onChange             | `function` | `undefined` | Callback for change in any item's expanded state (controlled)  |          |
```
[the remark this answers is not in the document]
divyanshu013: Fixed in https://github.com/razorpay/blade/pull/1189/commits/3fa17cbb5045c3be0a5d0309728e9604fdc6feef

packages/blade/src/components/Accordion/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: Confirmed that currently these will only be Blade Icons

packages/blade/src/components/Accordion/_decisions/decisions.md
parichay28, line 1: Will we be able to expand more than one accordion item at a time?

divyanshu013: Nope, that's a design restriction. If that changes in the future then yes we'll extend the capabilities on dev side too.

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,167 @@
...
+
+https://github.com/razorpay/blade/assets/6682655/1a02d680-1832-4a70-a0a3-1a1f56f2b8b9
+
+### Accordion close
+
+https://github.com/razorpay/blade/assets/6682655/b54eaeb9-7257-4da4-bf3b-89e10785e4f9
+
+## API
+
+Annotated component with props ([figma](https://www.figma.com/file/wzCTQ9kMjWWHKFu7kCzYGi/Accordion-breakdown?type=design&node-id=0-1&t=GfrkTz89y2M46wQN-0)):
```
cseas, line 27: Need access

Screenshot 2023-05-30 at 10 36 28 PM

cseas: Found the design [here][1], I can access this one. Please update the link.

[1]: https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=31097-721626&t=S5YhzX8uYWUwLZqj-4

divyanshu013: That one has restricted access. It's an annotated version of the component with props (screenshot included in the markdown). The full component figma is the one prior to this section.

cseas: Why add restricted links in a public doc even Razorpay employees don't have access to? 😕

divyanshu013: Just kept it for record, will clean up if it's causing inconvenience :)

chaitanyadeorukhkar (approved):

anuraghazra (approved):

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,167 @@
...
+  </AccordionItem>
+  <AccordionItem title="How can I transfer money to customers?" description="Just use Razorpay" />
+</Accordion>;
+```
+
+### Accordion
+
+| Prop                 | Type       | Default     | Description                                                    | Required |
+| -------------------- | ---------- | ----------- | -------------------------------------------------------------- | -------- |
+| defaultExpandedIndex | `number`   | `undefined` | Makes the passed item index expanded by default (uncontrolled) |          |
```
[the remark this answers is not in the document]
divyanshu013: There might be contextual cases where the most important item might not be the first one so proposed this as a `number` rather than a `boolean` that only allows default expanding the first item

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -0,0 +1,167 @@
...
+| expandedIndex        | `number`   | `undefined` | Expands the passed index (controlled)                          |          |
+| onChange             | `function` | `undefined` | Callback for change in any item's expanded state               |          |
+| showNumberPrefix     | `boolean`  | `false`     | Adds numeric index at the beginning of items                   |          |
+
+> **Note**
+>
+> - Also includes layout based styling props
+> - By default the accordion renders in all items collapsed state, at max only one item can be expanded at a time (design restriction)
```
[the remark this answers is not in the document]
divyanshu013: This is a design restriction that we should only allow expanding one item at a time. Should that change in the future we can also extend the usecase in accordion component.

PR #1201, opened 2023-05-25
feat(Box): add `ref` support

saurabhdaware:
- Adds support for ref in Box component
- Adds tests

### DIFF
--- a/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
+++ b/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
@@ -1,9 +1,11 @@
+import type { View } from 'react-native';
 import type { CSSObject } from 'styled-components';
 import type { MarginProps, PaddingProps, SpacingValueType } from './spacingTypes';
 import type { MakeObjectResponsive } from './responsiveTypes';
 import type { Theme } from '~components/BladeProvider';
 import type { Border } from '~tokens/global';
 import type { DotNotationColorStringToken, PickCSSByPlatform, TestID } from '~src/_helpers/types';
+import type { Platform } from '~utils';
 
 type LayoutProps = MakeObjectResponsive<
   {
@@ -206,4 +208,11 @@ type BaseBoxProps = Omit<BoxProps, keyof BoxVisualProps> &
     }
   >;
 
-export { BaseBoxProps, BoxProps, StyledPropsBlade, validBoxAsValues };
+// ref prop type
+type BoxRefType = Platform.Select<{
+  web: Omit<HTMLDivElement, 'style'>;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  native: View;
+}>;
+
+export { BaseBoxProps, BoxProps, BoxRefType, StyledPropsBlade, validBoxAsValues };
--- a/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
+++ b/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
@@ -210,8 +210,7 @@ type BaseBoxProps = Omit<BoxProps, keyof BoxVisualProps> &
 
 // ref prop type
 type BoxRefType = Platform.Select<{
-  web: Omit<HTMLDivElement, 'style'>;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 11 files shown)


packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -206,4 +208,11 @@ type BaseBoxProps = Omit<BoxProps, keyof BoxVisualProps> &
     }
   >;
 
-export { BaseBoxProps, BoxProps, StyledPropsBlade, validBoxAsValues };
+// ref prop type
+type BoxRefType = Platform.Select<{
+  web: Omit<HTMLDivElement, 'style'>;
```
saurabhdaware, line 213: I am removing style from type so that it will atleast warn consumers if they are doing something hacky

packages/blade/src/components/Box/__tests__/Box.web.test.tsx
```diff
@@ -96,4 +99,28 @@ describe('<Box />', () => {
     }
     console.error = tempConsoleError;
   });
+
+  it('should support ref on Box', async () => {
```
saurabhdaware, line 103: couldn't figure out how I can write test for react native because it doesn't support `addEventListener` and `focus` on Box wouldn't really work as its not button / link.

packages/blade/src/components/Box/Box.stories.tsx
```diff
@@ -97,4 +99,29 @@ AsSection.args = {
   as: 'section',
 } as BoxProps;
 
+export const WithRef = (args: BoxProps): JSX.Element => {
+  const ref: React.Ref<BoxRefType> = React.useRef(null);
```
anuraghazra, line 103:
```suggestion
  const ref = React.useRef<BoxRefType>(null);
```

packages/blade/src/components/Box/__tests__/Box.web.test.tsx
anuraghazra: You can just test if the ref isn't null or has the methods like focus or etc?

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
anuraghazra: 🤔 hmm okay so Box can also accept "as" prop so you need to ensure the this type will be compatible with that. 

See if just HTMLElement might work or not.

packages/blade/src/components/Box/Box.stories.tsx
```diff
@@ -97,4 +99,29 @@ AsSection.args = {
...
+      </Button>
+      <Box ref={ref} {...args}>
+        <Text>Hi from Box with ref</Text>
+      </Box>
+    </Box>
+  );
+};
+
+WithRef.args = {
+  marginTop: '800px',
```
anuraghazra, line 124: Let just put it directly on the box itself? does it need to be configurable?

anuraghazra (approved):
LGTM

packages/blade/src/components/Box/Box.stories.tsx
saurabhdaware: yes I mean this is what we follow for stories so it correct shows up in controls otherwise it shows blank in controls but has value in actual story

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -206,4 +208,11 @@ type BaseBoxProps = Omit<BoxProps, keyof BoxVisualProps> &
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
saurabhdaware: ops old comment. missed it.

divyanshu013 (approved):
Let's also add a basic test for native.

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
saurabhdaware: yup makes sense. turned it into HTMLElement.

divyanshu013 (dismissed):
:shipit:

divyanshu013 (approved):
:shipit:

PR #1205, opened 2023-05-25
feat(Box): add onMouse events

### LINKED ISSUE #1163: Hover pseudoClass and eventListener support on Blade components
It would be good to be able to do something like -
```ts
<Box _hover={{ cssStyles }} onHover={(e) => doSomething()} />
```

This could be beneficial for multiple use cases

- Sending analytics call if user hovers over some element
- Modify styles if user hovers over some element
- On razorpay.com navbar, we need the `onHover` method to show extended navbar items when hovering over menu items like 'Payments' or 'Banking+'

saurabhdaware:
- Add `onMouseOver`, `onMouseEnter`, `onScroll`, and `onMouseLeave` event
- Add tests and docs

Partially resolves #1163 (`onMouseOver` part is resolved. `_hover` part is not)

### DIFF
--- a/.changeset/warm-apples-try.md
+++ b/.changeset/warm-apples-try.md
@@ -2,4 +2,4 @@
 "@razorpay/blade": minor
 ---
 
-feat(Box): add `onMouseOver`, `onMouseEnter`, `onMouseLeave` events
+feat(Box): add `onScroll`, `onMouseOver`, `onMouseEnter`, `onMouseLeave` events
--- a/.changeset/warm-apples-try.md
+++ b/.changeset/warm-apples-try.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": minor
+---
+
+feat(Box): add `onMouseOver`, `onMouseEnter`, `onMouseLeave` events
--- a/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
+++ b/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
@@ -217,8 +217,9 @@ type BoxCallbackProps = Omit<
        * Use this for hoverable containers in cases like custom menus.
        */
       onMouseLeave: React.MouseEventHandler<HTMLElement>;
+      onScroll: React.UIEventHandler<HTMLElement>;
     };
-    native: Record<'onMouseOver' | 'onMouseEnter' | 'onMouseLeave', undefined>;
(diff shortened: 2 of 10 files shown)


chaitanyadeorukhkar (approved):

chaitanyadeorukhkar (dismissed):

.changeset/warm-apples-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": minor
+---
+
+feat(Box): add `onMouseOver`, `onMouseEnter`, `onMouseLeave` events
```
divyanshu013:
```suggestion
feat(Box): add `onMouseOver`, `onMouseEnter`, `onMouseLeave`, `onScroll` events
```

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -187,13 +187,51 @@ type StyledPropsBlade = Partial<
...
+type BoxCallbackProps = Omit<
+  Platform.Select<{
+    web: {
+      /**
+       * **Warning**
+       *
+       * Make sure to not use Box when you want to create a trigger that performs action on hover.
+       * You would probably want to render it as `button` using `styled.button` instead.
+       *
+       * Use this for hoverable containers in cases like custom menus.
```
divyanshu013, line 199: Don't know if this would work for custom menus because a lot of interactive things won't work like keyboard navigation, a11y (eg. <kbd>Tab</kbd>)

divyanshu013: Users will have to handle these things themselves (focus management, etc.)

.changeset/warm-apples-try.md
saurabhdaware: added already 🙈

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
saurabhdaware: yes yes I meant it can be used as base element there since now we can put `onMouse` events there. Keyboard events will anyway go to `window` or some focusable item like `link` that triggered the hover menu.

kamleshchandnani: we shall rather create eslint rules for all the warnings in jsdoc. because i doubt people barely read JSDoc unless stuck or need more details. Even i won't read when I'm typing in flow

PR #1208, opened 2023-05-26
feat(Button): add `href` to Button to render it as anchor tag

### LINKED ISSUE #1166: Button: Support rendering as anchor tag with href
We currently don't have a way where we can pass `href` to our Button component. Ideal API in my opinion would be where consumers can pass `href` prop and based on that, we render the button as anchor tag with that href.

```jsx
<Button href="/signup">Sign Up</Button> // renders as `<a href="/signup">Sign Up</a>`
```

## Original Request

              No, it doesn't matter which variant of button is used. The button component in blade doesn't support `to` or `href` attribute. My question is how do I redirect to another page using button component from blade (Is `onClick` the only option here?)

@saurabhdaware @divyanshu013

_Originally posted by @Kushagra-ag in https://github.com/razorpay/blade/issues/1165#issuecomment-1543885548_

saurabhdaware:
This PR adds `href`, `target`, `rel` property to Button

```jsx
<Button href="/">Something</Button> // turns into -> <a href="/">Something</a>
```

Resolves #1166 

TODO:
- [x] Tests
- [x] Snapshots


packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -372,10 +392,23 @@ const _BaseButton: React.ForwardRefRenderFunction<BladeElementRef, BaseButtonPro
...
   return (
     <StyledBaseButton
       ref={buttonRef as any}
-      accessibilityProps={{ ...makeAccessible({ role: 'button', label: accessibilityLabel }) }}
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error
+      // @ts-ignore: On React Native it will always be undefined but TS doesn't understand that
+      as={renderElement}
+      href={href}
+      target={target}
+      rel={rel}
```
chaitanyadeorukhkar, line 405: We auto-set rel based on href. Can we add that here?

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -22,9 +22,21 @@ const StyledPressable = styled(Animated.createAnimatedComponent(Pressable))<
...
 });
 
+const openURL = async (href: string): Promise<void> => {
+  try {
+    const canOpen = await Linking.canOpenURL(href);
+    if (canOpen) {
+      await Linking.openURL(href);
+    }
+  } catch {
+    console.warn(`[Blade: BaseLink]: Could not open the link "href=${href}"`);
```
chaitanyadeorukhkar, line 32:
```suggestion
    console.warn(`[Blade: BaseButton]: Could not open the link "href=${href}"`);
```

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -75,13 +87,27 @@ const _StyledBaseButton: React.ForwardRefRenderFunction<TextInput, StyledBaseBut
...
 
+  const handleOnPress = (event: GestureResponderEvent): void => {
+    if (href) {
+      void openURL(href);
+    }
+
+    if (onClick) {
+      /*
+      React Native's Pressable's onClick returns a GestureResponderEvent but our types expect a SyntheticEvent.
+      Until we have a way to handle platform specific types, we will have to ignore this TS error.
```
chaitanyadeorukhkar, line 98: We have a way to handle platform specific types, no?

packages/blade/src/components/Button/BaseButton/getStyledBaseButtonStyles.ts
```diff
@@ -29,6 +29,10 @@ const getBaseButtonStyles = ({
   paddingBottom: buttonPaddingBottom,
   paddingLeft: buttonPaddingLeft,
   paddingRight: buttonPaddingRight,
+  display: 'inline-flex',
+  justifyContent: 'center',
+  alignItems: 'center',
```
chaitanyadeorukhkar, line 34: Just check once if full width buttons are working okay on native. I recall facing some issues with full width when we give align and justify properties but I think that was with alignSelf

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
saurabhdaware: ops it was stale comment from BaseLink. Its not valid for BaseButton anyway. Removing it.

cseas:
Bump. This use-case is blocking our development.
- https://razorpay.slack.com/archives/CMQ3RBHEU/p1685441325576299
- https://razorpay.atlassian.net/browse/AQP-1101

packages/blade/src/components/Button/BaseButton/getStyledBaseButtonStyles.ts
saurabhdaware: image

Seems to look good.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
saurabhdaware: done

chaitanyadeorukhkar (dismissed):

divyanshu013 (dismissed):
Looks good to me.

Minor issues:
- disabled link is also working
- disabled link is also focussable (not sure about the exact html guidelines)

saurabhdaware:
> disabled link is also working
> disabled link is also focussable (not sure about the exact html guidelines)

Looks like links can't be disabled. Even in our Link component, `isDisabled` does not work when variant is not `anchor` https://codesandbox.io/s/blade-disabled-link-0q1rud?file=/src/App.tsx

I think same would end up being the case in Button as Link as well.

packages/blade/package.json
```diff
@@ -276,7 +276,7 @@
       "name": "Import Button only",
       "path": "./build/components/index.web.js",
       "import": "{ Button }",
-      "limit": "17 kb",
+      "limit": "17.5 kb",
```
divyanshu013, line 279: Button keeps growing 😅

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -15,19 +16,29 @@ import type {
...
   intent?: 'positive' | 'negative' | 'notice' | 'information' | 'neutral';
   contrast?: 'low' | 'high';
   icon?: IconComponent;
   iconPosition?: 'left' | 'right';
   onClick?: (event: SyntheticEvent) => void;
+  onBlur?: Platform.Select<{
+    native: (event: GestureResponderEvent) => void;
+    web: (event: React.FocusEvent<HTMLButtonElement>) => void;
+  }>;
+  onKeyDown?: Platform.Select<{
```
kamleshchandnani, line 38: adding these 2 events was an ask from any consumer?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -315,9 +335,11 @@ const _BaseButton: React.ForwardRefRenderFunction<BladeElementRef, BaseButtonPro
   },
   ref,
 ) => {
+  const isLink = Boolean(href);
   const childrenString = getStringFromReactText(children);
   const buttonRef = useBladeInnerRef(ref);
-  const disabled = isLoading || isDisabled;
+  // Button cannot be disabled when its rendered as Link
+  const disabled = isLoading || (isDisabled && !isLink);
```
kamleshchandnani, line 342: why so? if a link can be disabled then why not a button as a link? imagine an icon button - a particular user is not authorised to click it(regardless of a `button` or `a`) so link shouldn't be treated as special case?

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -18,13 +18,26 @@ const StyledPressable = styled(Animated.createAnimatedComponent(Pressable))<
   return {
     ...getStyledBaseButtonStyles(props),
     alignSelf: 'center',
+    display: 'flex',
     ...styledPropsCSSObject,
   };
 });
 
+const openURL = async (href: string): Promise<void> => {
```
kamleshchandnani, line 26: can't we use the link's openURL directly here?

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -75,13 +88,23 @@ const _StyledBaseButton: React.ForwardRefRenderFunction<TextInput, StyledBaseBut
     };
   });
 
+  const handleOnPress = (event: GestureResponderEvent): void => {
+    if (href) {
+      void openURL(href);
```
kamleshchandnani, line 93: it should return after this right? do we intend to execute onclick and href both?

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -9,6 +10,22 @@ import type { StringChildrenType, TestID } from '~src/_helpers/types';
...
+  /**
+   * Automatically renders button with `a` tag with `href` on web
+   */
+  href?: BaseButtonProps['href'];
+  /**
+   * anchor target attribute
+   *
+   * Should only be used alongside `href`
+   */
+  target?: BaseButtonProps['target'];
```
kamleshchandnani, line 22: can't we use these props from Link?

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -9,6 +10,22 @@ import type { StringChildrenType, TestID } from '~src/_helpers/types';
 import { assignWithoutSideEffects } from '~src/utils/assignWithoutSideEffects';
 
 type ButtonCommonProps = {
```
kamleshchandnani, line 12: what happens when someone passed onclick and href both?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
saurabhdaware: Link can't be disabled. Only `<Link variant="button" />` can be disabled.

If you go here, and change to `variant="anchor"` in story controls, it will end up getting enabled. https://blade.razorpay.com/?path=/story/components-link--disabled-link-button&globals=measureEnabled:false

In HTML itself, anchor tag can't be disabled.

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
saurabhdaware: Yes. People can have their analytics events in `onClick`. Just keeping this close to vanilla JS HTML behaviour. This is also similar in Link button

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
saurabhdaware: Didnt want to import Link component inside Button but I think we can move this to utils and use at both places. Didn't add it then because its a small function that just calls React Native's Linking.openURL and throws error and error message is different in both.

PR #1210, opened 2023-05-26
feat(Typography): expose color from other typography components

saurabhdaware:
We already exposed `color` prop from Text. This PR also exposes it from `Heading`, `Title`, and `Code` component.


saurabhdaware:
> Check on figma how these tokens are looking? Are they also like surface.text.subtle... for custom colors? Basically we should have same terminology to make it easy to copy the color values.

Checked. It won't be exact dot notation. Will be `/` like we see while building components. Saurav said it will be an overkill. Thread - https://razorpay.slack.com/archives/G01B3LQ9H0W/p1685435890916149

anuraghazra (dismissed):
Looks good to me: 

TODO: 

- [x] Add Snapshot tests (web+native)
- [x] Add storybook examples with the color prop so that we get chromatic snapshots

saurabhdaware:
> Check on figma how these tokens are looking? Are they also like surface.text.subtle... for custom colors? Basically we should have same terminology to make it easy to copy the color values.

This is how it will look on figma @divyanshu013 

https://razorpay.slack.com/archives/G01B3LQ9H0W/p1685521684101769?thread_ts=1685435890.916149&cid=G01B3LQ9H0W

anuraghazra (approved):

saurabhdaware:
Ignore the color changes on `Code` component, they are fixed in #1240

divyanshu013 (approved):
Looks good to me.

PR #1216, opened 2023-05-29
docs(blade): Tooltip API

anuraghazra:
## [View Formatted Content](https://github.com/razorpay/blade/blob/anu/tooltip-api/packages/blade/src/components/Tooltip/_decisions/decisions.md)


packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,107 @@
...
+| ------------- | ------------------------------------------------------------------------ | ----------- | ------------------------------------------------------------------------- | -------- |
+| content       | `string`                                                                 | `undefined` | Content of the tooltip                                                    | ✅       |
+| placement     | `top, top-start, top-end, left, right, bottom, bottom-start, bottom-end` | `top`       | Placement of tooltip                                                      | ✅       |
+| children      | `React.ReactNode`                                                        | `undefined` | Trigger component for tooltip, Accepts any interactive element            | ✅       |
+| isOpen        | `boolean`                                                                | `undefined` | Controls the tooltip state                                                |          |
+| defaultIsOpen | `boolean`                                                                | `undefined` | If true, the tooltip will be visible initially                            |          |
```
divyanshu013, line 28: Is there a usecase for this 🤔 i.e. keeping it open on mount in uncontrolled state (a UI should ideally have only one tooltip visible at a time I'm thinking - this way there could be multiple tooltips visible)

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,107 @@
...
+| onDismiss     | `Callback`                                                               | `undefined` | Called when tooltip is closed, either by mouseout or users pressing `Esc` |          |
```
divyanshu013, line 29: Do we need `onChange`? To hook into both visible and dismiss?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Design
+
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=40540-559304&t=tmTrf3xJU6oj59fM-0) to all variants of the Tooltip component
+
+## `Tooltip` API
+
+| Prop          | Type                                                                     | Default     | Description                                                               | Required |
+| ------------- | ------------------------------------------------------------------------ | ----------- | ------------------------------------------------------------------------- | -------- |
+| content       | `string`                                                                 | `undefined` | Content of the tooltip                                                    | ✅       |
```
divyanshu013, line 24: Do we've a usecase for ellipsizing the content if it's large?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,107 @@
...
+| placement     | `top, top-start, top-end, left, right, bottom, bottom-start, bottom-end` | `top`       | Placement of tooltip                                                      | ✅       |
```
divyanshu013, line 25: Do we've usecase for top-left, left-bottom, etc.?

packages/blade/src/components/Tooltip/_decisions/decisions.md
divyanshu013: Might make more sense along with `isOpen`

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,107 @@
...
+
+// default open
+<Tooltip defaultIsOpen={true} content="Amount reversed to customer bank account">
+  <IconButton icon={InfoIcon} accessibilityLabel="Refund" />
+</Tooltip>;
+```
+
+## Library
+
+We will be using [FloatingUI](https://floating-ui.com/) to position the tooltip & handle the basic tooltip logic. Floating UI supports both Web & ReactNative.
```
divyanshu013, line 60: Perhaps add few points on bundle size

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Implementation detail nuances:
+
+To make FloatingUI work with Blade components seamlessly there are few things we need to modify:
+
+1. Expose interaction props in all the trigger components
```
divyanshu013, line 66: We had some usecases for analytics things and doing navbar like popovers on website. So this might anyway be required.

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,107 @@
...
+  - onFocus
+  - onMouseLeave
+  - onMouseMove
+  - onPointerDown
+  - onPointerEnter
+- Native:
+  - onTouchStart
+  - onTouchEnd
+
+2. Expose the actual DOM node from ref instead of only exposing [certain methods via our useBladeInnerRef](https://github.com/razorpay/blade/blob/69a1bcef2f09ceaf6f910eaaca3076055fb059a2/packages/blade/src/hooks/useBladeInnerRef.web.ts#L26-L27) hook which we used to prevent component styling misuses. We need to expose the actual DOM node because FloatingUI internally does this [isElement() check](https://github.com/floating-ui/floating-ui/blob/b8990250568043b876e1c8fe42358fe337847ede/packages/react/src/hooks/useFloating.ts#L59) on the tooltip trigger element.
```
divyanshu013, line 78: I feel this might still be okay. Restricting methods on `ref` doesn't restrict users from using `document.` APIs to access the DOM node directly anyway.

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Motion
+
+Check out the [motion guidelines] for Tooltip.
```
divyanshu013, line 82: Link / screen recording / etc.

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Accessibility
+
+- Tooltip will have [`role=tooltip`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role)
+- `Esc` should close the tooltip
+- When using keyboard tooltip will apear when target element is focused
```
divyanshu013, line 88: Another nuance, does this mean tooltip would only work on focussable components?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,107 @@
...
+
+- https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/
+- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role
+- https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html
+- https://techservicesillinois.github.io/accessibility/aria-examples/tooltip.html
+- https://dequeuniversity.com/library/aria/tooltip
+
+## Open Questions
+
+- Which components should and should not be qualify as trigger?
```
divyanshu013, line 101: On design side are there any guidelines for this? Otherwise might be difficult to standardize this and we might have to support this on almost all components 🤔

packages/blade/src/components/Tooltip/_decisions/decisions.md
saurabhdaware, line 25: Do we need this if we expose styled props on Tooltip?

We can have a default position and consumers can position the tooltip the way they like with

```jsx
<Tooltip position="absolute" left="" right="" top="" bottom="" />
```

understandable that its a little bit more work but will be consistent and might be intuitive since this what they do for other components as well

saurabhdaware (commented):
Haven't reviewed rest of API decision yet. Just thought of one opinion regarding `placement` prop.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: I've seen other libraries do onVisible onHidden but I wanted to make the api similar to our other "popover" elements, let's take a public review and see if people need to track when the state is opened or not.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: No we only have maxWidth the height can be increased as needed tho from design side there are guidelines to not make tooltip content too long.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Didn't understand 🤔 top-left == top-start, left-bottom == bottom-start can you visually show the difference?

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Sure will add. on web side with all the necessary elements it will be around 10kb 

https://bundlejs.com/?q=%40floating-ui%2Freact&treeshake=%5B%7B%0A++arrow%2Cflip%2CFloatingArrow%2Coffset%2CuseFloating%2CuseFocus%2CuseHover%2CuseInteractions%2CuseTransitionStyles%2C%7D%5D&config=%7B%22esbuild%22%3A%7B%22external%22%3A%5B%22react%22%2C%22react-dom%22%5D%7D%7D

This handles everything, collision aware placement, arrow, automatic animations,  offsets, interactions (hover, focus)

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Yes ideally tooltip should only wrap interactive elements. 

Tho i've seen some libraries wrap the non-interactive elements with a `<span tabIndex={-1} />` to make it interactive.

packages/blade/src/components/Tooltip/_decisions/decisions.md
divyanshu013: So basically if the tooltip arrow was towards left or right:

```
------
|         |
|         |
|         |> (do we need to control this placement?)
------
```

anuraghazra: I don't think this position would work, `position` !== `placement`. 
Placement is automatically computed. while position is something the user will have to maintain. 

The placement is a suggestion, if user says `placement="top"` it might not always be on top if the space isn't there for it to be. it can automatically move to "bottom". 

Plus if we let consumers position the tooltip they will inevitability mess up the spacing between the the trigger and the tooltip content resulting in inconsistencies.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Yeah we could add these. Tho note that we are exposing onMouseMove (which is required for floatingUI) not onMouseOver, mouseMove should not be used for analytics.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: This needs to be decided upon. I'll ask rastogi to get some guideline on this.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Discussing this internally.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Not seeing this on the design, I think in this case, if the tooltip's height is longer and placement=left the arrow will just stay on the middle. 

image

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: @divyanshu013 discussed this with rastogi, it seems that from design side we discourage opening tooltip by default. 
So I think better to remove it. 

But there is another problem, from design side it's also recommended not to open multiple tooltips at once, if we provide users with isOpen & defaultIsOpen they could potentially open multiples. 

I can think of two solutions: 

1. Treat tooltips like stacks and only allow 1 to be opened even when multiple are triggered

```jsx
<Tooltip content="Tooltip content 1" />
<Tooltip content="Tooltip content 1" />
<Tooltip content="Tooltip content 1" /> // only the last one opens
```
Downside of this approach is that, it's quite unexpected & hard orchestrate for users if they have multiple tooltips in multiple places and files scattered.

 2. Don't allow isOpen & defaultIsOpen. only add events (onVisible, onHidden) for analytics requirements
 
 This is cleaner quoting @saurav12 
 
 > Tooltip is something that should be triggered on demand. Therefore, we should not show multiple tooltips simultaneously.

I think this is a viable approach.

divyanshu013: Let's re-word to go with approach (2), we can reassess if there are product usecases

anuraghazra: Will be removing these props so no need of re-word. If someone raises to add isOpen and other props we will justify our thoughts.

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -7,63 +7,66 @@ The tooltip typically provides additional context about the element or its funct
...
+| --------- | ------------------------------------------------------------------------ | ----------- | ----------------------------------------------------------------------- | -------- |
+| content   | `string`                                                                 | `undefined` | Content of the tooltip                                                  | ✅       |
+| placement | `top, top-start, top-end, left, right, bottom, bottom-start, bottom-end` | `top`       | Placement of tooltip                                                    | ✅       |
+| children  | `React.ReactNode`                                                        | `undefined` | Trigger component for tooltip, Accepts any interactive element or icons | ✅       |
+| onOpen    | `Callback`                                                               | `undefined` | Called when tooltip is opened                                           |          |
+| onClose   | `Callback`                                                               | `undefined` | Called when tooltip is closed                                           |          |
```
divyanshu013, line 31: Something to ponder: should we have a single `onChange` callback with `{ isOpen / isVisible}` callback param over this and why?

anuraghazra: Hmm yeah.. Kind of leaning towards two callbacks. 

I feel it's a matter of API choice, kind of subjective but here's my thought. 

**1. Naming**

onChange naming doesn't go very well with Tooltip, when we say on"Change" what exactly is changing? `onStateChange` might be better? 
<Tooltip onChange={(isOpen) => console.log(isOpen)} \/> 

**2. Ergonomics**

I feel (and this might be subjective) two callbacks provide a better ergonomic API that 1. 

```jsx
<Tooltip 
  onOpen={sendAnalyticsOnOpen} 
  onClose={sendAnalyticsOnClose} 
 />
```

vs

```jsx
<Tooltip 
  onChange={(isOpen) => {
    if (isOpen) {
      sendAnalyticsOnOpen()
    } else {
      sendAnalyticsOnClose();
    }
  }} 
 />
```

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -98,7 +101,7 @@ Resources:
 
 ## Open Questions
 
-- Which components should and should not be qualify as trigger? 
+- Which components should and should not be qualify as trigger?
```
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/Tooltip/_decisions/decisions.md
divyanshu013: Btw on collapsible I've proposed `onChange` rather than different callbacks for expanding / collapsing, let's also gather more feedback from the team

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,160 @@
...
+
+## `Tooltip` API
+
+| Prop               | Type                                                                     | Default     | Description                                                                                                                                              | Required |
+| ------------------ | ------------------------------------------------------------------------ | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| content            | `string`                                                                 | `undefined` | Content of the tooltip                                                                                                                                   | ✅       |
+| placement          | `top, top-start, top-end, left, right, bottom, bottom-start, bottom-end` | `top`       | Placement of tooltip                                                                                                                                     | ✅       |
```
divyanshu013, line 30: Shouldn't be required

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,160 @@
...
+| shouldWrapChildren | `boolean`                                                                | `false`     | If set to true, the tooltip trigger will be wrapped in a interactive `<div>`, this is useful for making non-interactive elements work as tooltip trigger |          |
```
divyanshu013, line 31: Should we term this to signify this should be used only for non interactive elements? Like icons. Eg:
- `isNonInteractive`
- `shouldWrapNonInteractiveElement`

divyanshu013 (approved):
Minor comments, looks good otherwise

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Yes, shouldWrapNonInteractiveElement sounds better tho a bit longer. 

@chaitanyadeorukhkar @saurabhdaware any suggestions ?

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: @chaitanyadeorukhkar @saurabhdaware any thoughts?

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: We decided this, we will be allowing all components - interactive, non-interactive, icons as triggers. Here's [slack discussion](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1685603611900939)

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Design
+
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=40540-559304&t=tmTrf3xJU6oj59fM-0) to all variants of the Tooltip component
+
+## `Tooltip` API
+
+| Prop               | Type                                                                     | Default     | Description                                                                                                                                              | Required |
+| ------------------ | ------------------------------------------------------------------------ | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| content            | `string`                                                                 | `undefined` | Content of the tooltip                                                                                                                                   | ✅       |
```
saurabhdaware, line 29: Should this be `title`? that is what we follow in other components. One difference is in other components we also have subtitle / description and here we don't but should we do it for consistency?

Anyway in HTML also that tooltip-like component itself is called `<a title="something">` so title doesn't seem to out of the way 🤔 ?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,160 @@
...
+
+// custom trigger
+<Tooltip content="Amount reversed to customer bank account" shouldWrapChildren>
+  <MyCustomButton>Click me<MyCustomButton>
+</Tooltip>
+```
+
+## Library
+
+We will be using [FloatingUI](https://floating-ui.com/) to position the tooltip & handle the basic tooltip logic. Floating UI supports both Web & ReactNative.
```
saurabhdaware, line 92: Offtopic: Nice!! Will we also be able to use this DropdownOverlay?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,160 @@
...
+
+```jsx
+import type { TooltipTriggerProps } from "@razorpay/blade/components";
+
+type MyCustomButtonProps = {} & TooltipTriggerProps
+const MyCustomButton = React.forwardRef(...);
+
+// custom trigger
+<Tooltip content="Amount reversed to customer bank account" shouldWrapChildren>
+  <MyCustomButton>Click me<MyCustomButton>
```
saurabhdaware, line 86: How to open tooltip in touch devices like mobile when the trigger is interactive element? Clicking on it will just do the action right?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,160 @@
...
+| content            | `string`                                                                 | `undefined` | Content of the tooltip                                                                                                                                   | ✅       |
+| children           | `React.ReactNode`                                                        | `undefined` | Trigger component for tooltip, Accepts any interactive element or icons                                                                                  | ✅       |
+| placement          | `top, top-start, top-end, left, right, bottom, bottom-start, bottom-end` | `top`       | Placement of tooltip                                                                                                                                     |          |
+| shouldWrapChildren | `boolean`                                                                | `false`     | If set to true, the tooltip trigger will be wrapped in a interactive `<div>`, this is useful for making non-interactive elements work as tooltip trigger |          |
```
saurabhdaware, line 32: +1 to `shouldWrapNonInteractiveElement`

saurabhdaware (commented):
Nice on point doc 🥳

packages/blade/src/components/Tooltip/_decisions/decisions.md
saurabhdaware: +1 to `shouldWrapNonInteractiveElement`

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Yes. most certainly, and will fix also handle all those placement bugs.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: On touch devices the design guideline is that we should not use a tooltip to wrap an interactive button or action. Beacuse that way the UX won't be right, what happens if clicking the button opens a modal? then the tooltip will be hidden behind the modal. 

So in tooltip designers will design it sparingly/accordingly based on their usecase (and mostly on non-interactive elements like info icon beside a Input)

// @saurav12 we might need to properly document mobile behaviour on figma side also.

packages/blade/src/components/Tooltip/_decisions/decisions.md
kamleshchandnani: but popover will be a different component right? how about calling it as `description` inline with our other components strategy? because it's not really a title and if we give title then its implicit that it might be bold or something?

packages/blade/src/components/Tooltip/_decisions/decisions.md
kamleshchandnani: of what if we make it more verbose? `showOnNonInteractiveElement` or `isChildrenNonInteractive` or `shouldMakeChildTrigger`(inspiration from https://twitter.com/ericchernuka/status/1666083106611421191)?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,199 @@
...
+| onOpen             | `Callback`                                                               | `undefined` | Called when tooltip is opened                                                                                                                            |          |
+| onClose            | `Callback`                                                               | `undefined` | Called when tooltip is closed                                                                                                                            |          |
+
```
kamleshchandnani, line 35: what happens when you render a disabled element say button? will the tooltip be visible?
```
<Tooltip content="Amount reversed to customer bank account" placement="top">
  <IconButton disabled icon={InfoIcon} />
</Tooltip>
```

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,199 @@
...
+| ------------------ | ------------------------------------------------------------------------ | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| content            | `string`                                                                 | `undefined` | Content of the tooltip                                                                                                                                   | ✅       |
+| children           | `React.ReactNode`                                                        | `undefined` | Trigger component for tooltip, Accepts any interactive element or icons                                                                                  | ✅       |
+| placement          | `top, top-start, top-end, left, right, bottom, bottom-start, bottom-end` | `top`       | Placement of tooltip                                                                                                                                     |          |
```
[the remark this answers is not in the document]
kamleshchandnani: also, will this position change if there's no space on the edge of the screen? for eg: we mention placement as right but there's no space. what happens then? lets mention that in the description

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,199 @@
...
+| shouldWrapChildren | `boolean`                                                                | `false`     | If set to true, the tooltip trigger will be wrapped in a interactive `<div>`, this is useful for making non-interactive elements work as tooltip trigger |          |
+| onOpen             | `Callback`                                                               | `undefined` | Called when tooltip is opened                                                                                                                            |          |
```
kamleshchandnani, line 33: will these function have any args? the type column shall have function signature right instead of `Callback`?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,199 @@
...
+To make custom elements work the components needs to expose:
+
+- ref
+- Web:
+  - onBlur
+  - onFocus
+  - onMouseLeave
+  - onMouseMove
+  - onPointerDown
+  - onPointerEnter
```
kamleshchandnani, line 73: what about mweb?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,199 @@
...
+- Native:
+  - onTouchStart
+  - onTouchEnd
+
+```jsx
+import type { TooltipTriggerProps } from "@razorpay/blade/components";
+
+type MyCustomButtonProps = {} & TooltipTriggerProps
+const MyCustomButton = React.forwardRef(...);
```
kamleshchandnani, line 82: when you do forwardref here how/why will tooltip need it? we'll get the ref by accessing ref of a children inside our Tooltip component?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,199 @@
...
+
+<details>
+  <summary>Implementation detail nuances:</summary>
+
+To make FloatingUI work with Blade components seamlessly there are few things we need to modify:
+
+1. Expose interaction props in all the trigger components
+
+- Web:
+  - onBlur
```
kamleshchandnani, line 118: so now all our consumers will have access to all these methods?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,199 @@
...
+</Tooltip>
+```
+
+> Note: If users don't wrap the non-interactive elements the tooltip won't apear
+
+### Custom Triggers
+
+Users will be able to use their own custom components as triggers:
+
+To make custom elements work the components needs to expose:
```
kamleshchandnani, line 64: whelp this sounds scary because so many things people will forget exposing one or the other thing. what if we provide some API like this
```
import type { TooltipTriggerProps } from "@razorpay/blade/components";

type MyCustomButtonProps = {} & TooltipTriggerProps
const MyCustomButton = ({children, ...props}) => <button {...props}>{children}</button>

// custom trigger
<Tooltip content="Amount reversed to customer bank account" shouldWrapChildren>
{(attributes) => <MyCustomButton {...attributes}>Click Me</MyCustomButton>}
</Tooltip>
```

might as well get rid of refs and all?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,199 @@
...
+  - onTouchStart
+  - onTouchEnd
+
+2. Expose the actual DOM node from ref instead of only exposing [certain methods via our useBladeInnerRef](https://github.com/razorpay/blade/blob/69a1bcef2f09ceaf6f910eaaca3076055fb059a2/packages/blade/src/hooks/useBladeInnerRef.web.ts#L26-L27) hook which we used to prevent component styling misuses. We need to expose the actual DOM node because FloatingUI internally does this [isElement() check](https://github.com/floating-ui/floating-ui/blob/b8990250568043b876e1c8fe42358fe337847ede/packages/react/src/hooks/useFloating.ts#L59) on the tooltip trigger element.
+
+</details>
+
+## Motion
+
+TBD
```
kamleshchandnani, line 134: lets add the animation once RK is done

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: But they will have the same API just that popover will behave a bit differently, eg: click to open and no auto dismiss. 

eg: 
```jsx
<Popover content={<p>some jsx</p>}>
  <Button>Trigger</Button>
</Popover>
```

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: `isChildrenNonInteractive` also sounds good, now we are getting into a bit of subjective territory. 

> Yeah so in that tweet, how radix does is, is basically polar opposite of our API.

packages/blade/src/components/Tooltip/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Yes it will flip to the other side based on the available space. 

Not in favour of `position` sounds like our css styled prop. placement is more accurate imo.

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,199 @@
...
+
+https://github.com/razorpay/blade/assets/35374649/612a96b9-34e1-492b-8e81-e04b896d79db
+
+Immediate tooltips makes sense when the tooltip trigger doesn't do any extra action on clicking, for example imagine in that search icon button if an immediate tooltip is used the tooltip will never have a chance to show up since the screen changes and it opens a new dialog/page.
+
+**Auto dismissal:**
+
+Another UX we need to think about is,
+
+In the long press example, once you release the press the tooltip automatically dismisses after 1s.
```
kamleshchandnani, line 185: 1s is too long? 300ms is what I've generally seen as a pattern most systems follow?

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Yes disabled elements can also have hover information associated with them, imagine a button with a tooltip showing some additional context on why the button is disabled.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Yes atleast in components which we are promoting as "triggers"

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: This was added before RK finalized the duration will update this.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: I mean sure your API that you are showing will just work as is: 

Here: 

```jsx
import type { TooltipTriggerProps } from "@razorpay/blade/components";

type MyCustomButtonProps = {} & TooltipTriggerProps
// forward ref is needed
const MyCustomButton = React.forwardRef(({ children, ...spreadingPropsFineToo }, ref) => {
  return (
    <button ref={ref} {...spreadingPropsFineToo}>
      {children}
    </button>
  );
});

// custom trigger
<Tooltip content="Amount reversed to customer bank account">
  <MyCustomButton>Click Me</MyCustomButton> // no need to manually spread because we use React.cloneElement
</Tooltip>
```

packages/blade/src/components/Tooltip/_decisions/decisions.md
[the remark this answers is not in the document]
saurabhdaware: +1 to not calling it `position`. Since we do actually support `position` in styled-props and it has values of `relative`, `absolute`

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,199 @@
+# Tooltip API Decisions <!-- omit in toc -->
+
+The tooltip typically provides additional context about the element or its function. A tooltip is always triggered by a `mouse hover` on desktop and `on tap` on mobile.
+
+<img src="./tooltip-thumbnail.png" width="380" alt="" />
```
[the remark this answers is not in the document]
anuraghazra: Yeah thumbnail isn't added.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: > we'll get the ref by accessing ref of a children inside our Tooltip component?

Didn't get this part but, ref will need to be exposed of the trigger. 

Here's a full example: 

```jsx
const CustomTrigger = React.forwardRef<
  HTMLDivElement,
  { children: React.ReactNode } & TooltipTriggerProps
>(({ children, ...props }, ref) => {
  return (
    <Box
      ref={ref}
      tabIndex={0}
      {...props}
    >
       {children}
    </Box>
  );
});

<TooltipComponent placement="bottom" content="A custom trigger">
  <CustomTrigger>Hover over me</CustomTrigger>
</TooltipComponent>
```

> when you do forwardref here how/why will tooltip need it? 

FloatingUI needs it to do placement calculation.

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,198 @@
...
+
+</details>
+
+## Motion
+
+TBD
+
+## Accessibility
+
+- Tooltip will have [`role=tooltip`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role)
```
kamleshchandnani, line 137: we don't need `aria-labelledby` and `aria-describedby` attrs? because on icons you might need those things for screen readers?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,198 @@
...
+| ------------------------------- | ------------------------------------------------------------------------ | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| content                         | `string`                                                                 | `undefined` | Content of the tooltip                                                                                                                                   | ✅       |
+| children                        | `React.ReactNode`                                                        | `undefined` | Trigger component for tooltip, Accepts any interactive element or icons                                                                                  | ✅       |
+| placement                       | `top, top-start, top-end, left, right, bottom, bottom-start, bottom-end` | `top`       | Placement of tooltip, the tooltip avoid collision with the edge of the screen and flip to the oppisite side even when placement is set.                  |          |
```
kamleshchandnani, line 31: okay there was a comment around  proposal to name this prop as `position`?

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Added.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Yes added.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Going ahead with placement for these reasons https://github.com/razorpay/blade/pull/1216#discussion_r1221461811

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Yes, we are already passing that via React.cloneElement, users will need to spread the props tho.

packages/blade/src/components/Tooltip/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: yes please add an image before we make this public?

packages/blade/src/components/Tooltip/_decisions/decisions.md
kamleshchandnani: can't find it? 🤔

packages/blade/src/components/Tooltip/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Thumbnail design me nehi hain.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: pushed check again? 
image

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Design
+
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=40540-559304&t=tmTrf3xJU6oj59fM-0) to all variants of the Tooltip component
+
+## `Tooltip` API
+
+| Prop         | Type                                                                     | Default     | Description                                                                                                                             | Required |
+| ------------ | ------------------------------------------------------------------------ | ----------- | --------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| content      | `string`                                                                 | `undefined` | Content of the tooltip                                                                                                                  | ✅       |
```
cseas, line 28: We have a use-case for having a link inside Tooltip. Please check the coin at the top of this page: https://razorpay.com/payment-pages/

cseas, line 28: Is it possible to add styling to the text inside tooltip? For example, bold, italics?

cseas, line 28: Can we have a newline inside the tooltip content?

akkhil7: +1, we've had similar use-cases where content had to be styled. What's the challenge of having JSX instead of string?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,221 @@
...
+| children     | `React.ReactNode`                                                        | `undefined` | Trigger component for tooltip, Accepts any interactive element or icons                                                                 | ✅       |
+| placement    | `top, top-start, top-end, left, right, bottom, bottom-start, bottom-end` | `top`       | Placement of tooltip, the tooltip avoid collision with the edge of the screen and flip to the oppisite side even when placement is set. |          |
+| onOpenChange | `({ isOpen }) => void`                                                   | `undefined` | Called when tooltip isOpen state is changed, this can be used to detect when tooltip opens or closed                                    |          |
+
```
akkhil7, line 32: Can we have the tooltip open by default? For eg, there are Tour components which are roughly the same but with a different logic in opening and closing of it. They're ordered and the first one in the order will be open by default. Thoughts on this?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,221 @@
...
+| content      | `string`                                                                 | `undefined` | Content of the tooltip                                                                                                                  | ✅       |
+| children     | `React.ReactNode`                                                        | `undefined` | Trigger component for tooltip, Accepts any interactive element or icons                                                                 | ✅       |
+| placement    | `top, top-start, top-end, left, right, bottom, bottom-start, bottom-end` | `top`       | Placement of tooltip, the tooltip avoid collision with the edge of the screen and flip to the oppisite side even when placement is set. |          |
+| onOpenChange | `({ isOpen }) => void`                                                   | `undefined` | Called when tooltip isOpen state is changed, this can be used to detect when tooltip opens or closed                                    |          |
```
nickisaacs, line 31: Does `onOpen` and `onClose` sound better?

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,221 @@
...
+## Library
+
+We will be using [FloatingUI](https://floating-ui.com/) to position the tooltip & handle the basic tooltip logic. Floating UI supports both Web & ReactNative.
+
+### Bundle Size
+
+With all the neccessary middlewares and features included for our usecase FloatingUI will be around **10kb** gzip.
+
+Check the [BundleJS Analysis](https://bundlejs.com/?q=%40floating-ui%2Freact&treeshake=%5B%7B%0A++arrow%2Cflip%2CFloatingArrow%2Coffset%2CuseFloating%2CuseFocus%2CuseHover%2CuseInteractions%2CuseTransitionStyles%2C%7D%5D&config=%7B%22esbuild%22%3A%7B%22external%22%3A%5B%22react%22%2C%22react-dom%22%5D%7D%7D)
+
```
iAmServerless, line 118: is this SEO friendly?

packages/blade/src/components/Tooltip/_decisions/decisions.md
Kushagra-ag: @anuraghazra so we will be to use props like onBlur, onFocus, etc. on all blade components?

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: @cseas Tooltip should only contain non-interactive elements inside of it for accessibility reasons. 

If you want to add interactive element inside like a button or link, a Popover or Toggletip will be a valid pattern where users will click the trigger and the Popover won't dismiss automatically or just by hovering out of the trigger. 

image

https://web.dev/building-a-tooltip-component/#accessibility

anuraghazra: No, tooltips will only contain short supplement text without newlines. if the content is longer that it needs newlines it should not be inside tooltip. 

Here are the content [guidelines](https://www.figma.com/file/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?type=design&node-id=175-812084&t=tdn24pTBE7PD3OY2-0) for tooltip: 

image

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Didn't understood 🤔 SEO as in? Do you mean tooltip's content should remain on DOM even if it's hidden?

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Yeah discussed this earlier also. 
**[Check the discussion here](https://github.com/razorpay/blade/pull/1216#discussion_r1212981863)**

We will go with 1 handler only since we are doing the same in few other components as well, Accordion, Collapse etc.

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: Yes, we will release Popover as a separate component.

cseas: Hoping the designers agree on this. Marking as resolved.

anuraghazra: @cseas @akkhil7 discussed this with design. 

We will only allow plain strings for now. 

We will be standardising the design to not have any bold/italic text. We've done design assessment of the Tooltip content & blade advocates are in-lined with just plain text on the tooltip. 

> +1, we've had similar use-cases where content had to be styled.

@akkhil7 these existing cases will also get migrated (first from the design side) when adopting blade, 
Having said that if there could be genuine usecases for bold/italic texts we can accommodate that later.

akkhil7 (approved):

packages/blade/src/components/Tooltip/_decisions/decisions.md
nickisaacs: i dont think this stops anyone from writing inline handlers here as well, but if we are following the same in other components its fine

nickisaacs (dismissed):

packages/blade/src/components/Tooltip/_decisions/decisions.md
anuraghazra: @Kushagra-ag not all blade components will expose the TooltipTrigger props. 

For example, it doesn't make sense to have the whole Checkbox be used as a trigger instead we should use the tooltip on InfoIcon. 

We will expose TooltipTriggers on: 
- Box
- Button
- Link
- IconButton

packages/blade/src/components/Tooltip/_decisions/decisions.md
```diff
@@ -0,0 +1,221 @@
...
+### Working with non-interactive triggers
+
+To make tooltip apear on hovering over non-interactive elements such as icons, badges, counters etc. We will provide a `TooltipInteractiveWrapper` component which will work as a minimal trigger:
+
+```js
+import { Tooltip, TooltipInteractiveWrapper } from '@razorpay/blade';
+
+// non-interactive element as trigger
+<Tooltip content="Amount reversed to customer bank account">
+  <TooltipInteractiveWrapper>
```
Rozamo, line 53: This might be a noob question but can we make it so that the `TooltipInteractiveWrapper`'s functionality is automatically added if the `Tooltip`'s children are not interactive?

anshulsahni (dismissed):

packages/blade/src/components/Tooltip/_decisions/decisions.md
prabhasjoshi: What if we only want to show a tooltip only when the button is disabled 
We can pass a boolean isDisabled to the button - but we will always have to wrap the button in a tooltip ...
will we have to pass either content as blank or non-blank to achieve this ?

kamleshchandnani (approved):

PR #1218, opened 2023-05-29
docs(Collapsible): add API design

divyanshu013:
View [formatted markdown](https://github.com/razorpay/blade/blob/div/collapsible-api/packages/blade/src/components/Collapsible/_decisions/decisions.md).

Also update related prop on Accordion spec to make the API consistent across.


packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+- `type` - will be `button` always
+
+### CollapsibleLink
+
+Built-in trigger for `Collapsible`, provides a subset of `Link` [properties](../../Link/_decisions/decisions.md).
+
+Supports ✅
+
+- `children`
+- `icon`
```
chaitanyadeorukhkar, line 100: Consumers will have to handle changing of icon here, right? So no animation of the icon right?

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+
+<div align="center">
+  <img src="./collapsible-thumbnail.png" width="380" alt="" />
+</div>
+
+## Design
+
+- [Figma - Collapsible](https://www.figma.com/file/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?type=design&node-id=79-629874&t=sVxH3DOnx3L3F9rO-0)
+
+## Interactions
```
chaitanyadeorukhkar, line 15: Can we add animations that show a collapsible button/link instead of accordion? Seems a bit confusing

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+- Some related properties from web will be applicable to native as well ([a11y guide](https://reactnative.dev/docs/accessibility))
+
+## Usage
+
+### Controlled state
+
+```tsx
+const App = () => {
+  const [isExpanded, setIsExpanded] = useState(false);
+  const onChange = ({ isExpanded }) => setIndex(isExpanded);
```
chaitanyadeorukhkar, line 146:
```suggestion
  const onChange = ({ isExpanded }) => setIsExpanded(isExpanded);
```

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+- `children`
+- `icon`
+- `isDisabled`
+- `size`
+- `accessibilityLabel`
+- `testID`
+
+Doesn't support ❌
+
+- `iconPosition` - will be `left` for passed `icon` since the `CollapsibleLink` renders a rotating chevron icon to its right
```
chaitanyadeorukhkar, line 108: So Link will have chevron icon by default but Button won't? We'll handle the icon to be up or down on Link but not on Button? This seems very confusing for the end user. Can we keep the behaviour consistent? Lets give icon control to the consumer entirely?

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+- `htmlTitle`
+- `variant` - will always be `button`
+- `onClick`
+- `href`, `target`, `rel` - will always be a button
```
chaitanyadeorukhkar, line 112: Then its not really a Link, right? Should we call it `CollapsibleLinkButton`?

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+### Controlled state
+
+```tsx
+const App = () => {
+  const [isExpanded, setIsExpanded] = useState(false);
+  const onChange = ({ isExpanded }) => setIndex(isExpanded);
+
+  return (
+    <Collapsible isExpanded={isExpanded} onChange={onChange}>
+      <CollapsibleLink>Answer to life, universe and everything</CollapsibleLink>
```
chaitanyadeorukhkar, line 150: Don't Panic! 🫢

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: Actually there is no figma prototype for this, we're reusing the same guidelines from accordion 😅

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: The caveat for giving control to users would be inability to animate that, the rotating chevron icon is there only for link-button but not for button.

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: Brainstorming on this, we've a similar usecase in `DropdownLink` / `DropdownLinkButton`, will prefer a similar terminology

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: I think for the link variant we can provide the rotating chevron (only let users add a left icon if they want). Need to reconfirm with RK once he's back but from the audits I couldn't find cases where a chevron icon won't work.

packages/blade/src/components/Collapsible/_decisions/decisions.md
chaitanyadeorukhkar: In that case should we restrict users from passing icons entirely? We'll handle the icon and its animation then for both Button & Link variants?

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: Going ahead with `CollapsibleLink` (like `DropdownLink`)

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: Assessing with design, will finalize based on design research / audits

chaitanyadeorukhkar: Cool. Rest of the API LGTM

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+>
+> - `Collapsible` will compose (`CollapsibleButton` or `CollapsibleLink`), `CollapsibleBody`. This ensures all interactions, a11y and animations work seamlessly at user end.
+> - `CollapsibleButton` and `CollapsibleLink` will support a subset of relevant props from `Button` and `Link` respectively.
+
+### Collapsible
+
+| Prop              | Type            | Default     | Description                                                        | Required |
+| ----------------- | --------------- | ----------- | ------------------------------------------------------------------ | -------- |
+| children          | `JSX`           | `undefined` | Composes `CollapsibleButton`, `CollapsibleLink`, `CollapsibleBody` | ✅       |
+| direction         | `bottom`, `top` | `bottom`    | Direction in which the content expands                             |          |
```
anuraghazra, line 61: How will this look in action? Do we have motion or designs for this? 

Like: 

content
button

When clicking the button the content height will animation from bottom to top?

divyanshu013: Yes, the content will expand which means the trigger will shift down (there are some usecases in the design)

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: Discussed with design, here's the outcome on dev API:

- We’ll let users pass all the props (including icons) (on Figma side there isn't a way to restrict this)
- We’ll not have right icon in link button trigger (because chevron takes that place)
- The chevron icon will animate (custom icons will not)
- No provision to disable rotating chevron icon on link button (note: button has no rotating chevron)

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+- `testID`
+
+Doesn't support ❌
+
+- `icon` - renders a chevron icon always
+- `iconPosition` - will be `right` always for the chevron icon (which animates)
+- `htmlTitle`
+- `variant` - will always be `button`
+- `onClick`
+- `href`, `target`, `rel` - will always be a button
```
anuraghazra, line 112: We will still render a link here? or will it look like a link and work like a button?

anuraghazra (approved):

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+
+- [Figma - Collapsible](https://www.figma.com/file/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?type=design&node-id=79-629874&t=sVxH3DOnx3L3F9rO-0)
+
+## Interactions
+
+Collapsible shares the opening / closing interaction with Accordion:
+
+### Accordion open
+
+https://github.com/razorpay/blade/assets/6682655/1a02d680-1832-4a70-a0a3-1a1f56f2b8b9
```
kamleshchandnani, line 21: lets put a motion of collapsible specifically?

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+- `iconPosition`
+- `isDisabled`
+- `testID`
+- `accessibilityLabel`
+
+Doesn't support ❌
+
+- `isFullWidth` - can visually break UI
+- `onClick`, `isLoading` - using `onClick` will break built-in interactions (unless we provide additional hooks), `onChange` should already provide a way for users to listen-in on `Collapsible` state. This might have usage for async with `isLoading` in the future but for now not exposing the prop
+- `type` - will be `button` always
```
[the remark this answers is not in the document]
divyanshu013: We anyway need a `CollapsibleLink` sub-component so it can automatically handle the toggling wiring logic. Might be good to restrict these props over here.
Basically `type` will not be available at all on `CollapsibleLink` because that would always be `button` so we can just avoid exposing the prop altogether.

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+For SEO purposes the collapsed content can be hidden with CSS.
+
+Behavior wise Collapsible's interaction is similar to Switch (toggles between binary states) _with slight differences_
+
+Collapsible shares the toggle interaction with Accordion. We can create abstractions that work for both plus other components such as tabs, switch, etc. (different from this public API spec):
+
+- Build collapsible interaction on top of a base collapsible component that can be reused for accordion
+- A disclosure based hook (references - [Chakra](https://chakra-ui.com/docs/hooks/use-disclosure), [Mantine](https://mantine.dev/hooks/use-disclosure/), [Paste](https://paste.twilio.design/components/disclosure))
+
+Find in page (automatically expanding the collapsible if someone does <kbd>Cmd</kbd> + <kbd>F</kbd>) is a tricky feat, need to check feasibility and should be taken up as an enhancement (references - [blog by Anurag](https://dev.to/anuraghazra/thoughts-on-find-in-page-with-custom-accordion-elements-5573), [`hidden="until-found"`](https://github.com/WICG/display-locking/blob/main/explainers/hidden-content-explainer.md)).
```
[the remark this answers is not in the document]
divyanshu013: Right, complexity seems high. Might be good idea to revisit in the future. For SEO purposes, the DOM content will be present (though hidden)

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Collapsible
+
+| Prop              | Type            | Default     | Description                                                        | Required |
+| ----------------- | --------------- | ----------- | ------------------------------------------------------------------ | -------- |
+| children          | `JSX`           | `undefined` | Composes `CollapsibleButton`, `CollapsibleLink`, `CollapsibleBody` | ✅       |
+| direction         | `bottom`, `top` | `bottom`    | Direction in which the content expands                             |          |
+| defaultIsExpanded | `boolean`       | `false`     | Expands the collapsible content by default (uncontrolled)          |          |
+| isExpanded        | `boolean`       | `undefined` | Expands the collapsible content (controlled)                       |          |
+| onChange          | `function`      | `undefined` | Callback for change in collapsible's expanded state                |          |
```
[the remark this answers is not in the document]
divyanshu013: Should we also change the terminology from `*expanded*` to `*open*` across other props? Also we're using the similar terminology in accordion.

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: Don't have it on Figma, will put a screen recording.

packages/blade/src/components/Collapsible/_decisions/decisions.md
kamleshchandnani: Can you ask RK to create one because its okay to have redundant but contextual animations esp to avoid confusion

packages/blade/src/components/Collapsible/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: not expanded. this is for the event handler.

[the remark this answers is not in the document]
kamleshchandnani: or we can also make it contextual. if we have `expanded` as a prop we can have `onExpandChange`?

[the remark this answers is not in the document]
divyanshu013: `onExpandChange` sounds good, will update

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Collapsible
+
+| Prop              | Type            | Default     | Description                                                        | Required |
+| ----------------- | --------------- | ----------- | ------------------------------------------------------------------ | -------- |
+| children          | `JSX`           | `undefined` | Composes `CollapsibleButton`, `CollapsibleLink`, `CollapsibleBody` | ✅       |
+| direction         | `bottom`, `top` | `bottom`    | Direction in which the content expands                             |          |
+| defaultIsExpanded | `boolean`       | `false`     | Expands the collapsible content by default (uncontrolled)          |          |
+| isExpanded        | `boolean`       | `undefined` | Expands the collapsible content (controlled)                       |          |
+| onExpandChange    | `function`      | `undefined` | Callback for change in collapsible's expanded state                |          |
```
saurabhdaware, line 64: should this be `defaultIsOpen`, `isOpen`, `onOpenChange` to [match it with Tooltip](https://github.com/razorpay/blade/pull/1216#discussion_r1220950828)?

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+>
+> - `Collapsible` will compose (`CollapsibleButton` or `CollapsibleLink`), `CollapsibleBody`. This ensures all interactions, a11y and animations work seamlessly at user end.
+> - `CollapsibleButton` and `CollapsibleLink` will support a subset of relevant props from `Button` and `Link` respectively.
+
+### Collapsible
+
+| Prop              | Type            | Default     | Description                                                        | Required |
+| ----------------- | --------------- | ----------- | ------------------------------------------------------------------ | -------- |
+| children          | `JSX`           | `undefined` | Composes `CollapsibleButton`, `CollapsibleLink`, `CollapsibleBody` | ✅       |
+| direction         | `bottom`, `top` | `bottom`    | Direction in which the content expands                             |          |
```
saurabhdaware, line 61: Should we call this `placement` to match it with Tooltip?

saurabhdaware (commented):
Looks good to me mostly. Have 2 doubts regarding the consistency of API with Tooltip

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -50,16 +50,16 @@ import { Accordion, AccordionItem } from '@razorpay/blade';
 | -------------------- | ---------- | ----------- | -------------------------------------------------------------- | -------- |
 | defaultExpandedIndex | `number`   | `undefined` | Makes the passed item index expanded by default (uncontrolled) |          |
 | expandedIndex        | `number`   | `undefined` | Expands the passed index (controlled)                          |          |
-| onChange             | `function` | `undefined` | Callback for change in any item's expanded state               |          |
+| onExpandChange       | `function` | `undefined` | Callback for change in any item's expanded state               |          |
```
saurabhdaware, line 53: Should we call this onOpenChange as well? refer to discussion above for the same on Collapsible

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: I feel this is a different kind of interaction and also differs from the behavior. Eg. the content still stays in the DOM (for SEO). Also `expanded` terminology is similar to the `aria-expanded` property which is also for similar disclosures types of interactions.

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: I think `direction` fits better with the interaction (which direction does the content expand in vs which placement does the content expand in) 😅

packages/blade/src/components/Accordion/_decisions/decisions.md
divyanshu013: Related discussion [here](https://github.com/razorpay/blade/pull/1218#discussion_r1226101854). Overall, this is to make it more contextual to collapsible kind of interactions.

cseas (approved):

packages/blade/src/components/Collapsible/_decisions/decisions.md
saurabhdaware: That is true. I think `placement` makes sense on Tooltip because it is not with respect to trigger. Trigger isn't part of the component. So it "places" itself. Whereas here `direction` makes sense because we're talking about it in the context of trigger 😅 

Like if we were to put this prop on `CollapsibleBody` then `placement` would've made sense because you're "placing" it wrt to the trigger and main component.

I think we can go with direction in that case.

packages/blade/src/components/Collapsible/_decisions/decisions.md
saurabhdaware: > Eg. the content still stays in the DOM (for SEO)

This is part of internal implementation so I think we can take call irrespective of that. Don't think the consumers would even know about this. We do usually tradeoff "most accurate term" for "consistent term in other components".

> Also expanded terminology is similar to the aria-expanded property which is also for similar disclosures types of interactions.

This is also a good point. "Collapsible is expanded / collapsed" does sound more natural than "Collapsible is open / close" 🤔 

Can take more inputs maybe? what do you think @kamleshchandnani @anuraghazra?

AnshulNautiyal:
@divyanshu013  

1. Do we support nested collapsible? 
2. Is collapsible content (which is hidden initially ) have it own space like border, shadow.. As per recording, it is look like attached to the base layout? 
3. In case of Modal, will it open on top of modal since  collapsible content is appear on top of it? (this answer may depend on 2nd question)
4. It support children props whose type is JSX..will it support any blade/custom made component in it? will it handle lazy load of components in itself or developer will take care of it?

packages/blade/src/components/Collapsible/_decisions/decisions.md
saurabhdaware: Oh just noticed we already discussed it here - https://github.com/razorpay/blade/pull/1218#discussion_r1224073355. Can go with `onExpandChange` then

packages/blade/src/components/Collapsible/_decisions/decisions.md
```diff
@@ -0,0 +1,186 @@
...
+        <Text>42</Text>
+      </CollapsibleBody>
+    </Collapsible>
+  );
+};
+```
+
+## Alternatives
+
+- Considered a more flexible API of using anything as a collapsible trigger instead of restricted wrappers over `Button` and `Link`. However, this can lead to inconsistencies and a11y issues _(because users will then have to handle all the a11y props and focus management)_.
```
anshulsahni, line 161: Sorry if I am not understanding it correctly, but by this do you mean that you don't encourage having custom triggers for Collapsible but the API gives button and link separately, which means we can easily skip having the button or link triggers and can make a fully controlled Collapsible component just using CollapsbileBody

anshulsahni (commented):
Added just one tiny comment

Kushagra-ag (approved):

packages/blade/src/components/Collapsible/_decisions/decisions.md
kamleshchandnani: also if you use CollapsibleBody without Collapsible it'll likely throw error, right @divyanshu013 ?

kamleshchandnani (approved):

saurabhdaware (approved):

packages/blade/src/components/Collapsible/_decisions/decisions.md
divyanshu013: Right, internally it uses a Collapsible context and would throw

anshulsahni (approved):

PR #1223, opened 2023-05-30
fix(SelectInput): single select value clear

### LINKED ISSUE #1211: Clearing selected input of controlled dropdown
Currently in the [controlled dropdown](https://blade.razorpay.com/?path=/story/components-dropdown-with-select--with-controlled-select&globals=measureEnabled:false) - the api allows us to pass the value to SelectInput.

This lets us select a specific value for example "Bangalore"
image
image

The problem is that we want the control to be able to clear away the selected value when the value to SelectInput is a falsey value.
as shown in this codesandbox - https://codesandbox.io/s/blade-controlled-select-forked-kc86ef?file=/App.tsx 
Select Bangalore > Bangalore gets selected.
Select null > Bangalore doesn't get cleared.

Our use case is the clearing of filters from a separate button.
https://drive.google.com/file/d/1xnotbTooBE1l0vD5L9-J-13OHZHEt7RG/view?usp=sharing 

The lack of this - leaves a part which is still uncontrolled. Thoughts ?

saurabhdaware:
resolves #1211 

Had a falsy value check. But we should allow empty string to be passed down so we can reset the value.

### DIFF
--- a/.changeset/six-points-shake.md
+++ b/.changeset/six-points-shake.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix(SelectInput): single select value clear
--- a/.changeset/six-points-shake.md
+++ b/.changeset/six-points-shake.md
@@ -1,5 +1,7 @@
 ---
-"@razorpay/blade": patch
+'@razorpay/blade': patch
 ---
 
 fix(SelectInput): single select value clear
+
+- You can pass `''` (empty string) in single select `value` prop to clear the selected value now.
--- a/packages/blade/src/components/Dropdown/docs/DropdownWithSelect.stories.tsx
+++ b/packages/blade/src/components/Dropdown/docs/DropdownWithSelect.stories.tsx
@@ -174,6 +174,60 @@ InternalControlledSelect.parameters = {
   },
 };
 
+// For chromatic and internal react native testing
+export const InternalControlledSingleSelect = (): JSX.Element => {
+  const [currentSelection, setCurrentSelection] = React.useState<string>('');
+
+  return (
+    <>
+      <Button
+        marginBottom="spacing.4"
+        onClick={() => {
+          setCurrentSelection('bangalore');
+        }}
+      >
+        Select Bangalore
+      </Button>
+      <Button
+        variant="secondary"
+        marginBottom="spacing.4"
+        marginLeft="spacing.4"
+        onClick={() => {
+          setCurrentSelection('');
+        }}
+      >
+        Clear Selection
+      </Button>
+      <Dropdown>
+        <SelectInput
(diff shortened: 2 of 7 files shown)


packages/blade/src/components/Dropdown/docs/DropdownWithSelect.stories.tsx
```diff
@@ -174,6 +174,60 @@ InternalControlledSelect.parameters = {
...
+        }}
+      >
+        Select Bangalore
+      </Button>
+      <Button
+        variant="secondary"
+        marginBottom="spacing.4"
+        marginLeft="spacing.4"
+        onClick={() => {
+          setCurrentSelection('');
```
divyanshu013, line 196: Have we documented this behavior anywhere? That you need to pass `''` to clear dropdown or reset selection. One might think to pass `null`, `undefined`, etc. for clearing the input

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -107,7 +107,9 @@ const _SelectInput = (
 
   const selectValues = (valuesToSelect: string | string[]): void => {
     if (options.length > 0) {
-      if (typeof valuesToSelect === 'string') {
+      if (isEmpty(valuesToSelect)) {
```
divyanshu013, line 110: A minor helpful comment that we treat empty string `''` as a way to clear input.

.changeset/six-points-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix(SelectInput): single select value clear
```
divyanshu013, line 5: Let's also add a line on what works now for clearing the input value (eg. pass empty string)

packages/blade/src/components/Dropdown/docs/DropdownWithSelect.stories.tsx
saurabhdaware: I have added "Clear Selection" button in Controlled Select Input story

.changeset/six-points-shake.md
saurabhdaware: done

packages/blade/src/components/Dropdown/docs/DropdownWithSelect.stories.tsx
saurabhdaware: https://61c19ee8d3d282003ac1d81c-splkqkkafz.chromatic.com/?path=/story/components-dropdown-with-select--with-controlled-select&globals=measureEnabled:false

divyanshu013 (approved):
Looks good to me.

PR #1226, opened 2023-05-30
fix: rebuild pods to resolve react native setup errors

saurabhdaware:
I followed pods and node reset https://reactnative.dev/docs/new-architecture-troubleshooting#cocoapods-and-node-reset

This fixed my react native setup. Earlier I thought I won't have to commit this but I have to do this again and again if I don't commit this. 

Let me know if this makes sense cause I have absolutely no clue what I am doing.


divyanshu013 (commented):
I think should be fine to commit these. Need expert opinion from @chaitanyadeorukhkar

saurabhdaware:
ok nope my setup breaks even when I have this in repo. Not sure what is wrong 😭

PR #1228, opened 2023-05-30
feat: add Modal component

chaitanyadeorukhkar:
- Adds Modal component
- Adds Tests
- Adds Documentation


packages/blade/src/components/Box/BaseBox/types/spacingTypes.ts
```diff
@@ -1,14 +1,13 @@
...
-import type { Platform } from '~utils';
 
 type ArrayOfMaxLength4<T> = readonly [T?, T?, T?, T?];
-type SpaceUnits = Platform.Select<{
-  web: 'px' | '%' | 'fr' | 'rem' | 'em' | 'vh' | 'vw';
-  native: 'px' | '%';
-}>;
-type SpacingValueType = DotNotationSpacingStringToken | `${string}${SpaceUnits}` | 'auto';
-
+type SpaceUnits = 'px' | '%' | 'fr' | 'rem' | 'em' | 'vh' | 'vw';
```
chaitanyadeorukhkar, line 5: Platform.Select is not working as we expect here. I think someone had also raised an issue about this earlier on slack. For now I've made all these values available on both platforms. If someone uses an unsupported value on RN, RN will throw big red errors

packages/blade/src/components/Box/BaseBox/types/spacingTypes.ts
```diff
@@ -1,14 +1,13 @@
...
+type SpacingValueType =
+  | DotNotationSpacingStringToken
+  | `${string}${SpaceUnits}`
+  | `calc(${string})`
```
chaitanyadeorukhkar, line 9: Added `calc` support

packages/blade/src/components/Modal/Modal.native.tsx
chaitanyadeorukhkar, line 1: Need to add a dummy export here

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -0,0 +1,209 @@
...
+const Modal = ({
+  isOpen = false,
+  children,
+  onDismiss,
+  initialFocusRef,
+  size = 'small',
+  accessibilityLabel,
+}: ModalProps): React.ReactElement => {
+  const { theme } = useTheme();
+  const [footerHeight, setFooterHeight] = useState(0);
```
chaitanyadeorukhkar, line 91: Need footer height to calculate scroll overlay offset. Added a comment about it in footer

packages/blade/src/components/Modal/ModalBody.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { useModalContext } from './ModalContext';
+import BaseBox from '~components/Box/BaseBox';
+import { assignWithoutSideEffects } from '~src/utils/assignWithoutSideEffects';
+import { MetaConstants, makeMotionTime, makeSize, metaAttribute } from '~utils';
+
+type ModalBodyProps = {
+  children: React.ReactNode;
+};
+
+const OverflowOverlay = styled(BaseBox)<{
```
chaitanyadeorukhkar, line 14: Ignore this for now

packages/blade/src/components/Box/BaseBox/types/spacingTypes.ts
divyanshu013: I was thinking perhaps we can just make this `string`, don't know if the typing is adding much value

packages/blade/src/components/Modal/Modal.native.tsx
```diff
@@ -0,0 +1,24 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  initialFocusRef?: React.MutableRefObject<any>;
+  size?: 'small' | 'medium' | 'large';
+  accessibilityLabel?: string;
+};
+
+const Modal = (): React.ReactElement => {
+  return (
+    <Text>
+      Modal Component is not available for Native mobile apps and we should use BottomSheet
```
divyanshu013, line 17: Should we throw an error instead?

packages/blade/src/tokens/global/border.ts
```diff
@@ -7,6 +7,8 @@ type BorderRadius = Readonly<{
   medium: 4;
   /** large: 8(px/rem/pt) */
   large: 8;
+  /** xlarge: 12(px/rem/pt) */
+  xlarge: 12;
```
divyanshu013, line 11: Assuming this token is also now added on figma?

packages/blade/src/components/Modal/index.tsx
```diff
@@ -0,0 +1 @@
+export { Modal } from './Modal';
```
divyanshu013: Todo: also export other sub components

packages/blade/src/components/Modal/ModalPortal.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const createModalRootWrapper = (id: string): HTMLElement => {
+  const element = document.createElement('div');
+  element.setAttribute('id', id);
+  document.body.appendChild(element);
+  return element;
+};
+
+const ModalPortal = ({ children }: ModalPortalProps): React.ReactElement => {
+  const wrapperID = 'modal-root-portal';
```
divyanshu013, line 15: Making it more specific so it doesn't conflict with any other random modal library, *just in case*
```suggestion
  const wrapperID = 'blade-modal-root-portal';
```

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -0,0 +1,238 @@
...
+      defaultInitialFocusRef,
+      isVisible,
+      footerHeight,
+      setFooterHeight,
+    }),
+    [isOpen, onDismiss, defaultInitialFocusRef, isVisible, footerHeight, setFooterHeight],
+  );
+  const handleKeyDown = (event: React.KeyboardEvent): void => {
+    // close modal on escape key press
+    if (event?.key === 'Escape' || event?.code === 'Escape') {
```
divyanshu013, line 176: We need to check both on `key` and `code`?

packages/blade/src/components/Modal/ModalBody.web.tsx
```diff
@@ -0,0 +1,107 @@
...
+const OverflowOverlay = styled(BaseBox)<{
+  scrollbarWidth: number;
+  footerHeight: number;
+  showOverlay: boolean;
+}>`
+  position: absolute;
+  left: 0%;
+  right: ${({ scrollbarWidth }) => `${scrollbarWidth}px`};
+  bottom: ${({ footerHeight }) => `${footerHeight}px`};
+  background: linear-gradient(
```
divyanshu013, line 23: Can you also add a comment that there are no tokens for the colors in this?

packages/blade/src/components/Modal/ModalBody.web.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const [scrollbarWidth, setScrollbarWidth] = useState(0);
+  const [scrollEnd, setScrollEnd] = useState(false);
+
+  const checkForScrollbar = useCallback(() => {
+    // Checks if the content has a scrollbar
+    // and sets the width of the scrollbar
+    const element = contentRef.current;
+    if (element) {
+      const hasScrollbar = element.scrollHeight > element.clientHeight;
+      setScrollbarWidth(element.offsetWidth - element.clientWidth);
```
divyanshu013, line 52: Not sure if this works across platforms and OSs, have you confirmed?

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -106,13 +106,22 @@ const Modal = ({
...
 }: ModalProps): React.ReactElement => {
-  const { theme } = useTheme();
+  const { theme, platform } = useTheme();
   const [footerHeight, setFooterHeight] = useState(0);
   const { isMounted, isVisible } = usePresence(isOpen, {
     transitionDuration: theme.motion.duration.xmoderate,
     initialEnter: true,
   });
 
+  // Warn consumer if modal is opened on mobile
```
divyanshu013, line 116: Minor: probably an overkill 😅

packages/blade/src/components/Modal/Modal.native.tsx
chaitanyadeorukhkar: We had a discussion on this with Abinash, Kamlesh & Anurag where we concluded on throwing a warning since an error could be disruptive to end user in case a developer forgets to check their flow on mWeb and ships to prod

packages/blade/src/tokens/global/border.ts
chaitanyadeorukhkar: There is an ongoing discussion. Will confirm this post that

packages/blade/src/components/Modal/index.tsx
chaitanyadeorukhkar: This is done

packages/blade/src/components/Modal/Modal.web.tsx
chaitanyadeorukhkar: Better to check both. During OTPInput implementation I had read that some older browser versions send either of these.

packages/blade/src/components/Modal/ModalBody.web.tsx
chaitanyadeorukhkar: We will need to tokenize this to ensure it works with multiple themes. This is pending discussion which we will close today.

packages/blade/src/components/Modal/ModalBody.web.tsx
chaitanyadeorukhkar: This doesn't work on macOS when scroll bar is on "auto" mode system-wide. Need to discuss with design if we want to spend more time figuring this out or be okay with the overlap

packages/blade/src/components/BaseHeaderFooter/BaseFooter.tsx
```diff
@@ -1,18 +1,24 @@
...
   children: React.ReactNode;
+  metaDataComponentName?: string;
   showDivider?: boolean;
 };
 
-const _BaseFooter = ({ children, showDivider = true }: BaseFooterProps): React.ReactElement => {
+const _BaseFooter = ({
+  children,
+  showDivider = true,
+  metaDataComponentName,
```
anuraghazra, line 15: nit: or just componentName

```suggestion
  metaComponentName,
```

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+  const focusOnInitialRef = React.useCallback(() => {
+    if (!initialFocusRef) {
+      // focus on close button
+      defaultInitialFocusRef.current?.focus();
+    } else {
+      // focus on the initialRef passed by the user
+      initialFocusRef.current?.focus();
+    }
+  }, [initialFocusRef]);
```
anuraghazra, line 151: All these logic probably can be removed. 

FloatingFocusManager supports focus return. 

```
<FloatingFocusManager
  initialFocus={initialFocusRef || defaultInitialFocusRef}
  returnFocus
  context={context}
  modal={true}
>
```

packages/blade/src/components/Modal/ModalBody.web.tsx
```diff
@@ -0,0 +1,107 @@
...
+type ModalBodyProps = {
+  children: React.ReactNode;
+};
+
+const OverflowOverlay = styled(BaseBox)<{
+  scrollbarWidth: number;
+  footerHeight: number;
+  showOverlay: boolean;
+}>`
+  position: absolute;
```
anuraghazra, line 19: Any particular reason to use string instead of object?

packages/blade/src/components/Modal/Modal.native.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { ModalHeader } from './ModalHeader';
+import type { ModalHeaderProps } from './ModalHeader';
+import { ModalFooter } from './ModalFooter';
+import type { ModalFooterProps } from './ModalFooter';
+import { ModalBody } from './ModalBody';
+import type { ModalBodyProps } from './ModalBody';
+import { Text } from '~components/Typography';
+
+type ModalProps = {
```
anuraghazra, line 10: Let's remove or reexport this type from a single place?

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -0,0 +1,245 @@
...
+      );
+    }
+  });
+
+  return (
+    <ModalPortal>
+      <ModalContext.Provider value={modalContext}>
+        {isMounted ? (
+          <FloatingFocusManager context={context} modal={true}>
+            <Box zIndex={999} position="fixed">
```
anuraghazra, line 208: Can we add a TODO here for future reference? 

IMO as with all the things we should keep zIndexes in check, We should also expose zIndex (elevation rip naming) tokens and have proper guidelines on stacking context on which components will take zIndex priority. 

eg: 

Popovers: 2000
Tooltip: 1000
BottomSheet: 500
Modal: 500


In most of our products people are using random zindexes and because of that there were few zindex issues anshul faced while implementing BottomSheet, where from their end they were using a ridiculously high zindex values.

anuraghazra:
@chaitanyadeorukhkar one important thing: 

In BottomSheet Header we had few augmentations that we did for the trailing component like for example: 

- Button will automatically become xsmall, tertiary
- Badge will always have size of medium
- etc etc

We basically enhance the trailing component. 
And we throw/warn users if they changed these props.

Are we doing that on Modal? This logic is not inside BaseHeader because different different components might have different restrictions & augmentations.

packages/blade/src/components/Modal/Modal.web.tsx
chaitanyadeorukhkar: As a starting point I'm abstracting Modal's zIndex into a common place. We should tokenize this

divyanshu013:
I think we can take the overflow overlay later as enhancement

packages/blade/src/components/Modal/Modal.native.tsx
chaitanyadeorukhkar: I've added this just to avoid TS errors at build time. This is a non-critical file and might be pointless to maintain this in single place too. I'll add a comment

packages/blade/src/components/BaseHeaderFooter/BaseHeader.tsx
```diff
@@ -51,6 +51,63 @@ type BaseHeaderProps = {
...
+            ', ',
+          )}\` component is accepted as trailing`,
+        );
+      }
+
+      const restrictedPropKeys = Object.keys(propRestrictionMap[trailingComponentType]);
+      for (const prop of restrictedPropKeys) {
+        if (trailing?.props?.hasOwnProperty(prop)) {
+          console.warn(
+            `[Blade BottomSheetHeader]: Do not pass "${prop}" to "${trailingComponentType}" while inside BottomSheetHeader trailing, because we override it.`,
```
anuraghazra, line 98: Let's use metaComponentName for the warning.

chaitanyadeorukhkar: Making it `Header`, metaComponentName will mostly be kebab case with smallcase letters. Will look weird.

packages/blade/src/components/Modal/Modal.web.tsx
anuraghazra: Weird, could be some sort of timing issue. But anyways we can pick up a refactor task later. [Tracking it here](https://github.com/razorpay/blade/issues/1269)

chaitanyadeorukhkar: Not related to timing, I think it might be related to how it manages focus. I can see my element nested within the element its adding focus to.

anuraghazra (approved):
LGTM

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -0,0 +1,243 @@
...
+  isOpen: boolean;
+  /**
+   *  Callback function when user clicks on close button or outside the modal or on pressing escape key.
+   */
+  onDismiss: () => void;
+  /**
+   *  Ref to the element to be focused on opening the modal.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  initialFocusRef?: React.MutableRefObject<any>;
```
anuraghazra, line 54: This eslint ignore might interfere with the jsdoc comment. let's move it to top of the file.

packages/blade/package.json
```diff
@@ -108,6 +108,7 @@
   "dependencies": {
     "@babel/runtime": "7.16.5",
     "@types/body-scroll-lock": "3.1.0",
+    "@floating-ui/react": "0.24.2",
```
anuraghazra, line 111: We will need to add this in peer dep also?

chaitanyadeorukhkar: Added

packages/blade/src/components/Box/BaseBox/types/spacingTypes.ts
saurabhdaware: So the issue that was raised should be unrelated to that. The types are broken because of forwardRef. Have found a fix for that, will create PR 🤔

anuraghazra (approved):
Looks good :shipit:

packages/blade/src/components/Box/BaseBox/types/spacingTypes.ts
saurabhdaware: Nice!!

divyanshu013 (commented):
Looks good overall, minor comments

packages/blade/package.json
chaitanyadeorukhkar: Removed lol. Forgot this was not needed on native. We add it when its needed on native too.

packages/blade/src/components/Modal/Modal.web.tsx
chaitanyadeorukhkar: Should be okay, right? Kamlesh had also asked to add this. In fact he suggested we throw an error but we took the call of throwing a warning instead.

packages/blade/src/components/Box/BaseBox/types/spacingTypes.ts
saurabhdaware: #1272

packages/blade/src/components/Modal/Modal.web.tsx
divyanshu013: Though this will also throw warnings on tablet size

divyanshu013 (approved):
Looks good to me, one query on the warning

packages/blade/src/components/Modal/Modal.web.tsx
chaitanyadeorukhkar: Should be fine. We want every non-desktop platform to use bottomsheet only

PR #1229, opened 2023-05-30
feat(blade): add textAlign to Box

anuraghazra:
Added textAlign property to Box. 

- Tested wrapping all the Blade components inside Box to see if anything breaking
- Tested on react-native

Here are the screenshot with all the components inside Box textAlign=center

image

image

image


packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -187,11 +187,11 @@ const Alert = ({
   const _title = title ? (
     <BaseBox marginBottom="spacing.2">
```
saurabhdaware, line 188: maybe add `textAlign="left"` here or the outermost element so you don't have to add on individual items?

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -15,6 +15,7 @@ type LayoutProps = MakeObjectResponsive<
     width: SpacingValueType;
     minWidth: SpacingValueType;
     maxWidth: SpacingValueType;
+    textAlign: CSSObject['textAlign'];
   } & PickCSSByPlatform<'display' | 'overflow' | 'overflowX' | 'overflowY'>
```
saurabhdaware, line 19:
```suggestion
  } & PickCSSByPlatform<'display' | 'overflow' | 'overflowX' | 'overflowY' | 'textAlign'>
```

packages/blade/src/components/Card/CardFooter.tsx
```diff
@@ -82,12 +82,12 @@ const _CardFooterLeading = ({ title, subtitle }: CardFooterLeadingProps): React.
   return (
     <BaseBox>
```
saurabhdaware, line 83: add `textAlign` here instead?

packages/blade/src/components/Typography/Heading/Heading.tsx
```diff
@@ -17,6 +17,7 @@ type HeadingCommonProps = {
   type?: TextTypes;
   contrast?: ColorContrastTypes;
   children: StringChildrenType;
+  textAlign?: BaseTextProps['textAlign'];
```
saurabhdaware, line 20: Should we avoid adding on `Heading` right now? I was thinking if we can remove from `Text` as well and just rely on parent's textAlign 🤔 

cause textAlign will stop working when we support something like `as="span"` no?

packages/blade/src/components/Alert/Alert.tsx
anuraghazra: Done.

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
anuraghazra: It was throwing some error before but not throwing now, maybe some issue was with TSServer. fixed.

packages/blade/src/components/Typography/Heading/Heading.tsx
anuraghazra: > Should we avoid adding on Heading right now? I was thinking if we can remove it from Text as well and just rely on parent's textAlign 🤔

Should be fine no harm on adding textAlign. 

> cause textAlign will stop working when we support something like as="span"
Span seems to respect textAlign property, it doesn't ignore it.

saurabhdaware: https://codesandbox.io/s/razorpay-blade-basic-forked-kmoo5u?file=/src/App.tsx:386-440

It will ignore no? or did you mean when we render Text component as span it works differently?

anuraghazra: Oh i see, I did the inheritance : 

```
      <div style={{ textAlign: "center" }}> 
        <span>hi</span>
      </div>
```

But anyway, it's fine it it ignores too if thats the default behaviour of CSS.

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
anuraghazra: Okay yeah this errors out on ReactNative. because react-native doesn't have textAlign styling on View props. 

I think it will be better to just keep the types same and ignore the prop on react-native. otherwise we will have to put `textAlign={'left' as never}` on all the places where we used it since it will work on web but fail on native

image

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -146,6 +146,7 @@ const _Badge = ({
         ) : null}
         <Text
           {...badgeTextSizes[size]}
+          textAlign="left"
```
saurabhdaware, line 149: this can also go to `StyledBadge`

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
saurabhdaware: `textAlign={castWebType('left')}` should fix it. This is what we do on other props as well.

packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -185,7 +185,7 @@ const Alert = ({
   );
 
   const _title = title ? (
-    <BaseBox marginBottom="spacing.2">
+    <BaseBox marginBottom="spacing.2" textAlign="left">
```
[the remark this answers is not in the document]
anuraghazra: Okay.

packages/blade/src/components/Badge/Badge.tsx
anuraghazra: Done.

saurabhdaware (dismissed):
Small comment - https://github.com/razorpay/blade/pull/1229/files?file-filters%5B%5D=.md&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&show-viewed-files=true#r1210216131

Rest looks good to me 🥳

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
anuraghazra: > his is what we do on other props as well.

Oh really? 

That is actually not the correct way to use CastWebType & CastNativeType. 
These cast functions are suppose to only work with `Branded<>` types (eg: types generated by Platform.Select) 

if you do `castWebType('some random type without brand')` this will return `never` 

So you are basically doing the same thing as `textAlign={'left' as never}`

saurabhdaware: no so we are using `Platform.Select`. That PickCSSByPlatform utility is using `Platform.Select` to decide which type to pick on which platform.

anuraghazra: I think let's do this: 

1. Remove all the wrong usage of castWebType and let's just directly cast them to native since they are either gonna break on web or on native if there is no overlap of the type
2. I'll change the textAlign to `PickCSSByPlatform` so that consumers will atleast get correct types. 
3. I'll also change castWeb & castNative to not accept any random type and only accept Branded types. 

Sounds good?

saurabhdaware: This one is the branded type though no? This is what PickCSSByPlatform does 

image

saurabhdaware: It does seem to work as expected (added screenshots in this PR) https://github.com/razorpay/blade/pull/1113 

Let me know if I missed something

saurabhdaware: Ohhhhhh understood. You meant we have to explicitly pass `as Type` to castWebType 🤦🏼 got it lol I for some reason thought it magically figures out type in web for that prop but yeah makes sense.

saurabhdaware: image
ahh yess this makes more sense 🤦🏼

saurabhdaware: I think let it be for this PR. Will create separate PR to fix these types.

anuraghazra: Yeah, let me to this atleast for the textAlign will fix other types later.

saurabhdaware (approved):

PR #1230, opened 2023-05-30
fix(Dropdown): add min-width and max-width in menu trigger dropdown

Jaynil1611:
Partially Fixes bug #1207 

~Adds `styled-props` support for the Dropdown Overlay component~

Adds minWidth and maxWidth in Menu Triggers for Dropdown

Also, fixes the storybook path in Dropdown docs.

[Jira](https://razorpay.atlassian.net/browse/RXVP-874)

PR #1232, opened 2023-05-31
fix(Dropdown): width in Menu triggers

saurabhdaware:
Stacked on #1230 

In SelectInput, we set width of dropdown according to the trigger. So if SelectInput takes 600px, Dropdown takes 600px.

However in Menus (Button triggers), We have minWidth and maxWidth depending on the content (these values are taken from figma)


packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -128,15 +128,18 @@ const _DropdownOverlay = ({
   return (
     <BaseBox position="relative">
       <StyledDropdownOverlay
-        width={width}
+        width={isMenu ? undefined : width}
+        minWidth="240px"
+        // in SelectInput, we don't want to set maxWidth because it takes width according to the trigger
+        maxWidth={isMenu ? '400px' : undefined}
+        left={isMenu ? 'spacing.0' : undefined}
```
Jaynil1611, line 135: There is one use case where keeping `left='0'` might overflow on the right.
Can we add styled props support for aligning the position of the dropdown overlay?

image

saurabhdaware: Started discussion for styled-props internally. Will update you on that. we can maybe merge minWidth and maxWidth changes first and will keep discussing styled-props parallely meanwhile?

PR #1242, opened 2023-06-01
feat: add generic `autoCompleteSuggestionType` in Inputs

### LINKED ISSUE #1159: Allow `autocomplete='on'` as an option for TextInput
For now we support `off` and a bunch of other specific (`phone`, `name`, etc.) but not a generic `on`. We have use-cases where `on` is required where a text input can be a mobile number or an email id.

Slack Ref: https://razorpay.slack.com/archives/C01H13RTF8V/p1683270900980509

MDN Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#:~:text=login%20fields.-,%22on%22,-The%20browser%20is

saurabhdaware:
Resolves #1159

Added a comment on the issue to show how it behaves in browser

### DIFF
--- a/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
+++ b/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
@@ -170,6 +170,10 @@ export type BaseInputProps = FormInputLabelProps &
      * - iOS: `textContentType`
      * - android: `autoComplete`
      *
+     * **Note:** `generic` autoCompleteSuggestionType becomes
+     * - `none` in Native
+     * - `on` in Web
+     *
      */
     autoCompleteSuggestionType?:
       | 'none'
@@ -186,7 +190,8 @@ export type BaseInputProps = FormInputLabelProps &
       | 'creditCardCSC'
       | 'creditCardExpiry'
       | 'creditCardExpiryMonth'
-      | 'creditCardExpiryYear';
+      | 'creditCardExpiryYear'
+      | 'generic';
     /**
      * Element to be rendered on the trailing slot of input field label
      */
@@ -261,6 +266,7 @@ export type BaseInputProps = FormInputLabelProps &
 
 const autoCompleteSuggestionTypeValues = [
   'none',
+  'generic',
   'name',
   'email',
   'username',
--- a/packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
+++ b/packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
@@ -32,6 +32,7 @@ type StyledComponentAutoCompleteAndroid =
 
 const autoCompleteSuggestionTypeAndroid = {
   none: 'off',
+  generic: 'off', // there isn't autocomplete="on" equivalent on ios
   name: 'name',
   email: 'email',
   username: 'username',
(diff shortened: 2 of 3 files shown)


packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -186,7 +190,8 @@ export type BaseInputProps = FormInputLabelProps &
       | 'creditCardCSC'
       | 'creditCardExpiry'
       | 'creditCardExpiryMonth'
-      | 'creditCardExpiryYear';
+      | 'creditCardExpiryYear'
+      | 'generic';
```
saurabhdaware, line 194: Do suggest better name if possible. Having `on` is too vague and anyway it doesn't exist on native

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -32,6 +32,7 @@ type StyledComponentAutoCompleteAndroid =
 
 const autoCompleteSuggestionTypeAndroid = {
   none: 'off',
+  generic: 'off', // there isn't autocomplete="on" equivalent on ios
```
saurabhdaware, line 35:
```suggestion
  generic: 'off', // there isn't autocomplete="on" equivalent on android
```

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
```diff
@@ -134,6 +135,8 @@ const _StyledBaseInput: React.ForwardRefRenderFunction<
       : undefined,
   };
 
+  console.log('HEREEE', commonProps.autoComplete, autoCompleteSuggestionType);
```
saurabhdaware, line 138:
```suggestion
```

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
divyanshu013: Should we call it `default`?

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -170,6 +170,10 @@ export type BaseInputProps = FormInputLabelProps &
      * - iOS: `textContentType`
      * - android: `autoComplete`
      *
+     * **Note:** `generic` autoCompleteSuggestionType becomes
+     * - `none` in Native
+     * - `on` in Web
```
divyanshu013, line 175: Can this be handled with platform specific types?

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
saurabhdaware: It can be confused with default value of that prop. Like `default` won't be the default value of autocomplete prop 😅 🙈

saurabhdaware: I think we can call it `on` and just disable that value in react native with platform-specific types.
https://github.com/razorpay/blade/pull/1242#discussion_r1212835771

Web developers will know what `on` means and will use it properly and it won't show on react native at all

PR #1248, opened 2023-06-02
fix: update migrate-typography script

chaitanyadeorukhkar:
Fixes the path to codemod. We will also need to release this fix to previous versions of blade.


divyanshu013 (approved):
I think we can skip doing this fix on older versions (anyway we don't have a seamless way to release a fix on older versions). Most consumers should have run the migration script on the version it was introduced on 🤔

PR #1251, opened 2023-06-02
test: add tests for Modal

chaitanyadeorukhkar:
Adds tests for Modal

**Pending:** 
- [ ] Add tests for native


packages/blade/src/components/Modal/__tests__/Modal.web.test.tsx
```diff
@@ -0,0 +1,244 @@
...
+          </Button>
+        </>
+      );
+    };
+
+    const { getByRole, queryByRole } = renderWithTheme(<Element />);
+    await user.click(getByRole('button'));
+    expect(queryByRole('dialog')).toBeInTheDocument();
+    await user.keyboard('{Escape}');
+    expect(onDismiss).toHaveBeenCalledTimes(1);
```
divyanshu013, line 129: You may also expect the dialog to not be there anymore after this point

packages/blade/src/components/Modal/__tests__/Modal.web.test.tsx
```diff
@@ -0,0 +1,244 @@
...
+            Toggle Modal
+          </Button>
+        </>
+      );
+    };
+
+    const { getByRole, getByTestId } = renderWithTheme(<Element />);
+    await user.click(getByRole('button'));
+    await user.click(getByTestId('modal-backdrop'));
+    expect(onDismiss).toHaveBeenCalledTimes(1);
```
divyanshu013, line 159: We can also expect the modal to not be rendering anymore

divyanshu013 (commented):
Looks good overall, I think similar tests for native would do

packages/blade/src/components/Modal/__tests__/Modal.web.test.tsx
chaitanyadeorukhkar: There is some issue with the state not changing on jest randomly which made the test very flaky. Initially I was asserting that only but switched to asserting call of `onDismiss`. If I run the test as `it.only` it passes but if I run all of them together it fails when trying to assert dialog's dismount :face

packages/blade/src/components/Modal/__tests__/Modal.web.test.tsx
chaitanyadeorukhkar: Same as this: https://github.com/razorpay/blade/pull/1251#discussion_r1218955256

packages/blade/src/components/Modal/__tests__/Modal.web.test.tsx
divyanshu013: Strange, might be due to animation but non-blocker

packages/blade/src/components/Modal/__tests__/Modal.native.test.tsx
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable @typescript-eslint/no-empty-function */
+import React from 'react';
+import { Modal, ModalBody } from '../Modal';
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
+import { Text } from '~components/Typography';
+
+describe('Modal', () => {
+  beforeEach(() => {
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
+  });
```
divyanshu013, line 10: Let's overwrite this only on appropriate blocks ref #1146

divyanshu013 (approved):
Looks good to me.

anuraghazra (approved):
Looks good.

anuraghazra:
Tracking on separate issue for useTrailingRestriction tests: https://github.com/razorpay/blade/issues/1270

PR #1254, opened 2023-06-03
fix: make Badge and Counter display inline-flex

### LINKED ISSUE #1241: Make `Badge` and `Counter` display inline-block
I think we should make Badge display inline-block by default since it is not taking the full width.

And Link followed by Text can be implemented this way right now-
```
<Text>Some Text <Link>Link</Link></Text>
```

_Originally posted by @saurabhdaware in https://github.com/razorpay/blade/issues/1214#issuecomment-1568268614_

If it has to be `flex`, we can probably make it `inline-flex`

archie252000:
Resolves #1241


packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -134,7 +134,7 @@ const _Badge = ({
         paddingLeft={horizontalPadding[size]}
         paddingTop={verticalPadding[size]}
         paddingBottom={verticalPadding[size]}
-        display="flex"
+        display="inline-flex"
```
saurabhdaware, line 137: `inline-flex` is not supported on react native. We'll have to add `isReactNative() ? 'flex' : 'inline-flex'` check here

saurabhdaware:
Thanks for the PR!! Added one comment.

anuraghazra:
Let's update the snapshots?

divyanshu013 (commented):
Please also add some relevant screenshots in PR description (also please validate for React Native)

archie252000:
using inline-flex is generating type errors

archie252000:
Noticed one thing ```isReactNative() ? 'flex': 'inline-flex'``` only seems to work only in Storybook story files. In components, it gives the aforementioned type error (error TS2769: No overload matches this call) on the npm pack.

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -134,7 +134,7 @@ const _Badge = ({
         paddingLeft={horizontalPadding[size]}
         paddingTop={verticalPadding[size]}
         paddingBottom={verticalPadding[size]}
-        display="flex"
+        display={isReactNative() ? "flex" : "inline-flex"}
```
anuraghazra, line 137:
```suggestion
        display={(isReactNative() ? "flex" : "inline-flex") as never}
```

This should fix the TS error. you can double check by running `yarn tsc:blade`

saurabhdaware: Oh yes this is the weird error we randomly face sometimes because of faulty ts plugin. Let me take a pull and fix it.

saurabhdaware: @archie252000 You can check and merge this PR into your branch - https://github.com/archie252000/blade/pull/1

I have fixed the build issue and updated the snapshot tests so checks should work fine with this.

archie252000: Yeah, it works. Build errors are gone. Thanks!

saurabhdaware (dismissed):
Thank you so much for contributing this @archie252000 !! Super helpful bug fix. This will be released in next version we release

saurabhdaware (approved):

saurabhdaware (approved):

archie252000:
> Thank you so much for contributing this @archie252000 !! Super helpful bug fix. This will be released in next version we release

The pleasure is all mine!

Issue #978, opened 2023-01-25
Create `Divider` component 


kamleshchandnani:
We have implemented `Divider` in two places now:
* Cards
* ActionList

This seems to increase the need for `Divider` as a generic component. So we shall create this as a component sooner rather than later

abhieshekumar:
folks will want to pick it up. As an exercise -
- tried out a [dummy divider component](https://codesandbox.io/s/divider-example-forked-coydmt?file=/src/Divider.tsx) (I know there is a whole process that happens before 🙈).
- uses Transient props, which prevents props from getting added to final HTML #714

Please let me know how to proceed? @kamleshchandnani @divyanshu013

divyanshu013:
For this component, you can go through the existing components (some are listed in the OP) and first list down the usecases and a proposed API here.

Once we've the possible usecases and an API we can proceed to the implementation (this will help minimizing code changes later).

abhieshekumar:
got it. thank @divyanshu013 not quiet sure what you mean by `listed in the OP`, can you please share?

divyanshu013:
I meant the ones listed in the parent comment [here](https://github.com/razorpay/blade/issues/978#issue-1556205532)

abhieshekumar:
Please check [this document](https://docs.google.com/document/d/1ZQbea8v8gGpwSYQAnpTcL3SCgyWYYzNX9AFwoZyv2PI/edit?usp=sharing)

divyanshu013:
Thanks for the doc. The design research is currently pending on this from our design team (current draft is [here](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=37369-560819&t=4gjztutJU1ndRBvb-0)).

Since this is pending a thorough design research we would like to currently scope the usecases only to what we've in the code i.e. the components like bottom sheet, cards, action list, etc. Things such as orientation, type, alignment etc should be skipped for now.

As next steps you can scope out the doc and propose an API that works only for our existing components (so you'll have to get rid of things which don't have a usecase in existing components - keeping things simpler).

After that perhaps draft out a PR that refactors one component to use this simple `Divider` component. We can review that and if it looks good you can refactor other components too.

abhieshekumar:
[...]

PR #1258, opened 2023-06-05
refactor: refactor divider component

abhieshekumar:
Addresses #978 
[Decision doc](https://docs.google.com/document/d/1kVIDa2MEU-xG4s9UBUVtGf1mrek2Xu30kDu4Ko93hL0/edit?usp=sharing)

### Horizontal Divider
image

### Vertical Divider
image


saurabhdaware:
Is this ready for review?

anuraghazra:
@abhieshekumar recently we refactored out the divider component on design side also. Please check the design for any discrepancies https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=42811-603965&t=qhjHg06dmpIzVOsc-0

abhieshekumar:
1. yes @saurabhdaware it is ready for review
2. @anuraghazra the initial idea was to just refactor all instances of divider and use a single component instead [[ref](https://github.com/razorpay/blade/issues/978#issuecomment-1566992084)]

chaitanyadeorukhkar:
@abhieshekumar now that we have the design ready, would you be willing to incorporate that in this PR itself? The design is not too complex, we will just need to add a few props to Divider component and change color/border-style based on that. I will help you out if you need any guidance here.

abhieshekumar:
@saurabhdaware can you please do a round of review?

packages/blade/src/components/Divider/Divider.web.tsx
```diff
@@ -0,0 +1,15 @@
+import styled from 'styled-components';
+import React from 'react';
+import type { DividerProps, StyledDividerProps } from './types';
+import { getDividerStyles, makeStyledDividerProps } from './getDividerStyles';
+
+const StyledDivider = styled.div<StyledDividerProps>(
```
anuraghazra, line 6: Use BaseBox.

packages/blade/src/components/Divider/getDividerStyles.ts
```diff
@@ -0,0 +1,43 @@
+import type { CSSObject } from 'styled-components';
+import { makeSize } from '../../utils/makeSize';
+import type { ColorContrast } from '../../tokens/theme/theme';
+import type { DividerProps, StyledDividerProps } from './types';
+import type { Theme } from '~components/BladeProvider';
+
+const makeStyledDividerProps = ({
+  orientation = 'horizontal',
+  style = 'solid',
```
anuraghazra, line 9: Why is the prop coming as style but we are returning it as "type"?

packages/blade/src/components/Divider/Divider.native.tsx
```diff
@@ -0,0 +1,15 @@
...
+import React from 'react';
+import type { DividerProps, StyledDividerProps } from './types';
+import { getDividerStyles, makeStyledDividerProps } from './getDividerStyles';
+
+const StyledDivider = styled.View<StyledDividerProps>(
+  ({ theme, orientation, type, variant, thickness, contrast }) =>
+    getDividerStyles({ orientation, type, variant, thickness, contrast, theme }),
+);
+
+const Divider = (props: DividerProps): React.ReactElement => (
```
anuraghazra, line 11: Were you able to run RN setup and test things out?

packages/blade/src/components/Divider/Divider.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export const DividerHorizontal: ComponentStory<
+  typeof DividerComponent
+> = DividerHorizontalTemplate.bind({});
+DividerHorizontal.storyName = 'Horizontal';
+
+const DividerVerticalTemplate: ComponentStory<typeof DividerComponent> = () => {
+  return (
+    <BaseBox height="100px" display="flex" alignItems="center" justifyContent="center">
+      <DividerComponent orientation="vertical" />
```
anuraghazra, line 74: Can we have a more real world storybook examples? 

eg: instead of rendering just the divider show it with content. like Divider between two text columns.

packages/blade/src/components/Accordion/AccordionItem.tsx
```diff
@@ -1,10 +1,10 @@
 import type { ReactElement, ReactNode } from 'react';
+import { Divider } from '../Divider';
```
chaitanyadeorukhkar, line 2: Lets move to aliased imports

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -139,7 +135,11 @@ const _ActionListSection = ({
       >
         {children}
       </BaseBox>
-      {_hideDivider && isReactNative() ? null : <ActionListSectionDivider />}
+      {_hideDivider && isReactNative() ? null : (
+        <Box marginX="spacing.3" marginY="spacing.1">
```
chaitanyadeorukhkar, line 139: Why have we added these margins here?

packages/blade/src/components/Divider/Divider.native.tsx
```diff
@@ -0,0 +1,15 @@
+import styled from 'styled-components/native';
```
chaitanyadeorukhkar, line 1: Instead of having `Divider.native` & `Divider.web` can we have a single `Divider` file and use `BaseBox` within it. BaseBox has `borderColor`, `borderStyle`, `width` & `height`

chaitanyadeorukhkar: This will simplify the complexity of code for Divider

abhieshekumar: done, however -
- could not find `borderStyle` on `BaseBox` (have used `style` prop for now). Please let me know if missed something.
- default `borderWidth` supports `none, thick, and thin`. However we also have a `thinner` variant here. Have also used style prop for that.
- Is it safe to assume that Divider will always be rendered inside a flex container -
    - If yes, then for vertical divider to take the full height have added `alignSelf: stretch` property.
    - If no, then how do we determine (to also add `height: 100%`) whether parent container is flex of not?
        - Tried to explore a bit, MUI uses a boolean prop `flexItem` for this scenario 🤔?

packages/blade/src/components/ActionList/ActionListItem.tsx
abhieshekumar: The [original ActionListSectionDivider](https://github.com/razorpay/blade/blob/84847e1511084b2bdd6f019b79e08e5398f13251/packages/blade/src/components/ActionList/ActionListItem.tsx#L85) component has margins, have tried to replicate that.

packages/blade/src/components/Divider/Divider.native.tsx
abhieshekumar: done, however - 
- the exisiting `borderWidth` allows `none, thin, and thick` but in our usecase there is a `thinner` option as well.
- also could not find `border style` prop on `baseBox`. Please let me know if missed something.
- for both above properties have used `style` prop
    - had tried using shorthand notation `borderLeft: ${borderStyle} ${width}` to avoid `style` prop but borderStyle was getting overridden with `solid`
- Also can we assume that Divider will always be render within a flex container?
    - if no, how do we determine whether parent container is flex? (ref [flexItem prop on MUI divider](https://mui.com/material-ui/api/divider/))
    - for now have used `flexGrow` and `alignItems` to auto adjust the width and height of divider.
    - let's please have this discussion?

packages/blade/src/components/Divider/Divider.tsx
```diff
@@ -0,0 +1,41 @@
...
+  thickness = 'thin',
+  contrast = 'low',
+}: DividerProps): React.ReactElement => {
+  const { theme } = useTheme();
+
+  const isDividerHorizontal = orientation === 'horizontal';
+  const borderPosition = isDividerHorizontal ? 'borderBottom' : 'borderLeft';
+
+  const borderWidth =
+    thickness === 'thinner' ? makeSize(0.5) : makeSize(theme.border.width[thickness]);
```
anuraghazra, line 21: Can you add a TODO "Tokenize thinner border width" and get this 0.5 from size[0.5], (add it to the size object if it's not there already)

anuraghazra (approved):
Small comment otherwise looks good.

chaitanyadeorukhkar (approved):

chaitanyadeorukhkar (commented):
Breaking on React native
Simulator Screenshot - iPhone 14 Pro - 2023-07-03 at 11 14 57

chaitanyadeorukhkar (commented):
Need to fix RN vertical divider

chaitanyadeorukhkar (approved):

Issue #1234, opened 2023-05-31
Types: fix platform specific Box types assertions


anuraghazra:
https://github.com/razorpay/blade/pull/1229#discussion_r1211150698

PR #1261, opened 2023-06-06
fix(Amount): Amount display property

saurabhdaware:
Amount component should ideally be `inline-flex` and not `flex` to avoid such issues - https://razorpay.slack.com/archives/CMQ3RBHEU/p1685967801896139?thread_ts=1685526177.654729&cid=CMQ3RBHEU

### DIFF
--- a/packages/blade/src/components/Amount/Amount.tsx
+++ b/packages/blade/src/components/Amount/Amount.tsx
@@ -13,7 +13,7 @@ import type { Feedback } from '~tokens/theme/theme';
 import type { BaseTextProps } from '~components/Typography/BaseText/types';
 import BaseBox from '~components/Box/BaseBox';
 import type { TestID } from '~src/_helpers/types';
-import { getPlatformType, metaAttribute, MetaConstants } from '~utils';
+import { castNativeType, castWebType, getPlatformType, metaAttribute, MetaConstants } from '~utils';
 import { getStyledProps } from '~components/Box/styledProps';
 import type { StyledPropsBlade } from '~components/Box/styledProps';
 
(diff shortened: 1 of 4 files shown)


packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -251,7 +251,8 @@ const Amount = ({
     <BaseBox
       paddingLeft="spacing.2"
       paddingRight="spacing.2"
-      display="flex"
+      // @TODO: fix casting of platform types. currently they all become `never` type
+      display={isReactNative ? castNativeType('flex') : castWebType('inline-flex')}
```
saurabhdaware, line 255: So this is incorrect usage of castNativeType and castWebType function. It should ideally be `castWebType('flex' as BrandType)` but still figuring out what is the best and reusable way of doing this for such props. For now I am letting it be here and we can change that all at once later since this is how I have done it in other places too.

anuraghazra: Just cast it to never manually. no need to castNative function indirections. 

```
(isReactNative ? 'flex' : 'inline-flex') as never
```

Ref: #1234

kamleshchandnani: yeah casting only works in platform specific files(.native or .web) 😢

anuraghazra: > yeah casting only works in platform specific files(.native or .web)

So kamlesh, in this particular case this isn't working because the string `inline-flex` is not a branded type, thus castNativeType will cause it to become `never (AKA not possible type)` more [info here](https://github.com/razorpay/blade/pull/1229#discussion_r1211133318)

anuraghazra (approved):

packages/blade/src/components/Amount/Amount.tsx
saurabhdaware: Created it here - https://github.com/razorpay/blade/issues/1267

Issue #638, opened 2022-08-17
Reusable focus ring


divyanshu013:
I see that we're using focus ring in different ways, in some components as CSS and at some places via helper functions. Probably can be taken up as a separate PR.

_Originally posted by @divyanshu013 in https://github.com/razorpay/blade/pull/613#discussion_r942610921_

saurabhdaware:
We have getFocusRingStyles utility now which I think @anuraghazra added

PR #1265, opened 2023-06-06
feat(blade): added tooltip component

anuraghazra:
Added Tooltip Component: 

[API Spec](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Tooltip/_decisions/decisions.md)

Web: 

https://github.com/razorpay/blade/assets/35374649/69cc791d-9d0e-41e3-98b4-6baab53f79e3

Native: 

https://github.com/razorpay/blade/assets/35374649/5c85ee32-149d-4a17-96af-1a1c105c5c07


packages/blade/src/components/Button/IconButton/IconButton.tsx
```diff
@@ -31,25 +31,46 @@ type IconButtonProps = {
...
       icon={icon}
       accessibilityLabel={accessibilityLabel}
+      onBlur={onBlur}
+      onFocus={onFocus}
+      onMouseLeave={onMouseLeave}
+      onMouseMove={onMouseMove}
+      onPointerDown={onPointerDown}
+      onPointerEnter={onPointerEnter}
+      onTouchEnd={onTouchEnd}
+      onTouchStart={onTouchStart}
```
divyanshu013, line 73: Maybe have a props map that we can perhaps do `{ ...triggerProps }`

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -324,8 +339,8 @@ const _BaseLink: React.ForwardRefRenderFunction<BladeElementRef, BaseLinkProps>
 
   return (
     <StyledBaseLink
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      ref={ref as any}
+      ref={ref as never}
+      // TODO Check if this is overriden
```
divyanshu013, line 343: Didn't get this todo

packages/blade/src/components/Tooltip/Tooltip.native.tsx
```diff
@@ -0,0 +1,102 @@
...
+  // that will be a wrong UX
+  // if (shouldWrapChildren) {
+  //   console.warn('[Blade Tooltip]: `shouldWrapChildren` prop does nothing on ReactNative');
+  // }
+
+  return (
+    <>
+      {shouldWrapChildren ? (
+        <Pressable
+          style={{ alignSelf: 'flex-start' }}
```
divyanshu013, line 64: Should use stylesheet I think, no literals

packages/blade/src/components/Tooltip/Tooltip.native.tsx
```diff
@@ -0,0 +1,102 @@
...
+          ref={refs.setReference}
+          collapsable={false}
+        >
+          {children}
+        </Pressable>
+      ) : (
+        React.cloneElement(children, {
+          onTouchEnd: handleOpen,
+          ref: refs.setReference,
+          style: { alignSelf: 'flex-start' },
```
divyanshu013, line 77: Won't this change the item's alignment 🤔

packages/blade/src/components/Tooltip/Tooltip.native.tsx
```diff
@@ -0,0 +1,102 @@
...
+        })
+      )}
+      <Modal collapsable={false} transparent visible={isOpen} animationType="fade">
+        <TouchableOpacity
+          style={{
+            flexShrink: 0,
+            flex: 1,
+          }}
```
divyanshu013, line 85: Stylesheet

packages/blade/src/components/Tooltip/Tooltip.native.tsx
```diff
@@ -0,0 +1,102 @@
...
+          {children}
+        </Pressable>
+      ) : (
+        React.cloneElement(children, {
+          onTouchEnd: handleOpen,
+          ref: refs.setReference,
+          style: { alignSelf: 'flex-start' },
+        })
+      )}
+      <Modal collapsable={false} transparent visible={isOpen} animationType="fade">
```
divyanshu013, line 80: Have you tested this both on ios / android in case there are any visual diff?

packages/blade/src/components/Tooltip/Tooltip.web.tsx
```diff
@@ -0,0 +1,153 @@
...
+        <TooltipInteractiveWrapper
+          tabIndex={0}
+          ref={refs.setReference as never}
+          {...makeAccessible({ label: content })}
+          {...getReferenceProps()}
+        >
+          {children}
+        </TooltipInteractiveWrapper>
+      ) : (
+        React.cloneElement(children, { ref: refs.setReference, ...getReferenceProps() })
```
divyanshu013, line 115: Can you add a comment on why we need to clone element?

packages/blade/tsconfig.json
```diff
@@ -11,6 +11,6 @@
     },
     // In Blade side we are targetting `.web` types
     // We can't add both `.native` & `.web` at once.
-    "moduleSuffixes": [".all", ".web", ""]
+    "moduleSuffixes": [".all", ".native", ""]
```
divyanshu013, line 14: Revert?

packages/blade/src/components/Tooltip/TooltipContent.tsx
```diff
@@ -0,0 +1,41 @@
...
+const TooltipContent = React.forwardRef<HTMLDivElement, TooltipContentProps>(
+  ({ children, arrow, style }, ref) => {
+    return (
+      <TooltipContentWrapper
+        position={isReactNative() ? 'absolute' : 'relative'}
+        paddingTop="spacing.3"
+        paddingBottom="spacing.3"
+        paddingLeft="spacing.4"
+        paddingRight="spacing.4"
+        maxWidth={{ base: '120px', l: '160px' }}
```
divyanshu013, line 21: Let's move to the size tokens

packages/blade/src/components/Tooltip/TooltipArrowNative.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const strokeWidth = theme.border.width.thin * 2;
+
+    if (!ref) {
+      console.warn('Floating UI: The `ref` prop is required for the `FloatingArrow`', 'component.');
+    }
+    if (!floating) {
+      return <></>;
+    }
+
+    const [side] = placement.split('-') as [Side];
```
divyanshu013, line 42: Minor: maybe refactor this logic, being used at other places as well

packages/blade/src/components/Tooltip/TooltipArrowNative.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    const svgX = width / 2;
+    const svgY = 0;
+    const dValue =
```
divyanshu013, line 46: Add a comment on what's happening here, what's this `dValue`

packages/blade/src/components/Tooltip/TooltipArrowNative.tsx
```diff
@@ -0,0 +1,106 @@
...
+      left: -90,
+      right: 90,
+    }[side];
+
+    let newStyles = {};
+    if (arrow) {
+      const { x, y } = arrow;
+      newStyles = {
+        width: '20px',
+        height: '20px',
```
divyanshu013, line 72: Let's use the size tokens

divyanshu013 (commented):
Please also add interactions screen recording / screenshots.

packages/blade/src/components/Tooltip/Tooltip.native.tsx
anuraghazra: Removed.

packages/blade/src/components/BladeProvider/BladeProvider.web.tsx
```diff
@@ -14,9 +15,11 @@ const BladeProvider = ({
 
   return (
     <ThemeContext.Provider value={themeContextValue}>
-      <StyledComponentThemeProvider theme={theme}>
-        <BottomSheetStackProvider>{children}</BottomSheetStackProvider>
-      </StyledComponentThemeProvider>
+      <FloatingDelayGroup delay={{ open: 300, close: 300 }}>
```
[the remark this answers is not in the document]
anuraghazra: moved out of the component

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -331,13 +328,17 @@ const _BaseButton: React.ForwardRefRenderFunction<BladeElementRef, BaseButtonPro
     children,
     accessibilityLabel,
     testID,
+    onFocus,
+    onMouseLeave,
+    onMouseMove,
+    onPointerDown,
+    onPointerEnter,
```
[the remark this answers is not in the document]
anuraghazra: Open to suggestion on how can we effectively abstract these out, even if we create some sort of wrapper we will have to accept these functions from the arguments, spread operator helps but in most of our components we are avoiding spread operator. 

```jsx
const Button = ({
    onFocus,
    onMouseLeave,
    onMouseMove,
    onPointerDown,
    onPointerEnter,
}) => {
  return <button 
      onBlur={onBlur}
      onFocus={onFocus}
      onMouseLeave={onMouseLeave}
      onMouseMove={onMouseMove}
      onPointerDown={onPointerDown}
      onPointerEnter={onPointerEnter}
   />
}
```

packages/blade/src/components/Button/IconButton/IconButton.tsx
anuraghazra: Would work, but then in the components we will have to accept 

```jsx
<Button triggerProps={} /> 
```

Not sure if this is going to be better than just accepting all the trigger props in the Button as is, like any other react component. 🤔

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
anuraghazra: Will fix this TODO. basically our trigger props are overriding this synthetic events for active state of link

packages/blade/src/components/Tooltip/Tooltip.native.tsx
anuraghazra: Done.

packages/blade/src/components/Tooltip/Tooltip.native.tsx
anuraghazra: This should be fine really, this is under tooltip content and tooltip contents will only have string text components. So i don't foresee any performance hits. And anyway the component will always render when isOpen state changes since it's under a Modal.

packages/blade/src/components/Tooltip/Tooltip.native.tsx
anuraghazra: No visual difference, the Modal I'm using just for the backdrop and opening state, it doesn't do any visual renders.

packages/blade/src/components/Tooltip/Tooltip.web.tsx
anuraghazra: Added

packages/blade/src/components/Tooltip/TooltipContent.tsx
anuraghazra: Done.

packages/blade/src/components/Tooltip/TooltipArrowNative.tsx
anuraghazra: Done.

packages/blade/src/components/Tooltip/TooltipArrowNative.tsx
anuraghazra: I really don't know 😆 ,  this is a modified version of [FloatingArrow](https://github.com/floating-ui/floating-ui/blob/master/packages/react/src/components/FloatingArrow.tsx) to work with ReactNative. 

I've added a comment at top.

kamleshchandnani:
the UX on native, we decided to not implement right? how do you differentiate click and tap on native?

anuraghazra:
> the UX on native

We will and have implemented that. There are guidelines from design not to use Tooltip with an interactive element on mobile, because we cannot detect if the user's intent to do the action or open the tooltip. 

We have discussed this and I've added details about the [UX in the API decision under open questions.](https://github.com/razorpay/blade/blob/anu/tooltip-api/packages/blade/src/components/Tooltip/_decisions/decisions.md#open-questions).

packages/blade/src/components/Tooltip/types.ts
```diff
@@ -0,0 +1,60 @@
...
+  content: string;
+  placement?: Exclude<
+    UseFloatingOptions['placement'],
+    'left-end' | 'left-start' | 'right-end' | 'right-start'
+  >;
+  children: React.ReactElement;
+  onOpenChange?: ({ isOpen }: { isOpen: boolean }) => void;
+};
+
+type TooltipTriggerProps = {
```
kamleshchandnani, line 17: I feel these shall rather be named as some common event handlers and we can create a type for them and then use it at all the places and not scope it just to TooltipTriggerProps?

packages/blade/src/components/Tooltip/types.ts
```diff
@@ -0,0 +1,60 @@
...
+  onTouchEnd?: Platform.Select<{
+    native: undefined | ((event: any) => void);
+    web: React.TouchEventHandler;
+  }>;
+};
+
+type TooltipContentProps = {
+  children: React.ReactNode;
+  style: CSSProperties;
+  arrow: React.ReactNode;
```
kamleshchandnani, line 43: what is this arrow for?

packages/blade/src/components/Tooltip/getTooltipContentWrapperStyles.ts
```diff
@@ -0,0 +1,25 @@
...
+  theme: Theme;
+  styles: React.CSSProperties;
+}): CSSObject => {
+  return {
+    backgroundColor: theme.colors.brand.gray[200].highContrast,
+    borderWidth: makeBorderSize(theme.border.width.thin),
+    borderRadius: makeBorderSize(theme.border.radius.medium),
+    borderColor: theme.colors.brand.gray[300].highContrast,
+    borderStyle: 'solid',
+    // TODO: fix RN shadow
```
kamleshchandnani, line 19: is this fixed?

packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.web.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const TooltipInteractiveWrapper = styled(BaseBox).attrs(() => {
+  return { tabIndex: -1 };
+})((props) => {
+  return {
+    display: 'inline-block',
+    '&:focus': {
+      borderRadius: makeBorderSize(props.theme.border.radius.medium),
+      // TODO: Replace with focus outline token
+      outline: `1px solid ${props.theme.colors.surface.background.level1.lowContrast}`,
```
kamleshchandnani, line 13: when we add an Interactive wrapper shouldn't we make the trigger element focussable?

packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.web.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { makeBorderSize } from '~utils';
+
+const TooltipInteractiveWrapper = styled(BaseBox).attrs(() => {
+  return { tabIndex: -1 };
+})((props) => {
+  return {
+    display: 'inline-block',
+    '&:focus': {
+      borderRadius: makeBorderSize(props.theme.border.radius.medium),
+      // TODO: Replace with focus outline token
```
kamleshchandnani, line 12: can you replace this token and then remove this Todo?

packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.native.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Pressable } from 'react-native';
+import styled from 'styled-components/native';
+
+const StyledPressable = styled(Pressable)(() => {
+  return { alignSelf: 'flex-start' };
+});
+
+const TooltipInteractiveWrapper = React.forwardRef<View, PressableProps>((props, ref) => {
+  return (
+    <StyledPressable ref={ref} collapsable={false} {...props}>
```
kamleshchandnani, line 12: so with this our non-interactive icons will now become pressable? and the default styling props of pressable will also apply?

packages/blade/src/components/Tooltip/TooltipContextWrapper.web.ts
```diff
@@ -0,0 +1,13 @@
+import type { CSSProperties } from 'react';
+import styled from 'styled-components';
+import { getTooltipContentWrapperStyles } from './getTooltipContentWrapperStyles';
+import type { TooltipContentWrapperProps } from './types';
+import BaseBox from '~components/Box/BaseBox';
+
+const TooltipContentWrapper = styled(BaseBox)<
```
kamleshchandnani, line 7: what does this tooltip content wrapper do?

packages/blade/src/components/Tooltip/TooltipContextWrapper.native.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const isHorizontal = side === 'left' || side === 'right';
+    const offset = isOppositeAxis ? -size[4] : size[4];
+
+    const translate = useSharedValue(offset);
+    const opacity = useSharedValue(0);
+
+    const easing = (theme.motion.easing.entrance.effective as unknown) as EasingFn;
+    const duration = theme.motion.duration.quick;
+
+    React.useEffect(() => {
```
kamleshchandnani, line 40: can you add some comments on why all these things are required?

packages/blade/src/components/Tooltip/types.ts
```diff
@@ -0,0 +1,60 @@
...
+    web: React.TouchEventHandler;
+  }>;
+  onTouchEnd?: Platform.Select<{
+    native: undefined | ((event: any) => void);
+    web: React.TouchEventHandler;
+  }>;
+};
+
+type TooltipContentProps = {
+  children: React.ReactNode;
```
kamleshchandnani, line 41: shouldn't this be string?

packages/blade/src/components/Tooltip/Tooltip.web.tsx
```diff
@@ -0,0 +1,129 @@
...
+              {content}
+            </TooltipContent>
+          </BaseBox>
+        </FloatingPortal>
+      )}
+    </>
+  );
+};
+
+export { Tooltip, TooltipProps };
```
kamleshchandnani, line 129: do we need to export `TooltipProps` from here? lets do all the re-export from the `index.ts` these things sometimes lead to circular deps and we bang our heads as to where is this issue happening.

packages/blade/src/components/Tooltip/types.ts
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import type { Side, UseFloatingOptions } from '@floating-ui/react';
+import type { CSSProperties } from 'react';
+import type { Platform } from '~utils';
+import type { BaseBoxProps } from '~components/Box/BaseBox';
+
+type TooltipProps = {
+  content: string;
```
kamleshchandnani, line 8: on api doc did we agree upon content or description?

packages/blade/src/components/Tooltip/Tooltip.web.tsx
```diff
@@ -0,0 +1,129 @@
...
+      if (open) {
+        setIsOpen(true);
+        onOpenChange?.({ isOpen: open });
+      } else {
+        setIsOpen(false);
+        onOpenChange?.({ isOpen: open });
+      }
+    },
+    middleware: [
+      shift({ crossAxis: false, padding: GAP }),
```
kamleshchandnani, line 60: what is this padding for?

packages/blade/src/components/Tooltip/Tooltip.web.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  return (
+    <>
+      {/* Cloning the trigger children to enhance it with ref and event handler */}
+      {React.cloneElement(children, {
+        ref: refs.setReference,
+        ...getReferenceProps(),
+        ...makeAccessible({ label: content }),
+      })}
+      {isMounted && (
```
kamleshchandnani, line 98: is this how their official page suggest to use tooltip by relying on isMounted?

packages/blade/src/components/Tooltip/Tooltip.native.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const Tooltip = ({
+  content,
+  children,
+  placement = 'left',
+  onOpenChange,
+}: TooltipProps): React.ReactElement => {
+  const { theme } = useTheme();
+  const [isOpen, setIsOpen] = React.useState(false);
+
```
kamleshchandnani, line 20: isn't a lot of pieces common in this and web implementation? and a lot of calculation parts and functions. possible to keep them at one central place? for eg: if we decide to tomorrow update the padding dev might forget update at both the places and things will become inconsistent

packages/blade/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,290 @@
...
+import { BankIcon, InfoIcon } from '~components/Icons';
+import { Link } from '~components/Link';
+import { Box } from '~components/Box';
+import { Text } from '~components/Typography';
+import { isReactNative } from '~utils';
+import { List, ListItem } from '~components/List';
+import { IconButton } from '~components/Button/IconButton';
+import StoryPageWrapper from '~src/_helpers/storybook/StoryPageWrapper';
+import { Sandbox } from '~src/_helpers/storybook/Sandbox';
+
```
kamleshchandnani, line 19: we shall also add a notice here to install the peerDependencies for making this work. also we shall add that to bottomsheet.

packages/blade/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,290 @@
...
+      flexShrink={0}
+      padding="spacing.5"
+      backgroundColor="surface.background.level3.lowContrast"
+      {...props}
+    >
+      <Text contrast="low">{children}</Text>
+    </Box>
+  );
+});
+
```
kamleshchandnani, line 111: lets also add a dedicated story for showing a11y specific things and props

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -63,7 +75,8 @@ type BaseLinkCommonProps = {
    */
   htmlTitle?: string;
 } & TestID &
-  StyledPropsBlade;
+  StyledPropsBlade &
+  Omit<TooltipTriggerProps, 'onBlur' | 'onMouseLeave'>;
```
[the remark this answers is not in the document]
anuraghazra: BaseLink already have onBlur & onMouseLeave defined, if we add then again they will get intersected and types will mess up.

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -335,7 +353,26 @@ const _BaseLink: React.ForwardRefRenderFunction<BladeElementRef, BaseLinkProps>
       target={target}
       rel={rel ?? defaultRel}
       onClick={handleOnClick}
-      onBlur={onBlur}
+      {...syntheticEvents}
+      onBlur={(event: any) => {
+        onBlur?.(event);
+        syntheticEvents.onBlur();
```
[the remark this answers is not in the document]
anuraghazra: these synthetic events are coming from `useInteraction` hook, that we use to provide apply active, press etc styles on react-native.

packages/blade/src/components/Link/Link/Link.tsx
```diff
@@ -103,40 +107,60 @@ export type LinkButtonVariantProps = LinkPropsWithOrWithoutIcon & {
...
-  target,
-  rel,
-  accessibilityLabel,
-  size = 'medium',
-  testID,
-  hitSlop,
-  htmlTitle,
-  ...styledProps
-}: LinkProps): ReactElement => {
+const _Link: React.ForwardRefRenderFunction<BladeElementRef, LinkProps> = (
```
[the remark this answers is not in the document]
anuraghazra: it's what React.forwardRef accepts. 

image

React.forwardRef(`component type`)

packages/blade/src/components/Link/Link/Link.tsx
```diff
@@ -103,40 +107,60 @@ export type LinkButtonVariantProps = LinkPropsWithOrWithoutIcon & {
...
+    onFocus,
+    onMouseLeave,
+    onMouseMove,
+    onPointerDown,
+    onPointerEnter,
+    onTouchStart,
+    onTouchEnd,
+    ...styledProps
+  },
+  ref,
```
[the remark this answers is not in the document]
anuraghazra: I think in Link we missed adding the ref.

packages/blade/src/components/Tooltip/Tooltip.native.tsx
anuraghazra: The overall structure might look very similar but web & native works quite differently in behaviour. 

In native it works like a Modal, while on web it's a proper mouse over tooltip, so it's best to keep things separated for now.

> if we decide to tomorrow update the padding dev

This file in particular doesn't have any styles which are common across platforms, all the places where we have styles I made sure to only have 1 abstracted out component. eg `TooltipContentWrapper`

packages/blade/src/components/Tooltip/Tooltip.stories.tsx
anuraghazra: 🤔 Not sure if we need a dedicated story for that, all the stories are a11y complete, just open the screenreader and good to go. Consumers don't need to do anything to for a11y.

packages/blade/src/hooks/useBladeInnerRef.web.ts
```diff
@@ -25,6 +25,7 @@ const useBladeInnerRef = (
       return {
         focus: (opts) => (handlers?.onFocus ? handlers.onFocus(opts) : element?.focus?.(opts)),
         scrollIntoView: (opts) => element?.scrollIntoView?.(opts),
+        getBoundingClientRect: () => element?.getBoundingClientRect(),
```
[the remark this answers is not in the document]
anuraghazra: Yeah dead code, will remove. 

there are still few components which are using useBladeInnerRef. 
Should I just remove this helper all together and expose the ref as is? 


image

packages/blade/src/components/Tooltip/utils.ts
```diff
@@ -0,0 +1,9 @@
+import type { Alignment, Placement, Side } from '@floating-ui/react';
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+const getPlacementParts = (placement: NonNullable<Placement>) => {
```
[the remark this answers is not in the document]
anuraghazra: Using this in few places to get the side, alignment in typesafe manner. 

```
const [side, alignment] = getPlacementParts('top-left');

side // top
alignment // left
```

packages/blade/src/components/Tooltip/types.ts
anuraghazra: That will work too, can rename this to `BladeCommonEvents` tho are we saying we will also expose other events here too in future?

packages/blade/src/components/Tooltip/types.ts
anuraghazra: I pass the Tooltip arrow component to render inside of this TooltipContent, 

```
<TooltipContent
  style={styles}
  arrow={
    <FloatingArrow
      ref={arrowRef}
      context={context}
      ...
    />
  }
>
  {content}
</TooltipContent>
```

packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.web.tsx
anuraghazra: Actually no, i started thinking that it should be focusable. and have tabIndex={0} but then realised a non-interactive component should always be non-interactive. 

Imagine someone adds a tooltip around an image or icon, the image should not suddenly become focusable, if it does it will confuse screen reader users that why does an image getting focus? "is this clickable?". 

Instead the image should define it's presence via the alt text and the associated label (aria-labelledby) which is added by the Tooltip. 

Same goes for other non-interactive elements, imagine a non-interactive InfoIcon besides an Input.
In this case: 

1. Any critical information should not be hidden under a Tooltip (eg: error text) 
2. if we associate supporting text in that InfoIcon, for example input's help text. For screen reader users that help text should also be associated with the Input component itself and not in an entirely different tooltip component.

packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.web.tsx
anuraghazra: Focus ring tokens aren't refactored out. We have many places where we are doing this. Ideally we should do this from design first.

packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.native.tsx
anuraghazra: Yep. Don't think Pressable has any default styling. It's a blank slate.

packages/blade/src/components/Tooltip/TooltipContextWrapper.web.ts
anuraghazra: It's literally the main wrapper. holds the arrow, content does styling etc

image

packages/blade/src/components/Tooltip/TooltipContextWrapper.native.tsx
anuraghazra: Do you mean 

"why do we need this useEffect" 
or 
"why do we need this `opacity.value = withDelay(duration, withTiming(isVisible ? 1 : 0, timings))`"

packages/blade/src/components/Tooltip/types.ts
anuraghazra: Yes changed, tho this is the internal type, public api me string hi hai.

packages/blade/src/components/Tooltip/Tooltip.web.tsx
anuraghazra: Cool. done.

packages/blade/src/components/Tooltip/Tooltip.web.tsx
anuraghazra: So that the tooltip never touches the edge of the screen, it will always have a 4px padding.

packages/blade/src/components/Tooltip/Tooltip.web.tsx
anuraghazra: Yes this is coming from their hook. [useTransitionStyles](https://floating-ui.com/docs/usetransition#usetransitionstyles)

packages/blade/src/components/Tooltip/TooltipArrowNative.tsx
anuraghazra: Done.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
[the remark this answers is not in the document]
anuraghazra: It looks okay for internal use, but the API seems a bit awkward for public use. 

Consumers would get confused on why are the most common event handlers are inside of an object prop called `commonEventHandlers`? And what would happen to other components where we already have onBlur, OnFocus on top level? 

Keeping them on the top level is more natural imo.

packages/blade/src/components/Box/Box.tsx
```diff
@@ -132,6 +132,14 @@ const makeBoxProps = (props: BoxProps): KeysRequired<Omit<BoxProps, 'testID' | '
     onScroll: props.onScroll,
 
     children: props.children,
+    tabIndex: props.tabIndex,
+    onBlur: props.onBlur,
+    onFocus: props.onFocus,
```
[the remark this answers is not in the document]
anuraghazra: Can remove things from Box, but then users won't be able to use Box to create custom tooltip trigger. 

Sounds be fine? thoughts?

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -35,6 +43,10 @@ type BaseLinkCommonProps = {
     native: (event: GestureResponderEvent) => void;
     web: (event: React.FocusEvent<HTMLButtonElement>) => void;
   }>;
+  onMouseLeave?: Platform.Select<{
```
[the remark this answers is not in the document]
anuraghazra: added

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -399,7 +400,7 @@ const _BaseButton: React.ForwardRefRenderFunction<BladeElementRef, BaseButtonPro
 
   return (
     <StyledBaseButton
-      ref={buttonRef as any}
+      ref={ref as never}
```
divyanshu013, line 403: Didn't get why the `never`?

packages/blade/src/components/Button/Button/Button.stories.tsx
```diff
@@ -75,6 +78,7 @@ export default {
   },
   argTypes: {
     ...getStyledPropsArgTypes(),
+    ...getCommonEventArgTypes(),
```
divyanshu013, line 81: The story table is becoming gigantic now 😅

packages/blade/src/components/Tooltip/Tooltip.native.tsx
```diff
@@ -0,0 +1,105 @@
...
+      if (!isOpen) {
+        setIsVisible(false);
+      }
+    }, theme.motion.duration.gentle);
+
+    if (isOpen) {
+      setIsVisible(true);
+    }
+    return () => clearTimeout(id);
+  }, [isOpen]);
```
divyanshu013, line 64: Some lint warning on this line

packages/blade/src/components/Tooltip/Tooltip.native.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  return (
+    <>
+      {/* Cloning the trigger children to enhance it with ref and event handler */}
+      {React.cloneElement(children, {
+        onTouchEnd: handleOpen,
```
divyanshu013, line 70: Can this override a passed `onTouchEnd` on the component?

packages/blade/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,290 @@
...
+import { Text } from '~components/Typography';
+import { isReactNative } from '~utils';
+import { List, ListItem } from '~components/List';
+import { IconButton } from '~components/Button/IconButton';
+import StoryPageWrapper from '~src/_helpers/storybook/StoryPageWrapper';
+import { Sandbox } from '~src/_helpers/storybook/Sandbox';
+
+const Page = (): React.ReactElement => {
+  return (
+    <StoryPageWrapper
```
divyanshu013, line 22: To be updated

packages/blade/src/components/Tooltip/__tests__/Tooltip.native.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+    expect(getByTestId(triggerId)).toBeTruthy();
+    fireEvent(getByTestId(triggerId), 'touchEnd');
+    expect(getByA11yLabel(tooltipContent)).toHaveProp('visible', true);
+
+    // close on clicking backdrop
+    fireEvent.press(getByTestId(modalBackdropId));
+
+    // wait for closing animation to finish
+    await act(async () => {
+      // theme.motion.duration.gentle
```
divyanshu013, line 74: You can also import the paymentTheme and pass the token directly

packages/blade/src/components/Tooltip/getTooltipContentWrapperStyles.ts
divyanshu013: Cleanup

anuraghazra: Fixed.

packages/blade/tsconfig.json
```diff
@@ -11,6 +11,6 @@
     },
     // In Blade side we are targetting `.web` types
     // We can't add both `.native` & `.web` at once.
-    "moduleSuffixes": [".all", ".web", ""]
+    "moduleSuffixes": [".native", ""]
```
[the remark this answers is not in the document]
divyanshu013: Cleanup

packages/blade/docs/guides/Installation.stories.mdx
```diff
@@ -40,10 +40,14 @@ Before you install the package, make sure that you have performed following step
 
 ## ⬇ Add blade to your application
 
-1. Install blade as a dependency. Blade has a peer dependency on `styled-components`, you can skip adding it if you already have it installed in your project.
+1. Install blade as a dependency.
```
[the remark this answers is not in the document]
anuraghazra: Added

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
anuraghazra: yeah any works too. not sure why i removed any. 

> tho quick tip: never is can be better in some cases because you don't need to ignore the eslint rule 😆

packages/blade/src/components/Button/Button/Button.stories.tsx
anuraghazra: Yeah agree, can check later if it's possible to make unneeded panels collapsed by default.

packages/blade/src/components/Tooltip/Tooltip.native.tsx
anuraghazra: Expected. No need to add `theme.motion.duration.gentle`

packages/blade/src/components/Tooltip/Tooltip.native.tsx
anuraghazra: Good catch, yes it will. also happening on web side, added a mergeProps utility to merge the props/functions and then pass into the cloneElement

packages/blade/src/components/Tooltip/Tooltip.stories.tsx
anuraghazra: Updated.

packages/blade/src/components/Tooltip/__tests__/Tooltip.native.test.tsx
anuraghazra: Done.

packages/blade/src/hooks/useBladeInnerRef.web.ts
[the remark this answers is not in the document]
anuraghazra: Let's take a separate task to remove useBladeInnerRef, Right now only the input components are still using them, Tooltip doesn't effect them.

packages/blade/src/components/Tooltip/utils.ts
```diff
@@ -6,4 +8,32 @@ const getPlacementParts = (placement: NonNullable<Placement>) => {
   return [side, alignment] as const;
 };
 
-export { getPlacementParts };
+/**
+ * Merges two sets of props.
+ *
+ * modified from: https://github.com/ariakit/ariakit/blob/f152585bbb698412ced42bcfa059038ef9d40100/packages/ariakit-react-core/src/utils/misc.ts#L48
+ */
+const mergeProps = <T extends Record<string, any>>(base: T, overrides: T) => {
```
divyanshu013, line 16: Some tests for this util might be helpful

divyanshu013 (approved):
Looks good to me.

packages/blade/src/components/Tooltip/utils.ts
anuraghazra: added

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
anuraghazra: Fixed.

packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.web.tsx
anuraghazra: Refer #638

packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.native.tsx
kamleshchandnani: when you press it it'll show that pressable effect. can you verify once? and also now we are actually doing the same thing you don't intend to do as [mentioned here](https://github.com/razorpay/blade/pull/1265#discussion_r1227942459)

packages/blade/docs/guides/Installation.stories.mdx
```diff
@@ -40,10 +40,14 @@ Before you install the package, make sure that you have performed following step
 
 ## ⬇ Add blade to your application
 
-1. Install blade as a dependency. Blade has a peer dependency on `styled-components`, you can skip adding it if you already have it installed in your project.
+1. Install blade as a dependency. 
+Blade has a peer dependency on few libraries, you can skip adding it if you already have it installed in your project.
+
+  - `styled-components`
+  - `@floating-ui/react`
```
kamleshchandnani, line 47: lets add that gorhom as well?

packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.native.tsx
anuraghazra: ReactNative is a different case, in react native Tooltip opens via click and works more like a "popover" thus it will have to be clickable.

packages/blade/docs/guides/Installation.stories.mdx
anuraghazra: That's for react-native I've already added it on the react-native section installation section. See 3rd point.

packages/blade/src/components/Tooltip/__tests__/Tooltip.web.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+    expect(screen.queryByRole('tooltip')).toBeInTheDocument();
+
+    await act(async () => {
+      jest.advanceTimersByTime(1);
+    });
+
+    expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
+  });
+
+  it('should open/close immediately on focus/blur without the default delay of 300ms', async () => {
```
kamleshchandnani, line 74: should just open immediately right? because in the test you're waiting for the animation to completely on blur so I'm assuming it'll close after 300ms?

packages/blade/src/components/Tooltip/__tests__/Tooltip.web.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+    expect(screen.queryByRole('tooltip')).toBeInTheDocument();
+
+    // close
+    fireEvent.blur(getByTestId(testId));
+    await act(async () => {
+      jest.advanceTimersByTime(animationDuration);
+    });
+    expect(screen.queryByRole('tooltip')).not.toBeInTheDocument();
+  });
+
```
kamleshchandnani, line 234: can we add a test to check for meta attributes so we are assured that our analytics won't be missed?

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -6,7 +6,7 @@ import type { Theme } from '~components/BladeProvider';
 import type { Border } from '~tokens/global';
 import type { DotNotationColorStringToken, PickCSSByPlatform, TestID } from '~src/_helpers/types';
 import type { Platform } from '~utils';
-import type { TooltipTriggerProps } from '~components/Tooltip/types';
+import type { BladeCommonEvents } from '~components/Tooltip/types';
```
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/Box/BaseBox/storybookArgTypes.ts
```diff
@@ -253,4 +253,69 @@ const getBaseBoxArgTypes = (): StorybookArgTypes<
   };
 };
 
-export { getBaseBoxArgTypes, getBoxArgTypes, getStyledPropsArgTypes };
+const getCommonEventArgTypes = (): StorybookArgTypes<Record<string, string>> => {
```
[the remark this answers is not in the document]
anuraghazra: Done.

packages/blade/src/components/Icons/_Svg/Svg/Svg.native.tsx
```diff
@@ -12,22 +12,25 @@ const StyledSvg = styled(SvgNative)((props) => {
...
-        fill={fill}
-        ref={ref}
-        {...styledProps}
-      >
-        {children}
-      </StyledSvg>
-    );
-  },
-);
+const _Svg: React.ForwardRefRenderFunction<any, SvgProps> = (
```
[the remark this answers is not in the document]
anuraghazra: `React.forwardRef<SectionList, ActionListBoxProps>()` this is the direct function invocation. 

This is just the return type of the React.forwardRef 

```
React.ForwardRefRenderFunction<any, SvgProps> 
```

Related to https://github.com/razorpay/blade/pull/1265#discussion_r1227903086

Need to do this way otherwise getting #701

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -76,7 +76,7 @@ type BaseLinkCommonProps = {
   htmlTitle?: string;
 } & TestID &
   StyledPropsBlade &
-  Omit<TooltipTriggerProps, 'onBlur' | 'onMouseLeave'>;
+  Omit<BladeCommonEvents, 'onBlur' | 'onMouseLeave'>;
```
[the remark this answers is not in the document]
anuraghazra: onBlur | onMouseLeave are already present on the `BaseLinkCommonProps` if we intersect again it will mess up the types thus omitting.

packages/blade/src/components/Tooltip/Tooltip.web.tsx
```diff
@@ -92,7 +92,7 @@ const Tooltip = ({
       {/* Cloning the trigger children to enhance it with ref and event handler */}
       {React.cloneElement(children, {
         ref: refs.setReference,
-        ...getReferenceProps(),
+        ...mergeProps(children.props, getReferenceProps()),
```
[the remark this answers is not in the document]
anuraghazra: Merges incoming props from users and the props coming form getReferenceProps. 

https://github.com/razorpay/blade/pull/1265#discussion_r1229197816

> This also merges the functions. 

Check the tests to understand [how it works: ](https://github.com/razorpay/blade/pull/1265/commits/44fa1e628eced9a1d0dc735f7216bd07c30cf90b)

packages/blade/src/components/Tooltip/__tests__/Tooltip.web.test.tsx
anuraghazra: It will close and open immediately on focus and blur, keyboard users can't express "I'm going to hover and stay here for a few miliseconds" intent they instantly focus/blur to the elements.

packages/blade/src/components/Tooltip/__tests__/Tooltip.web.test.tsx
anuraghazra: Done. But we need a reusable solution to this similar to "assertAccessible"

PR #1272, opened 2023-06-07
fix: jsdoc for components with refs

saurabhdaware:
Suuuper weird issue but when we had types like -
```ts
type StyledProps = {
  /** margin jsdoc */
  marginX: ''
}

type CheckBoxProps = {
  /** hello from jsdoc */
  myCheckBoxProps: number;
} & StyledProps;


const CheckBox = React.forwardRef(_CheckBox);
```

Then the output `.d.ts` file was this where instead of just pointing to CheckBoxProps and StyledProps types variables, it defined those again (badly). The ones that were statically available like `myCheckBoxProps` continued to have jsdoc but ones that were defined in variable like `StyledProps` or `testID` had their jsdocs stripped for some reason -
```ts
declare const CheckBox: React__default.ForwardRefExoticComponent<{
    /** hello from jsdoc */
  myCheckBoxProps: number;
} & {
  marginX: ''
}>
```

## Solution

Explicitly defining the export variable's type by pointing it to actual props seems to help solve the issue. JSDoc is working now as expected -
image

Definitely an additional overhead on top of existing components. Not sure how to solve this 😢

### DIFF
--- a/packages/blade/src/components/Box/Box.tsx
+++ b/packages/blade/src/components/Box/Box.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import BaseBox from './BaseBox';
 import type { BoxProps, BoxRefType, MakeValueResponsive } from './BaseBox/types';
 import { validBoxAsValues } from './BaseBox/types/propsTypes';
-import type { KeysRequired } from '~src/_helpers/types';
+import type { ForwardRefReturnType, KeysRequired } from '~src/_helpers/types';
 import { assignWithoutSideEffects, isReactNative, metaAttribute, MetaConstants } from '~utils';
 
 const validateBackgroundString = (stringBackgroundColorValue: string): void => {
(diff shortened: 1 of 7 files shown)


packages/blade/src/components/Box/Box.tsx
```diff
@@ -136,6 +136,37 @@ const makeBoxProps = (props: BoxProps): KeysRequired<Omit<BoxProps, 'testID' | '
   };
 };
 
+const _Box: React.ForwardRefRenderFunction<BoxRefType, BoxProps> = (props, ref) => {
```
saurabhdaware, line 139: diff is a bit messed up here but I essentially moved the jsdoc from `_Box` to `Box` (unrelated to this issue)

saurabhdaware:
Ops nope it wasn't refs, it was #1276. Closing this in favour of that.

PR #1275, opened 2023-06-07
chore: update overlay tokens

chaitanyadeorukhkar:


packages/blade/src/tokens/global/opacity.ts
```diff
@@ -26,10 +26,10 @@ export const opacity: Opacity = {
...
   2: 0.18,
   3: 0.32,
-  4: 1.0,
-  5: 1.0,
-  6: 1.0,
-  7: 1.0,
+  4: 0.48,
+  5: 0.56,
+  6: 0.64,
+  7: 0.72,
```
chaitanyadeorukhkar, line 32: We only need these two but have updated 4 & 5 as well to keep it meaningful. Design doesn't have a scale for this right now so this is a change on our side only.

kamleshchandnani: han design doesn't has a scale hence it was kept at `1`. can you keep it at `1` and only change what is know right now that is 4 & 5?

chaitanyadeorukhkar: Took this call with @saurav12 to change this as this makes more sense than keeping it 1. If we change 4 & 5 to the values we need and later we need more values between 0.32 & 0.64 (which saurav mentioned we will), it will be a breaking change to shift the scale. So we added 4,5, 6 & 7 such that 4 & 5 can change in the future if needed and we can use 6 & 7 right now

packages/blade/src/tokens/global/colors.ts
```diff
@@ -58,8 +63,8 @@ export type Color = Readonly<{
     blueGrayDark: ColorNeutralScale;
     navyGrayLight: ColorNeutralScale;
     navyGrayDark: ColorNeutralScale;
-    ashGrayLight: ColorNeutralScale;
-    ashGrayDark: ColorNeutralScale;
+    ashGrayLight: AshGrayColorNeutralScale;
```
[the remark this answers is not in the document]
kamleshchandnani: then why is that we have named it as a1100 in this case? do we have other values for the rest of the scale on design?

packages/blade/src/tokens/global/opacity.ts
kamleshchandnani: sure 👍

PR #1279, opened 2023-06-08
fix(Collapsible): interaction state and animation

divyanshu013:
- Fixes a bug in `BaseLink` with interaction state not resetting properly on blur
- Adds a regression test so we can catch such things in future
- Fixes animation of collapsible link chevron icon (now rotates in opposite direction)

### DIFF
--- a/packages/blade/src/components/Accordion/_decisions/decisions.md
+++ b/packages/blade/src/components/Accordion/_decisions/decisions.md
@@ -66,7 +66,7 @@ import { Accordion, AccordionItem } from '@razorpay/blade';
 | Key         | Type            | Default     | Description                                                                | Required |
 | ----------- | --------------- | ----------- | -------------------------------------------------------------------------- | -------- |
 | title       | `string`        | `undefined` | Title text content                                                         | ✅       |
-| description | `string`        | `undefined` | Body text content                                                          | ✅       |
+| description | `string`        | `undefined` | Body text content                                                          |          |
 | icon        | `IconComponent` | `undefined` | Renders a Blade icon as title prefix (requires `showNumberPrefix={false}`) |          |
 | children    | `JSX`           | `undefined` | Slot, renders any custom content                                           |          |
 
--- a/packages/blade/src/components/Collapsible/CollapsiblePanel.web.tsx
+++ b/packages/blade/src/components/Collapsible/CollapsiblePanel.web.tsx
@@ -44,7 +44,7 @@ const StyledCollapsiblePanel = styled.div<StyledCollapsiblePanelProps>((props) =
 });
 
 const CollapsiblePanel = ({ children }: CollapsiblePanelProps): ReactElement => {
-  const { isExpanded, defaultIsExpanded } = useCollapsibleContext();
+  const { isExpanded, defaultIsExpanded, direction } = useCollapsibleContext();
   const collapsiblePanelRef = useRef<HTMLDivElement>(null);
 
   /**
@@ -121,8 +121,12 @@ const CollapsiblePanel = ({ children }: CollapsiblePanelProps): ReactElement => 
       onTransitionEnd={onTransitionEnd}
     >
       <Box
-        // need margin here to prevent jank
-        marginY="spacing.5"
+        /**
+         * Need a margin inside the outside wrapper so this is
+         * included in height calculations and prevents jank
+         */
+        marginTop={direction === 'bottom' ? 'spacing.5' : 'spacing.0'}
+        marginBottom={direction === 'top' ? 'spacing.5' : 'spacing.0'}
       >
         {children}
       </Box>
--- a/packages/blade/src/components/Link/BaseLink/BaseLink.tsx
+++ b/packages/blade/src/components/Link/BaseLink/BaseLink.tsx
@@ -274,7 +274,12 @@ const _BaseLink: React.ForwardRefRenderFunction<BladeElementRef, BaseLinkProps> 
 ): ReactElement => {
   const [isVisited, setIsVisited] = useState(false);
   const childrenString = getStringFromReactText(children);
-  const { currentInteraction, setCurrentInteraction, ...syntheticEvents } = useInteraction();
+  const {
+    currentInteraction,
+    setCurrentInteraction,
+    onBlur: onBlurInteraction,
+    ...syntheticEvents
+  } = useInteraction();
   const { theme } = useTheme();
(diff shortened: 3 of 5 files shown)


packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -322,6 +327,14 @@ const _BaseLink: React.ForwardRefRenderFunction<BladeElementRef, BaseLinkProps>
     }
   };
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const handleBlur = (event: any): void => {
+    onBlurInteraction();
+    if (onBlur) {
+      onBlur(event);
+    }
```
anuraghazra, line 335:
```suggestion
onBlur?.(event);
```

packages/blade/src/components/Link/BaseLink/__tests__/BaseLink.web.test.tsx
```diff
@@ -391,4 +392,27 @@ describe('<BaseLink />', () => {
...
+
+  it('should handle styles for all interactions', () => {
+    const textContent = 'Slash';
+    const { getByText } = renderWithTheme(<BaseLink variant="button">{textContent}</BaseLink>);
+
+    // default state
+    const linkButtonText = getByText(textContent);
+    expect(linkButtonText).toHaveStyle({
+      color: paymentTheme.colors.onLight.action.text.link.default,
+    });
```
anuraghazra, line 404: snapshots might be better here?

anuraghazra (approved):
Looks good.

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
divyanshu013: I like the explicit conditional block at some places for readability 😅

packages/blade/src/components/Link/BaseLink/__tests__/BaseLink.web.test.tsx
divyanshu013: Wanted to add an explicit regression check because it's easy to miss these in snapshots otherwise, we already have lot of snapshots in this component.

packages/blade/src/components/Accordion/_decisions/decisions.md
```diff
@@ -66,7 +66,7 @@ import { Accordion, AccordionItem } from '@razorpay/blade';
 | Key         | Type            | Default     | Description                                                                | Required |
 | ----------- | --------------- | ----------- | -------------------------------------------------------------------------- | -------- |
 | title       | `string`        | `undefined` | Title text content                                                         | ✅       |
-| description | `string`        | `undefined` | Body text content                                                          | ✅       |
+| description | `string`        | `undefined` | Body text content                                                          |          |
```
kamleshchandnani, line 69: i don't know but i feel we shall remove description as a prop when we have children already. the initial idea was to have description always but now that's gone based on one of the use cases of X i feel we shall remove description all together 😞

packages/blade/src/components/Collapsible/CollapsiblePanel.web.tsx
```diff
@@ -121,8 +121,12 @@ const CollapsiblePanel = ({ children }: CollapsiblePanelProps): ReactElement =>
       onTransitionEnd={onTransitionEnd}
     >
       <Box
-        // need margin here to prevent jank
-        marginY="spacing.5"
+        /**
+         * Need a margin inside the outside wrapper so this is
+         * included in height calculations and prevents jank
+         */
+        marginTop={direction === 'bottom' ? 'spacing.5' : 'spacing.0'}
```
kamleshchandnani, line 128: have we confirmed this with design? because there might be cases where consumers might need lesser margin top. this is 16px not sure if we have this standardised from design side?

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -274,7 +274,12 @@ const _BaseLink: React.ForwardRefRenderFunction<BladeElementRef, BaseLinkProps>
 ): ReactElement => {
   const [isVisited, setIsVisited] = useState(false);
   const childrenString = getStringFromReactText(children);
-  const { currentInteraction, setCurrentInteraction, ...syntheticEvents } = useInteraction();
+  const {
+    currentInteraction,
+    setCurrentInteraction,
+    onBlur: onBlurInteraction,
```
kamleshchandnani, line 280: why do we need onBlur interaction?

packages/blade/src/components/Accordion/_decisions/decisions.md
divyanshu013: I think it would be beneficial to keep it:

- Majority of accordion items will have only description text (more convenient DX)
- Otherwise it'll be cumbersome for consumers to always wrap the passed text with correct props:

```jsx
<AccordionItem title="Title">
  <Text type="subtle">Description</Text> // <-- In most cases users will have to wrap this and remember type="subtle"
</AccordionItem>
```

packages/blade/src/components/Collapsible/CollapsiblePanel.web.tsx
divyanshu013: Yes, this is covered in the [design tokens spec](https://www.figma.com/file/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?type=design&node-id=79-629510&t=iylL78shTIJyxxW7-4)

image

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
divyanshu013: The `useInteraction` hook needs this for resetting the state to `default` when the link is blurred. This was earlier being passed but in some recent refactor got missed.

Have also added a regression test to catch these in the future in this PR.

packages/blade/src/components/Accordion/_decisions/decisions.md
kamleshchandnani: which should be fine. the persona of users you're talking about doesn't exist in our teams 🤣

```
<AccordionItem title="Title">
  <Text type="subtle">Description</Text> // <-- In most cases users will have to wrap this and remember type="subtle"
</AccordionItem>
```
I see this only being used majority of the times

kamleshchandnani (approved):

PR #1284, opened 2023-06-09
fix(dropdown): dropdown overlay positioning

### LINKED ISSUE #1245: Adjust DropdownOverlay position based on screensize and scroll position
If user is scrolled till bottom, DropdownOverlay should open on top. If trigger is on extreme right, overlay should open a bit towards left.

Similar usecase was mentioned in #1207. We partially solved that issue by giving minWidth and maxWidth but we still need to move overlay according to trigger position

Jaynil1611:
Fixes #1245


packages/blade/src/components/Dropdown/dropdownUtils.ts
```diff
@@ -324,3 +325,58 @@ export const makeInputDisplayValue = (selectedIndices: number[], options: Option
...
+  bottom?: SpacingValueType;
+  left?: SpacingValueType;
+  right?: SpacingValueType;
+}
+
+export const POSITION_THRESHOLDS = {
+  top: -400,
+  bottom: -400,
+  left: -400,
+  right: -400,
```
saurabhdaware, line 347: Where do these values come from?

saurabhdaware (commented):
Thanks for this contribution! This is huge for us. 

Added tiny comments below. Tried it out locally and it works great!

Can you revert the story changes and add a new story for this? In Select and in Menu trigger?


Also do take a pull of this branch once. I fixed that ts error.

packages/blade/src/components/Dropdown/dropdownUtils.ts
Jaynil1611: Since you have kept `max-width` for dropdown overlay as `400px`. I am reusing the same number to calculate height, width available in all positions

packages/blade/src/components/Dropdown/docs/stories.ts
```diff
@@ -604,6 +604,46 @@ const WithControlledMultiSelectStory = `
   export default App;
 `;
 
+const WithBottomAlignedSelectStory = `
```
anuraghazra, line 607: The Bottom aligned dropdown overlay suppose to go upwards? 

image

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -60,13 +63,39 @@ type DropdownOverlayProps = {
...
-  const { isOpen, triggererRef, hasLabelOnLeft, dropdownTriggerer } = useDropdown();
+  const { isOpen, triggererRef, hasLabelOnLeft, dropdownTriggerer, setIsOpen } = useDropdown();
   const { theme } = useTheme();
   const [display, setDisplay] = React.useState<'none' | 'block'>('none');
   const [width, setWidth] = React.useState<SpacingValueType>('100%');
+  const [dropdownPosition, setDropdownPosition] = useState<DropdownPosition>({});
 
   const isMenu = dropdownTriggerer !== 'SelectInput';
 
+  const detectOverflowMiddleware = {
```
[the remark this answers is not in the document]
anuraghazra: If it is needed can you show an example for me to visualise what edge case is this handling?

packages/blade/src/components/Dropdown/docs/stories.ts
Jaynil1611: Yes, the playground on local won't work. Check this deployed chromatic [link](https://61c19ee8d3d282003ac1d81c-gwxawxwngs.chromatic.com/?path=/story/components-dropdown-with-select--with-bottom-aligned-select&globals=showInternalComponents:true;measureEnabled:false)

packages/blade/src/components/Dropdown/dropdownUtils.ts
```diff
@@ -324,3 +325,58 @@ export const makeInputDisplayValue = (selectedIndices: number[], options: Option
   // When more than one item is selected, we display the count of items
   return `${selectedIndices.length} items selected`;
 };
+
+interface PositionProp {
```
anuraghazra, line 329: Let's use `type` if possible, interfaces can accidentally be augmented.

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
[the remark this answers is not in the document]
Jaynil1611: We did try out with `flip, autoPlacement, shift` provided out of the box by floating UI, but it didn't work as expected as it used `transform` property which didn't fit the use case here.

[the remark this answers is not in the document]
Jaynil1611: You can see with this [story](https://61c19ee8d3d282003ac1d81c-gwxawxwngs.chromatic.com/?path=/story/components-dropdown-with-button-and-link--with-right-aligned-menu&globals=showInternalComponents:true;measureEnabled:false) & change position of the menu button to the edges to see it in action.

packages/blade/package.json
```diff
@@ -283,14 +283,14 @@
...
-      "limit": "17.5 kb",
+      "limit": "19.5 kb",
       "running": false
     },
     {
       "name": "Import Text only",
       "path": "./build/components/index.web.js",
       "import": "{ Text }",
-      "limit": "13 kb",
+      "limit": "14.5 kb",
```
Jaynil1611, line 293: @anuraghazra I have increased limits here since the validate tree shaking job was failing. [CI](https://github.com/razorpay/blade/actions/runs/5313644080/jobs/9636524037)

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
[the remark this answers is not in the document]
anuraghazra: > We did try out with flip, autoPlacement, shift provided out of the box by floating UI

That's weird, okay we can refactor this later if needed

packages/blade/package.json
anuraghazra: Sounds good.

packages/blade/src/components/Dropdown/dropdownUtils.ts
anuraghazra: Let's add a comment here.

packages/blade/src/components/Dropdown/dropdownUtils.ts
```diff
@@ -324,3 +325,58 @@ export const makeInputDisplayValue = (selectedIndices: number[], options: Option
...
+};
+
+export const POSITION_THRESHOLDS = {
+  top: -400,
+  bottom: -400,
+  left: -400,
+  right: -400,
+};
+
+export const getDropdownOverlayPosition = ({
```
anuraghazra, line 350: Can we add few comments here? What this is doing?

packages/blade/src/components/Dropdown/dropdownUtils.ts
```diff
@@ -324,3 +325,58 @@ export const makeInputDisplayValue = (selectedIndices: number[], options: Option
...
+  triggererEl: HTMLButtonElement | null;
+}): DropdownPosition => {
+  const zeroSpacing: SpacingValueType = 'spacing.0';
+  const { bottom, right } = position;
+
+  const newPosition: DropdownPosition = { left: zeroSpacing };
+
+  if (!isMenu) {
+    if (bottom > POSITION_THRESHOLDS.bottom) {
+      newPosition.bottom = `${Number(triggererEl?.clientHeight) + 32}px`;
```
anuraghazra, line 366: Where is this 32px coming from?

anuraghazra: Let's extract to a variable or get this from the `size[32]` tokens

```
import { size } from '~tokens/global';
```

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
anuraghazra, line 74: Possible to put this middle ware outside of the function so it doesn't gets recreated everytime? 

Maybe for the `setDropdownPosition` you can pass it via arguments

anuraghazra (approved):
LGTM, few comments added

anuraghazra (approved):

PR #1294, opened 2023-06-13
fix(blade): BottomSheet unable to find ActionList component

anuraghazra:
Fixed react-native error "VirtualizedLists should never be nested inside plain ScrollViews"

ActionList's componentId accidentally got removed in this [commit](https://github.com/razorpay/blade/commit/fc8fef39b3eeac5d5f8ccec374a251d097475723#diff-56bfde9740ed80178b00a3439c5943a881552847e2caba74f7bc2d643962f899R159-R161), resulting in BottomSheet to not find the ActionList and satisfy this [condition](https://github.com/razorpay/blade/blob/36d8b2f563ae0ea33860d8bea214d207a3179d09/packages/blade/src/components/BottomSheet/BottomSheetBody.native.tsx#L21). 

> :( really need e2e tests to catch these issues

### DIFF
--- a/.changeset/lazy-bottles-learn.md
+++ b/.changeset/lazy-bottles-learn.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix(blade): BottomSheet unable to find ActionList component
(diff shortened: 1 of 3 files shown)


saurabhdaware (approved):
onu ;__;

PR #1295, opened 2023-06-13
feat(Accordion/Collapsible): PR7 - pre-release

divyanshu013:
Adds Accordion and Collapsible for pre-release


anuraghazra (approved):

packages/blade/src/components/Accordion/index.ts
```diff
@@ -1,2 +1,4 @@
-export * from './Accordion';
+export { Accordion as Accordion_PRE_RELEASE } from './Accordion';
```
kamleshchandnani, line 1: once we do stable release people will have to manually remove this PRE_RELEASE? 😢

kamleshchandnani (approved):
lgtm, except the pre-release export name.

packages/blade/src/components/Accordion/index.ts
divyanshu013: Yup they'll have to change `Accordion_PRE_RELEASE` imports to `Accordion`. I don't forsee any API changes so should be the only change.

PR #1300, opened 2023-06-14
fix(DropdownOverlay): DropdownOverlay width

### LINKED ISSUE #1299: DropdownOverlay is not opening relative to the SelectInput
Facing these issues which weren't there in the previous version:

Screenshot 2023-06-14 at 10 15 33 AM

Ideally, the SelectInput and DropdownOverlay should be of the same width, which was the earlier case.

saurabhdaware:
In `SelectInput` trigger, DropdownOverlay should not have minWidth and it should always take width of the input

Slack Issue: https://razorpay.slack.com/archives/CMQ3RBHEU/p1686718789586659

Resolves #1299

### DIFF
--- a/.changeset/fast-rings-count.md
+++ b/.changeset/fast-rings-count.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix(DropdownOverlay): DropdownOverlay width in SelectInput trigger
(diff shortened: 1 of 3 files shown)


divyanshu013 (approved):
Looks good to me.

divyanshu013 (approved):
:shipit:

PR #1309, opened 2023-06-15
fix(native): patch yoga.cpp react native internal

divyanshu013:
Followed @chaitanyadeorukhkar's suggestion after fixing some other errors and finally got the native builds working after this. This won't fix all the issues but it fixes a definite issue in an internal module.

Uses `patch-package` to patch the internal node module.

Run `yarn` after pulling this branch to let `patch-package` do its thing.

### DIFF
--- a/packages/blade/package.json
+++ b/packages/blade/package.json
@@ -103,7 +103,8 @@
     "watch:test:react-native": "yarn test:react-native --watch --onlyChanged",
     "chromatic": "npx chromatic",
     "publish-npm": "node ./scripts/publishToNpm.js",
-    "generate-github-npmrc": "node ./scripts/generateGitHubRegistryNpmrc.js"
+    "generate-github-npmrc": "node ./scripts/generateGitHubRegistryNpmrc.js",
+    "postinstall": "patch-package"
   },
   "dependencies": {
     "@babel/runtime": "7.16.5",
@@ -207,7 +208,9 @@
     "style-dictionary": "3.7.1",
     "styled-components": "5.3.5",
     "tsconfig-paths-webpack-plugin": "3.5.2",
-    "jsdom-testing-mocks": "1.9.0"
+    "jsdom-testing-mocks": "1.9.0",
+    "patch-package": "7.0.0",
+    "postinstall-postinstall": "2.1.0"
   },
   "peerDependencies": {
     "@gorhom/bottom-sheet": "^4",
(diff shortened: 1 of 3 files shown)


packages/blade/patches/react-native+0.66.5.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/node_modules/react-native/ReactCommon/yoga/yoga/Yoga.cpp b/node_modules/react-native/ReactCommon/yoga/yoga/Yoga.cpp
+index 2c68674..ea95e40 100644
+--- a/node_modules/react-native/ReactCommon/yoga/yoga/Yoga.cpp
++++ b/node_modules/react-native/ReactCommon/yoga/yoga/Yoga.cpp
+@@ -2229,7 +2229,7 @@ static float YGDistributeFreeSpaceSecondPass(
+         depth,
+         generationCount);
+     node->setLayoutHadOverflow(
+-        node->getLayout().hadOverflow() |
```
anuraghazra, line 9: do we have an upstream issue tracking this bug in react-native? This seems like a bug in RN itself. 

Or is it just related to our setup?

divyanshu013: This has been fixed in recent react-native release so once we upgrade we won't need this patch

saurabhdaware (approved):
Tried it locally and this did fix the setup 🕺🏼

packages/blade/package.json
```diff
@@ -207,7 +208,9 @@
     "style-dictionary": "3.7.1",
     "styled-components": "5.3.5",
     "tsconfig-paths-webpack-plugin": "3.5.2",
-    "jsdom-testing-mocks": "1.9.0"
+    "jsdom-testing-mocks": "1.9.0",
+    "patch-package": "7.0.0",
+    "postinstall-postinstall": "2.1.0"
```
kamleshchandnani, line 213: what's this sorcery `postinstall-postinstall` 😆

kamleshchandnani:
for me still ios and android fails 😢 probably something wrong with my setup

kamleshchandnani (approved):

packages/blade/package.json
divyanshu013: They [recommend](https://github.com/ds300/patch-package#why-use-postinstall-postinstall-with-yarn) using this for yarn legacy 😸

divyanshu013:
I think you might have some additional issues, I'd to also resolve some other build issues which were unrelated to this post macos update 🥲

PR #1315, opened 2023-06-16
docs(Tag): Tags API Decision

saurabhdaware:
[**✨ Pretty Document ✨**](https://github.com/razorpay/blade/blob/tag/api-decision/packages/blade/src/components/Tag/_decisions/decisions.md)

### DIFF
--- a/packages/blade/src/components/Tag/_decisions/decisions.md
+++ b/packages/blade/src/components/Tag/_decisions/decisions.md
@@ -98,7 +98,7 @@ On consumer end, the APIs would look like -
 <TextInput tagsSlot={} value="" onChange={} />
 ```
 
-**Alternate Prop Name**
+##### Alternate Prop Name
 
 - `leading` (felt confusing because it is leading to the invisible input tag and not the actual visible TextInput)
 - `valuePrefix` (Similar to `titleSuffix` from BaseHeader)
@@ -193,4 +193,4 @@ Refer to Keyboard Interactions of [react-select](https://react-select.com/home#g
 - Do we want to discuss Tag's integration with TextInput and TextArea? or should I remove it from API decision altogether?
   - We can discuss here
 - ~`<Tag />` or `<Tags />` 😅~ `<Tag />`
-- [`tagsSlot` vs `tagSlot` vs `valuePrefix` vs `leading` on Input components](#alternate-approach-value-prop)
+- [`tagsSlot` vs `tagSlot` vs `valuePrefix` vs `leading` on Input components](#alternate-prop-name)
--- a/packages/blade/src/components/Tag/_decisions/decisions.md
+++ b/packages/blade/src/components/Tag/_decisions/decisions.md
@@ -15,8 +15,8 @@ These are set of interactive keywords that help organise & categorise objects. T
 <Tag
   icon={CheckIcon}
   size="medium"
-  onDismiss={({ value }) => {
-    console.log('Close Icon Clicked', value);
+  onDismiss={() => {
+    console.log('Close Icon Clicked');
   }}
 >
   Unpaid
@@ -42,9 +42,8 @@ type TagProps = {
   /**
    * Callback when close icon on Tag is clicked
    *
-   * Gives a `value` argument which has same text as `children`
    */
-  onDismiss?: ({ value }: { value: ReactText }) => void;
+  onDismiss?: () => void;
 
   /**
    * Text that renders inside Tag
@@ -96,14 +95,14 @@ SelectInput and AutoComplete are not expected to have any changes in consumer AP
 
 On consumer end, the APIs would look like -
 
-#### New `tagsSlot` prop on `TextInput` and `TextArea`
+#### New `tags` prop on `TextInput` and `TextArea`
 
 - Can only be used in Controlled Input
   - Why?
     - This is comparitively a rare usecase so don't see any need of having another `hasTags` props that handles adding tags internally.
 
 ```jsx
-<TextInput tagsSlot={} value="" onChange={} />
+<TextInput tags={} value="" onChange={} />
 ```
 
 ##### Alternate Prop Name
@@ -131,8 +130,8 @@ function App() {
 
   return (
     <TextInput
-      tagsSlot={tags.map((tagName, index) => (
-        <Tag onDismiss={({ value }) => removeTag(value)}>{tagName}</Tag>
+      tags={tags.map((tagName, index) => (
+        <Tag onDismiss={() => removeTag(tagName)}>{tagName}</Tag>
       ))}
       value={inputValue}
       onChange={({ value }) => setInputValue(value)}
@@ -201,8 +200,14 @@ Refer to Keyboard Interactions of [react-select](https://react-select.com/home#g
 - Do we want to discuss Tag's integration with TextInput and TextArea? or should I remove it from API decision altogether?
   - We can discuss here
 - ~`<Tag />` or `<Tags />` 😅~ `<Tag />`
-- [`tagsSlot` vs `tagSlot` vs `valuePrefix` vs `leading` on Input components](#alternate-prop-name)
+  - We went with `<Tag />`
+- [`tagsSlot` vs `tagSlot` vs `valuePrefix` vs `leading` vs `tags` on Input components](#alternate-prop-name)
+  - We went with `tags`
 - Should we support `value` prop that allows us to set unique identifier value for Tag?
   ```jsx
   <Tag value="red">Red</Tag>
+  ```
+  - We decided not to support it right now as it can be statically handled by passing any id you like from `onDismiss`. Can be evaluated in future if required.
+  ```jsx
+  onDismiss={() => removeTag('red')}
   ```
--- a/packages/blade/src/components/Tag/_decisions/decisions.md
+++ b/packages/blade/src/components/Tag/_decisions/decisions.md
@@ -6,7 +6,8 @@ These are set of interactive keywords that help organise & categorise objects. T
 
 ## Design
 
-- [Tags Figma Design](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=29210-567576&t=HLRjz3OTofBtFdDl-0)
+- [Tags - Figma Design](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=29210-567576&t=HLRjz3OTofBtFdDl-0)
+- [Tags - Motion](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=29744-568562&mode=design&t=KI2KRJMzUpz8Ymr8-0)
 
 ## Proposed API
 
@@ -41,15 +42,15 @@ type TagProps = {
    * Text that renders inside Tag
    */
   children: ReactText;
+
+  /**
+   * Disable tag
+   */
+  isDisabled?: boolean;
 };
 ```
 
 _Didn't consider any alternate API. Kept `size` and `children` similar to `<Badge />` and `onDismiss` similar to `<Alert />` and `<Dropdown />`_
-
-## Accessibility
-
-- Ensure Close Icon is focussable
-- Close Icon to have appropriate `aria-label`
 
 ## Integration with Inputs
 
@@ -78,10 +79,6 @@ E.g. This is AutoComplete from Primer where the tags go into a slot and input ge
 SelectInput and AutoComplete are not expected to have any changes in consumer API as rendering of tags will be handled internally based on the selected value.
 
 ### Integration with TextInput and TextArea
-
-> **Warning**
->
-> Open Question: Do we want to discuss this right now? or do I remove it from API decision itself?
 
 > **Note**
 >
@@ -170,8 +167,22 @@ We can extend our `value` prop to accept JSX
 
 </details>
 
+## Accessibility
+
+- Ensure Close Icon is focussable
+- Close Icon to have appropriate `aria-label`
+- Special keyboard interaction when used inside Input components
+  - Pressing `Tab` should move focus into the input and not Tag
+  - Tag is focussable with Left and Right Arrow Keys
+  - Pressing `Backspace` on Tag should remove it
+  - Clicking on Tag should bring focus on Input
+- Tag should be disabled when used inside a disabled input
+
+Refer to Keyboard Interactions of [react-select](https://react-select.com/home#getting-started) for reference. Other Input components with Tag also follow similar keyboard interactions.
+
 ## References
 
+- [react-select](https://react-select.com/home#getting-started)
 - [`TagsInput` component by PluralSight Design System](https://design-system.pluralsight.com/components/tagsinput)
 - [`tokens` prop by Primer AutoComplete.Input](https://primer.style/react/Autocomplete/)
 - [Carbon just putting everything outside of Inputs](https://carbondesignsystem.com/components/tag/usage)
@@ -179,5 +190,7 @@ We can extend our `value` prop to accept JSX
 
 ## Open Questions
 
-1. Do we want to discuss Tag's integration with TextInput and TextArea? or should I remove it from API decision altogether?
-2. `<Tag />` or `<Tags />` 😅
+- Do we want to discuss Tag's integration with TextInput and TextArea? or should I remove it from API decision altogether?
+  - We can discuss here
+- ~`<Tag />` or `<Tags />` 😅~ `<Tag />`
+- [`tagsSlot` vs `tagSlot` vs `valuePrefix` vs `leading` on Input components](#alternate-approach-value-prop)
--- a/packages/blade/src/components/Tag/_decisions/decisions.md
+++ b/packages/blade/src/components/Tag/_decisions/decisions.md
@@ -35,8 +35,10 @@ type TagProps = {
 
   /**
    * Callback when close icon on Tag is clicked
+   *
+   * Gives a `value` argument which has same text as `children`
    */
-  onDismiss?: () => void;
+  onDismiss?: ({ value }: { value: TagProps['children'] }) => void;
 
   /**
    * Text that renders inside Tag
@@ -124,7 +126,7 @@ function App() {
   return (
     <TextInput
       tagsSlot={tags.map((tagName, index) => (
-        <Tag onDismiss={() => removeTag(tagName)}>{tagName}</Tag>
+        <Tag onDismiss={({ value }) => removeTag(value)}>{tagName}</Tag>
       ))}
       value={inputValue}
       onChange={({ value }) => setInputValue(value)}
--- a/packages/blade/src/components/Tag/_decisions/decisions.md
+++ b/packages/blade/src/components/Tag/_decisions/decisions.md
@@ -72,7 +72,6 @@ E.g. This is AutoComplete from Primer where the tags go into a slot and input ge
 <img width="419" alt="image" src="https://github.com/razorpay/blade/assets/30949385/be01cdfb-1171-4970-b381-76e4c44d1854">
 
 </details>
-<br/>
 
 ### Integration with SelectInput and AutoComplete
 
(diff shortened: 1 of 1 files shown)


kamleshchandnani:
>Do we keep TextInput and TextArea changes in scope of this API decision? or do we remove it from here? (context: we're not planning to scope it under implementation)

keep it here only. its better to co-locate all the applications we are going to provide for tags and easier for consumers to review

kamleshchandnani:
what about use cases of making the whole tag clickable like a toggle? had mentioned the use cases to Abhishek, need to close on that.

saurabhdaware:
> what about use cases of making the whole tag clickable like a toggle? had mentioned the use cases to Abhishek, need to close on that.

yup I think we're still waiting on that decision of Pills and Tags merge. So far only close button is clickable.

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Integration with SelectInput and AutoComplete
+
+SelectInput and AutoComplete are not expected to have any changes in consumer API as rendering of tags will be handled internally based on the selected value.
+
+### Integration with TextInput and TextArea
+
+> **Warning**
+>
+> Open Question: Do we want to discuss this right now? or do I remove it from API decision itself?
```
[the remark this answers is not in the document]
saurabhdaware: So the thing is implementation detail is relevant here. This way of implementation is what becomes a con in alternate API I had proposed. Because the div that has tags is separate from actual input, we can't use `value` prop for adding tags as well.

https://github.com/razorpay/blade/blob/8e47a3da1fc05263aaafa9cd97978133e10bfdf7/packages/blade/src/components/Tag/_decisions/decisions.md?plain=1#L168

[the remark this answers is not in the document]
saurabhdaware: I have anyway added it inside `<details>` tag so it won't take up too much space while reading API spec

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,183 @@
...
+_Didn't consider any alternate API. Kept `size` and `children` similar to `<Badge />` and `onDismiss` similar to `<Alert />` and `<Dropdown />`_
+
+## Accessibility
+
+- Ensure Close Icon is focussable
+- Close Icon to have appropriate `aria-label`
+
+## Integration with Inputs
+
+<img width="358" alt="image" src="https://github.com/razorpay/blade/assets/30949385/71cd236a-93e7-46f0-9da6-12d519b958d4">
```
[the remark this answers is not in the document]
saurabhdaware: I guess its because text height thing that Abhishek had showed in BaseHeader. I'll see how it looks in code while implementing.

sandeshdamkondwar:
can we add support for outline?

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,198 @@
...
+    setInputValue('');
+  }
+
+  const removeTag = (tagName) => {
+    setTags(tags.filter(tagNameValue) => tagNameValue !== tagName);
+  }
+
+  return (
+    <TextInput
+      tagsSlot={tags.map((tagName, index) => (
```
anuraghazra, line 128: Though now the prop name is somewhat tied to a specific component "tag", if we ever have future usecase of rendering anything different in here might not look correct. tagsSlot={badge|indicator etc}

saurabhdaware: Yup. Although can't see such usecases coming 🤔

saurabhdaware:
> can we add support for outline?

@sandeshdamkondwar Can you elaborate more? you mean have some border in design of Tag? or support custom `outline` css property?

There will be focus state on Tag by default which will add focus ring around it if that is what you meant.

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,198 @@
...
+        <Tag onDismiss={({ value }) => removeTag(value)}>{tagName}</Tag>
```
divyanshu013, line 129: Should we also think about some `id` based prop? Eg. for handling duplicate tags (having same names)

divyanshu013: We'll also need `key` prop here

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,198 @@
...
+   * Callback when close icon on Tag is clicked
+   *
+   * Gives a `value` argument which has same text as `children`
+   */
+  onDismiss?: ({ value }: { value: TagProps['children'] }) => void;
+
+  /**
+   * Text that renders inside Tag
+   */
+  children: ReactText;
```
divyanshu013, line 46: Slight difference, we pass the text value as `children` but the `onDismiss` callback receives `value` as a parameter. Should we consider `value` prop instead?

On a related note there are sometimes cases where people might want a different label and value so maybe we do need a `value` prop for such cases (eg. combining with text inputs) 🤔 

```tsx
<Tag value="red">Red color</Tag> // --> shows as Red color in the UI but `red` for form submissions, etc.

<Tag value="red">Red</Tag> // --> just capitalizing 
```

divyanshu013 (commented):
Looks good overall, some queries

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: `key` is automatically supported in react and typescript. It doesn't seem to throw any error.

By `id` you meant something like `value` as you mentioned in the comment below 🤔? Makes sense. Should it be optional?

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: I think we can support `value` as optional parameter. When value doesn't exist, the default value will just be the children string in  `value` of `onDismiss` 🤔

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,198 @@
...
+    setTags([...tags, inputValue]);
+    setInputValue('');
+  }
+
+  const removeTag = (tagName) => {
+    setTags(tags.filter(tagNameValue) => tagNameValue !== tagName);
+  }
+
+  return (
+    <TextInput
```
chaitanyadeorukhkar, line 127: How about adding tagsSlot on BaseTextInput only and using that for our internal components? This way we can avoid exposing this to consumers

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,198 @@
...
+
+- [react-select](https://react-select.com/home#getting-started)
+- [`TagsInput` component by PluralSight Design System](https://design-system.pluralsight.com/components/tagsinput)
+- [`tokens` prop by Primer AutoComplete.Input](https://primer.style/react/Autocomplete/)
+- [Carbon just putting everything outside of Inputs](https://carbondesignsystem.com/components/tag/usage)
+- [`elemBeforeInput` prop on Atlassian TextField](https://atlassian.design/components/textfield/examples#elements-before-and-after-input)
+
+## Open Questions
+
+- Do we want to discuss Tag's integration with TextInput and TextArea? or should I remove it from API decision altogether?
```
chaitanyadeorukhkar, line 195: If we add it on BaseTextInput only then its an implementation detail and we can remove it from here

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,198 @@
...
+  - We can discuss here
+- ~`<Tag />` or `<Tags />` 😅~ `<Tag />`
+- [`tagsSlot` vs `tagSlot` vs `valuePrefix` vs `leading` on Input components](#alternate-prop-name)
```
chaitanyadeorukhkar, line 198: If we're adding it to consumer-facing TextInput, we could name it `_tagsSlot` and throw warnings if a consumer tries to use it

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: It won't be just on BaseTextInput. It will also be publically available on TextInput and TextArea eventually since we have usecases where we want to enter email ids on TextArea which become tags as we type

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: It should be allowed for consumers as well. You can check the TextArea example here - https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=29212%3A571254&mode=dev

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: We might have to expose to our consumers - https://github.com/razorpay/blade/pull/1315#discussion_r1247724657

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: Gave this a thought and we might not need this because if someone has unique tags, they can also have unique ids on their end and pass that id instead of value returned from `onDismiss`.


E.g.
```jsx
const tags = [{ id: 'red-a', name: 'Red' }, { id: 'red-b', name: 'Red' }]

tags.map(([tagId, tagName], index) => (
     <Tag onDismiss={() => removeTag(tagId)}>{tagName}</Tag>
)
```

Initially I had thought its fine to not even have `value` argument from onDismiss but just added it incase someone wants to do this -`onDismiss={dismissHandler}` (then they won't have to pass the tagName separately)

Should we name argument to something like this? - 
1. tagName
2. name

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: Added more thoughts here - https://github.com/razorpay/blade/pull/1315#discussion_r1247728513

Regarding form submission, I think that won't be the case since tags are not input elements so they are irrelevant to forms (one of the reasons why I wanted to avoid `value` term)

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Integration with Inputs
+
+<img width="358" alt="image" src="https://github.com/razorpay/blade/assets/30949385/71cd236a-93e7-46f0-9da6-12d519b958d4">
+
+Tags can be added inside other input components such as -
+
+- TextArea
+- SelectInput (in Dropdown)
+- TextInput
```
anshulsahni, line 71: Why do we need Tags in `<TextInput/>`, I'm wondering what is the use case for using tags inside text input

packages/blade/src/components/Tag/_decisions/decisions.md
anshulsahni: Why are we even going `children` way if JSX is not allowed as children? Generally, the convention is that "children" is used where we have the freedom to render anything, since only text is being allowed I guess passing that via children would be confusing. I agreed with @saurabhdaware that having value since it's not a direct input component won't be a good idea, we can use text prop.

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,208 @@
...
+   * Decides the size of Tag
+   *
+   * @default medium
+   */
+  size?: 'medium' | 'large';
+
+  /**
+   * Any icon from Blade
+   */
+  icon?: IconComponent;
```
anshulsahni, line 40: No option of icon position?

packages/blade/src/components/Tag/_decisions/decisions.md
anshulsahni, line 1: Don't want to step into design decisions, but some options of changing the background colour and font colour should be allowed, right now I don't have a use case, but while using in landing pages, this use case might appear where a different background is required.

anshulsahni, line 1: I saw in the Figma design, constraints section that while using in a group like filters tags of different sizes shouldn't be used; are we planning to have a TagGroup component in future so those constraints can be enforced using that component?

packages/blade/src/components/Tag/_decisions/decisions.md
divyanshu013: I think `value` should suffice for the usecase I'd in mind.

packages/blade/src/components/Tag/_decisions/decisions.md
divyanshu013: I think `value` should be good enough, `key` / `id` based things can be skipped 👍

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: No. Tag has close icon button on trailing so any icon that is added on tag will always on leading part only.

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: Didn't go with TagGroup way because anyway I would people to loop over their tag data and render tags. Something like -
```js
tagsArray.map((tagName) => <Tag>{tagName}</Tag>)
```

So people by default will end up writing standard Tag sizes. It will also visibly look odd to have tags of different sizes side-by-side so it will be caught during development or during design review of page.

saurabhdaware: I think adding `variant`, `intent`, or `backgroundColor` and `color` won't be a breaking change in the API we have here so if such usecase comes, it will be possible to support that without any migration changes.

Most intent usecases we saw were actually Badge usecases and did not require removable Tags as such. Can explore this if new usecase comes.

packages/blade/src/components/Tag/_decisions/decisions.md
anshulsahni: We shouldn't have two ways of passing the same data, for such small & simple use cases, if a prop can suffice our use case then no point in supporting children here

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: So I have added image in Decision doc which has usecase of adding tags inside TextArea (Adding email id tags). Same usecase would also apply for TextInput where we don't want to show things on multiple lines

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: > Why are we even going children way if JSX is not allowed as children? Generally, the convention is that "children" is used where we have the freedom to render anything, since only text is being allowed I guess passing that via children would be confusing

Fair point. We use children because in cases like Badge, Tags, the text is semantically inside the visual part and it doesn't have a separate description or other components inside of it.

Don't really have much of a strong opinion though. Agree with your reasoning as well that children is usually more flexible. Our children btw is not string but ReactText so I do feel there are some cases where the code might get easier. E.g. -

```jsx
<Badge>{someBadgePrefixTitle} - {anotherBadgeTitle}</Badge>
// vs
<Badge title={`${someBadgePrefixTitle} - ${anotherBadgeTitle}`} />

```

Either ways we will have to keep it children to be consistent with Badge. Can look into changing it on all components from children to some other jsx prop if this is causing confusion?

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: Continued this discussion on slack. Summary was we can go without value prop because consumers can handle this on their own with something like this -
```jsx
<Tag onDismiss={() => doSomething('sm-1')}>Small</Tag>
<Tag onDismiss={() => doSomething('sm-2')}>Small</Tag>
```

This is also the approach that other design systems have
- [Carbon passes no argument from onDismiss param](https://react.carbondesignsystem.com/?path=/story/components-tag--playground&args=filter:!true)
- [Atlassian passes text prop which is the same prop that they use for display value as well](https://atlassian.design/components/tag/code)

We decided we can asses value prop later.

divyanshu013 (dismissed):
Looks good to me. We can keep the `value` prop assessment for later.

packages/blade/src/components/Tag/_decisions/decisions.md
```diff
@@ -0,0 +1,208 @@
...
+- [Carbon just putting everything outside of Inputs](https://carbondesignsystem.com/components/tag/usage)
+- [`elemBeforeInput` prop on Atlassian TextField](https://atlassian.design/components/textfield/examples#elements-before-and-after-input)
+
+## Open Questions
+
+- Do we want to discuss Tag's integration with TextInput and TextArea? or should I remove it from API decision altogether?
+  - We can discuss here
+- ~`<Tag />` or `<Tags />` 😅~ `<Tag />`
+- [`tagsSlot` vs `tagSlot` vs `valuePrefix` vs `leading` on Input components](#alternate-prop-name)
+- Should we support `value` prop that allows us to set unique identifier value for Tag?
```
kamleshchandnani, line 205: since tags are individual components what is the use case of passing the value prop to onDismiss?

packages/blade/src/components/Tag/_decisions/decisions.md
kamleshchandnani: what if we just call it as tags(that accepts array of tag)? since its literally that?

kamleshchandnani (dismissed):
overall lgtm

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: it can make writing custom handlers a bit simpler in some cases. `onDismiss={dismissHandler}` vs `onDismiss={() => dismissHandler(tagName)}`

saurabhdaware: Not a strong opinion though. This is what atlassian follows

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: By array of tags you mean something like `tags={['apple', 'banana']}`? or `tags={[<Tag>apple</Tag>, <Tag>banana</Tag>]}`

We can't do 1st one because we need `onDismiss` of Tag exposed as well so that closing it can remove that tag.

Used `tagsSlot` communicates that it's a slot where Tag components can be passed. But I think `tags={<Tag>apple</Tag><Tag>banana</Tag>}` also makes sense.

kamleshchandnani: meant this the prop name `tags` and the value as `tags={[<Tag>apple</Tag>, <Tag>banana</Tag>]}`

wouldn't this `tags={<Tag>apple</Tag><Tag>banana</Tag>}` be more difficult to type? like people misusing the jsx thing? with array of jsx you can at least type what typeof array elements we are expecting.

packages/blade/src/components/Tag/_decisions/decisions.md
kamleshchandnani: hmm, tbh value on non-input thing already makes it difficult to understand. might rather go ahead and do `{() => dismissHandler(tagName)}` because this seems more natural and I can do whatever data I want for this func. 🤔

packages/blade/src/components/Tag/_decisions/decisions.md
saurabhdaware: `tags` prop name makes sense. For TS, I was basically doing this `tags={<><Tag>apple</Tag><Tag>banana</Tag></>}`. I think what you mentioned can be done as well. We'll get to add some spacing between array items while rendering with array syntax. With jsx we might have to loop over with React.Children.

kamleshchandnani (approved):

divyanshu013 (approved):
:shipit:

PR #1319, opened 2023-06-18
fix: support `min-content`, `none`, `fit-content` on SpacingValueType

### LINKED ISSUE #1314: Box: Support "initial" and "fit-content" for width
## Problem

I'm working on implementing this design:

Screenshot 2023-06-16 at 2 53 14 PM

Which becomes this on mobile:

Screenshot 2023-06-16 at 2 55 01 PM

We want the last item to be on a separate line. Putting `display: block` on it messes up the alignment of the icon and text since their container has `display: flex`.

Code with the issue:
https://github.com/razorpay/frontend-website/pull/2194/files#diff-1f2c8c5821f57c8376ada25679cfc0b957d02788758ba99df29d8ca16b1599b5R96

## Suggestion

I could technically put another wrapper around the wrapper containing the icon and text and put a responsive `display: block` on that. But assuming the goal of the Layout primitives was to avoid *wrapocalypse* like this one.

Please support the "initial" and/or "fit-content" values for width on Box component. So we could write this:

```
<Box width={{ base: '100%', l: 'initial' }}>
```

This will render the correct layout:

Screenshot 2023-06-16 at 3 01 41 PM

archie252000:
Fixes #1314


packages/blade/src/components/Box/BaseBox/types/spacingTypes.ts
```diff
@@ -11,7 +11,9 @@ type SpacingValueType =
   | `min(${string})`
   | `max(${string})`
   | 'auto'
-  | 'none';
+  | 'none'
+  | 'initial'
+  | 'fit-content';
```
saurabhdaware, line 16: https://developer.mozilla.org/en-US/docs/Web/CSS/width#values

can we add `max-content` and `min-content` as well in this PR itself?

saurabhdaware (dismissed):

packages/blade/src/components/Box/BaseBox/types/spacingTypes.ts
archie252000: sure

packages/blade/src/components/Box/BaseBox/types/spacingTypes.ts
```diff
@@ -11,8 +11,13 @@ type SpacingValueType =
   | 'auto'
   | `min(${string})`
   | `max(${string})`
+  | 'auto'
```
saurabhdaware, line 14:
```suggestion
```

auto is already there in the list

saurabhdaware (approved):
Thank you for another helpful PR 👯 🥳

saurabhdaware (approved):
I committed wrong file to this branch while fixing prettier 🤦🏼 fixed it now. Thanks for the PR again 👯

archie252000:
@saurabhdaware you are welcome!

PR #1336, opened 2023-06-21
feat(Collapsible) - add native implementation

divyanshu013:
- This PR completes the Collapsible native implementation (Accordion isn't complete yet)
- Includes some refactors which were common for web and native
- Screen recording on ios and android below

https://github.com/razorpay/blade/assets/6682655/5c42c68a-23c7-43cf-8226-17fa20654aa9

https://github.com/razorpay/blade/assets/6682655/964b7bf2-621d-4e62-b8f6-a9227718fe8f

### DIFF
--- a/packages/blade/src/components/Accordion/AccordionButton.native.tsx
+++ b/packages/blade/src/components/Accordion/AccordionButton.native.tsx
@@ -2,9 +2,10 @@ import type { ReactElement } from 'react';
 import { StyledAccordionButton } from './StyledAccordionButton';
 import type { AccordionButtonProps } from './types';
 import { Heading } from '~components/Typography';
+import { MetaConstants, assignWithoutSideEffects } from '~utils';
 
 // TODO: implement with reanimated, pressable
-const AccordionButton = ({ children }: AccordionButtonProps): ReactElement => {
+const _AccordionButton = ({ children }: AccordionButtonProps): ReactElement => {
   return (
     <StyledAccordionButton
       // TODO: add logic
@@ -14,5 +15,9 @@ const AccordionButton = ({ children }: AccordionButtonProps): ReactElement => {
     </StyledAccordionButton>
   );
 };
+
+const AccordionButton = assignWithoutSideEffects(_AccordionButton, {
+  componentId: MetaConstants.AccordionButton,
+});
 
 export { AccordionButton };
--- a/packages/blade/src/components/Accordion/AccordionButton.native.tsx
+++ b/packages/blade/src/components/Accordion/AccordionButton.native.tsx
@@ -1,15 +1,14 @@
 import type { ReactElement } from 'react';
 import { StyledAccordionButton } from './StyledAccordionButton.native';
 import type { AccordionButtonProps } from './types';
-import { useAccordion } from './AccordionContext';
 import { Heading } from '~components/Typography';
 import { MetaConstants, assignWithoutSideEffects } from '~utils';
 import { useCollapsible } from '~components/Collapsible/CollapsibleContext';
 
 // TODO: implement with reanimated, pressable
 const _AccordionButton = ({ children }: AccordionButtonProps): ReactElement => {
-  const { onExpandChange, isExpanded, collapsibleBodyId } = useCollapsible();
-  const { showNumberPrefix, expandedIndex } = useAccordion();
+  const { onExpandChange, isExpanded } = useCollapsible();
+  // const { showNumberPrefix, expandedIndex } = useAccordion();
 
   return (
     <StyledAccordionButton
--- a/packages/blade/src/components/Accordion/AccordionButton.native.tsx
+++ b/packages/blade/src/components/Accordion/AccordionButton.native.tsx
@@ -1,15 +1,21 @@
 import type { ReactElement } from 'react';
-import { StyledAccordionButton } from './StyledAccordionButton';
+import { StyledAccordionButton } from './StyledAccordionButton.native';
 import type { AccordionButtonProps } from './types';
+import { useAccordion } from './AccordionContext';
 import { Heading } from '~components/Typography';
 import { MetaConstants, assignWithoutSideEffects } from '~utils';
+import { useCollapsible } from '~components/Collapsible/CollapsibleContext';
 
 // TODO: implement with reanimated, pressable
 const _AccordionButton = ({ children }: AccordionButtonProps): ReactElement => {
+  const { onExpandChange, isExpanded, collapsibleBodyId } = useCollapsible();
+  const { showNumberPrefix, expandedIndex } = useAccordion();
+
   return (
     <StyledAccordionButton
       // TODO: add logic
-      isExpanded={false}
+      isExpanded={isExpanded}
+      onPress={() => onExpandChange(!isExpanded)}
     >
       <Heading size="small">{children}</Heading>
     </StyledAccordionButton>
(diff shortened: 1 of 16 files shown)


packages/blade/src/components/Collapsible/CollapsibleBodyContent.native.tsx
```diff
@@ -1,13 +1,121 @@
...
+} from './commonStyles';
+import { nativeStyles } from './styles.native';
+import { Box } from '~components/Box';
+import { useTheme } from '~components/BladeProvider';
+import { castNativeType } from '~utils';
 
-// TODO: implement
-type CollapsibleBodyContentProps = {
-  children: ReactNode;
+type AnimatedStyledCollapsibleBodyContentProps = {
```
anuraghazra, line 25: longer than Java class names 💀

packages/blade/src/components/Collapsible/CollapsibleBodyContent.native.tsx
```diff
@@ -1,13 +1,121 @@
...
+    };
+  });
+
+  /**
+   * Tracks the height of content so we can animate height to and from 0 to the content's height.
+   * **Note:** We can't animate height from 0 to auto or vice-versa.
+   */
+  const onLayout: (event: LayoutChangeEvent) => void = useCallback(
+    (event) => {
+      if (event.nativeEvent.layout.height > layoutHeight) {
```
anuraghazra, line 85: Do we need this condition? Like onLayout will always run each time a layout change happens either when height increases or decreases. 

If we put this condition and imagine inside an accordion there is some sort of UI which can be removed or dismissed then the height should decrease but we will skip saving the height.

packages/blade/src/components/Collapsible/CollapsibleChevronIcon.native.tsx
```diff
@@ -0,0 +1,39 @@
...
+  });
+  const duration = castNativeType(getTransitionDuration(theme));
+  const easing = castNativeType(getTransitionEasing(theme));
+
+  const rotateZ = useDerivedValue(() =>
+    withTiming(isExpanded ? transformExpanded : transformCollapsed, { duration, easing }),
+  );
+  const animatedStyles = useAnimatedStyle(() => ({
+    // Interpolated string has to be passed, to make it work for both android and ios
+    transform: [{ rotateZ: `${rotateZ.value}deg` }],
```
anuraghazra, line 29: interesting, this is documented somewhere on the Reanimated docs? or did you stumble upon it randomly?

packages/blade/src/components/Collapsible/CollapsibleBodyContent.native.tsx
divyanshu013: Right, had left figuring this part out for Accordion implementation, modified the logic correctly in that.

packages/blade/src/components/Collapsible/CollapsibleChevronIcon.native.tsx
divyanshu013: Random searches and lot of pulling my hair 💀

packages/blade/src/components/Accordion/AccordionButton.native.tsx
```diff
@@ -1,18 +1,28 @@
 import type { ReactElement } from 'react';
-import { StyledAccordionButton } from './StyledAccordionButton';
+import { StyledAccordionButton } from './StyledAccordionButton.native';
```
chaitanyadeorukhkar, line 2:
```suggestion
import { StyledAccordionButton } from './StyledAccordionButton';
```

packages/blade/src/components/Collapsible/CollapsibleBodyContent.native.tsx
```diff
@@ -1,13 +1,121 @@
...
+    opacity.value = withTiming(getOpacity({ isExpanded }), { duration, easing });
+    height.value = withTiming(
+      // Animates the height to the measured value
+      isExpanded && layoutHeight ? layoutHeight : 0,
+      { duration, easing },
+      (isComplete) => {
+        // Only run this if the animation ran uninterrupted, for eg collapsing the content before it expanded fully
+        if (isComplete) {
+          // The callback `onAnimationComplete` has to be declared outside this, on JS thread
+          runOnJS(onAnimationComplete)();
```
chaitanyadeorukhkar, line 66: Nice!

packages/blade/src/components/Collapsible/CollapsibleBodyContent.native.tsx
```diff
@@ -1,13 +1,121 @@
...
+   */
+  const onLayout: (event: LayoutChangeEvent) => void = useCallback(
+    (event) => {
+      if (event.nativeEvent.layout.height > layoutHeight) {
+        /**
+         * We set `layoutHeight` if the native event's layout height is larger.
+         *
+         * The greater than comparison is needed because sometimes the native event's layout height is smaller than actual content height 🤯
+         * For example, this happens if you try to render a lengthy `Text` that wraps onto multiple lines.
+         * In this case the initial native event's layout height only counts height of `Text` as if it were single line.
```
chaitanyadeorukhkar, line 91: TIL

packages/blade/src/components/Collapsible/CollapsibleChevronIcon.native.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { isExpanded, direction } = useCollapsible();
+  const { theme } = useTheme();
+
+  const { transformExpanded, transformCollapsed } = getCollapsibleChevronIconTransforms({
+    direction,
+  });
+  const duration = castNativeType(getTransitionDuration(theme));
+  const easing = castNativeType(getTransitionEasing(theme));
+
+  const rotateZ = useDerivedValue(() =>
```
chaitanyadeorukhkar, line 24: Why are we using useDerivedValue here? No dependencies for useDerivedValue needed here?

packages/blade/src/components/Collapsible/CollapsibleChevronIcon.native.tsx
chaitanyadeorukhkar: Didn't get this? Anything different here? We're doing same for Spinner as well, right?

chaitanyadeorukhkar (approved):
:shipit:

packages/blade/src/components/Collapsible/CollapsibleChevronIcon.native.tsx
divyanshu013: Had to interpolate a string based on this value, don't need dependencies (this is being derived from `isExpanded`)

packages/blade/src/components/Accordion/AccordionButton.native.tsx
divyanshu013: It wasn't picking the typing correctly (i.e. `Pressable`) 😅 

So I changed it to `.native` explicitly because we're inside a native module

packages/blade/src/components/Collapsible/CollapsibleChevronIcon.native.tsx
divyanshu013: I think I might have been passing `rotateZ.value` directly earlier, lemme get rid of the comment. Weirdly didn't see any TS errors that time but now it's correctly throwing TS errors lol

divyanshu013: Getting rid of the comment in future PR

PR #1339, opened 2023-06-22
fix(dropdown): dynamic positioning of dropdown overlay

Jaynil1611:
Fixes: #1245 ,  #1344

Previous PR: #1284 
This PR is an enhancement on the previous PR which includes dynamic positioning of dropdown overlay depending on the height & width of the children rendered inside it.


saurabhdaware:
#1344 will also get resolved with this right? it seems to be fixed on storybook

Jaynil1611:
> #1344 will also get resolved with this right? it seems to be fixed on storybook

Yes, have added this issue in PR description

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -42,18 +42,20 @@ to {
...
       transform: translateY(${makeSize(props.theme.spacing[3])});
       opacity: 0;
       z-index: 99;
+      pointer-events: ${props.isOpen ? 'all' : 'none'};
     `,
 );
 
 type DropdownOverlayProps = {
-  children: React.ReactNode;
+  children: React.ReactElement;
```
saurabhdaware, line 58: is this same as ReactNode 🤔

saurabhdaware (dismissed):
Looks good to me. Big contributions! Thank you so much 🥳

packages/blade/src/components/Dropdown/docs/stories.ts
```diff
@@ -539,8 +539,6 @@ const WithControlledSelectStory = `
             <ActionList>
               <ActionListItem title="Mumbai" value="mumbai" />
               <ActionListItem title="Bangalore" value="bangalore" />
-              <ActionListItem title="Pune" value="pune" />
-              <ActionListItem title="Chennai" value="chennai" />
```
saurabhdaware, line 543: unrelated change?

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
Jaynil1611: Nope, ReactNode is a superset that includes ReactElement. [More Info](https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement)

packages/blade/src/components/Dropdown/docs/stories.ts
Jaynil1611: @saurabhdaware There is not enough space to render all options which flips the position of the dropdown in controlled select [story](https://61c19ee8d3d282003ac1d81c-fanhpizsbk.chromatic.com/?path=/docs/components-dropdown-dropdown--page).
image

saurabhdaware: oh got it. btw, when there is no enough space on top and bottom both, I think we should prefer opening it in bottom only right?

Jaynil1611: Yeah, I have made the change. Please review once

saurabhdaware (approved):

PR #1346, opened 2023-06-23
feat(blade): added bottomsheet body zero padding option

anuraghazra:
Added padding.0 option in BottomSheetBody.

When the padding is set at zero the header grab handle becomes floating so that images and other elements can overlap it. 

react-native: 

image

Web: 

image


packages/blade/src/components/BottomSheet/BottomSheet.native.tsx
```diff
@@ -168,6 +170,10 @@ const _BottomSheet = ({
   const renderHandle = React.useCallback((): React.ReactElement => {
```
chaitanyadeorukhkar, line 170: Why was this inside a useCallback? Can't it be just a component as is?

anuraghazra: Gorhom's render handles and other slots only accepts React.FunctionElement. 

I could've just done `handleComponent={BottomSheetHandle}` but i also need to pass additional props to the components like zIndex, isHeaderEmpty, setHeaderHeight etc thus using a render function. 

I cannot also do this: 
```
            handleComponent={
              <Handle
                isHeaderEmpty={isHeaderEmpty}
                zIndex={zIndex}
                // header={header}
                setHeaderHeight={setHeaderHeight}
              />
            }
```

it will throw "expect a string or class/function but got object"

packages/blade/src/components/BottomSheet/BottomSheetHeader.native.tsx
```diff
@@ -17,11 +17,20 @@ const _BottomSheetHeader = ({
...
 
+  React.useEffect(() => {
+    setIsHeaderEmpty(isHeaderEmpty);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isHeaderEmpty]);
+
   return (
-    <BaseBox backgroundColor="white" overflow="visible" flexShrink={0}>
+    <BaseBox
+      backgroundColor={isHeaderEmpty ? undefined : 'white'}
```
chaitanyadeorukhkar, line 30: Why are we hardcoded white here? Should follow theme right?

chaitanyadeorukhkar (dismissed):

packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
```diff
@@ -82,6 +83,8 @@ const _BottomSheet = ({
   const [headerHeight, setHeaderHeight] = React.useState(0);
   const [footerHeight, setFooterHeight] = React.useState(0);
   const [grabHandleHeight, setGrabHandleHeight] = React.useState(0);
+  const [hasBodyPadding, setHasBodyPadding] = React.useState(true);
+  const [isHeaderEmpty, setIsHeaderEmpty] = React.useState(false);
```
saurabhdaware, line 87: BottomSheet.web and BottomSheet.native seems to share a lot of state code. Should we move this out into a custom hook?

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
```diff
@@ -85,6 +85,24 @@ describe('<BottomSheet />', () => {
     mockConsoleError.mockRestore();
   });
 
+  it('should render empty header with padding 0', () => {
+    const mockConsoleError = jest.spyOn(console, 'error').mockImplementation();
```
saurabhdaware, line 89: expected to mock console.error here?

packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
anuraghazra: On the surface, yes - it does share few bits of state code and logic. 

But in practice how we are using these states and the logic between native & web implementation has subtle differences which if we extract out will cause more overhead and introduce `ifocalypse`. 

Discussed this earlier with chaitanya also, it's better to keep things separated out for now looking at the complexity of the two platforms.

packages/blade/src/components/BottomSheet/__tests__/BottomSheet.web.test.tsx
anuraghazra: Removed.

saurabhdaware (approved):
:shipit:

.changeset/many-parrots-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
```
saurabhdaware, line 2:
```suggestion
"@razorpay/blade": minor
```

🤔 ?

saurabhdaware (approved):

Issue #812, opened 2022-11-03
Should subheading render as h6?


saurabhdaware:
```html
<Title>Main Heading</Title>
<Heading variant="subheading">Something something</Heading>
<Text>Hi there!!</Text>
```

will render as 

```html
<h1>Main Heading</h1>
<h6>Something something</h6>
<p>Hi there!!</p>
```

is this expected semantics for accessibility? 🤔 shouldn't subheading not be an `h*` tag?

<img width="814" alt="image" src="https://user-images.githubusercontent.com/30949385/199714631-059a9bfc-a2cf-4741-8433-18a7e36fd9fc.png">

// cc @anuraghazra

divyanshu013:
MDN recommends [not skipping heading levels](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#usage_notes) so we might need to revisit later. Maybe an `as` prop support?

saurabhdaware:
React Docs has one nice approach in their example where they create `Section` component and based on the nesting of Section, they decide heading level.

<img width="1078" alt="image" src="https://user-images.githubusercontent.com/30949385/236428384-7346c6dd-74c7-4c83-9097-c736e7d10935.png">

https://react.dev/reference/react/useContext#examples

divyanshu013:
Not sure if automatically determining the h level is good idea but the approach is nice _(could probably be utilized for certain validations if needed)_

divyanshu013:
Discovered another possible usecase of `<Heading as="span">` while working on Accordion:

> need to render a heading inside a button element (accordion header) but apparently heading inside button is invalid (but button inside heading is valid)

![image](https://github.com/razorpay/blade/assets/6682655/3ebf3f62-fd56-4f02-aedd-c7495cf89e8c)

cseas:
All Heading and Title components should support the `"as"` prop with at least h1 to h4.

Posting some examples of heading hierarchy violations from recent designs:

<img width="1220" alt="Screenshot 2023-06-12 at 1 38 10 AM" src="https://github.com/razorpay/blade/assets/29686866/56a7845c-a033-4c17-b347-e2eaa6c1714e">

<img width="837" alt="Screenshot 2023-06-12 at 1 38 50 AM" src="https://github.com/razorpay/blade/assets/29686866/9bd2e539-5064-45f8-81e8-9e1955728487">

[...]

PR #1360, opened 2023-06-26
test(Accordion/Collapsible): add tests

divyanshu013:
- Also bumps up `jest-native` version which includes extra matchers
- Update the markup to remove earlier hacks because https://github.com/razorpay/blade/issues/812 is now resolved

> **Note**
>
> Large diff showing because of snapshots 😅


packages/blade/src/components/Accordion/AccordionButton.web.tsx
```diff
@@ -14,16 +14,10 @@ const _AccordionButton = ({ index, icon: Icon, children }: AccordionButtonProps)
 
   const toggleCollapse = (): void => onExpandChange(!isExpanded);
   const onClick = (): void => toggleCollapse();
-  const onKeyDown: KeyboardEventHandler<HTMLDivElement> = (event) => {
-    const SPACE_KEY = ' ';
-    if (event.key === SPACE_KEY || event.key === 'Enter') {
```
anuraghazra, line 19: Noice.

packages/blade/src/components/Accordion/StyledAccordionButton.web.tsx
```diff
@@ -8,7 +8,7 @@ import {
 } from './commonStyles';
 import { castWebType } from '~utils';
 
-const StyledAccordionButton = styled.div<StyledAccordionButtonProps>((props) => {
+const StyledAccordionButton = styled.button<StyledAccordionButtonProps>((props) => {
```
anuraghazra, line 11: We can't use BaseButton? Is this because you want the markdown to be button but don't want the styles?

divyanshu013: Yup exactly, the styles are quite different from BaseButton's

packages/blade/src/components/Accordion/AccordionButton.web.tsx
divyanshu013: `useThePlatform` 😸

packages/blade/src/components/Accordion/__tests__/Accordion.native.test.tsx
```diff
@@ -0,0 +1,195 @@
+import { fireEvent, waitFor } from '@testing-library/react-native';
+import { useState } from 'react';
+import { Accordion } from '../Accordion';
+import { AccordionItem } from '../AccordionItem';
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.native';
+import { Button } from '~components/Button';
+
+describe('<Accordion />', () => {
+  // Accordion uses Collapsible which uses animations and requestAnimationFrame which makes RN tests throw warnings
```
anuraghazra, line 9: 🤔 So does js automatically mocks requestAnimationFrame in RN?

packages/blade/src/components/Accordion/__tests__/Accordion.native.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+    );
+
+    // Clicking first accordion item
+    const trigger1 = getAllByRole('button')[0];
+    expect(trigger1).toHaveAccessibilityState({ expanded: false });
+    expect(getByText(description1)).not.toBeVisible();
+
+    fireEvent.press(trigger1);
+    // wait for animation to finish
+    await waitFor(() => expect(getByText(description1)).toBeVisible());
```
anuraghazra, line 65: we should also assert that description2 should not be visible.

packages/blade/src/components/Accordion/__tests__/Accordion.native.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+    const { getByText, getAllByRole } = renderWithTheme(<ControlledAccordionExample />);
+
+    /**
+     * - Button 1 -> External Button 1
+     * - Button 2 -> External Button 2
+     * - Button 3 -> External Button 3
+     * - Button 4 -> External Button Collapse
+     * - Button 5 -> AccordionItem button 1
+     * - Button 6 -> AccordionItem button 2
```
anuraghazra, line 138: 😅 kind of hard to track via indices, can we save them in variable and reference these?

packages/blade/src/components/Accordion/__tests__/Accordion.web.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+import { AccordionItem } from '../AccordionItem';
+import { Accordion } from '../Accordion';
+import renderWithTheme from '~src/_helpers/testing/renderWithTheme.web';
+import assertAccessible from '~src/_helpers/testing/assertAccessible.web';
+import { Button } from '~components/Button';
+
+describe('<Accordion />', () => {
+  it('should render', () => {
+    const { container } = renderWithTheme(
+      <Accordion>
```
anuraghazra, line 13: question: do we need to expose styled props on root Accordion for layout purposes?

packages/blade/src/components/Accordion/__tests__/Accordion.web.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+    expect(getAllByRole('region')[0]).toHaveAttribute('aria-hidden', 'false');
+
+    expect(onExpandChange).toHaveBeenCalledTimes(1);
+    expect(onExpandChange).toHaveBeenLastCalledWith({ expandedIndex: 0 });
+
+    // Clicking second accordion item
+    const trigger2 = getByRole('button', { name: button2 });
+    expect(trigger2).toHaveAttribute('aria-expanded', 'false');
+    expect(getByText(description2)).not.toBeVisible();
+    expect(getAllByRole('region', { hidden: true })[1]).toHaveAttribute('aria-hidden', 'true');
```
anuraghazra, line 68: getByRole(region) is referring to the accordion description or accordion item?

anuraghazra: If it is AccordionItem shouldn't this be button? 🤔

packages/blade/src/components/Accordion/__tests__/Accordion.native.test.tsx
divyanshu013: `getByText` seems bugged when collapsing back, even if the container has `0` height and `aria-hidden="true"` it still shows the text content as visible. Spent quite a bit of time but couldn't figure out why.

Have added a relevant comment in collapsible tests [here](https://github.com/razorpay/blade/pull/1360/files#diff-777282071cca5e5b49f7bc5a2bba126981123d4d7146480ce9c8b070f1042fa7R126)

packages/blade/src/components/Accordion/__tests__/Accordion.native.test.tsx
divyanshu013: [Yes](https://jestjs.io/docs/jest-object#jestusefaketimersfaketimersconfig) it seems, since we're using JSDOM env

packages/blade/src/components/Accordion/__tests__/Accordion.web.test.tsx
divyanshu013: Right, root Accordion has a top level `BaseBox` which accepts these

packages/blade/src/components/Accordion/__tests__/Accordion.web.test.tsx
divyanshu013: It actually refers to the expanding / collapsing region from user perspective. In code this refers to `CollapsibleBody`

packages/blade/src/components/Accordion/__tests__/Accordion.web.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+        <AccordionItem
+          title="How can I setup QR Codes?"
+          description="Just use Razorpay. You may also check our docs for detailed instructions. Please use the search functionality to ask your queries."
+        />
+        <AccordionItem title="How can I setup Subscriptions?" description="Just use Razorpay." />
+      </Accordion>,
+    );
+
+    await assertAccessible(getAllByRole('button')[0]);
+    await assertAccessible(getAllByRole('region', { hidden: true })[0]);
```
anuraghazra, line 194: Probably be better to assert accessibility on the whole tree instead of single elements, otherwise missing ID bindings and references might not get picked up as errors. 

await assertAccessible(container);

packages/blade/src/components/Collapsible/__tests__/Collapsible.web.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+          </Box>
+        </CollapsibleBody>
+      </Collapsible>,
+    );
+
+    const collapsibleTrigger = getByRole('button');
+    const collapsiblePanel = getByRole('region');
+
+    await assertAccessible(collapsibleTrigger);
+    await assertAccessible(collapsiblePanel);
```
anuraghazra, line 185: same here, assert whole tree.

anuraghazra (approved):
Mostly looks good, minor comments.

packages/blade/src/components/Collapsible/__tests__/Collapsible.web.test.tsx
divyanshu013: Done in https://github.com/razorpay/blade/pull/1360/commits/5362504ae49ec5bbbae577c5543de2e9ba15b3cc

packages/blade/src/components/Accordion/__tests__/Accordion.web.test.tsx
divyanshu013: Done in https://github.com/razorpay/blade/commit/5362504ae49ec5bbbae577c5543de2e9ba15b3cc

packages/blade/src/components/Accordion/__tests__/Accordion.native.test.tsx
divyanshu013: Done in https://github.com/razorpay/blade/commit/5362504ae49ec5bbbae577c5543de2e9ba15b3cc

anuraghazra (approved):

packages/blade/src/components/Accordion/__tests__/Accordion.web.test.tsx
chaitanyadeorukhkar: So Accordion will also have styled props, right? We will have to pass it down to the BaseBox?

chaitanyadeorukhkar (approved):

PR #1361, opened 2023-06-27
feat(Tag): add Tag component

saurabhdaware:
This PR contains
- Tag component
- Documentation
- Tests

What this PR does NOT have 
- Integration with SelectInput or other Inputs (That is here - #1459)


packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
```diff
@@ -175,7 +175,10 @@ const getAllProps = (
     props.borderBottomWidth ||
     props.borderTopWidth ||
     props.borderLeftWidth ||
-    props.borderRightWidth;
+    props.borderRightWidth ||
+    props.borderWidth ||
```
saurabhdaware, line 179: `borderStyle` was not being added when we pass just `borderWidth` or `borderColor`

packages/blade/src/components/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,159 @@
...
+          onSubmit={(e) => {
+            e.preventDefault();
+            addTag();
+          }}
+        >
+          <TextInput
+            label="Tag Label"
+            value={inputValue}
+            onChange={({ value }) => setInputValue(value ?? '')}
+            {...{ onSubmit: isReactNative() ? () => addTag() : undefined }}
```
chaitanyadeorukhkar, line 143: This is allowing me to add empty tags, which shouldn't even render since children prop is mandatory. Can we do an empty check here?

packages/blade/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,99 @@
...
+  icon: Icon,
+  onDismiss,
+  children,
+  isDisabled,
+  testID,
+  ...styledProps
+}: TagProps): React.ReactElement | null => {
+  const [isVisible, setIsVisible] = React.useState(true);
+
+  if (!isVisible) {
```
chaitanyadeorukhkar, line 55: So we handle whether or not to show the UI of tag or null internally. The Tag component itself doesn't really get unmounted, right? If the consumer wanted to show or hide this tag conditionally, there is no way for them to do that? And we have to have faith that the consumer has implemented `onDismiss` handler

Imagine this scenario where we submit `selectedItems` to an API call:
```tsx
const selectedItems = ['Item A', 'Item B', 'Item C']

return selectedItems.map((item) => 
  <Tag 
     onDismiss={()=>removeSlectedItem(item)} // This will remove item from selectedItems array
  >
  {item}
  </Tag>
)
```
Everything is correct here and when the tag is dismissed, the array is changed, but what if the consumer doesn't implement onDismiss?

```tsx
const selectedItems = ['Item A', 'Item B', 'Item C']

return selectedItems.map((item) => 
  <Tag>
  {item}
  </Tag>
)
```
Now, on UI for the end user, selected items will change as and when they dismiss each Item  because we handle that state internally but on code, `selectedItems` array will remain as is and when the API call is made, it will send the full `selectedItems` array which may not be the end-user's intention.


tldr; we shouldn't handle this internally, let consumers handle the visibility of the tag - we can just give them a `onDismiss` callback and they can hide the dismissed item on their end. This will ensure that there will never be a mismatch for the end user. Thoughts?

saurabhdaware: Yupp thanks for pointing out. I actually implemented it this way at the start but then moved to having uncontrolled dismiss to keep it consistent with how we do it in `Alert` right now. In Alert also when we press close icon, we remove Alert and then consumer has to take care of keeping states in sync by calling state update in `onDismiss`

I also saw this is how Ant design and Atlassian design system do as well 
- https://atlassian.design/components/tag/examples#removable
- https://ant.design/components/tag

On the other hand Carbon follows what you're saying
- https://react.carbondesignsystem.com/?path=/story/components-tag--playground&args=filter:!true

I couldn't really find any reasoning for this approach but kept it this way to keep consistency with Ant design and Atlassian's design system.

Would like more opinions on this. Don't really have a strong opinion on either approaches.

chaitanyadeorukhkar: Its okay on Alert since its just a notification to the user and not necessarily something that will impact the user if the onDismiss is not handled by the developer but on Tags it can have a significant impact

packages/blade/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,160 @@
...
+   */
+  children: StringChildrenType;
+
+  /**
+   * Disable tag
+   */
+  isDisabled?: boolean;
+
+  /**
+   *
```
chaitanyadeorukhkar, line 48: Should we mention that this is an internal only tag and shouldn't be used in the jsdoc?

packages/blade/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,160 @@
...
+      flexWrap="nowrap"
+      backgroundColor="brand.gray.a100.lowContrast"
+      borderRadius="max"
+      padding={size === 'medium' ? mediumPadding : largePadding}
+      {...getStyledProps(styledProps)}
+      {...metaAttribute({ name: MetaConstants.Tag, testID })}
+      isTagFocussed={_isTagFocussed}
+    >
+      <Box display="flex" flexDirection="row" flexWrap="nowrap" alignItems="center">
+        <Desktop>
```
chaitanyadeorukhkar, line 145: Maybe use a better component name that's easier to understand? `<ShowOnDesktop>` & `<ShowOnMobile>` maybe?

packages/blade/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,160 @@
...
+  return (
+    <FocussableTag
+      display={(isReactNative() ? 'flex' : 'inline-flex') as never}
+      alignSelf={isReactNative() ? 'center' : undefined}
+      flexDirection="row"
+      flexWrap="nowrap"
+      backgroundColor="brand.gray.a100.lowContrast"
+      borderRadius="max"
+      padding={size === 'medium' ? mediumPadding : largePadding}
+      {...getStyledProps(styledProps)}
```
chaitanyadeorukhkar, line 140: Does adding a display prop break the UI for this? If yes, then we should add a wrapper around it?

chaitanyadeorukhkar (dismissed):

packages/blade/src/components/Tag/Tag.tsx
saurabhdaware: ops yes it was. I removed one internal wrapper and added one wrapper externally

packages/blade/src/components/Tag/Tag.tsx
saurabhdaware: done

packages/blade/src/components/Tag/Tag.tsx
saurabhdaware: done

chaitanyadeorukhkar (dismissed):

packages/blade/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,221 @@
...
+   *
+   * @private
+   *
+   */
+  _isTagFocussed?: boolean;
+} & StyledPropsBlade &
+  TestID;
+
+const FocussableTag = styled(BaseBox)<{ isTagFocussed: TagProps['_isTagFocussed'] }>((props) => {
+  if (props.isTagFocussed) {
```
kamleshchandnani, line 62: need focus ring component soon now 😞

packages/blade/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,221 @@
...
+        padding={size === 'medium' ? mediumPadding : largePadding}
+        isTagFocussed={_isTagFocussed}
+      >
+        <ShowOnDesktop>
+          {getLeadingIcon({ size: 'small' })}
+          {getTagText({ size: 'small' })}
+          {getCloseIcon({ size: 'small' })}
+        </ShowOnDesktop>
+        <ShowOnMobile>
+          {getLeadingIcon({ size: size === 'large' ? 'medium' : 'small' })}
```
kamleshchandnani, line 212: why can't we just compute size based on screen and then pass that directly instead of 2 separate components?

packages/blade/src/components/Tag/__tests__/Tag.native.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  it('should call onDismiss', () => {
+    const dismissHandler = jest.fn();
+    const { queryAllByText, getAllByLabelText } = renderWithTheme(
+      <Tag onDismiss={dismissHandler}>in:User</Tag>,
+    );
+    expect(queryAllByText('in:User')[1]).toBeTruthy();
+    fireEvent.press(getAllByLabelText('Close in:User tag')[1]);
+    expect(dismissHandler).toBeCalledWith();
```
kamleshchandnani, line 27: assert that the `in:User` is falsy as well since it shall be dismissed?

packages/blade/src/components/Tag/__tests__/Tag.web.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it('should call onDismiss', async () => {
+    const user = userEvents.setup();
+    const dismissHandler = jest.fn();
+    const { getByRole, queryAllByText } = renderWithTheme(
+      <Tag onDismiss={dismissHandler}>in:User</Tag>,
+    );
+
+    const [_desktopTagText, mobileTagText] = queryAllByText('in:User');
```
kamleshchandnani, line 30: what is desktop tag and mobile tag? 🤔

packages/blade/src/components/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,176 @@
...
+        <Tag {...args} onDismiss={() => setIsTagVisible(false)}>
+          {children}
+        </Tag>
+      ) : null}
+    </Box>
+  );
+};
+
+export const Default = TagTemplate.bind({});
+Default.args = {
```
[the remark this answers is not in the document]
saurabhdaware: It is for creating overall focus ring on tag when used inside Input components. Which is I marked it internal.

packages/blade/src/components/Tag/Tag.tsx
saurabhdaware: That will make size dependent on client-side javascript. I wanted to avoid that since it changes height and can cause some layout shift in SSR apps. This one uses media query to switch sizes.

packages/blade/src/components/Tag/__tests__/Tag.native.test.tsx
saurabhdaware: It won't be falsy actually. We expect consumers to handle the closing state no so when dismiss is called, consumer removes the tag. From our end for unit test, just calling onDismiss handles the cases.

packages/blade/src/components/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        export default App;
+        `}
+      </Sandbox>
+    </StoryPageWrapper>
+  );
+};
+
+export default {
```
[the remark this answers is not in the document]
saurabhdaware: size is consistent on figma and dev and also predictable because its same on other components so unlikely that they will miss 🤔

packages/blade/src/components/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,176 @@
...
+          )
+        }
+
+        export default App;
+        `}
+      </Sandbox>
+    </StoryPageWrapper>
+  );
+};
+
```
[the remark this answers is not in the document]
saurabhdaware: The biggest real life usecase is using it inside AutoComplete 🙈 

I have one Dropdown story where I had temporarily used Badge, I can change it to Tags.

packages/blade/src/components/Button/IconButton/types.ts
```diff
@@ -1,12 +1,14 @@
+import type { IconButtonProps } from './IconButton';
 import type { IconComponent } from '~components/Icons';
 import type { TestID } from '~utils/types';
 import type { BladeCommonEvents } from '~components/types';
 
 export type StyledIconButtonProps = {
   icon: IconComponent;
   onClick: () => void;
-  size: 'medium' | 'large';
```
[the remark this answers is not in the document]
saurabhdaware: done. pushing in next commit.

packages/blade/src/components/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,176 @@
...
+export const Default = TagTemplate.bind({});
+Default.args = {
+  children: 'Unpaid',
+  icon: 'FileTextIcon',
+} as TagProps & { icon: string };
+
+export const Disabled = TagTemplate.bind({});
```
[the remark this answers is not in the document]
saurabhdaware: Will have to change the focus ring on our IconButton components. Mentioned it to design here - https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps?node-id=29744:568562&mode=dev#509768673

packages/blade/src/components/Tag/__tests__/Tag.web.test.tsx
saurabhdaware: There will tags of 2 sizes in DOM and we switch between them with media query (explained it [here](https://github.com/razorpay/blade/pull/1361#discussion_r1275745922))

packages/blade/src/components/Tag/Tag.stories.tsx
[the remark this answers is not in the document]
kamleshchandnani: so do we release it as it is?

packages/blade/src/components/Tag/__tests__/Tag.native.test.tsx
kamleshchandnani: but we shall have a story that captures hiding of tag as well right?

packages/blade/src/components/Tag/Tag.tsx
kamleshchandnani: they layout shift will anyways get triggered because our typography scale is diff for mobile and desktop and that gets evaluated client side only 🤔

saurabhdaware: > they layout shift will anyways get triggered because our typography scale is diff for mobile and desktop and that gets evaluated client side only 🤔

Lollll didn't know about this. Yup then just isMobile check makes more sense.

packages/blade/src/components/Tag/__tests__/Tag.native.test.tsx
saurabhdaware: I tried this but its not really testing any additional behaviour apart from React's useState. Feels like too much mocking for this usecase in particular. In something like [Dropdown we test the usage behaviour](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx#L398-L423) because it deals with our internal props as well such as `value` `onChange` etc. Which is not the case with Tag.

image

packages/blade/src/components/Tag/Tag.tsx
saurabhdaware: updated the code to use isMobile

packages/blade/src/components/Tag/Tag.stories.tsx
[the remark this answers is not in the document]
saurabhdaware: I guess yes. Don't see any way to solve this without touching IconButton's focus ring. We were also discussing about increasing focus area of icon button so that will also change the focus ring styles by default.

kamleshchandnani (approved):
lets gooo 🚀

Issue #1327, opened 2023-06-20
Feature/Background Image props on Box


Kushagra-ag:
Request for adding `background-image` prop on the blade `Box` component.

Along with this other props that control the sizing and position of background image will also be needed. Example of some - 
`background-size, background-position, background-origin, background-repeat`

cseas:
Use-case:

https://github.com/razorpay/frontend-website/pull/2214/files#diff-864784c3dc2a876993920d9e235dadbebacde22e9f874722bd517956441688a8R29

PR #1369, opened 2023-06-27
feat: add background-image, size, position, origin, repeat props

chaitanyadeorukhkar:
Closes: #1327

### DIFF
--- a/.changeset/tame-buses-compete.md
+++ b/.changeset/tame-buses-compete.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+feat: add background-image, size, position, origin, repeat props
--- a/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
+++ b/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
@@ -153,11 +153,6 @@ type BaseBoxVisualProps = MakeObjectResponsive<
       | BackgroundColorString<'surface'>
       | BackgroundColorString<'action'>
       | (string & Record<never, never>);
-    backgroundImage: CSSObject['backgroundImage'];
-    backgroundSize: CSSObject['backgroundSize'];
-    backgroundPosition: CSSObject['backgroundPosition'];
-    backgroundOrigin: CSSObject['backgroundOrigin'];
-    backgroundRepeat: CSSObject['backgroundRepeat'];
     lineHeight: SpacingValueType;
     touchAction: CSSObject['touchAction'];
     userSelect: CSSObject['userSelect'];
--- a/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
+++ b/packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
@@ -125,6 +125,11 @@ type BaseBoxVisualProps = MakeObjectResponsive<
       | BackgroundColorString<'surface'>
       | BackgroundColorString<'action'>
       | (string & Record<never, never>);
+    backgroundImage: CSSObject['backgroundImage'];
+    backgroundSize: CSSObject['backgroundSize'];
+    backgroundPosition: CSSObject['backgroundPosition'];
+    backgroundOrigin: CSSObject['backgroundOrigin'];
+    backgroundRepeat: CSSObject['backgroundRepeat'];
     lineHeight: SpacingValueType;
     touchAction: CSSObject['touchAction'];
     userSelect: CSSObject['userSelect'];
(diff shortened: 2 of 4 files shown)


packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -155,6 +160,36 @@ type BaseBoxVisualProps = MakeObjectResponsive<
 
 type BoxVisualProps = MakeObjectResponsive<{
   backgroundColor: BackgroundColorString<'surface'>;
+  /**
+   * **Warning**
```
chaitanyadeorukhkar, line 164: Adding warning until we pick up #1234

snitin315 (dismissed):
LGTM :shipit:

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -125,6 +125,11 @@ type BaseBoxVisualProps = MakeObjectResponsive<
       | BackgroundColorString<'surface'>
       | BackgroundColorString<'action'>
       | (string & Record<never, never>);
+    backgroundImage: CSSObject['backgroundImage'];
```
saurabhdaware, line 128: there is `PickByCSSPlatform` type below from line 150. Can we add these prop there? they will handle cross-platform and make it `undefined` on React Native so they won't need warnings.

.changeset/tame-buses-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
```
snitin315, line 2: Should it be `minor`? since we are adding more props.

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
saurabhdaware: That issue only exists for our repo no? so should we have this warning? On react native projects it will anyway be `undefined`

saurabhdaware: Also should we move `background*` and `border*` props to some `CommonVisualPropsType` kind of object? then we won't have to add these types twice.

I had initially created `BoxVisualProps` and `BaseBoxVisualProps` because visual props deferred between Box and BaseBox. They had different types (E.g. backgroundColor for Box has different type from backgroundColor for BaseBox). This is not the case anymore for backgroundImage, border props.

chaitanyadeorukhkar: Yeah makes sense, let me clean it up a bit

chaitanyadeorukhkar: I've refactored it to make it a bit cleaner but the `PickCSSByPlatform` is not working as intended anyway. Not even for the other types like `placeSelf` that were there already. See if you can take a quick look at it otherwise we can raise an issue for it for fixing `PickCSSByPlatform`

.changeset/tame-buses-compete.md
chaitanyadeorukhkar: Since its only addition of props which can happen often, we stick to patch. Only for new component or features we bump it with minor and for anything that's breaking existing consumers, we make it major

saurabhdaware: oh I've been bumping minor for new props all this time 🙈

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -142,6 +135,32 @@ type BaseBoxVisualProps = MakeObjectResponsive<
...
+    backgroundColor:
+      | BackgroundColorString<'feedback'>
+      | BackgroundColorString<'surface'>
+      | BackgroundColorString<'action'>
+      | (string & Record<never, never>);
+    backgroundImage: CSSObject['backgroundImage'];
+    backgroundSize: CSSObject['backgroundSize'];
+    backgroundPosition: CSSObject['backgroundPosition'];
+    backgroundOrigin: CSSObject['backgroundOrigin'];
+    backgroundRepeat: CSSObject['backgroundRepeat'];
```
saurabhdaware, line 160: I can see these are added to `CommonBoxVisualProps` now so we can remove from here right?

chaitanyadeorukhkar: Yeah forgot to remove those

saurabhdaware (approved):

Issue #1306, opened 2023-06-15
Counter: support for blue intent


abhieshekumar:
### Issue

Based on design the counter component should have support for intent=blue. But it somehow got missed during development.

1. [Figma](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=8222-70827&t=Knx7ByenuRUAr9Yc-4).
2. [Slack](https://razorpay.slack.com/archives/CMQ3RBHEU/p1686816575171229?thread_ts=1686814984.724319&cid=CMQ3RBHEU)

Please consider adding the intent as it can be used by designers but not developers.

PR #1374, opened 2023-06-28
feat: add blue variant to Counter component & change intent to variant

chaitanyadeorukhkar:
Fixes: #1306 

## ⚠️ Changes for Counter component
Changes the `intent` prop to `variant` since we support more than Feedback colors with the addition of `blue` color for Counter. We will continue to support `intent` prop for backward compatibility but it will be deprecated in an upcoming major release.

### DIFF
--- a/packages/blade/src/components/Counter/Counter.stories.tsx
+++ b/packages/blade/src/components/Counter/Counter.stories.tsx
@@ -58,7 +58,7 @@ const CounterTemplate: ComponentStory<typeof CounterComponent> = ({ ...args }) =
 export const Counter = CounterTemplate.bind({});
 Counter.args = {
   value: 20,
-  intent: 'neutral',
+  variant: 'neutral',
   contrast: 'low',
 };
 Counter.storyName = 'Default';
@@ -67,13 +67,13 @@ export const Max = CounterTemplate.bind({});
 Max.args = {
   value: 120,
   max: 99,
-  intent: 'neutral',
+  variant: 'neutral',
   contrast: 'high',
 };
 Max.storyName = 'Max';
 
 const CountersWithVariantTemplate: ComponentStory<typeof CounterComponent> = ({ ...args }) => {
-  const intents = ['positive', 'negative', 'notice', 'information', 'neutral'] as const;
+  const variants = ['positive', 'negative', 'notice', 'information', 'neutral'] as const;
 
   return (
     <BaseBox display="flex" flexDirection="column">
@@ -85,13 +85,13 @@ const CountersWithVariantTemplate: ComponentStory<typeof CounterComponent> = ({ 
         paddingBottom="spacing.5"
         flexWrap="wrap"
       >
-        {intents.map((intent) => (
+        {variants.map((variant) => (
           <CounterComponent
             {...args}
-            key={intent}
+            key={variant}
             marginRight="spacing.3"
             marginTop="spacing.2"
-            intent={intent}
+            variant={variant}
             contrast="low"
           />
         ))}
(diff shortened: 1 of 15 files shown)


packages/blade/src/components/Counter/Counter.tsx
```diff
@@ -47,24 +47,36 @@ type ColorProps = {
...
-    textColor: `feedback.text.${intent}.${contrast}Contrast`,
-    backgroundColor: `feedback.background.${intent}.${contrast}Contrast`,
+    textColor: 'feedback.text.neutral.lowContrast',
+    backgroundColor: 'feedback.background.neutral.lowContrast',
   };
+  if (isFeedbackVariant(variant)) {
+    props.textColor = `feedback.text.${variant}.${contrast}Contrast`;
+    props.backgroundColor = `feedback.background.${variant}.${contrast}Contrast`;
+  } else {
+    props.textColor = `badge.text.${variant}.${contrast}Contrast`;
```
chaitanyadeorukhkar, line 70: I've used badge's tokens only since we want Badge and Counter to be related. Any changes to badge should also reflect it on counter.

anuraghazra (approved):

packages/blade/src/components/Counter/Counter.tsx
kamleshchandnani: can we rather abstract this out as `feedback`? so all of the counters, badges, alerts, toasts(in future) etc can use the common thing?

packages/blade/src/components/Counter/Counter.stories.tsx
```diff
@@ -67,13 +67,13 @@ export const Max = CounterTemplate.bind({});
...
   max: 99,
-  intent: 'neutral',
+  variant: 'neutral',
   contrast: 'high',
 };
 Max.storyName = 'Max';
 
 const CountersWithVariantTemplate: ComponentStory<typeof CounterComponent> = ({ ...args }) => {
-  const intents = ['positive', 'negative', 'notice', 'information', 'neutral'] as const;
+  const variants = ['positive', 'negative', 'notice', 'information', 'neutral'] as const;
```
kamleshchandnani, line 76: add `blue`?

packages/blade/src/components/Counter/Counter.tsx
```diff
@@ -47,24 +54,37 @@ type ColorProps = {
...
   contrast = 'low',
 }: {
-  intent: NonNullable<CounterProps['intent']>;
+  variant: NonNullable<CounterProps['variant']>;
   contrast: NonNullable<CounterProps['contrast']>;
 }): ColorProps => {
   const props: ColorProps = {
-    textColor: `feedback.text.${intent}.${contrast}Contrast`,
-    backgroundColor: `feedback.background.${intent}.${contrast}Contrast`,
+    textColor: 'feedback.text.neutral.lowContrast',
```
kamleshchandnani, line 70: why hardcode it? you're already setting the default value on the args above right?

packages/blade/src/components/Counter/Counter.tsx
chaitanyadeorukhkar: You mean adding `blue` to `feedback`? That was intentionally not done because we don't want every feedback component to support blue

packages/blade/src/components/Counter/Counter.tsx
chaitanyadeorukhkar: Yes but variant can contain `blue` and there is no `feedback.text.blue.lowContrast`, that check happens below. ColorProps type is required just to make sure it never gets undefined in the future so assigning a value here is needed.  Other option is doing it all inline which looks too complicated for no reason. Just had to pick one approach. 

```
const props: ColorProps = {
    textColor: isFeedback
      ? `feedback.text.${variant}.${contrast}Contrast`
      : `badge.text.${variant}.${contrast}Contrast`,
    backgroundColor: isFeedback
      ? `feedback.background.${variant}.${contrast}Contrast`
      : `badge.background.${variant}.${contrast}Contrast`,
  };
```

kamleshchandnani (approved):

PR #1378, opened 2023-06-28
fix: make OTPInput fluid width

chaitanyadeorukhkar:
We encountered use-cases where OTPInput as fixed width would not make sense. Our initial plan was to add `isFullWidth` as a prop to ensure there would be no breaking change but when we compared the existing use-cases of OTPInput where it's fixed width along with other Blade elements like Input & Button, it made more visual sense for it to be of a fluid width. So we asked ourselves what we want to encourage as a default behaviour - Let it be fluid so that it can take up the same width as rest of the elements in a form or let it be fixed which might lead it to look out of place compared to rest of the elements in the form. We realised that we want to encourage a fluid width by default to maintain consistency with rest of the form elements. Even though this leads to a breaking change and it would've been easier to make it opt-in, this would be the right thing to do. Otherwise in the future every developer would have to add isFullWidth for every OTPInput component because design would want it to be of fluid width.


.changeset/great-dingos-sleep.md
```diff
@@ -0,0 +1,28 @@
...
+### Migration guide
+
+**It is recommended to let the OTPInput take the entire width of the parent form in order to ensure consistency with the rest of the Input elements.**
+
+If you still want to keep the UI for OTPInput on your existing screens the same, you can do the following:
+1. Wrap `OTPInput` with a `Box` component and assign a `maxWidth` of `168px` (for `otpLength` of 4) or `256px` (for `otpLength` of 6)
+```diff
++ <Box maxWidth='256px' />
+  <OTPInput label='Enter OTP' otpLength={6} />
++ </Box>
```
parichay28, line 19: @chaitanyadeorukhkar While the effort to migrate might be low, would it still make sense to include an explicit prop for fluid width as opt-in?

chaitanyadeorukhkar: Yeah, that was the initial plan but when we compared the existing use-cases of OTPInput where it's fixed width along with other Blade elements like Input & Button, it made more visual sense for it to be of a fluid width. So we asked ourselves what we want to encourage as a default behaviour - Let it be fluid so that it can take up the same width as rest of the elements in a form or let it be fixed which might lead it to look out of place compared to rest of the elements in the form. We realised that we want to encourage a fluid width by default to maintain consistency with rest of the form elements. Even though this leads to a breaking change and it would've been easier to make it opt-in, this would be the right thing to do. Otherwise in the future every developer would have to add `isFullWidth` for every OTPInput component because design would want it to be of fluid width.

parichay28: Makes sense, thanks for the detailed reply ✨

.changeset/great-dingos-sleep.md
```diff
@@ -0,0 +1,28 @@
+---
+"@razorpay/blade": major
+---
+
+chore: make OTPInput fluid width (#104)
```
saurabhdaware, line 5: what is `#104` 🤔

saurabhdaware (approved):
Looks good to me

kamleshchandnani (dismissed):
changeset and the PR title has some issue being linked. lets remove them before merging

chaitanyadeorukhkar:
> changeset and the PR title has some issue being linked. lets remove them before merging

Looks like you've updated it. Can you approve the PR?

kamleshchandnani (approved):

PR #1379, opened 2023-06-28
docs(blade): added Skeleton API 

anuraghazra:
## [View Formatted Content](https://github.com/razorpay/blade/blob/anu/skeleton-api/packages/blade/src/components/Skeleton/_decisions/decisions.md)


packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,326 @@
...
+Skeleton Loader is a static / animated placeholder for the information that is still loading. It mimic the structure and look of the entire view. 
+
+The skeleton blocks can have 3 types of shape
+- circle 
+- rectangle
+- pill
+
+Each of these shapes can be achieved by adjusting the width, height and border radius of the blocks. Properties of skeleton loaders.
+
+<img src="./skeleton-thumbnail.png" width="380" />
```
chaitanyadeorukhkar, line 11: Doesn't render

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,326 @@
...
+
+## Design
+- [Figma - Skeleton](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=16508%3A258522&mode=dev)
+
+
+## API
+
+| Prop          | Type        | Default     | Description                                                                                                                                             | Required |
+| ------------- | ----------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| type          | `TextTypes` | `undefined` | Predefined variants for Skeleton's dimensions when working with Typography components. (if width/height layout props are set they will take precedence) |          |
```
chaitanyadeorukhkar, line 21: As discussed, this might not be worth adding and would only lead to maintenance overheads.

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,326 @@
...
+Check the motion [here](https://www.figma.com/proto/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=16502-258345&scaling=min-zoom&page-id=16498%3A256331):
+
+https://github.com/razorpay/blade/assets/35374649/8298efde-f977-4aa3-bc2d-92fb50a8c830
+
+
+## Accessibility
+
+- All Skeleton nodes will have aria-hidden to hide them from screen readers
+- All Skeleton nodes will have `cursor: progress`
+- Consumers will have to add aria-busy in the respective container which is being rendered
```
chaitanyadeorukhkar, line 311: I'm not sure if this will be followed by consumers. Thoughts about exposing a `SkeletonWrapper` that only adds `aria-busy`? Seems like an overkill but I'm sure a lot of consumers would miss adding aria-busy

anuraghazra: Yes to option is to expose SkeletonGroup or similar wrapper component to handle it, but we need to make sure to enforce people to use SkeletonGroup. 

But even with SkeletonGroup, it might not be correct, imagine 5 cards with SkeletonGroup component inside each card's body now the page will have 5 aria-busy, but ideally there should be only 1 aria-busy for that whole section.

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,327 @@
...
+| -------- | ----------- | ------- | -------------------------------------------- | -------- |
+| contrast | `low, high` | `low`   | The contrast of the Skeleton to be rendered. |          |
+
+Skeleton will have subset of Box props to help users layout the Skeleton blocks accordingly. 
+
+Layout: 
+
+- borderRadius
+- height
+- minHeight
```
kamleshchandnani, line 29: why do we need min* properties for skeleton? 🤔

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,327 @@
...
+- maxHeight
+- width
+- minWidth
+- maxWidth
+
+Flex:
```
kamleshchandnani, line 35: why do we need Flex for skeleton?

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,327 @@
+# Skeleton <!-- omit in toc -->
+Skeleton Loader is a static / animated placeholder for the information that is still loading. It mimic the structure and look of the entire view. 
+
+The skeleton blocks can have 3 types of shape
```
kamleshchandnani, line 4: do we need to mention this explicitly? we have borderadius to tackle this already which is also implied? right this is leads to a confusion that we'll have a variant as a prop which is not the case

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,327 @@
...
+    <Text>{data.content}</Text>
+  </CardBody>
+</Card>
+```
+
+We discussed this, while it might look simple this approach also has few downsides: 
+
+- if we provide `isLoading` prop it might cause a lot of jumps and shifts on the page because we won't know until API responds if the CardHeader has title/prefix or subtitle set or not and it could be removed after the API is done loading.
+- Providing `isLoading` prop for all Blade components (where applicable) might not be the most flexible approach.
+- With `isLoading` prop consumers will anyways have to opt out and use custom Skeleton templates for the CardBody content so there's not much benefit to providing it out of the box.
```
kamleshchandnani, line 182: make it more understandable? the context is missing what exactly we are talking about here?

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,327 @@
...
+Check the motion [here](https://www.figma.com/proto/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=16502-258345&scaling=min-zoom&page-id=16498%3A256331):
+
+https://github.com/razorpay/blade/assets/35374649/8298efde-f977-4aa3-bc2d-92fb50a8c830
+
+
+## Accessibility
+
+- All Skeleton nodes will have aria-hidden to hide them from screen readers
+- All Skeleton nodes will have `cursor: progress`
+- Consumers will have to add aria-busy in the respective container which is being rendered
```
kamleshchandnani, line 196: add a code snippet for this?

kamleshchandnani (approved):
mostly looks good

packages/blade/src/components/Skeleton/_decisions/decisions.md
anuraghazra: Resolved, not doing this.

packages/blade/src/components/Skeleton/_decisions/decisions.md
anuraghazra: We will need to support all layout related props otherwise consumers won't be able to replicate complex structures like Card, Alert etc. 

image

packages/blade/src/components/Skeleton/_decisions/decisions.md
anuraghazra: Makes sense, removed.

packages/blade/src/components/Skeleton/_decisions/decisions.md
anuraghazra: Context is there, if you scroll up you have a codeblock with <Card isLoading \/> prop example, that's what we are talking about here. 

I've made the `A question that might arise is "Why not just give isLoading prop for Card?` bold though, maybe you missed it and didn't get what we are talking about.

> Open to suggestion on wording changes

packages/blade/src/components/Skeleton/_decisions/decisions.md
anuraghazra: Don't have a solid working example right now, I will add accessibility story in the storybook itself. There are few nitty gritties i need to handle and test for first. 

While the reference blog that I linked does suggest to use aria-busy, i need to see can the loading state be communicated via our `LiveAnnouncer` component also?

packages/blade/src/components/Skeleton/_decisions/decisions.md
anuraghazra: Same as https://github.com/razorpay/blade/pull/1379#discussion_r1259382751, we need to expose all layout props to ensure consumers can build complex. 

If a consumer's custom component can have maxWidth, the skeleton should also have maxWidth and related props.

packages/blade/src/components/Skeleton/_decisions/decisions.md
anuraghazra: Added thumbnail

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,326 @@
+# Skeleton <!-- omit in toc -->
+Skeleton Loader is a static / animated placeholder for the information that is still loading. It mimic the structure and look of the entire view. 
+
+The skeleton blocks can have 3 types of shape
+- circle 
+- rectangle
+- pill
+
+Each of these shapes can be achieved by adjusting the width, height and border radius of the blocks. Properties of skeleton loaders.
```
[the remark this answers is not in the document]
anuraghazra: Removed this line.

packages/blade/src/components/Skeleton/_decisions/decisions.md
anuraghazra: Will explore a11y demos on storybook https://github.com/razorpay/blade/pull/1379#discussion_r1259421017

packages/blade/src/components/Skeleton/_decisions/decisions.md
kamleshchandnani: okay here's my proposal:
> With `isLoading` prop consumers will anyways have to opt out and use custom Skeleton templates for the CardBody because CardBody accepts any JSX, so there's not much benefit to providing it out of the box because people will need to tweak it for different use cases

chaitanyadeorukhkar (approved):
LGTM

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,320 @@
...
+## Open Questions
+
+- How will Skeleton loader work with existing components like ModalHeader/Footer etc?
+  - Ans: we will let the consumer handle the skeleton loading for Card components
+
+
+<details>
+
+  <summary>
+    Should we infer dimensions or provide predefined variants?
```
divyanshu013, line 204: Wondering if we should provide couple of predefined variants like circle, rectangle which would automatically add required properties such as border radius for circle would be 50%, for rectangle border radius will be say 4px. Otherwise there might be lot of repetition required where people have to pass border radius 4px to any rectangular skeleton they're using (plus if in future we want to tweak that).

Another common usecase that comes to mind is where people want to show skeleton loaders for a piece of text they're rendering. Since we don't have a predefined variant for it, consumers will have to handle spacing, number of lines of skeleton loader, etc. Slight inconvenience.

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,320 @@
...
+        <Box>
+          <Box display="flex">
+            <Box>
+              <Skeleton width="40%" height="30px" />
+              <Skeleton width="30%" height="20px" />
+            </Box>
+            <Skeleton width="30px" height="50px" />
+          </Box>
+          <Skeleton width="100%" height="100px" />
+        </Box>
```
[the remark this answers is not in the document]
anuraghazra: Sorry forgot the elevation prop. We've added that on Box for this exact usecase. 

elevation="lowRaised" borderRadius="medium" will suffice.

packages/blade/src/components/Skeleton/_decisions/decisions.md
```diff
@@ -0,0 +1,320 @@
...
+      <CardBody>
+        {isLoading ? <Skeleton width="100%" height="100px" /> : <Text>{data.content}</Text>}
+      </CardBody>
+    </Card>
+  );
+};
+
+```
+
+**Q. A question that might arise is "Why not just give isLoading prop for Card?"**
```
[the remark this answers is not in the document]
anuraghazra: Makes sense, moved.

packages/blade/src/components/Skeleton/_decisions/decisions.md
anuraghazra: > Another common usecase that comes to mind is where people want to show skeleton loaders for a piece of text

Yeah we actually had a long discussion on this please check the "Should we infer dimensions" open question where we explored this and why we decided not to do it as phase-1

packages/blade/src/components/Skeleton/_decisions/decisions.md
anuraghazra: Updated.

packages/blade/src/components/Skeleton/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: There will be certain other things such as padding in card so users will have to match those in this `Box` wrapper. If card padding changes in a future release users will have to manually update such Box occurences (which is why I was suggesting this example might not be ideal - it recommends a practice to replicate a card by creating their own box wrapper.

Can we replace the outer `Box` by a `Card`?

packages/blade/src/components/Skeleton/_decisions/decisions.md
divyanshu013: Is this planned as future enhancement? Can be added in the doc.

packages/blade/src/components/Skeleton/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: > it recommends a practice to replicate a card by creating their own box wrapper.

I see, can update the examples. but agree that in general people will need to make reusable composite components for skeleton loaders, otherwise lot of copy pasting will happening.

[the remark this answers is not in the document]
anuraghazra: > Can we replace the outer Box by a Card?

We have both examples - 1. Skeleton inside `<Card />`  2. Skeleton inside `<CardBody />`

Since the root `<Card />` doesn't allow any other components as it's children other than Card specific ones, if consumer wants to make the whole card as a skeleton they will need to recreate the wrapper also.

kamleshchandnani (approved):

packages/blade/src/components/Skeleton/_decisions/decisions.md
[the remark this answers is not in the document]
divyanshu013: Need for recreating the wrapper seems non-ideal in terms of DX (and ensuring consistency eg. padding, etc.) 🤔

anshulsahni (approved):

PR #1380, opened 2023-06-29
feat: figma plugin to measure coverage

kamleshchandnani:
As of today, our typical Design ↔️ Dev lifecycle looks like this:

A designer designs the feature ➡️ Hand-off to Dev ➡️ Dev analyses the design and asks the Designer to build it Blade ➡️ Designer claims they have used Blade and the cycle repeats.

Now, in this entire process there's no way to identify where are the actual gaps, what is built using Blade, what is not built using Blade, etc.

To address all of the above problems, designers before doing a hand-off to dev can run this plugin and the plugin will put a card with the coverage and highlight the non-blade parts. This will make it easier for:

1. **Designers** - To always get instant feedback to validate if the designs are cohesive and in line with our design language.
2. **Devs** - To quickly spot what is built using Blade and what isn't and act/plan accordingly for their feature deliveries.

This Figma plugin will give us Blade coverage for a particular design file.

It calculates the coverage and puts a tiny coverage card on top of each frame with the following metrics.

https://github.com/razorpay/blade/assets/11384858/00534d4c-8731-4109-8b61-f503a76fa9b4


packages/plugin-figma-blade-coverage/src/bladeLibraryConstants.ts
```diff
@@ -0,0 +1,8217 @@
+export const paymentLightThemeData = {
+  components: {
```
anuraghazra, line 2: Did we manually collected these IDs or do we have a script to generate? 

If manually, let's add a TODO to automate this, seems like a huge effort to maintain these IDs manually. (any additional colors, components, tokens could get missed)

packages/plugin-figma-blade-coverage/src/main.ts
```diff
@@ -0,0 +1,429 @@
...
+      'bladeCoverageType#45789:0': bladeCoverageType,
+      'bladeCoverage#45789:1': `${bladeCoverage.toFixed(2)}%`,
+      'totalLayers#45789:2': totalLayers.toString().padStart(2, '0'),
+      'bladeComponents#45789:3': bladeComponents.toString().padStart(2, '0'),
+      'nonBladeComponents#45789:4': nonBladeComponents.toString().padStart(2, '0'),
+      'nonBladeTextStyles#45789:5': nonBladeTextStyles.toString().padStart(2, '0'),
+      'nonBladeColorStyles#45789:6': nonBladeColorStyles.toString().padStart(2, '0'),
+    });
+
+    const detachedCoverageCard = coverageCardInstance.detachInstance();
```
anuraghazra, line 132: 🤔 why are we detaching our own Coverage Card?

packages/plugin-figma-blade-coverage/src/main.ts
```diff
@@ -0,0 +1,429 @@
...
+          let isMixedColorStyleOfBlade = false;
+          let traversedNodeColorStyleId = '';
+
+          /** the textSyleId can have figma.mixed. so in that case we need to go character by character
+           * and do getRangeTextStyleId(charIndex,charIndex+1) instead of textStyleId
+           * */
+          if (traversedNode?.textStyleId === figma.mixed) {
+            isMixedTextStyleOfBlade = traversedNode.characters
+              .split('')
+              .every((character, index) => {
```
anuraghazra, line 202: Do we have to go to every character? 
Can we do `getRangeTextStyleId(0, traversedNode.character.length)`?

packages/plugin-figma-blade-coverage/src/main.ts
```diff
@@ -0,0 +1,429 @@
...
+        if (
+          traversedNode.type === 'INSTANCE' &&
+          (BLADE_COMPONENT_IDS.includes(
+            (traversedNode.mainComponent?.parent as ComponentSetNode)?.key ?? '',
+          ) ||
+            BLADE_COMPONENT_IDS.includes(traversedNode.mainComponent?.key ?? ''))
+        ) {
+          // we shall stop traversal further if we have found that an instance is Blade instance
+          // if we keep traversing then chances are the metrics will be skewed because Blade components are composed of non-blade themselves
+          // in code analytics we can add "data-*" to all the children till leaf nodes but over here we can't hence we stop
```
anuraghazra, line 313: Such a small detail that comes again and again.

packages/plugin-figma-blade-coverage/src/main.ts
```diff
@@ -0,0 +1,429 @@
...
+  // plugin used
+  // const pluginUsageCount = await incrementTotalUseCountAsync();
+
+  // const analytics = AnalyticsBrowser.load({ writeKey: '6lpfX5loXnTbBFVo2FbJHpjins0hGaC4' });
+  // await analytics.track('Blade Coverage Plugin Used', {
+  //   pluginUsageCount,
+  //   fileName: figma.currentPage.parent?.name,
+  //   pageName: figma.currentPage.name,
+  // });
+  figma.skipInvisibleInstanceChildren = true;
```
anuraghazra, line 369: noice

packages/plugin-figma-blade-coverage/src/bladeLibraryConstants.ts
kamleshchandnani: there's a script in the utils directory

packages/plugin-figma-blade-coverage/src/main.ts
kamleshchandnani: good question but that's because of how figma handles component instances 🤦 

so essentially I want to increase the width of the progress bar and also change the color basis the percentage coverage. now since it's a component I can't change the width based on coverage because you can't change anything in an instance except properties and the properties are just Text properties in this card. 

One way is to add variants for color change but I can't have a variant handle the progress bar width which is dynamic.

packages/plugin-figma-blade-coverage/src/main.ts
kamleshchandnani: if your text style has multiple styles then figma will return `figma.mixed` instead of a style id but it doesn't tells me what is different in that. So if you have something like this 

Qty: **2** _units_

in this figma will return the style as mixed hence i need to do character by character to check if each character has the color style/text style from blade 😢

packages/plugin-figma-blade-coverage/src/bladeLibraryConstants.ts
anuraghazra: Oh noice!

packages/plugin-figma-blade-coverage/src/main.ts
anuraghazra: I see, makes sense.

packages/plugin-figma-blade-coverage/src/main.ts
kamleshchandnani: yes in my research and digging through the docs and forums i couldn't find a better way, and logically as well it's against the instance philosophy. but yeah even I'm not liking it and if you find something better let me know we can replace it.

packages/plugin-figma-blade-coverage/src/main.ts
```diff
@@ -0,0 +1,429 @@
...
+  figma.notify('Calculating Coverage', { timeout: Infinity });
+
+  let nodes: SceneNode[] = [];
+  if (figma.currentPage.selection.length > 0) {
+    // you already have the selection, run the plugin
+    //@ts-expect-error type 'readonly SceneNode[]' is 'readonly' and cannot be assigned to the mutable type 'SceneNode[]'
```
anuraghazra, line 375: you can make the nodes readonly

```
  let nodes: readonly SceneNode[] = [];
```

Also make the getPageMainFrameNodes accept `nodes: readonly SceneNode[]`

Better to specify nodes are readonly so that we accidentally don't modify consumer's design.

packages/plugin-figma-blade-coverage/src/main.ts
```diff
@@ -0,0 +1,429 @@
...
+              bladeTextStyles++;
+            } else {
+              nonBladeTextStyles++;
+              highlightNonBladeNode(traversedNode);
+            }
+          }
+
+          /** the fillstyleId can have figma.mixed. so in that case we need to go character by character
+           * and do getRangeFillStyleId(4,5) instead of fillStyleId
+           * */
```
anuraghazra, line 230: nit: beautify these comments, getting a bit hard to read and looks ugly. 

```
/**
 * The textSyleId can have figma.mixed. so in that case we need to go character by character
 * and do getRangeTextStyleId(charIndex,charIndex+1) instead of textStyleId
 */
```

> ps: vscode automatically creates multiline comments nicely if you try /** and press enter

anuraghazra (dismissed):
Mostly looks good, there are few commented out analytics code I saw though.

packages/plugin-figma-blade-coverage/src/main.ts
kamleshchandnani: done

packages/plugin-figma-blade-coverage/src/main.ts
kamleshchandnani: done

Issue #1283, opened 2023-06-09
Feature/Add isRequired prop on Radiogroup component


Kushagra-ag:
Right now, for Radio Group component, `necessityIndicator` prop controls the visible asterisk sign and the `required` html prop.
We can separate these 2 roles with the addition of an `isRequired` prop on the blade component itself similar to what has been done for the TextInput Blade component.

PR #1384, opened 2023-07-01
feat: add isRequired prop to RadioGroup

archie252000:
Fixes: #1283

### DIFF
--- a/packages/blade/src/components/Radio/Radio.tsx
+++ b/packages/blade/src/components/Radio/Radio.tsx
@@ -39,6 +39,7 @@ type RadioProps = {
    * @default false
    */
   isDisabled?: boolean;
+  isRequired?: boolean;
   /**
    * Size of the radios
    *
@@ -49,7 +50,7 @@ type RadioProps = {
   StyledPropsBlade;
 
 const _Radio: React.ForwardRefRenderFunction<BladeElementRef, RadioProps> = (
-  { value, children, helpText, isDisabled, size = 'medium', testID, ...styledProps },
+  { value, children, helpText, isDisabled, isRequired, size = 'medium', testID, ...styledProps },
   ref,
 ) => {
   const groupProps = useRadioGroupContext();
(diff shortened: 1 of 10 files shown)


chaitanyadeorukhkar (dismissed):
@archie252000 thanks for another contribution! 🙌 

LGTM. @anuraghazra can you take a look. Also, shall we do the same for Checkbox as well?

packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -84,7 +86,7 @@ const _Radio: React.ForwardRefRenderFunction<BladeElementRef, RadioProps> = (
     isChecked,
     hasError,
     isDisabled: _isDisabled,
-    isRequired: groupProps.necessityIndicator === 'required',
+    isRequired: _isRequired,
```
anuraghazra, line 89: this might be breaking change, since existing customers who already used neccessityIndicator = required will now have to make sure to add this additional prop.

anuraghazra:
@archie252000 yes let's also do this for Checkbox.

archie252000:
> @archie252000 thanks for another contribution! 🙌
> 
> LGTM. @anuraghazra can you take a look. Also, shall we do the same for Checkbox as well?

Thanks!

archie252000:
> @archie252000 yes let's also do this for Checkbox.

sure

packages/blade/src/components/Radio/Radio.tsx
chaitanyadeorukhkar: We could avoid this by making it required if necessityIndicator is 'required' as well as controlling it with `isRequired`.

chaitanyadeorukhkar:
@archie252000 we had a discussion internally and decided to make this a non-breaking change. So we will need to ensure that existing use-cases do not break. Our logic needs to be like: `shouldBeRequired = necessityIndicator==='required' || isRequired`

archie252000:
@chaitanyadeorukhkar Ok, will do that.

chaitanyadeorukhkar:
@archie252000 I'm making the changes, Need to add a few more things I will push it and add you as a co-author

chaitanyadeorukhkar:
Closing in favour of #1414

Issue #1101, opened 2023-04-03
Document utilities in blade


saurabhdaware:
we don't document our utilities that are exposed from blade documented anywhere. Will be better to document usage and usecases somewhere

nashcheez:
Relevant [thread & request](https://razorpay.slack.com/archives/CMQ3RBHEU/p1680505182215379).

divyanshu013:
Also to consider the utlis we want consumers to use eg. spacing, responsive (media queries), theme, etc.

saurabhdaware:
We should do this before writing documentation - https://github.com/razorpay/blade/issues/1136

We'll have to move few utils to internal utils and not expose them to consumer. Currently it's a bit mixed.

PR #1391, opened 2023-07-04
docs(utils): add documentation for `useTheme`

chaitanyadeorukhkar:
Partially re-solves: #1101

### DIFF
--- a/packages/blade/docs/utils/useTheme.stories.mdx
+++ b/packages/blade/docs/utils/useTheme.stories.mdx
@@ -14,7 +14,7 @@ import { Meta } from '@storybook/addon-docs';
 
 ### Overview
 
-`useTheme` is a custom React hook provided by the `@razorpay/blade/components`  that allows you to access the current theme and color scheme of your application.
+`useTheme` is a custom React hook provided by the `@razorpay/blade/utils`  that allows you to access the current theme and color scheme of your application.
 
 The `useTheme` hook returns a `ThemeContext` object that contains the following properties:
 
@@ -31,7 +31,7 @@ To use the `useTheme` hook, you must first wrap your application with a `Blad
 
 ```tsx
 import React from 'react';
-import { useTheme, Heading, Text, Button } from '@razorpay/blade/components';
+import { useTheme, Heading, Text, Button } from '@razorpay/blade/utis';
 
 const MyComponent = () => {
   const { theme, colorScheme, platform, setColorScheme } = useTheme();
--- a/packages/blade/docs/utils/useTheme.stories.mdx
+++ b/packages/blade/docs/utils/useTheme.stories.mdx
@@ -1,0 +1,108 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Utils/useTheme" />
+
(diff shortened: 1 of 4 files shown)


packages/blade/docs/utils/useTheme.stories.mdx
```diff
@@ -0,0 +1,108 @@
...
+- [Usage](#usage)
+- [Types](#types)
+  - [`Theme`](#theme)
+  - [`ColorScheme`](#colorscheme)
+  - [`Platform`](#platform)
+  - [`ColorSchemeInput`](#colorschemeinput)
+
+### Overview
+
+`useTheme` is a custom React hook provided by the `@razorpay/blade/components`  that allows you to access the current theme and color scheme of your application.
```
saurabhdaware, line 17: now that we're properly documenting this, should we expose this from `@razorpay/blade/utils` instead? that's where other hooks are exported from so this seems a bit confusing.


If we want to avoid breaking change, we can probably export from both and document the `@razorpay/blade/utils` everywhere 🤔

saurabhdaware (commented):
Documentation looks good to me. Only one doubt about export if useTheme

packages/blade/docs/utils/useTheme.stories.mdx
chaitanyadeorukhkar: Makes sense. Putting it up for a vote.

saurabhdaware (approved):

packages/blade/docs/utils/useTheme.stories.mdx
kamleshchandnani: what was the conclusion lets put that as well?

chaitanyadeorukhkar: Yes, put it up for a vote - got max votes for putting it in both places

chaitanyadeorukhkar: I've updated the documentation also accordingly

packages/blade/src/utils/index.ts
```diff
@@ -15,3 +15,5 @@ export * from './makeSize';
 // https://github.com/razorpay/blade/pull/1193#issuecomment-1609122359
 export * from './toTitleCase';
 export * from './usePrevious';
+
+export { default as useTheme, ThemeContext } from '~components/BladeProvider/useTheme';
```
kamleshchandnani, line 19: don't think so we shall have interdependencies like this. we wanted to keep them separate hence separate packages. Its okay to have component specific hooks to be imported from components. also utils for meant for things which people can use generally but useTheme can't be used generally without blade's context

packages/blade/docs/utils/useTheme.stories.mdx
```diff
@@ -0,0 +1,108 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Utils/useTheme" />
+
+## useTheme <!-- omit in TOC -->
+
+- [Overview](#overview)
```
kamleshchandnani, line 7: use our list component? this looks very weird when rendered

PR #1399, opened 2023-07-05
test: added skeleton tests

anuraghazra:


packages/blade/src/components/Skeleton/__tests__/Skeleton.native.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { Skeleton } from '../Skeleton';
+import renderWithTheme from '~utils/testing/renderWithTheme.native';
+import { Box } from '~components/Box';
+
+describe('<Skeleton />', () => {
+  it('should render skeleton', () => {
+    const { toJSON } = renderWithTheme(<Skeleton width="100%" height="50px" />);
+    expect(toJSON()).toMatchSnapshot();
+  });
+
```
kamleshchandnani, line 13: add test for checking meta attribute?

packages/blade/src/components/Skeleton/__tests__/Skeleton.native.test.tsx
```diff
@@ -20,4 +21,18 @@ describe('<Skeleton />', () => {
     );
     expect(toJSON).toMatchSnapshot();
   });
+
+  it('should be hidden', () => {
```
kamleshchandnani, line 25: probably in future PR we can create a util test that we can just include in all the test by default by passing the name of the attribute so we don't have to write this again and again

kamleshchandnani (approved):

packages/blade/src/components/Skeleton/__tests__/Skeleton.native.test.tsx
anuraghazra: Sure, opened an issue to track https://github.com/razorpay/blade/issues/1443

chaitanyadeorukhkar (approved):

PR #1401, opened 2023-07-05
fix(blade): tooltip zIndex issue 

### LINKED ISSUE #1372: Tooltip is rendered outside the visible viewport
While using the tooltip component in the X dashboard, I encountered this issue where the component is rendered outside the viewport. I am trying to render a tooltip on Blade's Text component. I've attached the below video to help you understand this issue better.

[Figma Link](https://www.figma.com/file/kxLqOeCiJc0Qzh6RmoJnX6/Vendor-Advances-(without-PO)?type=design&node-id=801-297390&mode=design&t=MUuu9VQToc8rDkpV-4)

Here is the code snippet I am trying to run:
```js
      <Tooltip content="Total advance selected should be ≤ Payable amount" placement="bottom">
        <TooltipInteractiveWrapper>
          <Text type="placeholder" size="small">
            {notes}
          </Text>
        </TooltipInteractiveWrapper>
      </Tooltip>
```

https://github.com/razorpay/blade/assets/48921037/870b7eca-d8cb-4558-93a5-1b3330c32b9f

anuraghazra:
fixes #1372

- Added default zIndex 100 to tooltip
- Added zIndex prop to Tooltip so that consumers can customize it


packages/blade/src/components/Tooltip/Tooltip.web.tsx
```diff
@@ -103,6 +104,7 @@ const Tooltip = ({
             ref={refs.setFloating}
             style={floatingStyles}
             pointerEvents="none"
+            zIndex={zIndex}
```
anuraghazra, line 107: Added zIndex prop to Tooltip, because we don't know if consumer's wrapper will have 100 or 200 zIndex it can be larger. 

In that cases better to let consumers customise it.

anuraghazra: Removed and assigned a higher value. 

Raised an issue to track the refactor: https://github.com/razorpay/blade/issues/1402

chaitanyadeorukhkar (approved):

PR #1415, opened 2023-07-10
feat: standardise isRequired & necessityIndicator across `Checkbox`, `Radio` & `Input` components

chaitanyadeorukhkar:
Fixes: #1283 

Co-authored-by: archie252000 <arc252000@gmail.com>


anuraghazra (approved):
Looks good

packages/blade/src/components/Checkbox/Checkbox.tsx
```diff
@@ -173,6 +173,9 @@ const _Checkbox: React.ForwardRefRenderFunction<BladeElementRef, CheckboxProps>
   const _validationState = validationState ?? groupProps?.validationState;
   const _hasError = _validationState === 'error';
   const _isDisabled = isDisabled ?? groupProps?.isDisabled;
+  const _isRequired = Boolean(
```
saurabhdaware, line 176: The slack message you linked in description seems to say that `necessityIndicator` should add visual asterick and `isRequired` prop should add `required` HTML attribute right?

This implementation seems to add `required` attribute even when `necessityIndicator` is set 🤔 

That is what we follow in other components as well - https://razorpay.slack.com/archives/C01H13RTF8V/p1686301617044819?thread_ts=1685690380.948209&cid=C01H13RTF8V

saurabhdaware: Oh ok understood, so you mean now just passing `necessityIndicator="require"` will add asterick and set `required` prop.


But if I want require inputs without asterick, then I can just use `isRequired` 🤔 ? is this what we're doing?

chaitanyadeorukhkar: Yup, that's the idea

anuraghazra (approved):

Issue #1400, opened 2023-07-05
Lowercase color token output generated by razorsharp plugin


Jaynil1611:
@burhanuday Attaching a screenshot below to help understand this issue better.
Ideally color token should have been `feedback.icon.info.lowContrast`. This is happening for all colors token outputs where cameCasing is used.

<img width="342" alt="Screenshot 2023-07-05 at 12 56 14 PM" src="https://github.com/razorpay/blade/assets/48921037/735d5d6c-042b-4e73-bd31-45c3770eccad">

PR #1429, opened 2023-07-12
fix: bugs in razorsharp

burhanuday:
Closes https://github.com/razorpay/blade/issues/1400

1. Fix bug with `justifyContent` and `alignItems` generating incorrect values
2. Fix bug with incorrect value of color tokens getting generated
3. Improve algo that generates width, height, maxWidth, maxHeight, and flex grow properties


snitin315 (approved):
:shipit:

Issue #1164, opened 2023-05-09
Feature/display attribute for elements other than <Box>


Kushagra-ag:
This is a good to have feature as there's a workaround for this issue at present.

Usecase - Developer wants to hide the below link component on mobile.
```
<Link display={{ base: 'none', l: 'block' }} > I am link </Link>
```
This is not possible currently as `Link` component doesn't support display attribute.

Workaround - 
```
<Box display={{ base: 'none', l: 'block' }} >
  <Link> I am link </Link>
</Box>
```

saurabhdaware:
So we don't support `display` property because it can be misused (E.g. someone might just make Badge as a display `block` which will change how the Badge looks)

One way could be to support just `none` and `default` on `display` but I feel then we'll deviate from actual possible CSS values. We can take a call if more people request for this.

For now, if you want, you can also create maybe `<Mobile />` and `<Desktop />` (or any other name) type of wrapper components in your repo if you feel writing display object again and again is too verbose. 

```html
<Mobile>
  <Button>Mobile Button</Button>
</Mobile>
<Desktop>
  <Button>Desktop Button</Button>
</Desktop>
```

divyanshu013:
Agreed, we brainstormed this a lot on [layout RFC](https://github.com/razorpay/blade/blob/master/rfcs/2023-01-06-layout.md#layout-based-styling-props) and decided to not support `display` props intentionally on the presentational components (because they can mess the component's layout).

saurabhdaware:
I think we can close this for now. If we get more people requesting this, we can take a look again.

Kushagra-ag:
One caveat here is unnecessary DOM element that is added in the workaround suggested

divyanshu013:
Yes, that is an inconvenience. Unfortunately that is the best way currently to prevent components misuse. Open to other suggestions that can check all the boxes in the RFC.

saurabhdaware:
Reopening this because we got another request https://razorpay.slack.com/archives/CMQ3RBHEU/p1685967672996579?thread_ts=1685526177.654729&cid=CMQ3RBHEU

cseas:
> someone might just make Badge as a display block which will change how the Badge looks

Pretty sure this is impossible. The devs are only trying to match what the designers have built, not ship broken designs on purpose. In hopes of preventing impossible cynical scenarios, we're blocking devs from many genuine use-cases that are a no-brainer for Blade to support from the beginning. [[Example][1]]

[...]

PR #1431, opened 2023-07-12
feat: support `display` prop for all components with styled props

chaitanyadeorukhkar:
fixes: #1164 

Components that didn't need any changes (display prop wasn't breaking anything):
- Button
- Checkbox, CheckboxGroup
- Radio, RadioGroup
- List
- Spinner
- Typography
- Link 
- Code(it breaks with certain display props but adding wrappers on typography would be very messy and its okay if people break it since it would be unusable in the broken state)

Not added display prop since it doesn't support any styled props
- Tooltip
- Bottomsheet
- Modal

Misc fixes in this PR:
- Added styledProps to Switch
- Added `testID` & meta attribute to Dropdown

Win 🧘‍♂️:
Screenshot 2023-07-13 at 4 41 42 PM


packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -293,29 +293,32 @@ const Alert = ({
   }
 
   return (
-    <StyledAlert
-      intent={intent}
-      contrastType={contrastType}
-      isFullWidth={isFullWidth}
-      isDesktop={isDesktop}
-      textAlign={'left' as never}
+    <BaseBox
```
saurabhdaware, line 296: I think all Box wrappers should have `display: inline-block` by default right? otherwise they'll take up full width even when the component inside it is inlined

packages/blade/src/components/Divider/Divider.stories.tsx
```diff
@@ -133,13 +135,17 @@ const DividerVerticalTemplate: ComponentStory<typeof DividerComponent> = () => {
           flexDirection="row"
           height="50px"
         >
-          <Button variant="secondary" marginRight="spacing.4" isFullWidth>
-            Sign up
-          </Button>
+          <BaseBox display="flex" flex={1}>
```
saurabhdaware, line 138: is this a related change?

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -182,8 +184,13 @@ const _Dropdown = ({
   return (
     <BottomSheetAndDropdownGlueContext.Provider value={BottomSheetAndDropdownGlueContextValue}>
       <DropdownContext.Provider value={contextValue}>
-        <BaseBox position="relative" textAlign={'left' as never} {...getStyledProps(styledProps)}>
-          {children}
+        <BaseBox
```
saurabhdaware, line 187: is wrapper needed on Dropdown?

packages/blade/src/components/Alert/Alert.tsx
chaitanyadeorukhkar: Better to keep it unset by default? Since we allow people to change the display prop on the wrapper, they can tweak it however they want. 

Maybe some specific components that needs to have a specific display prop, like `inline`, we can replicate that on the wrapper component by default but allow consumers to override it anyway. Let me see if I can find any such instances in this PR

packages/blade/src/components/Divider/Divider.stories.tsx
chaitanyadeorukhkar: Not really but story was breaking on RN so just fixed it

packages/blade/src/components/Dropdown/Dropdown.tsx
chaitanyadeorukhkar: Yeah, without it Dropdown was breaking especially if you make it flex, the sub-components get messed up

packages/blade/src/components/Alert/Alert.tsx
chaitanyadeorukhkar: Did this for amount, other places it seems to not need it. Can you check once?

saurabhdaware: Can we do it on Counter, Badge, Indicators too? components that are small in width and usually used along with text by default?

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: Ohh do we want to consider the flex scenarios as well? I was thinking more along the lines of that components shouldn't break when we pass `block` or `inline-block` to it. Because while doing responsiveness, those are the props that usually people use to "show" an element.

E.g.

```jsx
<Badge display={{ base: 'none', m: 'block' }}>I am badge visible on desktop</Badge>
```

Things breaking when we pass "flex" should be fine right? those will break the UI itself so it will be noticeable and won't be shipped. E.g. I can't think of scenario where someone would want to make Dropdown a display flex.

chaitanyadeorukhkar: I think we'll have to consider all scenarios right? It shouldn't break the component's visuals. Imagine someone shipped this code without testing on smaller screen and it breaks the component's UI itself. 

```tsx
<Badge display={{ base: 'none', s: 'flex' }}>I am badge visible on desktop</Badge>
```

packages/blade/src/components/Alert/Alert.tsx
chaitanyadeorukhkar: So Counter, Badge & Indicators have flex wrappers so it shouldn't be much of a problem even if they make it inline

chaitanyadeorukhkar: Counter, Badge, & Indicators have been flex since forever now. With this change, they can now make it inline. I think its better to keep the way it is for now without changing it. I will make the wrapper also flex so that it remains as it was before

chaitanyadeorukhkar: Especially because these components are used within Card, Modal & Bottomsheet. Don't want to impact it there by enforcing it as inline here

saurabhdaware: Oh actually we did try to make Badge inline-flex but I just noticed that we didn't change display property of the wrapper container https://github.com/razorpay/blade/pull/1254

We can do this separately then but we should make it inline-flex / inline-block as it was raised by our consumers as well https://github.com/razorpay/blade/issues/1241, https://razorpay.slack.com/archives/CMQ3RBHEU/p1685526734598399?thread_ts=1685526177.654729&cid=CMQ3RBHEU.

Lets take it up separately since this change is big in itself already.

kamleshchandnani: lets create an issue to track it?

chaitanyadeorukhkar: Saurabh already did #1447

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -649,6 +649,7 @@ export const BaseInput = React.forwardRef<HTMLInputElement, BaseInputProps>(
           justifyContent={isLabelLeftPositioned ? 'center' : undefined}
           alignItems={isLabelLeftPositioned ? 'center' : undefined}
           position="relative"
+          width="100%"
```
kamleshchandnani, line 652: why this?

kamleshchandnani:
@chaitanyadeorukhkar resolve conflicts?

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
chaitanyadeorukhkar: Without this, the input will break if we change the display property

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -124,40 +124,40 @@ const _Badge = ({
   } as const;
 
   return (
-    <StyledBadge
-      backgroundColor={backgroundColor}
-      size={size}
-      textAlign={'left' as never}
+    <BaseBox
+      display="flex"
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: getStyledProps() will prioritize it. 

```tsx
<BaseBox
      display="flex"
      {...metaAttribute({ name: MetaConstants.Badge, testID })}
      {...getStyledProps(styledProps)} // user's display prop will be added from here
    >
      <StyledBadge backgroundColor={backgroundColor} size={size} textAlign={'left' as never}>
        <BaseBox
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -251,32 +251,36 @@ const _Amount = ({
...
-        marginRight="spacing.1"
-        fontWeight={currencyWeight}
-        fontSize={currencyFontSize}
-        color={currencyColor}
-        as={isReactNative ? undefined : 'span'}
+      <BaseBox
+        paddingLeft="spacing.2"
+        paddingRight="spacing.2"
+        // @TODO: fix casting of platform types. currently they all become `never` type
+        display={isReactNative ? castNativeType('flex') : castWebType('inline-flex')}
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Making the display prop of new wrapper same as the earlier one's to ensure nothing breaks in existing implementations. See the code below this, it has this too:
```tsx
        display={isReactNative ? castNativeType('flex') : castWebType('inline-flex')}
```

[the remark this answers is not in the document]
kamleshchandnani: TODO things fixing for this file?

packages/blade/src/components/Accordion/Accordion.tsx
```diff
@@ -95,14 +95,14 @@ const Accordion = ({
   return (
     <AccordionContext.Provider value={accordionContext}>
       <BaseBox
-        minWidth={MIN_WIDTH}
-        maxWidth={MAX_WIDTH}
         {...metaAttribute({ name: MetaConstants.Accordion, testID })}
```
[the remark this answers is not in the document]
kamleshchandnani: `metaAttribute` also needs to be passed to this `<BaseBox minWidth={MIN_WIDTH} maxWidth={MAX_WIDTH} width="100%">` as well?

packages/blade/src/components/Amount/Amount.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Tried looking into it, might take a bit longer - its a weird issue. Lets not block the PR for it. We can look at it when working on some component where similar use-case would come up

packages/blade/src/components/Accordion/Accordion.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: No... the meta Attribute will be there only on the root level wrapper, right?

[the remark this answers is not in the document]
kamleshchandnani: no, after @anuraghazra fix we need to add it to all the nodes otherwise the coverage will mess up

.changeset/lovely-pants-breathe.md
```diff
@@ -0,0 +1,7 @@
+---
+"@razorpay/blade": minor
+---
+
+feat: support `display` prop for all components
```
snitin315, line 5: This might be misleading.
```suggestion
feat: support `display` prop for all components with styled props
```

packages/blade/src/components/Accordion/Accordion.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Its basebox, so we don't need to add anything. Confirmed with @anuraghazra

kamleshchandnani (approved):

kamleshchandnani (approved):

PR #1432, opened 2023-07-13
feat: add analytics for tracking usage

kamleshchandnani:
This PR:
* integrates with segment analytics service
* sends blade coverage metrics whenever the plugin is run

Attaching the graphs for reference of what data we'll be able to derive. this is similar to our code coverage analytics script
 

Screenshot 2023-07-13 at 5 31 18 PM
Screenshot 2023-07-13 at 5 31 39 PM


packages/plugin-figma-blade-coverage/build-figma-plugin.main.js
```diff
@@ -0,0 +1,23 @@
+require('dotenv').config();
+
+module.exports = function buildFigma(buildOptions) {
+  const define = {};
+
+  // eslint-disable-next-line guard-for-in
+  for (const secretKey in process.env) {
+    let secretValue = process.env[secretKey];
+    if (secretKey === 'SEGMENT_WRITE_KEY') {
+      secretValue = btoa(process.env[secretKey]);
```
anuraghazra, line 10: 🤔 why are we doing this btoa?

packages/plugin-figma-blade-coverage/src/utils/sendAnalytics.ts
```diff
@@ -0,0 +1,40 @@
...
+}): Promise<void> => {
+  const pluginUsageCount = await incrementTotalUseCountAsync();
+  await fetch('https://api.segment.io/v1/track', {
+    method: 'POST',
+    headers: {
+      'content-type': 'application/json',
+      Authorization: `Basic ${process.env.SEGMENT_WRITE_KEY}`,
+    },
+    body: JSON.stringify({
+      userId: figma.currentUser?.id ?? `logged-out-user-${Date.now()}`,
```
anuraghazra, line 18: can logged out users even run the plugin?

packages/plugin-figma-blade-coverage/src/utils/sendAnalytics.ts
```diff
@@ -0,0 +1,40 @@
+import { incrementTotalUseCountAsync } from '@create-figma-plugin/utilities';
+
+export const sendAnalytics = async ({
+  eventName,
+  properties,
+}: {
+  eventName: string;
+  properties: object;
```
anuraghazra, line 8: let's add proper types here for `properties` which things we accept/send

packages/plugin-figma-blade-coverage/src/utils/sendAnalytics.ts
kamleshchandnani: nupe. but because figma type for this is nullable and segment userId expects it to be string i had to do this

packages/plugin-figma-blade-coverage/src/utils/sendAnalytics.ts
kamleshchandnani: honestly not sure about the properties right now we would want to send. since its a generic properties object i feel its okay to send anything and everything we pass to this object?

packages/plugin-figma-blade-coverage/build-figma-plugin.main.js
kamleshchandnani: because it'll be passed in headers as Basic Auth. I saw you were passing writekey in body in the code analytics but segment's docs suggest this and it makes sense to pass this in header for security reasons too
Screenshot 2023-07-13 at 6 20 46 PM

Issue #1326, opened 2023-06-20
[Feat]: Support negative spacing tokens in Styled Props


anuraghazra:
Originally raised by @AnshulNautiyal - [Internal Thread](https://razorpay.slack.com/archives/CJY7W3WFM/p1687239763785379?thread_ts=1682600861.191649&cid=CJY7W3WFM) 

Because we don't support `margin="-spacing.5"` he had to do `margin={-${theme.spacing[5]}px}`

Now `-${theme.spacing[5]}px` throws a TS error because by default TS will resolve this type as `string` and not the concrete `-16px` literal type, thus I suggested him to use:

```jsx
margin={`-${theme.spacing[5]}px` as const}
```

or alternatively we can also do 

```jsx
import { makeSpace } from "@razorpay/blade/utils";

margin={makeSpace(theme.spacing[5])}
```

But i think this is a bit verbose and an inconvenience to opt into interpolation to do negative margins, we should support negative variants for these also:

```jsx
margin="-spacing.5"
```

archie252000:
I think introducing negative values within spacing will make more sense.
We can do this:
```
export const spacing: Spacing = {
  0: 0,
  1: 2,
 -1: -2    //Add negative value
  2: 4,
 -2: -4    //Add negative value
  .
  .
  .
};
```
So this syntax becomes valid ```margin="spacing.-5"```

saurabhdaware:
> ```js
> import { makeSpace } from "@razorpay/blade/utils";
>
> margin={makeSpace(-theme.spacing[5])}
> ```

I am actually favouring this more.

Negative spacings are not as common and adding them to autocomplete might overwhelm the autocomplete. Right now I can type "spacing.5" and hit enter and it will always be the correct spacing.5 but with this addition, I will have to choose between negative and positive variant from AutoComplete of editor everytime.

Introducing this in token-level might also introduce more maintenance. We will have to make sure our positive spacings are in sync with negative spacings.

saurabhdaware:
I don't seem to see any typescript error mentioned in original issue description 🤔 

Was it some intermediate error that got fixed now @anuraghazra?

`theme.spacing['1']` has a type of number so it will correctly type it as `-2px` right?

<img width="412" alt="image" src="https://github.com/razorpay/blade/assets/30949385/fcaeccff-a2d7-4426-855e-5357a21cd37a">

anuraghazra:
> I don't seem to see any typescript error mentioned in original issue description 🤔

[...]

PR #1435, opened 2023-07-13
feat: support negative spacing tokens in styled props

archie252000:
fixes: #1326


chaitanyadeorukhkar:
@archie252000 thanks for the PR! I don't think we should add negative tokens. This becomes a bit difficult to scale and we will have to add a negative token to every positive token we add in the future. Rather we should see if we can allow using the same tokens but with a negation. We might have to explore few other options. I'd suggest lets discuss it on the issue itself and conclude what is the right approach first.

archie252000:
@chaitanyadeorukhkar got it.

archie252000:
closing this pr for now

PR #1440, opened 2023-07-14
feat: add support for ListItemText within List

chaitanyadeorukhkar:
enables consumers to pass custom color, weight, etc. properties to
text inside ListItem


packages/blade/src/components/List/ListItem.tsx
```diff
@@ -111,12 +111,13 @@ const _ListItem = ({
     if (
       typeof child === 'string' ||
       isValidAllowedChildren(child, 'ListItemLink') ||
+      isValidAllowedChildren(child, 'ListItemText') ||
       isValidAllowedChildren(child, 'ListItemCode')
```
anuraghazra, line 115: Can we put all these ids in a separate file? Same how we are doing in other components.

packages/blade/src/components/List/ListItemText.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+type ListItemTextProps = Omit<TextProps<{ variant: TextVariant }>, 'variant' | 'size'>;
+
+const _ListItemText = ({ children, testID, ...props }: ListItemTextProps): ReactElement => {
+  const { size } = useListContext();
+
+  return <Text {...props} size={size} children={children} testID={testID} />;
+};
+
+const ListItemText = assignWithoutSideEffects(_ListItemText, { componentId: 'ListItemText' });
```
anuraghazra, line 15: same with this componentId

packages/blade/src/components/List/ListItemText.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { Text } from '../Typography';
+import { useListContext } from './ListContext';
+import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
+
+type ListItemTextProps = Omit<TextProps<{ variant: TextVariant }>, 'variant' | 'size'>;
+
+const _ListItemText = ({ children, testID, ...props }: ListItemTextProps): ReactElement => {
+  const { size } = useListContext();
+
+  return <Text {...props} size={size} children={children} testID={testID} />;
```
anuraghazra, line 12: we can just do

```
<Text {...props} size={size} />;
```

do we need to destructure and set again? 🤔

anuraghazra (dismissed):
LGTM

packages/blade/src/components/List/ListItemText.tsx
chaitanyadeorukhkar: It should be fine since Text destructures it anyway

packages/blade/src/components/List/List.tsx
```diff
@@ -129,7 +128,7 @@ const _List = ({
   );
 };
 
-const List = assignWithoutSideEffects(_List, { componentId: ComponentIds.List });
+const List = assignWithoutSideEffects(_List, { componentId: MetaConstants.List });
```
saurabhdaware, line 132: Why this change 🤔? we usually use ComponentIds here no?

saurabhdaware (approved):
Lookss good to me. Added minor doubt above.

packages/blade/src/components/List/List.tsx
chaitanyadeorukhkar: This was becoming too confusing, using 2 constants for the same thing. We should use the same naming convention for component name on meta data as well as id. Otherwise we'll maintain 2 constants for every component.

Issue #1269, opened 2023-06-07
Refactor: BottomSheet & Modal focus management logic


anuraghazra:
> Reference: https://github.com/razorpay/blade/pull/1228#discussion_r1219150725

For Modal we are using [FloatingFocusManager](https://floating-ui.com/docs/floatingfocusmanager) which comes with focus return, initial focus, focus trapping etc features, but right now we are using custom logic to achieve those behaviours. 

We can refactor them to remove a lot of code in both BottomSheet and Modal. 

Similarly in [BottomSheet web](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1685443630811999) we can utilize [FloatingOverlay](https://floating-ui.com/docs/FloatingOverlay) which handles stacking of zIndexes.

chaitanyadeorukhkar:
We have completed this partially with https://github.com/razorpay/blade/pull/1340 and implemented this for Modal. Need to do the same for BottomSheet

PR #1441, opened 2023-07-16
refactor: BottomSheet focus management logic 

archie252000:
Fixes: #1269


packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
```diff
@@ -433,31 +412,43 @@ const _BottomSheet = ({
...
-          />
-          {children}
-        </BaseBox>
-      </BottomSheetSurface>
+      {isMounted ? (
+        <FloatingFocusManager
+          returnFocus
+          initialFocus={initialFocusRef ?? defaultInitialFocusRef}
+          context={context}
+          modal={true}
```
chaitanyadeorukhkar, line 420: Why do we need to make it `modal=true`? This isn't a modal, right?

packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
```diff
@@ -433,31 +412,43 @@ const _BottomSheet = ({
...
-          <BottomSheetGrabHandle
-            ref={grabHandleRef}
-            isHeaderFloating={isHeaderFloating}
-            {...metaAttribute({ name: ComponentIds.BottomSheetGrabHandle })}
-            {...bind()}
-          />
-          {children}
-        </BaseBox>
-      </BottomSheetSurface>
+      {isMounted ? (
```
chaitanyadeorukhkar, line 415: Why do we need to conditionally render this?

packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
archie252000: It is because in the documentation of [FloatingFocusManager](https://floating-ui.com/docs/FloatingFocusManager) the modal prop is there to ensure that the focus remains within the floating component. It does not primarily indicate that the component is modal but the behavior of "focus" is modal.

chaitanyadeorukhkar: Makes sense

packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
chaitanyadeorukhkar: Oh okay. @anuraghazra can you cross check this once?

archie252000:
@chaitanyadeorukhkar yeah, sure. I will look into it.

packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
anuraghazra: This might break integration with Dropdown, check the above check where we have `!isMounted && !isInsideDropdown`

The bottomsheet needs to be rendered and not unmount when using with Dropdown.

archie252000: @anuraghazra Oh, ok. Do you think we can use ```isOpen``` ?

archie252000: @anuraghazra So you mean when ```!isMounted == true``` and ```!isInsideDropdown == false``` in that case we should not return ```null``` in the conditional rendering?

anuraghazra: @archie252000 never mind that logic, this is for an edge case where imagine the bottomsheet has a input inside of it and you close the bottomsheet while being focused on the input in that case we want to destroy the react tree so the input is blurred. 

Your code should work as is.

anuraghazra (approved):
LGTM

packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
archie252000: @anuraghazra I found a solution that is passing the test for ```<BottomSheet /> › should compose with Dropdown multi select```. Basically Conditionally wrapping the contents of BottomSheet with FloatingFocusManager instead of conditionally rendering the whole component. What do you think about that?

snitin315 (commented):
I'll finish it.

chaitanyadeorukhkar:
@snitin315 we also need to expose `zIndex` as a prop in this PR after using floating UI. Similar to #1482

snitin315:
Thanks for the PR, closing in favor of #1489

PR #1461, opened 2023-07-21
feat: strip off logs & errors in production builds

snitin315:
- Fixes https://github.com/razorpay/blade/issues/272
- Wrapped all errors inside a `__DEV__` flag.
- Create two different bundles `index.development.web.js` and `index.production.web.js`.
- Updated `package.json` exports to [include `development` & `production` exports](https://nodejs.org/api/packages.html#resolving-user-conditions).
- [Webpack automatically resolves to the correct bundle based on the `mode` option in config](https://webpack.js.org/guides/package-exports/#providing-devtools-or-production-optimizations). Users don't have to do anything.
- In react native, `__DEV__` is `true` in debug mode & `false` in release mode. We don't strip but suppress the error in case of react native.

### DIFF
--- a/packages/blade/jest.web.config.js
+++ b/packages/blade/jest.web.config.js
@@ -24,6 +24,9 @@ const baseConfig = {
     '^\\~utils/(.*)': '<rootDir>/src/utils/$1',
     '^\\~tokens/(.*)': '<rootDir>/src/tokens/$1',
   },
+  globals: {
+    __DEV__: true,
+  },
 };
 
 module.exports = {
--- a/packages/blade/package.json
+++ b/packages/blade/package.json
@@ -86,9 +86,9 @@
     "types:copy-declarations:native": "copyfiles -u 1 \"src/**/*.d.ts\" build/types/native",
     "build:generate-root-imports": "rollup -c && node ./scripts/generateRootImports.js",
     "build:clean-declarations": "rm -rf build/types",
-    "build:react-dev": "FRAMEWORK=REACT NODE_ENV=development rollup -c && node ./scripts/generateRootImports.js",
-    "build:react-prod": "FRAMEWORK=REACT NODE_ENV=production rollup -c && node ./scripts/generateRootImports.js",
-    "build:react-native": "FRAMEWORK=REACT_NATIVE rollup -c && node ./scripts/generateRootImports.js",
+    "build:react-dev": "FRAMEWORK=REACT NODE_ENV=development rollup -c",
+    "build:react-prod": "FRAMEWORK=REACT NODE_ENV=production rollup -c",
+    "build:react-native": "FRAMEWORK=REACT_NATIVE rollup -c",
     "build:generate-css-variables": "GENERATE_CSS_VARIABLES=true FRAMEWORK=REACT rollup -c && node ./scripts/generateCssThemeVariables.js",
     "build:clean-theme-bundle": "rm -rf build/js-bundle-for-css",
     "react-native:get-stories": "sb-rn-get-stories --config-path=./.storybook/react-native && yarn prettier --write ./.storybook/react-native/storybook.requires.js",
@@ -254,11 +254,7 @@
   "bundlemon": {
     "files": [
       {
-        "friendlyName": "Development Web Components",
-        "path": "./build/components/index.development.web.js"
-      },
-      {
-        "friendlyName": "Production Web Components",
+        "friendlyName": "Web Components",
         "path": "./build/components/index.production.web.js"
       },
       {
@@ -278,11 +274,7 @@
         "path": "./build/css/*.css"
       },
       {
-        "friendlyName": "Development Web Utils",
-        "path": "./build/utils/index.development.web.js"
-      },
-      {
-        "friendlyName": "Production Web Utils",
+        "friendlyName": "Web Utils",
         "path": "./build/utils/index.production.web.js"
       },
       {
--- a/packages/blade/package.json
+++ b/packages/blade/package.json
@@ -39,7 +39,9 @@
       },
       "default": {
         "types": "./build/components/index.d.ts",
-        "default": "./build/components/index.web.js"
+        "development": "./build/components/index.development.web.js",
+        "production": "./build/components/index.production.web.js",
+        "default": "./build/components/index.development.web.js"
       }
     },
     "./tokens": {
@@ -49,7 +51,9 @@
       },
       "default": {
         "types": "./build/tokens/index.d.ts",
-        "default": "./build/tokens/index.web.js"
+        "development": "./build/tokens/index.development.web.js",
+        "production": "./build/tokens/index.production.web.js",
+        "default": "./build/tokens/index.development.web.js"
       }
     },
     "./utils": {
@@ -59,7 +63,9 @@
       },
       "default": {
         "types": "./build/utils/index.d.ts",
-        "default": "./build/utils/index.web.js"
+        "development": "./build/utils/index.development.web.js",
+        "production": "./build/utils/index.production.web.js",
+        "default": "./build/utils/index.development.web.js"
       }
     }
   },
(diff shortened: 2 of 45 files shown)


packages/blade/package.json
```diff
@@ -246,16 +254,20 @@
   "bundlemon": {
     "files": [
       {
-        "friendlyName": "Web Components",
-        "path": "./build/components/index.web.js"
+        "friendlyName": "Development Web Components",
```
kamleshchandnani, line 257: shall we keep a track of dev bundles? this might bloat up bundlemon table?

packages/blade/rollup.config.js
```diff
@@ -61,13 +62,17 @@ const getWebConfig = ({ exportCategory }) => ({
...
+      file: `${outputRootDirectory}/${exportCategory}/index.${process.env.NODE_ENV}.web.js`,
       format: 'esm',
       sourcemap: true,
     },
   ],
   external: (id) => id.includes('@babel/runtime'),
   plugins: [
+    pluginReplace({
+      __DEV__: process.env.NODE_ENV !== 'production',
+      preventAssignment: true,
```
kamleshchandnani, line 74: do we really need this? because we don't do assignments and this will always be set via env vars 🤔

packages/blade/src/components/ActionList/actionListUtils.ts
```diff
@@ -180,25 +182,29 @@ const validateActionListItemProps = ({
...
   React.Children.map(leading, (child) => {
-    if (
-      !isValidAllowedChildren(child, componentIds.ActionListItemIcon) &&
-      !isValidAllowedChildren(child, componentIds.ActionListItemText) &&
-      !isValidAllowedChildren(child, componentIds.ActionListItemAsset)
-    ) {
-      throw new Error(
-        `[ActionListItem]: Only ${componentIds.ActionListItemIcon}, ${componentIds.ActionListItemAsset}, and ${componentIds.ActionListItemText} are allowed in leading prop`,
-      );
+    if (__DEV__) {
```
kamleshchandnani, line 198: put the entire children.map under the dev check? why to even iterate if its not dev mode

packages/blade/src/components/ActionList/actionListUtils.ts
```diff
@@ -180,25 +182,29 @@ const validateActionListItemProps = ({
   trailing: ActionListItemProps['trailing'];
 }): void => {
   React.Children.map(trailing, (child) => {
```
kamleshchandnani, line 184: put the entire children.map under the dev check? why to even iterate if its not dev mode

packages/blade/src/components/Accordion/AccordionButton.native.tsx
```diff
@@ -79,8 +79,10 @@ const _AccordionButton = ({ index, icon: Icon, children }: AccordionButtonProps)
       <Icon size="medium" color={iconColor} marginRight="spacing.3" marginY="spacing.2" />
     );
 
-    if (_index && _icon) {
-      throw new Error(`[Blade: Accordion]: showNumberPrefix and icon shouldn't be used together`);
+    if (__DEV__) {
```
kamleshchandnani, line 82: so when the mode is prod this is evaluated to false and hence the whole block is removed by the minifier?

packages/blade/package.json
```diff
@@ -80,7 +86,8 @@
     "types:copy-declarations:native": "copyfiles -u 1 \"src/**/*.d.ts\" build/types/native",
     "build:generate-root-imports": "rollup -c && node ./scripts/generateRootImports.js",
     "build:clean-declarations": "rm -rf build/types",
-    "build:react": "FRAMEWORK=REACT rollup -c && node ./scripts/generateRootImports.js",
+    "build:react-dev": "FRAMEWORK=REACT NODE_ENV=development rollup -c && node ./scripts/generateRootImports.js",
```
kamleshchandnani, line 89: @snitin315 we shall remove `node ./scripts/generateRootImports.js` from all the places. there's already `build:generate-root-imports` which does this job so this is redundant and has no effect but increases the build time. Its not related to your changes but some place this got missed and added to package.json in the past I feel, can you remove it now?

snitin315: @kamleshchandnani Sure, I'll remove it.

packages/blade/src/components/Accordion/AccordionButton.native.tsx
snitin315: Yes.

packages/blade/rollup.config.js
snitin315: Yes, we need to set __DEV__ during build time.

packages/blade/package.json
snitin315: I think we can keep the prod bundle only, initially I added this to check the diff in dev and prod sizes.

packages/blade/rollup.config.js
kamleshchandnani: the DEV is fine, I was talking about `preventAssignment`, the docs says that we need this to avoid things like `process.env.NODE_ENV` getting reset when the program is running.

snitin315: @kamleshchandnani No, we don't really need this but rollup was logging these warnings so I added it.

Screenshot 2023-07-26 at 3 24 56 PM

packages/blade/src/components/Accordion/AccordionButton.native.tsx
snitin315: | Dev bundle | Prod Bundle |
| ------------|-------------|
| Screenshot 2023-07-26 at 3 32 15 PM | Screenshot 2023-07-26 at 3 33 35 PM |

kamleshchandnani: yes i had checked out this branch and checked locally :D

kamleshchandnani (approved):

kamleshchandnani (dismissed):

packages/blade/jest.web.config.js
```diff
@@ -24,6 +24,9 @@ const baseConfig = {
     '^\\~utils/(.*)': '<rootDir>/src/utils/$1',
     '^\\~tokens/(.*)': '<rootDir>/src/tokens/$1',
   },
+  globals: {
+    __DEV__: true,
```
chaitanyadeorukhkar, line 28: Won't our consumer need to also add this in their config? What happens if this is missing? Does it simply ignore the error or throws error that __DEV__ not found?

packages/blade/package.json
```diff
@@ -247,15 +255,15 @@
...
+        "path": "./build/components/index.production.web.js"
       },
       {
         "friendlyName": "React Native Components",
         "path": "./build/components/index.native.js"
       },
       {
         "friendlyName": "Web Tokens",
-        "path": "./build/tokens/index.web.js"
+        "path": "./build/tokens/index.development.web.js"
```
chaitanyadeorukhkar, line 266: Shouldn't this be production?
```suggestion
        "path": "./build/tokens/index.production.web.js"
```

packages/blade/rollup.config.js
```diff
@@ -61,13 +62,17 @@ const getWebConfig = ({ exportCategory }) => ({
   input: `${inputRootDirectory}/${exportCategory}/index.ts`,
   output: [
     {
-      file: `${outputRootDirectory}/${exportCategory}/index.web.js`,
+      file: `${outputRootDirectory}/${exportCategory}/index.${process.env.NODE_ENV}.web.js`,
```
chaitanyadeorukhkar, line 65: Lets have a check  in rollup to ensure NODE_ENV is set? Otherwise it might end up creating `index.undefined.web.js` 😆

packages/blade/package.json
snitin315: Yes.

packages/blade/rollup.config.js
snitin315: I've passed `NODE_ENV` in our build scripts. Ideally, we should fallback to development:
```suggestion
      file: `${outputRootDirectory}/${exportCategory}/index.${process.env.NODE_ENV || 'development'}.web.js`,
```

chaitanyadeorukhkar (approved):

PR #1465, opened 2023-07-24
fix: coverage script parent selector & add highlight nodes option

anuraghazra:
- Fixed parent selector in coverage script plugin
- Added highlight node option 

https://github.com/razorpay/blade/assets/35374649/df4c0ede-c440-4be8-9ce2-f2e7149d8915

### DIFF
--- a/packages/blade-coverage-extension/background_script.js
+++ b/packages/blade-coverage-extension/background_script.js
@@ -1,5 +1,5 @@
 /* eslint-disable no-undef */
-const calculateBladeCoverage = () => {
+const calculateBladeCoverage = (shouldHighlightNodes) => {
   /**
    * Checks if DOM node is hidden or not
    */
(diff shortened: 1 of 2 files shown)


kamleshchandnani:
can we just highlight non-blade nodes? otherwise it'll lead to a lot of clutter for consumers to figure out what they shall focus upon

packages/blade-coverage-extension/background_script.js
```diff
@@ -40,17 +40,27 @@ const calculateBladeCoverage = () => {
...
   const bladeNodes = bladeNodeElements.length;
   const bladeCoverage = Number(((bladeNodes / totalNodes) * 100).toFixed(2));
 
+  if (shouldHighlightNodes) {
+    totalNodeElements.forEach((node) => {
+      node.style.outline = '1px solid rgba(255, 0, 0, 0.5)';
+    });
+
+    bladeNodeElements.forEach((node) => {
+      node.style.outline = '1px solid green';
```
kamleshchandnani, line 56: can we just highlight non-blade nodes? otherwise it'll lead to a lot of clutter for consumers to figure out what they shall focus upon

kamleshchandnani (dismissed):

packages/blade-coverage-extension/background_script.js
anuraghazra: makes sense.

kamleshchandnani (approved):

Issue #1286, opened 2023-06-12
`height` prop for Card component


InvincibleZeal:
**Problem Statement** 

I built a carousel that has `Card` components as tiles. Since the `Card` API doesn't support `height` as a prop, I can't make it span the entire height of the parent `Carousel Slider` Box due to which the cards look uneven at different points. Setting a fixed height to a parent `Box` component looks weird for this use case. 

<img width="1290" alt="Screenshot 2023-06-12 at 11 32 26 AM" src="https://github.com/razorpay/blade/assets/42371840/ac0d9834-fd2c-4613-81a6-85a32f513743">

Check the implementation in [this PR](https://github.com/razorpay/frontend-website/pull/2160)

**Expectation**

Provide `height` as a prop on the `Card` and `CardBody` (for `justifyContent: space-between`) API so it can be set as 100% which will make it stretch over the Carousel Slider height making all the cards look even.

<img width="1291" alt="Screenshot 2023-06-12 at 11 32 12 AM" src="https://github.com/razorpay/blade/assets/42371840/41617fd6-684b-46ff-bb7b-c61ccc52dac4">

nikhil-2911:
Can you please assign this issue to me?

chaitanyadeorukhkar:
@InvincibleZeal you can have a Box inside CardBody with a fixed height, that should solve this, right? The dimensions of the Card should be decided by its content, having a Box of a specific height inside CardBody makes sense here

cseas:
Is there an issue with providing the height prop? Agree that there could be many different ways to work around this but we already have a workaround in place. Issue was created because developers experienced friction in implementing designs due to limitations in Blade API.

chaitanyadeorukhkar:
@cseas there is no issue with it but better to keep separation of concern. Since already have a layout component, it would be better if that takes care of such instances otherwise we will bloat up our component APIs for things that have straightforward solutions. Card with an empty body doesn't do much and will anyway always have some layout inside it. We can anyway consider adding this if we see more friction or more similar asks.

PranjalMaithani:
Same issue here. I have a flex/grid container which has 3 cards. I want to use flex with `space-between` on the card component to place the 'Know more' link at the bottom. But the issue is that the `CardBody` div is not occupying 100% of the height of the parent `Card` component.

[...]

PR #1470, opened 2023-07-25
docs(blade): added carousel API

anuraghazra:
### [View Formatted](https://github.com/razorpay/blade/blob/anu/carousel-api/packages/blade/src/components/Carousel/_decisions/decisions.md)


packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,150 @@
...
+    <Testimonial />
+  </CarouselItem>
+
+  <CarouselItem>
+    <Testimonial />
+  </CarouselItem>
+</Carousel>
+```
+
+<img src="./example-usage-1.png" width="70%" />
```
chaitanyadeorukhkar, line 69: The example image has navigation buttons that point in the same direction

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Usage with Card
```
chaitanyadeorukhkar, line 71: Why do we need to mention this in API doc?

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,150 @@
...
+    <TestimonialCard />
+  </CarouselItem>
+</Carousel>
+```
+
+## Accessibility
+
+We need to follow the APG Pattern For [Carousel](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-2-tablist/)
+
+## Motion
```
chaitanyadeorukhkar, line 121: Does the navigation button go in a disabled state when reached either ends or does it keep looping?

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,150 @@
...
+
+https://github.com/razorpay/blade/assets/35374649/6876f348-1b0a-4e9a-800a-300025f54b5c
+
+## Open Questions
+
+- Should we call it `navigationButtonStyle` or `navigationButtonVariant`?
```
chaitanyadeorukhkar, line 127: `navigationButtonVariant`. Style would make someone thing it accepts a style object

chaitanyadeorukhkar: Also we call it `indicatorVariant` so better keep terminology same?

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,150 @@
...
+**web**
+
+On mweb & dweb we can use [CSS scroll snap](https://ishadeed.com/article/css-scroll-snap/) feature, this covers all our common usecase with bleed smooth scrolling etc.
+
+**react-native**
+
+on rnweb, we cannot use CSS scroll snap so might might need to opt for a library.
+But further exploration lead me to this [SO question](https://stackoverflow.com/questions/39849648/horizontal-scrollview-snapping-react-native) and [this expo example](https://snack.expo.io/H1CnjIeDb), turns our that ScrollView does support scroll snapping so if this works for our usecase it will be helpful.
+
+But we might need to see if this might cause an performance issue since we won't be using FlatList.
```
chaitanyadeorukhkar, line 140: Why can't we use FlatList? It has horizontal scroll

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Libraries
+
+**web**
+
+On mweb & dweb we can use [CSS scroll snap](https://ishadeed.com/article/css-scroll-snap/) feature, this covers all our common usecase with bleed smooth scrolling etc.
+
+**react-native**
+
+on rnweb, we cannot use CSS scroll snap so might might need to opt for a library.
```
chaitanyadeorukhkar, line 137: We're not using rnweb, we're using just react native. Sounds misleading.

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Accessibility
+
+We need to follow the APG Pattern For [Carousel](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-2-tablist/)
+
+## Motion
+
+https://github.com/razorpay/blade/assets/35374649/6876f348-1b0a-4e9a-800a-300025f54b5c
+
+## Open Questions
```
chaitanyadeorukhkar, line 125: Are we planning to support drag behaviour? If yes, can we add something about it in the doc. If not, why not?

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Put it as an example on how to compose commonly used components together.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: From the motion docs, it seems that it loops - but i'll confirm the UX once with RK (ideally it should loop tho)

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Yes, also raised this on figma, variant seems more appropriate

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Yes, Drag will be supported on mobile devices and is already supported with CSS scroll snap & that ScrollView thing.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: FlatList doesn't support `snapToInterval, snapToAlignment`

Two things: 
- Even if snapToAlignment was supported in FlatList - we might again need to do hacks to convert the JSX to FlatList accepted data array. 
- Given that carousels should not have hundreds of elements, I wonder if FlatList is even needed or not. At most it might have 10 (even that sounds way too many)

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: bug in figma design, raised.

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,150 @@
...
+| children                 | `CarouselItem`                         | `undefined`                              | Carousel slides                                                                                                                                                                              | ✅        |
+| autoPlay                 | `boolean`                              | `false`                                  | If true, the carousel will autoplay                                                                                                                                                          |          |
+| visibleItems             | `1,2,3,auto`                           | `1`                                      | Total number of carousel items to show at once, if set to 1,2 or 3 all the CarouselItem's width will remain the same but if set to `auto` the carousel items will take up space responsively |          |
```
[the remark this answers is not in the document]
anuraghazra: Yes. it will break given there is not enough space, we might need to change visibleItems to 1 on mobile devices.

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,150 @@
...
+| ------------------------ | -------------------------------------- | ---------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| children                 | `CarouselItem`                         | `undefined`                              | Carousel slides                                                                                                                                                                              | ✅        |
+| autoPlay                 | `boolean`                              | `false`                                  | If true, the carousel will autoplay                                                                                                                                                          |          |
```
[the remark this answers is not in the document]
anuraghazra: fine with either, but autoPlay is a more standard prop name here. 

Used by most major libs: 
- https://flickity.metafizzy.co/#autoplay
- http://kenwheeler.github.io/slick/
- https://www.npmjs.com/package/react-responsive-carousel

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,150 @@
...
+| visibleItems             | `1,2,3,auto`                           | `1`                                      | Total number of carousel items to show at once, if set to 1,2 or 3 all the CarouselItem's width will remain the same but if set to `auto` the carousel items will take up space responsively |          |
+| bleed                    | `none,right,left,both`                 | `none`                                   | Sets the carousel items to bleed on either side button                                                                                                                                       |          |
```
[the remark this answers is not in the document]
anuraghazra: Bleed is removed now, we have automatic bleed.

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,161 @@
...
+| children                 | `CarouselItem`                         | `undefined`                              | Carousel slides                                                                                                                                                                                              | ✅        |
+| autoPlay                 | `boolean`                              | `false`                                  | If true, the carousel will autoplay                                                                                                                                                                          |          |
+| visibleItems             | `1,2,3,auto`                           | `1`                                      | Total number of carousel items to show at once, if set to 1,2 or 3 all the CarouselItem's width will remain the same but if set to `auto` the carousel items will take up space responsively                 |          |
```
anuraghazra, line 19: should we make the visibleItems: `1 | 2 | 3 | auto`? (`auto` being the default.)

Not really fond of using `undefined` as a valid value to switch between variants. (auto also makes more sense i think)

iAmServerless: auto should be default. Majority of the usecases are where 80% area on mobile device is the item and we want to show some part of next and previous item. Also they might also needs some blur effect.

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,161 @@
...
+
+<img src="./example-visibleitems-1.png" width="70%" />
+
+v2: visibleItems: 2 
+
+<img src="./example-visibleitems-2.png" width="70%" />
+
+v2: visibleItems: 3
+
+<img src="./example-visibleitems-3.png" width="70%" />
```
iAmServerless, line 124: Can we add few mobile variants as well?

Kushagra-ag:
@anuraghazra would it be possible to control the size of the carousel slides based on whether the slide is focussed or not? (Mobile usecase only)

Attaching image for example - Here the slide that is focussed is slightly larger than the other slides

Screenshot 2023-08-04 at 3 35 54 PM

anuraghazra:
> would it be possible to control the size of the carousel slides based on whether the slide is focussed or not?

You can probably, use the onChange to get the current slide index and add transform: scale to the selected slide from your end, but this isn't a pattern that we are providing out of the box right now.

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,161 @@
...
+| visibleItems             | `1,2,3,auto`                           | `1`                                      | Total number of carousel items to show at once, if set to 1,2 or 3 all the CarouselItem's width will remain the same but if set to `auto` the carousel items will take up space responsively                 |          |
+| carouselItemWidth        | `ResponsiveProp<width>`                        | `undefined`                              | Sets the width of the carousel items, this can be used with visibleItems: auto to achive automatic bleed                                                                                                     |          |
+| carouselItemAlignment    | `start, end, center`                   | `start`                                  | Sets the align-items CSS property on carousel container which specifies how the carousel items will align if their heights are different                                                                     |          |
```
divyanshu013, line 21: What if people want to make all items have equal height (same as height of the largest carousel item)?

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,161 @@
...
+| shouldAddStartEndMargin  | `boolean`                              | `false`                                  | If true, adds extra margin before and after the first/last slides so that they align in center, this prop can be used to achive carousel which is in [GST page](https://razorpay.com/gst-number-search/pan/) |          |
+| showIndicators           | `boolean`                              | `true`                                   | Toggles the visibility of indicators                                                                                                                                                                         |          |
+| showOverlay              | `boolean`                              | `false`                                  | Toggles the visibility of overlay                                                                                                                                                                            |          |
```
divyanshu013, line 24: Should we make this more specific? eg. scroll overlay

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,161 @@
...
+| navigationButtonStyle    | `filled,stroke`                        | `filled`                                 | Styles to be used depending on the emphasis you want to give to the navigation buttons                                                                                                                       |          |
+| indicatorVariant         | `gray,white,blue`                      | `gray`                                   | Variants to be used depending on the background and the emphasis you want to give to the indicator.                                                                                                          |          |
+| onChange                 | `(slideIndex: number) => void`         | `undefined`                              | Callback which gets fired everytime a slide changes, useful for analytic events (only runs on user interaction not on autoPlay)                                                                              |          |
```
divyanshu013, line 30: On few other components we made `onChange` terminology behavior specific (like `onExpandChange`). Should we do that here as well?

divyanshu013 (approved):
Looks good to me, minor queries

lakshay-dutt:
@anuraghazra  possible to add a `scrollToIndex` handler?
Few sections have external elements that control the carousel scroll.

Screenshot 2023-08-07 at 3 08 20 PM

anuraghazra:
> possible to add a scrollToIndex handler?
We can look into it, but even with scrollToIndex it won't be trivial to recreate the POS page carousel since the interaction is quite different than any normal carousel.  

- It's controlled via the window scroll
- doesn't snap to specific slides, while blade uses scroll snaps
- there are no buttons/indicators instead clicking on the side of the pages causes the next/prev slide to appear

It might be easier to recreate the POS page with Chip (when it releases) + Card component than retrofitting carousel.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Right, i think for that we can also expose 'stretch' and people can do: 

```
      <CarouselItem>
        <Box backgroundColor="brand.primary.300" height="100%" />
      </CarouselItem>
```

Though in Cards we should expose height prop then. otherwise they won't be able to set 100% height

lakshay-dutt:
@anuraghazra I mentioned POS just as one the examples, there are cases in the past in which the active item of the carousel was controlled through an external element.

anuraghazra:
> active item of the carousel was controlled through

Sure if you just have the usecase for `scrollToIndex` i think we can provide that.

packages/blade/src/components/Carousel/_decisions/decisions.md
kamleshchandnani, line 30: how about `ref`? for example, I want to scroll into a particular carousel item?

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,161 @@
...
+| accessibilityLabel                 | `string`         | `undefined`                              | Accessibility label for the carousel, this will let screen reader users know what content the carousel holds (eg: "Product carousel")                                                                              |          |
+
+Additionally we will also expose few styled props, namely margin/width etc so that users can add negative margins to create end to end bleed carousel.
```
kamleshchandnani, line 33: why do we need this now negative margin thing? because we already have bleed capability right?

packages/blade/src/components/Carousel/_decisions/decisions.md
divyanshu013: Cool, might be good idea to validate so usecases like #1286 are easier to build as carousels

kamleshchandnani: yes we anyways have an issue for adding height to card

anshulsahni (dismissed):

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,161 @@
...
+| showIndicators           | `boolean`                              | `true`                                   | Toggles the visibility of indicators                                                                                                                                                                         |          |
+| showOverlay              | `boolean`                              | `false`                                  | Toggles the visibility of overlay                                                                                                                                                                            |          |
+| overlayColor             | `BrandColorTokens, SurfaceColorTokens` | `surface.background.level1.highContrast` | Changes the color of the overlay, so that consumers can blend the overlay with the background color                                                                                                          |          |
```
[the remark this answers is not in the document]
anuraghazra: Sure i guess that makes sense.

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,161 @@
...
+| navigationButtonPosition | `bottom,side`                          | `bottom`                                 | Sets the position of navigation button                                                                                                                                                                       |          |
+| navigationButtonSpacing  | `SpacingTokens`                        | `spacing.4`                              | Spacing between navigation button and slides when bleed is set to none                                                                                                                                       |          |
```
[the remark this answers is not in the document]
anuraghazra: Will get this checked with RK if this is even needed

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: > @anuraghazra I mentioned POS just as an example, there are cases in the past in which the active item of the carousel was controlled through an external element.

Yeah usecase hain, we can probably expose `goToSlide` method via a ref.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Bleed is for showing partial carousel items, Negative margin will be needed to make the carousel container stretch end to end

```
<body marginX="20px">
  <Carousel marginX="-20px" />
</body>
```

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: We might just remove this, we will just have overlayColor, but then we can rename it to `scrollOverlayColor`

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: I think in here it should be fine, there is no ambiguity what is "changing" thus onChange makes sense. I feel `onSlideChange` might be redundant.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: added

packages/blade/src/components/Carousel/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Added

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Removed. and renamed to scrollOverlayColor

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Will add the height prop

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: In normal variants it loops, but in case of `visibleItems: autofit & shouldAddStartEndSpacing` we remove the next/prev button.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Removed.

packages/blade/src/components/Carousel/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: We set it to 1 on mobile devices.

packages/blade/src/components/Carousel/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: yep autoPlay is more relatable and known here

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -21,16 +21,15 @@ Carousel is a UI component that allows the display and navigation of a set of co
...
 | onChange                 | `(slideIndex: number) => void`         | `undefined`                              | Callback which gets fired everytime a slide changes, useful for analytic events (only runs on user interaction not on autoPlay)                                                                              |          |
 | accessibilityLabel       | `string`                               | `undefined`                              | Accessibility label for the carousel, this will let screen reader users know what content the carousel holds (eg: "Product carousel")                                                                        |          |
 
-Additionally we will also expose few styled props, namely margin/width etc so that users can add negative margins to create end to end bleed carousel.
+- Additionally we will also expose few styled props, namely margin/width etc so that users can add negative margins to create end to end bleed carousel.
+- To let consumers scroll to any specific slide we will expose a `ref` with `goToSlide` method.
```
divyanshu013, line 32: Won't it be better to add controlled behavior instead 🤔 That can handle other cases as well where users might want more control

anuraghazra: Agree, will need to add controlled/uncontrolled states for that then. 

I think let's take it as an enhancement in upcoming sprints, need to fastrack the release for now.

packages/blade/src/components/Carousel/_decisions/decisions.md
kamleshchandnani: yes we shall

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,175 @@
...
+> NOTE: To ensure bleed happens you need to specify a fixed amount of carousel item's width via the `carouselItemWidth` prop. Otherwise the carousel item will take up the 100% of the container width and won't bleed
+
+
+v1: visibleItems: autofit + auto bleed + shouldAddStartEndMargin
+
+<img src="./example-visibleitems-undefined-with-margin.png" width="70%" />
+
+v1: visibleItems: autofit + auto bleed
+
+<img src="./example-visibleitems-undefined-without-margin.png" width="70%" />
```
kamleshchandnani, line 100: finally will we show the indicators here? I think we decided we'll hide it? can we update the image?

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,175 @@
...
+| autoPlay                 | `boolean`                              | `false`                                  | If true, the carousel will autoplay                                                                                                                                                                          |          |
+| visibleItems             | `1,2,3,autofit`                        | `1`                                      | Total number of carousel items to show at once, if set to 1,2 or 3 all the CarouselItem's width will remain the same but if set to `autofit` the carousel items will take up available space responsively    |          |
+| carouselItemWidth        | `ResponsiveProp<width>`                | `undefined`                              | Sets the width of the carousel items, this can be used with visibleItems: auto to achive automatic bleed                                                                                                     |          |
```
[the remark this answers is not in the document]
anuraghazra: This prop needs to be on Carousel, if we put it on CarouselItem people will be able to give different widths to each carousel item which is not expected. 

CarouselItem component will have no props it's just to hold the content, like a slot.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Outdated screenshot, we won't show indicators here. will update.

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,175 @@
...
+| children                 | `CarouselItem`                         | `undefined`                              | Carousel slides                                                                                                                                                                                              | ✅        |
+| autoPlay                 | `boolean`                              | `false`                                  | If true, the carousel will autoplay                                                                                                                                                                          |          |
+| visibleItems             | `1,2,3,autofit`                        | `1`                                      | Total number of carousel items to show at once, if set to 1,2 or 3 all the CarouselItem's width will remain the same but if set to `autofit` the carousel items will take up available space responsively    |          |
```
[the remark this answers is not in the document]
anuraghazra: all the carousel item widths will always be same, and consumers can change it via the carouselItemWidth prop. 

Modified the text to: 


Total number of carousel items to show at once, can be set to 1,2,3 or if set to `autofit` the carousel items will take up 100% of the container, consumers can pass fixed width to carousel items via the carouselItemWidth prop to achive bleed.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Updated and added note about the indicator

packages/blade/src/components/Carousel/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: oh yeah.

kamleshchandnani (approved):

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,176 @@
...
+| onChange                 | `(slideIndex: number) => void`         | `undefined`                              | Callback which gets fired everytime a slide changes, useful for analytic events (only runs on user interaction not on autoPlay)                                                                              |          |
+| accessibilityLabel       | `string`                               | `undefined`                              | Accessibility label for the carousel, this will let screen reader users know what content the carousel holds (eg: "Product carousel")                                                                        |          |
+
+- Additionally we will also expose few styled props, namely margin/width etc so that users can add negative margins to create end to end bleed carousel.
+- To let consumers scroll to any specific slide we will expose a `ref` with `goToSlide` method.
```
pcprashant, line 32: Should we provide next, previous method ? If we intend to provide only ref then we can provide these methods on ref also.

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,176 @@
...
+| ------------------------ | -------------------------------------- | ---------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | -------- |
+| children                 | `CarouselItem`                         | `undefined`                              | Carousel slides                                                                                                                                                                                              | ✅        |
+| autoPlay                 | `boolean`                              | `false`                                  | If true, the carousel will autoplay                                                                                                                                                                          |          |
```
pcprashant, line 18: Should we provide delay prop to decide the delay ? Also another prop to identify if it needs to be a loop ?

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,176 @@
...
+
+On mweb & dweb we can use [CSS scroll snap](https://ishadeed.com/article/css-scroll-snap/) feature, this covers all our common usecase with bleed smooth scrolling etc.
+
+**react-native**
+
+on rnweb, we cannot use CSS scroll snap so might might need to opt for a library.
+But further exploration lead me to this [SO question](https://stackoverflow.com/questions/39849648/horizontal-scrollview-snapping-react-native) and [this expo example](https://snack.expo.io/H1CnjIeDb), turns our that ScrollView does support scroll snapping so if this works for our usecase it will be helpful.
+
+But we might need to see if this might cause an performance issue since we won't be using FlatList.
+
```
pcprashant, line 167: How does the swipe behaviour work on mobile web ? Also if there are no indicators on dweb can clicking on the items will slide the items ?

packages/blade/src/components/Carousel/_decisions/decisions.md
```diff
@@ -0,0 +1,176 @@
...
+| children                 | `CarouselItem`                         | `undefined`                              | Carousel slides                                                                                                                                                                                              | ✅        |
+| autoPlay                 | `boolean`                              | `false`                                  | If true, the carousel will autoplay                                                                                                                                                                          |          |
+| visibleItems             | `1,2,3,autofit`                        | `1`                                      | Total number of carousel items to show at once, can be set to 1,2,3 or if set to `autofit` the carousel items will take up 100% of the container, consumers can pass fixed width to carousel items via the carouselItemWidth prop to achive bleed.     |          |
```
pcprashant, line 19: If visible items are set to 3, then how many items will slide on clicking navigation button ?

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: We will actually look into adding controlled state in the carousel, so no need to expose these refs.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: If the usecase comes in we will add a delay prop but right now we've standardized it to 6s. 

> Also another prop to identify if it needs to be a loop
The looping behaviour is standardised from design side only so probably no.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: Yes swipe will be available. 

> Also if there are no navigation buttons on dweb can clicking on the items will slide the items ?
Nav buttons are mandatory in dweb.

packages/blade/src/components/Carousel/_decisions/decisions.md
anuraghazra: 3 items.

PR #1471, opened 2023-07-25
feat(Button): expose `intent` prop

snitin315:
Screenshot 2023-07-26 at 12 51 45 PM


packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -29,6 +29,8 @@ type ButtonCommonProps = {
   rel?: BaseButtonProps['rel'];
   variant?: 'primary' | 'secondary' | 'tertiary';
   size?: 'xsmall' | 'small' | 'medium' | 'large';
+  intent?: 'positive' | 'negative' | 'notice';
+  contrast?: 'low' | 'high';
```
saurabhdaware, line 33: I don't think so we'll need contrast prop. We can achieve the same result by passing `variant="secondary"` and `intent="negative"` right?

snitin315: @saurabhdaware  If `variant` and `intent` both are defined, `intent` takes priority which makes `variant` ineffective/redundant.

snitin315: > Can we change this? so variant will only decide if the button is filled, outlined etc. But intent will decide the color. And when no intent is passed, the color will be primary blue that we have right now 🤔

@saurabhdaware No. variant secondary and low contrast have different designs.

saurabhdaware: I think that contrast was decided for Alert component earlier. In our case of Button, we don't really need contrast. We need our intent color to support different variants of Button 🤔 

In the context of Button, the contrast might have different meaning (e.g. I would expect blue button to be white button on passing contrast the way we do in text component)

Can discuss this with the team once and take a call.

snitin315: Update: we decided to set contrast `high` by default and expose only `intent`.

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -105,6 +110,7 @@ const _Button: React.ForwardRefRenderFunction<BladeElementRef, ButtonProps> = (
       size={size}
       type={type}
       variant={variant}
+      intent={isAllowedIntent ? intent : undefined}
```
saurabhdaware, line 113: image

Can you check how we can make this work? Basically setting `variant` and `intent` both. So we get outlined button when variant is secondary. Might require some changes in BaseButton https://razorpay.slack.com/archives/G01B3LQ9H0W/p1690350610167359?thread_ts=1690347620.242949&cid=G01B3LQ9H0W

snitin315: @saurabhdaware From the discussion I recall that it is not something we want to allow.

PR #1473, opened 2023-07-26
chore: upgrde jest to v29

snitin315:
### Dependency Changes

> **Note**
> Didn't upgrade `@testing-library/react` to v13+ since it drops support for React 17 and earlier.

- Upgraded the following dependencies to their latest version:
  - `@testing-library/react-hooks` to `v8`
  - `@testing-library/react-native` to `v12.1.3`
  - `babel-jest`, `jest`, and `jest-environment-jsdom` to `v29.6.1`
  - `jest-axe` to v8
  -  `jest-styled-components` to `v7.1.1`
  -  `react-native-reanimated` to `v2.14.4`, for compatibility with Jest v28+

### Jest Config Changes

- Configured `defaultIncludeHiddenElements: true` for `@testing-library/react-native` to keep the old behavior. [Since v12, all queries exclude elements hidden from accessibility by default.](https://callstack.github.io/react-native-testing-library/docs/migration-v12#1-all-queries-exclude-elements-hidden-from-accessibility-by-default)

- Added [`--forceExit`](https://jestjs.io/docs/next/cli#--forceexit) to the `test:react` and `test:react-native`.
  -  These scripts without `--forceexit` now fail on CI when Jest doesn't; shut down cleanly (Existing behavior, for [example BottomSheet tests](https://github.com/razorpay/blade/actions/runs/5691126042/job/15425725319?pr=1483#step:8:1924))
  - Ideally, we should detect the cause and refactor the corresponding tests, which can be tackled as a separate issue.
    
     Screenshot 2023-07-31 at 11 02 16 AM

- Added workaround for `jest-styled-components` snapshot serializer.
    - https://github.com/styled-components/jest-styled-components/issues/429

### Source Code Changes

- In the `makeAccessible.native.ts` util, explicitly add `accessible={true}` prop as well along with other accessibility props. 
   - [This was required because `*ByRole` queries now return only accessibility elements](https://callstack.github.io/react-native-testing-library/docs/migration-v12#2-byrole-queries-now-return-only-accessibility-elements), either explicitly marked with `accessible` prop or implicit ones where this status is derived from component type itself (e.g `Text`, `TextInput`, `Switch`, but not `View`).

### Test Changes

- In Jest v28, [Fake timers were refactored to allow passing options.](https://jestjs.io/docs/28.x/upgrading-to-jest28#fake-timers). 
  - Added `legacyFakeTimers: true` in tests which were breaking to use the old fake timers implementation instead of one backed by `@sinonjs/fake-timers`.  The default is `false`.
  
    ```diff
    - jest.useFakeTimers();
    + jest.useFakeTimers({
    +  legacyFakeTimers: true,
    + });
    ```
- `getByA11yValue` is deprecated, refactored  `ProgressBar.native.test.tsx` to use `expect(progressbar).toHaveAccessibilityValue({})`

  Screenshot 2023-07-29 at 9 15 54 AM

- `getAllByA11yRole`, `getByA11yLabel` aliases were removed in favor of `getAllByRole`, `getByLabelText` respectively.
  -  https://github.com/callstack/react-native-testing-library/pull/881

### Snapshot Changes

- As announced in the [Jest 28 blog post](https://jestjs.io/blog/2022/04/25/jest-28#future), Jest 29 has changed the default snapshot formatting to `{escapeString: false, printBasicPrototype: false}`. This makes snapshots both more readable and more copy-pasteable.

### GitHub Action Changes

- Refactored validate workflow to enable [Jest Sharding](https://jestjs.io/docs/next/cli#--shard), which has reduced the total workflow time by 3-4 minutes.


packages/blade/src/components/Divider/__tests__/__snapshots__/Divider.web.test.tsx.snap
```diff
@@ -1,196 +1,41 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`<Divider /> should render Divider within flex container 1`] = `
-.c0 {
-  display: -webkit-box;
-  display: -webkit-flex;
-  display: -ms-flexbox;
-  display: flex;
-}
```
anuraghazra, line 9: it seems that all the generated CSS got removed from snapshots this should not happen as we deliberately enabled snapshots to include the CSS rules. Maybe check styled-components repo for jest 29 guides

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
```diff
@@ -11,6 +11,8 @@ import type { AccessibilityMap, AccessibilityProps } from './types';
 export const makeAccessible = (props: Partial<AccessibilityProps>): Record<string, unknown> => {
   const newProps: Record<string, any> = {};
 
+  newProps.accessible = true;
```
anuraghazra, line 14: This might cause issues on react-native and change accessibility behaviours, because now anywhere we have used makeAccessible will cause that element to become a single selectable element and won't be able to select the nested elements. 

https://reactnative.dev/docs/accessibility#accessible

kamleshchandnani:
@snitin315 rebase with master first so chromatic optimisations can kick in?

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
snitin315: @anuraghazra But we are adding `accessible={true}` to the elements only which have other accessibility props, like [`accessibilityLabel`](https://reactnative.dev/docs/accessibility#accessibilitylabel), etc so it should be fine. Having other accessible props means, we do want to make the component accessible, right?

[Now, it is mandatory to have `accessible={true}`, otherwise `*ByRole` queries won't return these elements.](https://callstack.github.io/react-native-testing-library/docs/migration-v12#2-byrole-queries-now-return-only-accessibility-elements)

packages/blade/src/components/Divider/__tests__/__snapshots__/Divider.web.test.tsx.snap
snitin315: Fixed, it seems `jest-styled-components` is not compatible with jest v28+.
- https://github.com/styled-components/jest-styled-components/issues/429

packages/blade/patches/react-native-reanimated+2.12.0.patch
```diff
@@ -0,0 +1,34 @@
+diff --git a/node_modules/react-native-reanimated/lib/reanimated2/jestUtils.js b/node_modules/react-native-reanimated/lib/reanimated2/jestUtils.js
```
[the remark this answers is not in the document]
snitin315: Upgraded reanimated and it works!

I don't know why that issue is still open, I thought it wasn't fixed in master.

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
anuraghazra: Thik hai i'll test the a11y on my device once and let you know if this is causing some issue.

.github/workflows/blade-validate.yml
```diff
@@ -6,7 +6,7 @@ env:
   GITHUB_ACCESS_TOKEN: ${{ secrets.CI_BOT_TOKEN }}
 
 jobs:
-  validate:
+  lint:
```
snitin315, line 9:
```suggestion
  validate:
```

packages/blade/src/utils/makeAccessible/makeAccessible.native.ts
anuraghazra: Okay tested each component manually on RN, most of them looks good. 

There are some minor changes like in Indicator component now consumers won't be able to go into the text node and read out the "Success" text 

image

But i think this is fine because it still announces the accessibilityLabel as "Status ok". 
Anyway if we want to manually disable accessible prop we can just pass `accessible: false` when using makeAccessible, so we will just need to be a bit careful.

.github/actions/install-dependencies/action.yml
```diff
@@ -18,8 +18,14 @@ runs:
         key: ${{ runner.os }}-blade-${{ hashFiles('**/yarn.lock') }}
 
     # cache miss - install packages with `yarn --frozen-lockfile`
-    # cache hit - download packages from the cache
+    # cache hit - download packages from the cache and explicitly run postinstall script
```
anuraghazra, line 21: is this something newly changed or github is just showing the old diff?

packages/blade/patches/bundlemon+2.0.1.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/node_modules/bundlemon/lib/main/analyzer/pathUtils.js b/node_modules/bundlemon/lib/main/analyzer/pathUtils.js
+index e9ce416..1c41a0c 100644
+--- a/node_modules/bundlemon/lib/main/analyzer/pathUtils.js
++++ b/node_modules/bundlemon/lib/main/analyzer/pathUtils.js
+@@ -42,6 +42,7 @@ function getAllPaths(dirPath) {
+         const files = entries.filter((file) => !file.isDirectory()).map((file) => path.join(dirPath, file.name));
+         const folders = entries.filter((folder) => folder.isDirectory());
+         for (const folder of folders)
++            if (folder.name !== 'node_modules')
```
anuraghazra, line 9: whats with bundlemon?

anuraghazra (approved):
LGTM

packages/blade/patches/bundlemon+2.0.1.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/node_modules/bundlemon/lib/main/analyzer/pathUtils.js b/node_modules/bundlemon/lib/main/analyzer/pathUtils.js
```
chaitanyadeorukhkar, line 1: Raise an issue on bundlemon?

chaitanyadeorukhkar (approved):

PR #1477, opened 2023-07-27
fix: remove className from Button, svg, Link

saurabhdaware:
while using `useStyledProps`, the filtering of styled-props was happening inside of the Styled Component. But in cases where we were just doing `...styledProps` over a styled-component, props like className were ending up in DOM.

Issue: https://razorpay.slack.com/archives/G01B3LQ9H0W/p1690458021769119

Because props were getting spread, few props like accessibility props or data props were also getting added by-default for internal components which shouldn't have been the case.

So after this change, we have to explicitly define accessibility props on BaseButton and BaseLink. I have added `accessibilityProps` object which lets you pass a11y props to base components on button and link.


packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -35,6 +36,13 @@ type ButtonCommonProps = {
   isLoading?: boolean;
   accessibilityLabel?: string;
   type?: 'button' | 'reset' | 'submit';
+
+  /**
+   * It is exposed for internal usage with tooltip.
+   *
+   * @private
+   */
+  'aria-describedby'?: string;
```
chaitanyadeorukhkar, line 45: Shoudn't we make it `ariaDescribedBy`? would make more sense with rest of naming

packages/blade/src/utils/metaAttribute/__tests__/metaAttribute.native.test.tsx
```diff
@@ -1,13 +1,12 @@
...
-import { Button } from '~components/Button';
 import renderWithTheme from '~utils/testing/renderWithTheme.native';
+import BaseBox from '~components/Box/BaseBox';
 
 describe('metaAttribute', () => {
   it('add the correct meta attributes', () => {
-    const { getByRole } = renderWithTheme(
-      <Button {...metaAttribute({ name: 'Button', testID: 'button-test' })}>{'Click'}</Button>,
+    const { getByTestId } = renderWithTheme(
+      <BaseBox {...metaAttribute({ name: 'new-box', testID: 'box-test' })}>{'Click'}</BaseBox>,
```
chaitanyadeorukhkar, line 8:
```suggestion
      <BaseBox {...metaAttribute({ name: 'new-box', testID: 'box-test' })} />,
```

packages/blade/src/utils/metaAttribute/__tests__/metaAttribute.web.test.tsx
```diff
@@ -1,13 +1,13 @@
...
-import { Button } from '~components/Button';
 import renderWithTheme from '~utils/testing/renderWithTheme.web';
+import BaseBox from '~components/Box/BaseBox';
 
 describe('metaAttribute', () => {
   it('add the correct meta attributes', () => {
-    const { getByRole } = renderWithTheme(
-      <Button {...metaAttribute({ name: 'Button', testID: 'button-test' })}>{'Click'}</Button>,
+    const { getByText } = renderWithTheme(
+      <BaseBox {...metaAttribute({ name: 'new-box', testID: 'box-test' })}>{'Click'}</BaseBox>,
```
chaitanyadeorukhkar, line 8:
```suggestion
      <BaseBox {...metaAttribute({ name: 'new-box', testID: 'box-test' })} />,
```

chaitanyadeorukhkar (approved):

kamleshchandnani:
but this was already handled right? i remember raising this when we were adding styled props support?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -317,14 +319,16 @@ const _BaseButton: React.ForwardRefRenderFunction<BladeElementRef, BaseButtonPro
...
     testID,
     onFocus,
     onMouseLeave,
     onMouseMove,
     onPointerDown,
     onPointerEnter,
-    ...styledProps
+    accessibilityProps,
+    onTouchEnd,
+    onTouchStart,
```
anuraghazra, line 330: new props?

packages/blade/src/components/Button/Button/Button.tsx
anuraghazra: Can we do this: 

Remove this type-definition so users don't get the intellisence? 
In the tooltip we won't get any errors either way because we spread a `Record` type and in case of the error that will cause in this line: we can either ignore it or cast the type

```
describedBy: (rest as Record<string, string>['aria-describedby'],
```

packages/blade/src/components/Dropdown/DropdownButton.tsx
```diff
@@ -54,17 +53,16 @@ const _DropdownButton = ({
...
       testID={testID}
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       ref={triggererRef as any}
-      {...makeAccessible({
+      accessibilityProps={{
+        label: accessibilityLabel,
         hasPopup: getActionListContainerRole(hasFooterAction, 'DropdownButton'),
         expanded: isOpen,
         controls: `${dropdownBaseId}-actionlist`,
-        role: 'combobox',
```
anuraghazra, line 65: where did the role go?

anuraghazra (approved):
LGTM

packages/blade/src/components/Dropdown/DropdownButton.tsx
saurabhdaware: Removed it. It was previously getting overriden to role button and the screenreader seemed to announce it correctly with that too so kept the behaviour same. Will reevaluate what should be the role of trigger during autocomplete

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
saurabhdaware: They were part of Common events type but were not explicitly added in basebutton and baselink. They were still getting passed and we were using them so added them explicitly

packages/blade/src/components/Button/Button/Button.tsx
saurabhdaware: Can't think of any scenario where people would try to use this prop. We anyway have jsdoc with warning so should be fine. Having it in type will help us track which props are expected to be on dom.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -60,7 +62,7 @@ type BaseButtonCommonProps = {
   }>;
   type?: 'button' | 'reset' | 'submit';
   isLoading?: boolean;
-  accessibilityLabel?: string;
+  accessibilityProps?: Partial<AccessibilityProps>;
```
kamleshchandnani, line 65: didn't understand why this?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -432,13 +436,15 @@ const _BaseButton: React.ForwardRefRenderFunction<BladeElementRef, BaseButtonPro
...
       onKeyDown={onKeyDown}
+      onTouchStart={onTouchStart}
+      onTouchEnd={onTouchEnd}
       type={type}
       borderWidth={borderWidth}
       borderRadius={borderRadius}
       motionDuration={motionDuration}
       motionEasing={motionEasing}
       {...metaAttribute({ name: MetaConstants.Button, testID })}
-      {...styledProps}
```
kamleshchandnani, line 441: so we shall never use `...styledProps` now and use `...getStyledProps(rest)` always?

packages/blade/src/components/Button/Button/Button.tsx
kamleshchandnani: but if you out `ariaDescribedBy` via our makeA11yprops it will render `aria-describedby` only right?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
saurabhdaware: Earlier all HTML props were getting spread no so we could do `{...makeAccessible({ /* accessibility props */ })}` and they were getting added (we did this at multiple places)

Now that we're not spreading all props. We need an alternative way to pass accessibility props to these base components. This is that alternative way.

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
saurabhdaware: Mostly yes. If you know that props are filtered then you can still do `...styledProps` (e.g. if I am filtering it on top component somewhere and pass it to lets say `StyledButton`, then I don't have to filter them in StyledButton again.

But yes in general, it's best to do that anyway.

packages/blade/src/components/Button/Button/Button.tsx
saurabhdaware: Yup. But in tooltip, we're passing `aria-describedby` directly (we have to because otherwise we'll have to put the checks saying if component is `Button` or `Link` use describedBy else use `aria-describedby` in custom components.

kamleshchandnani (approved):

PR #1484, opened 2023-07-28
fix: chromatic optimisations

kamleshchandnani:


packages/blade/.storybook/react/preview.js
```diff
@@ -43,12 +43,50 @@ export const parameters = {
           'makeTypographySize',
           'useTheme',
           'overrideTheme',
+          'Accessibility',
         ],
         'Components',
+        [
```
kamleshchandnani, line 49: just restructured this and grouped similar components, right now it was big chaos when I'm looking for accordion and I don't even know if collapsible also exists

packages/blade/src/components/Icons/Icons.stories.tsx
```diff
@@ -114,36 +115,50 @@ Icon.args = {
...
-                textOverflow: 'ellipsis',
-                textAlign: 'center',
-              }}
-            >
-              {icon}
-            </BaseBox>
-          </BaseBox>
-        );
-      })}
+      {isChromatic() ? (
```
kamleshchandnani, line 154: rendering only one icon in chromatic

packages/blade/src/components/Icons/Icons.stories.tsx
```diff
@@ -114,36 +115,50 @@ Icon.args = {
   icon: 'CreditCardIcon',
 };
 
+const IconList = ({
+  IconComponent,
+  iconName,
+  ...args
+}: {
+  IconComponent: IconComponent;
+  iconName: keyof typeof iconMap;
```
kamleshchandnani, line 124: @anuraghazra can I get the keynames as union type here from the `iconMap` object?

anuraghazra: Yes, you need to make the iconMap to use as const. 

```
const iconMap = {
...
...
} as const;
```

I will recommend exporting a type from here also
```
type IconNames = keyof typeof iconMap;
```

> because i can see an error in BaseLink.stories where you need to now do `iconMap[(icon as unknown) as IconNames]` instead. (just check that story you'll see the error)

kamleshchandnani: did that as const but it still shows `string` instead of union of keys
Screenshot 2023-07-28 at 4 25 28 PM

packages/blade/src/components/Icons/Icons.stories.tsx
kamleshchandnani: Screenshot 2023-07-28 at 4 22 09 PM

packages/blade/src/components/Icons/Icons.stories.tsx
anuraghazra: Did you removed the `const iconMap: Record<string, IconComponent>` type? it needs to be removed so that TS doesn't widen the object type.

packages/blade/src/components/Icons/Icons.stories.tsx
anuraghazra: Should we? because for chromatic it should not matter component ke andar if we are rendering 1 or 1000 subcomponents right? the chromatic will only take 1 snapshot either way.

packages/blade/.storybook/react/preview.js
```diff
@@ -43,12 +43,50 @@ export const parameters = {
...
+          'Modal',
+          'BottomSheet',
+          'Divider',
+          'List',
+          'Accordion',
+          'Collapsible',
+          // feedback group
+          'Indicator',
+          'Badge',
+          'Counter',
```
saurabhdaware, line 74: This sorting makes sense here because it has this `// feedback group`, `// Layout and complex interactive elements` headings.

But when I see it in storybook, it feels a bit random. The consumers won't understand this grouping because they won't have the detailed context no? The alphabetical order felt a bit simpler to read 🙈 

image

saurabhdaware: Ideally there should've been option to put more subheadings in storybook 🙈

packages/blade/.storybook/react/preview.js
```diff
@@ -43,12 +43,50 @@ export const parameters = {
...
           'useTheme',
           'overrideTheme',
+          'Accessibility',
         ],
         'Components',
+        [
+          'Typography',
+          'Amount',
+          'Icons',
+          'Button',
```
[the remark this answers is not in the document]
kamleshchandnani: they automatically get added. this is just to define the order

packages/blade/.storybook/react/preview.js
```diff
@@ -43,12 +43,50 @@ export const parameters = {
...
+          'Tooltip',
+          // others
+          'BaseHeaderFooter',
+        ],
         'Recipes',
       ],
     },
   },
+  // Notifies Chromatic to pause the animations when they finish for all stories.
+  chromatic: { pauseAnimationAtEnd: true },
```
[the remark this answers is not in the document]
kamleshchandnani: the css animations, this is for motion tokens

packages/blade/.storybook/react/preview.js
kamleshchandnani: we can put more subheadings we can create groups as well. I thought to do it but then i can't find components from top level. should we do that?

packages/blade/src/components/Icons/Icons.stories.tsx
kamleshchandnani: hmm true 🤔

packages/blade/.storybook/react/preview.js
saurabhdaware: They end up like directory no? or will it look like proper heading? and will it stay expanded 🤔 ?

kamleshchandnani: If I put `/` then it goes as a subdirectory eg: `Components/Feedback` but I can do `Components-Feedback` like that

saurabhdaware: Oh that would be great. I tried putting `-` locally but didn't seem to work 🤔 might be missing somehing

kamleshchandnani: looks like this. honestly doesn't look very intuitive 🙈 
Screenshot 2023-08-03 at 1 06 57 PM

the other is this option with `/` which creates sub directories. This still looks better but then discoverability becomes a challenge.yes people should use `search` but you know what i mean here 😆 
Screenshot 2023-08-03 at 1 08 38 PM

kamleshchandnani: @saurabhdaware 👆

saurabhdaware: Should we let it be alphabetic order then? Not sure if they will be familiar with the terminology of Feedback group etc. They will see Alert on design and will come to our docs to search "Alert" itself 🤔

kamleshchandnani: yes isliye i thought to group it internally and not confuse consumers but when you start looking for components you'll find related things nearby

kamleshchandnani:
closing in favor of #1518

Issue #1436, opened 2023-07-13
Expose `isOpen` on Dropdown


saurabhdaware:
We don't allow controlled closing / opening of the Dropdown. But there are some valid requirements for that coming up. E.g. - https://github.com/razorpay/blade/issues/1334#issuecomment-1620269052

**Some challenges**
- Once we let consumers handle open / close state, we have to make sure things like keyboard events (pressing ESC should close Dropdown, pressing down arrow should open the dropdown, pressing tab should move to next item, etc) are handled by consumers
- We will have to expose `useDropdown` hook that returns `triggerProps` so that events on all that handling of events gets simpler. 

```jsx
const ControlledSelectTrigger = ({  isOpen, setIsOpen }) => {
  const { triggerProps } = useDropdown();
  return <SelectInput {...triggerProps({ isOpen, setIsOpen })} label="Controlled Select Trigger" />
}

const App = () => {
  const [isOpen, setIsOpen] = React.useState(false);
  return (
    <Dropdown isOpen={isOpen} onDismiss={() => setIsOpen(false)}>
      <ControlledSelectTrigger isOpen={isOpen} setIsOpen={setIsOpen} />
      <DropdownOverlay>
      </DropdownOverlay>
    </Dropdown>
  );
}
```

Issue #1416, opened 2023-07-11
Unable to close Multi-select dropdown with footer button


ShivamKSRazorpay:
I am unable to close the Multi-select dropdown with footer button. I have used the “hidden button focus” hack with DropdownOverlay and DropdownFooter button as shown [HERE](https://codesandbox.io/s/confident-christian-jj5dsn?file=/App.tsx). But this hack doesn’t work with Dropdown BottomSheet and BottomSheetFooter button as shown [HERE](https://codesandbox.io/s/misty-pine-vmn89g?file=/App.tsx:1447-1464).

What is an ideal way to close the DropdownOverlay or BottomSheet with a Dropdown action button click?

chaitanyadeorukhkar:
@saurabhdaware can you check this?

ShivamKSRazorpay:
@saurabhdaware  Thanks for picking it up! Any ETA on the fix? Need it for the Transactions Revamp project delivery.

ShivamKSRazorpay:
@saurabhdaware  this is not completely done, right?
I see this still needs to be done https://github.com/razorpay/blade/issues/1436

saurabhdaware:
@ShivamKSRazorpay Yup ideally we would need something like `isOpen` on Dropdown component itself to do something like this. Have created #1436 for that.

To get you unblocked, we are allowing `isOpen` to BottomSheet to take priority over Dropdown's internal state so you can use that for BottomSheet - https://codesandbox.io/s/modest-matan-4d29f7?file=/App.tsx

But yes for Dropdown on desktop, the current solution is hacky. We'll solve that with something like `isOpen`

PR #1485, opened 2023-07-28
fix: Expose isOpen and setIsOpen in Dropdown

archie252000:
Fixes: #1436

Also provides a fix for #1416 (Tested it manually)


saurabhdaware:
We're yet to decide API on this. I'll let you know once we have something. I was thinking of something like this-

```jsx
<Dropdown isOpen={} onOpenChange={} />
// or if we want to make non-breaking change then
<Dropdown isOpen={} onOpen={} onDismiss={} />
```

I will update here once we have that. We can keep it in draft till then maybe?

archie252000:
@saurabhdaware sure!

saurabhdaware:
Sorry it took time. Got into some other task. 

We decided to go with 

```jsx
<Dropdown isOpen={} onOpenChange={} />
```

and deprecate the existing `onDismiss`. I'll merge your branch into a stacked branch and make changes. Thanks for the PR! Fixes an important thing for us :D

saurabhdaware (approved):

Issue #1033, opened 2023-03-02
Web file imported in native


chaitanyadeorukhkar:
https://github.com/razorpay/blade/blob/c994b0e7e827d8734d2e1d6b62c708bb1121e22b/packages/blade/src/components/Dropdown/DropdownOverlay.native.tsx#L5

Since we're only importing a type, it's not breaking anything but we should avoid importing from `.web` files to a `.native` file

chaitanyadeorukhkar:
@saurabhdaware can you check this whenever possible?

divyanshu013:
We should add a linting rule for this

saurabhdaware:
yeah sure I'll check

PR #1495, opened 2023-08-02
chore: add new `no-cross-platform-imports` eslint rule

snitin315:
Fixes https://github.com/razorpay/blade/issues/1033

- Created a new package `slint-plugin-blade`.
- Created a custom rule `no-cross-platform-imports` to avoid importing from .web files to .native files and vice-versa.
- In the 
Screenshot 2023-08-02 at 9 19 51 AM
Screenshot 2023-08-02 at 9 24 36 AM


package.json
```diff
@@ -21,7 +22,8 @@
     "lint": "eslint --ext js,jsx,ts,tsx packages && stylelint '**/*.{js,jsx,ts,tsx}'",
     "publish-npm": "lerna run --scope @razorpay/blade publish-npm",
     "release": "lerna run --scope @razorpay/blade generate-github-npmrc && changeset publish",
-    "build": "lerna run --scope @razorpay/blade build"
+    "build": "lerna run --scope @razorpay/blade build",
+    "postinstall": "yarn workspace eslint-plugin-blade build"
```
chaitanyadeorukhkar, line 26: why can't we use lerna here? We're using lerna in above scripts, lets do the same thing everywhere?

snitin315: @chaitanyadeorukhkar We can use Lerna here, I've also updated the `tsc:blade` script to use lerna for consistency.

anuraghazra:
Noice

packages/eslint-plugin-blade/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+  "extends": "../../tsconfig.json",
+  "include": ["src"],
+  "exclude": [],
+  "compilerOptions": {
+    "baseUrl": ".",
+    "outDir": "build",
+    "module": "commonjs",
+    "noEmit": false,
+    "paths": {
+      "~src/*": ["src/*"]
```
chaitanyadeorukhkar, line 11: Why this?

.github/actions/install-dependencies/action.yml
```diff
@@ -18,8 +18,13 @@ runs:
...
 
     # cache miss - install packages with `yarn --frozen-lockfile`
-    # cache hit - download packages from the cache
+    # cache hit - download packages from the cache and explicitly run postinstall script
     - name: install packages if cache miss
       run: yarn --frozen-lockfile
       shell: bash
       if: steps.cache.outputs.cache-hit != 'true'
+
+    - name: Run postinstall script
```
chaitanyadeorukhkar, line 27: Why do we need to run this explicitly? won't yarn run it anyway? Or is it that when there is cache-hit, it will assume yarn is done and skip post install?

.eslintrc.js
```diff
@@ -129,8 +129,9 @@ module.exports = {
         'plugin:@typescript-eslint/recommended-requiring-type-checking',
         'plugin:prettier/recommended',
       ],
-      plugins: ['@typescript-eslint', 'jsx-a11y', 'no-only-tests'],
+      plugins: ['@typescript-eslint', 'jsx-a11y', 'no-only-tests', 'blade'],
```
chaitanyadeorukhkar, line 132: Should we call it `@razorpay/blade-eslint`? Just blade seems too confusing

packages/eslint-plugin-blade/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "@types/node": "18.17.1",
+    "@types/jest": "29.5.3",
+    "jest": "29.6.1",
+    "ts-jest": "29.1.1"
+  },
+  "scripts": {
+    "build": "tsc -b",
+    "lint": "eslint .",
+    "test": "jest",
+    "postinstall": "npm run build"
```
chaitanyadeorukhkar, line 19: Why npm? Use yarn here?

.github/actions/install-dependencies/action.yml
snitin315: We don't run `yarn` when it's cache-hit. Hence it's required.

packages/eslint-plugin-blade/tsconfig.json
snitin315: Removed, this is really not required.

.eslintrc.js
snitin315: `@razorpay/blade-eslint` is not possible.

Screenshot 2023-08-07 at 6 42 51 AM


As per the [naming convention](https://eslint.org/docs/latest/extend/plugins#name-a-plugin), we can have `@razorpay-blade-eslint/eslint-plugin`, `@razorpay-blade/eslint-plugin`, `eslint-plugin-razorpay-blade-eslint`, `eslint-plugin-razorpay-blade`.


I've updated it to `@razorpay-blade/eslint-plugin` in the plugin's key we define `@razorpay-blade`.

snitin315: I tried with `@razorpay-blade-eslint/eslint-plugin` and `eslint-plugin-razorpay-blade` names but they doesn't seem to work on CI for some unknown reason.

Screenshot 2023-08-07 at 7 06 39 AM
Screenshot 2023-08-07 at 7 19 38 AM

Let's keep `eslint-plugin-blade` only for now. We don't publish this package anywhere. WDYT?

packages/eslint-plugin-blade/package.json
snitin315: it should be yarn, my bad.

.eslintrc.js
chaitanyadeorukhkar: Okay. Lets keep it eslint-plugin-blade

chaitanyadeorukhkar, line 132: This will change too right?
```suggestion
      plugins: ['@typescript-eslint', 'jsx-a11y', 'no-only-tests', 'eslint-plugin-blade'],
```

chaitanyadeorukhkar (dismissed):
1 comment if we need to change plugin array in our eslintrc. Rest LGTM

.eslintrc.js
snitin315: @chaitanyadeorukhkar both works, ESLint will automatically add the `eslint-plugin` prefix.

chaitanyadeorukhkar: Okay

packages/eslint-plugin-blade/src/index.ts
```diff
@@ -0,0 +1,5 @@
+import noCrossPlatformImportsRule from './rules/no-cross-platform-imports';
```
saurabhdaware, line 1: eslint-plugin-blade sounds like a eslint plugin that our consumers should use to lint blade errors. Can we add this plugin code without creating a separate package 🤔 ? Since its specific to our repo? or maybe a new open-source plugin `eslint-plugin-no-cross-platform-imports`?

snitin315: @saurabhdaware We are not publishing `eslint-plugin-blade` anywhere, this is for internal use only and not exposed to consumers.

saurabhdaware: oh cool then

anuraghazra: Similar thoughts, what happens if later down the line we want to expose a consumer end eslint plugin for blade? 

Should we call our internal plugin `esling-plugin-blade-internal`?

package.json
```diff
@@ -10,18 +10,20 @@
...
+      "packages/razorsharp",
+      "packages/eslint-plugin-blade"
     ],
     "nohoist": [
       "**"
     ]
   },
   "scripts": {
-    "tsc:blade": "yarn workspace @razorpay/blade typecheck",
+    "tsc:blade": "lerna run --scope @razorpay/blade typecheck",
```
anuraghazra, line 21: why are we changing this to use lerna? 
Ideally we should stick to workspaces only, maybe later down the line we can remove lerna since lerna has a new successor (nx) now and isn't even being used very much in blade.

packages/eslint-plugin-blade/src/rules/no-cross-platform-imports.ts
```diff
@@ -0,0 +1,43 @@
...
+    schema: [], // no options
+  },
+  create: (context) => ({
+    ImportDeclaration(node) {
+      const sourceValue = node.source.value;
+      const isWebImport = sourceValue.endsWith('.web');
+      const isNativeImport = sourceValue.endsWith('.native');
+
+      if (
+        isWebImport &&
```
anuraghazra, line 22: I wonder if we have another check for: 
Ideally files which runs on both platform should not import platform specific files

- .native or .web file being imported in file with no extension (files which run on both platforms)

anuraghazra (dismissed):
LGTM, minor comment.

package.json
snitin315: For consistency https://github.com/razorpay/blade/pull/1495#discussion_r1281821818

packages/eslint-plugin-blade/src/index.ts
snitin315: I understand the concern, but the likelihood of creating a consumer-facing eslint plugin is very low. We have proper types and TS checks in place for blade. We can rename the plugin later if required, also we can add as many rules as we want so maybe be can use the same plugin in the future.

packages/eslint-plugin-blade/src/rules/no-cross-platform-imports.ts
snitin315: >.native or .web file being imported in file with no extension (files which run on both platforms)

What's the issue in importing if the file can run cross-platform?

snitin315: We don't at the moment. Should I add it?

snitin315: Added in https://github.com/razorpay/blade/pull/1495/commits/f24a76c2857e3bee6fab57075d20074316c36667

anuraghazra (approved):

anuraghazra (approved):

Issue #1447, opened 2023-07-18
Make Badge, Counter, Indicator as display inline-flex / inline-block


saurabhdaware:
Some elements like Badge, Counter, Tag, Indicator are expected to be inlined elements that can be used next to text or other components. They also don't take up full space so should not be block.

We can change these component (and also identify other components where this might be required)

              Oh actually we did try to make Badge inline-flex but I just noticed that we didn't change display property of the wrapper container https://github.com/razorpay/blade/pull/1254

We can do this separately then but we should make it inline-flex / inline-block as it was raised by our consumers as well https://github.com/razorpay/blade/issues/1241, https://razorpay.slack.com/archives/CMQ3RBHEU/p1685526734598399?thread_ts=1685526177.654729&cid=CMQ3RBHEU.

Lets take it up separately since this change is big in itself already.

_Originally posted by @saurabhdaware in https://github.com/razorpay/blade/pull/1431#discussion_r1266651693_

saurabhdaware:
On Badge, it was added on inner div and not outer div. So outer div is still display: flex.

For other components like Counter, Indicator, we did not add.

chaitanyadeorukhkar:
We've not had any such feedback from consumers. Now that we allow them to pass a custom display prop, they should be able to make it inline anyway. We should do this, I'll assign it to @snitin315 since he's working on issues but lets make it a P1?

PR #1499, opened 2023-08-03
feat: implement `inline-flex` display for `Badge`,`Counter`, & `Indicator`

snitin315:
Resolve https://github.com/razorpay/blade/issues/1447


packages/blade/src/components/Indicator/Indicator.tsx
```diff
@@ -88,21 +88,24 @@ const Indicator = ({
...
     <BaseBox
-      display="flex"
+      display={(isWeb ? 'inline-flex' : 'flex') as never}
       {...a11yProps}
       {...metaAttribute({ name: MetaConstants.Indicator, testID })}
       {...getStyledProps(styledProps)}
     >
-      <BaseBox display="flex" flexDirection="row" alignItems="center">
+      <BaseBox
+        display={(isWeb ? 'inline-flex' : 'flex') as never}
```
saurabhdaware, line 105: we don't have to make inner item `inline-` right? if outer item is inner-flex the inner item will anyway not extend the width 🤔

.changeset/late-boats-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+feat: implement `inline-flex` display for `Badge`, `Counter`, & `Indicator` components
```
saurabhdaware, line 5:
```suggestion
feat: make `Badge`, `Counter`, & `Indicator` components `inline-flex` by default
```

packages/blade/src/components/Indicator/Indicator.tsx
snitin315: @saurabhdaware I added it since `Badge` and `Counter` have it as well. I feel it should not cause any issue

saurabhdaware: I think we should remove it from Badge and Counter instead. It was incorrectly fixed earlier. 

Won't cause an issue but we have to add that extra check everywhere so better to remove it if its not needed

snitin315: I agree, will remove it.

saurabhdaware:
By removing the check I meant keeping it static `display="flex"`
```diff
- display={isReactNative() ? 'flex' : 'inline-flex'}
+ display="flex"
```

When we remove it, it becomes display block by default on web which breaks the UI of element
image

kamleshchandnani:
@snitin315 resolve conflicts?

saurabhdaware (dismissed):

packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -131,7 +131,7 @@ const _Badge = ({
 
   return (
     <BaseBox
-      display="flex"
+      display={(isReactNative() ? 'flex' : 'inline-flex') as never}
```
chaitanyadeorukhkar, line 134: as `never` 😢 

@anuraghazra iska kuch karna padega long term ke liye. Not a PR blocker but we need to figure out something soon. Upgrade RN me bhi similar issues the

.changeset/late-boats-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
```
chaitanyadeorukhkar, line 2: Lets make it minor? It might break snapshots for some folks

chaitanyadeorukhkar (dismissed):

.changeset/late-boats-report.md
snitin315:
```suggestion
"@razorpay/blade": minor
```

chaitanyadeorukhkar: You are so lazy man 😆

chaitanyadeorukhkar (approved):

saurabhdaware (approved):

Issue #1449, opened 2023-07-19
Add `onDrag*`, `onDrop` handlers, and pointerEvents on Box


inaseem:
Please add the following props to blade components. These props should be supported(atleast on Box) but aren't.

`onDragEnter`
`onDragOver`
`onDragLeave`
`onDrop`
`pointerEvents`
`placeItems` https://github.com/razorpay/blade/issues/1381

saurabhdaware:
Can you also add where you're using onDrag and onDrop handlers?

inaseem:
In XPayroll, we are creating a FileUpload component which should support files drag and drop on the component. I was using Box to wrap the component, but since none of the drag and drop events were available, I had to resort to using a styled.div. 

<img width="380" alt="image" src="https://github.com/razorpay/blade/assets/6770106/5d3f87e8-300b-4835-b309-1417f8a7a88d">

Issue #1381, opened 2023-06-30
Box: placeItems


cseas:
The `place-items` grid property is not supported by Box component.

```tsx
      <Box display="grid" placeItems="center">
        {children}
      </Box>
```

## Workaround

We can specify both alignItems and justifyItems separately to have the same effect.

```tsx
      <Box display="grid" alignItems="center" justifyItems="center">
        {children}
      </Box>
```

chaitanyadeorukhkar:
This shorthand support does not exist on react-native and we want to maintain API consistency across platforms. You can continue using `justifyContent` and `alignItems` to achieve this. Closing this since we won't be adding this and there is a proper solution already.

cseas:
> You can continue using justifyContent and alignItems

justifyContent is different, we need to use `justifyItems`. Even I had to look it up. 🙈 

Do you think there could be a way to add this warning/workaround to show up in the IDE itself? 🤔 
How do other similar libs handle this?

saurabhdaware:
Discussed it internally. Can look into adding this since we have `placeSelf` on styled-props already.

Issue #1394, opened 2023-07-04
☂️ Polygon Support


saurabhdaware:
There were multiple places where this is coming back as individual prop request so creating this umbrella issue to track everything regarding polygons on landing pages

Previous Context: 
- https://razorpay.slack.com/archives/C01H13RTF8V/p1685421988808309

**Problem Statement:** How to build these polygons on frontend-website?

<img width="500" alt="image" src="https://github.com/razorpay/blade/assets/30949385/642b48fe-1cdf-48d9-8c83-b8cc083266b3"> <img width="200" alt="image" src="https://github.com/razorpay/blade/assets/30949385/5616b0b9-62ab-4681-9487-8d9c46a41746">

**Existing Solution:** Build these as custom styled-components in frontend-website repo since it's limited to razorpay.com

**Consumer Concerns:** [Building these as custom components is an overhead](https://docs.google.com/spreadsheets/d/1RnjvDwNOAMnrLN2-2KIyEBwwX0cAmKK2JTtI3gNK2rA/edit#gid=0&range=12:12). developer experience will be better if 
- Box supports all props that are required for building polygons
- We standardise polygons on blade and maybe create a Polygon component

**Props we will have to support on Box for Polygons:**

- role
- backgroundImage
- transform
- transformOrigin
- boxShadow
- clipPath

saurabhdaware:
## Opinions

### Regarding supporting all polygon props on Box

Props we will have to support for Polygons

- role
- backgroundImage
- transform
- transformOrigin
- boxShadow
- clipPath

I don't see much harm in just supporting all props that polygon requires. My concerns are more on the overall call side-
- Should we start treating Box as primitive component for building other custom components? Because right now the usecase at hand is creating polygons. In future we'll have lots of other custom components with props specific to the project.

### Regarding creating custom Polygon component

I am not much in favour of creating Polygon component from blade because we haven't seen these polygons much outside of razorpay.com. Will also prefer to wait for design revamp and see what type of designs we eventually build to avoid duplicate efforts.

cseas:
> For frontend-website teams: Create a reusable Polygon styled-component in frontend-website common components to get unblocked

This is done (partially). For Payment Pages, we had to take up adhoc of coming up with a custom primitive to implement this in a clean way since Blade's Box had too many restrictions for this use-case. Code for reference available [here][1].

To elaborate on the full requirement, this wasn't a one-off implementation even for a single page. We needed a reusable component that:
- can accept different background colors/gradients
[...]

PR #1500, opened 2023-08-03
feat(Box): add `draggable`,`onDrag*`,`onDrop`,`pointerEvents`,`placeItems`

snitin315:
Resolve https://github.com/razorpay/blade/issues/1449
Resolve https://github.com/razorpay/blade/issues/1381

We're excited to unveil the latest update for `Box`! Enjoy draggable elements, onDrag* & onDrop events, pointerEvents, and placeItems props. Elevate your UI with unparalleled control and flexibility. 

https://github.com/razorpay/blade/assets/46647141/e6ab16bf-c94f-4fb3-8ce1-9e1a921105a2

### DIFF
--- a/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
+++ b/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
@@ -135,6 +135,8 @@ const getAllProps = (
     justifyItems: getResponsiveValue(props.justifyItems, breakpoint),
     justifyContent: getResponsiveValue(props.justifyContent, breakpoint),
     justifySelf: getResponsiveValue(props.justifySelf, breakpoint),
+    placeSelf: getResponsiveValue(props.placeSelf, breakpoint),
+    placeItems: getResponsiveValue(props.placeItems),
     order: getResponsiveValue(props.order, breakpoint),
     position: getResponsiveValue(props.position, breakpoint),
     zIndex: getResponsiveValue(props.zIndex, breakpoint),
@@ -223,6 +225,9 @@ const getAllProps = (
     ...(!isReactNative() && {
       boxShadow: getElevationValue(props.elevation, props.theme, breakpoint),
     }),
+    transform: props.transform,
+    transformOrigin: props.transformOrigin,
+    clipPath: props.clipPath,
   };
 };
 
--- a/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
+++ b/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
@@ -225,9 +225,6 @@ const getAllProps = (
     ...(!isReactNative() && {
       boxShadow: getElevationValue(props.elevation, props.theme, breakpoint),
     }),
-    transform: props.transform,
-    transformOrigin: props.transformOrigin,
-    clipPath: props.clipPath,
   };
 };
 
--- a/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
+++ b/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
@@ -136,7 +136,7 @@ const getAllProps = (
     justifyContent: getResponsiveValue(props.justifyContent, breakpoint),
     justifySelf: getResponsiveValue(props.justifySelf, breakpoint),
     placeSelf: getResponsiveValue(props.placeSelf, breakpoint),
-    placeItems: getResponsiveValue(props.placeItems),
+    placeItems: getResponsiveValue(props.placeItems, breakpoint),
     order: getResponsiveValue(props.order, breakpoint),
     position: getResponsiveValue(props.position, breakpoint),
     zIndex: getResponsiveValue(props.zIndex, breakpoint),
--- a/packages/blade/src/components/Box/Box.tsx
+++ b/packages/blade/src/components/Box/Box.tsx
@@ -129,9 +129,6 @@ const makeBoxProps = (props: BoxProps): KeysRequired<Omit<BoxProps, 'testID' | '
     backgroundOrigin: props.backgroundOrigin,
     backgroundRepeat: props.backgroundRepeat,
     elevation: props.elevation,
-    transform: props.transform,
-    transformOrigin: props.transformOrigin,
-    clipPath: props.clipPath,
 
     // Border
     borderWidth: props.borderWidth,
--- a/packages/blade/src/components/Box/Box.tsx
+++ b/packages/blade/src/components/Box/Box.tsx
@@ -74,6 +74,7 @@ const makeBoxProps = (props: BoxProps): KeysRequired<Omit<BoxProps, 'testID' | '
     justifyContent: props.justifyContent,
     justifySelf: props.justifySelf,
     placeSelf: props.placeSelf,
+    placeItems: props.placeItems,
     order: props.order,
 
     // Grid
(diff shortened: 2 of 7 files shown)


packages/blade/src/components/Box/styledProps/getStyledProps.ts
```diff
@@ -54,7 +54,9 @@ const makeStyledProps = (props: StyledPropsInputType): KeysRequired<StyledPropsB
     display: props.display,
     justifySelf: props.justifySelf,
     placeSelf: props.placeSelf,
+    placeItems: props.placeItems,
```
saurabhdaware, line 57: We can skip these props from styled-props. They are only relevant for Box no

snitin315: Since we already allow `placeSelf`, makes sense to add `placeItems` as well. Can skip `pointerEvents` though. Not sure.

saurabhdaware: No for example `placeItems` is for the scenario where we place the items inside of the container in center. We don't want people to move the items inside of the button.

Which is why we support `alignSelf` and `justifySelf` on styled props but we don't support `alignItems` and justifyItems` on it

saurabhdaware:
Can you add these props as well https://github.com/razorpay/blade/issues/1394#issuecomment-1637783662? 2 issues will be resolved then.

snitin315:
@saurabhdaware Added.

packages/blade/src/components/Box/styledProps/getStyledProps.ts
```diff
@@ -48,7 +48,9 @@ const removeUndefinedStyledProps = (obj: StyledPropsInputType): StyledPropsInput
  *
  * **Use `getStyledProps` instead if you're using this for adding styledProps!**
  */
-const makeStyledProps = (props: StyledPropsInputType): KeysRequired<StyledPropsBlade> => {
+const makeStyledProps = (
+  props: StyledPropsInputType,
+): KeysRequired<Omit<StyledPropsBlade, 'pointerEvents' | 'placeItems'>> => {
```
saurabhdaware, line 53:
```suggestion
): KeysRequired<StyledPropsBlade> => {
```

This is expected to be required. You can add `pointerEvents: props.pointerEvents` and `placeItems: props.placeItems` in the code.

This function basically filters the props so it makes sure that props are being passed to the BaseBox correctly. The placeItems won't work as expected without passing it down from here

packages/blade/src/components/Box/Box.tsx
```diff
@@ -123,6 +124,9 @@ const makeBoxProps = (props: BoxProps): KeysRequired<Omit<BoxProps, 'testID' | '
     backgroundOrigin: props.backgroundOrigin,
     backgroundRepeat: props.backgroundRepeat,
     elevation: props.elevation,
+    transform: props.transform,
+    transformOrigin: props.transformOrigin,
+    clipPath: props.clipPath,
```
saurabhdaware, line 129: So all props that we add here need to be handled in BaseBox.

The way it works is, BaseBox is superset of Box and styled-props. Everything we add here needs to be handled in BaseBox.

You can checkout how we handle it here - 

https://github.com/razorpay/blade/blob/d84a656f178bf25cf44bd84038cbc3481e58126e/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts#L207-L211

This makes the props work. It handles there types and also makes the responsiveness work with props. E.g. - `pointerEvents={{ base: 'none', m: 'all' }}`

packages/blade/src/components/Box/Box.tsx
```diff
@@ -147,6 +151,16 @@ const makeBoxProps = (props: BoxProps): KeysRequired<Omit<BoxProps, 'testID' | '
...
     onScroll: props.onScroll,
 
+    // Drag and Drop
+    draggable: props.draggable,
+    onDragStart: props.onDragStart,
+    onDragEnd: props.onDragEnd,
+    onDragEnter: props.onDragEnter,
+    onDragLeave: props.onDragLeave,
+    onDragOver: props.onDragOver,
+    onDrop: props.onDrop,
```
saurabhdaware, line 161: I think the event handlers and `draggable` will get handled by themselves (since they are HTML attributes and not CSS properties) but you might have to add pointerEvents, placeItems, and other props to baseBoxStyles

packages/blade/src/components/Box/Box.tsx
saurabhdaware: Offtopic: Looks like we also missed adding `placeItems` here. Can you add it while adding other props?

packages/blade/src/components/Box/styledProps/getStyledProps.ts
snitin315: I didn't add this as per https://github.com/razorpay/blade/pull/1500#discussion_r1282655823

snitin315:
TS is not happy, will check it.

packages/blade/src/components/Box/styledProps/getStyledProps.ts
saurabhdaware: Right in that case that `placeItems` value shouldn't exist on StyledPropsBlade type.

StyledPropsBlade is the type of styled-props that we support and `placeItems` and `pointerEvents` are not the styled-props.

snitin315: I didn't get this, so should we use `KeysRequired<StyledPropsBlade>` or `Omit<StyledPropsBlade, 'pointerEvents' | 'placeItems'>>?`

snitin315:
> Can you add these props as well [#1394 (comment)](https://github.com/razorpay/blade/issues/1394#issuecomment-1637783662)? 2 issues will be resolved then.

@saurabhdaware Do we want these props to accept responsive values?

packages/blade/src/components/Box/styledProps/getStyledProps.ts
saurabhdaware: We should use `KeysRequired<StyledPropsBlade>` and remove the `placeItems` and `pointerEvents` from actual `StyledPropsBlade` type in basebox types file

saurabhdaware: Basically you're having to omit it because that prop exists in that type right? but that prop shouldn't exist in that type.

saurabhdaware:
> Do we want these props to accept responsive values?

Yup all CSS properties on Box and Styled Props should accept responsive values for consistency

snitin315:
> Can you add these props as well [#1394 (comment)](https://github.com/razorpay/blade/issues/1394#issuecomment-1637783662)? 2 issues will be resolved then.

@saurabhdaware I reverted these prop changes from this PR, it seems they need additional handling. For example, `transform` expects an object instead of `string`. I will create a separate PR for them.

packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
```diff
@@ -135,6 +135,8 @@ const getAllProps = (
     justifyItems: getResponsiveValue(props.justifyItems, breakpoint),
     justifyContent: getResponsiveValue(props.justifyContent, breakpoint),
     justifySelf: getResponsiveValue(props.justifySelf, breakpoint),
+    placeSelf: getResponsiveValue(props.placeSelf, breakpoint),
+    placeItems: getResponsiveValue(props.placeItems),
```
saurabhdaware, line 139:
```suggestion
    placeItems: getResponsiveValue(props.placeItems, breakpoint),
```

saurabhdaware (approved):

kamleshchandnani:
@snitin315 we also need to set aria roles when draggable is set? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-grabbed

PR #1503, opened 2023-08-04
feat(Amount): add support for more currencies

snitin315:


packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -43,53 +43,275 @@ const amountLineHeights: Record<
   'title-medium': 600,
 } as const;
 
-type CurrencyPrefixMapping = {
-  [key: string]: {
-    'currency-symbol': string;
-    'currency-code': string;
-  };
-};
```
snitin315, line 51: Removed this so that I can do `type Currency = keyof typeof currencyPrefixMapping` to get all valid currencies.

packages/blade/src/components/Amount/__tests__/Amount.web.test.tsx
```diff
@@ -80,10 +82,14 @@ describe('<Amount />', () => {
     expect(container).toMatchSnapshot();
   });
 
-  it('should render MYR currency Amount ', () => {
-    const { container } = renderWithTheme(<Amount currency="MYR" value={1000} />);
-    expect(container).toMatchSnapshot();
-  });
+  for (const currency of Object.keys(currencyPrefixMapping)) {
```
chaitanyadeorukhkar, line 85: That's a lot of snapshots! Since currency is just string, lets just check for 1 or 2 currencies?

packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -43,53 +43,275 @@ const amountLineHeights: Record<
...
-const currencyPrefixMapping: CurrencyPrefixMapping = {
-  INR: {
-    'currency-symbol': '₹',
-    'currency-code': 'INR',
-  },
-  MYR: {
-    'currency-symbol': 'RM',
-    'currency-code': 'MYR',
-  },
+// All the supported currency codes are taken from https://github.com/razorpay/dashboard/blob/1e64ca67130aa0e9452e5c35111df261706f25b6/web/js/merchant/constants/currency.js#L3
```
chaitanyadeorukhkar, line 46: Lets avoid putting such links. Lets just say we took this from Razorpay's Merchant Dashboard codebase

packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -98,5 +320,6 @@ export {
   affixFontSizes,
   currencyPrefixMapping,
   currencyLocaleMapping,
-  currencyAbbreviationsMapping,
+  getCurrencyAbbreviations,
+  Currency,
 };
```
chaitanyadeorukhkar, line 325: Export type independently. We've had issues in the past in certain cases when types weren't exported independently
```suggestion
};

export type { Currency };
```

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -1,8 +1,9 @@
 import type { ReactElement } from 'react';
```
chaitanyadeorukhkar, line 1: Now that we're supporting so many currencies it could lead to some confusion on conversion. Lets add a note in Amount's JSDOC as well as story clarifying that the Amount component will render the value as is without performing any currency conversions.

packages/blade/src/components/Amount/amountTokens.ts
snitin315: We should add a lint rule for this.

packages/blade/src/components/Amount/__tests__/Amount.web.test.tsx
snitin315: Makes sense, updated.

chaitanyadeorukhkar (dismissed):
LGTM. Lets get a design sign off before we merge this @snitin315.

kamleshchandnani (commented):
overall LGTM

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -18,7 +19,7 @@ const Page = (): React.ReactElement => {
           'https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?node-id=28012%3A580578&t=WfWp7qiwZ3lvvbdw-0',
       }}
       componentName="Amount"
-      componentDescription="Amounts are used to show small amount of color coded metadata, which are ideal for getting user attention."
+      componentDescription="Amounts are used to show small amount of color coded metadata, which are ideal for getting user attention. Note that this component only displays the provided value in the specified currency, it does not perform any currency conversion."
```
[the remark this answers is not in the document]
snitin315: Done. I didn't know about the `note` prop.

Screenshot 2023-08-09 at 3 17 46 PM

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -168,10 +169,10 @@ HumanizeSuffix.args = {
 HumanizeSuffix.storyName = 'Humanize Suffix';
 
 const AmountCurrencyTemplate: ComponentStory<typeof AmountComponent> = (args) => {
-  const values = ['INR', 'MYR'] as const;
+  const values = Object.keys(currencyPrefixMapping);
```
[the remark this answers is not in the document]
snitin315: TS is not happy with `as const` anymore :(

Screenshot 2023-08-09 at 3 17 29 PM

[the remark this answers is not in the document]
anuraghazra: Unfortunately, Object.keys will return `string[]` irrespective of if we did `as const` or not.

kamleshchandnani (approved):
lets goooo 🚀

PR #1512, opened 2023-08-07
fix(SelectInput): truncate text in select input

### LINKED ISSUE #1510: Dropdown breaks when selecting option with large length
When selecting an option that has big length, dropdown text overflow. It should get truncated ideally.
https://github.com/razorpay/blade/assets/39096639/f6d690a5-6c4c-4be8-b6b1-fd08503e0cd8

image

saurabhdaware:
Resolves #1510 

## Before and After

image

image


snitin315 (dismissed):
LGTM. Some snaps need to be updated.

anuraghazra (approved):

PR #1516, opened 2023-08-08
feat(Box): add support for `transparent` background color

snitin315:
https://github.com/razorpay/blade/assets/46647141/51595886-6c29-46f5-8d6e-38e215468ff3

### DIFF
--- a/packages/blade/src/components/Box/Box.tsx
+++ b/packages/blade/src/components/Box/Box.tsx
@@ -12,7 +12,8 @@ const validateBackgroundString = (stringBackgroundColorValue: string): void => {
   if (__DEV__) {
     if (
       !stringBackgroundColorValue.startsWith('surface.background') &&
-      !stringBackgroundColorValue.startsWith('brand.')
+      !stringBackgroundColorValue.startsWith('brand.') &&
+      stringBackgroundColorValue !== 'transparent'
     ) {
       throwBladeError({
         message: `Oops! Currently you can only use \`surface.background.*\` and \`brand.*\` tokens with backgroundColor property but we received \`${stringBackgroundColorValue}\` instead.\n\n Do you have a usecase of using other values? Create an issue on https://github.com/razorpay/blade repo to let us know and we can discuss ✨`,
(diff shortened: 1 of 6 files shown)


packages/blade/src/components/Box/Box.tsx
```diff
@@ -12,7 +12,8 @@ const validateBackgroundString = (stringBackgroundColorValue: string): void => {
   if (__DEV__) {
     if (
       !stringBackgroundColorValue.startsWith('surface.background') &&
-      !stringBackgroundColorValue.startsWith('brand.')
+      !stringBackgroundColorValue.startsWith('brand.') &&
+      stringBackgroundColorValue !== 'transparent'
     ) {
       throwBladeError({
```
kamleshchandnani, line 18: add transparent here as well in the message??

packages/blade/src/components/Box/__tests__/Box.web.test.tsx
```diff
@@ -119,6 +119,32 @@ describe('<Box />', () => {
     `);
   });
 
+  it('should accept "transparent" value for backgroundColor', () => {
```
kamleshchandnani, line 122: I'm not sure if this test gives us something meaningful, ideally how do I check that transparent shall remove background and undefined shall apply the base?

snitin315: This just verifies `background-color: 'transparent'` was applied, and can be removed.

snitin315: >I check that transparent shall remove background and undefined shall apply the base?

We can see it from styles in the snap:

1. For `backgroundColor={{ base: 'brand.primary.300', l: undefined }}`:

```
.c0 {
        background-color: hsla(218,89%,51%,0.09);
      }
```

2. For `backgroundColor={{ base: 'brand.primary.300', l: 'transparent' }}`:

```
.c0 {
        background-color: hsla(218,89%,51%,0.09);
      }
      @media screen and (min-width:1024px) {
        .c0 {
          background-color: transparent;
        }
      }
```

packages/blade/src/components/Box/Box.tsx
snitin315: Done.

saurabhdaware (approved):

anuraghazra (approved):

PR #1517, opened 2023-08-08
build: upgrade storybook to v7

chaitanyadeorukhkar:
Upgrades to v7 for storybook react & react native


chaitanyadeorukhkar:
There is a bug on storybook/react-native v7-alpha3 due to which even if the story renders, we cannot interact with it. Hopefully, this should be resolved with future versions of `storybook/react-native`. We tried debugging this with storybook's maintainer Danny, but couldn't come to a final solution. He will keep us in loop for any future updates or early alpha versions we can try out. Ref: [Discord discussion](https://discord.com/channels/486522875931656193/1136972171135176785). 

Closing this for now with much sadness 😞  (Please don't delete the branch)

**Stories render but cannot navigate:**
image

PR #1518, opened 2023-08-08
add kitchensink for snapshot

kamleshchandnani:


packages/blade/.storybook/react/preview.tsx
```diff
@@ -46,6 +48,42 @@ export const parameters = {
           'overrideTheme',
         ],
         'Components',
+        [
+          'Button',
```
kamleshchandnani, line 52: @saurabhdaware lets resume [this](https://github.com/razorpay/blade/pull/1484#discussion_r1280246321) discussion here

PR #1521, opened 2023-08-09
docs(AutoComplete): API Decisions

saurabhdaware:
### Changelog Since API PR Creation

- `rows: '1' | '3' | 'expandable'` --> `maxRows: 'single' | 'multiple' | 'expandable'`
- Label inside the input will be added with `labelPosition: 'inside-input'` prop instead of `accessibilityLabel` (Checkout [With Different Label Positions](https://github.com/razorpay/blade/blob/feat/autocomplete/api-decision/packages/blade/src/components/AutoComplete/_decisions/decisions.md#with-different-label-positions) section in doc)
```diff
<AutoComplete
-  showTags={false}
-  accessibilityLabel="Status"
+  label="Status"
+  labelPosition="inside-input"
/>
```
- `showTags` is removed. We do always show tags now unless labelPosition is inside-input, then we show label inside the input instead of all tags. 
- `selected` and `onSelectChange` --> `value` and `onChange`
- Earlier `value` and `onChange` --> `inputValue` and `onInputValueChange`
- #### Changes in Controlled API
  Earlier Controlled API for AutoComplete was to directly filter over ActionListItem component with `.map`. Although during initial implementation we figured out that it is impossible to implement this API without explicitly taking array of all items. ([Which is how filtering works in AriaKit as well](https://ariakit.org/examples/combobox-filtering)). 

  Because we're giving the rendering control to consumers, here blade doesn't have any way to know what is the 'total list of items'
  ```js
  <Dropdown>
    <AutoComplete />
    <DropdownOverlay>
      <ActionList>
        {filteredItems.map((item) => <ActionListItem title={item.title} value={item.value} /> )}
      </ActionList>
    </DropdownOverlay>
  </Dropdown>
  ```

  ##### New API

  To tackle the implementation limitation, we have changed the Controlled API to use new `filteredValues` prop which can be controlled by state on consumer end to tell blade which items to show in filtering. E.g. in below example, 'bangalore' will be filtered out and only mumbai and pune will be shown
  ```js
  <Dropdown>
    <AutoComplete filteredValues={['mumbai', 'pune']} />
    <DropdownOverlay>
      <ActionList>
        <ActionListItem title="Mumbai" value="mumbai" />
        <ActionListItem title="Pune" value="pune" />
      </ActionList>
    </DropdownOverlay>
  </Dropdown>
  ```

  Checkout [Controlled AutoComplete Full Example](https://github.com/razorpay/blade/blob/feat/autocomplete/api-decision/packages/blade/src/components/AutoComplete/_decisions/decisions.md#controlled-autocomplete) for closer to real-world example


packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+| showTags           | Boolean to show or hide tags inside AutoComplete ([Checkout "With Tags Outside" Usage](#with-tags-outside))          | boolean                                                         | true              |
+| value              | Controlled state of the value inside input                                                                           | string                                                          |                   |
+| onChange           | Callback when input value changes                                                                                    | (inputValue: string) => void                                    |                   |
+| selected           | Controlled state of the selected items ([Checkout "Controlled AutoComplete" Usage](#controlled-autocomplete))        | string[]                                                        |                   |
+| onSelectChange     | Callback when selected items change                                                                                  | ({ name, values }: { name: string; values: string[] } ) => void |                   |
```
anuraghazra, line 26: nit: maybe onSelectionChange.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+
+</details>
+
+### In BottomSheet
+
+> **Warning**
+>
+> We will have to turn BottomSheetHeader (and maybe BaseHeader) to a slot to allow adding Input element there.
+
+**Open Question:** Should we also remove `leading` `trailing` etc props from Header? since same can be implemented inside the slot. This is what we follow for Footer.
```
anuraghazra, line 143: Probably shouldn't do this, Header is already standardised across Modal & BottomSheet and any deviations will also needed to be carried to the BaseHeader. 

And anyways if people want to opt out of Header altogether they can just put things in the Body.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+If not tags, what to show inside AutoComplete :thinking:? Checkout this table to understand -
+
+| Code                                                            | Output                                                                                                                     |
+| --------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------- |
+| `<AutoComplete label="Status" showTags={false} />`              | <img width="220" alt="image" src="https://github.com/razorpay/blade/assets/30949385/c8a30845-812c-493f-a51f-97c233880960"> |
+| `<AutoComplete accessibilityLabel="Status" showTags={false} />` | <img width="200" alt="image" src="https://github.com/razorpay/blade/assets/30949385/83cf63d5-ccc4-4bb9-bf0d-ce91dc5c1b09"> |
+
+So when input has label, we show number of selected items.
+
+When input doesn't have a label, we show accessibilityLabel inside the input.
```
anuraghazra, line 182: this seems a bit unexpected, accessibilityLabel i think of as a invisible prop which enables screen readers to get more information about the current component. 

Now we are saying that if accessibilityLabel is provided we will render that text. 

Instead it makes sense that: 
- accessibilityLabel -> show "2 more"
- label -> show "Status"

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+- [Ariakit - Combobox Filtering](https://ariakit.org/examples/combobox-filtering)
+  - Used for Controlled Filtering Inspiration
+- [React Select](https://react-select.com/home)
+  - Used for Accessibility and Keyboard Navigations Inspiration
+
+## Open Questions
+
+- **Q1:** Is everyone sold on `selected` and `onSelectChange` API? It is inconsistent with SelectInput. Might have to change it in SelectInput as well.
+- **Q2:** What to do with BottomSheetHeader (and BaseHeader)? Should we make it slot like Footer? Should we remove title, subtitle, etc?
+- **Q3:** `rows: '1' | '3' | 'expandable'` vs `height: 'single-line' | 'multi-line' | 'expandable'`
```
anuraghazra, line 240: 1, 3, expandable is more flexible. in future we can add 2, 5 or more.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+      <AutoComplete
+        label="Select City"
+        // Controlled input value
+        // Sets the input value of input element in AutoComplete
+        value={currentInputValue}
+        // Triggers when input value changes
+        onChange={(inputValue) => {
+          if (inputValue) {
+            // filtering logic
+            setCities(allCities.filter((city) => city.includes(inputValue)));
```
anuraghazra, line 108: We can probably have an example with derived state, instead of manually setting filtered items. 

Similar to how ariakit is doing: https://ariakit.org/examples/combobox-filtering#filtering-items

> Yes just same thing but two ways of doing but our examples are probably people will copy paste.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: You mean having a `showLabel` prop?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: So what will be the way to put input in header in this case? opt out of Header? or you mean we can go with API I proposed above where we keep leading trailing but also add children?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: Can you explain in more React-noob terms? I didn't understand that ariakit example 🙈

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: > 1, 3, expandable is more flexible. in future we can add 2, 5 or more.

I am wondering if such usecase comes up in future, should we add more values here or should we internally define the max-height.

Because if we decide to change max height in future, the `rows:` API will also change. Whereas with `height:` API, we can define what we mean by multiline

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+There are 3 controlled states that are relevant to AutoComplete
+
+1. Controlled state of filtering items
+2. Controlled state of the Input where user types
+3. Controlled state of selected items
+
+```jsx
+<AutoComplete
+  // For controlling the input value
+  value={}
```
kamleshchandnani, line 66: what are the cases for value and selected as 2 diff things? because what is selected will be displayed in the input right? 🤔

packages/blade/src/components/AutoComplete/_decisions/decisions.md
kamleshchandnani: The API you proposed is fine only. No other learning curve i can plug this right directly? and also do we need this? for eg: the ideal solution on mobile is to have an input and then open a bottomsheet right? so is this a valid pattern in dropdown's context? probably we can have 2 separate example to demo this? one with normal dropdown outside dropdown and one with dropdown inside?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+        <ActionListItem title="Hyderabad" value="hyderabad" />
+      </ActionList>
+    </BottomSheetBody>
+  </BottomSheet>
+</Dropdown>
+```
+
+### With Tags Outside
+
+When possible selections are more than 11, it is advised to put tags outside of the Input.
```
kamleshchandnani, line 167: why the magic number 11?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+<img width="700" alt="image" src="https://github.com/razorpay/blade/assets/30949385/8906eb6c-fae0-4e99-bd3f-d7e0edc60537">
+
+We add `showTags={false}` to AutoComplete to remove tags from input. Then we can use Controlled state to show tags outside in the UI.
+
+If not tags, what to show inside AutoComplete :thinking:? Checkout this table to understand -
+
+| Code                                                            | Output                                                                                                                     |
+| --------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------- |
+| `<AutoComplete label="Status" showTags={false} />`              | <img width="220" alt="image" src="https://github.com/razorpay/blade/assets/30949385/c8a30845-812c-493f-a51f-97c233880960"> |
+| `<AutoComplete accessibilityLabel="Status" showTags={false} />` | <img width="200" alt="image" src="https://github.com/razorpay/blade/assets/30949385/83cf63d5-ccc4-4bb9-bf0d-ce91dc5c1b09"> |
```
kamleshchandnani, line 178: add a description that when the label is not available we show the a11y label inside input. also on parallel thought, this looks weird visually. What if we create a hidden label and assign it the a11y label here and still keep the items selected as the first option? I also understand that doesn't sets context but the tradeoff is either we show something irrelevant i.e just a11y label or we show the items selected and leave it up to consumers to use the right copywriting

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+
+So when input has label, we show number of selected items.
```
kamleshchandnani, line 180: move this inside the table itself for better context?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+    </Tag>
+  )}
+</Box>;
+```
+
+### With Different Inactive - Active States
+
+| Code                                 | Inactive                                                                                                                   | Active                                                                                                                     |
+| ------------------------------------ | -------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------- |
+| `<AutoComplete rows="1" />`          | <img width="266" alt="image" src="https://github.com/razorpay/blade/assets/30949385/74b849f3-f1b9-4922-b717-70ffc7ffd56e"> | <img width="276" alt="image" src="https://github.com/razorpay/blade/assets/30949385/055bc4f4-fb57-4129-b27a-0f1966c8ad26"> |
```
kamleshchandnani, line 216: add some description that the number of tags shown will be dependent on the width of the Input?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
kamleshchandnani: so right now i can only have 1 | 3 what if i want to have other numbers too? why that restriction?

kamleshchandnani:
so well written doc ❤️ Very nice balance of Code + content + visual images 🎉

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: consistency mostly. Places where you can't layout shift UI needs "1" and places where you can shift UI, there we can standardise the number of rows. We settled on 3 there.

Without restriction, I would also be able to say rows="99" and make it infinitely increase

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+
+| **Props**          | **Description**                                                                                                      | **Type**                                                        | **Default Value** |
+| ------------------ | -------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------- | ----------------- |
+| label              | label of input                                                                                                       | string                                                          |                   |
+| accessibilityLabel | aria-label of input when label is not defined                                                                        | string                                                          |                   |
+| rows               | height restrictions of input ([Checkout "Inactive - Active States" Usage](#with-different-inactive---active-states)) | '1' \| '3' \| 'expandable'                                      | '3'               |
```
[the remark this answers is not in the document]
saurabhdaware: aha yes good point! Can decide between numberOfLines vs inputHeight="single" | "multiline"

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: Took it from design guidelines. Let me ask Abhishek 😆

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: In this value is `"O"` and selected is `['apple', 'bob-gordon-american-elderberry']`
image


We need 2 states because in 1 state, we can't differentiate between what is being written inside input vs what is selected.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: Like you said it doesn't set the context.

Imagine in this example (tbh it should just have visible label in this case but imagine such scenarios) we just show "4 items selected" but 4 items selected of what? Showing number of items selected makes sense when there is visual label to add context. 

When there is no label, there is nothing to suggest what exactly is selected. Imagine there are 3 such dropdowns side-by-side all showing "2 items selected" "4 items selected" "3 items selected" with no context.

Ideally label should always exist in such scenarios but when it isn't given, next most relevant thing to show is accessibilityLabel since that is mandatory when label is omitted.

image

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,240 @@
...
+### Controlled AutoComplete
+
+There are 3 controlled states that are relevant to AutoComplete
+
+1. Controlled state of filtering items
+2. Controlled state of the Input where user types
+3. Controlled state of selected items
+
+```jsx
+<AutoComplete
```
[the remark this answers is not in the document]
saurabhdaware: Didn't understand 🤔 You mean move that full code example from inside of `<details>` to here? the `onChange` is required because that makes the AutoComplete filter its items.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,248 @@
...
+
+| **Props**          | **Description**                                                                                                      | **Type**                                                        | **Default Value** |
+| ------------------ | -------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------- | ----------------- |
+| label              | label of input                                                                                                       | string                                                          |                   |
+| accessibilityLabel | aria-label of input when label is not defined                                                                        | string                                                          |                   |
+| labelPosition      | Label Position like other inputs but with new option `inside-input`                                                  | 'top' \| 'left' \| 'inside-input'                               | 'top'             |
```
kamleshchandnani, line 21: can we mention here only in a note that this applies to select input as well and not other input elements or maybe a section explaining the rationale of this entire scenario and why only we want to support it on select input families and then link the section here?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,248 @@
...
+| maxRows            | height restrictions of input ([Checkout "Inactive - Active States" Usage](#with-different-inactive---active-states)) | 'single' \| 'multiple' \| 'expandable'                          | '3'               |
```
kamleshchandnani, line 22:
```suggestion
| maxRows            | height restrictions of input ([Checkout "Inactive - Active States" Usage](#with-different-inactive---active-states)) | 'single' \| 'multiple' \| 'expandable'                          | 'single'               |
```

kamleshchandnani: or expandable depends what we decide

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,248 @@
...
+
+1. Controlled state of filtering items
+2. Controlled state of the Input where user types
+3. Controlled state of selected items
+
+```jsx
+<AutoComplete
+  // For controlling the input value
+  inputValue={}
+  onInputValueChange={(inputValue) => {}}
```
kamleshchandnani, line 67: nit: just swap the value and input value for better contextual readability?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: no don't move the full example but a few imp parts. right now it's hard to understand how/where is this `cities` coming from so filtering criteria is looking very confusing

you can just use the state variable, single-line and inputvaluechange and abstract out rest of the parts

packages/blade/src/components/AutoComplete/_decisions/decisions.md
kamleshchandnani: we now know right that in slot we'll only allow allow autocomplete so nothing else shall be removed as that can be added in slot that means it still remains as it is?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,248 @@
...
+</Dropdown>
+```
+
+### With Tags Outside
+
+When possible selections are more than 11, it is advised to put tags outside of the Input.
+
+> **Note**
+>
+> This screenshot is outdated. The input will show "Status (13 Selected)" instead of just "Status". Will update this soon.
```
kamleshchandnani, line 171: update this?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
kamleshchandnani: ??

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Props
+
+> **Note**
+>
+> Below this section, I have added [examples of common AutoComplete usage](#usage). Check them out to understand usage of these props in details.
+
+| **Props**          | **Description**                                                                                                      | **Type**                                                        | **Default Value** |
+| ------------------ | -------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------- | ----------------- |
+| label              | label of input                                                                                                       | string                                                          |                   |
```
kamleshchandnani, line 19: the type here would be `React.ReactText | React.ReactText[]` right to allow use cases like 
`Status: ${selectedOptions.length} selected`

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,248 @@
...
+| -------------------------------------------------------------- | --------------------- | ---------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------- |
+| `<AutoComplete label="Status" /> `                             | With Label Present    | <img width="500" src="https://github.com/razorpay/blade/assets/30949385/0c559daa-5cdc-4d71-9221-6073b7969b1b" /> | <img width="500" src="https://github.com/razorpay/blade/assets/30949385/9314fd11-96b7-46ef-963f-283b393cb9fd" /> |
+| `<AutoComplete label="Status" labelPosition="inside-input" />` | Without Label Present | <img width="500" src="https://github.com/razorpay/blade/assets/30949385/2cb1d659-4a98-4f63-96d8-204ead4a681f" /> | <img width="500" src="https://github.com/razorpay/blade/assets/30949385/0b219c4c-2df2-4cfe-a32d-4c02809ab440" /> |
```
kamleshchandnani, line 210: when we provide label position inside then label is what we show as prefix always? shouldn't that be configurable by user for better copywriting content?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: Yup I'll add a note

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: I think 'multiple` might be better default since it has the best UX among these 3. Only place where multiple can't use is in header where we cant have layout shifts. There people will realize its breaking UI and use single anyway.

I'll check with Abhishek also once

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: Oh it won't take the entire string. It will just take the first part of it (the label part). The selection part will be appended on our end by us

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: > when we provide label position inside then label is what we show as prefix always? shouldn't that be configurable by user for better copywriting content?

I am not sure if people will spend time getting it right. It might end up becoming inconsistent where 2 dropdowns on same page have 2 different ways of telling what is selected and what is label.

saurabhdaware: > for inside-input and multiselect why don't we do it like we do for label outside thing? show Status: 1 tag and +6more?

1. It takes more space which we won't have in smaller inputs.
2. Its a fairly rare usecase so we just wanted to do the easiest thing possible rather than spending extra time on this. E.g. Our number of tags to show and `+x more` calculations will need some extra fixing when we add another text element in the input.
3. Even when it is used, it is mostly expected to be used in headers where they are probably showing tags outside. In forms, they should just show label.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: Yupp we're going with what is proposed in doc. I'll remove this open question note.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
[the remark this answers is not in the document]
saurabhdaware: we're going with maxRows

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: we changed it to 10. Its just roughly the number of tags after which scrolling horizontally to deselect becomes difficult. If there are more tags and height is constrained, you have to scroll too much to reach the last tag

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,257 @@
...
+
+<AutoComplete
+  // For controlling the selections
+  value={selections}
+  onChange={({ name, values }) => {
+    setSelections(values)
+  }}
+  // For controlling the input value
+  inputValue={}
+  onInputValueChange={(inputValue) => {
```
nickisaacs, line 77: can we pass a promise here instead? We usually use autocomplete when we have to query APIs to fetch data. Passing promises here allows the autocomplete component to abort the previous promise/result whenever input changes before the API call has been resolved. It will relieve the caller from handling such things.

The same promise can also be used for static data

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Props
+
+> **Note**
+>
+> Below this section, I have added [examples of common AutoComplete usage](#usage). Check them out to understand usage of these props in details.
+
+| **Props**          | **Description**                                                                                                                             | **Type**                                                        | **Default Value** |
+| ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------- | ----------------- |
+| label              | label of input                                                                                                                              | string                                                          |                   |
```
nickisaacs, line 19: can we add a disabled / pending prop? for when a form submission is happening

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,257 @@
...
+| accessibilityLabel | aria-label of input when label is not defined                                                                                               | string                                                          |                   |
```
nashcheez, line 20: Hopefully we have an internal mechanism to mandate the `accessibilityLabel` when `label` is not defined? If yes, this should be mentioned in the props table.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,257 @@
...
+
+| Code                                    | Inactive                                                                                                                   | Active                                                                                                                     |
+| --------------------------------------- | -------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------- |
+| `<AutoComplete maxRows="single" />`     | <img width="266" alt="image" src="https://github.com/razorpay/blade/assets/30949385/74b849f3-f1b9-4922-b717-70ffc7ffd56e"> | <img width="276" alt="image" src="https://github.com/razorpay/blade/assets/30949385/055bc4f4-fb57-4129-b27a-0f1966c8ad26"> |
+| `<AutoComplete maxRows="multiple" />`   | <img width="420" alt="image" src="https://github.com/razorpay/blade/assets/30949385/ed71318c-889e-407e-86bf-fe4f5902dedf"> | <img width="416" alt="image" src="https://github.com/razorpay/blade/assets/30949385/8a636a3c-2afe-43a2-b283-c5ef40c14362"> |
```
nashcheez, line 229: With `maxRows="multiple"` are we limiting the number of rows to 3? If yes, this should be mentioned in the API decisions. If not, how are we deciding where to place the "+x more" indicator?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: Do you mean something like this? this will be possible yes. Let me know if you meant something else.

```js
onInputValueChange={async (inputValue) => {
   const newFilteredData = await fetchData(inputValue);
   setFilteredData(newFilteredData);
}}

// ...

filteredData.map((data) => <ActionList title={data.title} />
```

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: Yup its consistent for all inputs. Will add in the description as well.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: > With maxRows="multiple" are we limiting the number of rows to 3?

yes. Will add it in doc.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
nickisaacs: yeah but who does debouncing / aborting of requests when a new query is triggered, would the component handle it or is it left to the user

saurabhdaware: It will be upto the user. This callback will fire on every input change so debounce is something that consumers can add while calling the API.

How much you want to delay abort is contextual to how you're fetching data so we won't have enough context to take call there.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: added a description column in this table with explaination

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: done

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,253 @@
...
+| ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------- | ----------------- |
+| label              | label of input                                                                                                                              | string                                                          |                   |
+| accessibilityLabel | aria-label of input when label is not defined (required when label is not passed)                                                           | string                                                          |                   |
+| labelPosition      | Label Position like other inputs but with new option `inside-input`. This value new value will be added to `AutoComplete` and `SelectInput` | 'top' \| 'left' \| 'inside-input'                               | 'top'             |
```
cseas, line 21:
```suggestion
| labelPosition      | Label Position like other inputs but with new option `inside-input`. This new value will be added to `AutoComplete` and `SelectInput` | 'top' \| 'left' \| 'inside-input'                               | 'top'             |
```

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,253 @@
...
+| `<AutoComplete maxRows="single" />`     | Input always stays 1 line                                   | <img width="266" alt="image" src="https://github.com/razorpay/blade/assets/30949385/74b849f3-f1b9-4922-b717-70ffc7ffd56e"> | <img width="276" alt="image" src="https://github.com/razorpay/blade/assets/30949385/055bc4f4-fb57-4129-b27a-0f1966c8ad26"> |
+| `<AutoComplete maxRows="multiple" />`   | Input can go upto 3 lines                                   | <img width="420" alt="image" src="https://github.com/razorpay/blade/assets/30949385/ed71318c-889e-407e-86bf-fe4f5902dedf"> | <img width="416" alt="image" src="https://github.com/razorpay/blade/assets/30949385/8a636a3c-2afe-43a2-b283-c5ef40c14362"> |
+| `<AutoComplete maxRows="expandable" />` | Input is 1 line initially but expands upto 3 lines on click | <img width="385" alt="image" src="https://github.com/razorpay/blade/assets/30949385/21b63bc4-804a-4180-953c-119a2b676cd5"> | <img width="394" alt="image" src="https://github.com/razorpay/blade/assets/30949385/1131a1bd-2989-4bfc-b659-207088be3bfb"> |
```
cseas, line 226: There's no chevron up icon in the expanded state screenshot here. Is that intended? How do we collapse it back?

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,253 @@
...
+
+AutoComplete will be built as a new trigger for existing Dropdown component.
+
+Depending on where and how AutoComplete is used, the API can differ. Checkout examples below on how code will look like for a particular usage.
+
+- [Basic Usage](#basic-usage)
+- [Controlled AutoComplete](#controlled-autocomplete)
+- [In BottomSheet](#in-bottomsheet)
+- [With Tags Outside](#with-tags-outside)
+- [With Different Inactive - Active States](#with-different-inactive---active-statess)
```
cseas, line 38:
```suggestion
- [With Different Inactive - Active States](#with-different-inactive---active-states)
```

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,253 @@
...
+| -------------------------------------------------------------- | --------------------- | ---------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------- |
+| `<AutoComplete label="Status" /> `                             | With Label Present    | <img width="500" src="https://github.com/razorpay/blade/assets/30949385/0c559daa-5cdc-4d71-9221-6073b7969b1b" /> | <img width="500" src="https://github.com/razorpay/blade/assets/30949385/9314fd11-96b7-46ef-963f-283b393cb9fd" /> |
+| `<AutoComplete label="Status" labelPosition="inside-input" />` | Without Label Present | <img width="500" src="https://github.com/razorpay/blade/assets/30949385/2cb1d659-4a98-4f63-96d8-204ead4a681f" /> | <img width="500" src="https://github.com/razorpay/blade/assets/30949385/0b219c4c-2df2-4cfe-a32d-4c02809ab440" /> |
+
+### With Different Inactive - Active States
```
cseas, line 216: Please add Positive/Negative intent examples

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: ops not it will exist. It got cropped I guess. Will change image.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,253 @@
...
+| onInputValueChange | Callback when input value changes                                                                                                           | (inputValue: string) => void                                    |                   |
+| value              | Controlled state of the selected items ([Checkout "Controlled AutoComplete" Usage](#controlled-autocomplete))                               | string[]                                                        |                   |
+| onChange           | Callback when selected items change                                                                                                         | ({ name, values }: { name: string; values: string[] } ) => void |                   |
+
+## Usage
+
+AutoComplete will be built as a new trigger for existing Dropdown component.
+
+Depending on where and how AutoComplete is used, the API can differ. Checkout examples below on how code will look like for a particular usage.
```
Jaynil1611, line 32: @saurabhdaware I couldn't find any example for a normal autocomplete dropdown without select input.
We have a use case where we want to show a single select autocomplete dropdown. [Figma](https://www.figma.com/file/cdTS7TAdvN4vhGiIH27s25/Purchase-Orders-V1?type=design&node-id=746-296934&mode=dev)

Also, the links you've attached allow a single select autocomplete dropdown. [ComboBox](https://ariakit.org/examples/combobox-filtering).

saurabhdaware: I think once we release AutoComplete, design team can replace this with our version. I only see chevron icon missing here in design which will by default be added to all autocompletes from our end

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: yup this will be supported. This is basically `validationState` variable on input. 

I'll add a note about supporting common input properties in doc.

divyanshu013 (dismissed):
Looks good to me.

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: The filtering API is changed due to implementation limitations, but I think same thing applies to new API as well so resolving the conversation.

divyanshu013 (commented):
There might be some discoverability issues with `filteredValues` based API but seems like the best way forward. Are all the props optional? Can we work with TS to suggest users some props are required when the opt into controlled mode?

saurabhdaware:
> There might be some discoverability issues with filteredValues based API but seems like the best way forward. Are all the props optional? Can we work with TS to suggest users some props are required when the opt into controlled mode?

There isn't any isControlled prop to figure out if the component is controlled or not. Adding those props is makes that part controlled.

E.g. in AutoComplete, you can have onInputValue and inputValue controlled but selections and filtering uncontrolled. Or selections controlled and input values uncontrolled.

The only scenario where we might be able to do that is when we have `filteredValues` in that case we might be able to make `onInputValueChange` required. Although not sure if that is necessary since if you use filteredValues without handling input change, AutoComplete won't work in the first place. It will show static values always. I do have plans to add better examples in documentation to show all controlled states correctly

divyanshu013 (dismissed):
:shipit:

pcprashant:
**Custom Filter Options:**
- Issue: The current spec doesn't seem to offer the ability to provide custom filter logic for the uncontrolled variation of the dropdown.
- Suggestion: Can we introduce a filterOptions prop for this? It can be useful, considering there are different use-cases like implementing a fuzzy search, localization-related requirements (like removing accents), and other custom filtering based on values.

**Resetting Input Behavior:**
- Issue: The spec is unclear on when the input value gets reset.
- Question: Is the input reset on making a selection?
- Suggestion: It would be beneficial to have a prop to dictate reset behavior, allowing the user to determine if the input should reset on blur, close, or upon selection.

**Grouped Options Example:**
- Issue: While the spec talks about options, there's no clear example showcasing grouped options and custom labels.
- Suggestion: It would be helpful to include an example demonstrating the usage of grouped options along with custom labeling.

**Close on Select Behavior:**
- Issue: The behavior post-selection isn't articulated.
- Suggestion: A prop to decide if the dropdown should close upon selection would be valuable, especially for multi-selection scenarios.

**Loading State:**
- Issue: Handling of asynchronous data loading states isn't addressed.
- Suggestion: Introducing an isLoading prop to display a loader within the input would be handy for cases where data loading happens asynchronously.

**Filtering Selected Items:**

- Issue: The spec doesn't discuss the behavior of already selected items in the dropdown.
- Suggestion: Can we provide a prop that determines if the selected items should be filtered out from the dropdown options?

**Ref-based Control:**
- Issue: The ability to control the dropdown externally isn't mentioned.
- Question: Can the dropdown be opened or closed using its ref from external elements?


// @saurabhdaware @kamleshchandnani

saurabhdaware:
> Custom Filter Options:

The exact usecase is mentioned in [Controlled AutoComplete API](https://github.com/razorpay/blade/blob/feat/autocomplete/api-decision/packages/blade/src/components/AutoComplete/_decisions/decisions.md#controlled-autocomplete). We have `filteredValues` prop.


> Resetting Input Behavior:

This is also mentioned in the example of Controlled AutoComplete API above ^^. `inputValue` and `onInputValue` props are for this usecase. There will be a predefined pattern where we clear inputs on selection.

> Grouped Options Example:

Can you elaborate what you mean by grouped options? if you mean having different sections, then that is supported in our ActionList today - [ActionListSection API](https://github.com/razorpay/blade/blob/feat/autocomplete/api-decision/packages/blade/src/components/Dropdown/_decisions/decisions.md#actionlistsection)

> Close on Select Behavior:

We have defined a standard UX pattern here for both Select and AutoComplete that -
- Single Select: Always closes on selecting item
- Multi Select: Always stays open after selections.

It might not be a good UX pattern to close multiselect after every selection. Do ask your designers to reach out to our designers if you have such usecase.

> Loading State:

I dont think so AutoComplete or SelectInput should be the elements that go into loading state. When data is loading, its loading inside DropdownOverlay right? you can use our SkeletonLoader or Spinner component there while data is loading.

It's also subjective. For example if you want to load 100 items, and you have first 10 items. You can just not show the rest 90 items while they are loading and add to ActionList when they are done loading.

This can be implemented in current Dropdown and SelectInput as well. Do check with your designer and let me know if there is any UX loading pattern that they are expecting that is not possible to be built using AutoComplete or SelectInput

> Filtering Selected Items:

Here as well we have defined a standard pattern that selected items will stay in ActionList and they will be marked as selected. Users can click them again and mark as non-selected in multiselect or select a different item in single select. Let me know if your designer is expecting to remove the selected items, I can check with our designers then.

> Ref-based Control:

 [Controlled AutoComplete API](https://github.com/razorpay/blade/blob/feat/autocomplete/api-decision/packages/blade/src/components/AutoComplete/_decisions/decisions.md#controlled-autocomplete) mentions about controlled selections, controlled inputValue and controlled filtering. 

> Question: Can the dropdown be opened or closed using its ref from external elements?

This is a good point. For this we have P0 issue open - #1436. We'll be fixing this after AutoComplete release so Dropdown's overlay can be controlled as well.

cseas (approved):

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -14,16 +14,19 @@ A SelectInput where you can type inside the input to filter through the items.
...
+| onInputValueChange | Callback when input value changes                                                                                                           | (inputValue: string) => void                                    |                                                  |
+| value              | Controlled state of the selected items ([Checkout "Controlled AutoComplete" Usage](#controlled-autocomplete))                               | string[]                                                        |                                                  |
+| onChange           | Callback when selected items change                                                                                                         | ({ name, values }: { name: string; values: string[] } ) => void |                                                  |
+
+_This is a list of unique props to AutoComplete, apart from these all common input props will be supported such as `validationState`, `helpText`, `errorText`, etc_
```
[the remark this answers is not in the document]
saurabhdaware: added edb0009b1ca1258bb5a0c425dbec0d815317fef9

kamleshchandnani (approved):

saurabhdaware:
Followup from UX Concerns we had on call. 

Hey @pcprashant, I discussed your concerns with the design team. These are the things that came out of it -

> - closing dropdown on every selection in multiselect 

We have decided to follow default behaviours of AriaKit, and Twilio. Regarding the concerns on People not knowing how to close Dropdown, we do allow having Close button on Footer although that shouldn't be needed as well since clicking outside is a common pattern that users usually know about.

> - clearing the input on dismiss of dropdown

We have decided to keep retaining the value as default behaviour similar to [Twilio](https://paste.twilio.design/components/multiselect-combobox). In case of form submissions, you can validate your AutoComplete input the way you would validate other inputs. If you feel there is any case where value should be cleared, its possible to implement that with combination of onDismiss and inputValue controls. We will add example in documentation for this.


> - loading state on input

As per design, loading state should be on the dropdown overlay and not input when items are loading (Inspirations: Primer). Your point regarding having to write more code in this case was valid. In future, we can think of adding a isLoading prop on DropdownOverlay to help you do this with less code but for now it should be possible by adding your Spinner to DropdownOverlay when things are loading.

cseas (approved):

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,288 @@
...
+| `<AutoComplete maxRows="multiple" />`   | Input can go upto 3 lines                                   | <img width="420" alt="image" src="https://github.com/razorpay/blade/assets/30949385/ed71318c-889e-407e-86bf-fe4f5902dedf"> | <img width="416" alt="image" src="https://github.com/razorpay/blade/assets/30949385/8a636a3c-2afe-43a2-b283-c5ef40c14362"> |
+| `<AutoComplete maxRows="expandable" />` | Input is 1 line initially but expands upto 3 lines on click | <img width="385" alt="image" src="https://github.com/razorpay/blade/assets/30949385/21b63bc4-804a-4180-953c-119a2b676cd5"> | <img width="416" alt="image" src="https://github.com/razorpay/blade/assets/30949385/8a636a3c-2afe-43a2-b283-c5ef40c14362"> |
+
+## Accessibility
+
+AutoComplete will continue to follow [same keyboard navigations as SelectInput](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Dropdown/_decisions/decisions.md#accessibility).
+
+## Referrences
```
akkhil7, line 267:
```suggestion
## References
```

packages/blade/src/components/AutoComplete/_decisions/decisions.md
```diff
@@ -0,0 +1,288 @@
...
+        // Selected Items in AutoComplete
+        value={currentSelections}
+        onChange={({ values }) => {
+          setCurrentSelections(values);
+        }}
+        // controls the values that should be shown to user
+        filteredValues={filteredValues}
+      />
+      <DropdownOverlay>
+        <ActionList>
```
akkhil7, line 166: question: Does ActionList automatically handle the already selected items and move it down to the bottom, because showing it in the dropdown could cause clutter and could hinder the user interaction, right? Or is it upto the dev. I meant something like - `[...selectableItems, ...selectedItems]` ?

saurabhdaware: No it will keep its place because moving items might be a bit more chaotic UX than keeping them there and having check mark next to it.

We have a good visual indicator of checkbox in actionlistitems so it will distinguish it properly with non-selected items. We follow similar UX as [AriaKit](https://ariakit.org/examples/combobox-multiple)

packages/blade/src/components/AutoComplete/_decisions/decisions.md
saurabhdaware: Will merge this with spelling mistake only otherwise I'll have to ask people to approve again. The approvals get dismissed 😆

saurabhdaware:
@pcprashant will merge this based on the [thumbs up](https://razorpay.slack.com/archives/C01H13RTF8V/p1694080171707359?thread_ts=1693908393.845889&cid=C01H13RTF8V). Let me know if you have anymore comments, I can send followup PR if any change is required.

Issue #1528, opened 2023-08-10
Upgrading to v7+ fails on react native due to errors from floating-ui/react


chaitanyadeorukhkar:
Ref: https://razorpay.slack.com/archives/C01H13RTF8V/p1691652926526559

PR #1529, opened 2023-08-10
fix: react native errors with floating-ui/react

chaitanyadeorukhkar:
Fixes: #1528 

We were importing `detectOverflow` from `@floating-ui/react` in a common file that was being imported in web and native. 

Originally introduced in: https://github.com/razorpay/blade/pull/1284

### DIFF
--- a/packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
+++ b/packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
@@ -2,9 +2,10 @@ import React from 'react';
 import throttle from 'lodash/throttle';
 import styled, { keyframes, css } from 'styled-components';
 import type { FlattenSimpleInterpolation } from 'styled-components';
-import { autoUpdate, useFloating } from '@floating-ui/react';
+import { autoUpdate, detectOverflow, useFloating } from '@floating-ui/react';
+import type { MiddlewareState } from '@floating-ui/react';
 import type { DropdownPosition } from './dropdownUtils';
-import { componentIds, getDropdownOverflowMiddleware } from './dropdownUtils';
+import { componentIds } from './dropdownUtils';
 import { useDropdown } from './useDropdown';
 import { StyledDropdownOverlay } from './StyledDropdownOverlay';
 import type { DropdownOverlayProps } from './types';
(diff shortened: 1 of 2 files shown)


anuraghazra (approved):

snitin315 (dismissed):

packages/blade/src/components/Tooltip/types.ts
```diff
@@ -1,8 +1,10 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import type { Side, UseFloatingOptions } from '@floating-ui/react';
```
anuraghazra, line 2: Very weird, it's a type import but still resolving the module. type imports are supposed to be stripped out from runtime. Not sure why this is occurring.

chaitanyadeorukhkar: I can see `@floating-ui/react` import on native bundle only coming it for `useFloatingOptions` so this is the only place where it could come from. I'll send Shubhnik this PR to try out before we conclude

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -58,6 +59,99 @@ const AnimatedOverlay = styled(StyledDropdownOverlay)<{
     `,
 );
 
+/**
+ * This function calculates the position of dropdown overlay with respect to dropdown trigger element.
+ * For non-menus (e.g SelectInput), position is flipped if overflow is on bottom.
+ * For menus (e.g. DropdownButton), position is flipped if overflow is on right or bottom.
+ * Additional spacing is added to clientHeight to provide spacing above the dropdown trigger.
+ */
+const getDropdownOverlayPosition = ({
```
anuraghazra, line 68: Oh okay so it wasn't because of type imports. That makes sense now.

anuraghazra (approved):
:shipit:

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
chaitanyadeorukhkar: Yup

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -2,9 +2,10 @@ import React from 'react';
 import throttle from 'lodash/throttle';
 import styled, { keyframes, css } from 'styled-components';
 import type { FlattenSimpleInterpolation } from 'styled-components';
-import { autoUpdate, useFloating } from '@floating-ui/react';
+import { autoUpdate, detectOverflow, useFloating } from '@floating-ui/react';
```
snitin315, line 5: Maybe we should update our eslint rule for this case

chaitanyadeorukhkar: How? 
1. Its okay to import type from a web package in a common file
2. Its okay to import anything platform specific in a platform-specific file

Wont this be a complex check?

PR #1535, opened 2023-08-14
docs(Chip): add API decisions

snitin315:
This pull request introduces a comprehensive **API Design Document for the `Chip` and `ChipGroup` components**, facilitating a clear understanding and streamlined development of these crucial UI elements.

> **Note**
>
> You can view the API Design Document in a more reader-friendly format by following [this link](https://github.com/razorpay/blade/blob/feat/chips-api/packages/blade/src/components/Chip/_decisions/decisions.md).

**Changes Introduced:**

- Inclusion of a detailed API specification for the `Chip` and `ChipGroup` components.
- Explication of the accessibility considerations, elaborating on the adherence to ARIA patterns for radio and checkbox interactions.

This document will serve as a valuable reference point for the development team, facilitating coherent discussions and informed decisions as we progress toward the implementation phase. We would greatly encourage feedback and suggestions as we strive for the best possible outcome.


packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,121 @@
...
+| **size**                | `'xsmall' \| 'small' \| 'medium' \| 'large'`                         | `'small'`    | Specifies the size of the rendered Chip                                                                                                            | ❌       |
+| **inputType**           | `'checkbox' \| 'radio'`                                              | `'checkbox'` | Determines the Chip's input type                                                                                                                   | ❌       |
+| **value (html native)** | `string`                                                             | `undefined`  | The value of the input field in Chip, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) | ❌       |
+| **variant**             | `'positive' \| 'negative' \| 'notice' \| 'information' \| 'neutral'` | `'neutral'`  | Sets the Chip's visual variant                                                                                                                     | ❌       |
```
anuraghazra, line 38: since we are saying Chip cannot be used without ChipGroup, we can also make these props available on ChipGroup so that consumers doesn't have to repeat them each time. 

I think we need to atleast have these props on group: `variant`, `size`, `isDisabled`

```jsx
<ChipGroup variant="positive">
...
</ChipGroup>
```

Similar to how we have in Checkbox and Radio Group

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,121 @@
...
+| **isChecked**           | `boolean`                                                            | `false`      | Controls the Chip's checked state                                                                                                                  | ❌       |
+| **isDisabled**          | `boolean`                                                            | `false`      | Disables or enables the Chip                                                                                                                       | ❌       |
+| **size**                | `'xsmall' \| 'small' \| 'medium' \| 'large'`                         | `'small'`    | Specifies the size of the rendered Chip                                                                                                            | ❌       |
+| **inputType**           | `'checkbox' \| 'radio'`                                              | `'checkbox'` | Determines the Chip's input type                                                                                                                   | ❌       |
```
anuraghazra, line 36: This probably doesn't need to be exposed to consumers, you can detect this internally since ChipGroup will have selectionType prop.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,121 @@
...
+This adherence to established accessibility practices is further substantiated by the following guidelines:
+
+- **Radio Accessibility** - The `Chip` component follows the [W3C ARIA Radio Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/) to ensure that users utilizing assistive technologies can seamlessly interact with and understand the radio selection mechanism.
+- **Checkbox Accessibility** - Similarly, the `Chip` component aligns with the [W3C ARIA Checkbox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/) to guarantee that users with disabilities can effectively engage with the checkbox selection paradigm.
+
+These deliberate design choices manifest in a user experience that respects accessibility guidelines, enabling users of diverse abilities to interact with the components with equal ease.
+
+## Open Questions
+
+- Should we provide a `name` prop to the `ChipGroup` component, similar to the `RadioGroup` component?
```
anuraghazra, line 121: Probably without the `name` prop a radios won't work. 

Although we can set a default random id to the name prop internally, we should expose a `name` prop so that consumers can provide a human readable ID to the group for form submission purposes. 

This is already handled in useRadioGroup.

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: @anuraghazra I feel all chips should be of the same size within ChipGroup, hence we can provide `size` prop only for `ChipGroup`. WDYT?

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,127 @@
...
+| **isDisabled**          | `boolean`                                                            | `false`     | Disables or enables the Chip                                                                                                                       | ❌       |
+| **size**                | `'xsmall' \| 'small' \| 'medium' \| 'large'`                         | `'small'`   | Specifies the size of the rendered Chip                                                                                                            | ❌       |
+| **value (html native)** | `string`                                                             | `undefined` | The value of the input field in Chip, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) | ❌       |
+| **variant**             | `'positive' \| 'negative' \| 'notice' \| 'information' \| 'neutral'` | `undefined` | Sets the Chip's visual variant                                                                                                                     | ❌       |
```
saurabhdaware, line 35: Should this be `intent` or `variant` 🤔

snitin315: `variant`, similar to `badge`.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,127 @@
...
+This adherence to established accessibility practices is further substantiated by the following guidelines:
+
+- **Radio Accessibility** - The `Chip` component follows the [W3C ARIA Radio Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/radio/) to ensure that users utilizing assistive technologies can seamlessly interact with and understand the radio selection mechanism.
+- **Checkbox Accessibility** - Similarly, the `Chip` component aligns with the [W3C ARIA Checkbox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/) to guarantee that users with disabilities can effectively engage with the checkbox selection paradigm.
+
+Utilizing the established `useRadio`, `useCheckbox`, `useRadioGroup`, & `useCheckboxGroup` hooks hooks ensures that the `Chip` and `ChipGroup` components inherently maintain accessibility as a fundamental aspect.
+
+These deliberate design choices manifest in a user experience that respects accessibility guidelines, enabling users of diverse abilities to interact with the components with equal ease.
+
+## Open Questions
```
snitin315, line 122: https://github.com/razorpay/blade/pull/1535#discussion_r1295362618

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,127 @@
...
+| **selectionType** | `'single' \| 'multiple'`                                                 | `'single'`  | Defines the selection behavior within the ChipGroup component                                                                  | ❌       |
+| **size**          | `'xsmall' \| 'small' \| 'medium' \| 'large'`                             | `'small'`   | Specifies the size of the rendered Chips withing the ChipGroup                                                                 | ❌       |
+| **value**         | `string \|string[]`                                                      | `undefined` | Acts as a controlled component by specifying the ChipGroup value                                                               | ❌       |
+| **variant**       | `'positive' \| 'negative' \| 'notice' \| 'information' \| 'neutral'`     | `'neutral'` | Sets the ChipGroups's visual variant, it will propagate down to all the Chips                                                  | ❌       |
```
akkhil7, line 49: question: will there be a use-case where every chip will be a single variant that is not `neutral`?

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,127 @@
...
+
+## `Chip` API
+
+> **Note**
+>
+> The `Chip` component is inherently tied to the `ChipGroup` and cannot be utilized outside its context.
+
+| Prop                    | Type                                                                 | Default     | Description                                                                                                                                        | Required |
+| ----------------------- | -------------------------------------------------------------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| **children**            | `string`                                                             | `undefined` | Renders the label text for the Chip                                                                                                                | ✅       |
```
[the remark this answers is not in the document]
snitin315: @akkhil7 The text will truncate if it exceeds the max width. In such cases it is advised to use checkboxes/radio, Chips should be used for short labels.
 
Screenshot 2023-08-17 at 2 50 42 PM

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: There can be. I have it as one of the open questions https://github.com/razorpay/blade/pull/1535#discussion_r1295362618

packages/blade/src/components/Chip/_decisions/decisions.md
anuraghazra: Sounds good to me, size/variant probably you can put only on group. But ekbar designers se check kar lena

packages/blade/src/components/Chip/_decisions/decisions.md
akkhil7: I meant from a design standpoint, would there be use-cases (might need to check with design team)? I can think of chips being all neutral which would be the default variant. But I can't think of chips all being any other variant. If so, then I guess the control can be left to the individual Chip and not ChipGroup. 

P.S: My whole assumption could be wrong.

packages/blade/src/components/Chip/_decisions/decisions.md
saurabhdaware: For Badge there was chance of `blue` or `indigo` coming up. We currently use `intent` on other components 🤔 

Here's details of that discussion - https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Badge/_decisions/decisions.md#variant-prop

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,127 @@
...
+### Basic/Uncontrolled Usage
+
+Single Chip selection:
+
+```jsx
+<ChipGroup defaultValue="1">
+  <Chip value="1">Single chip</Chip>
+  <Chip value="2">Can be selected</Chip>
+  <Chip value="3">At a time</Chip>
+</ChipGroup>
```
kamleshchandnani, line 64: please add better example in Razorpay's context

kamleshchandnani (changes requested):

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,127 @@
...
+| **children**      | `React.ReactNode`                                                        | `undefined` | Accepts multiple `Chip` components                                                                                             | ✅       |
+| **defaultValue**  | `string \|string[]`                                                      | `undefined` | Sets the initial value of the Chip group                                                                                       | ❌       |
+| **isDisabled**    | `boolean`                                                                | `false`     | Disables or enables ChipGroup, it will propagate down to all the Chips                                                         | ❌       |
+| **name**          | `string`                                                                 | `undefined` | The name of the chip group, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name) | ❌       |
```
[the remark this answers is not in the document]
snitin315: @kamleshchandnani Chip is an input element, it will be either a radio or checkbox based on the `selectionType` prop.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,127 @@
...
+| **icon**                | `Icon`                                                               | `undefined` | Displays the Blade Icon component within the Chip                                                                                                  | ❌       |
+| **isDisabled**          | `boolean`                                                            | `false`     | Disables or enables the Chip                                                                                                                       | ❌       |
+| **size**                | `'xsmall' \| 'small' \| 'medium' \| 'large'`                         | `'small'`   | Specifies the size of the rendered Chip                                                                                                            | ❌       |
+| **value (html native)** | `string`                                                             | `undefined` | The value of the input field in Chip, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) | ❌       |
```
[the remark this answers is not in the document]
snitin315: @kamleshchandnani This is similar to the [Radio API](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Radio/_decisions/decisions.md#radio). This value will also help in filtering the chips when we have `defaultValue` on `ChipGroup`.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,127 @@
...
+| ----------------------- | -------------------------------------------------------------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| **children**            | `string`                                                             | `undefined` | Renders the label text for the Chip                                                                                                                | ✅       |
+| **icon**                | `Icon`                                                               | `undefined` | Displays the Blade Icon component within the Chip                                                                                                  | ❌       |
+| **isDisabled**          | `boolean`                                                            | `false`     | Disables or enables the Chip                                                                                                                       | ❌       |
```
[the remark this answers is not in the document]
snitin315: `disabled` state is provided in the design. Abhishek and @saurav12 can give more insights on usecases.

packages/blade/src/components/Chip/_decisions/decisions.md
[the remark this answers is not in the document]
snitin315: @kamleshchandnani it's not the text value, that will be the children. `<Chip value="submission-value"> Show on Screen</Chip>`

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,127 @@
...
+| **defaultValue**  | `string \|string[]`                                                      | `undefined` | Sets the initial value of the Chip group                                                                                       | ❌       |
+| **isDisabled**    | `boolean`                                                                | `false`     | Disables or enables ChipGroup, it will propagate down to all the Chips                                                         | ❌       |
+| **name**          | `string`                                                                 | `undefined` | The name of the chip group, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name) | ❌       |
+| **onChange**      | `({ name, value }: { name: string; value: string \| string[] }) => void` | `undefined` | Calls a function on any state change within the ChipGroup                                                                      | ❌       |
```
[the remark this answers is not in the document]
snitin315: @kamleshchandnani The type of `value` depends on the `selectionType` prop. If its `single` value will be of `string` type and if it's `multiple` it will be `string[]` simulating radio/checkbox-like behavior respectively. So users won't have to add a check like `Array.isArray(value)`, since it will be either one of the cases only.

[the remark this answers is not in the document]
kamleshchandnani: yes i understand that so the call we took in select was that we'll always make this array so if its single value the array will always have single value. because imagine if you make the dropdown multiple/single at runtime based on API response or if you change it manually in code you don't have to worry about the onchange function.so better DX and less error prone

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,140 @@
...
+
+## `Chip` API
+
+> **Note**
+>
+> The `Chip` component is inherently tied to the `ChipGroup` and cannot be utilized outside its context.
+
+| Prop                    | Type      | Default     | Description                                                                                                                                        | Required |
+| ----------------------- | --------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| **children**            | `string`  | `undefined` | Renders the label text for the Chip                                                                                                                | ✅       |
```
kamleshchandnani, line 30: wouldn't be children because we might want to support the template strings `Bussiness ${api.busineessType}`. We shall do `StringChildrenType` from our `utils` which we have created for use in Radio/Checkbox/Text which is nothing but `React.ReactText | React.ReactText[];`

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,140 @@
...
+| **selectionType** | `'single' \| 'multiple'`                                                 | `'single'`  | Defines the selection behavior within the ChipGroup component, simulating radio/checkbox like behavior within the group        | ❌       |
+| **size**          | `'xsmall' \| 'small' \| 'medium' \| 'large'`                             | `'small'`   | Specifies the size of the rendered Chips withing the ChipGroup                                                                 | ❌       |
+| **value**         | `string \|string[]`                                                      | `undefined` | Acts as a controlled component by specifying the ChipGroup value                                                               | ❌       |
+| **intent**        | `'positive' \| 'negative' \| 'notice' \| 'information' \| 'neutral'`     | `'neutral'` | Sets the ChipGroups's visual variant, it will propagate down to all the Chips                                                  | ❌       |
```
kamleshchandnani: lets confirm with design what all intents we really want to support

packages/blade/src/components/Chip/_decisions/decisions.md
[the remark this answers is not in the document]
snitin315: @kamleshchandnani I see, makes sense. I'll update it.

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: Updated.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,140 @@
...
+
+```jsx
+function Single() {
+  // string value when selectionType is 'single' (default)
+  const [selected, setSelected] = useState('proprietorship');
+
+  return (
+    <Box>
+      <Text> Select Business type: </Text>
+      <ChipGroup selectionType="single" value={value} onChange={({ value }) => setSelected(value)}>
```
kamleshchandnani, line 91:
```suggestion
      <ChipGroup selectionType="single" value={selected} onChange={({ value }) => setSelected(value)}>
```
??

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,140 @@
...
+function Multiple() {
+  // array of strings value when selectionType is 'multiple'
+  const [selected, setSelected] = useState(['in-progress', 'pending']);
+
+  return (
+    <Box>
+      <Text> Quick Filters: </Text>
+      <ChipGroup
+        selectionType="multiple"
+        value={value}
```
kamleshchandnani, line 109:
```suggestion
        value={selected}
```
??

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: We discussed this internally, and we decided that all Chips should have same intent within a ChipGroup, hence I've updated the API to remove the `intent` prop from `Chip`

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: Yes, in code I've used `StringChildrenType` only. Let me update it in the doc too.

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: My bad!

packages/blade/src/components/Chip/_decisions/decisions.md
anuraghazra: > but badge had use case for intent+custom colors hence we made it variant here i don't see that

I would lean towards variant, intent is not scalable. 

Abhi nehi hain but bad mein we might have usecase for blue or white chips (similar to how we need white button/links). 
If we go with intent we will later have to make a breaking change, similarly how we had to do in Counter.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,144 @@
...
+Here are a few illustrative examples showcasing the utilization of the `Chip` and `ChipGroup` components:
+
+### Basic/Uncontrolled Usage
+
+Single Chip selection:
+
+```jsx
+<Box>
+  <Text> Select Business type: </Text>
+  <ChipGroup defaultValue={['proprietorship']}>
```
kamleshchandnani, line 60:
```suggestion
  <ChipGroup defaultValue="proprietorship">
```

the example says single selection 🙈

kamleshchandnani (dismissed):
one tiny typo, rest lgtm

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,144 @@
...
+  </ChipGroup>
+</Box>
+```
+
+### Controlled Usage
+
+```jsx
+function Single() {
+  // Only one value in array when selectionType is 'single' (default)
+  const [selected, setSelected] = useState(['proprietorship']);
```
snitin315, line 86:
```suggestion
  // string value when selectionType is 'single' (default)
  const [selected, setSelected] = useState('proprietorship');
```

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,144 @@
...
+## `Chip` API
+
+> **Note**
+>
+> The `Chip` component is inherently tied to the `ChipGroup` and cannot be utilized outside its context.
+
+| Prop                    | Type                                   | Default     | Description                                                                                                                                        | Required |
+| ----------------------- | -------------------------------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| **children**            | `React.ReactText \| React.ReactText[]` | `undefined` | Renders the label text for the Chip                                                                                                                | ✅       |
+| **icon**                | `Icon`                                 | `undefined` | Displays the Blade Icon component within the Chip                                                                                                  | ❌       |
```
cseas, line 31: Trailing icon not supported?

Screenshot 2023-08-22 at 3 51 08 PM

[Design][1]

[1]: https://www.figma.com/file/xZP4W0zCEVyjpxn12XtQ96/GPT-for-lead-generation?type=design&node-id=561-171566&mode=design&t=18xlkeF6bSuCXqX6-4

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,144 @@
...
+
+## `Chip` API
+
+> **Note**
+>
+> The `Chip` component is inherently tied to the `ChipGroup` and cannot be utilized outside its context.
+
+| Prop                    | Type                                   | Default     | Description                                                                                                                                        | Required |
+| ----------------------- | -------------------------------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| **children**            | `React.ReactText \| React.ReactText[]` | `undefined` | Renders the label text for the Chip                                                                                                                | ✅       |
```
cseas, line 30: If we're restricting the users to only pass a string, please add a `textTransform` prop on `Chip` to control the text casing. 

We've had earlier Chip designs that used upper case while the Blade design examples are using capital case.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,144 @@
...
+
+Here are a few illustrative examples showcasing the utilization of the `Chip` and `ChipGroup` components:
+
+### Basic/Uncontrolled Usage
+
+Single Chip selection:
+
+```jsx
+<Box>
+  <Text> Select Business type: </Text>
```
cseas, line 59: Please add a `label` prop similar to other input components. Using `Text` for label here would disrupt the UI consistency. The use-case of Chip we've seen was used alongside other Input components.

Screenshot 2023-08-22 at 4 00 56 PM

divyanshu013: Right, matching the form UI would need additional efforts. Eg. you likely need a margin bottom additionally to the `Text` which should match what `label` already applies to other input fields.
Also, how would this affect a11y of forms if we don't have a label?

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,144 @@
...
+        <Chip value="refunded">Refunded</Chip>
+        <Chip value="failed">Failed</Chip>
+        <Chip value="pending">Pending</Chip>
+        <Chip value="in-progress">In Progress</Chip>
+      </ChipGroup>
+    </Box>
+  );
+}
+```
+
```
divyanshu013, line 125: Might be good to add a form submission based example as well

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: Yes,  `<Chip> {value.toUpperCase()} </Chip>` is allowed.

cseas: All consumers can write a javascript util for transformText yes but why go to JS for something that can be done with CSS?

cseas: @snitin315 Can you please add this example in the docs? Generally we'd use CSS for this use-case, not JS.

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: @cseas For consistency we will support leading icons only, the designs will need an update.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,144 @@
...
+| **selectionType** | `'single' \| 'multiple'`                                         | `'single'`  | Defines the selection behavior within the ChipGroup component, simulating radio/checkbox like behavior within the group        | ❌       |
+| **size**          | `'xsmall' \| 'small' \| 'medium' \| 'large'`                     | `'small'`   | Specifies the size of the rendered Chips withing the ChipGroup                                                                 | ❌       |
+| **value**         | `string \|string[]`                                              | `undefined` | Acts as a controlled component by specifying the ChipGroup value                                                               | ❌       |
+| **intent**        | `'positive' \| 'negative' \| 'neutral'`                          | `'neutral'` | Sets the ChipGroups's visual variant, it will propagate down to all the Chips                                                  | ❌       |
```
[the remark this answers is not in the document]
snitin315: No, all intent and size should be the same in a chip group. IMO `Yes/No` should be buttons.

packages/blade/src/components/Chip/_decisions/decisions.md
kamleshchandnani: what is the difference between passing it as prop `textTransform: upperCase` vs doing `text.toUpperCase()` inside children?

cseas: For styling changes we usually go for CSS, doing `string.toUpperCase` is generally not the first solution that comes to mind. In my experience, it's common/expected to get review comments on PRs as well for "if we can do it with CSS, why use JS". It's fine if we want to keep it this way but it'd be nice to have it documented as an example since the API isn't intuitive for the use-case IMO.

They also [behave slightly differently][1] but that's probably an edge case, not the main concern here.

[1]: https://stackoverflow.com/a/53731879/7435656

kamleshchandnani (dismissed):

anshulsahni (commented):
Added one question for @saurav12

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,152 @@
...
+| **name**               | `string`                                                         | `undefined` | The name of the chip group, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name) | ❌       |
+| **onChange**           | `({ name, values }: { name: string; values: string[] }) => void` | `undefined` | Calls a function on any state change within the ChipGroup                                                                      | ❌       |
+| **selectionType**      | `'single' \| 'multiple'`                                         | `'single'`  | Defines the selection behavior within the ChipGroup component, simulating radio/checkbox like behavior within the group        | ❌       |
+| **size**               | `'xsmall' \| 'small' \| 'medium' \| 'large'`                     | `'small'`   | Specifies the size of the rendered Chips withing the ChipGroup                                                                 | ❌       |
```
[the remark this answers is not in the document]
saurav12: @anshulsahni This time we've considered a large size as well that can work for landing pages. Also, got it reviewed by advocates including Acquisition team.

Being said that, it might not solve the use-case for assisted discovery because that is still a custom component and cannot be made generic.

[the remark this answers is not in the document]
anshulsahni: @saurav12 One doubt, the large variant is present in other components, I don't see an variant larger than that, any thing different with the larger variant of Chip component vs. other components?

packages/blade/src/components/Chip/_decisions/decisions.md
saurabhdaware: True but problem is variant isn't scalable either because variant has a contexual meaning based on components.

E.g. if we have negative button, we might have both. 

```jsx
<Button variant="secondary" intent="negative" />
```

So not sure if we can use variant as default either 🤔

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Design
+
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=52377%3A27686&mode=design&t=pFzCay0x4a6sTcoQ-1) to all variants of the `Chip` component.
+
+## `Chip` API
+
+> **Note**
+>
+> The `Chip` component is inherently tied to the `ChipGroup` and cannot be utilized outside its context.
```
nashcheez, line 26: We should explain a bit on why this decision was taken for everyone's visibility.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,152 @@
...
+| **value (html native)** | `string`                                | `undefined` | The value of the input field in Chip, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value) | ❌       |
+
+## `ChipGroup` API
+
+| Prop                   | Type                                                             | Default     | Description                                                                                                                    | Required |
+| ---------------------- | ---------------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------ | -------- |
+| **accessibilityLabel** | `string`                                                         | `undefined` | Accessibility label for the ChipGroup                                                                                          | ✅       |
```
nashcheez, line 40: Mention that the required behaviour of `accessibilityLabel` is intentional until we introduce a Label component of varying sizes for all our inout elements - this will be in the pipeline.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,152 @@
...
+| **selectionType**      | `'single' \| 'multiple'`                                         | `'single'`  | Defines the selection behavior within the ChipGroup component, simulating radio/checkbox like behavior within the group        | ❌       |
+| **size**               | `'xsmall' \| 'small' \| 'medium' \| 'large'`                     | `'small'`   | Specifies the size of the rendered Chips withing the ChipGroup                                                                 | ❌       |
+| **value**              | `string \|string[]`                                              | `undefined` | Acts as a controlled component by specifying the ChipGroup value                                                               | ❌       |
+| **intent**             | `'positive' \| 'negative' \| 'neutral'`                          | `'neutral'` | Sets the ChipGroups's visual variant, it will propagate down to all the Chips                                                  | ❌       |
+
+## Examples
```
nashcheez, line 51: Mention an example where the intent of a Chip can override the intent of the ChipGroup. Take the use case of the Yes/No feedback selection. This example should also explain why intent was required on both Chip and ChipGroup.

snitin315: Added.

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: Added.

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: updated

packages/blade/src/components/Chip/_decisions/decisions.md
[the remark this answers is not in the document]
snitin315: Yes, updated the API to allow `intent` on individual chips.

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: For Ally, we have added an `accessibiilityLabel` prop. We are solving labels of various sizes probably by introducing a separate `Label` component for all input components. This development is currently in the pipeline.

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: I feel the controlled/uncontrolled usage examples should suffice.

divyanshu013: I meant from perspective of a form specifically so we can see in action what would it take for consumers to build forms this way. For eg. form values, matching the label styling, spacing et al.

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,196 @@
...
+| **children**            | `React.ReactText \| React.ReactText[]` | `undefined` | Renders the label text for the Chip                                                                                                                | ✅       |
+| **icon**                | `Icon`                                 | `undefined` | Displays the Blade Icon component within the Chip                                                                                                  | ❌       |
+| **isDisabled**          | `boolean`                              | `false`     | Disables or enables the Chip                                                                                                                       | ❌       |
+| **intent**              | `'positive' \| 'negative' \| 'none'`   | `'none'`    | Sets the ChipGroups's visual variant, it will propagate down to all the Chips                                                                      | ❌       |
```
pcprashant, line 33: Is the description correct ?

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,196 @@
...
+>
+> The `Chip` component is inherently tied to the `ChipGroup` and cannot be utilized outside its context. This is because the chips are designed to be used in multiples since they represent collections of objects.
+
+| Prop                    | Type                                   | Default     | Description                                                                                                                                        | Required |
+| ----------------------- | -------------------------------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| **children**            | `React.ReactText \| React.ReactText[]` | `undefined` | Renders the label text for the Chip                                                                                                                | ✅       |
+| **icon**                | `Icon`                                 | `undefined` | Displays the Blade Icon component within the Chip                                                                                                  | ❌       |
```
pcprashant, line 31: we may require avatars to serve as the leading icons in some instances // @kamleshchandnani 
Screenshot 2023-08-28 at 4 29 38 PM

packages/blade/src/components/Chip/_decisions/decisions.md
```diff
@@ -0,0 +1,196 @@
...
+
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=52377%3A27686&mode=design&t=pFzCay0x4a6sTcoQ-1) to all variants of the `Chip` component.
+
+## `Chip` API
+
+> **Note**
+>
+> The `Chip` component is inherently tied to the `ChipGroup` and cannot be utilized outside its context. This is because the chips are designed to be used in multiples since they represent collections of objects.
+
+| Prop                    | Type                                   | Default     | Description                                                                                                                                        | Required |
```
pcprashant, line 28: Can a chip be dismissible? Additionally, what distinguishes a tag from a chip at its core?

pcprashant (changes requested):

packages/blade/src/components/Chip/_decisions/decisions.md
pcprashant: If we treat the chip as a form element, it should come with a label. Should we introduce an 'accessibilityLabel' as a distinct prop, instead of using a 'label' prop? Once the new Label component is finalised, we should update the chip component accordingly.

packages/blade/src/components/Chip/_decisions/decisions.md
kamleshchandnani: Blade advocates didn't raise this. so we'll add this as enhancement once we have the use case. Moreover, we don't even have an avatar component

packages/blade/src/components/Chip/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: @pcprashant I think you shall start getting involved at the blade advocates level call. ye sb waha decisions le liye the why filled vs why not. there's a big research in the doc attached on the channel 😄 

TL;DR that was too close to a primary button making UX non-differentiating between what is a button vs a chip

packages/blade/src/components/Chip/_decisions/decisions.md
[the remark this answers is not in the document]
saurav12: @anshulsahni We have aligned the larger variant of the `Chip` component to the `Button` component that we have in the DS. We've got feedback that this can be used as an actionable entity on Desktop devices (for ex. It can be on the onboarding screen, assisted discovery, and partnership onboarding) and hence it should be similar to actionable components (i.e. buttons).

packages/blade/src/components/Chip/_decisions/decisions.md
kamleshchandnani: @snitin315 can you add the differences from the design doc here?

snitin315: I'll add it in the doc too.

Screenshot 2023-08-28 at 5 41 04 PM

packages/blade/src/components/Chip/_decisions/decisions.md
snitin315: We have already added an `accessibilityLabel` prop, as the name suggests it's for A11y purposes only and will be visually hidden in the UI.

packages/blade/src/components/Chip/_decisions/decisions.md
kamleshchandnani: @snitin315 can you add this to tags and badges API doc as well?

snitin315: Will add in separate PR.

packages/blade/src/components/Chip/_decisions/decisions.md
pcprashant: @snitin315 what is the reason behind not keeping the Label prop ?

PR #1538, opened 2023-08-16
feat: add new `Chip` and `ChipGroup` components

snitin315:
Introducing the New `Chip` and `ChipGroup` Components! 🚀

- API - https://github.com/razorpay/blade/pull/1535
- Tests - https://github.com/razorpay/blade/pull/1547

### API TODO

- [x] Change `neutral` to `none`.

### Storybook TODO

- [x] use contextual icons right now its all info icon
- [x] Redundant stories:
  - [x] Default story which is similar to “Single Selection”. probably name the default as single selection
  - [x]  Probably Rename Default Multi Select and Default Single Select as Unctrolled with Default Value because that’s what it does
  - [x] the controlled selection shall show a better real life example basically show a dropdown whose value change selects the chip
- [x] For intents take a contextual example of “Yes/No”
- [x] defaultValue and value controls shall only be visible when they can do something meaningful right now they are not usable on all stories
- [x] BUG when i’m on default story and click multiple it breaks
- [x] remove children prop from the prop table since its of no use to consumers
- [x] the size prop is sorted alphabetically, we shall sort it in the incremental order xsmall -> small -> medium -> large
- [x] intent doesn’t work on default story
- [x] how are we showing chipghroup and chips props individually? right now i see we only show chipgroup’s prop so consumers will be confused on what all things are supported by chip and how do they play with it? esp for eg: intent of individual chips can be changed
- [x]  add a story for showing .toUpperCase()

### Design TODO

- [x] there are some a11y color contrast issues
- [x] When we select a chip the background color is still there even if i hover out of the chip. the hover shall not be tied to clicked state. it shall reset once you hover out?
- [x] the hover color is too dark that it almost blends with the text color.
- [x] why is hover color of all the chips is gray regardless of the intent? for eg: in dropdown when we hve destructive button the hover color is negative feedback token color


packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,237 @@
...
+  const chipTextColor =
+    chipColorTokens.text[_isDisabled ? 'disabled' : _isChecked && _intent ? _intent : 'unchecked'];
+  const chipBackgroundColor =
+    chipColorTokens.background[_isChecked && _intent ? _intent : 'unchecked'][
+      _isDisabled ? 'disabled' : 'default'
+    ];
+  const chipBorderColor =
+    chipColorTokens.border[_isChecked && _intent ? _intent : 'unchecked'][
+      _isDisabled ? 'disabled' : 'default'
+    ];
```
anuraghazra, line 155: phew, quite a lot of ternaries. 1 level deep ternaries are fine but this is a bit hard to read. 

Can we simplify? 

```jsx
  let textVariant = 'unchecked';
  if (_isChecked && _intent) {
    textVariant = _intent;
  }
  if (_isDisabled) {
    textVariant = 'disabled';
  }
  const chipTextColor = chipColorTokens.text[textVariant];

  let intentVariant = 'unchecked';
  const stateVariant = _isDisabled ? 'disabled' : 'default';
  if (_isChecked && _intent) {
    intentVariant = _intent;
  }
  const chipBackgroundColor = chipColorTokens.background[intentVariant][stateVariant];
  const chipBorderColor = chipColorTokens.border[intentVariant][stateVariant];
```

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,237 @@
...
+          <BaseBox display="flex" alignItems="center" flexDirection="row">
+            <SelectorInput
+              hoverTokens={getChipHoverTokens(_intent)}
+              isChecked={state.isChecked}
+              isDisabled={_isDisabled}
+              inputProps={inputProps}
+              ref={ref}
+            />
+            <AnimatedChip
+              borderColor={chipBorderColor as never}
```
anuraghazra, line 185: why as never here?

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,237 @@
...
+              isPressed={isPressed}
+              isDesktop={matchedDeviceType === 'desktop'}
+            >
+              <BaseBox
+                display="flex"
+                flexDirection="row"
+                justifyContent="center"
+                alignItems="center"
+                overflow="hidden"
+                backgroundColor={chipBackgroundColor as never}
```
anuraghazra, line 196: why as never here?

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,237 @@
...
+                borderRadius="max"
+                borderWidth={['xsmall', 'small'].includes(_size) ? 'thinner' : 'thin'}
+                paddingLeft={
+                  chipHorizontalPaddingTokens[Boolean(Icon) ? 'icon' : 'default'].left[_size]
+                }
+                paddingRight={
+                  chipHorizontalPaddingTokens[Boolean(Icon) ? 'icon' : 'default'].right[_size]
+                }
+                style={{
+                  height: chipHeightTokens[_size],
```
anuraghazra, line 206: you can put height as a prop. 

```
height={`${chipHeightTokens[_size]}px`}
```

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,237 @@
...
+                  borderColor: _isChecked ? 'inherit' : 'transparent',
+                }}
+              >
+                {Icon ? (
+                  <BaseBox paddingRight="spacing.3" display="flex">
+                    <Icon color={chipTextColor as never} size={chipIconSizes[_size]} />
```
anuraghazra, line 212: same here. 

If the TS is throwing error that means we need to change that `ColorTokens` type in chipTokens. 

I seems like currently it's way to big, it includes literally all the color tokens: 
image

packages/blade/src/components/Chip/chipTokens.ts
```diff
@@ -0,0 +1,199 @@
...
+  },
+  large: {
+    variant: 'body',
+    size: 'large',
+  },
+} as const;
+
+const chipMotionTokens = {
+  duration: 'motion.duration.xquick',
+  timingFunction: 'motion.easing.standard.effective',
```
anuraghazra, line 187:
```suggestion
  easing: 'motion.easing.standard.effective',
```

packages/blade/src/components/Chip/AnimatedChip.native.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const chipAnimation = useAnimatedStyle(() => {
+    return {
+      transform: [
+        {
+          scale: scaleDownAnimation.value,
+        },
+      ],
+    };
+  }, []);
```
anuraghazra, line 44:
```jsx
  const chipAnimation = useAnimatedStyle(() => {
    return {
      transform: [
        {
          scale: withTiming(isPressed ? 0.92 : 1, {
            duration,
            easing,
          }),
        },
      ],
    };
  }, [isPressed]);
```

Just this would work, you can remove the useEffect stuff

anuraghazra (changes requested):
Added few comments.

packages/blade/src/components/Chip/Chip.tsx
snitin315: Nice, I didn't know about this util.

packages/blade/src/components/Chip/Chip.tsx
snitin315: Fixed types

packages/blade/src/components/Chip/Chip.tsx
snitin315: Fixed types

packages/blade/src/components/Chip/Chip.tsx
snitin315: Yes, updated 👍🏻

packages/blade/src/components/Chip/Chip.tsx
snitin315: Fixed types

snitin315: @anuraghazra I updated it to use `IconColor` type, but it doesn't accept `brand.primary.500` yet used, should I add it to the Icon props?

snitin315: I just confirmed with Abhishek, that these should be text color only. Should I update the type of IconProps?

snitin315: @anuraghazra We can do it in a non-breaking way `IconColorType = currentType | TextProps['color']`

kamleshchandnani: we already have this type supported right? `TextIconColors` for icons?

kamleshchandnani:
@snitin315 i did a master back merge because of chromatic on this branch

packages/blade/src/components/Chip/Chip.tsx
snitin315: @kamleshchandnani TextIconColors are `surface.text.*` tokens. We need `brand.primary.500` and `feedback.text.<intent>.lowContrast`

snitin315:
The build is failing, I'll fix it.

packages/blade/src/components/Chip/chipTokens.ts
```diff
@@ -0,0 +1,209 @@
...
+
+const chipHeightTokens: Record<NonNullable<ChipGroupProps['size']>, number> = {
+  xsmall: size[24],
+  small: size[28],
+  medium: size[36],
+  large: size[48],
+};
+
+type ChipHorizontalPaddingTokens = {
+  default: Record<
```
kamleshchandnani, line 36: maybe rename it to `withoutIcon` and `withIcon`? right now its not very clearly understandable what does default mean in this context

packages/blade/src/components/Chip/chipTokens.ts
```diff
@@ -0,0 +1,209 @@
...
+};
+
+type ColorTokens = DotNotationColorStringToken<Theme['colors']>;
+type TextColorTokens = BaseTextProps['color'];
+type IconColorTokens = IconProps['color'];
+type ChipColorTokens = {
+  text: Record<string, TextColorTokens>;
+  icon: Record<string, IconColorTokens | 'brand.primary.500'>;
+  background: Record<string, Record<string, ColorTokens | 'transparent'>>;
+  border: Record<string, Record<string, ColorTokens>>;
```
kamleshchandnani, line 84: same for `border`

packages/blade/src/components/Chip/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './Chip';
```
kamleshchandnani, line 1: why export everything? export only required things? like type and the component? otherwise all internal components will be exported.

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,250 @@
...
+} from './chipTokens';
+import { AnimatedChip } from './AnimatedChip';
+import type { IconComponent } from '~components/Icons';
+import { metaAttribute, MetaConstants } from '~utils/metaAttribute';
+import { getStyledProps } from '~components/Box/styledProps';
+import type { StyledPropsBlade } from '~components/Box/styledProps';
+import BaseBox from '~components/Box/BaseBox';
+import { SelectorLabel } from '~components/Form/Selector/SelectorLabel';
+import { SelectorInput } from '~components/Form/Selector/SelectorInput';
+import type { BladeElementRef } from '~utils/useBladeInnerRef';
```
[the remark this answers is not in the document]
snitin315: Fixed.

packages/blade/src/components/Chip/ChipGroup/ChipGroup.tsx
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
```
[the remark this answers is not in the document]
snitin315: Yes, we can refactor this.

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,250 @@
...
+  isChecked,
+  event,
+  value,
+}: {
+  isChecked: boolean;
+  event?: React.ChangeEvent;
+  value?: string;
+}) => void;
+
+type ChipProps = {
```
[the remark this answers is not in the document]
snitin315: Moved.

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,250 @@
...
+   * Displays the Blade Icon component within the Chip
+   * Accepts a component of type `IconComponent` from Blade.
+   *
+   */
+  icon?: IconComponent;
+  /**
+   * Sets the Chip's visual intent
+   *
+   */
+  intent?: 'positive' | 'negative' | 'neutral';
```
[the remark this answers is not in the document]
snitin315: Updated it to `none`.

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+  const _isDisabled = isDisabled ?? groupProps?.isDisabled;
+  const _name = groupProps?.name;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+  const _isChecked = groupProps?.state?.isChecked(value!);
+  // Check if the defaultValue property of groupProps is undefined
+  const defaultChecked =
+    typeof groupProps?.defaultValue === 'undefined'
+      ? undefined // If undefined, defaultChecked is also undefined
+      : groupProps?.defaultValue?.includes(value as string); // If multiple selection, check if value is in defaultValue array
```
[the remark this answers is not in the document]
snitin315: We make sure it's an array internally. We convert it to `[providedValue]`. 
https://github.com/razorpay/blade/blob/7f6ae07fbb95a25b2fec8f4688b1512a81937c82/packages/blade/src/components/Chip/useChipGroup.ts#L39-L43

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,250 @@
...
+        moduleName: 'Chip',
+        message:
+          '<Chip /> component should only be used within the context of a <ChipGroup /> component',
+      });
+    }
+  }
+
+  const _isDisabled = isDisabled ?? groupProps?.isDisabled;
+  const _name = groupProps?.name;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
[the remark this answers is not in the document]
snitin315: I think there is a bug in this rule because `value!` is actually converting the type from `string | undefined` to `string` here. I can see [multiple instances of this rule being disabled.](https://github.com/search?q=repo%3Arazorpay%2Fblade%20%40typescript-eslint%2Fno-unnecessary-type-assertion&type=code)

packages/blade/src/components/Chip/index.ts
snitin315: Currently, we are exporting only Chip and ChipProps here. but I do see the point here. We should update all components to export only the required types & components. Maybe we can add an eslint rule for this. Currently, there are many components with `export * from './component';` exports.

packages/blade/src/components/Chip/chipTokens.ts
snitin315: Updated.

packages/blade/src/components/Chip/chipTokens.ts
```diff
@@ -0,0 +1,209 @@
...
+  },
+};
+
+type ColorTokens = DotNotationColorStringToken<Theme['colors']>;
+type TextColorTokens = BaseTextProps['color'];
+type IconColorTokens = IconProps['color'];
+type ChipColorTokens = {
+  text: Record<string, TextColorTokens>;
+  icon: Record<string, IconColorTokens | 'brand.primary.500'>;
+  background: Record<string, Record<string, ColorTokens | 'transparent'>>;
```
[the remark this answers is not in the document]
snitin315: Updated types

packages/blade/src/components/Chip/chipTokens.ts
snitin315: Updated types.

packages/blade/src/components/Chip/AnimatedChip.web.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { chipMotionTokens } from './chipTokens';
+import BaseBox from '~components/Box/BaseBox';
+import { makeMotionTime } from '~utils/makeMotionTime';
+import { castWebType } from '~utils';
+
+const AnimatedChip = styled(BaseBox)<AnimatedChipProps>((props) => {
+  const easing = getIn(props.theme, chipMotionTokens.easing);
+  const duration = castWebType(makeMotionTime(getIn(props.theme, chipMotionTokens.duration)));
+  return {
+    ...getAnimatedChipStyles(props),
```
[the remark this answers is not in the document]
snitin315: I think it's fine, this is used inside `Chip` only.

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const _isChecked = groupProps?.state?.isChecked(value!);
+  // Check if the defaultValue property of groupProps is undefined
+  const defaultChecked =
+    typeof groupProps?.defaultValue === 'undefined'
+      ? undefined // If undefined, defaultChecked is also undefined
+      : groupProps?.defaultValue?.includes(value as string); // If multiple selection, check if value is in defaultValue array
+  const useChip = groupProps?.selectionType === 'single' ? useRadio : useCheckbox;
```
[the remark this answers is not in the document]
snitin315: That's because `role` and other props are optional on useCheckbox.

kamleshchandnani:
we shouldn't disable the option itself if its breaking. lets fix it? for `default` story we need as many knobs possible because its a playground. 

Screenshot 2023-08-28 at 10 42 21 AM

snitin315:
>we shouldn't disable the option itself if its breaking. lets fix it? for default story we need as many knobs possible because its a playground.

@kamleshchandnani I think it might be a bug in the storybook itself because it works in docs view but not story view. Furthermore I feel disabling makes sense because `value` and `defaultValue`  props value would be `select`/`multi-select` based on the `selectionType` which could unintentionally break the story when playing around. Hence we have separate story for Single & multi selection.

kamleshchandnani:
>When we select a chip the background color is still there even if i hover out of the chip. the hover shall not be tied to clicked state. it shall reset once you hover out?

@snitin315 this says its done? what was the conclusion?

kamleshchandnani:
> > we shouldn't disable the option itself if its breaking. lets fix it? for default story we need as many knobs possible because its a playground.
> 
> @kamleshchandnani I think it might be a bug in the storybook itself because it works in docs view but not story view. Furthermore I feel disabling makes sense because `value` and `defaultValue` props value would be `select`/`multi-select` based on the `selectionType` which could unintentionally break the story when playing around. Hence we have separate story for Single & multi selection.

oh this is why its breaking. as a workaround what you can do is based on the toggle of selectionType you can change passing the default value from single string to array of string and issue might be fixed?

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,564 @@
...
+        {...args}
+      >
+        <ChipComponent ref={chipRef} value="Proprietorship">
+          Proprietorship
+        </ChipComponent>
+        <ChipComponent value="Public">Public</ChipComponent>
+        <ChipComponent value="Small Business">Small Business</ChipComponent>
+      </ChipGroupComponent>
+      <Box maxWidth="200px">
+        <Button isFullWidth={false} onClick={() => chipRef?.current?.focus()}>
```
kamleshchandnani, line 548: shouldn't the chip be selected on focus when the selection type is single since its radio?

Also, can we add a button to reset the selection in this same story so that way we'll cover all the cases of refs and use cases too

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,564 @@
...
+          Click to focus the Chip
+        </Button>
+      </Box>
+    </Box>
+  );
+};
+
```
kamleshchandnani, line 555: lets add Chip only story?

maybe nest it like Checkbox

`Chip & ChipGroup/Chipgroup/all stories of chipgroup` and `Chip & ChipGroup/Chipgroup/all stories of chip`

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,564 @@
...
+export const ChipWithIcon = ChipWithIconTemplate.bind({});
+ChipWithIcon.storyName = 'With Icon';
+ChipWithIcon.args = {
+  accessibilityLabel: 'Quick Filters:',
+};
+
+const AllChipSizesTemplate: ComponentStory<typeof ChipGroupComponent> = ({ children, ...args }) => {
```
kamleshchandnani, line 421: why do we need this? this seems kitchen sink in the middle of stories. we anyway have knobs which are working fine. if you feel we shall only have sizes maybe? but sizes and with/without icons is too much in the centre of the doc

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,564 @@
...
+        ))}
+      </ChipGroupComponent>
+    </Box>
+  );
+};
+
+export const MultiSelection = MultiSelectChipTemplate.bind({});
+MultiSelection.storyName = 'Multi Selection';
+MultiSelection.args = {
+  accessibilityLabel: 'Quick Filters:',
```
kamleshchandnani, line 248: lets use previous example of Business type itself. Quick filters isn't fitting well here. what you can do is put a dropdown with mutliselect option which will also select the multiple chips

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,564 @@
...
+          <ChipComponent key={chipValue} value={chipValue} icon={args.icon}>
+            {chipValue}
+          </ChipComponent>
+        ))}
+      </ChipGroupComponent>
+    </Box>
+  );
+};
+
+export const Default = ChipTemplate.bind({});
```
kamleshchandnani, line 205:
```suggestion
export const SingleSelection = ChipTemplate.bind({});
```
??

packages/blade/src/components/Chip/ChipGroup/ChipGroup.tsx
```diff
@@ -0,0 +1,88 @@
...
+      throwBladeError({
+        moduleName: 'ChipGroup',
+        message: `When "selectionType" is "single", the "value" prop must be a string value, but an array was provided.`,
+      });
+    }
+  }
+
+  return (
+    <ChipGroupProvider value={contextValue}>
+      <BaseBox {...getStyledProps(styledProps)}>
```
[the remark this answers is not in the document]
snitin315: We define it on `SelectorGroupField`, 

https://github.com/razorpay/blade/blob/88e68ec1e7b674163e44fd7208ac819e5c5c8fec/packages/blade/src/components/Chip/ChipGroup.tsx#L60

https://github.com/razorpay/blade/blob/88e68ec1e7b674163e44fd7208ac819e5c5c8fec/packages/blade/src/components/Form/Selector/SelectorGroupField.tsx#L39

packages/blade/src/components/Chip/Chip.stories.tsx
snitin315: @kamleshchandnani Business type is actually `singular` in nature, you can either be Public or Proprietor.

packages/blade/src/components/Chip/Chip.stories.tsx
snitin315: Yes, I'll update

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,564 @@
...
+      </Text>
+
+      <ChipGroupComponent defaultValue="Proprietorship" {...args}>
+        {chipValues.map((chipValue: string) => (
+          <ChipComponent key={chipValue} value={chipValue}>
+            {chipValue.toUpperCase()}
+          </ChipComponent>
+        ))}
+      </ChipGroupComponent>
+    </Box>
```
kamleshchandnani, line 523: lets put a code component here with an example stating that you can pass `value.toUpperCase()` to the children to make it uppercase because on storybook they parse the JS so they can't really identify how to make this work

Screenshot 2023-08-28 at 1 15 08 PM

packages/blade/src/components/Chip/Chip.stories.tsx
kamleshchandnani: so changhe context maybe use what else you're looking wala context

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -544,16 +544,24 @@ export const chipRef: ComponentStory<typeof ChipComponent> = (args) => {
...
             chipRef?.current?.focus();
             chipRef?.current?.click();
           }}
         >
           Click to focus the Chip
         </Button>
+        <Button
+          isFullWidth={true}
+          onClick={() => {
+            chipRef?.current?.blur();
```
kamleshchandnani, line 560: we shall reset the selection and not just remove the focus

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -544,16 +544,24 @@ export const chipRef: ComponentStory<typeof ChipComponent> = (args) => {
...
-          isFullWidth={false}
+          isFullWidth={true}
           onClick={() => {
             chipRef?.current?.focus();
             chipRef?.current?.click();
           }}
         >
           Click to focus the Chip
         </Button>
+        <Button
```
kamleshchandnani, line 557: make this button secondary

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -544,16 +544,24 @@ export const chipRef: ComponentStory<typeof ChipComponent> = (args) => {
...
+      <Box maxWidth="400px" display="flex" flexDirection="row" gap="spacing.3">
         <Button
-          isFullWidth={false}
+          isFullWidth={true}
           onClick={() => {
             chipRef?.current?.focus();
             chipRef?.current?.click();
           }}
         >
           Click to focus the Chip
```
kamleshchandnani, line 555: Select Proprietorship

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -544,16 +544,24 @@ export const chipRef: ComponentStory<typeof ChipComponent> = (args) => {
...
         </Button>
+        <Button
+          isFullWidth={true}
+          onClick={() => {
+            chipRef?.current?.blur();
+          }}
+        >
+          Click to remove focus
```
kamleshchandnani, line 563: Reset Selection

packages/blade/src/components/Chip/Chip.stories.tsx
snitin315: Updated.

packages/blade/src/components/Chip/Chip.stories.tsx
snitin315: Updated.

packages/blade/src/components/Chip/Chip.stories.tsx
snitin315: Updated.

snitin315:
- [X] the default story has selection type disabled. we shall enable it? maybe because you removed default story because of this comment probably name the default as single selection
   - [x] default value doesn’t work/has no effect for  single selection story. We need to probably pass  key prop so the component can re-render on every change(this actually doesn’t work on many stories)
- [x] Icon props not working for Multiselection story
    - [x] the default value and value prop shows just a big box for multi select story it shall have a drop down right to change the defaultvalue? value prop can be disabled in this scenario
- [x] the a11y label shall be something meaningful Select Business type:this is a label we shall have Select one/any business type from below options something that will be read by screen readers so has to be human understandable
- [x] selectionType wherever disabled can we show its value displayed for eg where stories are single the value for selectionType prop shall also display the same?
- [x] Can you add a story for Chip separetely to show its props? check Checkbox stories we have further nesting Chip/ChipGroup and Chip/Chip

// @kamleshchandnani

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,291 @@
...
+    >
+      <Title>Usage</Title>
+      <Sandbox showConsole editorHeight={400} editorWidthPercentage={60}>
+        {`
+          import { Box, Chip, ChipGroup, Text } from '@razorpay/blade/components';
+
+          function App(): React.ReactElement {
+            return (
+              <Box>
+                <Text> Select Business type: </Text>
```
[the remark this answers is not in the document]
kamleshchandnani: apply the same suggestion for all the chip stories

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,291 @@
...
+        <Code size="medium"> value.toUpperCase() </Code> as the children.
+      </Text>
+    </Box>
+  );
+};
+
+export const TextTransformationUppercase = TextTransformationTemplate.bind({});
+TextTransformationUppercase.storyName = 'Text Transformation (Uppercase)';
+
+export const chipRef: ComponentStory<typeof ChipComponent> = (args) => {
```
kamleshchandnani, line 243: chip ref doesn't makes sense for `Chip` component so just keep this only in `ChipGroup`

packages/blade/src/components/Chip/Chip.stories.tsx
snitin315: added

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,291 @@
...
+  ThumbsDownIcon,
+} from '~components/Icons';
+import { Sandbox } from '~utils/storybook/Sandbox';
+import StoryPageWrapper from '~utils/storybook/StoryPageWrapper';
+import type { BladeElementRef } from '~utils/types';
+import { Button } from '~components/Button';
+import { Box } from '~components/Box';
+import { getStyledPropsArgTypes } from '~components/Box/BaseBox/storybookArgTypes';
+import iconMap from '~components/Icons/iconMap';
+
```
kamleshchandnani, line 21: can you sort the story so chipgroup comes first in the navbar?
Screenshot 2023-08-29 at 10 33 11 AM

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,291 @@
...
+const Page = (): React.ReactElement => {
+  return (
```
kamleshchandnani, line 23: remove figma+sandpack example for chip and put an alert which says Chips cannot be used individually and shall only be used with `ChipGroup` and provide a link to the ChipGroup Stories. Also, add the description that this story is only meant for demonstrating props of chip component.

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,291 @@
...
+      mapping: iconMap,
+      table: {
+        category: propsCategory.CHIP,
+        type: {
+          summary: 'IconComponent',
+        },
+      },
+    },
+    intent: {
+      description: 'Sets the intent of the Chip',
```
kamleshchandnani, line 106: the intent is not re-rendering the chip. this happens in docs mode itself

packages/blade/src/components/Chip/Chip.stories.tsx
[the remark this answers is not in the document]
kamleshchandnani: you can actually remove all the stories of chip, just keep the default one to play around with the props

packages/blade/src/components/Chip/ChipGroup.stories.tsx
```diff
@@ -0,0 +1,463 @@
...
+      componentDescription="Chips represents a collection of selectable objects which enable users to make selections, filter content, and trigger relevant actions. Chips can have either single selection or multiple (based on context)."
+      componentName="ChipGroup"
+      imports={`import { Chip, ChipGroup } from '@razorpay/blade/components';\nimport type { ChipProps, ChipGroupProps } from '@razorpay/blade/components';`}
+      figmaURL={{
+        paymentTheme:
+          'https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=52377%3A23885&mode=design&t=y7gUIBIzzNMRd3w6-1',
+        bankingTheme:
+          'https://www.figma.com/file/sAdplk2uYnI2ILnDKUxycW/Blade---Banking-Dark?type=design&node-id=18358%3A3135&mode=design&t=FzNrQV6ZZaLoxzcj-1',
+      }}
+    >
```
kamleshchandnani, line 28: put an alert above the Props table mentioning that these are just ChipGroup props and link to chip stories for props of chips

packages/blade/src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,291 @@
...
+export default {
+  title: 'Components/Chip/Chip',
+  args: {
+    isDisabled: false,
+  },
+  argTypes: {
+    // ChipGroup props
+    isDisabled: {
+      description:
+        'Disables or enables `ChipGroup`, it will propagate down to all the children `Chip` components.',
```
kamleshchandnani, line 71:
```suggestion
        'Disables or enables `Chip`
```

packages/blade/src/components/Chip/Chip.stories.tsx
kamleshchandnani, line 106:
```suggestion
      description: 'Sets the intent of the Chip. This overwrites the intent set by the parent `ChipGroup` component',
```

packages/blade/src/components/Chip/ChipGroup.stories.tsx
```diff
@@ -0,0 +1,463 @@
...
+    'Wallet on My App',
+    'Offer discounts, Pay Later & EMI options',
+  ];
+  const [values, setValues] = React.useState(['Automated Payment Links']);
+  return (
+    <Box display="flex" gap="spacing.5" flexDirection="column" minHeight="200px">
+      <Dropdown marginRight="spacing.4">
+        <DropdownButton size="small">What other capabilities are you looking for?</DropdownButton>
+        <DropdownOverlay>
+          <ActionList>
```
kamleshchandnani, line 370: this shall have selection type multiple right?  it looks like this right now. @saurabhdaware is this a valid multiselect?
Screenshot 2023-08-29 at 11 04 48 AM

snitin315: Fixed

packages/blade/src/components/Chip/Chip.stories.tsx
snitin315: Checking the docs https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -173,7 +186,7 @@ const _Chip: React.ForwardRefRenderFunction<BladeElementRef, ChipProps> = (
                 justifyContent="center"
                 alignItems="center"
                 overflow="hidden"
-                backgroundColor={chipBackgroundColor}
+                backgroundColor={isHovered ? 'transparent' : chipBackgroundColor}
```
kamleshchandnani, line 189: there's a flicker that happens because we change this on hover. 

https://github.com/razorpay/blade/assets/11384858/a625f3d6-061a-48f1-9e92-213937dc4435

snitin315: Fixed, can you check again.

packages/blade/src/components/Chip/Chip.stories.tsx
snitin315: Unable to do it without compromising `Chip` folder position in the sidebar.

kamleshchandnani (approved):

PR #1540, opened 2023-08-16
build: storybook upgrade v7 for web & native

chaitanyadeorukhkar:


chaitanyadeorukhkar:
Rebasing

chaitanyadeorukhkar:
**DO NOT DELETE BRANCH**. Closing this yet again. 2 things are broken:

1. StorybookWrapper doesn't work with mdx stories (Modal, Dropdown). We need to replace usages of all `@storybook/addon-docs` like Title, Subtitle, etc. with our custom components to fix this. This is an easier fix but we might lose out on few things like copy link to exact MDX heading.
2. Bottomsheet story is broken with storybook/react-native v7 alpha 5 upgrade. Confirmed that Bottomsheet component itself is okay but storybook isn't playing well with Grohom bottom sheet package. Maybe its not picking up the right platform files. We need to spend more time to figure this out.

PR #1547, opened 2023-08-21
test: add tests for `Chip` and `ChipGroup` components

snitin315:


packages/blade/src/components/Chip/__tests__/Chip.native.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+        </Chip>
+      </ChipGroup>,
+    );
+    expect(toJSON()).toMatchSnapshot();
+    expect(getByText('Apple')).toBeDefined();
+    expect(getByText('Mango')).toBeDefined();
+  });
+
+  it('should throw error if used outside ChipGroup', () => {
+    try {
```
anuraghazra, line 42: let's use toThrow. 

```
expect(() =>
      renderWithTheme(
        <Alert
          description="Currently you can only accept payments in international currencies using PayPal."
          // @ts-expect-error testing failure case when there is no primary action passed
          actions={{
            secondary: { text: 'Link', onClick: onClickSecondary },
          }}
        />,
      ),
    ).toThrow(`[Blade: Alert]: SecondaryAction is allowed only when PrimaryAction is defined.`);
```

The issue with try catch is that, if the error never happens (eg: say you remove the runtime check) the catch clause will never run thus test will pass.

packages/blade/src/components/Chip/__tests__/Chip.web.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+    );
+    expect(container).toMatchSnapshot();
+    expect(getByRole('radio', { name: 'Apple' })).toBeInTheDocument();
+    expect(getByRole('radio', { name: 'Mango' })).toBeInTheDocument();
+  });
+
+  it('should throw error if used outside ChipGroup', () => {
+    try {
+      renderWithTheme(<Chip value="apple">Apple</Chip>);
+    } catch (error: unknown) {
```
anuraghazra, line 44: same here let's change all the try catches

packages/blade/src/components/Chip/__tests__/Chip.web.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+        </Chip>
+        <Chip value="mango">Mango</Chip>
+      </ChipGroup>,
+    );
+    expect(container).toMatchSnapshot();
+    expect(getByRole('radio', { name: 'Apple' })).toBeDisabled();
+    expect(getByRole('radio', { name: 'Mango' })).not.toBeDisabled();
+  });
+
+  it('user should be able set checked state', async () => {
```
anuraghazra, line 67: let's add test for selectionType single and multiple.

packages/blade/src/components/Chip/__tests__/Chip.web.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+          Apple
+        </Chip>
+        <Chip value="mango" testID="chip-mango-test">
+          Mango
+        </Chip>
+      </ChipGroup>,
+    );
+    expect(getByTestId('chip-apple-test')).toBeTruthy();
+    expect(getByTestId('chip-mango-test')).toBeTruthy();
+  });
```
anuraghazra, line 137: let's add assertAccessible test

```
it('should pass general a11y', async () => {
    const { container } = renderWithTheme();

    await assertAccessible(container);
  });
```

packages/blade/src/components/Chip/__tests__/ChipGroup.native.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+        const chips = getAllByRole(isSingleSelection ? 'radio' : 'checkbox');
+        const apple = chips.find((chip) => chip.props.value === 'apple');
+        const mango = chips.find((chip) => chip.props.value === 'mango');
+        const orange = chips.find((chip) => chip.props.value === 'orange');
+
+        expect(apple?.props.accessibilityState.checked).toBeTruthy();
+        expect(onChange).not.toBeCalled();
+
+        if (isSingleSelection) {
+          fireEvent.press(mango as ReactTestInstance);
```
anuraghazra, line 127: Let's try to keep the tests as pure as possible. 

Code duplication should be fine, let's keep separate test suites for `single select` and `multi select` rather than adding if statements and dynamically generating the test. 

```
it('should work with controlled state when selectionType=single', () => {
  // code for single selection without any external logic
})

it('should work with controlled state when selectionType=multiple', () => {
  // code for multi selection without any external logic
})
```

packages/blade/src/components/Chip/__tests__/__snapshots__/ChipGroup.native.test.tsx.snap
```diff
@@ -0,0 +1,10513 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
anuraghazra, line 1: Oh man, 10k plus snapshots. 

TBH i'm not sure at this point this will be much of use, can we reduce few snapshots? I think it's getting this big because we are snapshotting [every possible](https://github.com/razorpay/blade/pull/1547/files#diff-2ca24370c91077981af15bcac075041c67c52dd367133334f1e6d23e1255731bR8-R10) variant. 

I would say it's better to do visual testing for these than adding this many snapshots

packages/blade/src/components/Chip/__tests__/ChipGroup.native.test.tsx
snitin315: Updated.

packages/blade/src/components/Chip/__tests__/__snapshots__/ChipGroup.native.test.tsx.snap
snitin315: I have removed snaps for all sizes and variants.

packages/blade/src/components/Chip/__tests__/Chip.native.test.tsx
snitin315: Updated, I have seen similar usage of try-catch in a few other tests, we should use` toThrow()/toThrowError()` for consistency. maybe we can add an eslint rule for this.

packages/blade/src/components/Chip/__tests__/Chip.web.test.tsx
snitin315: Added, it's covered in ChipGroup tests too.

packages/blade/src/components/Chip/__tests__/Chip.web.test.tsx
snitin315: Nice, I didn't know about this util.

packages/blade/src/components/Chip/__tests__/Chip.native.test.tsx
anuraghazra: Yes that would be nice, if there is a third party eslint rule for this then let's add, otherwise rehne do.

anuraghazra (dismissed):
LGTM

packages/blade/src/components/Chip/__tests__/ChipGroup.web.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+import type { ChipGroupProps } from '../ChipGroup/ChipGroup';
+import { ChipGroup } from '../ChipGroup/ChipGroup';
+import renderWithTheme from '~utils/testing/renderWithTheme.web';
+
+const selectionTypes: ChipGroupProps['selectionType'][] = ['single', 'multiple'];
+const sizes: ChipGroupProps['size'][] = ['xsmall', 'small', 'medium', 'large'];
+const intents: ChipGroupProps['intent'][] = ['neutral', 'positive', 'negative'];
+
+describe('<ChipGroup />', () => {
+  for (const selectionType of selectionTypes) {
```
kamleshchandnani, line 13: brilliant 👏

packages/blade/src/components/Chip/__tests__/ChipGroup.web.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+        const { container, getByRole } = renderWithTheme(
+          <ChipGroup accessibilityLabel="Select fruits" selectionType={selectionType}>
+            <Chip value="apple">Apple</Chip>
+            <Chip value="mango">Mango</Chip>
+            <Chip value="orange">Orange</Chip>
+          </ChipGroup>,
+        );
+
+        const chipRole = selectionType === 'single' ? 'radio' : 'checkbox';
+        expect(getByRole('group')).toBeInTheDocument();
```
kamleshchandnani, line 25: will the roll be a group for a radio as well?

packages/blade/src/components/Chip/__tests__/ChipGroup.web.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+import { ChipGroup } from '../ChipGroup/ChipGroup';
+import renderWithTheme from '~utils/testing/renderWithTheme.web';
+
+const selectionTypes: ChipGroupProps['selectionType'][] = ['single', 'multiple'];
+const sizes: ChipGroupProps['size'][] = ['xsmall', 'small', 'medium', 'large'];
+const intents: ChipGroupProps['intent'][] = ['neutral', 'positive', 'negative'];
+
+describe('<ChipGroup />', () => {
+  for (const selectionType of selectionTypes) {
+    describe(`selectionType="${selectionType}"`, () => {
```
kamleshchandnani, line 14: why do you need a describe here? you already have it for multiple behavior already?

packages/blade/src/components/Chip/__tests__/ChipGroup.web.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+    expect(fn).toBeCalledWith(['mango']);
+    await user.keyboard('[ArrowDown]');
+    expect(getByLabelText('Orange')).toHaveFocus();
+    expect(fn).toBeCalledWith(['orange']);
+    await user.keyboard('[ArrowDown]');
+    expect(getByLabelText('Apple')).toHaveFocus();
+    expect(fn).toBeCalledWith(['apple']);
+  });
+
+  it('should work in uncontrolled mode when selectionType="single"', async () => {
```
kamleshchandnani, line 148: typo selectiontype multiple?

packages/blade/src/components/Chip/__tests__/ChipGroup.web.test.tsx
snitin315: This is to provide a better output on the terminal.
Screenshot 2023-08-28 at 8 20 48 AM

packages/blade/src/components/Chip/__tests__/ChipGroup.web.test.tsx
snitin315: yes, good catch.

packages/blade/src/components/Chip/__tests__/ChipGroup.web.test.tsx
snitin315: Thanks!

packages/blade/src/components/Chip/__tests__/ChipGroup.web.test.tsx
snitin315: @anuraghazra Can you confirm? Currently, I see we have implemented `group` in radio web and `radiogroup ` on RN
https://github.com/razorpay/blade/blob/f13a18d51cfd1b013fa7d573a724d4dba4eee000/packages/blade/src/components/Radio/RadioGroup/RadioGroup.tsx#L139

kamleshchandnani (approved):
overall LGTM except that one role thing which is open

packages/blade/src/components/Chip/__tests__/ChipGroup.web.test.tsx
anuraghazra: Actually in web & native both should be `radiogroup` only, i'm not sure why i put `group` on web for RadioGroup seems like a mistake.

snitin315: I'll update it for Chip. For radio I'll open a separate PR.

kamleshchandnani (approved):

kamleshchandnani:
@snitin315 sorry i merged the main branch 🙈 can you rebase this branch now?

snitin315:
Rebasing.

Issue #1539, opened 2023-08-16
Accordion occupies the full width when screen size > 1024px


sumitd94:
When we have the screen size greater than 1024px, we have the min-width set to 400px in the accordion & it occupies the full width in the modal

<img width="450" alt="Screenshot 2023-08-16 at 3 24 00 PM" src="https://github.com/razorpay/blade/assets/29303618/a048e76b-32e7-4946-ba0b-750a1377d264">

Looks like we will also have to increase the max-width of the entire modal so that the left and right margin get's aligned correctly (of the accordion).

iamabhay17:
hey, I would like to work on in this issue

chaitanyadeorukhkar:
Seems okay to me? Can you help me with a reproducable example?
<img width="1509" alt="Screenshot 2023-08-22 at 12 07 26 PM" src="https://github.com/razorpay/blade/assets/24487274/5e761799-6190-4d38-a55f-2c49b47cecf3">

PR #1552, opened 2023-08-22
chore: make accordion min-width 360px for large screens

chaitanyadeorukhkar:
Fixes: #1539


packages/blade/src/components/Accordion/Accordion.tsx
```diff
@@ -46,7 +46,7 @@ type AccordionProps = {
 const MIN_WIDTH: BoxProps['minWidth'] = {
   s: makeSize(size[200]),
   m: makeSize(size[360]),
-  l: makeSize(size[400]),
+  l: makeSize(size[360]),
```
anuraghazra, line 49: is 360 also the same min width for modal?

anuraghazra (approved):

PR #1558, opened 2023-08-24
refactor(blade): remove useBladeInnerRef

### LINKED ISSUE #1305: refactor: remove useBladeInnerRef
When working on Tooltip we had to expose the whole DOM node to make FloatingUI work properly, thus in few components we removed the useBladeInnerRef abstraction. 

I think it will be better to just remove it altogether at this point. 

Reference: https://github.com/razorpay/blade/pull/1265/files#r1227922177

anuraghazra:
fixes #1305


packages/blade/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -139,7 +138,7 @@ export const ControlledAndUncontrolled = _ControlledAndUncontrolled.bind({});
 
 export const checkboxRef: ComponentStory<typeof CheckboxComponent> = () => {
   // eslint-disable-next-line react-hooks/rules-of-hooks
-  const checkboxRef = React.useRef<BladeElementRef>(null);
+  const checkboxRef = React.useRef<HTMLInputElement>(null);
```
chaitanyadeorukhkar, line 141: Is this not breaking on react native?

chaitanyadeorukhkar (dismissed):
Check on react native once

packages/blade/src/components/Checkbox/Checkbox.stories.tsx
anuraghazra: No won't be a breaking change, underline types are kept wide in purpose so that it doesn't throw any type errors. 

Plus since we removed the whole useBladeInnerRef thing it doesn't make sense for consumers to use BladeElementRef anymore they can, on RN they will do `useRef<View>` or `useRef<Pressable>` etc

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
```diff
@@ -334,7 +334,7 @@ const OTPInput = ({
             hideLabelText={true}
             id={`${inputId}-${index}`}
             textAlign="center"
-            ref={ref}
+            ref={ref as never}
```
kamleshchandnani, line 337: why never?

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
```diff
@@ -212,7 +205,10 @@ const _SelectInput = (
       {!isReactNative() ? (
         <VisuallyHidden>
           <input
-            ref={inputRef as React.Ref<HTMLInputElement>}
+            onFocus={() => {
```
kamleshchandnani, line 208: why do we do this onFocus? isn't it handled already?

packages/blade/src/utils/useMergeRefs.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  if (typeof ref === 'function') {
+    ref(value);
+    return;
+  }
+
+  try {
+    ref.current = value;
+  } catch (error: unknown) {
+    throw new Error(`Cannot assign value '${value}' to ref '${ref}'`);
```
kamleshchandnani, line 18: use out throwError util?

packages/blade/src/utils/useMergeRefs.ts
```diff
@@ -0,0 +1,48 @@
...
+function assignRef<T = any>(ref: ReactRef<T> | null | undefined, value: T) {
+  if (ref == null) return;
+
+  if (typeof ref === 'function') {
+    ref(value);
+    return;
+  }
+
+  try {
+    ref.current = value;
```
kamleshchandnani, line 16: are we assigning ref.current(incoming ref.current) to internal ref? basically, does it mean ref.current=internalref? if yes then how will focus of internal ref work?

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
anuraghazra: Check the old diff, it's as same as the previous logic just that we were using the ref's focus event to trigger this.

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
anuraghazra: Because I NEVER want to mess with absolute mess of React's ref types.

Issue #1353, opened 2023-06-23
Replace Custom Dropdown with Floating UI Implementation


saurabhdaware:
We didn't come across Floating UI while building Dropdown but it seems like a perfect library for our usecase.

Example: https://codesandbox.io/s/admiring-lamport-5wt3yg?file=/src/DropdownMenu.tsx

Things to explore
- [ ] How can header and footer focus be implemented in this?
- [ ] Tags implementation

Will simplify few things for future
- [ ] Nested Menus will get easy if usecase comes
- [ ] Blurring logic is very hacky right now. If Floating UI happens to support header and footer, blurring logic will get simpler
- [ ] Will remove lots of custom code

anuraghazra:
@saurabhdaware did we do this? I remember you replacing this with FloatingUI

saurabhdaware:
> @saurabhdaware did we do this? I remember you replacing this with FloatingUI

The dropdown's open, close, and handling screenwidth logic is on floating ui. But there are still few things like typeahead, etc that can be moved so will keep it open for now

PR #1559, opened 2023-08-24
refactor(DropdownOverlay): remove custom overlay code and use floating ui

### LINKED ISSUE #1546: Not able to control height of the dropdown overlay
- Currently there is no way to control the height of the dropdown overlay.
- This problem exist when we have a dropdown inside a modal, and the modal height is too small to occupy the dropdown options within it

Sample screen shot

Screenshot 2023-08-18 at 6 59 07 PM

### LINKED ISSUE #1556: Dropdown Overlay not coming on top with parent as overflow: scroll
- If the Dropdown component is used inside a parent element with overflow: scroll. The dropdown overlay is not coming on top of the elements.

- The same is working if we remove overflow from the parent but our use case involves overflow elements.

image

saurabhdaware:
- Remove custom logic from DropdownOverlay
- Add floating-ui implementation

fixes #1556 
fixes #1546


packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -144,6 +149,7 @@ const _Dropdown = ({
       setIsKeydownPressed,
       dropdownBaseId,
       triggererRef,
+      triggererWrapperRef,
```
kamleshchandnani, line 152: this needs to be set by consumers?

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -215,78 +84,21 @@ const _DropdownOverlay = ({ children, testID }: DropdownOverlayProps): React.Rea
...
-          left="spacing.0"
-          height="100%"
-          width="100%"
-          onClick={() => setIsOpen(false)}
-        />
-      ) : null}
-      <AnimatedOverlay
+    <BaseBox
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      ref={refs.setFloating as any}
```
kamleshchandnani, line 90: maybe as `never` so you NEVER mess with it? 😆

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -228,6 +228,7 @@ type BaseInputCommonProps = FormInputLabelProps &
      * true if popup is in expanded state
      */
     isPopupExpanded?: boolean;
+    setInputWrapperRef?: (node: HTMLDivElement) => void;
```
kamleshchandnani, line 231: we already have ref exposed right? what is the ref about?

packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
```
kamleshchandnani, line 1: no changes on RN?

kamleshchandnani:
resolve conflicts

packages/blade/src/components/Dropdown/docs/DropdownWithButton.stories.tsx
```diff
@@ -161,6 +161,57 @@ InternalMenu.parameters = {
   },
 };
 
+export const InternalAutoPositioning = (): React.ReactElement => {
```
kamleshchandnani, line 164: why mark this as Internal? this is a good to have for consumer's ref too?

kamleshchandnani: event these
Screenshot 2023-08-25 at 11 03 35 AM

saurabhdaware: We already have two examples for auto positioning 
- https://blade.razorpay.com/?path=/story/components-dropdown-with-select--with-bottom-aligned-select
- https://blade.razorpay.com/?path=/story/components-dropdown-with-button-and-link--with-right-aligned-menu


I just needed one example where I can test locally and see all positions at one place since this is important scenario for Dropdown

packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
saurabhdaware: nope the bug was specific to scroll overflow in web. Although in RN also we can use floating-ui/react-native. I'll do it later since its not that important and in RN people are expected to use bottomsheet anyway so not spending much time on it.

I do have task to relook at Dropdown and ActionList and move several parts of it to floating-ui. Will do this later #1353

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
saurabhdaware: image

We have ref exposed for the actual input field (red one in this screenshot)

But to place dropdown and size dropdown correctly with input, we need reference to the wrapper element (blue one in screenshot)

Earlier we had statically defined size of trailing, then checking width of trailing, if trailing exists or not, etc then add that to input element to get final width but thats too complex and likely to break in future. And with floating-ui, it requires even more hacky code to work.

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: nope its internally set on Select and AutoComplete

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
saurabhdaware: image

In most places where we use `any` actually. Can relook into it later if required.

kamleshchandnani (dismissed):

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -22,6 +22,7 @@ import BaseBox from '~components/Box/BaseBox';
 
 const OVERLAY_OFFSET: number = size['8'];
 const OVERLAY_PADDING: number = size['12']; // doesn't have to be exact. Just rough padding for floating ui to decide to show overlay on top or bottom
+const OVERLAY_ZINDEX = 1001;
```
kamleshchandnani, line 25: Can we use our zindex token? chaitanya had created for modal? because it might get overridden by our consumers as in the past?

saurabhdaware: https://github.com/razorpay/blade/blob/da72785595e4b76ccbc5940b66b5d950a5aaedd2/packages/blade/src/components/Modal/modalTokens.ts#L21

We haven't created zIndex tokens yet 🤔

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -1,211 +1,81 @@
...
-  const { refs } = useFloating({
+  const { refs, floatingStyles, context } = useFloating({
     open: isOpen,
     onOpenChange: setIsOpen,
-    strategy: 'absolute',
+    strategy: 'fixed',
     placement: 'bottom-start',
     elements: {
-      reference: triggererRef.current,
+      reference: triggererWrapperRef.current ?? triggererRef.current,
```
anuraghazra, line 48: can we add a comment why triggererWrapperRef ?? triggerRef is needed?

kamleshchandnani: I think yeah add this in code
https://github.com/razorpay/blade/pull/1559#discussion_r1305438920

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
saurabhdaware: Actually we also allow overriding the zIndex on modal now. We didn't add it in Dropdown earlier. I'll add it here.

packages/blade/src/components/Dropdown/docs/_KitchenSink.Dropdown.stories.tsx
```diff
@@ -12,14 +12,16 @@ const allStories = [
...
-      {allStories.map((Story) => {
-        return (
-          <>
-            <Heading>{Story.storyName}</Heading>
-            <Story />
-          </>
-        );
-      })}
+      {allStories
+        .filter((Story) => Story.storyName.startsWith('Internal'))
```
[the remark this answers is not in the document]
saurabhdaware: Other stories are sandpack stories in Dropdown which are not stable and keep on changing based on loading state.

So earlier also we had snapshots disabled for them and enabled for Internal stories

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
saurabhdaware: done

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
saurabhdaware: done

anuraghazra (approved):

Issue #1087, opened 2023-03-23
Upgrade to new rollup version


saurabhdaware:
Can't we upgrade our rollup version and use this?

_Originally posted by @chaitanyadeorukhkar in https://github.com/razorpay/blade/pull/1045#discussion_r1143179398_
            
We would be able to remove custom babel plugin of manualPureFunction with it.

TODO:
- Check new features that it offers
- Check breaking changes
- Upgrade

saurabhdaware:
> Any specific version we are targeting.

We are looking to move to latest version

> Any specific feature we are looking at after upgrade

One particular we're looking for is -
- https://rollupjs.org/configuration-options/#treeshake-manualpurefunctions (we currently have custom babel plugin)

I think rest are the things we can pick one-by-one when usecase comes and that feature is required. Currently we can just focus to migrate and fix breaking changes.

saurabhdaware:
awesome! thanks 👯 Let us know if you need some help.

kamleshchandnani:
@loombarishav do you have an ETA to finish this?

loombarishav:
I picked this task in Q1 S4 sprint, but due to some P0 feature delivery (3 weeks work), I moved this to backlog. And after that i forgot to pick this task again. My bad. 
Surely Will pick and complete this in next sprint (Already packed heavily in this sprint).
@kamleshchandnani

snitin315:
@saurabhdaware I tried this new option but it doesn't seem to be working as expected. I see size-limit crossing if I remove the babel-plugin and use this option.

<img width="632" alt="Screenshot 2023-09-04 at 8 45 56 AM" src="https://github.com/razorpay/blade/assets/46647141/bab818f1-42d2-4821-8159-4e8c4dc68501">

<img width="761" alt="Screenshot 2023-09-04 at 8 44 30 AM" src="https://github.com/razorpay/blade/assets/46647141/b49327a9-02d1-42cc-a4ce-e710dbe69420">

anuraghazra:
Done as part of #1563

PR #1563, opened 2023-08-25
build: upgrade rollup to v3

snitin315:
## Rollup Changes

- Fixes https://github.com/razorpay/blade/issues/701
- Upgrades Rollup and related plugins to the latest version.
- Upgrades `rollup-plugin-dts` to v5, v6 is incompatible with the current Nodejs version on CI.
- [In rollup config,](https://rollupjs.org/migration/#using-configuration-files) refactored `__dirname` usage in favor of `improt.meta.url` for ESM.
  Screenshot 2023-08-31 at 9 16 27 AM

### Metrics

We didn't see a significant difference in build times, this is because in v3.0.0  rollup team hasn't done much related to caching/build times. They have majorly dropped support for older node versions, removed deprecated options, updated some defaults, etc. They do have a cache option (true by default) but it's useful in watch mode only.

#### Master (Before)

-  `yarn build`  - **62s**
- `yarn build:react-dev/prod` - **12s**
- `yarn build:react-native` - **10s**
- `yarn build:generate-css-variables` - **4s**
- `yarn build:generate-root-imports` - **3s**
- `yarn build:generate-types` - **17s**

#### Rollup Upgrade (After)

- `yarn build` - **57s**
- `yarn build:react-dev/prod` - **10s**
- `yarn build:react-native` - **9s**
- `yarn build:generate-css-variables` - **4s**
- `yarn build:generate-root-imports` - **3s**
- `yarn build:generate-types` - **18s** (This has increased because we now transform alias imports to relative imports in types declarations)

## Typescript Changes

- Use the [`typescript-transform-paths` package](https://www.npmjs.com/package/typescript-transform-paths) to transform alias imports in `.d.ts` file to normal relative imports, this is to make sure `rollup-plugin-dts` doesn't crash like mentioned in  https://github.com/razorpay/blade/issues/701
- Add [`ttypescript` package](https://www.npmjs.com/package/ttypescript) to support custom transformers in the `tsconfig.json`. It's compatible with the current ts version.


packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -1,16 +1,16 @@
...
   FormInputLabelProps,
   FormInputValidationProps,
   FormInputHandleOnEvent,
   FormInputOnEvent,
   FormHintProps,
-} from '../../Form';
-import { StyledBaseInput } from './StyledBaseInput';
-import { BaseInputVisuals } from './BaseInputVisuals';
-import { BaseInputWrapper } from './BaseInputWrapper';
+} from '~components/Form';
```
snitin315, line 13: - Updated this import to reproduce this issue https://github.com/razorpay/blade/issues/701.
- With this change, the build fails on master branch.

packages/blade/src/components/BladeProvider/BladeProvider.native.tsx
```diff
@@ -19,7 +15,7 @@ const BladeProvider = ({
   const { theme, themeContextValue } = useBladeProvider({ initialColorScheme, themeTokens });
 
   return (
-    <GestureHandlerRootView style={gestureHandlerStyle}>
+    <GestureHandlerRootView style={{ flex: 1 }}>
```
snitin315, line 18: This fails on master too.

packages/blade/.babelrc.js
```diff
@@ -1,5 +1,3 @@
-const manualPureFunctions = require('./scripts/manualPureFunctions');
```
snitin315, line 1: Fixes https://github.com/razorpay/blade/issues/1087

packages/blade/rollup.config.mjs
```diff
@@ -51,10 +52,19 @@ const themeBundleCategories = ['tokens', 'utils'];
 
 const aliases = pluginAlias({
   entries: [
-    { find: '~src', replacement: `${__dirname}/${inputRootDirectory}` },
-    { find: '~components', replacement: `${__dirname}/${inputRootDirectory}/components` },
-    { find: '~utils', replacement: `${__dirname}/${inputRootDirectory}/utils` },
-    { find: '~tokens', replacement: `${__dirname}/${inputRootDirectory}/tokens` },
+    { find: '~src', replacement: fileURLToPath(new URL(inputRootDirectory, import.meta.url)) },
```
anuraghazra, line 55: What is this fileURLToPath doing, and can't this be done with the node:path module? 🤔

packages/blade/src/components/BladeProvider/BladeProvider.native.tsx
anuraghazra: Let's revert this anyways, on the root provider best to avoid plain objects since it might cause unnecessary rerender of whole root if any state changes inside useBladeProvider.

packages/blade/tsconfig.json
```diff
@@ -10,8 +10,12 @@
...
       "~src/*": ["src/*"]
     },
-    // In Blade side we are targetting `.web` types
+    // In Blade side we are targeting `.web` types
     // We can't add both `.native` & `.web` at once.
-    "moduleSuffixes": [".all", ".web", ""]
+    "moduleSuffixes": [".all", ".web", ""],
+    "plugins": [
+      // Transform paths in output .d.ts files
+      { "transform": "typescript-transform-paths", "afterDeclarations": true }
```
anuraghazra, line 18: Add a TODO here to replace this with official compiler transform plugins once TS supports this: 

```
// TODO: Replace ttypescript when official compiler transforms are supported https://github.com/microsoft/TypeScript/issues/54276
```

anuraghazra (commented):
TODO: 

- [x] Build blade and bootstrap a new project and check if intellisense & types are correctly resoling
- [x] Do the same with react-native types (you can just change the moduleSuffixes on the bootstrapped project to use `.native` extension)
- [x] Once you manually test, go to https://arethetypeswrong.github.io and test the packaged build for any type related issues

packages/blade/rollup.config.mjs
snitin315: > can't this be done with the node:path module?

No, this is also [recommended from rollup.](https://rollupjs.org/command-line-interface/#caveats-when-using-native-node-es-modules)

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
kamleshchandnani: can you also check for relative imports in other files there were quite a few and replace them with their aliases?

kamleshchandnani:
@snitin315 can you put the stats from slack to the PR description here mentioning about the before/after of rollup upgrade?

packages/blade/src/components/BottomSheet/__tests__/__snapshots__/BottomSheet.web.test.tsx.snap
```diff
@@ -312,47 +353,6 @@ exports[`<BottomSheet /> should render Header/Footer/Body properly 1`] = `
   -webkit-box-orient: vertical;
 }
 
-.c24 {
-  border: none;
```
snitin315, line 316: This just moved up, no idea why.

packages/blade/src/components/BottomSheet/__tests__/__snapshots__/BottomSheet.web.test.tsx.snap
```diff
@@ -896,19 +896,6 @@ exports[`<BottomSheet /> should render empty header with padding 0 1`] = `
   border-radius: 9999px;
 }
 
-.c11 {
-  color: hsla(217,56%,17%,1);
```
snitin315, line 900: This moved down.

packages/blade/src/components/BottomSheet/__tests__/__snapshots__/BottomSheet.web.test.tsx.snap
snitin315: Only changed relative imports to alias imports.

kamleshchandnani (approved):

Issue #1533, opened 2023-08-11
 IconComponent: adding default props for size and colour


abhieshekumar:
1. If we simply do `<ArrowRightIcon />`  or `<ArrowRightIcon size="medium"/>` it doesn’t display the icon. They expect explicit props for size and colour. Can we please look into adding default props for the same?
2. Also if we only give colour prop `<ArrowRightIcon color="feedback.icon.neutral.lowContrast"/>`, the size exceeds beyond `2xlarge` (attached screenshot).
3. [Slack](https://razorpay.slack.com/archives/CMQ3RBHEU/p1691681309025459)
![image](https://github.com/razorpay/blade/assets/24494390/e3f7f0b5-86b9-4bc9-b231-c5fdab86326b)

PR #1567, opened 2023-08-28
fix: Add default props for size and colour in IconComponent

archie252000:
Fixes: #1533

### DIFF
--- a/packages/blade/src/components/Icons/ArrowDownRightIcon/ArrowDownRightIcon.tsx
+++ b/packages/blade/src/components/Icons/ArrowDownRightIcon/ArrowDownRightIcon.tsx
@@ -1,8 +1,13 @@
 import { Svg, Path } from '../_Svg';
 import type { IconComponent } from '..';
 import useIconProps from '../useIconProps';
+import { defaultProps } from '../defaults';
 
-const ArrowDownRightIcon: IconComponent = ({ size, color, ...styledProps }) => {
+const ArrowDownRightIcon: IconComponent = ({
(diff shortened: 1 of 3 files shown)


archie252000:
It currently only adds default props for ```Arrow

packages/blade/src/components/Icons/ArrowDownRightIcon/ArrowDownRightIcon.tsx
```diff
@@ -1,8 +1,13 @@
...
 import type { IconComponent } from '..';
 import useIconProps from '../useIconProps';
+import { defaultProps } from '../defaults';
 
-const ArrowDownRightIcon: IconComponent = ({ size, color, ...styledProps }) => {
+const ArrowDownRightIcon: IconComponent = ({
+  size = defaultProps.size,
+  color = defaultProps.color,
+  ...styledProps
+}) => {
```
anuraghazra, line 10: fyi before you try to change this on all 260 icons, 

It might be better to change the defaults inside the useIconProps hook. 

> though we can also change it on all 260 icons individually, we actually have a script to autogenerate icons.

archie252000: I think it will be better to change this for each icon individually so that the user can use the components right away without providing any props. Can you please tell me about the script?

snitin315:
@archie252000 Thanks for the PR, closing it in favor of https://github.com/razorpay/blade/pull/1603. We need to implement this on Priority.

PR #1569, opened 2023-08-29
fix(RadioGroup): use `radiogroup` as a11y role

snitin315:
Follow up https://github.com/razorpay/blade/pull/1547#discussion_r1307136951


.changeset/chatty-coats-wave.md
```diff
@@ -0,0 +1,7 @@
+---
+"@razorpay/blade": patch
+---
+
+fix(RadioGroup): use `radiogroup` as a11y role
+
+We've upgraded the RadioGroup component with 'radiogroup' role, significantly boosting accessibility for screen readers and assistive technologies.
```
anuraghazra, line 7: Let's communicate whats changed: 

```suggestion
Fixed accessibility of RadioGroup component by changing it's role to `radiogroup` instead of `group`
```

anuraghazra (approved):

packages/blade/src/components/Radio/__tests__/RadioGroup.web.test.tsx
```diff
@@ -20,7 +20,7 @@ describe('<RadioGroup />', () => {
       </RadioGroup>,
     );
     expect(container).toMatchSnapshot();
-    expect(getByRole('group')).toHaveTextContent(labelText);
+    expect(getByRole('radiogroup')).toHaveTextContent(labelText);
```
kamleshchandnani, line 23: web pe to we are setting `group` only right? so why are we checking `radiogroup` here?

snitin315: >web pe to we are setting group only right? so why are we checking radiogroup here?

RadioGroup should be `radiogroup` & CheckboxGroup should be `group` on both web and native. 

>also the comment says that it shall be radiogroup only for both?

The comment is only for the RadioGroup component.

@anuraghazra Can you confirm?

anuraghazra: Yes it should be [group only](https://russmaxdesign.github.io/accessible-forms/fieldset-checkbox-group-role-group.html), there is nothing called a "checkboxgroup" and using radiogroup for checkbox will be misleading for screen reader users.

kamleshchandnani (approved):

Issue #1574, opened 2023-08-29
Allow tooltip to be shown when button is disabled


parichay28:
Current implementation doesn't show tooltip on hover in case button is in disabled state.

Usecase: Use tooltip to let users know why certain CTAs are in disabled state

Issue Repro: https://codesandbox.io/s/p84w4g?file=/App.tsx&utm_medium=sandpack

anuraghazra:
@parichay28 to make non-interactive components like disabled button interactive you need to wrap the button in TooltipInteractiveWrapper. 

Demo: https://codesandbox.io/s/distracted-smoke-c5hxk2?file=/App.tsx

parichay28:
Ahh missed this 🙈, thanks for sharing. Closing this issue

PR #1584, opened 2023-08-30
docs: Popover API Decision

anuraghazra:
## [See Formatted Doc](https://github.com/razorpay/blade/blob/anu/popover-api/packages/blade/src/components/Popover/_decisions/decisions.md)


packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
+# Popover API Decisions <!-- omit in toc -->
+
+The popover component is used to provide context as well as enable users to take certain actions on it. They are designed to display relevant information without requiring the user to navigate to a different page or dialog. Popovers are triggered on click or onload.
+
+<img src="./assets/popover-thumbnail.png" width="380" alt="Popover Thumbnail" />
+
+- [Design](#design)
+- [Basic Features](#basic-features)
```
kamleshchandnani, line 8: add before this section

* Link to Tooltip decision
* Comparision should come here before even we start popover related things?

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+- [Motion](#motion)
+- [Accessibility](#accessibility)
+- [Open Question](#open-question)
+- [References](#references)
+
+## Design
+
+[Figma Link](https://www.figma.com/file/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?type=design&node-id=2875-40870&mode=design&t=mTbpUCQOr7kHk6UT-0) to all variants of the Popover component
+
+## Basic Features
```
kamleshchandnani, line 29: very nice 👌

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+
+- Controlled & Uncontrolled state
+- Can be triggered on click or on load
```
kamleshchandnani, line 32: do we support on hover? because technically i can open it with anything right since it has controlled variant?

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+| defaultIsOpen   | `boolean`                                                                                                              | `undefined` | Uncontrolled state of the popover                                                                                                     |          |
+| onOpenChange    | `({ isOpen }) => void`                                                                                                 | `undefined` | Called when popover isOpen state is changed, this can be used to detect when popover opens or closed                                  |          |
+| initialFocusRef | `React.RefObject`                                                                                                      | `undefined` | The ref of the element that should receive focus when the popover opens.                                                              |          |  |
```
kamleshchandnani, line 48: onDismiss button click? consumers might need it for triggering some actions or analytics

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+| headerTitle     | `string`                                                                                                               | `undefined` | Title of the header                                                                                                                   | ✅        |
+| headerLeading   | `React.ReactNode`                                                                                                      | `undefined` | Leading element to be placed before header title, recommended components are Icon or Asset. The icon's size will be set to large      |          |
+| footerContent   | `React.ReactNode`                                                                                                      | `undefined` | Slottable content of the footer                                                                                                       |          |
```
kamleshchandnani, line 41: how about just `footer` since footer doesn't has any header-content-footer trio

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+#### Working with non-interactive triggers
+
+To make popover apear on clicking a non-interactive elements such as icons, badges, counters etc. We will provide a `PopoverInteractiveWrapper` component which will work as a minimal trigger:
+
+```js
+import { Popover, PopoverInteractiveWrapper } from '@razorpay/blade';
+
+// non-interactive element as trigger
+<Popover content="Amount reversed to customer bank account">
+  <PopoverInteractiveWrapper>
```
kamleshchandnani, line 134: is there anything special in `PopoverInteractiveWrapper` compared to the tooltip one? if not can we use the same wrapper here as well?

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+      // or pick the onMouseDown, onMouseUp.. etc events and add them individually
+      {...props}
+    >
+       {children}
+    </button>
+  );
+});
+
+// Custom trigger
+<Popover content="Amount reversed to customer bank account">
```
kamleshchandnani, line 173:
```suggestion
<Popover headerTitle="Refund Status" content="Amount reversed to customer bank account">
```

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+  <PopoverInteractiveWrapper>
+    <Counter value={100} />
+  </PopoverInteractiveWrapper>
+</Popover>;
+```
+
+> Note: If users don't wrap the non-interactive elements in `PopoverInteractiveWrapper` the popover won't apear
+
+### Custom Triggers
```
kamleshchandnani, line 142:
```suggestion
### Custom Interactive Triggers
```

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+
+Users will be able to use their own custom components as triggers:
```
kamleshchandnani, line 144:
```suggestion
Users will be able to use their own custom interactive components(mention the names of these components) as triggers:
```

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+- [Accessibility](#accessibility)
+- [Open Question](#open-question)
+- [References](#references)
+
+## Design
+
+[Figma Link](https://www.figma.com/file/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?type=design&node-id=2875-40870&mode=design&t=mTbpUCQOr7kHk6UT-0) to all variants of the Popover component
+
+## Basic Features
+
```
kamleshchandnani, line 30: mention somehwere that popovers are not auto dismissible unline tooltip

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+- Proper guidelines on how to use the component
+- etc
+
+This will be a separate component and will be released in future, since the scope & complexity of Guided Popovers will be much larger than normal Popovers.
+
+## Tooltips vs Popovers vs Guided Popovers
+
+|                         	| Tooltip                                          	| Popover                                                        	| GuidedPopover                                                  	|
+|-------------------------	|--------------------------------------------------	|----------------------------------------------------------------	|----------------------------------------------------------------	|
+| **Use Case**            	| Set context                                      	| Set context<br>Let user take action                            	| Highlight new feature<br>Let users take action                 	|
```
kamleshchandnani, line 196: add a point for dismiss interaction?

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+| **Trigger Interaction** 	| Hover                                            	| onClick<br>onLoad                                              	| onClick<br>onLoad                                              	|
+| **Triggered By**        	| Interactive elements<br>Non-interactive elements 	| Interactive elements<br>Non-interactive elements               	| On page load<br>Nav buttons of current popover                 	|
```
kamleshchandnani, line 198: even popovers can open on page loads right? its not mentioned here

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+| **Content**             	| Heading<br>Body Text (Normal + Bold)            	| Heading<br>Rich Body Text<br>Links<br>Buttons<br>Assets/ Media 	| Heading<br>Rich Body Text<br>Links<br>Buttons<br>Assets/ Media 	|
+| **Header**              	| Plain Text                                       	| Rich Text<br>Leading Icon<br>Leading Asset                     	| Rich Text<br>Leading Icon<br>Leading Asset                     	|
```
kamleshchandnani, line 200: how can you say header is rich text for popover? its still plain text right? it can just have leading assets right?

kamleshchandnani, line 200:
```suggestion
| **Header**              	| Plain Text                                       	| Rich Text<br>Leading Asset(Icon/Avatar/etc.)                     	| Rich Text<br>Leading Icon<br>Leading Asset                     	|
```

icon is also a type of asset

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+      text: "Secondary me",
+      onClick: () => {},
+    },
+  }}
+>
+  <IconButton icon={InfoIcon} />
+</Popover>
+```
+
+**Conclusion:** We should go ahead with Flexible API because even if we go with constrained API, users will anyway still be able to pass React.ReactNode as `footerContent` and could just skip the `footerActions` altogether.
```
kamleshchandnani, line 269:
```suggestion
**Conclusion:** We will go ahead with Flexible API as this will mostly be controlled by design and design has the slots and permutations and combinations in place of what is allowed and what is disallowed.
```

packages/blade/src/components/Popover/_decisions/decisions.md
kamleshchandnani: Can you also add comparision to tooltip decisions?

anuraghazra: > Link to Tooltip decision

Added

> Comparision should come here before even we start popover related things?

Might get a bit overwhelming at first, let them go through the basic functionality first and then they will eventually reach the comparison.

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: Technically yes they can hack around it, but we don't provide that feature nor recommend it do so.

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: 🤔 yeah but right now onOpenChange will be called when close button is clicked or click outside popover, adding onDismissClick might get a bit redundant.

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: > (mention the names of these components)

Anything they like. as long as they expose the refs and events.

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: I think this could be a good point for comparison.

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: Already added at the end. "Dismiss Button"

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: added

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: Right.

anuraghazra: We are differentiating this from design side also, in the slot when you choose it will have icon & asset (slot) as separate things.

kamleshchandnani: probably call it as icon and other assets on design as well?

packages/blade/src/components/Popover/_decisions/decisions.md
kamleshchandnani: no i mean dismiss button change handler

packages/blade/src/components/Popover/_decisions/decisions.md
kamleshchandnani: but then how will we diff between interactive and non-interactive and when to put the wrapper for making the underlying element interactive?

packages/blade/src/components/Popover/_decisions/decisions.md
kamleshchandnani: there are 2 events that will happen here

onDismiss will also trigger onOpenChange because that's an action user is performing explicitly and onOpenChange is a handler that we are triggering when the state of the tooltip is changing right?

packages/blade/src/components/Popover/_decisions/decisions.md
kamleshchandnani: lets make sure we put that in bold in storybook

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: Sure, but should onDismiss also be called on clicking outside the popover or just clicking on close button only? 

clicking outside popover to close it is also an user action. 🤔

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+</Popover>;
+```
+
+> Note: If users don't wrap the non-interactive elements in `PopoverInteractiveWrapper` the popover won't apear
+
+### Custom Interactive Triggers
+
+Users will be able to use their own custom interactive components as triggers:
+
+To make custom elements work the components needs to expose it's inner ref, this can be achived by using `React.forwardRef`:
```
nashcheez, line 146:
```suggestion
To make custom elements work, the component needs to expose its inner ref, this can be achieved by using `React.forwardRef`:
```

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+
+Apart from normal Popovers while we audited the components used in Razorpay, we also found instances of Guided Popovers (Tour component), which can be used to guide users through a flow. 
+We will be adding support for Guided Popovers in future.
+
+Guided Popovers will have additional features like: 
+
+- Ability to add multiple steps in a flow without juggling with state management
+- Mask & backdrop support
+- Proper guidelines on how to use the component
+- etc
```
nashcheez, line 188:
```suggestion
```

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: PopoverInteractiveWrapper is a convenience wrapper that we provide to consumers, technically it is a "Custom Trigger" just that we expose this custom trigger so that consumer doesn't have to wire up the custom trigger for each time they are using a non-interactive element since the wiring up code is quite boilerplate-y.  

Also imagine this case: 
https://github.com/razorpay/blade/issues/1574

```jsx
    <Tooltip content="Hello world" placement="bottom">
      <TooltipInteractiveWrapper>
        <Button isDisabled={true}>Hover over me</Button>
      </TooltipInteractiveWrapper>
    </Tooltip>
```

If a Button is disabled, it won't respond to any events (onClick, mouseDown etc), in this case if you want to open the popover you need to wrap it in PopoverInteractiveWrapper.

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: Added

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: Fixed

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+
+- Controlled & Uncontrolled state
+- Can be triggered on click or on load
+- Can be placed on different placements (top, bottom, left, right)
+
+## `Popover` API
+
+| Prop            | Type                                                                                                                   | Default     | Description                                                                                                                           | Required |
+| --------------- | ---------------------------------------------------------------------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| headerTitle     | `string`                                                                                                               | `undefined` | Title of the header                                                                                                                   | ✅        |
```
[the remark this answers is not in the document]
anuraghazra: We could do that, but we also have leading prop, and the header itself is a separate section inside Popover, prefixing with "header" helps with the intent that the "title" prop is for header's title, because in some design system they use the "title" prop as the content of the popover. (eg: [mui](https://mui.com/material-ui/react-tooltip/#basic-tooltip))

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+| headerLeading   | `React.ReactNode`                                                                                                      | `undefined` | Leading element to be placed before header title, recommended components are Icon or Asset. The icon's size will be set to large      |          |
+| footerContent   | `React.ReactNode`                                                                                                      | `undefined` | Slottable content of the footer                                                                                                       |          |
+| content         | `string`                                                                                                               | `undefined` | Content of the popover                                                                                                                |          |
```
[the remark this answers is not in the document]
anuraghazra: Right a typo, that should be ReactNode,

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: Yeah design mein it's "Asset" only, and when you select the asset it repalced the content with a slot.

packages/blade/src/components/Popover/_decisions/decisions.md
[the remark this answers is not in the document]
nashcheez: In Blade's terminology I think `title` and `description` sound better than `headerTitle` and `content`. The title in the Modal is also called `title` and not `headerTitle` even if it appears in the ModalHeader primarily because it defines the content inside the modal.

`headerLeading` is also an unclear name on what the API is supposed to do. Depending on what the vocabulary is we can also call it `titleLeadingAsset` or `headerLeadingAsset`.

packages/blade/src/components/Popover/_decisions/decisions.md
kamleshchandnani: no only explicitly clicking on the button you can call it `onDismissButtonClick` for better context and understanding. outside click to you'll trigger `onOpenChange` na

packages/blade/src/components/Popover/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: agree with abinash on this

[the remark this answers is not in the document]
anuraghazra: Okay it seems we will need a larger discussion around this, lets discuss this tommorow.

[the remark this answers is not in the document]
nashcheez: Also, noticed - we have followed the same naming conventions in Card and BottomSheet - `title` in `ComponentHeader`. Need to follow a similar convention, don't think we have ever had a prop called `headerTitle` yet.

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/Popover/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Discussed, going ahead without prefixing. 

headerTitle -> title
headerLeading -> titleLeading

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: the change handler is a implementation detail / prop, don't think it qualifies as a distinctive difference. 

If there is not Dismiss button, it's implied that there can't be onDismissButtonClick.

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+
+- Controlled & Uncontrolled state
+- Can be triggered on click or on load
+- Can be placed on different placements (top, bottom, left, right)
+
+## `Popover` API
+
+| Prop            | Type                                                                                                                   | Default     | Description                                                                                                                           | Required |
+| --------------- | ---------------------------------------------------------------------------------------------------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| title     | `string`                                                                                                               | `undefined` | Title of the header                                                                                                                   | ✅        |
```
cseas, line 39: Why is `title` required? There are use-cases where we only want to render just the content. Example, on this page the Tooltip/Popover contains just text with a link.

https://razorpay.com/payment-links/

Screenshot 2023-09-07 at 1 18 16 PM

anuraghazra: Typo, title isn't a required prop

packages/blade/src/components/Popover/_decisions/decisions.md
```diff
@@ -0,0 +1,275 @@
...
+| isOpen          | `boolean`                                                                                                              | `undefined` | Controlled state of the popover                                                                                                       |          |
+| defaultIsOpen   | `boolean`                                                                                                              | `undefined` | Uncontrolled state of the popover                                                                                                     |          |
+| onOpenChange    | `({ isOpen }) => void`                                                                                                 | `undefined` | Called when popover isOpen state is changed, this can be used to detect when popover opens or closed                                  |          |
+| onDismissButtonClick    | `() => void`                                                                                                 | `undefined` | Called when dismiss button is clicked, useful for sending analytic events                                  |          |
```
divyanshu013, line 48: Separate event handler for dismiss seems bit redundant because we already have an event handler that runs for both show / hide 🤔

divyanshu013 (commented):
Looks good to me.

packages/blade/src/components/Popover/_decisions/decisions.md
anuraghazra: Yeah kinda agree 🤔  could cause confusion actually, we added onDismissButtonClick for analytics purpose. https://github.com/razorpay/blade/pull/1584#discussion_r1315529213

@kamleshchandnani maybe we can re-evaluate this, the question is what is the intent that analytics team will be interested in? It will most probably be "closing of popup" doesn't matter how. 

Plus it's more convenient to click the popup by clicking outside than clicking the dismiss button, thus if someone adds analytic events in onDismissButtonClick, the analytics team might not get all the "closed popup" events.

divyanshu013: Did we decide anything on this?

kamleshchandnani: While there are trade-offs the thing is product when it comes to analytics they track everything. But we can comeback on this and might add this as an enhancement. Think of it this way if product wants to track ho many people dismissed the popover via moving the cursor outside vs clicking on the dismiss button and that can change a lot of things for product. but lets park it for now

kamleshchandnani (approved):

pcprashant:
Looks good 👍

PR #1586, opened 2023-08-30
feat(AutoComplete): add AutoComplete trigger

saurabhdaware:
> **Note**
> 
> Ignore the documentation and tests in this. I will be creating separate PR for them stacked on top of this
>
> Also, there aren't that many changes in this file. The diff is a mess because I moved files (so at some places it thinks I deleted and created new file). Adding comments on those scenarios to make it easy for review.

This PR
- Moves most logic from SelectInput --> BaseDropdownInputTrigger 
- Adds `AutoComplete` component extending from the same BaseDropdownInputTrigger component


chaitanyadeorukhkar:
Shouldn't this story take up full width?
Screenshot 2023-09-07 at 3 37 31 PM

chaitanyadeorukhkar:
No empty state?
Screenshot 2023-09-07 at 3 38 41 PM

saurabhdaware:
> Shouldn't this story take up full width?

Yup it should. The stories here are just for my testing. I am writing actual documentation in #1608

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
```diff
@@ -0,0 +1,198 @@
+import React from 'react';
```
chaitanyadeorukhkar, line 1: Should we keep autocomplete outside the dropdown component? like `components/AutoComplete`? Since it is used in Dropdown as well as Bottomsheet?

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
```diff
@@ -0,0 +1,198 @@
...
+import { useDropdown } from '../../Dropdown/useDropdown';
+import type { AutoCompleteProps, BaseDropdownInputTriggerProps } from './types';
+import { BaseDropdownInputTrigger } from './BaseDropdownInputTrigger';
+import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
+import BaseBox from '~components/Box/BaseBox';
+import { componentIds } from '~components/Dropdown/dropdownUtils';
+import { BladeElementRef } from '~utils/types';
```
chaitanyadeorukhkar, line 8:
```suggestion
import type { BladeElementRef } from '~utils/types';
```

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
```diff
@@ -0,0 +1,198 @@
...
+    options,
+    setFilteredValues: setGlobalFilteredValues,
+    activeTagIndex,
+    setActiveTagIndex,
+    setActiveIndex,
+    filteredValues: globalFilteredValues,
+    selectionType,
+  } = useDropdown();
+
+  React.useEffect((): void => {
```
chaitanyadeorukhkar, line 42: Since its a lot of specific logic, add comments that could be helpful for later

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
```diff
@@ -0,0 +1,198 @@
...
+    }
+    onBaseDropdownInputKeydown?.(e);
+  };
+
+  const onSelectionChange: BaseDropdownInputTriggerProps['onChange'] = ({ values }) => {
+    if (selectionType === 'multiple') {
+      setInputValue('');
+      props.onInputValueChange?.({ name: props.name, value: '' });
+      setActiveTagIndex(-1);
+      setGlobalFilteredValues(getOptionValues());
```
chaitanyadeorukhkar, line 102: What is this doing exactly? Its setting the filtered values to be the same as all options? Do we need to do that on ever selection change?

saurabhdaware:
> No empty state?

Ops yes left it for the end since it was simple and forgot. Added it here - 203b5d68b681196c75d7583ccd07c3e555f600d9

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+  const {
+    isOpen,
+    options,
+    setFilteredValues: setGlobalFilteredValues,
+    hasAutoCompleteInBottomSheetHeader,
+    setHasAutoCompleteInBottomSheetHeader,
+    onTriggerClick,
+    dropdownTriggerer,
+  } = useDropdown();
```
chaitanyadeorukhkar, line 136: This would fail if used outside dropdown context, right? I don't see `useDropdown` doing a context check either?

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+  const getOptionValues = React.useCallback(() => {
+    return options.map((option) => option.value);
+  }, [options]);
```
chaitanyadeorukhkar, line 140: Since its an array, won't `getOptionValues` get triggered on every render? Does `useCallback` handle that and actually deeply check array changes? 🤔

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
```diff
@@ -0,0 +1,198 @@
...
+  const { onSelectionChange, onTriggerKeydown, onInputValueChange } = useAutoCompleteHandlers({
+    props,
+    inputValue,
+    setInputValue,
+    getOptionValues,
+  });
+
+  React.useEffect(() => {
+    if (dropdownTriggerer !== 'AutoComplete') {
+      // When AutoComplete is mounted but not as trigger,
```
chaitanyadeorukhkar, line 151: Thats a big assumption, no? Can we not use Bottomsheet context to check this?

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
```diff
@@ -0,0 +1,198 @@
...
+      // we assume its in header of BottomSheet
+      setHasAutoCompleteInBottomSheetHeader(true);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  React.useEffect(() => {
+    if (isOpen && !inputValue) {
+      setGlobalFilteredValues(getOptionValues());
```
chaitanyadeorukhkar, line 160: We're doing this very often and in multiple places. I'm still not sure what we're doing with this and if its safe to do it so many times 😅

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
```diff
@@ -0,0 +1,198 @@
...
+      <BaseDropdownInputTrigger
+        {...props}
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ref={ref as any}
+        onChange={onSelectionChange}
+        isSelectInput={false}
+        inputValue={inputValue}
+        onTriggerKeydown={onTriggerKeydown}
+        onInputValueChange={onInputValueChange}
+        onTriggerClick={(e) => {
```
chaitanyadeorukhkar, line 182: nit:
```suggestion
        onTriggerClick={(triggerEvent) => {
```

packages/blade/src/components/Input/DropdownInputTriggers/BaseDropdownInputTrigger.tsx
```diff
@@ -0,0 +1,283 @@
...
+import { getTagsGroup } from '~components/Tag/getTagsGroup';
+import type { BladeElementRef } from '~utils/types';
+
+const useControlledDropdownInput = (
+  props: Pick<
+    BaseDropdownInputTriggerProps,
+    'onChange' | 'onSelectionChange' | 'name' | 'value' | 'defaultValue'
+  >,
+): void => {
+  const isFirstRenderRef = React.useRef(true);
```
chaitanyadeorukhkar, line 19: We could have a common util for this just like we have `useUpdateEffect`, we can have `useFirstRender`?

packages/blade/src/components/Input/SelectInput/SelectInput.tsx
chaitanyadeorukhkar, line 1: Nice!

packages/blade/src/components/ActionList/ActionListNoResults.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const ActionListNoResults = (): React.ReactElement => {
+  return (
+    <Box textAlign="center" padding="spacing.5">
+      <SearchIcon size="xlarge" color="surface.text.normal.lowContrast" />
+      <Text marginTop="spacing.5" weight="bold">
+        No Search Result Found
+      </Text>
+      <Text marginTop="spacing.3" size="small" type="subtle">
+        Try Searching for a Different Text
```
chaitanyadeorukhkar, line 13: Is this on design? Can we not let our consumers also decide what to show here?

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
```diff
@@ -38,7 +38,7 @@ const _DropdownOverlay = ({
   const { theme } = useTheme();
   const bottomSheetAndDropdownGlue = useBottomSheetAndDropdownGlue();
 
-  const isMenu = dropdownTriggerer !== 'SelectInput';
+  const isMenu = dropdownTriggerer !== 'SelectInput' && dropdownTriggerer !== 'AutoComplete';
```
chaitanyadeorukhkar, line 41: Wondering if we should store dropdownTrigger as an enum. Its getting big now

packages/blade/src/components/Dropdown/useDropdown.ts
```diff
@@ -64,9 +71,9 @@ type DropdownContextType = {
   /** common baseId which is prepended to multiple other ids inside this dropdown  */
   dropdownBaseId: string;
   /** Which element has triggered the dropdown */
-  dropdownTriggerer?: 'SelectInput' | 'DropdownButton';
+  dropdownTriggerer?: 'SelectInput' | 'DropdownButton' | 'AutoComplete' | 'DropdownLink';
```
chaitanyadeorukhkar, line 74: Lets store them as constants?

packages/blade/src/components/Dropdown/useDropdown.ts
```diff
@@ -79,12 +86,12 @@ type DropdownContextType = {
...
   setHasFooterAction: (value: boolean) => void;
+
   /**
-   * We need to know the label's position because when it is on left, the overlay takes the width of input.
-   * Rest of the times, we can set 100% width
+   * Apart from dropdownTriggerer prop, we also set this boolean because in BottomSheet, the initial trigger can be Select but also have autocomplete inside of it
    */
-  hasLabelOnLeft: boolean;
-  setHasLabelOnLeft: (value: boolean) => void;
+  hasAutoCompleteInBottomSheetHeader: boolean;
```
chaitanyadeorukhkar, line 93: Isn't this changing the existing behaviour for `hasLabelOnLeft`?

packages/blade/src/components/Dropdown/useDropdown.ts
```diff
@@ -216,6 +224,7 @@ const useDropdown = (): UseDropdownReturnValue => {
     setChangeCallbackTriggerer,
     isControlled,
     setControlledValueIndices,
+    filteredValues,
     ...rest
```
chaitanyadeorukhkar, line 228: nit: Why `rest`? There are only a few other values, lets destruct them all here na?

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -865,7 +868,7 @@ export const BaseInput = React.forwardRef<BladeElementRef, BaseInputProps>(
               inputWrapperRef={inputWrapperRef}
             >
               <StyledBaseInput
-                as={isReactNative ? undefined : as}
+                as={as}
```
chaitanyadeorukhkar, line 871: Won't this break RN if someone passes `as`?

packages/blade/src/components/Input/BaseInput/BaseInputTagSlot.web.tsx
```diff
@@ -96,10 +99,11 @@ const BaseInputTagSlot = ({
     tags,
     maxTagRows,
     visibleTagsCountRef,
+    showAllTags,
   });
 
   React.useEffect(() => {
-    slotRef.current?.scrollTo?.({
+    slotRef.current?.scrollTo({
```
chaitanyadeorukhkar, line 106: Why did we remove the check for scrollTo? We sure it won't ever fail?

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -251,7 +251,7 @@ const _StyledBaseInput: React.ForwardRefRenderFunction<
     isFocused: currentInteraction === 'active',
   };
 
-  return hasPopup ? (
+  return renderAs === 'button' ? (
```
chaitanyadeorukhkar, line 254: logic change? Won't it affect popup?

chaitanyadeorukhkar:
Padding seems incorrect on RN
Screenshot 2023-09-07 at 4 41 08 PM

chaitanyadeorukhkar:
Whitespace seems a bit too much on native?
Screenshot 2023-09-07 at 4 42 56 PM

chaitanyadeorukhkar:
The interaction on clicking "1 more" has a weird active state when I press on it. Shouldn't it be similar to rest of our active states? Also the animation is abrupt - we should add some transition for "1 more" appearing and disappearing

https://github.com/razorpay/blade/assets/24487274/485cf93d-7e3b-4c3b-b1b8-b9312d65569c

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
saurabhdaware: so Dropdown as a whole is used inside BottomSheet actually. So it isn't much different from SelectInput. 

https://blade.razorpay.com/?path=/docs/components-bottomsheet--with-dropdown-single-select

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
saurabhdaware: yup when you select an item, you clear the input field so all options become visible for next selection. I'll add some comments

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
saurabhdaware: It actually works outside dropdown context. It just won't be able to do filtering etc. When used outside dropdown context, it renders like a simple list.

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
saurabhdaware: It's a flat array so I thought it wouldn't get re-render. Let me check 🤔

packages/blade/src/components/ActionList/ActionListNoResults.tsx
saurabhdaware: With Controlled API we do allow that. Can think of exposing additional props if its something that people want to override very often.


On design we have this but it will be only possible with controlled API since it has too much of project-specific info- 

image

Maybe I should not title case it and use `Try searching for a different value`. Will check this with designers at the end. 🤔

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
saurabhdaware: we can add if more triggers come up. I don't think so apart from Select and AutoComplete there will be any other input trigger anytime soon 🤔

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
saurabhdaware: it is handled inside `StyledBaseInput`. We do want to let `as` pass here so that depending on as="button" vs as="input" we can render the correct RN element.

packages/blade/src/components/Input/BaseInput/BaseInputTagSlot.web.tsx
saurabhdaware: yup that referenced element is in the same file only and we have overflow there.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
saurabhdaware: I used popup terminology to refer to dropdown earlier 🙈 So it was always rendering button in Dropdown. But with AutoComplete, our deciding factor is this renderAs prop. We render button only when renderAs is button. Earlier we didn't have input component + dropdown together so this logic worked fine.

saurabhdaware:
> Padding seems incorrect on RN

Yes i noticed. Fixed it.

> I can only click on the icon or the "Select Option" Text on RN to trigger bottomsheet

The whitespace should also be clickable 🤔 Is it not working?

> I'm assuming on RN we aren't going to allow Dropdown being used right? Only bottomsheet?

Yup. We do have basic code to make it work but design recommendation is always bottomsheet.

saurabhdaware:
> Whitespace seems a bit too much on native?

Was the next tag too large? It basically tries to keep space for next tag + the +x more text. If it's not able to fit both, it doesn't add that tag.

packages/blade/src/components/Dropdown/useDropdown.ts
saurabhdaware: oh yup we'll do

packages/blade/src/components/Dropdown/useDropdown.ts
saurabhdaware: ops sorry for confusion. I removed `hasLabelOnLeft` because that type wasn't used anymore. floating-ui migration of DropdownOverlay removed its purpose but it existed in DropdownContext.

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
saurabhdaware: This is actually simpler logic than having to check React.Children of header. Also not a big assumption since it does what it is expected to do. Set a flag when AutoComplete is not used as a trigger. This flag is used to disable enable some trigger properties on that AutoComplete then. As far as we don't decide to show AutoComplete as trigger and inside header both, I can't think of scenario where this breaks 🤔  


We'll have to go into checking the children with React.Children then. Wanted to avoid that wherever possible.

Also this will kind of allow you to put AutoComplete anywhere inside of the Dropdown so in future if we decide to AutoComplete in DropdownHeader, it can work on this logic.

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
saurabhdaware: Yup realised that as well. It is safe but will check if I can refactor this into separate hook maybe

packages/blade/src/components/Dropdown/useDropdown.ts
saurabhdaware: done

packages/blade/src/components/Dropdown/useDropdown.ts
saurabhdaware: Basically removing `hasLabelOnLeft` is unrelated to this PR

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
saurabhdaware: Its comparatively lesser times. It still gets triggered multiple times but thats expected and its not doing any too expensive operation.
image

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
saurabhdaware: rephrased the comment a bit

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
saurabhdaware: moved some of it inside useAutoComplete hook. Also added comment for each useEffect block. Don't think so we can remove any of them as they are all doing something expected.

packages/blade/src/components/Input/DropdownInputTriggers/BaseDropdownInputTrigger.tsx
saurabhdaware: done

packages/blade/src/components/Dropdown/DropdownOverlay.web.tsx
saurabhdaware: I moved it to constants

packages/blade/src/components/Dropdown/useDropdown.ts
saurabhdaware: oh I remembered. So we re-export all dropdown context values from useDropdown + some additional functions.

If we destructure here, it becomes 2 more places to go on adding every value we add to dropdown context. Its already a bit pain because we have to add it to to value in provider, type, initial type, dependency array of Dropdown's useMemo.

I think its better to leave it to rest cause there are around 14 properties that we'll have destructure here and then again add them to return value of this function

packages/blade/src/components/Input/BaseInput/BaseInputTagSlot.web.tsx
saurabhdaware: oh I remembered why I added that earlier. It fails in tests because jsdom can't scroll. Added it back just to on safer side anyway.

packages/blade/src/components/ActionList/ActionListNoResults.tsx
chaitanyadeorukhkar: So if I want to simply change the text, I'll have to change my logic and use controlled API instead. I suspect this use-case might come up soon and we might have to let users customise this text. Anyway, makes sense to defer that decision until the use-case comes up

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -106,6 +111,11 @@ const _Dropdown = ({
   }, [isOpen]);
 
   const close = React.useCallback(() => {
+    // try {
+    //   throw new Error('checking stack');
+    // } catch (err: unknown) {
+    //   console.log(err);
+    // }
```
chaitanyadeorukhkar, line 118: Remove if not needed?

packages/blade/src/components/Input/BaseInput/BaseInputTagSlot.native.tsx
```diff
@@ -73,8 +73,8 @@ const BaseInputTagSlot = ({
   return (
     <BaseBox
       justifyContent="flex-start"
-      paddingY={hasTags ? 'spacing.1' : 'spacing.0'}
-      paddingX={hasTags ? 'spacing.3' : 'spacing.0'}
+      paddingY="spacing.1"
+      paddingX="spacing.4"
```
chaitanyadeorukhkar, line 77: Assuming this isn't affecting baseinput without tags

packages/blade/src/components/Input/DropdownInputTriggers/types.ts
```diff
@@ -0,0 +1,125 @@
...
+   *
+   * Although similar to onChange, this is the callback that we can internally use without interferring with controlled props
+   * This is used in changing AutoComplete's internal states when item gets selected
+   */
+  onSelectionChange?: () => void;
+};
+
+export type SelectInputProps = DropdownInputTriggersProps;
+
+// @TODO add jsdoc in documentation PR
```
chaitanyadeorukhkar, line 120: Todo ^

packages/blade/src/components/index.ts
```diff
@@ -17,7 +17,7 @@ export * from './Input/TextInput';
 export * from './Input/PasswordInput';
 export * from './Input/TextArea';
 export * from './Input/OTPInput';
-export * from './Input/SelectInput';
+export * from './Input/DropdownInputTriggers';
```
chaitanyadeorukhkar, line 20: Make sure no existing exports are restricted now that there is a change here?

chaitanyadeorukhkar (approved):

chaitanyadeorukhkar:
@saurabhdaware seems like there are some unresolved comments for RN. I'm getting it to run on my local to see what's still unresolved but can you also take a look?

packages/blade/src/components/index.ts
saurabhdaware: This diff is because of directory rename. The index.ts is not really changed so it continues to export the same things it was exporting earlier.

saurabhdaware:
Resolved the RN issues you mentioned. Added more details in our [feedback thread on slack](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1694580259302029?thread_ts=1694514473.311699&cid=G01B3LQ9H0W)

chaitanyadeorukhkar (approved):

PR #1595, opened 2023-09-01
feat(Typography): add new Display component

snitin315:
This PR adds a new Typography component `Display`.


packages/blade/src/components/Typography/Display/Display.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <Text as="span" weight="bold">
+          size prop
+        </Text>{' '}
+        passed
+      </Text>
+      <List>
+        <ListItem>small: h3</ListItem>
+        <ListItem>medium: h2</ListItem>
+        <ListItem>large: h1</ListItem>
+        <ListItem>xlarge: h1</ListItem>
```
anuraghazra, line 129: By default display should always render `h1`, because it's the largest of them all, even it's small size is same has the Title.

packages/blade/src/components/Typography/Display/Display.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+    size: {
+      options: ['small', 'medium', 'large', 'xlarge'],
+      control: {
+        type: 'radio',
+      },
+      table: {
+        type: {
+          summary: '"small" | "medium" | "large" | "xlarge"',
+        },
+      },
```
anuraghazra, line 65: Do we ned to manually do this?

packages/blade/src/components/Typography/Display/Display.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const colorContrast: keyof ColorContrast = contrast ? `${contrast}Contrast` : 'lowContrast';
+  const props: Omit<BaseTextProps, 'children'> = {
+    color: color ?? `surface.text.${type ?? 'normal'}.${colorContrast}`,
+    fontSize: 1100,
+    fontWeight: 'bold',
+    fontStyle: 'normal',
+    lineHeight: 900,
+    fontFamily: 'text',
+    accessibilityProps: isPlatformWeb ? {} : { role: 'heading' },
+    componentName: 'title',
```
anuraghazra, line 50:
```suggestion
    componentName: 'display',
```

anuraghazra (approved):
Mostly looks good.

packages/blade/src/components/Typography/Display/Display.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+  title: 'Components/Typography/Display',
+  component: DisplayComponent,
+  args: {
+    size: 'small',
+    type: 'normal',
+    children: 'Power your finance, grow your business',
+    contrast: 'low',
+    as: undefined,
+  },
+  argTypes: {
```
kamleshchandnani, line 55: once you select the color prop, it doesn't have any effect if you try to change the type prop. try selecting the color as `primary.300` for example and then try to change the `type` prop

packages/blade/src/components/Typography/Display/__tests__/Display.web.test.tsx
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
```
kamleshchandnani, line 1: add ssr test

packages/blade/src/components/Typography/Display/Display.stories.tsx
snitin315: @kamleshchandnani I believe this is intentional for all typography elements as we have these type tokens for surface text only. Otherwise, it's a bug which we will need to fix for all typography elements.

Screenshot 2023-09-04 at 7 16 25 AM

packages/blade/src/components/Typography/Display/__tests__/Display.web.test.tsx
snitin315: We need to add this for other typography components too, I can add them in a separate PR.

packages/blade/src/components/Typography/Display/Display.stories.tsx
snitin315: This makes sense. @kamleshchandnani I've updated it to be `h1` always. If `Title large` and `Display small` exists on the same screen, the Display should not be `h3`.

anuraghazra (approved):
Mostly looks good, hope you've checked it on RN once how it looks.

packages/blade/src/components/Typography/Display/Display.tsx
```diff
@@ -47,28 +47,25 @@ const getProps = ({
...
     accessibilityProps: isPlatformWeb ? {} : { role: 'heading' },
-    componentName: 'title',
+    componentName: 'display',
     testID,
   };
 
   if (size === 'small') {
     props.fontSize = 1100;
     props.lineHeight = 900;
-    props.as = isPlatformWeb ? 'h3' : undefined;
```
kamleshchandnani: why did we remove this?

packages/blade/src/components/Typography/BaseText/types.ts
```diff
@@ -64,7 +64,7 @@ export type BaseTextProps = {
    * React Native only
    */
   numberOfLines?: number;
-  componentName?: 'base-text' | 'text' | 'title' | 'heading' | 'code';
```
kamleshchandnani, line 67: @snitin315 can you check with design that do we need to update this in our amount components? because amount component is basically permutations and combinations of all the typography elements

packages/blade/src/components/Typography/Display/Display.stories.tsx
snitin315: Yeah. This is to make sure the options are sorted by size.

kamleshchandnani (approved):

PR #1599, opened 2023-09-01
fix(blade): firefox carousel fix

### LINKED ISSUE #1598: Carousel: onChange event is being triggered >1 times if the arrow buttons are clicked
https://razorpay.slack.com/archives/CMQ3RBHEU/p1693559846715769

anuraghazra:
fixes #1598 

Before: 

https://razorpay.slack.com/archives/CMQ3RBHEU/p1693801064286809?thread_ts=1693559846.715769&cid=CMQ3RBHEU

After: 
https://razorpay.slack.com/archives/CMQ3RBHEU/p1693819528345219?thread_ts=1693559846.715769&cid=CMQ3RBHEU

So there was two bugs: 
1. In blade: where in firefox on initial render sometimes it will try to mount carousel items with a smooth scrolling transition because we added `scroll-behaviour: smooth`, this caused the `shouldAddStartEndSpacing` calculation to start from the wrong position. 
3. in website: inside carousel website team added the Media component which loads the image but because it didn't had width set it was causing layoutshifts and in turn messing with the scrolling of the carousel items


packages/blade/src/components/Carousel/Carousel.web.tsx
```diff
@@ -400,7 +402,7 @@ const Carousel = ({
 
   const carouselContext = React.useMemo<CarouselContextProps>(() => {
     return {
-      visibleItems,
+      visibleItems: _visibleItems,
```
anuraghazra, line 405: fixes a bug where when you add `<Carousel visibleItems={2} />` and open the component on mobile device it still stays as visibleItems, but instead on mobile we should automatically change to use 1 visibleItems.

packages/blade/src/components/Carousel/Carousel.native.tsx
```diff
@@ -154,7 +155,7 @@ const Carousel = ({
     },
   );
 
-  React.useEffect(() => {
+  useUpdateEffect(() => {
```
chaitanyadeorukhkar, line 158: Why wasn't just useEffect working here?

chaitanyadeorukhkar, line 158: Rather call it - `useFirstRenderEffect()`?

anuraghazra: No it's not first render effect. It skips first render and only runs un subsequent renders thus called "update effect". 

"useUpdateEffect" is a very commonly used name for this hook throughout different libraries.

chaitanyadeorukhkar: I thought you're checking first mount 🤦

packages/blade/src/utils/useUpdateEffect.ts
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+const useUpdateEffect: typeof React.useEffect = (effect, deps) => {
```
chaitanyadeorukhkar, line 3: Add some comments here on what this does

packages/blade/src/utils/useUpdateEffect.ts
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
chaitanyadeorukhkar, line 1: Add tests

chaitanyadeorukhkar: Add tests and comments in subsequent PR if needed since its a P0

packages/blade/src/utils/useUpdateEffect.ts
anuraghazra: Yep just added! 👍

packages/blade/src/utils/useUpdateEffect.ts
anuraghazra: added, just realized most of our hooks don't have tests, we should add them.

anuraghazra:
- Ah just realized we already had that useUpdateEffect, it was called useDidUpdate. so removed it and added tests for useDidUpdate. 

- Also had to make few changes to fix that visibleItems bug, it was interfering with the autofit variant.

divyanshu013:
I'd added `useDidUpdate` util but we scoped it inside `Collapsible`. It also has tests, you can probably use the same.

anuraghazra:
🤯 Carousel also had useDidUpdate? 😆 let me make sure only 1 is there

packages/blade/src/utils/useDidUpdate.web.test.ts
```diff
@@ -18,4 +18,18 @@ describe('useDidUpdate', () => {
     // Should be called after first render, anytime dependencies change
     expect(effectCallback).toHaveBeenCalledTimes(1);
   });
+
+  it('should not run the effect when the dependencies do not change', () => {
```
kamleshchandnani, line 22: can you add ssr test to make sure nothing gets called on server?

packages/blade/src/components/Carousel/Indicators/Indicators.tsx
```diff
@@ -14,8 +14,7 @@ const Indicators = (props: IndicatorsProps): React.ReactElement => {
       {...makeAccessible({ role: 'tablist', label: 'Slides' })}
     >
       {new Array(props.totalItems).fill(0).map((_, idx) => {
-        const isResponsive = visibleItems === 'autofit';
-        let _visibleItems = visibleItems as 1 | 2 | 3;
+        let _visibleItems = visibleItems!;
```
kamleshchandnani, line 17: didn't quite understand why are are now getting isresponsive from context? can you elaborate?

packages/blade/src/utils/useDidUpdate.web.test.ts
anuraghazra: it's impossible to get a call on server, since useEffect doesn't run there.

packages/blade/src/components/Carousel/Indicators/Indicators.tsx
anuraghazra: Previously the isResponsive was getting derived from `visibleItems === 'autofit'` which was causing issues because this check will always fail on mobile since on the main Carousel.web file we check for `isMobile` and set visibleItems to `1`.

To fix this I separated isResponsive & visibleItems to different values to reduce ambiguity.

packages/blade/src/utils/useDidUpdate.web.test.ts
kamleshchandnani: yes so your test shall have toNotHaveBeenCalled() to true right?

anuraghazra: Added.

kamleshchandnani (approved):

Issue #1611, opened 2023-09-08
Add chromatic snapshot tests for ActionList


saurabhdaware:
Current Dropdown snapshots don't take it's ActionList into consideration because its hidden in the initial snapshots. We should have tests of ActionList itself to avoid issues like #1610

harshalkaigaonkar:
Hey @saurabhdaware , i can take this up, if it's still open..

ashutosh887:
@saurabhdaware I think this is merged already

saurabhdaware:
@ashutosh887 no we just have ActionList documentation. We have snapshots on Dropdown component but they are taken with Dropdown closed so don't cover ActionList as part of it.

We have to create a ActionList.stories.tsx and render multiple ActionList combinations there and then create a Kitchen Sink story where we read those stories and render them on one page for snapshots

KitchenSink stories are here - https://blade.razorpay.com/?path=/docs/components-kitchensink-accordion--accordion

saurabhdaware:
Yup you can take this @harshalkaigaonkar. Thanks 🥳

harshalkaigaonkar:
@saurabhdaware , anything specific needed here for combinations, some kind of list?

PR #1612, opened 2023-09-08
fix(ActionList): last section divider

### LINKED ISSUE #1610: ActionListSection has unwanted divider at the end
ActionListSection has an unwanted divider at the end.
Screenshot 2023-09-07 at 5 10 46 PM

saurabhdaware:
Resolves #1610 

Adding role separator wouldn't work on Divider since divider component doesn't accept the role property. This made the css selector for last actionlist section divider invalid. There was also additional div around divider.

https://github.com/razorpay/blade/blob/2fd09500feefd32a6237be48e88c3874f4f588c6/packages/blade/src/components/ActionList/styles/StyledListBoxWrapper.web.tsx#L12

```js
<Divider
  {...makeAccessible({ role: 'separator' })
/>
```

I have added `role="separator"` to Divider component by default since that is closer to the `<hr />` behaviour https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/separator_role

Also removed ActionListSectionDivider component as we don't need that anymore. I'll maybe add some snapshot tests later to test out this scenario. Created issue for that - #1611 (just finishing up AutoComplete right now so dont want to spend too much time on this task)

You can try the change on - InternalSectionListPerformance story


packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -139,11 +125,7 @@ const _ActionListSection = ({
       >
         {children}
       </BaseBox>
-      {_hideDivider && isReactNative() ? null : (
-        <Box marginX="spacing.3" marginY="spacing.1">
-          <ActionListSectionDivider />
-        </Box>
-      )}
+      {_hideDivider && isReactNative() ? null : <Divider marginX="spacing.3" marginY="spacing.1" />}
```
anuraghazra, line 128: 🤔 how did this fix the bug? I can see no logic changes.

packages/blade/src/components/Card/__tests__/__snapshots__/Card.native.test.tsx.snap
```diff
@@ -1092,6 +1093,7 @@ exports[`<Card /> should render a Card with Header 1`] = `
           </View>
         </View>
         <View
+          accessible={true}
```
anuraghazra, line 1096: How is Card's snapshot updating 🤔

anuraghazra: Also this should not be accessible={true}

packages/blade/src/components/Divider/Divider.tsx
```diff
@@ -94,6 +95,9 @@ const Divider = ({
       {...borderColor}
       {...metaAttribute({ name: MetaConstants.Divider, testID })}
       {...getStyledProps(styledProps)}
+      {...makeAccessible({
```
anuraghazra, line 98: I see now why we are getting accessible={true} 

Let's explicitly override this prop in RN to be false. Otherise dividers will unnecessarily receive focus.

anuraghazra (changes requested):
Mostly looks good, minor questions.

packages/blade/src/components/Divider/Divider.tsx
saurabhdaware: ahh got it. I removed the makeAccessible call itself from react native.

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: this didn't fix the bug. This is just refactor to remove additional wrapper and nesting of divider. Adding `role="separator"` fixed the bug. More info in PR description

anuraghazra (approved):
Looks good to me.

saurabhdaware:
I'll merge this on L1 since its a P0. @snitin315 let me know if you have any comments, I'll take them up as followup PR.

snitin315 (commented):
LGTM

PR #1613, opened 2023-09-08
feat(AutoComplete): add labelPosition

saurabhdaware:
Adds `labelPosition: inside-input` for AutoComplete and SelectInput


packages/blade/src/components/Input/BaseInput/BaseInputTagSlot.native.tsx
```diff
@@ -52,6 +53,26 @@ const ScrollableTagSlotContainer = ({
...
+  children,
+  handleOnClick,
+}: {
+  children: StringChildrenType;
+  handleOnClick: BaseInputTagSlotProps['handleOnClick'];
+}): React.ReactElement => {
+  return (
+    <TouchableWithoutFeedback
+      onPress={() => {
+        handleOnClick?.({ name: '', value: '' });
```
chaitanyadeorukhkar, line 66: Why is this going as blank? Shouldn't we send nothing instead?

chaitanyadeorukhkar (approved):

packages/blade/src/components/Input/BaseInput/BaseInputTagSlot.native.tsx
saurabhdaware: let me see if we can actually send the correct values here. Reliably sending value might be tricky since in AutoComplete and Select, inputs in themselves don't know what value is. We should be able to send the correct name though. 

I am maintaining a todo list of these task. Will create a followup PR with fix for all this

PR #1615, opened 2023-09-08
fix(blade): carousel bug in safari

### LINKED ISSUE #1617: Carousel not working on safari browser
Raised by @lakshay-dutt in https://razorpay.slack.com/archives/CMQ3RBHEU/p1694155233398949?thread_ts=1693559846.715769&cid=CMQ3RBHEU

anuraghazra:
Fixes #1617 

### What happened? 

Turns out that safari doesn't include the margin right of carousel item into consideration when calculating the `scrollWidth`, thus safari wasn't able to scroll to the last element since it didn't even had enough space to scroll. 

Run the code snippet in this SO question to understand whats happening: 
https://stackoverflow.com/questions/75509058/safari-does-not-include-margins-to-the-scroll-width

### Fix
I could not use the :after trick mentioned in the SO, because we are already using :after :before for scroll overlays, to fix the bug I added an extra `div` at the end of the carousel with the width of the available space. 

image

### DIFF
--- a/.changeset/metal-years-smash.md
+++ b/.changeset/metal-years-smash.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix(blade): carousel bug in safari
--- a/.changeset/new-balloons-smash.md
+++ b/.changeset/new-balloons-smash.md
@@ -1,5 +1,0 @@
----
-"@razorpay/blade": minor
----
-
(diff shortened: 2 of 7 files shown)


chaitanyadeorukhkar (approved):

anuraghazra:
@chaitanyadeorukhkar please also test on your safari mobile.

kamleshchandnani (approved):

PR #1616, opened 2023-09-08
chore(blade): added id to box

anuraghazra:
Added ID prop to box.

### DIFF
--- a/.changeset/green-lobsters-rule.md
+++ b/.changeset/green-lobsters-rule.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+chore(blade): added id to Box component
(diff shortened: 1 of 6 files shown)


.changeset/green-lobsters-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+chore(blade): added id to Box component
```
snitin315, line 5: It's a feature. Should be minor because we are adding a new prop.

packages/blade/src/components/Box/__tests__/Box.web.test.tsx
```diff
@@ -194,4 +194,11 @@ describe('<Box />', () => {
     await user.unhover(getByText('Hoverable Text'));
     expect(onMouseLeaveCallback).toBeCalled();
   });
+
+  it('should accept id prop', () => {
+    const text = 'Box id';
```
snitin315, line 199: Add in native as well?

anuraghazra: added

saurabhdaware (approved):

PR #1630, opened 2023-09-13
chore: white label support docs tests

chaitanyadeorukhkar:
- Adds documentation for `createTheme`
- Adds Theme Playground
- Adds tests for `createTheme`
- Deprecates `overrideTheme`

### [Demo Link](https://61c19ee8d3d282003ac1d81c-uqvvpjbzmh.chromatic.com/?path=/docs/guides-themeing-playground--page)


packages/blade/.storybook/react/global.css
```diff
@@ -0,0 +1,4 @@
+* {
+  /* Need this to ensure mdx stories don't break visually*/
+  box-sizing: border-box;
```
chaitanyadeorukhkar, line 3: Storybook changes this randomly, have to set it so that this doesn't keep changing

packages/blade/src/components/Carousel/Carousel.stories.tsx
```diff
@@ -281,8 +281,8 @@ const CarouselTestimonialTemplate: ComponentStory<typeof CarouselComponent> = (p
   );
 };
 
-export const Default = CarouselTestimonialTemplate.bind({});
-Default.args = {
+export const DefaultCarousel = CarouselTestimonialTemplate.bind({});
```
chaitanyadeorukhkar, line 284: Changin this so that "Default" storyName isn't the same for multiple components. Example: Checkbox could have an export Default as well as Carousel so when we compose all the stories together, its hard to know which "Default" storyName is associated with which component. Having this separation helps in mapping storyName to component name in `BrandedComponentKitchenSink.tsx`

packages/blade/tsconfig-generate-types.native.json
```diff
@@ -9,7 +9,8 @@
     "**/**.stories.tsx",
     "**/**.stories.internal.tsx",
     "**/**.web.ts",
-    "**/**.web.tsx"
+    "**/**.web.tsx",
+    "**/ThemePlayground/**.tsx"
```
chaitanyadeorukhkar, line 13: ThemePlayground import from stories, stories are already ignored but because we don't ignore ThemePlayground, it starts typechecking stories too

packages/blade/.storybook/react/preview.js
```diff
@@ -48,7 +49,6 @@ export const parameters = {
           'makeSpace',
           'makeTypographySize',
           'useTheme',
-          'overrideTheme',
```
kamleshchandnani, line 51: can you add the Theming here? right now it shows in guides but we want it to be in `utils`

Also rename Theming ➡️ Theming(White-Label)

so the structure can be
```
utils/
  	Theming(White-Label)/
			createTheme Function/
			Theme Playground/
```

packages/blade/src/storybook-recipes/ThemePlayground/BrandedComponentKitchenSink.tsx
```diff
@@ -0,0 +1,94 @@
...
+  SimpleModal: 'Modal',
+  ProgressBarIndeterminate: 'ProgressBar',
+  RequiredRadio: 'Radio',
+  DefaultCheckedSwitch: 'Switch',
+  TextInputHelpText: 'TextInput',
+  OTPInputHelpText: 'OTPInput',
+  InternalAutoCompleteControlled: 'AutoComplete',
+};
+
+const allStories = [];
```
kamleshchandnani, line 55: nit: why do you need to call push? you can directly add values to `allStories` array?

```
const allStories = [
...Object.values(composeStories(buttonStories))
]
```
like this?

packages/blade/src/tokens/theme/createTheme.ts
```diff
@@ -347,6 +347,7 @@ export const createTheme = ({ brandColor }: { brandColor: ColorInput }): ThemeTo
   const brandedThemeTokens = overrideTheme({
     baseThemeTokens: paymentTheme,
     overrides: {
+      name: `custom-${tinycolor(brandColor).toHex()}`,
```
kamleshchandnani, line 350: ideally we shall only allow hex values because hsl to hex doesn't give you the same color always there are discrepancies sometimes. but we can ignore for that but add this somewhere in code so we can come back later

kamleshchandnani (approved):
tiny comments, overall lgtm

packages/blade/src/tokens/theme/createTheme.ts
chaitanyadeorukhkar: Our backend sends different formats 😢

PR #1631, opened 2023-09-13
feat(blade): add ref to OTP Input

### LINKED ISSUE #1633: Not able to focus OtpInput (controlled)
Use Case: OtpInput is disabled on first load. once otp is sent successfully, otp field gets enabled and task is it should gain focus.
I tried using autofocus prop, but it did not work. also "ref" is not allowed in props.

Please allow "ref" prop for each input in OtpInput field.
https://razorpay.slack.com/archives/CMQ3RBHEU/p1694589493069019

anuraghazra:
fixes #1633

### DIFF
--- a/.changeset/shy-ears-cry.md
+++ b/.changeset/shy-ears-cry.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+feat(blade): add ref to OTP Input
--- a/.changeset/shy-ears-cry.md
+++ b/.changeset/shy-ears-cry.md
@@ -1,5 +1,5 @@
 ---
-"@razorpay/blade": patch
+"@razorpay/blade": minor
 ---
 
 feat(blade): add ref to OTP Input
--- a/packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
+++ b/packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
@@ -6,6 +6,8 @@ import { OTPInput as OTPInputComponent } from './OTPInput';
 import { Sandbox } from '~utils/storybook/Sandbox';
 import StoryPageWrapper from '~utils/storybook/StoryPageWrapper';
 import { getStyledPropsArgTypes } from '~components/Box/BaseBox/storybookArgTypes';
+import { Box } from '~components/Box';
+import { Button } from '~components/Button';
 
 const propsCategory = {
   BASE_PROPS: 'OTPInput Props',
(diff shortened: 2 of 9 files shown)


packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
```diff
@@ -243,3 +245,37 @@ const OTPInputControlledTemplate: ComponentStory<typeof OTPInputComponent> = ()
...
+    </Box>
+  );
+};
+
+inputRef.storyName = 'OTP Input Ref';
+inputRef.parameters = {
+  docs: {
+    description: {
+      story:
+        'OTP component exposes the `ref` prop. You can use the ref to programatically focus the input. The OTPs ref prop exposes an array of refs for each input field. This will enable you to focus on any input field via `inputRef.current[index].focus()`',
```
snitin315, line 278:
```suggestion
        'The OTP component offers a `ref` prop for programmatically focusing on its input fields. This prop exposes an array of individual refs for each input, allowing you to focus on a particular field using `inputRef.current[index].focus()`.',
```

packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
```diff
@@ -243,3 +245,37 @@ const OTPInputControlledTemplate: ComponentStory<typeof OTPInputComponent> = ()
...
+        onChange={({ name, value }): void => console.log({ name, value })}
+      />
+      <Button
+        onClick={() => {
+          console.log(inputRef);
+          inputRef?.current[1].focus();
+        }}
+      >
+        Click to focus the OTPInput
+      </Button>
```
snitin315, line 268: For more interactivity, I feel we can make this a dropdown and provide options as Focus 1st input, Focus 2nd input, ...

snitin315 (commented):
Can we add tests?

packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
anuraghazra: Focusing after dropdown state change might cause issues since dropdown's focus will try to go to the target and we will try to move the focus to the OTP input.

anuraghazra: Okay, added a dropdown with an explicit button to focus on.

anuraghazra:
Added tests

chaitanyadeorukhkar (dismissed):

snitin315 (dismissed):
Need to fix lint error, otherwise LGTM.

.changeset/shy-ears-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": minor
+---
+
+feat(blade): add ref to OTP Input
```
snitin315, line 5: Maybe we should link to the ref story here.

chaitanyadeorukhkar (approved):

PR #1632, opened 2023-09-13
feat(AutoComplete): Comment Resolutions, Snapshot Updates, Chores

saurabhdaware:
Additional comment resolutions, snapshots updates, and chores


packages/blade/ios/Podfile.lock
```diff
@@ -673,4 +673,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: d221e9c85520c3a61813f4bdbd8daf13d66d4f59
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.12.1
```
saurabhdaware, line 676: reverts to master's version of Podfile.lock

saurabhdaware:
One test that is failing is failing due to different id generated on component. It happens when there is some change in master that is not pulled in this. It should get fixed when we merge this to integration branch and pull master on integration branch

packages/blade/src/components/Input/DropdownInputTriggers/__tests__/AutoComplete.ssr.test.tsx
```diff
@@ -3,7 +3,7 @@ import userEvent from '@testing-library/user-event';
 import { waitFor } from '@testing-library/react';
 import { Dropdown, DropdownOverlay } from '~components/Dropdown';
 import { DropdownFooter, DropdownHeader } from '~components/Dropdown/DropdownHeaderFooter';
-import renderWithSSR from '~utils/testing/renderWithSSR';
+import renderWithSSR from '~utils/testing/renderWithSSR.web';
```
chaitanyadeorukhkar, line 6: Why .web? It should take the right extension anyway right?

chaitanyadeorukhkar (approved):

packages/blade/src/components/Input/DropdownInputTriggers/__tests__/AutoComplete.ssr.test.tsx
saurabhdaware: nope it seems to fail on ssr storybook. This is how we seem to do in all ssr files.

saurabhdaware:
@chaitanyadeorukhkar Fixed for native as well

https://github.com/razorpay/blade/assets/30949385/1ffd06d9-f738-46ac-9c48-a537c2a26fe8

chaitanyadeorukhkar (approved):

Issue #1386, opened 2023-07-03
Tooltip component is overriding the aria-label of trigger element


anuraghazra:
> Reference: https://razorpay.slack.com/archives/CMQ3RBHEU/p1688128540647929
> Demo: https://codesandbox.io/s/infallible-matan-yc7stq?file=/App.tsx

We are redundantly passing aria-label to the trigger element which is overriding the computed aria-label of the Input, it should work as is with aria-describedby only. 

(Need to also see how it will work with RN)

archie252000:
@anuraghazra Can you please explain the desired behaviour? Should ```...makeAccessible({ label: content })``` not override the ```aira-label``` if the trigger element has it already?

PR #1635, opened 2023-09-15
fix(Tooltip): prevent overriding the `aria-label` attribute

snitin315:
Fixes https://github.com/razorpay/blade/issues/1386

The issue exists on the web only because we don't set the `accessibilityLabel` prop to trigger on react-native.

https://github.com/razorpay/blade/blob/7c116d54164124651fe420a93f182fb4caf49cf9/packages/blade/src/components/Tooltip/Tooltip.native.tsx#L72-L80

### DIFF
--- a/packages/blade/src/components/Tooltip/Tooltip.web.tsx
+++ b/packages/blade/src/components/Tooltip/Tooltip.web.tsx
@@ -98,7 +98,7 @@ const Tooltip = ({
       {React.cloneElement(children, {
         ref: refs.setReference,
         ...mergeProps(children.props, getReferenceProps()),
-        ...makeAccessible({ label: content }),
+        ...makeAccessible({ describedBy: content }),
       })}
       {isMounted && (
         <FloatingPortal>
--- a/packages/blade/src/components/Tooltip/Tooltip.web.tsx
+++ b/packages/blade/src/components/Tooltip/Tooltip.web.tsx
@@ -97,8 +97,8 @@ const Tooltip = ({
       {/* Cloning the trigger children to enhance it with ref and event handler */}
       {React.cloneElement(children, {
         ref: refs.setReference,
+        ...makeAccessible({ label: content }),
         ...mergeProps(children.props, getReferenceProps()),
-        ...makeAccessible({ describedBy: content }),
       })}
       {isMounted && (
         <FloatingPortal>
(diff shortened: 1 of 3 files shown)


anuraghazra (approved):

packages/blade/src/components/Tooltip/__tests__/Tooltip.web.test.tsx
```diff
@@ -53,7 +53,7 @@ describe('<Tooltip />', () => {
 
   it('should render tooltip with custom zIndex', () => {
     const buttonText = 'Hover me';
-    const { container, getByRole, queryByRole } = renderWithTheme(
+    const { baseElement, getByRole, queryByRole } = renderWithTheme(
```
saurabhdaware, line 56: expected change?

packages/blade/src/components/Tooltip/Tooltip.web.tsx
```diff
@@ -98,7 +98,7 @@ const Tooltip = ({
       {React.cloneElement(children, {
         ref: refs.setReference,
         ...mergeProps(children.props, getReferenceProps()),
-        ...makeAccessible({ label: content }),
+        ...makeAccessible({ describedBy: content }),
```
saurabhdaware, line 101: did you mean `description` here? `describedby` expects id of the description no?

Also why this change?

image

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby#example

packages/blade/src/components/Tooltip/__tests__/__snapshots__/Tooltip.web.test.tsx.snap
```diff
@@ -175,7 +186,7 @@ exports[`<Tooltip /> should render 1`] = `
 <body>
   <div>
     <button
-      aria-describedby=":r0:"
+      aria-describedby="Hello world"
```
saurabhdaware, line 189: `:r0:` would be expected no?

packages/blade/src/components/Tooltip/__tests__/Tooltip.web.test.tsx
snitin315: Yes, we want to snapshot the open tooltip and not the trigger.

packages/blade/src/components/Tooltip/Tooltip.web.tsx
anuraghazra: Oh right, it should not be the content

snitin315: Yeah, I updated the logic not to override the aria-label if provided by the user. We're not adding `aria-description` as [recommended in the docs.](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description)

@anuraghazra @saurabhdaware

anuraghazra (approved):

saurabhdaware (approved):

PR #1636, opened 2023-09-15
feat(Dropdown): add isOpen and onOpenChange

### LINKED ISSUE #1416: Unable to close Multi-select dropdown with footer button

### LINKED ISSUE #1436: Expose `isOpen` on Dropdown

saurabhdaware:
PR to add `isOpen` and `onOpenChange`. 

Extends on top of #1485 

Fixes #1436 
Fixes #1416


packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -88,22 +90,28 @@ const _Dropdown = ({
...
     }
 
-    if (!isOpen && onDismiss) {
-      onDismiss();
+    onOpenChange?.(isOpen);
+    if (!isOpen) {
+      onDismiss?.();
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isOpen]);
```
anuraghazra, line 104: We can use the onDidUpdate hook instead?

saurabhdaware: aha nice! didn't know about this hook. Added

packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
```diff
@@ -533,11 +552,29 @@ describe('<Dropdown />', () => {
 describe('<Dropdown /> with <DropdownButton />', () => {
   // Skipping this test because the id that `useId` generates seems to be different and flaky between local and CI.
   // Have to figure out solution to that and then enable this again.
-  it.skip('should render menu and make items clickable', async () => {
```
saurabhdaware, line 536: Unrelated change: I broke this test into 2 parts. The snapshot and non-snapshot assertions.

This will allow us to do non-snapshot assertions at least while the snapshots are being skipped because of flaky id generation

packages/blade/src/components/Dropdown/types.ts
```diff
@@ -1,8 +1,40 @@
+import type React from 'react';
 import type { StyledPropsBlade } from '~components/Box/styledProps';
 import type { TestID } from '~utils/types';
 
 type DropdownProps = {
+  /**
+   * Control open / close state of the Dropdown component
+   *
+   * Make sure to not misuse this prop to create custom triggers. Use DropdownButton instead for such usecases.
```
kamleshchandnani, line 9: probably remove this line `Make sure to not misuse this prop to create custom triggers`? because somethings are better being ambiguous than explicit :D

kamleshchandnani (dismissed):
LGTM 🚀

packages/blade/src/components/Dropdown/types.ts
saurabhdaware: haha yes someone will read this and think: "Woahhh we can do that?? 😆"

anuraghazra (approved):

PR #1639, opened 2023-09-15
feat(ActionListItem): add ActionListItemBadge and titleSuffix option

saurabhdaware:
Adds-
- ActionListItemBadge
- ActionListItemBadgeGroup
- `titleSuffix` prop on ActionListItem
- Documentation for it


packages/blade/src/utils/storybook/Sandbox/Sandbox.web.tsx
```diff
@@ -21,7 +21,7 @@ import { Button } from '~components/Button';
 const getBladeVersion = (): string => {
   // We don't publish codesandbox ci on master so version is not present
   const isMaster = process.env.GITHUB_REF === 'refs/heads/master';
-  const sha = process.env.GITHUB_SHA ?? '6f76c2e1d36a5f0076ee76470b36cd54b7d4801b'; // @todo: for development, remove before merging
```
saurabhdaware, line 24: And I, in fact, did not remove it before merging 😆

packages/blade/src/components/ActionList/actionListUtils.ts
```diff
@@ -196,6 +198,18 @@ const validateActionListItemProps = ({
       }
     });
 
+    React.Children.map(titleSuffix, (child) => {
+      if (
+        !isValidAllowedChildren(child, componentIds.ActionListItemBadge) &&
+        !isValidAllowedChildren(child, componentIds.ActionListItemBadgeGroup)
+      ) {
+        throwBladeError({
+          message: `Only ${componentIds.ActionListItemBadge} and ${componentIds.ActionListItemBadgeGroup} are allowed in trailing prop`,
```
kamleshchandnani, line 207:
```suggestion
          message: `Only ${componentIds.ActionListItemBadge} and ${componentIds.ActionListItemBadgeGroup} are allowed in titleSuffix prop`,
```

packages/blade/src/components/ActionList/docs/stories.ts
```diff
@@ -83,15 +83,23 @@ const ActionListItem = `
...
 
   function App(): React.ReactElement {
     return (
-    <Box>
+    <Box backgroundColor="surface.background.level2.lowContrast">
       <ActionList>
         <ActionListItem 
           title="Title"
+          titleSuffix={(
+            <ActionListItemBadgeGroup>
```
kamleshchandnani, line 98: can we make this example more contextual with the use case where this was raised? something with `in: something` IIRC?

packages/blade/src/components/ActionList/docs/stories.ts
```diff
@@ -83,15 +83,23 @@ const ActionListItem = `
...
+              <ActionListItemBadge>as: Keyword</ActionListItemBadge>
```
kamleshchandnani, line 99: can we have better stories like "With Badges for items" something on those lines? we can have all the components of ActionList with their props table listed and then different examples which are use cases centric but full examples instead of just individual sub-component examples? so that way it'll be easier for our consumers to scan

kamleshchandnani (commented):
overall LGTM, tiny documentation changes needed

kamleshchandnani:
tests are in different PR? because the builds here are failing

packages/blade/src/components/ActionList/docs/stories.ts
saurabhdaware: I updated our AutoComplete controlled filtering example to use badges since it is mostly used in controlled filtering context to search on keywords

image

Regarding ActionList components with tables, it already exists - https://61c19ee8d3d282003ac1d81c-lqbhyaeney.chromatic.com/?path=/docs/components-dropdown-actionlist--page

I added link to that AutoComplete filtering example with badges in ActionListItemBadge section

packages/blade/src/components/ActionList/docs/stories.ts
saurabhdaware: This one is just individual action list item example to show all capabilities so it has to be vague cause there is no context around.
image

Added more contextual example here - https://github.com/razorpay/blade/pull/1639#discussion_r1328477137

kamleshchandnani (approved):

packages/blade/src/utils/storybook/Sandbox/Sandbox.web.tsx
anuraghazra: Was it causing any issues?

packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -154,6 +164,30 @@ const ActionListItemIcon = assignWithoutSideEffects(_ActionListItemIcon, {
...
+    </Box>
+  );
+};
+
+const ActionListItemBadgeGroup = assignWithoutSideEffects(_ActionListItemBadgeGroup, {
+  componentId: componentIds.ActionListItemBadgeGroup,
+});
+
+const _ActionListItemBadge = (props: BadgeProps): React.ReactElement => {
+  return <Badge size="medium" marginLeft="spacing.3" {...props} />;
```
anuraghazra, line 184: is badge size controlled by us or can be overriden by consumer? 

If we don't intend to let consumer override let's put size at the end.

anuraghazra (approved):

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: It will be consistent on design mostly. They can override though, its not breaking the UI so much. We just set the default size on our end.

PR #1640, opened 2023-09-15
fix(AutoComplete): Storybook panel and React Native Icon Alignment

saurabhdaware:
## Storybook Docs Page Fix

I remember why I had kept SelectInput separate 😆 So we can only define 1 meta info in 1 story. In WithSelectInput I want to disable docs panel but in SelectInput we want to keep docs panel to show arguments. When we keep both stories on same level, they seem to clash and this is how it is showing up when you open WithAutoComplete story on master 
https://blade.razorpay.com/?path=/docs/components-dropdown-with-autocomplete--with-single-select

image

We have to nest it inside /Stories or keep it separate for it to work properly 😢 I nested it inside stories for now. Anyway the main Dropdown documentation links to these prop tables

## React Native Icon Alignment Fix

The trailing icon on TextInput and PasswordInput React Native got aligned to top. This PR fixes that issue by defining alignSelf: center in code

https://github.com/razorpay/blade/pull/1640/files?file-filters%5B%5D=.md&file-filters%5B%5D=.tsx&show-viewed-files=true#r1327058653


packages/blade/src/components/Input/BaseInput/BaseInputVisuals.tsx
```diff
@@ -12,11 +12,12 @@ type InputVisuals = Pick<
 
 const getVisualContainerStyles = (): Pick<
   BaseBoxProps,
-  'display' | 'flexDirection' | 'alignItems'
+  'display' | 'flexDirection' | 'alignItems' | 'alignSelf'
 > => ({
   display: 'flex',
   flexDirection: 'row',
   alignItems: 'center',
+  alignSelf: 'center',
```
saurabhdaware, line 20: aligns the trailing icons to center like before. The trailing container's size was made full height to make chevron icon clickable from anywhere. So the elements inside of it got top-aligned. By setting alignSelf: center here, we make them center again while keeping height full.
 
image

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.stories.tsx
```diff
@@ -25,7 +25,7 @@ const propsCategory = {
 };
 
 export default {
-  title: 'Components/Dropdown/With AutoComplete',
+  title: 'Components/Dropdown/With AutoComplete/Stories',
```
chaitanyadeorukhkar, line 28: Its hard to discover because its called "Stories". Can we rename the nested folder itself to "Props Playground"? Would be easier to discover

```suggestion
  title: 'Components/Dropdown/With AutoComplete/Props Playground',
```

saurabhdaware: best!

saurabhdaware: done

chaitanyadeorukhkar (approved):

chaitanyadeorukhkar (approved):

PR #1647, opened 2023-09-18
docs: improve documentation aesthetics

### LINKED ISSUE #1255: docs: Add Blade logo
I was searching for the Blade logo and couldn't find it anywhere in the docs.

Should be there on the Intro page ideally:
https://blade.razorpay.com/?path=/docs/guides-intro--page

saurabhdaware:
Resolves #1255

- Make Documentation Look Better
  image

- Make README Look Better
  image

- Fixes Code Highlighting Issues
  image


README.md
```diff
@@ -1,39 +1,41 @@
...
-<p align="center">
+  <a href="https://npmjs.org/package/@razorpay/blade"><img alt="Blade Latest Version" src="https://img.shields.io/github/package-json/v/razorpay/blade?style=for-the-badge&labelColor=322&logo=npm&label=@razorpay/Blade&color=darkred&filename=packages%2Fblade%2Fpackage.json"></a> &nbsp;<a href="https://blade.razorpay.com/"><img alt="Documentation blade.razorpay.com" src="https://img.shields.io/badge/Documentation-blade.razorpay.com-0648EF?style=for-the-badge&labelColor=0012AD&logo=readthedocs&logoColor=eee"/></a> &nbsp;<a href="https://github.com/razorpay/blade/tree/master/CONTRIBUTING.md"><img alt="Discord Join Chat" src="https://img.shields.io/badge/Contributions-Open-333333?style=for-the-badge&logo=github&logoColor=ffffff&labelColor=111111"/></a></p>
+
+<h1 aria-hidden="true"></h1>
 
-### [`@razorpay/blade`](https://github.com/razorpay/blade/tree/master/packages/blade)
+<br/>
 
-This package is under active development.
+Blade is a Design System that powers [Razorpay](https://razorpay.com/).
```
snitin315, line 19:
```suggestion
Blade is the Design System that powers [Razorpay](https://razorpay.com/). 
```

packages/blade/docs/guides/Intro.stories.mdx
```diff
@@ -2,36 +2,46 @@ import { Meta } from '@storybook/addon-docs';
...
-      src="https://img.shields.io/badge/license-MIT-blue.svg?style=flat-square&colorA=444444"
-      alt="Blade is released under the MIT license."
+      alt="Open for Contributions"
+      src="https://img.shields.io/badge/Contributions-Open-333333?style=for-the-badge&logo=github&logoColor=ffffff&labelColor=111111"
     />
   </a>
 </p>
+<br />
+
+Blade is a Design System that powers [Razorpay](https://razorpay.com/)
```
snitin315, line 30:
```suggestion
Blade is the Design System that powers [Razorpay](https://razorpay.com/)
```

packages/blade/src/utils/storybook/StoryPageWrapper.tsx
```diff
@@ -116,6 +125,22 @@ const StoryPageWrapper = (props: StoryPageWrapperTypes): React.ReactElement => {
...
               <BaseBox id="properties-ref">
                 <Title>Properties</Title>
+                {props.apiDecisionLink === '' ? null : (
+                  <Text marginY="spacing.5">
+                    You can checkout API Decisions for {props.componentName} at{' '}
+                    <Link
+                      target="_blank"
+                      href={
+                        props.apiDecisionLink ??
+                        `https://github.com/razorpay/blade/blob/master/packages/blade/src/components/${props.componentName}/_decisions/decisions.md`
```
snitin315, line 135: So, by default we use this link, unless explicitly mentioned via prop. In case it doesn't have apiDescision we need to pass an empty string `apiDecisionLink=''`. Is that right?

packages/blade/src/components/Input/PasswordInput/PasswordInput.stories.tsx
```diff
@@ -15,6 +15,7 @@ const Page = (): ReactElement => {
   return (
     <StoryPageWrapper
       componentName="PasswordInput"
+      apiDecisionLink="https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Input/PasswordInput/_decisions/decisions.md"
```
snitin315, line 18: Currently, it opens API descision on GitHub, can we not host it on storybook itself?

packages/blade/src/utils/storybook/StoryPageWrapper.tsx
saurabhdaware: yup correct!

packages/blade/src/components/Input/PasswordInput/PasswordInput.stories.tsx
saurabhdaware: A bit difficult to do in current setup (I am not sure if it will be able to resolve locally referenced images correctly). 

Also the alignments on storybook are a mess when used with markdown or mdx so need to fix that before doing this 😢 .

packages/blade/src/utils/storybook/StoryPageWrapper.tsx
```diff
@@ -116,6 +125,22 @@ const StoryPageWrapper = (props: StoryPageWrapperTypes): React.ReactElement => {
...
+                      }
+                    >
+                      {trimmedAPIDecisionLink
+                        ? trimmedAPIDecisionLink
+                        : `/blade/src/components/${props.componentName}/_decisions/decisions.md`}
```
anuraghazra, line 140: Let's put the Component's name instead of the API decision path?

anuraghazra: eg: 

You can checkout API Decisions for [Button here](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Button/_decisions/decisions.md)

packages/blade/src/components/Button/IconButton/IconButton.stories.tsx
```diff
@@ -14,6 +14,7 @@ const Page = (): ReactElement => {
     <StoryPageWrapper
       componentName="IconButton"
       componentDescription="Useful for making clickable icons. For example - close button for modals, inputs, etc."
+      apiDecisionLink=""
```
anuraghazra, line 17: Why is this empty? Let's make it optional instead?

saurabhdaware: when we skip this prop, apiDecisionLink is auto-generated from componentName prop since our URL format is mostly constant.

This prop is to override that link. So you can override it with new link, or override it with blank string to remove that section itself.

We can't treat undefined here because undefined would just create a API decision link. Maybe I can change this to null instead.

snitin315: Maybe this will be more meaningful.
```suggestion
      apiDecisionLink="none"
```

packages/blade/src/utils/storybook/StoryPageWrapper.tsx
saurabhdaware: image

done

packages/blade/src/components/Button/IconButton/IconButton.stories.tsx
saurabhdaware: done. Added `apiDecisionLink={null}` instead

packages/blade/.storybook/react/manager-head.html
```diff
@@ -1,3 +1,59 @@
...
+
+  .sidebar-item[data-selected="true"]+a[href]:focus {
+    opacity: 1 !important;
+  }
+
+  .sidebar-item[data-selected="true"] svg {
+    color: var(--blade-surface-action-icon-active) !important;
+  }
+
+  .sto-1va4tas {
```
anuraghazra, line 56: I hope this isn't a dynamic style

anuraghazra (approved):

packages/blade/.storybook/react/manager-head.html
saurabhdaware: no I too checked. It stays static. This is how CRED has also overriden styles of their storybook.

I have anyway tried to avoid this until I couldn't so wouldn't break too many things even if it changes.

PR #1661, opened 2023-09-21
feat: radio & checkbox integration with card

anuraghazra:


packages/blade/src/components/Checkbox/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -56,7 +56,7 @@ type CheckboxGroupProps = {
   /**
    * Renders the label of the checkbox group
    */
-  label: string;
+  label?: string;
```
kamleshchandnani, line 59: i'm more concerned with the scary patterns emerging with this 😢 people not using label because they want to use a large label 😭

packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -57,10 +56,10 @@ const _Radio: React.ForwardRefRenderFunction<BladeElementRef, RadioProps> = (
 
   if (__DEV__) {
     if (!isInsideGroup) {
-      throwBladeError({
-        moduleName: 'Radio',
-        message: 'Cannot use <Radio /> outside of <RadioGroup />',
-      });
+      // throwBladeError({
```
kamleshchandnani, line 59: delete this

packages/blade/src/components/Checkbox/Checkbox.tsx
```diff
@@ -50,7 +50,7 @@ type CheckboxProps = {
   /**
    * Sets the label of the checkbox
    */
-  children: React.ReactNode;
+  children?: React.ReactNode;
```
kamleshchandnani, line 53: why is it `React.ReactNode` here and in radio `StringChildrenType`?

packages/blade/src/components/Radio/RadioGroup/RadioGroup.tsx
```diff
@@ -55,7 +55,7 @@ type RadioGroupProps = {
   /**
    * Renders the label of the radio group
    */
-  label: string;
+  label?: string;
```
kamleshchandnani, line 58: now no one can stop anyone from doing 😭 

```
<Title> My Label </Title>
<CheckboxGroup>
	<Checkbox>Option1</Checkbox>
	<Checkbox>Option2</Checkbox>
</CheckboxGroup>
```

kamleshchandnani (dismissed):
mostly lgtm but i've concerns that this will lead to chaos 😢

packages/blade/src/components/Radio/Radio.tsx
anuraghazra: Yeah don't need to delete this check actually, since we could achieve the usecase while being inside RadioGroup only.

packages/blade/src/components/Checkbox/Checkbox.tsx
anuraghazra: No idea actually, seems like a miss. But changing this might cause breaking change.

packages/blade/src/components/Radio/RadioGroup/RadioGroup.tsx
anuraghazra: Yeah can't do much, tradeoffs. 😿

kamleshchandnani (approved):

PR #1667, opened 2023-09-25
fix: change autocomplete search logic from startsWith to includes

chaitanyadeorukhkar:


saurabhdaware (approved):

saurabhdaware (commented):
I think we can go with just includes for now like [Twilio](https://paste.twilio.design/components/multiselect-combobox#usemultiselectcombobox-state-hook)

[AriaKit](https://ariakit.org/examples/combobox-filtering) seems to sort elements by startsWith and filter with includes but we can do that as enhancement later

PR #1671, opened 2023-09-25
fix(AutoComplete): controlled single selection

saurabhdaware:
AutoComplete when used in single selection controlled version, was not changing its input value on selection change.

In AutoComplete single select, the value displayed inside AutoComplete is inputValue. When we select item in controlled version, we also have to set its inputValue to the selection

### DIFF
--- a/packages/blade/src/components/Input/DropdownInputTriggers/types.ts
+++ b/packages/blade/src/components/Input/DropdownInputTriggers/types.ts
@@ -40,11 +40,9 @@ type DropdownInputTriggersCommonProps = Pick<
   defaultValue?: string | string[];
   onChange?: ({ name, values }: { name?: string; values: string[] }) => void;
   /**
-   * Handles the controlled selection change.
+   * Syncs the selected value to inputValue in AutoComplete
    *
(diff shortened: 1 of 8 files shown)


packages/blade/src/components/Input/DropdownInputTriggers/types.ts
```diff
@@ -39,6 +39,12 @@ type DropdownInputTriggersCommonProps = Pick<
    */
   defaultValue?: string | string[];
   onChange?: ({ name, values }: { name?: string; values: string[] }) => void;
+  /**
+   * Syncs the selected value to inputValue in AutoComplete
+   *
+   * Only needed in single select AutoComplete because inputValue is expected to be same as selected value there
+   */
+  syncInputValueWithSelection?: (value: string) => void;
```
chaitanyadeorukhkar, line 47: This is internal only, right?

saurabhdaware: yup

chaitanyadeorukhkar (approved):

snitin315 (approved):

chaitanyadeorukhkar (approved):

PR #1693, opened 2023-10-04
docs(blade): Tabs API Decisions

anuraghazra:
## [Formatted Document](https://github.com/razorpay/blade/blob/anu/tabs-api/packages/blade/src/components/Tabs/_decisions/decisions.md)


packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,299 @@
...
+- Controlled and Uncontrolled
+- Filled and Bordered variants
+- Horizontal and Vertical orientation
+- Medium and Large size
+
+## Anatomy
+
+<img src="./tabs-anatomy.png" width="100%" alt="Tabs thumbnail" />
+
+```jsx
```
chaitanyadeorukhkar, line 37: How about
```jsx
<Tabs>
  <TabList>
     <TabItem>
     </TabItem>
  </TabList>

  <TabPanel>
  </TabPanel>
</Tabs>

```

chaitanyadeorukhkar: TabPanel still seems very decoupled from TabItem, maybe we can nest them together somehow?

chaitanyadeorukhkar: How about something straightforward? What is the downside of this?
```jsx
<Tabs>
     <Tab value={} leading={} trailing={} title={}>
        {...the jsx content for the content of the tab}
     </Tab>
</Tabs>
```

anuraghazra: @chaitanyadeorukhkar while the API that you proposed looks much compact and straightforward there are major downsides to this API

### 1. Quite tricky/hack to implement

With the API that you proposed it will be quite tricky to map the JSX to the HTML structure that a typical tab requires. 

A tab's HTML structure is as follows

```html
<div>
  <div role="tablist">
    <button role="tab">tab item 1</button>
    <button role="tab">tab item 2</button>
    <button role="tab">tab item 3</button>
  </div>
  
  <div>
    <section role="tabpanel" />
    <section role="tabpanel" />
    <section role="tabpanel" />
  </div>
</div>
```

We need to achieve the above HTML structure, and to do this with the API that you proposed will get very hacky since now we need to use React.cloneElement and save those JSX nodes in state and render them in appropriate place. 

```html
<Tabs>
      // this will be the button with role="tab", need to React.cloneElement and move it to the right place, and while placing also need to make sure it has a parent of role="tablist"
     <Tab value={} leading={} trailing={} title={}>  
        // this is role="tabpanel" need to React.cloneElement and move it to the right place
        {...the jsx content for the content of the tab} 
     </Tab>
</Tabs>
```

### 2. The JSX Doesn't map well with the visual hierarchy or the HTML structure

If you notice the HTML structure pointed out in point #1 you'll see that it's 1:1 with the JSX API that i'm proposing. 

The current API that is proposed also maps 1:1 with the visual hierarchy of the component, but the second one doesn't map very well.  

Frame 1 (11)

### 3. Extensibility

The current API is also much more easier to extend in case any future usecase arises, for example let's say we want to use these tabs like a segmented control (which works like radio buttons but doesn't associate with any "tabpanels" we can simply remove the tabpanels and change the underlying html structure to acomodate for any future usecases. 

Check how reshaped is allowing this [extensibility](https://reshaped.so/content/docs/components/tabs#using-in-forms) - if you pass `name` prop it will convert the tabs to work like a Radio. 

### 4. API Familiarity 

The current API that is proposed is also much more familiar and is similar to how many other design systems are approaching compared to the second one: 

- [Reshaped](https://reshaped.so/content/docs/components/tabs)
- [Chakra](https://chakra-ui.com/docs/components/tabs/usage)
- [MUI](https://mui.com/material-ui/react-tabs/#experimental-api)
- [Atlassian](https://atlassian.design/components/tabs/examples)
- [RadixUI](https://www.radix-ui.com/primitives/docs/components/tabs)
- [HeadlessUI](https://headlessui.com/react/tabs)

anuraghazra: Added

anuraghazra: https://github.com/razorpay/blade/pull/1693#discussion_r1345552656

Also changed the naming of the current components to not use the plural suffix. instead of TabsItem -> TabItem

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+   */
+  size?: 'medium' | 'large';
+  /**
+   * The variant of the tabs.
+   */
+  variant?: 'bordered' | 'filled';
+  /**
+   * If `true`, the TabItems will grow to use all the available space.
+   */
+  autoWidth?: boolean;
```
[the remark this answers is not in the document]
Rozamo: Okay. So in this case the `TabItems` would stretch to occupy the whole `Tabs`, right?

Rozamo:
Can we also have some sort of [scrollable](https://mui.com/material-ui/react-tabs/#scrollable-tabs) feature? Would be helpful in smaller screens.

anuraghazra:
> Can we also have some sort of [scrollable](https://mui.com/material-ui/react-tabs/#scrollable-tabs) feature? Would be helpful in smaller screens.

Yes that will be supported out of the box. Check the design [here](https://www.figma.com/file/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?node-id=3733%3A10298&mode=dev): 

image

packages/blade/src/components/Tabs/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Yeah check the design [here](https://www.figma.com/file/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?node-id=3733%3A10954&mode=dev)

image

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+   */
+  value: string;
+  /**
+   * The content of the tab panel.
+   */
+  children: React.ReactNode;
+};
+```
+
+### Examples:
```
divyanshu013, line 135: Examples missing or wrong heading level?

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+```js
+<Tabs>
+  <TabList>
+    <TabItem value="payments">Payments</TabItem>
+    <TabItem value="refunds">Refunds</TabItem>
+    <TabItem value="disputes">Disputes</TabItem>
+    <TabItem value="settlements">Settlements</TabItem>
+  </TabList>
+
+  <TabPanel value="payments">
```
divyanshu013, line 148: One con of this API is I'm not sure if we can make `value` type safe? For example, easy to mis-spell `value` in tab item and panel

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+  trailing?: React.ReactNode;
+  /**
+   * If `true`, the tab item will be disabled.
+   */
+  isDisabled?: boolean;
+};
+```
+
+## `TabPanel` API
+
```
divyanshu013, line 121: Can we employ lazy loading for various tab panels? Maybe an example will help. Also will the tab panels be unmounted and removed from DOM when unselected?

divyanshu013 (commented):
Looks good overall

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: Yeah I think that we can't do anything. 

But should be fine, either way you should not have more than 7 or 8 tabs (maybe max 10), so don't foresee any maintainability issues. 

People can also do this

```
const paymentTabValue = "payments"

<TabItem value={paymentTabValue} />
<TabContent value={paymentTabValue} />
```

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: 🤔 Hmm. 

> Could probably expose a lazy prop
When you say "lazy loading tabs" do you mean to have the Tab's content be a Suspense component? Or do you mean a "lazy" prop which will control the mounted behaviour of the tabs, eg: lazy=true ---> unmounts on selection change, lazy=false ---> doesn't unmount just hides via display: none.

> Also will the tab panels be unmounted and removed from DOM when unselected?
right now we unmount unselected tabs.

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+   */
+  defaultValue?: string;
+  /**
+   * Callback fired when the value changes.
+   */
+  onChange?: (value: string) => void;
+  /**
+   * The orientation of the tabs.
+   */
+  orientation?: 'horizontal' | 'vertical';
```
kamleshchandnani, line 75: what if we call this as `direction: row | column`?

packages/blade/src/components/Tabs/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: autoWidth is confusing too. because i'm unable to understand what will it do? will it adjust the width of the tab or tablist or tab panel as it affects the width of the tabList. can we name it something like `isFullWidthTabList`?

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+```
+
+## `TabItem` API
+
+```ts
+type TabItemProps = {
+  /**
+   * The label of the tab item.
+   */
+  children: string;
```
kamleshchandnani, line 98: shouldn't this be `StringChildrenType`?

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+  /**
+   * The value of the tab item.
+   */
+  value: string;
+  /**
+   * Leading element of the tab item.
+   * Can be used to render an icon.
```
kamleshchandnani, line 105: are we putting a guardrail for this?

packages/blade/src/components/Tabs/_decisions/decisions.md
kamleshchandnani: >right now we unmount unselected tabs.

this could be trickier because in some product cases we might need to retain the state of the children(for eg checkbox or an accordion in the previous selected tab)

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+  /**
+   * The value of the tab panel. This will be used to match the selected tab.
+   */
+  value: string;
+  /**
+   * The content of the tab panel.
+   */
+  children: React.ReactNode;
+};
+```
```
kamleshchandnani, line 133: add `TabList` props?

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+    <TabItem value="payments">Payments</TabItem>
+    <TabItem value="refunds">Refunds</TabItem>
+    <TabItem value="disputes">Disputes</TabItem>
+    <TabItem value="settlements">Settlements</TabItem>
+  </TabList>
+
+  <TabPanel value="payments">
+    <Payments />
+  </TabPanel>
+  <TabPanel value="refunds">
```
kamleshchandnani, line 151: do we give the styling of tabpanel like a card or something or just an empty box and consumer decides the styling of panel?

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+  <TabPanel value="disputes">
+    <Disputes />
+  </TabPanel>
+  <TabPanel value="settlements">
+    <Settlements />
+  </TabPanel>
+</Tabs>
+```
+
+<img src="./example-1.png" width="380" alt="Tabs Example 1" />
```
kamleshchandnani, line 163: can you attach an example that is the output of the above code? this seems a made screenshot from existing page?

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+    </TabItem>
+    <TabItem value="excluded" leading={SlashIcon}>
+      Excluded
+    </TabItem>
+  </TabList>
+  ...
+</Tabs>
+```
+
+<img src="./example-4.png" width="380" alt="Tabs Example 2" />
```
kamleshchandnani, line 247: will the UI look like this exactly?

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Filled Variant
+
+```jsx
+<Tabs defaultValue="refunds">
```
kamleshchandnani, line 252: update this. typo here. it doesn't has filled variant and the default value is something else

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+<img src="./example-2.png" width="380" alt="Tabs Example 2" />
+
+## With Tooltip
+
+```jsx
+<Tabs>
+  <TabList>
+    <TabItem value="payments">Payments</TabItem>
+    <TabItem value="refunds">
+      <Tooltip content="Return (money) in restitution, repayment, or balancing of accounts.">
```
kamleshchandnani, line 280: the props table above says that the children will be string? but we are passing JSX here. will this work?

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+    <Text>Tab Panel Content For Disputes</Text>
+  </Tab>
+</Tabs>
+```
+
+We decided not to go with this API because of these downsides: 
+
+#### 1. Quite tricky/hack to implement
+
+With the API that you proposed it will be quite tricky to map the JSX to the HTML structure that a typical tab requires.
```
kamleshchandnani, line 315:
```suggestion
With this API it will be quite tricky to map the JSX to the HTML structure that a typical tab requires. 
```

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+  
+  <div>
+    <section role="tabpanel"></section>
+    <section role="tabpanel"></section>
+    <section role="tabpanel"></section>
+  </div>
+</div>
+```
+
+We need to achieve the above HTML structure, and to do this with the API that you proposed will get very hacky since now we need to use React.cloneElement and save those JSX nodes in state and render them in appropriate place.
```
kamleshchandnani, line 335:
```suggestion
We need to achieve the above HTML structure, and to do this with the above API it will get very hacky since now we need to use React.cloneElement and save those JSX nodes in state and render them in appropriate place. 
```

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+
+- [Reshaped](https://reshaped.so/content/docs/components/tabs)
+- [Chakra](https://chakra-ui.com/docs/components/tabs/usage)
+- [MUI](https://mui.com/material-ui/react-tabs/#experimental-api)
+- [Atlassian](https://atlassian.design/components/tabs/examples)
+- [RadixUI](https://www.radix-ui.com/primitives/docs/components/tabs)
+- [HeadlessUI](https://headlessui.com/react/tabs)
+
+
+## Motion
```
kamleshchandnani, line 373: put the motion video of tabs switching for both vertical+horizontal orientation and bordered and filled variant

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: > this could be trickier because in some product cases we might need to retain the state of the children(for eg checkbox or an accordion in the previous selected tab)

This can be solved, by adding lazy prop, on react-native side, [react-native-tab-view](https://reactnavigation.org/docs/tab-view/) also provides a similar API. Should be simple enough to implement.

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: Yeah sure, we can do this to enable interpolation.

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: TabList doesn't have any props, it's just used for structuring and internal logic management.

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: It's an empty box, no padding, no margin, no styling.

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: Oh wait right, we have to keep this React.ReactNode then to accommodate tooltip usecases.

packages/blade/src/components/Tabs/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: isFullWidthTabList, too verbose. 

I've seen design systems use "isFitted" prop. which might be better?

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: orientation I think sounds more natural here, direction might be confused with the flex property or the CSS direction attribute. (plus could cause issues if we have to support rtl, ltr)

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: Yes.

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: No it's not from an existing page, the screenshot is from the guidelines example RK added.

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: Yeah mostly, omitting whatever inside the <TabPanels \/>

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: Updated.

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: There are many variants, attached the motion prototype instead. People can open the prototype and inspect the motion.

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: Updated to use React.ReactNode.

packages/blade/src/components/Tabs/_decisions/decisions.md
anuraghazra: Updated.

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+
+Check the motion [documentation](https://www.figma.com/proto/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?type=design&node-id=4005-31894&t=b3OlPHxPuFH1BAIZ-0&scaling=min-zoom&page-id=3698%3A13772) here.
+
+## Accessibility
+
+Tabs will follow the [WAI-ARIA Tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) pattern.
+
+- [Keyboard navigation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#keyboardinteraction)
+
+## Open Questions
```
pcprashant, line 383: Are the tabs SEO-friendly by default, or do we need to make any adjustments?
Will the tab panels be server-rendered to facilitate SEO?

anuraghazra: By default all TabPanels be mounted on DOM and we will show/hide them via CSS to ensure SEO. 

And additionally, we will also introduce a `isLazy` prop which if present, will unmount the TabPanels from the DOM, could be useful if imagine inside your TabPanel you are doing API requests and don't want to mount/init them.

pcprashant (approved):

packages/blade/src/components/Tabs/_decisions/decisions.md
kamleshchandnani: yes so it'll least have children right? better to add in the spec to avoid any confusion

packages/blade/src/components/Tabs/_decisions/decisions.md
```diff
@@ -0,0 +1,385 @@
...
+- [Chakra](https://chakra-ui.com/docs/components/tabs/usage)
+- [MUI](https://mui.com/material-ui/react-tabs/#experimental-api)
+- [Atlassian](https://atlassian.design/components/tabs/examples)
+- [RadixUI](https://www.radix-ui.com/primitives/docs/components/tabs)
+- [HeadlessUI](https://headlessui.com/react/tabs)
+
+
+## Motion
+
+Check the motion [documentation](https://www.figma.com/proto/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?type=design&node-id=4005-31894&t=b3OlPHxPuFH1BAIZ-0&scaling=min-zoom&page-id=3698%3A13772) here.
```
kamleshchandnani, line 375:
```suggestion
Check the motion [prototype](https://www.figma.com/proto/LSG77hEeVYDk7j7WV7OMJE/Blade-DSL---Components-Guideline?type=design&node-id=4005-31894&t=b3OlPHxPuFH1BAIZ-0&scaling=min-zoom&page-id=3698%3A13772) here.
```

packages/blade/src/components/Tabs/_decisions/decisions.md
kamleshchandnani: can you record a tiny video from the prototype and it as a sample here? want it to act as a cue for readers to open the prorotype

packages/blade/src/components/Tabs/_decisions/decisions.md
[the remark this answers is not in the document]
kamleshchandnani: again adds a learning curve. i'm okay with verbose name as far as it conveys the purpose without me fiddling around and guessing what will it do.

packages/blade/src/components/Tabs/_decisions/decisions.md
kamleshchandnani: @anuraghazra bumping this up. link will easy be missed and rarely people will actually open it

anuraghazra: Added

anuraghazra:
> Can we have some sort of CSS such that the user knows that the tabs are scrollable? It can happen that the width of the visible tabs might be fitting 100% and the hidden tab is not visible.

This is something we can look into as an enhancement later on, in most cases you should avoid scrolling tabs and keep the tab items to 4 or 5 items only, and in case of mobile it's better to make the tab items `autoWidth` to avoid scrolling. 

Refer to the design guidelines for optimal number of tabs.

kamleshchandnani (approved):

PR #1694, opened 2023-10-04
feat(blade): tabs web implementation

anuraghazra:
- [PR 1](https://github.com/razorpay/blade/pull/1694) - Web Implementation (this PR)
- [PR 2](https://github.com/razorpay/blade/pull/1707) - Native Implementation 
- [PR 3](https://github.com/razorpay/blade/pull/1715) - Documentation


packages/blade/src/components/Tabs/TabIndicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+        position="absolute"
+        left="0px"
+        top="0px"
+        borderRadius="small"
+        backgroundColor="brand.primary.300"
+        style={{
+          ...transitionProps,
+          width: `${activeElementDimensions.width}px`,
+          height: `${activeElementDimensions.height}px`,
+          transform: `translate(${activeElementDimensions.x}px, ${activeElementDimensions.y}px)`,
```
saurabhdaware, line 116: nit: makeSize?

packages/blade/src/components/Tabs/TabItem.tsx
```diff
@@ -0,0 +1,166 @@
...
+    cursor: 'pointer',
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: isVertical ? 'left' : 'center',
+    gap: makeSpace(theme.spacing[3]),
+    width: isFullWidthTabItem ? '100%' : undefined,
+    paddingTop: makeSpace(get(theme, paddingY[_variant][orientation][device][size!])),
+    paddingBottom: makeSpace(get(theme, paddingY[_variant][orientation][device][size!])),
+    paddingLeft: makeSpace(get(theme, paddingX[_variant][orientation][device][size!])),
+    paddingRight: makeSpace(get(theme, paddingX[_variant][orientation][device][size!])),
```
saurabhdaware, line 51: should we use media queries instead? since its not too dificult here and don't think so our `isMobile` is server-render compatible so there might be flash of different padding in SSR.

packages/blade/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,61 @@
...
+      setSelectedValue,
+      size,
+      variant,
+      isLazy,
+    ],
+  );
+
+  return (
+    <TabsContext.Provider value={contextValue}>
+      <BaseBox display="flex" flexDirection={isVertical ? 'row' : 'column'}>
```
saurabhdaware, line 54: TODO: pending to add metaAttributes and other things?

packages/blade/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import type { TabsProps } from './types';
+import { TabsContext } from './TabsContext';
+import { useControllableState } from '~utils/useControllable';
+import { useId } from '~utils/useId';
+import BaseBox from '~components/Box/BaseBox';
+
+const Tabs = ({
```
saurabhdaware, line 8: Add basic usage example in JSDoc here maybe?

saurabhdaware (commented):
looks good to me mostly. minor comments.

packages/blade/src/components/Tabs/TabIndicator.tsx
anuraghazra: It's fine I think, this component is only for web. makeSize makes the LOC too long and hard to read.

packages/blade/src/components/Tabs/Tabs.tsx
anuraghazra: Yeah. Will add this on upstream PRs for both web and RN

packages/blade/src/components/Tabs/TabItem.tsx
anuraghazra: Okay, can you check this? 

https://github.com/razorpay/blade/pull/1694/commits/01ce3aab88dd7d69ebf5ce15ec65a79f87e54c05

packages/blade/src/components/Tabs/Tabs.tsx
anuraghazra: Added

anuraghazra:
For documentation and storybook examples check out the Tab Documentation [PR](https://github.com/razorpay/blade/pull/1715)

saurabhdaware (approved):
Looks good 🥳

packages/blade/src/components/Tabs/TabsContext.ts
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import type { TabsProps } from './types';
+import { throwBladeError } from '~utils/logger';
+import type { ControllableStateSetter } from '~utils/useControllable';
+
+type TabsContextProps =
+  | (Pick<TabsProps, 'size' | 'variant' | 'isFullWidthTabItem' | 'isLazy'> & {
+      isVertical: boolean;
+      baseId: string;
+      selectedValue: string;
```
kamleshchandnani, line 10: would it make more sense if we call this as `selectedTab` rather than `selectedValue`?

packages/blade/src/components/Tabs/tabTokens.ts
```diff
@@ -0,0 +1,174 @@
+import type { TabsProps } from './types';
+import type { DotNotationSpacingStringToken } from '~utils/types';
+
+type Devices = 'desktop' | 'mobile';
```
kamleshchandnani, line 4: can we use this type from `useBreakpoint`?

packages/blade/src/components/Tabs/tabTokens.ts
```diff
@@ -0,0 +1,174 @@
...
+type TabSizes = NonNullable<TabsProps['size']>;
+type TabVariants = Exclude<NonNullable<TabsProps['variant']>, 'borderless'>;
+type TabOrientation = NonNullable<TabsProps['orientation']>;
+
+type TabItemPadding = Record<
+  TabVariants,
+  Record<TabOrientation, Record<Devices, Record<TabSizes, DotNotationSpacingStringToken>>>
+>;
+
+const paddingY: TabItemPadding = {
```
kamleshchandnani, line 14: does it seriously needs this complex padding so many permutations and combinations?

packages/blade/src/components/Tabs/types.ts
```diff
@@ -0,0 +1,100 @@
...
+   */
+  value: string;
+  /**
+   * Leading element of the tab item.
+   * Can be used to render an icon.
+   */
+  leading?: React.ReactNode;
+  /**
+   * Trailing element of the tab item.
+   * Can be used to render an badge/counter.
```
kamleshchandnani, line 66:
```suggestion
   * Can be used to render a Badge/Counter component.
```

packages/blade/src/components/Tabs/types.ts
```diff
@@ -0,0 +1,100 @@
...
+  href?: string;
+  /**
+   * Callback fired when the tab item is clicked.
+   */
+  onClick?: (event: React.MouseEvent) => void;
+};
+
+type TabPanelProps = {
+  /**
+   * The value of the tab panel. This will be used to match the selected tab.
```
kamleshchandnani, line 91:
```suggestion
   * The value of the tab panel same as the corresponding TabItem's value to match the selected TabItem.
```

packages/blade/src/components/Tabs/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  const [
+    validatedTrailingComponent,
+    setValidatedTrailingComponent,
+  ] = React.useState<React.ReactElement | null>(null);
+
+  // validate and restrict sub component props in trailing prop
+  useIsomorphicLayoutEffect(() => {
+    if (React.isValidElement(trailing)) {
+      const trailingComponentType = getComponentId(trailing) as TrailingComponents;
+      const restrictedProps = propRestrictionMap[trailingComponentType]?.[tabItemSize];
```
kamleshchandnani, line 56: why do we call it propRestriction? we only need to check on component restriction right?

kamleshchandnani: also, the size shouldn't be restricted to `tabItemSize` right they should be mutually exclusive, they are same today but not sure if they'll always have 1:1 mapping?

packages/blade/src/components/Tabs/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+          });
+        }
+
+        const restrictedPropKeys = Object.keys(restrictedProps);
+        for (const prop of restrictedPropKeys) {
+          if (trailing?.props?.hasOwnProperty(prop)) {
+            logger({
+              message: `Do not pass "${prop}" to "${trailingComponentType}" while inside TabsItem trailing, because we override it.`,
+              moduleName: 'TabsItem',
+              type: 'warn',
```
kamleshchandnani, line 71: shouldn't this be error instead of warning? we shall break the component rendering IMO in these kind of cases

packages/blade/src/components/Tabs/utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  trailing: React.ReactNode,
+  tabItemSize: NonNullable<TabsProps['size']>,
+): React.ReactNode => {
+  const [
+    validatedTrailingComponent,
+    setValidatedTrailingComponent,
+  ] = React.useState<React.ReactElement | null>(null);
+
+  // validate and restrict sub component props in trailing prop
+  useIsomorphicLayoutEffect(() => {
```
kamleshchandnani, line 53: do we use this kind of method for all the restrictions? haven't seen in the previous components 🤔

packages/blade/src/utils/useControllable.ts
```diff
@@ -3,6 +3,17 @@
 /* eslint-disable react-hooks/exhaustive-deps */
 import * as React from 'react';
 
+type ControllableStateSetter<T> = (
+  /**
+   * Sets the state to the given value
```
kamleshchandnani, line 8: this is that hacky way or we found the main issue?

packages/blade/src/components/Tabs/TabIndicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+});
+
+const TabIndicator = ({
+  tabListContainerRef,
+}: {
+  tabListContainerRef: React.RefObject<HTMLElement | null>;
+}): React.ReactElement => {
+  const { theme } = useTheme();
+  const { selectedValue, baseId, variant } = useTabsContext();
+  const [hasMeasured, setHasMeasured] = React.useState(false);
```
kamleshchandnani, line 27: what does `hasMeasured` mean?

packages/blade/src/components/Tabs/TabIndicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const [activeElementDimensions, setActiveElementDimensions] = React.useState({
+    width: 0,
+    height: 0,
+    x: 0,
+    y: 0,
+  });
+
+  // Update the dimensions of the active element
+  const updateDimensions = React.useCallback(() => {
```
kamleshchandnani, line 36: why do we change the width and height when the tabitem is selected?

packages/blade/src/components/Tabs/TabIndicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+    });
+  }, [baseId, selectedValue, tabListContainerRef, variant]);
+
+  // Update the dimensions when the selected value changes
+  useIsomorphicLayoutEffect(() => {
+    if (!selectedValue) return;
+
+    updateDimensions();
+
+    const id = requestAnimationFrame(() => {
```
kamleshchandnani, line 60: i hope this won't break in website again 🙈

packages/blade/src/components/Tabs/TabIndicator.tsx
```diff
@@ -0,0 +1,139 @@
...
+    };
+  }, [baseId, selectedValue]);
+
+  // Update the dimensions when the window resizes or when the font loads
+  React.useEffect(() => {
+    if (!tabListContainerRef.current) return;
+
+    // check for FontFace API support
+    // FontFaceAPI is widely supported but better to be safe than sorry
+    if ('fonts' in document) {
```
kamleshchandnani, line 77: why do we do this in tabs? we already have fallback fonts right?

packages/blade/src/components/Tabs/TabPanel.tsx
```diff
@@ -0,0 +1,42 @@
...
+    return (
+      <>
+        {isSelected ? (
+          <BaseBox
+            id={panelId}
+            tabIndex={0}
+            {...makeAccessible({ role: 'tabpanel', labelledBy: tabItemId })}
+          >
+            {children}
+          </BaseBox>
```
kamleshchandnani, line 24: why not extract this and the component below as one component?

packages/blade/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,91 @@
...
+  defaultValue,
+  value,
+  onChange,
+  orientation = 'horizontal',
+  size = 'medium',
+  variant = 'bordered',
+  isFullWidthTabItem = false,
+  isLazy = false,
+}: TabsProps): React.ReactElement => {
+  const baseId = useId('tabs');
```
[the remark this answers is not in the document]
anuraghazra: Do we know if all our consumers on React18? If so we can use React.useId otherwise it will throw error on their end.

packages/blade/src/components/Tabs/TabsContext.ts
anuraghazra: both works but value makes more sense because even the public API our nomenclature is `value` not `selectedTab` 

<TabPanel value="payments" \/>

packages/blade/src/components/Tabs/tabTokens.ts
anuraghazra: We have a `DeviceType` type defined there, I've exported it.

packages/blade/src/components/Tabs/tabTokens.ts
```diff
@@ -0,0 +1,174 @@
+import type { TabsProps } from './types';
+import type { DotNotationSpacingStringToken } from '~utils/types';
+
+type Devices = 'desktop' | 'mobile';
+type TabSizes = NonNullable<TabsProps['size']>;
+type TabVariants = Exclude<NonNullable<TabsProps['variant']>, 'borderless'>;
```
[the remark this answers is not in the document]
anuraghazra: bordered & borderless are essentially the same variants with same padding.

packages/blade/src/components/Tabs/tabTokens.ts
anuraghazra: :p you are looking at the simplified version, previously it was [monstrous](https://github.com/razorpay/blade/pull/1694/commits/9a5d240a88c9422514a2d67fa131857c0e93cff8#diff-18f9bb44412c8a625a3f751777eb97fd77db23be76807564f80c6bd07fdd63f4R25-R188). 

This approach is a whole lot easier to maintain and update later than using ternaries to switch between different variants.

packages/blade/src/components/Tabs/utils.ts
anuraghazra: > why do we call it propRestriction? we only need to check on component restriction right?

We don't just check, we override/enhance the component based on the `propRestrictionMap`

> also, the size shouldn't be restricted to tabItemSize right they should be mutually exclusive, they are same today but not sure if they'll always have 1:1 mapping?

Didn't understand what do you mean by mutually exclusive. 
They are not mutually exclusive check above for mappings: 

```ts
const iconSizeMap = {
  medium: 'medium',
  large: 'large',
} as const;

const badgeSizeMap = {
  medium: 'small',
  large: 'medium',
} as const;

const counterSizeMap = {
  medium: 'small',
  large: 'small',
} as const;
```

packages/blade/src/components/Tabs/utils.ts
anuraghazra: Let's say internally we override `size` prop of Badge to always be `large`. 
And consumer again passes `size=medium` but this won't have an effect because we override it internally. 

In this case even if consumer passes a size prop nothing breaks visually, it's essentially becomes a NOOP. 

Throwing error for each and every case will cause disruption while using this component instead it's better to let them know this has no effect even if you do it. (it's even worse on RN because RN will show a full page error dialog)

-----

If passing some prop can cause visual breakage in the component only then we should opt for throwing errors, like how we throw error if they pass anything other than Badge/Counter.

packages/blade/src/components/Tabs/utils.ts
anuraghazra: BottomSheet/BaseHeader uses this. 

It's same as any other inline runtime checks that we do, just extracted to a hook because this is a lot of code.

packages/blade/src/utils/useControllable.ts
anuraghazra: I spent quite a while on that bug even created a [reproduction](https://codesandbox.io/s/usecontrollable-usefirstrender-bug-th7545) of this to debug the issue, could not figure out why this would behave like they way it does. 

There's no point further spending time on this. Because anyways the "hacky" way is much much cleaner than messing around with useRef, useEffect, useisFirstRender, useControllable complexities. 

**Goal:**

So our main goal is that, we don't want to call the onChange when we set the initial value of the tabs. 

**Intent:**

With the "hacky" way the intent is simple: 
_"When setting initial value, don't call onChange"_ 

With the useIsFirstRender the intent becomes:
_"When setting initial value, don't call onChange, and since we set initial value on first render we need to skip calling onChange on the first render"_


Overall the hacky way is much more straightforward than messing with React's lifecycle.

packages/blade/src/components/Tabs/TabIndicator.tsx
anuraghazra: It tracks if we are finished with measuring the TabItem's width which we need to calculate to animate the TabIndicator.

packages/blade/src/components/Tabs/TabIndicator.tsx
anuraghazra: This is the bottom blue indicator that animates from 1 tabitem to another tabitem, we need recalculate and update it's dimensions. 🤔

packages/blade/src/components/Tabs/TabIndicator.tsx
anuraghazra: This is `animationFrame`, `idleCallback` which broke previously. 

rAF is supported on all major browsers & we are already using rAF in multiple places in blade.

packages/blade/src/components/Tabs/TabIndicator.tsx
anuraghazra: This is waiting for fonts to be fully loaded and recalculating that blue bottom indicator's dimensions, because on initial render the fonts are not yet loaded and that time the dimensions are different. 

Check the video: See how after the fonts are loaded the indicator adjusts it's width to accommodate the new width of the tab item: 

https://github.com/razorpay/blade/assets/35374649/093e8657-86dd-418a-a686-4207eef3b901

packages/blade/src/components/Tabs/TabsContext.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const TabsContext = React.createContext<TabsContextProps>(null);
+
+const useTabsContext = (): NonNullable<TabsContextProps> => {
+  const context = React.useContext(TabsContext);
+
+  if (!context) {
+    throwBladeError({
+      moduleName: 'Tabs',
+      message: 'Tabs compound components cannot be rendered outside the Tabs component',
```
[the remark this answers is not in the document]
anuraghazra: done.

packages/blade/src/components/Tabs/tabTokens.ts
anuraghazra: done.

packages/blade/src/components/Tabs/types.ts
```diff
@@ -0,0 +1,100 @@
...
+   * If `true`, the tab item will be disabled.
+   */
+  isDisabled?: boolean;
+  /**
+   * If set the tab item will be rendered as a link.
+   * This can be used to create a tab item that redirects to another page or integrate with react-router.
+   *
+   * @default undefined
+   */
+  href?: string;
```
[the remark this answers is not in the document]
anuraghazra: Don't think there is any pattern where a clicking on a TabItem it opens the panel on a new tab. So no need of rel/target.

saurabhdaware (approved):

kamleshchandnani (approved):

packages/blade/src/components/Tabs/tabTokens.ts
nashcheez: @coderabbit

.changeset/breezy-dancers-breathe.md
```diff
@@ -0,0 +1,9 @@
+---
+"@razorpay/blade": minor
+---
+
+feat(blade): tabs web implementation
```
chaitanyadeorukhkar, line 5: Change changeset text. This will go in our changelog

chaitanyadeorukhkar (dismissed):

chaitanyadeorukhkar (approved):

kamleshchandnani (approved):

Issue #1696, opened 2023-10-09
Getting validateDOMNesting warning while using ListItemText inside ListItem


tarungupta9:
`Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>`
Please refer codesandbox to see issue replication: https://codesandbox.io/s/jolly-bash-d92v55?file=/App.tsx

tarungupta9:
Fix PR: https://github.com/razorpay/blade/pull/1697

PR #1697, opened 2023-10-10
feat(ListItemText): making span as default tag for ListItemText

tarungupta9:
Issue: https://github.com/razorpay/blade/issues/1696

Context: `Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>`
`<ListItemText />` uses `<p>` tag by default and `<ListItem />` renders a `<p>` tag as well which is causing the aforementioned issue.

Solution: ListItemText uses Text component which renders `<p>` tag by default. However, for almost all the use-cases ListItemText will get rendered inside ListItem and we can make `<span>` as default

https://razorpay.slack.com/archives/CMQ3RBHEU/p1696501813630969

### DIFF
--- a/.changeset/bright-pumpkins-push.md
+++ b/.changeset/bright-pumpkins-push.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+feat(ListItemText): make span as default tag for ListItemText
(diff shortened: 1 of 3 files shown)


chaitanyadeorukhkar (approved):

snitin315 (changes requested):
Code LGTM, we need to fix CI

tarungupta9:
> Code LGTM, we need to fix CI

@snitin315, checking

snitin315 (approved):
LGTM, thank you for your contribution ❤️

Issue #1679, opened 2023-09-28
Add Visibility and Opacity Properties to the Box Component


ajaybrahmarzp:
Currently, the Box component provided by Blade lacks support for the visibility and opacity CSS properties. These properties are essential for controlling the visibility and transparency of elements, which is a common requirement in modern web applications.

PR #1702, opened 2023-10-11
feat(Box): add `transform`, `transformOrigin`, `clipPath`, `opacity`, & `visibility` props

snitin315:
Fixes https://github.com/razorpay/blade/issues/1394
Fixes https://github.com/razorpay/blade/issues/1679

Screenshot 2023-10-11 at 5 33 26 PM

### DIFF
--- a/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
+++ b/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
@@ -222,9 +222,15 @@ const getAllProps = (
     userSelect: getResponsiveValue(props.userSelect, breakpoint),
     pointerEvents: getResponsiveValue(props.pointerEvents),
     opacity: getResponsiveValue(props.opacity, breakpoint),
+    visibility: getResponsiveValue(props.visibility, breakpoint),
     ...(!isReactNative() && {
       boxShadow: getElevationValue(props.elevation, props.theme, breakpoint),
     }),
+
+    // Polygon support
+    transform: getResponsiveValue(props.transform as string, breakpoint),
+    transformOrigin: getResponsiveValue(props.transformOrigin, breakpoint),
+    clipPath: getResponsiveValue(props.clipPath, breakpoint),
   };
 };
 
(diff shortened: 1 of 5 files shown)


packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -97,6 +97,10 @@ type GridProps = MakeObjectResponsive<
   >
 >;
 
+type PolygonSupportProps = MakeObjectResponsive<
+  PickCSSByPlatform<'transform' | 'transformOrigin' | 'clipPath'>
+>;
```
saurabhdaware, line 102: Should we add this to `CommonBoxVisualProps`? Since that is where we have other equivalent props and `transform` and `transformOrigin` are not polygon-specific props.

saurabhdaware (approved):

packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
```diff
@@ -222,9 +222,15 @@ const getAllProps = (
     userSelect: getResponsiveValue(props.userSelect, breakpoint),
     pointerEvents: getResponsiveValue(props.pointerEvents),
     opacity: getResponsiveValue(props.opacity, breakpoint),
+    visibility: getResponsiveValue(props.visibility, breakpoint),
     ...(!isReactNative() && {
       boxShadow: getElevationValue(props.elevation, props.theme, breakpoint),
     }),
+
+    // Polygon support
+    transform: getResponsiveValue(props.transform as string, breakpoint),
```
kamleshchandnani, line 231: do we need this now since we are approaching rebranding?

snitin315: Does re-branding ensure no more polygons?

anuraghazra (approved):

packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts
saurabhdaware: Talked to Chaitanya about this today. There are going to be polygons after rebranding too but they will be built on frontend-website repo.

I think we still might have to expose this to enable them to build Polygons using Box component (since that covers major part of homepage)

PR #1706, opened 2023-10-12
feat(blade-coverage-script): filter out svgs and media elements

anuraghazra:
Pulls in changes from https://github.com/razorpay/frontend-core-tools/pull/96

### DIFF
--- a/packages/blade-coverage-extension/background_script.js
+++ b/packages/blade-coverage-extension/background_script.js
@@ -28,7 +28,16 @@ const calculateBladeCoverage = (shouldHighlightNodes) => {
     return false;
   };
 
+  /**
+   * Checks if DOM node is a media element or not
+   */
+  const isMediaElement = (element) => {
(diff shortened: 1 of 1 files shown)


packages/blade-coverage-extension/background_script.js
```diff
@@ -48,7 +77,11 @@ const calculateBladeCoverage = (shouldHighlightNodes) => {
 
   const totalNodes = totalNodeElements.length;
   const bladeNodes = bladeNodeElements.length;
-  const bladeCoverage = Number(((bladeNodes / totalNodes) * 100).toFixed(2));
+  let bladeCoverage = Number(((bladeNodes / totalNodes) * 100).toFixed(2));
+  // NaN guard
+  if (totalNodes === 0) {
```
kamleshchandnani, line 82: when will totalNodes ever be 0? only when the page is blank?

PR #1707, opened 2023-10-12
feat(blade): tabs react-native implementation

anuraghazra:
- [PR 1](https://github.com/razorpay/blade/pull/1694) - Web Implementation
- [PR 2](https://github.com/razorpay/blade/pull/1707) - Native Implementation (this PR)
- [PR 3](https://github.com/razorpay/blade/pull/1715) - Documentation


packages/blade/src/components/Tabs/TabIndicator.native.tsx
```diff
@@ -0,0 +1,41 @@
...
+        if (!isFilled) return props.getTabWidth(index);
+        if (index === props.navigationState.routes.length - 1) {
+          return props.getTabWidth(index) - theme.spacing[2] * 3;
+        }
+        return props.getTabWidth(index);
+      }}
+      style={{
+        ...(isFilled
+          ? {
+              height: props.layout.height - theme.border.width.thick - theme.spacing[2] * 2,
```
chaitanyadeorukhkar, line 26: I see tab navigation is giving the layout height & width. I hope its well optimised and doesn't cause too many re-renders since its passing layout (that could change) in props 🤔

packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -0,0 +1,214 @@
...
+const getTabs = (children: React.ReactNode): React.ReactElement[] => {
+  const childs = React.Children.toArray(children);
+  const tabList = childs.find((child) => getComponentId(child) === 'TabList');
+  if (!tabList || !React.isValidElement(tabList)) throw new Error('TabList is required');
+
+  return tabList.props.children;
+};
+
+const getTabPanels = (children: React.ReactNode): Record<string, () => React.ReactElement> => {
+  const childs = React.Children.toArray(children) as React.ReactElement[];
```
chaitanyadeorukhkar, line 36:
```suggestion
  const children = React.Children.toArray(children) as React.ReactElement[];
```

packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -0,0 +1,214 @@
...
+  return childs
+    .filter((child) => getComponentId(child) === 'TabPanel')
+    .reduce((prev, curr) => {
+      return { ...prev, [curr.props.value]: () => curr.props.children };
+    }, {});
+};
+
+const getRoutes = (tabs: React.ReactElement[]) => {
```
chaitanyadeorukhkar, line 44: What is routes?

packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const [selectedValue, setSelectedValue] = useControllableState({
+    defaultValue,
+    value,
+    onChange: (value) => {
+      if (isFirstRender) return;
+      onChange?.(value);
+    },
+  });
+
+  const index = getRouteIndexFromValue({
```
chaitanyadeorukhkar, line 91: more contextual?
```suggestion
  const routeIndex = getRouteIndexFromValue({
```

packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+  const setIndex = React.useCallback(
+    (index: number) => {
+      const value = getRouteValueFromIndex({ index, routes });
+      setSelectedValue(() => value);
+    },
+    [routes, setSelectedValue],
+  );
+
+  const isVertical = false;
```
chaitanyadeorukhkar, line 104: add comment why its always false

packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -0,0 +1,214 @@
...
+      const { title, leading, trailing } = route;
+      const selectedState = focused ? 'selected' : 'unselected';
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      const validatedTrailingComponent = useTabsItemPropRestriction(trailing, size);
+
+      return (
+        <StyledTabButton autoWidth={!autoWidth && !isFilled} variant={variant} size={size}>
+          <Box display="flex" alignItems="center" flexDirection="row" gap="spacing.3">
+            {leading
+              ? React.cloneElement(leading as React.ReactElement, {
```
chaitanyadeorukhkar, line 132: What are we doing here? Cloning I'm assuming an Icon element and adding size & color?

Why this way? See how we're doing it in other places like List. We should have a different component like `TabIcon` which only has limited props. Cloning and adding sizes this way will lead to confusion because users might pass a size and then when it doesn't work they will raise an issue.

packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -0,0 +1,214 @@
...
+                  size: iconSizeMap[size],
+                  color: `surface.action.icon.default.lowContrast`,
+                })
+              : null}
+            <Text
+              color={textColor[selectedState].default}
+              size={size === 'medium' ? 'medium' : 'large'}
+              weight={size === 'medium' ? 'bold' : 'regular'}
```
chaitanyadeorukhkar, line 140: Save as mapping somewhere. Don't hardcode it here directly. I'm assuming same will be applicable for web too

packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -0,0 +1,214 @@
...
+          padding: 0,
+          margin: 0,
+          minHeight: 0,
+        }}
+        contentContainerStyle={
+          isFilled
+            ? {}
+            : {
+                borderBottomColor: theme.colors.surface.border.normal.lowContrast,
+                borderBottomWidth: 1.5,
```
chaitanyadeorukhkar, line 169: Use tokens

packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -0,0 +1,214 @@
...
+      <TabView
+        navigationState={{
+          index,
+          routes,
+        }}
+        renderScene={SafeSceneMap(panels)}
+        renderTabBar={renderTabBar}
+        onIndexChange={setIndex}
+        initialLayout={initialLayout}
+        lazy={false}
```
chaitanyadeorukhkar, line 208: Why not lazy load it? Give it to consumers as an option. On our mobile apps we're using lazy load in places. If we don't, the initial load is very laggy since it makes all the API calls for all the tabs at once.

packages/blade/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -12,42 +12,19 @@ import { Box } from '~components/Box';
 export default {
```
chaitanyadeorukhkar, line 12: RN stories are very empty right now. Can you at least add some basic stories for variants & sizes? Or will that be done in documentation PR?

packages/blade/src/components/Tabs/TabIndicator.native.tsx
anuraghazra: Mostly everything is handled by the library only, and we are already using the [optimization](https://reactnavigation.org/docs/tab-view/#optimize-large-number-of-routes) mentioned in the docs so should be fine.

packages/blade/src/components/Tabs/TabItem.native.tsx
```diff
@@ -0,0 +1,31 @@
...
+    flexDirection: 'row',
+    width: autoWidth ? '100%' : undefined,
+    paddingTop: makeSpace(get(theme, paddings[variant].horizontal.mobile.top[size!])),
+    paddingBottom: makeSpace(get(theme, paddings[variant].horizontal.mobile.bottom[size!])),
+    paddingLeft: makeSpace(get(theme, paddings[variant].horizontal.mobile.left[size!])),
+    paddingRight: makeSpace(get(theme, paddings[variant].horizontal.mobile.right[size!])),
+  };
+});
+
+const TabItem = (_props: TabItemProps): React.ReactElement => {
```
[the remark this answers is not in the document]
anuraghazra: So internally we just loop over the TabItems and [extract](https://github.com/razorpay/blade/pull/1707/files#diff-867b33d70dec2cbec73a53d6f39a4cdfd13fdd6a35a9dfd194b07e967db4bdc2R46-R53) it's props. This component itself doesn't need to render anything. 

Since rendering is handled by the react-native-tab-view library's [renderTabBar](https://reactnavigation.org/docs/tab-view/#rendertabbar) etc APIs

packages/blade/src/components/Tabs/TabList.native.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { componentIds } from './utils';
+import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
+
+const _TabList = (_props: { children: React.ReactNode }): React.ReactElement => {
```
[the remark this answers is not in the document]
anuraghazra: See https://github.com/razorpay/blade/pull/1707#discussion_r1368285955

packages/blade/src/components/Tabs/Tabs.native.tsx
anuraghazra: It's an array of route object which TabView component expects. They basically map to the TabItems. 

https://reactnavigation.org/docs/tab-view/#tabview-props

packages/blade/src/components/Tabs/Tabs.native.tsx
anuraghazra: Added

packages/blade/src/components/Tabs/Tabs.native.tsx
anuraghazra: > Cloning and adding sizes this way will lead to confusion

I've already modified this to not use cloneElement. 


> We should have a different component like TabIcon which only has limited props.

No, we are already doing this this way in newer components like BottomSheet, BaseHeader etc. Introducing new component for every single Parent component is not very ideal. 

Since we override both color/size prop in icon, consumers will simply use it like so: 

```
<TabItem icon={InfoIcon} />
```

No need to introduce new TabItemIcon component.

packages/blade/src/components/Tabs/Tabs.native.tsx
anuraghazra: Don't actually need to weight wala mapping. It's always bold.

packages/blade/src/components/Tabs/Tabs.native.tsx
anuraghazra: Added isLazy prop so consumers can decide this.

packages/blade/src/components/Tabs/Tabs.stories.tsx
anuraghazra: Yes more stories are added in the documentation PR. Check #1715

packages/blade/src/components/Tabs/TabItem.native.tsx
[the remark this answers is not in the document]
chaitanyadeorukhkar: Write a comment in these files to avoid confusion?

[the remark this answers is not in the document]
anuraghazra: Done.

chaitanyadeorukhkar (approved):

saurabhdaware (approved):
Looks good to me

Issue #1676, opened 2023-09-27
In controlled Autocomplete, selected item is getting updated without explicit setState call 


sujoys12:
In the controlled component, the value should be updated only if the consumer explicitly updates it, i.e. makes a setCurrentSelection call inside the onChange callback.
But in this case, it's getting updated without the setState call.

Code sandbox - https://codesandbox.io/s/elated-vaughan-csqj3f

snitin315:
@sujoys12 I believe this is expected behavior because we are explicitly passing `value={currentSelection}` to the Autocomplete. So it will be selected whenever we update the value via `setCurrentSelection` regardless of where this is being called.

saurabhdaware:
> So it will be selected whenever we update the value via setCurrentSelection regardless of where this is being called

@snitin315 I think in this case the issue is that it is updating even when you don't call `setCurrentSelection` (when you go and change dropdown from UI without calling `setCurrentSelection` in onChange)

I am able to reproduce it by going to the demo mentioned in the issue, and clicking on the input, selecting item. It selects the item but it shouldn't because we're not calling setCurrentSelection in its onChange

snitin315:
@saurabhdaware Gotcha! Raised a fix.

PR #1714, opened 2023-10-15
fix(Autocomplete): enhance input handling for controlled scenarios

snitin315:
Fix https://github.com/razorpay/blade/issues/1676

### DIFF
--- a/packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
+++ b/packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
@@ -128,7 +128,9 @@ const useAutoComplete = ({
         value: displayText,
       });
       // Use displayText as inputValue only if its not controlled by user
-      setInputValue((props.value as string) ?? displayText ?? '');
+      if (typeof props.value === 'undefined') {
+        setInputValue(displayText ?? '');
+      }
     }
     props.onChange?.({ name: props.name, values });
   };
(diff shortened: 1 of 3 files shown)


anuraghazra (approved):
Looks good, possible to add a test?

snitin315:
I believe testing is not required for this.

kamleshchandnani:
> I believe testing is not required for this.

@snitin315 why do you say that? I feel we shall add a test for this to make sure this doesn't occur in future?

snitin315:
I will add one, in this case adding a test case takes more time than writing the fix. Also, it's super rare not to call `setValue` in `onChange`, in controlled scenarios.

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.tsx
```diff
@@ -127,7 +127,8 @@ const useAutoComplete = ({
         name: props.name,
         value: displayText,
       });
-      setInputValue(displayText ?? '');
+      // Use displayText as inputValue only if its not controlled by user
+      setInputValue((props.value as string) ?? displayText ?? '');
```
saurabhdaware, line 131: displayText is supposed to be different from value.

E.g. if `value="mumbai"`, and we have that value mapped to `"Mumbai"` in ActionListItem, then the value displayed in input is "Mumbai" and not "mumbai".


I think what we need to do here might be to not call `setInputValue` when value is being controlled 🤔

snitin315: Let me check if that fixes the issue.

saurabhdaware (dismissed):
Looks good to me

saurabhdaware (approved):

Issue #1721, opened 2023-10-18
Multi-select dropdown component breaks with certain props


anilprzp:
### Environment
Blade version - `10.15.0`
Browser - Chrome v112.0.5615.137

### Description
Passing a certain set of props to multi-select dropdown is breaking the component. For reproducing the bug - 
1. Go to the multi-select example on storybook - https://master--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/story/components-dropdown-with-select--with-multi-select
2. Paste the following snippet with specific set of props - 
    ```App.tsx
    import { 
      Dropdown, 
      DropdownOverlay,
      SelectInput,
      ActionList,
      ActionListItem,
    } from '@razorpay/blade/components';
    
    function App(): React.ReactElement {
      return (
        <Dropdown 
          selectionType="multiple"
        >
          <SelectInput
            label="City"
            placeholder="Select your City"
            name="action"
            value={["mysore"]} // value prop is added extra
            onChange={({ name, values }) => {
              console.log({ name, values });
            }}
          />
          <DropdownOverlay>
            <ActionList>
              <ActionListItem title="Mumbai" value="mumbai" isDisabled /* isDisabled is added extra */ />
              <ActionListItem title="Pune" value="pune" />
              <ActionListItem title="Bangalore" value="bangalore" />
              <ActionListItem title="Mysore" value="mysore" />
            </ActionList>
          </DropdownOverlay>
        </Dropdown>
      )
    }
    
    export default App;
    ```

Please find the screen recordings of how to reproduce the bug - 

https://github.com/razorpay/blade/assets/95009160/8e91ab1b-fec5-487f-ba4c-0f964c0a9aba

https://github.com/razorpay/blade/assets/95009160/801f5928-9f97-4563-ac01-e0b03b57c62f

anilprzp:
**Please note that this is a potential release blocker for XPayroll team**

anuraghazra:
@anilprzp we will prioritize this ASAP.

saurabhdaware:
@anilprzp can you also share the code for your X Payroll component?

The error in sandpack seem due to the react refresh in dev-server. It works when the page is refreshed properly - https://codesandbox.io/s/snowy-water-687dxg

The error on X Payroll might be due to some other race condition. I can take a look

anilprzp:
@saurabhdaware here are the references - 
[...]

PR #1729, opened 2023-10-18
fix(Dropdown): enhance tag selection handling

snitin315:
Fix https://github.com/razorpay/blade/issues/1721

### DIFF
--- a/packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
+++ b/packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
@@ -520,6 +520,81 @@ describe('<Dropdown />', () => {
     Object.defineProperty(HTMLElement.prototype, 'clientWidth', { configurable: true, value: 0 });
   });
 
+  // https://github.com/razorpay/blade/issues/1721
+  it('should handle controlled props & disabled options with multi select', async () => {
+    const ControlledDropdown = (): React.ReactElement => {
+      const [currentSelection, setCurrentSelection] = React.useState<string[]>([]);
+      const [isDropdownOpen, setIsDropdownOpen] = React.useState(false);
+
+      return (
+        <>
+          <Button
+            onClick={() => {
+              if (!currentSelection.includes('bangalore')) {
+                setCurrentSelection([...currentSelection, 'bangalore']);
+              }
+            }}
+          >
+            Select Bangalore
+          </Button>
+          <Button onClick={() => setIsDropdownOpen(!isDropdownOpen)}>Toggle Dropdown</Button>
+
+          <Dropdown
+            isOpen={isDropdownOpen}
+            onOpenChange={setIsDropdownOpen}
+            selectionType="multiple"
+          >
+            <SelectInput
+              label="Select City"
+              value={currentSelection}
+              onChange={(args) => {
+                if (args) {
+                  setCurrentSelection(args.values);
+                }
+              }}
+            />
+            <DropdownOverlay>
+              <ActionList>
+                <ActionListItem title="Delhi" value="delhi" isDisabled />
+                <ActionListItem title="Bangalore" value="bangalore" />
+                <ActionListItem title="Pune" value="pune" />
+                <ActionListItem title="Chennai" value="chennai" />
+              </ActionList>
+            </DropdownOverlay>
+          </Dropdown>
+        </>
+      );
+    };
+
+    // JSDOM doesn't calculate layouts so always return 0 https://github.com/testing-library/react-testing-library/issues/353#issuecomment-481248489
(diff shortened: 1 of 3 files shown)


packages/blade/src/components/Input/DropdownInputTriggers/BaseDropdownInputTrigger.tsx
```diff
@@ -194,7 +194,7 @@ const _BaseDropdownInputTrigger = (
       }
 
       return getTagsGroup({
-        tags: selectedIndices.map((selectedIndex) => options[selectedIndex].title),
+        tags: selectedIndices.map((selectedIndex) => options[selectedIndex]?.title),
```
kamleshchandnani, line 197: didn't quite understand the fix? if we mark an ActionListItem as disabled then  title prop doesn't exist?

snitin315: Yeah, I think so. It's also hard to reproduce locally. @saurabhdaware you may have more context here.

snitin315: We follow similar pattern at other places too:

https://github.com/razorpay/blade/blob/de82795aab4917d04be94c2b42605ae1f280c268/packages/blade/src/components/Dropdown/dropdownUtils.ts#L322

saurabhdaware: will try out locally and check

saurabhdaware: Will need to check their usage - https://github.com/razorpay/blade/issues/1721#issuecomment-1770084559

I wasn't able to replicate. The bug on sandbox is some react refresh dev-server issue which gets fixed on page refresh.

saurabhdaware: So the issue on sandbox is different from issue in actual code. They both end up failing on the same point and throw same error but reasons are different.

I think putting `?` should be fine here. It will stop the error. We can see if it works out for them. My guess is this will remove the error but still won't let them achieve what they are trying to achieve.

The issue is the options are dynamic in their case and seem to change their properties. Everytime the ActionListItem changes, the options need to be recalculated and can intermediately become unavailable.

saurabhdaware (approved):
Looks good to me. I don't see any harm of adding `?`. Need to check if this actually solves their issue though.

packages/blade/src/components/Dropdown/__tests__/Dropdown.web.test.tsx
```diff
@@ -520,6 +520,81 @@ describe('<Dropdown />', () => {
...
+              value={currentSelection}
+              onChange={(args) => {
+                if (args) {
+                  setCurrentSelection(args.values);
+                }
+              }}
+            />
+            <DropdownOverlay>
+              <ActionList>
+                <ActionListItem title="Delhi" value="delhi" isDisabled />
```
saurabhdaware, line 558: This test won't exactly test that github issue. This case is anyway working fine - https://codesandbox.io/s/snowy-water-687dxg?file=/App.tsx. It's just sandpack sometimes acts up. Things work fine when you refresh page.

Their issue is dynamic ActionListItem coming from options and then some properties are changing based on selection of items. A bit difficult to replicate that in tests.

snitin315:
@saurabhdaware it does solve the issue on codesandbox. you can [try here](https://61c19ee8d3d282003ac1d81c-otjcdpuwdz.chromatic.com/?path=/story/components-dropdown-with-select--with-multi-select) or [here](https://codesandbox.io/s/razorpay-blade-basic-9522tq)

saurabhdaware:
> @saurabhdaware it does solve the issue on codesandbox. you can [try here](https://61c19ee8d3d282003ac1d81c-otjcdpuwdz.chromatic.com/?path=/story/components-dropdown-with-select--with-multi-select) or [here](https://codesandbox.io/s/razorpay-blade-basic-9522tq)

Yup yup I meant,
- Having question mark would remove that error because the input becomes undefined and we print nothing. 

  Although this also means that it is getting called with `options` as undefined or changed options. Basically its a generic error that gets thrown when options change dynamically. So we can't exactly pinpoint why that error was thrown as options can change dynamically for various reasons. That case mentioned in sandpack, was sandpack specific where options were changing dynamically because of dev-server.

- In the actual Payroll example, options seem to change dynamically because they mark certain options disabled on selection of item.

So your change will fix and remove that error but options being undefined might be causing unexpected behaviour. I think they can take a look at that in their codebase. If there's something else that needs changed from our end, we can create separate PR.

We can merge this one after reviews.

snitin315:
I am merging this as a small change.

PR #1730, opened 2023-10-19
docs: add Table API Decisions

chaitanyadeorukhkar:
### [View Formatted File](https://github.com/razorpay/blade/blob/docs/table-api/packages/blade/src/components/Table/_decisions/decisions.md#api)

> This API has already been reviewed by a few frontend leads and the decisions have been made in consensus with them

### Summary by CodeRabbit

- New Feature: Added search functionality to the UI, allowing users to easily search for specific items.
- Bug Fix: Resolved an issue where the application would crash when loading large datasets.
- Documentation: Updated the README file with detailed instructions on how to set up and configure the application.
- Refactor: Optimized the database query logic for improved performance.
- Style: Applied consistent formatting and naming conventions throughout the codebase.
- Test: Added unit tests to ensure the correctness of critical functions.
- Chore: Upgraded dependencies to the latest versions to improve security and compatibility.
- Revert: Rolled back a previous change that caused unintended side effects.


packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,375 @@
...
+
+<Table 
+    data={data} // This contains the actual data to be rendered in the table which would be retrieved from some API
+    selectionType='single' 
+    onSelectionChange={onSelectionChange} 
+    sortFunctions={sortFunctions} 
+    onSortChange={onSortChange}
+    isLoading={false}
+    cellDensity='normal'
+    showZebraStripes={true}
```
kamleshchandnani, line 108: we shall have a better name for this

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,375 @@
...
+// Composable Table
+
+<Table 
+    data={data} // This contains the actual data to be rendered in the table which would be retrieved from some API
+    selectionType='single' 
+    onSelectionChange={onSelectionChange} 
+    sortFunctions={sortFunctions} 
+    onSortChange={onSortChange}
+    isLoading={false}
+    cellDensity='normal'
```
kamleshchandnani, line 107: we don't have a way of customizing density per cell right? so this shall be row density rather? also maybe we can call it something more simple like row spacing

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,375 @@
...
+        <TableCell>
+          {tableData.firstName}
+        </TableCell>
+        <TableCell>
+          {tableData.lastName}
+        </TableCell>
+        <TableCell>
+          {tableData.balance}
+          <Badge>Low</Badge>
+        </TableCell>
```
kamleshchandnani, line 150: to enhance the reading the devs in real will put it as iteration right? so it won't look that verbose

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,375 @@
...
+  console.log(selectedItems);
+};
+
+const columnDefinition = [
+    {
+        header: 'First Name',
+        headerKey: 'firstName',
+        cell: (row) => row.firstName,
+        isSticky: true,
+        columnStart: undefined,
```
kamleshchandnani, line 221: what does columnStart and columnEnd mean?

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,375 @@
...
+};
+
+const onSelectionChange = (selectedItems) => {
+  console.log(selectedItems);
+};
+
+const columnDefinition = [
+    {
+        header: 'First Name',
+        headerKey: 'firstName',
```
kamleshchandnani, line 218: what is headerKey required for?

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,375 @@
...
+        columnStart: 1,
+        columnEnd: 4,
+    },
+    {
+        footer: computedTotalBalance,
+    }
+];
+
+const pagination = {
+  showRowCountPicker?: boolean
```
kamleshchandnani, line 264: what is row count picker?

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,375 @@
...
+- Most popular table libraries like AG Grid, React Table, etc. use a compact API which is battle tested and developers are more familiar with the same
+- It would be relatively easier to swap out the table library in the future if we decide to do so since the API design is similar to other popular table libraries
+
+##### Cons
+- Compact API is less intuitive and does not resemble the native HTML structure of a table
+- It is harder to extend and add new feature to the table component since it is a single component
+- We would have to import the entire table component even if we are using only a few features of the table resulting in a higher bundle size for the consumer
+- Internally, we would have to use a composable API to create the compact API anyway since the library we have chosen has a composable API
+
+#### Decision
```
kamleshchandnani, line 308: can you also add what API are our existing products using?

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,375 @@
...
+
+### What are we looking for in a library?
+We had a few requirements that we were looking for in a library:
+- Well maintained and has a good community around it
+- Reasonable bundle size
+- Supports all the features that we need today and in the future (including features marked out of scope)
+- Flexible styling to ensure we can match the Razorpay's design language
+- Flexible API to ensure we can build both the Composable & Compact APIs
+
+### Libraries evaluated
```
kamleshchandnani, line 349: compare the table libraries by putting them under a table 😄

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: This section will always be verbose. If you see there is no "iteration" happening anywhere for the data. This defining what a single row looks like so balance will always have a badge in this case

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: Its for grouping cells. Especially in footer, one cell could span multiple columns like the last column shown in the image here. I will add more details in the props table
Screenshot 2023-10-25 at 5 18 07 PM

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: It decides how many rows should be shown per page
Screenshot 2023-10-25 at 5 19 27 PM

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,541 @@
...
+ 
+| Project         | Composable API Table Instances | Compact API Table Instances |
+| --------------- | ------------------------------ | --------------------------- |
+| PG Dashboard    | 101                            | 81                          |
+| Admin Dashboard | 19                             | 196                         |
+| X Dashboard     | 6                              | 11                          |
+
+
+##### Blade Team's Recommendation
+**We recommend** going ahead with the **Compact API** since it gives us the following advantages:
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: We can still do `onClick` and custom rendering based on cell value with Compact API as well. There will be feature parity between both approaches. I don't foresee anything on composable that can't be done on compact and vice versa

```tsx
const columns = [{
        header: 'First Name',
        cell: (row) => <TableCell onClick={console.log}>row.firstName</TableCell>,
    }]
```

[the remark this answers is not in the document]
nickisaacs: yup yup, its possible. We had to do the same thing you've shown here.

The composable API looked more cleaner than this.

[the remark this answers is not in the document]
chaitanyadeorukhkar: Yeah makes sense. On the projects you work on right now can you see if developers are writing composable or compact tables?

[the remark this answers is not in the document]
chaitanyadeorukhkar: @anilprzp would love to hear your inputs and experience here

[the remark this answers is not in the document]
nickisaacs: Couple of more things I thought of:

- attaching refs to a table row. Since the row is internal to the table we won't be able to do that directly
- colspan attributes (I suppose this can be solved if the cell function allows only table cell)

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,541 @@
...
+- React Table is possibly our 2nd best option if we ever need more complex tables in the future but as of now it would be an overkill for us
+
+#### Small sized library - React Table Library
+- [React Table Library (@table-library/react-table-library)](https://github.com/table-library/react-table-library) is the best example of a small sized library. It has all the features we need and is relatively well maintained.
+- It contains all the features we need today as well as the ones we can forsee needing in the future
+- As of today, React Table Library's bundle size is 23kb (7kb gzipped).
+- React Table Library has a Composable API using which we can build our Composable as well as Compact API
+- React Table Library also exposes a minimal Compact API but we cannot use that directly to build our Compact API since there isn't a parity between their own Composable & Compact APIs
+- React Table Library is our best option as of today since it has all the features we could need and is relatively lightweight
+
```
divyanshu013, line 535: We should perhaps also consider built-in virtualization for tables, otherwise consumers may have to do the wiring themselves 🤔

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,541 @@
...
+##### Blade Team's Recommendation
+**We recommend** going ahead with the **Compact API** since it gives us the following advantages:
+
+- It is the most popular approach within the Table libraries ecosystem which makes it easier for us to swap out the table library in the future
+- It is easier to consume since it is more concise and requires less effort to create a table
+- TypeScript would be able to help us with type checking the API better since it is a single component
+- Learning curve is lesser since its a single component and not a set of components that need to be composed together
+- As per our findings, 70% of the tables at Razorpay are built using the Compact API which would make it relatively easier for us to migrate them to Blade's Table component as well
+
```
divyanshu013, line 364: Was going through tanstack/table and looks like they also switch to a sort of "composable API" pattern for customizations ([example](https://tanstack.com/table/v8/docs/examples/react/editable-data)), would we need this kind of support as well? Possibly requiring us sometimes to create the table structure manually with table primitives (`th`, `tbody`, etc.) instead of relying on `Table` component.

Sooner than later, usecases will evolve so I like the pattern from tanstack table

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,541 @@
...
+| [AG Grid](https://www.ag-grid.com/)                                         | Very heavy and has a lot of features we don't need            | 75kb (15kb gzipped) (+ additional supporting libraries) | Challenging to build | Easy to build | Extremely well maintained. Has paid Enterprise support. |
+| [Tanstack's React Table](https://react-table.tanstack.com/)                 | Relatively heavy and has a lot of features we don't need      | 60kb (14kb gzipped)                                     | Challenging to build | Easy to build | Very well maintained                                    |
+| [React Table Library](https://github.com/table-library/react-table-library) | Relatively lightweight and has all the features we could need | 23kb (7kb gzipped)                                      | Easy to build        | Easy to build | Relatively well maintained                              | ✅          |
```
divyanshu013, line 541: What about tables for native? Can we use the same libraries? Any other things that may change (eg. virtualization)

packages/blade/src/components/Table/_decisions/decisions.md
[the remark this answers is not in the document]
chaitanyadeorukhkar: colspan should be easily solvable. Refs could be tricky but I'm wondering why we need refs on individual rows? Won't that be _a lot_ of refs

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: As long as we use proper pagination, virtualization might not be needed at the moment. But there is support for virtualization for all the table libraries we have evaluated and selected.

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: Tanstack has a very weird API. They're sorta composable and compact at the same time. They seem to have done this to enable framework-agnostic APIs. They do have a massive learning curve though.

Do you forsee any use-cases where we need composable and compact might not work?

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: We're scoping Table only for web & mweb. Native has lists instead of tables. Will add this to the doc

packages/blade/src/components/Table/_decisions/decisions.md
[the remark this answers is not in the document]
nickisaacs: Was thinking of cases where we might need to :
- reorder rows by drag and drop
- calculate the height of the row for certain animations

packages/blade/src/components/Table/_decisions/decisions.md
divyanshu013: Mostly for customizability (eg like supporting the above editable data usecase)

chaitanyadeorukhkar: We could have an `isEditable` & `onTextChange` prop in the columns definition prop for each row which should solve the same with compact API as well, right?

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,541 @@
...
+    - [Medium sized library - Tanstack's React Table](#medium-sized-library---tanstacks-react-table)
+    - [Small sized library - React Table Library](#small-sized-library---react-table-library)
+    - [Comparison](#comparison-1)
+
+
+## Features
+- Column Sorting
+- Row Selection - Single & Multiple
+- Pagination
+- Bulk Actions Toolbar
```
anilprzp, line 41: We should allow toolbar for each row as well. I can see in an illustration below, row level toolbar is shown as its own separate `Actions` column at the end of all columns. This will not be an optimal design for horizontally scrolling tables as the `Actions` column will be hidden away at the very end.

Instead the row level toolbar can be shown as an hover element when cursor is on top of the row. This allows user to perform inline actions instantly.

Screenshot 2023-10-30 at 11 55 09 AM

packages/blade/src/components/Table/_decisions/decisions.md
divyanshu013: Yes, should work for this but for more customizations we'll have to introduce new set of props then

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -56,6 +56,8 @@ We don't have enough use-cases for the following features at Razorpay and hence
 - Nested Tables
 - Hiding Columns
 - Editable Rows
+- Table for React Native
+  - Mobile Apps will be using Lists instead of Tables to convey information
```
divyanshu013, line 60: Is this confirmed from their side? There could be potential table related UIs on mobile

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,543 @@
...
+  firstName: (array) => array.sort((a, b) => a.name.localeCompare(b.name)),
+};
+
+const onSelectionChange = (selectedItems) => {
+  console.log(selectedItems);
+};
+
+const columnDefinition = [
+    {
+        header: 'First Name',
```
anilprzp, line 223: There can be instances where the column header itself might have to be a react component to support a tooltip or any other visual aid.

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,543 @@
...
+- It is the most popular approach within the Table libraries ecosystem which makes it easier for us to swap out the table library in the future
+- It is easier to consume since it is more concise and requires less effort to create a table
+- TypeScript would be able to help us with type checking the API better since it is a single component
+- Learning curve is lesser since its a single component and not a set of components that need to be composed together
+- As per our findings, 70% of the tables at Razorpay are built using the Compact API which would make it relatively easier for us to migrate them to Blade's Table component as well
+
+##### Conclusion
+[To be concluded]
+
+### Final API
```
anilprzp, line 370: Although API is not yet final, but would be great to have loading and refetch states as well for the table. This have a very good use case when pairing table with react-query style data fetching libraries. Below is a loading state illustration and for refetching a progress bar on top can be indicative of a background refresh.
Screenshot 2023-10-30 at 1 14 29 PM

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: The design team did confirm. I will still tag @shridhar52war @nikhil-rzp here

shridhar52war: Table components are generally not used. Alternative to the table component for mobile design cards, lists, or expandable sections are used usually. It's a design call.

packages/blade/src/components/Table/_decisions/decisions.md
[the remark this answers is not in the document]
chaitanyadeorukhkar: Do we have any such use-cases in our products?

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: That would be true even with composable API... we will have to add new props on other table components instead of the single table component

packages/blade/src/components/Table/_decisions/decisions.md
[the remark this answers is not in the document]
nickisaacs: I remember we needed to calculate the height of a row for some reason. But I can't remember what the use case was.

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: Loading state is already a part of the API. Do we have any such use-cases right now where we have progress bar on top or refetch mechanism in our products?

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: Its not equal spacing on all sides so calling it something like `rowSpacing='spacing.5'` is misleading. We could call it `rowDensity` instead of `cellDensity` though

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: Required for sorting internally. But this can be something like `key` or `id`. Some unique identifier

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: The problem with such a design is that it hides crucial information like arrears and additions. This will be use-case specific. We can't make a generalized decision to hide the last two rows to show an action. Action can always have a place on the top toolbar or it can be a dedicated column that can be fixed on the right side.

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: That is allowed. Header can be a string or JSX

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,543 @@
...
+    toolbarItemsSelectedTitle={`${selectedUsersCount} Users selected`}
+    toolbarActions={<Button>Export</Button>}
+    isStickyHeader={true}
+    isStickyFooter={true}
+/>
+```
+
+##### Pros
+- Compact API is more concise and requires less effort to create a table
+- Most popular table libraries like AG Grid, React Table, etc. use a compact API which is battle tested and developers are more familiar with the same
```
pcprashant, line 305: Has React table etc. used compact API to reduce verbosity only or there are other reasons ?

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,543 @@
...
+
+##### Pros
+- Composable API is more intuitive and resembles the native HTML structure of a table
+- It is easier to understand since the API design is similar to rest of the Blade components
+- It is easier to extend and add new feature to individual table components
+- We could leverage tree-shaking to only import the components that are being used in the consumer's table
+
+##### Cons
+- Composable API is more verbose and requires more effort to create a table by composing together multiple components
+- Most popular table libraries like AG Grid, React Table, etc. use a compact API which is battle tested and developers are more familiar with the same
```
pcprashant, line 194: React table is headless table library, You have to compose your own table using the data provided from React table. Not sure this falls into Compact API.

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,543 @@
...
+      labelPosition='left'
+      showRowCountPicker
+      rowCount='25'
+      onPaginationChange={console.log}
+      onRowCountPickerChange={console.log}
+      /> 
+   </>
+  )
+}
+</Table>
```
pcprashant, line 183: Can you add an example of column grouping etc. which will cause verbosity issues ?

packages/blade/src/components/Table/_decisions/decisions.md
pcprashant: https://codesandbox.io/p/sandbox/github/tanstack/table/tree/main/examples/react/column-groups?embed=1&file=%2Fsrc%2Fmain.tsx%3A1%2C1 Can we replicate this example?

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: Using APIs that define columns and layout instead of components that compose columns and layouts easier when you need to support multiple frameworks which is what react table does

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: It sort of does both compact and composable. It uses the ideas of compact API to create column objects and then the consumers have to map over it and compose their tables. Its not _truly_ composable in a sense that it doesn't expose composable components hence I categorized it as compact API

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: Column grouping isn't what causes verbosity issues. Any table, even a simple one will be less verbose with compact API compared to composable API

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,663 @@
...
+- Sticky Header
+- Sticky Footer
+- Cell Density - Normal & Comfortable
+
+## Out of scope
+We don't have enough use-cases for the following features at Razorpay and hence scoped them out of our Table component. If you are aware of any of these features being used at Razorpay, please let us know by creating an issue - we will evaluate it and scope it for future releases.
+
+- Column Reordering
+- Column Resizing
+- Column Filtering
```
cseas, line 73: Filtering is pretty common. If Blade doesn't standardise a pattern, different teams will keep coming up with different designs for filters.

Screenshot 2023-11-07 at 2 09 28 PM

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,663 @@
...
+- Search
```
cseas, line 74: Almost all merchant dashboard tables have need of search. Same concern as above, do we want to keep search controlled by external input fields? How do we maintain design/UX consistency this way?

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,663 @@
...
+| --------------------------------------------------------------------------- | ------------------------------------------------------------- | ------------------------------------------------------- | -------------------- | ------------- | ------------------------------------------------------- | ---------- |
+| [AG Grid](https://www.ag-grid.com/)                                         | Very heavy and has a lot of features we don't need            | 75kb (15kb gzipped) (+ additional supporting libraries) | Challenging to build | Easy to build | Extremely well maintained. Has paid Enterprise support. |
+| [Tanstack's React Table](https://react-table.tanstack.com/)                 | Relatively heavy and has a lot of features we don't need      | 60kb (14kb gzipped)                                     | Challenging to build | Easy to build | Very well maintained                                    |
```
cseas, line 657: Doesn't tree shaking eliminate bundle size concerns? 🤔 
If we know a tree shakeable library has a lot of features we don't need, the bundle size for us is going to be a lot lesser than the full package as well, right?

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: We explored this but standardising this is quite tricky for now since every product has a very specific use-case. This was discussed with design leadership as well and they concluded to keep this out of scope and focus only on the core table element. We can explore this in the future though.

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: Considering the worst-case scenario here. Since we have not spent a lot of time with these libraries we don't know the nuances and limitations of their tree-shaking abilities - what can and can't be tree-shaked. That is anyway not the only criteria in deciding a library. All our use-cases are met with the chosen library and it allows us a higher level of flexibility in terms of designing our API.

cseas (dismissed):

packages/blade/src/components/Table/_decisions/decisions.md
```diff
@@ -0,0 +1,659 @@
...
+- [Tanstack's React Table](https://react-table.tanstack.com/) is the best example of a medium sized library. It has all the features we need and is very well maintained. However, it is also relatively heavy and has a lot of features that we don't need.
+- As of today, React Table's bundle size is 60kb (14kb gzipped).
+- Tanstack's React  Table has been around for a longer duration and is extremely well maintained. It has relatively less features than AG Grid but also has a lot of features that we don't need.
+- Tanstack's React Table is built in a framework agnostic manner which meant they had to use a lot of abstractions to ensure it works well with all the frameworks. This makes it harder to understand, maintain and extend the library.
+- Replicating the Composable API would be a challenging task with React Table since it internally uses API similar to the Compact API. We would have to build a lot of abstractions to ensure we can build the Composable API on top of the Compact API.
+- React Table is possibly our 2nd best option if we ever need more complex tables in the future but as of now it would be an overkill for us
+
+#### Small sized library - React Table Library
```
sudhakar-selva, line 641: chore: did we explore https://react-spectrum.adobe.com/react-aria/Table.html? seems to be offering headless component but bundle size could be of higher end(based on imported components)

packages/blade/src/components/Table/_decisions/decisions.md
abhieshekumar: @anilprzp @chaitanyadeorukhkar folks would this be a use case?

https://github.com/razorpay/blade/assets/24494390/e07fdcf0-70f6-45c8-b596-e577d80582c0

packages/blade/src/components/Table/_decisions/decisions.md
abhieshekumar: @anilprzp @chaitanyadeorukhkar folks will this be a use case?

https://github.com/razorpay/blade/assets/24494390/8cd34b40-a723-413c-9c1b-4b2c0a72ff55

chaitanyadeorukhkar: I will forward this to design and check.

packages/blade/src/components/Table/_decisions/decisions.md
chaitanyadeorukhkar: Seems similar to the libraries we have evaluated already. Anything specific about this library that you think would be helpful?

PR #1737, opened 2023-10-21
fix(ActionListItem): Width based Truncation 

### LINKED ISSUE #1699: Allow width of action list to more than Dropdown button
As per the screenshot, the names in the list can be longer requiring the need for the width of dropdown list to be more than the dropdown button.
One way is to increase width of DropdownButton, but that takes more space in the filters form.
If there could be a way to handle this width manually/automatically so that contents are visible in dropdown.
Screenshot 2023-10-06 at 11 40 57 PM (1)

Saksham-13:
## Description

This pull request adds a new prop to the `Text` component called `shouldTruncate` which takes in a boolean value. When `shouldTruncate` is set to `true`, the `Text` component will apply CSS to truncate the text if it exceeds the container's width.

The `shouldTruncate` prop is then set to `true` for the `ActionListItem` component to enable text truncation for the dropdown overlay.

## Changes Made

- Added a new prop called `shouldTruncate` to the `Text` component.
- Added CSS to the `Text` component to truncate text if it exceeds the container's width and the shouldTruncate prop is flagged true.
- Set `shouldTruncate` to `true` for the `ActionListItem` component.

## Screenshots

image

## Related Issue

Closes #1699

### DIFF
--- a/packages/blade/src/components/Typography/BaseText/getBaseTextStyles.ts
+++ b/packages/blade/src/components/Typography/BaseText/getBaseTextStyles.ts
@@ -12,6 +12,7 @@ const getBaseTextStyles = ({
   fontStyle = 'normal',
   textDecorationLine = 'none',
   numberOfLines,
+  shouldTruncate,
   lineHeight = 100,
   textAlign,
   theme,
@@ -22,6 +23,7 @@ const getBaseTextStyles = ({
   const themeFontWeight = theme.typography.fonts.weight[fontWeight];
   const themeLineHeight = makeTypographySize(theme.typography.lineHeights[lineHeight]);
   let truncateStyles: CSSObject = {};
+  let truncateWidthStyles: CSSObject = {};
   if (numberOfLines !== undefined) {
     if (isReactNative()) {
       truncateStyles = {};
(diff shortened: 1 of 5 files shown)


packages/blade/src/components/Typography/BaseText/getBaseTextStyles.ts
```diff
@@ -35,6 +37,18 @@ const getBaseTextStyles = ({
       };
     }
   }
+  if(shouldTruncate !== undefined) {
```
saurabhdaware, line 40: We have `truncateAfterLines={1}` prop on Text which is expected to do the similar thing. Any idea why that one is not working? can that prop be used instead of new prop?
https://codesandbox.io/s/7gtqwk?file=/App.tsx&utm_medium=sandpack

saurabhdaware (commented):
Hiiii! Thanks for the PR. Added one comment below

saurabhdaware:
Hi @Saksham-13 , thanks for the PR. We're closing it for now since it has been stale and there's several conflicts. But let me know if you decide to work on it again, we can reopen.

Thanks for the efforts!

PR #1738, opened 2023-10-22
feat(TextInput): add onClick prop

### LINKED ISSUE #1560: onClick for the TextInput component
We are adding a datepicker component. For this we are using the TextInput component to be the trigger to open the datepicker. Unfortunately, there is no onClick handler right now to open the datepicker.

The onFocus prop is not sufficient, as we don't want to trigger the datepicker when the user lands on the input component via keyboard navigation. It should trigger either for `onClick` or `onFocus + enter key`.

The 2nd trigger can be achieved by us by using onFocus to add an keyboard event listener for an enter key before showing the datepicker, but we need the onClick for mouse users too.

rk9595:
Resolves #1560.

I have passed onClick prop of BaseInput to TextInput.
Hope this is the right way.
Wrote a test too.
Please let me know of any changes, thanks.

### DIFF
--- a/packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
+++ b/packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
@@ -167,7 +167,9 @@ const _StyledBaseInput: React.ForwardRefRenderFunction<
       onInput={(event: React.ChangeEvent<HTMLInputElement>) => {
         handleOnInput?.({ name, value: event });
       }}
-      onClick={handleOnClick}
+      onClick={(event) => {
+        handleOnClick?.({ name, value: event });
+      }}
       autoCapitalize={autoCapitalize}
       {...commonProps}
       {...props}
--- a/packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
+++ b/packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
@@ -129,10 +129,6 @@ const _StyledBaseInput: React.ForwardRefRenderFunction<
       : undefined,
   };
 
-  // const handleInputClick = (event: React.MouseEvent<HTMLInputElement>): void => {
-  //   handleOnClick?.({ name, value: event });
-  // };
-
   return props.as === 'button' ? (
     <StyledBaseNativeButton
       // @ts-expect-error: TS doesnt understand that this will always be `button`
--- a/packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
+++ b/packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
@@ -129,6 +129,10 @@ const _StyledBaseInput: React.ForwardRefRenderFunction<
       : undefined,
   };
 
+  // const handleInputClick = (event: React.MouseEvent<HTMLInputElement>): void => {
+  //   handleOnClick?.({ name, value: event });
+  // };
+
   return props.as === 'button' ? (
     <StyledBaseNativeButton
       // @ts-expect-error: TS doesnt understand that this will always be `button`
@@ -167,6 +171,7 @@ const _StyledBaseInput: React.ForwardRefRenderFunction<
       onInput={(event: React.ChangeEvent<HTMLInputElement>) => {
         handleOnInput?.({ name, value: event });
       }}
+      onClick={handleOnClick}
       autoCapitalize={autoCapitalize}
       {...commonProps}
       {...props}
--- a/packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
+++ b/packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
@@ -173,7 +173,6 @@ describe('<TextInput />', () => {
     const { getByLabelText } = renderWithTheme(<TextInput label={label} onClick={onClick} />);
 
     const input = getByLabelText(label);
-    console.log('input', input);
     await userEvent.click(input);
     //should be called for onClick
     expect(onClick).toHaveBeenCalledTimes(1);
--- a/packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
+++ b/packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
@@ -167,6 +167,18 @@ describe('<TextInput />', () => {
     expect(onFocus).toHaveBeenCalledWith({ name, value: userName });
   });
 
+  it('should handle onClick', async () => {
+    const onClick = jest.fn();
+    const label = 'Enter name';
+    const { getByLabelText } = renderWithTheme(<TextInput label={label} onClick={onClick} />);
+
+    const input = getByLabelText(label);
+
+    await userEvent.click(input);
+    //should be called for onClick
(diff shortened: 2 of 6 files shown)


packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
```diff
@@ -167,6 +167,18 @@ describe('<TextInput />', () => {
...
+  it('should handle onClick', async () => {
+    const onClick = jest.fn();
+    const label = 'Enter name';
+    const { getByLabelText } = renderWithTheme(<TextInput label={label} onClick={onClick} />);
+
+    const input = getByLabelText(label);
+
+    await userEvent.click(input);
+    //should be called for onClick
+    expect(onClick).toHaveBeenCalledTimes(1);
```
saurabhdaware, line 179: This test seems to fail 🤔 
image

saurabhdaware (commented):
Thanks for the PR 🥳. Added one comment below

packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
```diff
@@ -167,6 +167,18 @@ describe('<TextInput />', () => {
     expect(onFocus).toHaveBeenCalledWith({ name, value: userName });
   });
 
+  it('should handle onClick', async () => {
```
anuraghazra, line 170: should probably also add a test for react-native.

packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
rk9595: Ohh, will look into it.

packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
rk9595: Noted

packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
rk9595: On debugging the issue, it seemed the input element wasn't receiving the onClick prop. Is it because of event bubbling?
I guess the main DOM input element is in StyledBaseInput , so added the onClick handler to StyledBaseNativeInput. Ran the test, they seem to be passing. I used watch:test:react to test only the changed files. Let me know if it resolves the issue. Thanks. 
Screenshot 2023-10-24 at 9 48 22 PM

rk9595: Sorry about the 2 silly commits , forgot to remove the console.logs and unnecessary code snippets.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
```diff
@@ -167,6 +167,7 @@ const _StyledBaseInput: React.ForwardRefRenderFunction<
       onInput={(event: React.ChangeEvent<HTMLInputElement>) => {
         handleOnInput?.({ name, value: event });
       }}
+      onClick={handleOnClick}
```
saurabhdaware, line 170: aha right! earlier we allowed onClick only on button renders. Makes sense to add this. Can you also add similar thing in react native file? I see its missing there as well.

packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
saurabhdaware: haha that's ok we don't really care much about commits since they are squashed and merged in master while merging. Feel free to commit as per your comfort :D

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
rk9595: Sure, need to do a brief research on ReactNative, never used it.

rk9595: added the test for ReactNative, which is failing as it is not able to receive onClick prop. 
Tried some approaches like wrapping StyledNativeBaseInput  with `TouchableOpacity <TouchableOpacity  
      onPress={(): void => {
        onClick?.({ name, value: props.value }); 
      }}
    >
      <StyledNativeBaseInput`, but lot of the other tests are failing.
And how do I render TextInput as Button? I thought to add the test at least for the Button renders , but there is no prop such as 'as' to mention the type as 'button' or 'input'.

saurabhdaware: I think we can pass it down to `onFocus` of TextInput 

https://stackoverflow.com/a/54708179

I don't think so there is any other clean way of achieving this on React Native 😢

rk9595: Ohh, got it.

rk9595: Hey @saurabhdaware @anuraghazra Not really sure how to proceed ahead on this issue for React native. Would appreciate some specific steps to resolve the issue. And from the original issue description, the second use case for onFocus+keyboard seems to be handled. Thanks.

snitin315: @rk9595 We have [StyledBaseInput.native.tsx](https://github.com/razorpay/blade/blob/48a4d4aaaf73840f00932e4f7d190d0102b642e9/packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx) file as well which would change things fir react native.

You can follow our [contributing guide](https://github.com/razorpay/blade/blob/master/CONTRIBUTING.md#contributing-guidelines-for-blade) to setup react native dev environment locally.

rk9595:
Ok, noted.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
saurabhdaware: Hi @rk9595, sorry for the late review. We were also not sure of the solution so had to pull and figure out few changes. I have made a PR to your fork with additional changes for react native and tests for it. You can merge those changes and they will get reflected in this PR - https://github.com/rk9595/blade/pull/1

saurabhdaware:
Hi @rk9595, I have created PR here to handle clicks on RN. You can merge that PR to have those changes reflected here - https://github.com/rk9595/blade/pull/1

saurabhdaware (approved):
Yay thanks! Looks good to me.

@rk9595  Can you fill this form if you haven't already? https://docs.google.com/forms/d/e/1FAIpQLSeiH-sDUDrmULyEqxCLEk3ZpEC1nE-L9PoJ3Urfn4UVffrsKg/viewform

rk9595:
Hey, thanks! I have filled the above form in October itself.

saurabhdaware:
> Hey, thanks! I have filled the above form in October itself. But haven't received any communication from Razorpay.

Oh cool. Yup we're in the process of finalising and dispatching goodies. Thanks

PR #1742, opened 2023-10-23
fix(AutoComplete, SelectInput): disable tags in input when `isDisabled` is set to true

### LINKED ISSUE #1739: Dropdown: validate functionality of isDisabled in multi select dropdown
- There is a use case to display the options that were selected by user at some point in time as read-only.
- Have tried using `isDisabled` prop on `<SelectInput/>` and eventually on `<ActionListItem/>`
- However unable to get the desired behaviour.
- Please help get this validated.
- [Codesandbox](https://codesandbox.io/s/dropdown-disabled-queries-kxplvx?file=/App.tsx)

saurabhdaware:
Resolves #1739 

Marks tags inside the input and value of the input as disabled when `isDisabled` is passed

### DIFF
--- a/packages/blade/src/components/Dropdown/docs/DropdownWithSelect.stories.tsx
+++ b/packages/blade/src/components/Dropdown/docs/DropdownWithSelect.stories.tsx
@@ -299,6 +299,29 @@ export const InternalSelect = (): React.ReactElement => {
   );
 };
 
+export const InternalDisabledSelect = (): React.ReactElement => {
+  const [isDisabled, setIsDisabled] = React.useState(false);
+  return (
+    <Box padding="spacing.5" maxWidth="400px">
+      <Button marginBottom="spacing.4" isFullWidth onClick={() => setIsDisabled(!isDisabled)}>
+        Toggle Input
+      </Button>
+      <Dropdown selectionType="multiple">
+        <SelectInput label="Select fruits" isDisabled={isDisabled} />
+        <DropdownOverlay>
+          <ActionList>
+            <ActionListItem title="Apples" value="Apples" />
+            <ActionListItem title="Appricots" value="Appricots" />
+            <ActionListItem title="Cherries" value="Cherries" />
+            <ActionListItem title="Crab apples" value="Crab apples" />
+            <ActionListItem title="Jambolan" value="Jambolan" />
+          </ActionList>
+        </DropdownOverlay>
+      </Dropdown>
+    </Box>
+  );
+};
+
 export const InternalAutoPositioning = (): React.ReactElement => {
   return (
     <Box>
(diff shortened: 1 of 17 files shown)


packages/blade/src/components/Input/DropdownInputTriggers/InputChevronIcon.native.tsx
```diff
@@ -14,12 +14,12 @@ const InputChevronIcon = (props: SelectChevronIconProps): React.ReactElement =>
     <StyledChevronIconContainer accessibilityLabel="Chevron Icon" onPress={props.onClick}>
       <ChevronDownIcon
         display={props.isOpen ? 'none' : 'flex'}
-        color="surface.text.normal.lowContrast"
+        color={`surface.text.${props.isDisabled ? 'placeholder' : 'normal'}.lowContrast`}
```
anuraghazra, line 17: nit: extract to a variable in the component body.

anuraghazra (dismissed):
Mostly lgtm.

packages/blade/src/components/Input/DropdownInputTriggers/InputChevronIcon.native.tsx
saurabhdaware: done. Also moved this into separate component so it can be reused in web and native both

anuraghazra (dismissed):

packages/blade/src/components/Dropdown/__tests__/Dropdown.native.test.tsx
```diff
@@ -48,6 +48,24 @@ describe('<Dropdown />', () => {
...
+        <DropdownOverlay>
+          <ActionList>
+            <ActionListItem title="Mango" value="mango" />
+            <ActionListItem title="Apple" value="apple" />
+          </ActionList>
+        </DropdownOverlay>
+      </Dropdown>,
+    );
+
+    expect(getByTestId('dropdown-overlay').props.display).toBe('none');
```
kamleshchandnani, line 64: we shall not do test id unless absolutely necessary? the query by selector aren't helping in this case?

packages/blade/src/components/Dropdown/docs/DropdownWithSelect.stories.tsx
```diff
@@ -299,6 +299,29 @@ export const InternalSelect = (): React.ReactElement => {
   );
 };
 
+export const InternalDisabledSelect = (): React.ReactElement => {
+  const [isDisabled, setIsDisabled] = React.useState(false);
+  return (
+    <Box padding="spacing.5" maxWidth="400px">
+      <Button marginBottom="spacing.4" isFullWidth onClick={() => setIsDisabled(!isDisabled)}>
+        Toggle Input
```
kamleshchandnani, line 307: shouldn't the caption be Toggle Disabled State?

packages/blade/src/components/Dropdown/__tests__/Dropdown.native.test.tsx
saurabhdaware: There isn't any `.toBeVisible` equivalent in React Native that works as expected so I am checking the `display` prop.

Now the display prop is set on a `View` that doesn't have any accessibility role as such so to point to that exact div, we have to rely on getByTestId in React Native.

In Web, we can just get the ActionListItem element inside and check `toBeVisible` on it.

packages/blade/src/components/Dropdown/docs/DropdownWithSelect.stories.tsx
saurabhdaware: Yup right. Resolved in 0eaae714b9264c398313a222891e18e4053750be

kamleshchandnani (approved):

PR #1744, opened 2023-10-25
feat(Typography): accept `brand.secondary.500` as valid `color` prop

snitin315:


packages/blade/src/components/Typography/BaseText/types.ts
```diff
@@ -42,6 +42,7 @@ export type BaseTextProps = {
   id?: string;
   color?:
     | BrandPrimaryColors
+    | 'brand.secondary.500'
```
saurabhdaware, line 45: Lets expose all secondary shades like primary? they might need different shade at some place. Anyway we'll have to expose everything in future since brand refresh uses secondary colors a lot.

snitin315: @saurabhdaware Currently [we have only `500` in secondary.](https://blade.razorpay.com/?path=/story/guides-intro--page)

Screenshot 2023-10-25 at 12 40 39 PM

anuraghazra: Extract the type from the theme object instead of hardcoding it, otherwise we might forget to update later: 

```js
type BrandSecondaryColors = `brand.secondary.${keyof Theme['colors']['brand']['secondary']}`;
```

snitin315: @saurabhdaware @anuraghazra Done, refactored.

saurabhdaware (approved):

anuraghazra (approved):

PR #1784, opened 2023-11-02
fix(blade): fix jsdom & ssr test setup

anuraghazra:
Fixes JSDOM & SSR testing setup. 

**Changes:** 

- Added `pretendToBeVisual` option to JSDOM contractor - this is needed otherwise JSDOM doesn't mock requestAnimationFrame and other browser utilties. That's why our tests fails and we had to do [manual mocks.](https://github.com/razorpay/blade/blob/2d07f56cfbfc9e4337f12d65c92328dcd3f6c7db/packages/blade/src/components/Dropdown/__tests__/Dropdown.ssr.test.tsx#L37-L41) 
- Copied all JSDOM globals to Node globals - This was a major issue because of this while rendering we were getting errors that some window.method is not defined: For example in BottomSheet floatingUI was throwing [navigator isn't defined](https://github.com/razorpay/blade/blob/2d07f56cfbfc9e4337f12d65c92328dcd3f6c7db/packages/blade/src/components/BottomSheet/__tests__/BottomSheet.ssr.test.tsx#L14) error, which was fixed after this change. 
- Put a `try-catch` wrapper in RTL hydration function and mocked console.warn/error to ensure we catch any hydration mismatch errors. 

**Review Tip:** Filter out snapshots on github when reviewing the code, the CSR snapshots were updated to include proper styles - I think previously the styles were not getting calculated because our JSDOM setup was broken, now that is fixed styled-component started working. 

### Summary by CodeRabbit

- Bug Fix: Fixed JSDOM and SSR testing setup in multiple components (`BottomSheet`, `Dropdown`, `Tabs`, `ChipGroup`, `AutoComplete`, `Box`) by adding the `pretendToBeVisual` option to the JSDOM constructor, copying JSDOM globals to Node globals, wrapping the RTL hydration function in a try-catch block, and mocking console.warn/error. These changes aim to resolve errors related to undefined window methods and catch hydration mismatch errors.
- Refactor: Modified the rendering of the `<Box>` component by adding CSS classes and modifying the display property. Also removed test cases related to error handling for unsupported values and props.
- Test: Updated snapshots in `useId.ts` to include proper styles for JSDOM and SSR testing setup fixes.


packages/blade/src/components/Box/__tests__/Box.ssr.test.tsx
```diff
@@ -61,9 +73,17 @@ describe('<Box />', () => {
       );
     } catch (err: unknown) {
       expect(err).toMatchInlineSnapshot(`
-        [Error: [Blade: Box]: Oops! Currently you can only use \`transparent\`, \`surface.background.*\`, and \`brand.*\` tokens with backgroundColor property but we received \`red\` instead.
+        [Error: The above error occurred in the <Box> component:
 
-         Do you have a usecase of using other values? Create an issue on https://github.com/razorpay/blade repo to let us know and we can discuss ✨]
+            at backgroundColor (/Users/anurag.hazra/Documents/razorpay/projects/blade/packages/blade/src/components/Box/Box.tsx:222:13)
+            at children (/Users/anurag.hazra/Documents/razorpay/projects/blade/packages/blade/src/components/BottomSheet/BottomSheetStack.tsx:26:3)
```
snitin315, line 79: This path is specific to your machine. need to normalize to check for substring only.

packages/blade/src/utils/testing/renderWithSSR.web.tsx
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-implicit-any-catch */
```
snitin315, line 2: cover4_new_1200x768

anuraghazra: image

packages/blade/src/utils/testing/renderWithSSR.web.tsx
```diff
@@ -20,17 +25,43 @@ const renderWithSSR = (ui: ReactElement): RenderResult => {
   );
 
   const { JSDOM } = jsdom;
-  const dom = new JSDOM(html);
-  // @ts-expect-error ignoring this error since we are trying to make tests work between node and jsdom environments
-  global.window = dom.window;
-  global.document = dom.window.document;
+  const dom = new JSDOM(html, { pretendToBeVisual: true, url: 'http://localhost/' });
```
kamleshchandnani, line 28: was this the only issue?

anuraghazra: Check the PR description the first two points that I've mentioned were majorly the issue.

kamleshchandnani (approved):

kamleshchandnani (approved):

packages/blade/src/components/Box/__tests__/Box.ssr.test.tsx
anuraghazra: Oops! checking. 

This should even error like this.

anuraghazra: Thanks for catching this, removed this tests, we should not even need to test for these since this exact test are already present on the .web.ts files. 

SSR test should contain SSR related flows & smoke tests.

packages/blade/src/utils/useId.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable react-hooks/rules-of-hooks */
```
snitin315, line 2: Why this? there are no changes in this file which would lead to eslint errors.

snitin315 (dismissed):
LGTM. One comment.

packages/blade/src/utils/useId.ts
anuraghazra: removed.

snitin315 (approved):

PR #1787, opened 2023-11-02
fix(blade): remove extra margin left on carousel

anuraghazra:
Fixes: https://github.com/razorpay/blade/issues/1788

### Summary by CodeRabbit

- Style: Modified the `marginLeft` property of the `StyledCarouselItem` component in `CarouselItem.web.tsx` to have a value of `40%` instead of `0` when the `shouldHaveStartSpacing` prop is true. This change improves the spacing of carousel items and enhances the visual presentation for the end-user.

### DIFF
--- a/packages/blade/src/components/Carousel/CarouselItem.web.tsx
+++ b/packages/blade/src/components/Carousel/CarouselItem.web.tsx
@@ -34,7 +34,7 @@ const StyledCarouselItem = styled(BaseBox)<StyledCarouselItemProps>(
       ...(isResponsive && {
         width: '100%',
         scrollSnapAlign: isMobile || !shouldAddStartEndSpacing ? 'start' : 'center',
-        marginLeft: shouldHaveStartSpacing ? '100%' : 0,
+        marginLeft: shouldHaveStartSpacing ? '40%' : 0,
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/Carousel/CarouselItem.web.tsx
```diff
@@ -34,7 +34,7 @@ const StyledCarouselItem = styled(BaseBox)<StyledCarouselItemProps>(
       ...(isResponsive && {
         width: '100%',
         scrollSnapAlign: isMobile || !shouldAddStartEndSpacing ? 'start' : 'center',
-        marginLeft: shouldHaveStartSpacing ? '100%' : 0,
+        marginLeft: shouldHaveStartSpacing ? '40%' : 0,
```
kamleshchandnani, line 37: what was the reason of this miss? was it scroll gesture?

anuraghazra: This only happens if you use trackpad to purposely scroll past the left side too much. 

Otherwise not an experience breaker.

kamleshchandnani (approved):

PR #1792, opened 2023-11-06
fix: defaultValidator is not fn

saurabhdaware:


saurabhdaware:
This breaks in their frontend bundler but works when used with vite in node template. Closing this PR for now because we went with node template for our examples.

PR #1793, opened 2023-11-06
docs: Tour Component API Decisions

anuraghazra:
### Summary by CodeRabbit

- New Feature: Implemented a search functionality in the user interface, allowing users to easily search for specific items.
- Bug Fix: Fixed a critical bug that caused the application to crash when processing large datasets.
- Documentation: Updated the API documentation with comprehensive examples and improved clarity.
- Refactor: Refactored the codebase to improve modularity and maintainability.
- Style: Applied consistent coding style throughout the project for better readability.
- Test: Added additional unit tests to improve test coverage and ensure code reliability.
- Chore: Upgraded dependencies to their latest versions to leverage new features and security patches.
- Revert: Reverted a previous change that introduced a regression in performance.


packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,718 @@
...
+  /**
+   * Array of steps to be rendered
+   *
+   * The order of the steps will be the order in which they are rendered depending on the `activeStep` prop
+   */
+  steps: Step[];
+  /**
+   * Whether the tour is visible or not
+   */
+  isOpen: boolean;
```
chaitanyadeorukhkar, line 85: Should we just call it `isVisible?`

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,718 @@
...
+   */
+  activeStep: number;
+}
+```
+
+### `TourStep` API
+
+TourStep is an enhancer component which is used to wrap the element that needs to be highlighted with a specific unique identifier.
+
+This component is needed because in react-native there is no `id` prop or `getElementById` platform API, but to keep the API consistent between web & native, we will also expose this component for web.
```
chaitanyadeorukhkar, line 109: Also because we don't expose id from our components

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,718 @@
...
+```
+
+## Usage
+
+```jsx
+import { Tour } from '@razorpay/blade/components';
+import type { TourSteps } from '@razorpay/blade/components';
+
+const Footer = ({ gotToNext, goToPrev, stopTour, activeStep, totalStep }) => {
+  const isLast = activeStep === totalStep;
```
chaitanyadeorukhkar, line 132:
```suggestion
  const isLast = activeStep === totalSteps;
```

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,718 @@
...
+
+And on the web, there was a problem that even though simply adding `id` could work, not all our components had `id` prop. That means consumers will have to wrap everything with a `Box`.
+
+- We decided **not** to go with the `id` approach on web, and instead go with the same implementation as the [react-native POC](https://github.com/razorpay/blade/compare/master...anu/tour-rn-poc#diff-4fe985a90d9ce955346ffc61e152a98272f4d02e044111d30241eb63c8fcf1b1R168-R183), where we keep track of refs of the elements via the `TourStep` component.
+  - Pros:
+    - API is now same on both RN & Web
+    - No need to wrap everything with Box, consumers can use the TourStep enhancer component.
+    - Simplified implementation, since now we don't need to maintain two types of implementation 1 for native (with refs) 1 for web (with ids).
+
+**2. Tour needs to wrap the whole app:**
```
chaitanyadeorukhkar, line 360: Its needs to wrap on screen levels right? Not on App level

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,718 @@
...
+### Approach-1: Prefix `id` to avoid conflicts
+
+It will look like this:
+
+```jsx
+
+import { Tour } from '@razorpay/blade/components';
+import type { TourSteps } from '@razorpay/blade/components';
+
+const globalSteps: TourSteps = [
```
chaitanyadeorukhkar, line 392: This section needs to be updated based on recent discussions

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,718 @@
...
+      <Button id="paymentlink-step-1">Click me</Button>
+      <Box id="paymentlink-step-2" padding="spacing.5">
+        <Text>Some content</Text>
+      </Box>
+    </Box>
+  );
+};
+```
+
+### Approach-2: Use `TourStep` to avoid conflicts
```
chaitanyadeorukhkar, line 456: Needs to be updated based on discussions

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,718 @@
...
+  <Button>Click me</Button>
+</TourStep>;
+
+// file: PaymentLinks.tsx
+<TourStep tourKey="paymentlinks" id="step-1">
+  <Button>Click me</Button>
+</TourStep>;
+```
+
+### Technical Challenge in React Native
```
chaitanyadeorukhkar, line 481: Why is this a challenge? We will wrap it on screen level and that should be fine right?

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: Yeah I initially called it visible only, but prefer to use isOpen terminology since it's already a well defined nomenclature in all our components, Tooltip, Popover etc might as well go with same for Tour.

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: Yes will add this point.

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,718 @@
...
+> NOTE: The below sections are discussions, decisions & challenges we faced while designing the API for the `Tour` component.
+> If you are interested in diving deeper into the rabbit hole, you can read the below sections, or else you can skip them.
+
+## Open Questions And Technical Challenges
+
+## React Native Specifics
+
+On react-native there is 2 differences in the API compared to web:
+
+1. `TourStep` component is needed.
```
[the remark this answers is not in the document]
anuraghazra: I've kept these discussion points for archiving and future reference, you can check the "conclusions" after each section where I talked about the updated decisions. 

I've also added note above that this section is largely for discussions & future reference and can be skipped if when reviewing this API doc.

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: Same as https://github.com/razorpay/blade/pull/1793#discussion_r1383101447

Just kept it for future reference, check the "conclusions" below.

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: Check the conclusion section below: 

> **Conclusion:**
> Given the complexity of solving this, we decided that:
> - Consumers can wrap the `<Tour />` component closer to the module they need the tour for, instead of wrapping the whole `App` in a single tour. And we will also document this on the storybook.
> - There may not be that many cases for multiple tour flows in the same page

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: It's still a technical challenge, because if consumer wants to add two different Tours in the same React subtree they will have to wrap 1 tour over another which will cause issues since React.Context will take the nearest context. 

```
<Tour tourKey="global">
  <Tour tourKey="payments">
    <Example1 /> // this will get all the states from the `payments` tour
    <Example2 /> // this will also get all the states from the `payments` tour
  </Tour>
</Tour>
```

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,723 @@
+# Tour API Decisions <!-- omit in toc -->
+
+The Tour component is used to provide context as well as enable users to take certain actions on it. These are used to highlight a new feature or provide a guided tour to a new user.
+
+<img src="./assets/tour-thumbnail.png" width="380" alt="Tooltip Thumbnail" />
+
+- [Design](#design)
+- [Features](#features)
+- [API](#api)
```
kamleshchandnani, line 9: Put references to Popover API decision?

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,723 @@
...
+    - [**2. How will we have two different tour flows in the same page?**](#2-how-will-we-have-two-different-tour-flows-in-the-same-page)
+    - [3. Segmentation / Lack of control](#3-segmentation--lack-of-control)
+    - [4. Not possible to maintain a consistent API / Implementation between web \& native](#4-not-possible-to-maintain-a-consistent-api--implementation-between-web--native)
+    - [Conclusion](#conclusion)
+- [MoMs](#moms)
+- [References](#references)
+
+## Design
+
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=40540-559304&t=tmTrf3xJU6oj59fM-0) to all variants of the Popover component
```
kamleshchandnani, line 32:
```suggestion
[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=40540-559304&t=tmTrf3xJU6oj59fM-0) to all variants of the Tour component
```

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,723 @@
...
+
+## API
+
+### `Tour` API
+
+```jsx
+// Step will have similar props as the Popover component, With extra Tour related props.
+type Step = {
+  /**
+   * Unique identifier for the tour step
```
kamleshchandnani, line 47: Maybe just mention identifier? the moment we say unique it tells me that i need to use some random number or uuid to have this identifier?

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,723 @@
...
+   */
+  name: string;
+  /**
+   * Content of the Popover
+   */
+  content: ({ gotToNext, goToPrev, stopTour, activeStep, totalSteps }) => React.ReactNode;
```
kamleshchandnani, line 53:
```suggestion
  content: ({ gotToNext, goToPrevious, stopTour, activeStep, totalSteps }) => React.ReactNode;
```

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,723 @@
...
+[Figma Link](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade---Payment-Light?type=design&node-id=40540-559304&t=tmTrf3xJU6oj59fM-0) to all variants of the Popover component
+
+## Features
+
+- [x] **Tour:** A guided tour with multiple steps
+- [x] **Masking:** Mask the rest of the page except the highlighted element.
+
+## API
+
+### `Tour` API
```
kamleshchandnani, line 41: before diving directly into the API can you describe in few words the composition of Tour from a consumer lens basically put an example usage? and then describe in 2-3 lines? doesn't has to be a detailed example like below usage section but something that is abstracted and only explains about the composition of tour

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,723 @@
...
+  const handleStepChange = (step: number) => {
+    setActiveStep(step);
+    // do something with the step
+    // ie: send an analytic event
+  };
+
+  return (
+    <Box>
+      <Button onClick={() => setIsOpen(true)}>Show Tour</Button>
+      <Tour
```
kamleshchandnani, line 211: can't see `<DashboardPage/>` wrapped around `<Tour/>`?

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: We should mention "unique", since it has to be unique, you cannot have duplicate identifiers in the steps array.

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: Done.

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: Updated.

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: right, fixed.

kamleshchandnani (dismissed):

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,766 @@
...
+
+**A few points to consider:**
+
+- This is absolutely needed on `react-native`
+- Technically we can also expose this component for `web` too and make the API surface same, although it will largely be useless because on web we can simply use the `id` prop to query the element and highlight it.
+
+The `web` implementation will look like this:
+
+```jsx
+const TourStep = ({ id, children }) => {
```
sudhakar-selva, line 374: chore: when using ref, are we planning to merge the ref from TourStep and component prop? might be useful for cases where ref is used on underlying children.

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,766 @@
...
+   */
+  titleLeading?: React.ReactNode;
+  /**
+   * Placement of Popover
+   * @default "top"
+   */
+  placement?: UseFloatingOptions['placement'];
+}
+
+type TourProps = {
```
cseas, line 111: Figma design has 3 variations of footer buttons
1. single
2. double with text
3. double with icons

Screenshot 2023-11-07 at 9 42 54 PM

Aren't these going to be part of the component/API? Looks like the Footer part of the design is arbitrary and left for consumers to build as they like. This might lead to inconsistency in spacing, button sizes, icons, etc.

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,766 @@
...
+
+Check/Provide feedback for the motion for Tour component [here](https://razorpay.slack.com/archives/C0274H7QRC1/p1697520879941439)
+
+https://github.com/razorpay/blade/assets/35374649/5830c059-fbd6-461e-915c-e3a98e930735
+
+---
+
+---
+
+> NOTE: The below sections are discussions, decisions & challenges we faced while designing the API for the `Tour` component.
```
divyanshu013, line 306: Might be good to move to a different md file

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,766 @@
...
+  const isLast = activeStep === totalStep;
+  const isFirst = activeStep === 0;
+  return (
+    <Box>
+      <Text>
+        {activeStep} / {totalStep}
+      </Text>
+      {!isFirst && <Button onClick={gotToNext}>Prev</Button>}
+      {isLast ? <Button onClick={stopTour}>Done</Button> : <Button onClick={goToPrev}>Next</Button>}
+    </Box>
```
divyanshu013, line 187: Will also need to add spacings here

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,766 @@
...
+  );
+};
+
+const steps: TourSteps = [
+  {
+    name: 'step-1',
+    title: 'Step 1',
+    content: ({ activeStep }) => (
+      <Box>
```
divyanshu013, line 196: Do we need a Box wrapper always? Might not be intuitive for consumers

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,766 @@
...
+   */
+  isOpen: boolean;
+  /**
+   * Callback when the tour is opened or closed
+   */
+  onOpenChange: ({ isOpen: boolean }) => void;
+  /**
+   * Callback which fires when the tour has reached the last step.
+   */
+  onFinish: () => void;
```
divyanshu013, line 129: Seems bit redundant with `onStepChange` in the picture

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,766 @@
...
+TourStep is an enhancer component which is used to wrap the element that needs to be highlighted with a specific unique identifier.
+
+This component is needed because:
+
+- in react-native there is no `id` prop or `getElementById` platform API.
+- We don't expose `id` for every blade component, currently only Box does.
+
+But to keep the API consistent between web & native, we will also expose this component for web.
+
+> Note: on web, this component doesn't attach an `id` to the DOM element, instead it uses `ref` to collect the DOM element and save it to the state inside the `Tour` component, See below [discussions](#react-native-specifics) to know more about this approach.
```
divyanshu013, line 152: > uses `ref` to collect the DOM element and save it to the state inside the `Tour` component

Interesting 🤯 

We store the state in a ref? Not sure about implementation but can there be any state sync issues if root of `Tour` re-renders?

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: Yes, we will ensure that consumer's ref doesn't get overriden. We will merge consumers ref and our internal ref together.

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: No you don't need any wrapper. I'll remove these unneeded boxes.

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: Actually, the JSDOC is wrong. 

This was bit of a gray area in terms of differences between onStepChange vs onFinish when i started writing the API. 

But after I've done the implementation it's more clear now, 
The usecase is creating interruptible tours, this is an existing usecase in Dashboard:  

There can be possibility that you need to interrupt the tour or skip the tour in the middle, so when when you call `stopTour` function from the `steps` render prop only then this onFinish will get called. 

```
const steps = [
  {
    name: 'step-1',
    content: ({ stopTour }) => <Button onClick={stopTour}>Skip Tour</Button>
  }
]

<Tour onFinish={() => "Called when stopTour is called"} />
```

Check this example of [interruptible tour](https://github.com/razorpay/blade/pull/1802/files#diff-2ad1c41c9aaa02282767d4d760cd3ff77f7d54c09b0b905a2b866be807b04a33R908-R994)

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: > We store the state in a ref

Yes, store the refs in the state. 

> Not sure about implementation but can there be any state sync issues if root of Tour re-renders?

We did this POC initially for ReactNative first because on RN we have to do this way since there is no `id` or `getElementById` and it worked out pretty well. 

You can check the POC implementation of it [here](https://github.com/razorpay/blade/compare/master...anu/tour-rn-poc#diff-4fe985a90d9ce955346ffc61e152a98272f4d02e044111d30241eb63c8fcf1b1R163-R183): 

Basic idea is that we store the refs in state in a map so that we can query those refs by it's `name` 

```jsx
// map of refs
// eg: { 'step-1': domNodeRef }
const [refIdMap, setRefIdMap] = React.useState(new Map<string, React.RefObject<Element>>());
```

And the `TourStep` component ensures to properly attach and remove the refs from the Map: 

```jsx
const TourStep = ({ name, children }) => {
  const ref = React.useRef(null);
  const { attachStep, removeStep } = React.useContext(TourContext);

  React.useEffect(() => {
    if (!ref) return;
    attachStep(name ref);

    return () => {
      removeStep(name);
    };
  }, [ref, attachStep, id, removeStep]);

  const child = children as any;
  return React.cloneElement(child, { ...child.props, ref });
};
```

But later on realised that this can also be used in web, and there were few advantages to this too: 

Pros:
- API is now same on both RN & Web
- No need to wrap everything with Box, consumers can use the TourStep enhancer component. (Since most components from blade exposes `ref` but only Box component exposes `id`)
- Simplified implementation, since now we don't need to maintain two types of implementation 1 for native (with refs) 1 for web (with ids).

Check the [conclusion](https://github.com/razorpay/blade/blob/anu/tour-api/packages/blade/src/components/Tour/_decisions/decisions.md#react-native-specifics) of react native specifics section to read more on why we approached it this way. 


> Not sure about implementation but can there be any state sync issues if root of Tour re-renders?

We've implemented this already didn't saw any weird issues with state desync. Once we save those refs on state those DOM nodes stay `live` so didn't faced any issues with this.

packages/blade/src/components/Tour/_decisions/decisions.md
```diff
@@ -0,0 +1,766 @@
...
+// Step will have similar props as the Popover component, With extra Tour related props.
+type Step = {
+  /**
+   * Unique identifier for the tour step
+   */
+  name: string;
+  /**
+   * Content of the Popover
+   */
+  content: ({ gotToNext, goToPrev, stopTour, activeStep, totalSteps }) => React.ReactNode;
```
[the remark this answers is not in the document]
anuraghazra: > Shouldn't description be a mandatory prop with an optional

So there's actually no `description` prop on Figma. 

This description text is there for designer's ease of use, so that each time they use the Popover they don't have to pull in a Text component to put on the slot. 
Designers can remove this default description text add anything they want. So on dev it's just the content `children`

packages/blade/src/components/Tour/_decisions/decisions.md
anuraghazra: @divyanshu013 @cseas discussed this & updated the API. 

We've decided to expose a new component `TourFooter` which consumers can use to achieve a consistent footer design. 

And we've also kept the same `footer` render prop API for flexibility so that if there's any special usecase based on the product consumers can have the option to create their own footer.

cseas (approved):

kamleshchandnani (approved):

PR #1797, opened 2023-11-07
feat: Add `Table` component

chaitanyadeorukhkar:
Adds Table Component

Documentation: #1834


packages/blade/src/components/Table/Table.native.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export type TableNode<Item> = Item & {
+  id: string;
+};
+
+export type TableData<Item> = {
+  nodes: TableNode<Item>[];
+};
+
+export type TableProps<Item> = {
```
anuraghazra, line 16: Let's have types.ts file and put all the reusable types there? So you don't have to copy these types for RN file again?

packages/blade/src/components/Table/Table.web.tsx
```diff
@@ -0,0 +1,539 @@
...
+};
+
+// Get the number of TableHeaderCell components.
+// This is very complicated but the only way to iterate through the structure and get number of header cells.
+// Assuming number of header cells is the same as number of columns
+const getTableHeaderCellCount = (children: (data: []) => React.ReactElement): number => {
+  const tableRootComponent = children([]);
+  if (tableRootComponent && React.isValidElement(tableRootComponent)) {
+    const tableComponentArray = React.Children.toArray(tableRootComponent);
+    if (React.isValidElement(tableComponentArray[0])) {
```
anuraghazra, line 156: nit: see if we can use guard clauses / early returns to reduce the indentation here. 
Nested ifs right now making the code hard to read.

packages/blade/src/components/Table/TableBody.web.tsx
```diff
@@ -0,0 +1,346 @@
...
+          '& .row-select-single-selected:nth-child(even) td, .row-select-selected:nth-child(even) td ': {
+            backgroundColor: getIn(theme.colors, tableRow.stripe.backgroundColorSelected),
+          },
+          '& .row-select-single-selected:nth-child(even):hover:not(.disabled-row) td, .row-select-selected:nth-child(even):hover:not(.disabled-row) td ': {
+            backgroundColor: getIn(theme.colors, tableRow.stripe.backgroundColorSelectedHover),
+          },
+          '& .row-select-single-selected:nth-child(even):focus:not(.disabled-row) td, .row-select-selected:nth-child(even):focus:not(.disabled-row) td ': {
+            backgroundColor: getIn(theme.colors, tableRow.stripe.backgroundColorSelectedFocus),
+          },
+          '& .row-select-single-selected:nth-child(even):active:not(.disabled-row) td, .row-select-selected:nth-child(even):active:not(.disabled-row) td ': {
```
anuraghazra, line 120: Where are these class names come from? the react table library?

packages/blade/src/components/Table/Table.web.tsx
```diff
@@ -0,0 +1,539 @@
...
+  `
+  }`
+    : '';
+  const firstColumnStickyFooterCellCSS = isFirstColumnSticky
+    ? `
+  &:nth-of-type(1) {
+    left: 0 !important;
+    position: sticky !important;
+    z-index: ${firstColumnStickyHeaderFooterZIndex} !important;
+  }
```
anuraghazra, line 251: 🤔 we are using strings here now? instead of objects?

anuraghazra:
https://github.com/razorpay/blade/assets/35374649/605d804e-331f-4ce2-b6b3-07c1af64bfc7

In the pagination the forward button should also change the icon when focusing.

packages/blade/src/components/Table/TableBody.web.tsx
```diff
@@ -0,0 +1,346 @@
...
+  surfaceLevel: SurfaceLevels;
+}>(({ theme, isSelectable, surfaceLevel }) => ({
+  '&&&': {
+    height: '100%',
+    backgroundColor: getIn(theme.colors, `surface.background.level${surfaceLevel}.lowContrast`),
+    '& > div:first-child': {
+      pointerEvents: isSelectable ? 'none' : 'auto',
+      alignSelf: 'stretch',
+    },
+    '&: focus-visible': {
```
anuraghazra, line 169:
```suggestion
    '&:focus-visible': {
```

packages/blade/src/components/Table/TablePagination.web.tsx
```diff
@@ -0,0 +1,422 @@
...
+const pageSizeOptions: NonNullable<TablePaginationProps['defaultPageSize']>[] = [10, 25, 50];
+
+const PageSelectionButton = styled.button<{ isSelected?: boolean }>(({ theme, isSelected }) => ({
+  backgroundColor: isSelected
+    ? getIn(theme.colors, tablePagination.pageSelectionButton.backgroundColorSelected)
+    : 'transparent',
+  border: 'none',
+  cursor: 'pointer',
+  height: '32px',
+  width: '32px',
```
anuraghazra, line 71: nit: can be moved to the tokens file as constant + get the 32px from `size` object.

packages/blade/src/components/Table/Table.native.tsx
chaitanyadeorukhkar: Done

packages/blade/src/components/Table/TableBody.web.tsx
chaitanyadeorukhkar: Yup

packages/blade/src/components/Table/Table.web.tsx
chaitanyadeorukhkar: This is a hook given by react table library and it only accepts a string :(

packages/blade/src/components/Table/TablePagination.web.tsx
chaitanyadeorukhkar: Done

packages/blade/src/components/Table/TableBody.web.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+const StyledRow = styled(Row)<{ isSelectable: boolean; showStripes?: boolean }>(
+  ({ theme, isSelectable }) => ({
+    '&&&': {
+      backgroundColor: 'transparent',
+      ...(isSelectable && {
+        '&:hover:not(.disabled-row) td': {
+          transition: `background-color ${makeMotionTime(
+            getIn(theme.motion, tableRow.backgroundColorMotionDuration),
+          )} ${getIn(theme.motion, tableRow.backgroundColorMotionEasing)}`,
```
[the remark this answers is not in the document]
chaitanyadeorukhkar: Done

packages/blade/package.json
```diff
@@ -135,6 +136,7 @@
     "@babel/preset-react": "7.16.5",
     "@babel/preset-typescript": "7.16.5",
     "@codesandbox/sandpack-react": "1.16.0",
+    "@emotion/react": "11.11.1",
```
anuraghazra, line 139: Oh wait, does that mean now consumers will have bundle of both styled comp and emotion?

packages/blade/src/components/Table/TableFooter.web.tsx
```diff
@@ -0,0 +1,94 @@
...
+    },
+  },
+}));
+
+const _TableFooterCell = ({ children }: TableFooterCellProps): React.ReactElement => {
+  const { surfaceLevel } = useTableContext();
+  const isChildrenString = typeof children === 'string';
+  return (
+    <StyledFooterCell
+      $surfaceLevel={surfaceLevel}
```
anuraghazra, line 76: why the transient prop?

packages/blade/package.json
chaitanyadeorukhkar: Yup thats a down side of this library. Mentioned it in the API decision 😢 

Doesnt add much though

packages/blade/src/components/Table/TableFooter.web.tsx
chaitanyadeorukhkar: It adds it to the dom and you see lots of warnings in console

anuraghazra (dismissed):
minor comments

packages/blade/package.json
anuraghazra: I see, so does the library marks @emotion/react as peerDep? 

It should otherwise all the components will have to carry emotion bundle this is what I was exploring also. 

Also, let's move `@table-library/react-table-library` to peerDep, we don't need to include this on our bundle if a consumer doesn't use Table keeping in mind treeshaking/codesplitting things.

chaitanyadeorukhkar: Checking

packages/blade/package.json
```diff
@@ -242,6 +242,7 @@
   },
   "peerDependencies": {
     "@table-library/react-table-library": "4.1.7",
+    "@emotion/react": "11.11.1",
```
anuraghazra: Another thing we should do that we didn't on other peerDeps, is instead of pinning them we should use ranges, otherwise consumers will have to use the exact version down the the patch otherwise there will be mismatches.  

```
@emotion/react: ^11.11.1
```

chaitanyadeorukhkar: Yup, done!

PR #1803, opened 2023-11-09
feat(rfcs): Font Loading Strategy RFC

saurabhdaware:
[**✨ Formatted Version ✨**](https://github.com/razorpay/blade/blob/rfc/font-loading-strategy/rfcs/2023-11-08-font-loading-strategy.md)

### Summary by CodeRabbit

- Documentation: Added a new RFC (Request for Comments) document for the Font Loading Strategy in the Blade library. The document outlines different font loading strategies and provides code examples for self-hosted font loading, hosting fonts on a CDN, and loading fonts from existing third-party CDNs. This RFC aims to improve the understanding and selection of font loading strategies for developers using the Blade library. No changes were made to the code logic or functionality.


rfcs/2023-11-08-font-loading-strategy.md
```diff
@@ -0,0 +1,211 @@
...
+Similar to how we load Lato font from `@fontsource/lato` right now, Consumers add import from `@razorpay/blade/fonts.css`
+
+```js
+import '@razorpay/blade/fonts.css';
+```
+
+**Pros:**
+
+- Easy Implementation
+- Self-hosted so does not require additional DNS connection and thus fonts load faster
```
kamleshchandnani, line 146:
```suggestion
- Self-hosted so does not require additional DNS connection and thus fonts load faster(no preconnect required)
```

rfcs/2023-11-08-font-loading-strategy.md
```diff
@@ -0,0 +1,211 @@
...
+  - [Criteria for Font Loading Strategy](#criteria-for-font-loading-strategy)
+  - [Explored Solutions](#explored-solutions)
+- [Implementation Details](#implementation-details)
+- [References](#references)
+
+## Summary
+
+After Brand Refresh, we have 2 major fonts. `Inter` and `Tasa Orbiter`.
+
+So far when we had Lato, we could load it from `@fontsource/lato`. But Tasa Orbiter does not exist on Lato and we also selectively load weights from Inter and Lato.
```
kamleshchandnani, line 25:
```suggestion
So far when we had Lato, we could load it from `@fontsource/lato`. But Tasa Orbiter does not exist on fontsource and we also selectively load weights from Inter and Lato.
```

rfcs/2023-11-08-font-loading-strategy.md
```diff
@@ -0,0 +1,211 @@
...
+
+These are the weights and styles being used right now
+
+**Tasa Orbiter:** 400, 500, 600
+**Inter:** 400, 600
+
+**Conclusion**
+
+- We use Tasa Orbiter as variable font since its 32kb in size compared to ~90kb (30kb x 3 weights) of static font
+- We use Inter as static font. We use lighter version of Inter from [@fontsource/inter](https://www.npmjs.com/package/@fontsource/inter?activeTab=code) as static fonts there are 70kb each compared to ~200kb on Google Fonts version. The variable font of Inter from Google Fonts is ~800kb.
```
kamleshchandnani, line 191: can't we use variable fonts for inter too as for Tasa and store it locally like we do for tasa?

rfcs/2023-11-08-font-loading-strategy.md
```diff
@@ -0,0 +1,211 @@
...
+https://github.com/razorpay/blade/assets/30949385/b2b12cbe-2c93-4af9-bc7f-7b94ed36b90b
+
+https://github.com/razorpay/blade/assets/30949385/d468cfc1-c1a9-44ad-9d85-a3ea37320ce7 -->
+
+#### 2. Self Hosted: Using CSS Imports from Blade (✅ Preferred)
+
+Similar to how we load Lato font from `@fontsource/lato` right now, Consumers add import from `@razorpay/blade/fonts.css`
+
+```js
+import '@razorpay/blade/fonts.css';
```
kamleshchandnani, line 140: wouldn't this be render blocking when the consumers are using it as this will be at the beginning of the file?

rfcs/2023-11-08-font-loading-strategy.md
```diff
@@ -0,0 +1,211 @@
...
+
+</details>
+
+<!-- #### Difference between loading fonts without preload and with preload
+
+https://github.com/razorpay/blade/assets/30949385/b2b12cbe-2c93-4af9-bc7f-7b94ed36b90b
+
+https://github.com/razorpay/blade/assets/30949385/d468cfc1-c1a9-44ad-9d85-a3ea37320ce7 -->
+
+#### 2. Self Hosted: Using CSS Imports from Blade (✅ Preferred)
```
kamleshchandnani, line 135: can we mention about some perf boosts by adding approaches like service worker precache, etc.?

saurabhdaware: we cannot use it inside service worker in CSR apps. We won't know the CSS file URLs as they are randomly generated with hash on build.

rfcs/2023-11-08-font-loading-strategy.md
saurabhdaware: I guess bundler will load this css file and add it to HTML like a normal `<link rel="stylesheet"` so its same as normally linking CSS only
image

If you mean render-blocking in sense how normal stylesheet links are render-blocking in HTML, don't think so we can do much about it.

rfcs/2023-11-08-font-loading-strategy.md
saurabhdaware: we can. Its just heavier. Its 200kb vs ~140kb for addition of our required weights. So 60kb extra 🤔

saurabhdaware: Added a table here to represent information better

rfcs/2023-11-08-font-loading-strategy.md
kamleshchandnani: we can tie it with workbox precache webpack plugin and it will auto-add the hashes bit. [Ref](https://developer.chrome.com/docs/workbox/precaching-with-workbox/#side-by-side-comparison)

kamleshchandnani (approved):

PR #1814, opened 2023-11-10
feat(Collapsible): mark width prop as external

### LINKED ISSUE #1810: collapsible: make width for CollapsibleBody as 100%
1. Collapsible body doesn't seem to grow it's width, therefore it's content are unable to take up available width.
2. [Slack](https://razorpay.slack.com/archives/CMQ3RBHEU/p1699555218542609)
3. [Sandbox](https://codesandbox.io/s/eager-neco-shd5yk?file=/App.tsx)

abhieshekumar:
Resolves #1810 

- Changes prop `_width` to `width`


packages/blade/src/components/Collapsible/CollapsibleBody.tsx
```diff
@@ -23,7 +23,7 @@ const _CollapsibleBody = ({ children, testID, _width }: CollapsibleBodyProps): R
     <BaseBox
       id={collapsibleBodyId}
       // Just React Native things, need this 100% so collapsed content flows correctly inside Accordion
-      width={_width}
+      width={_width ?? '100%'}
```
abhieshekumar, line 26: ques: should it also have a check for platform? 

```js
width={isReactNative()? _width : '100%'}
```

saurabhdaware (commented):
Hey @abhieshekumar, we discussed this in team and we think allowing `width` as prop might be less of breaking change than having 100% width by default.


So we can rename that `_width` prop as `width` and mark it external and let people use it whenever they require a different width.

Setting 100% width can break the existing UIs so maybe its better to not set that. What do you think?

saurabhdaware (approved):

.changeset/strange-fans-impress.md
```diff
@@ -0,0 +1,14 @@
...
+---
+
+feat(Collapsible): mark width prop as external
+
+> [!WARNING]
+>
+> If you were using internal `_width` prop, you will have to rename it to `width`
+
+```diff
+- <CollapsibleBody _width={} />
```
kamleshchandnani, line 12: I see 17 instances of `_width` prop. this will be a breaking change for all these customers right?
Screenshot 2023-12-12 at 12 39 24 PM

saurabhdaware: @kamleshchandnani can you share link of this chart? I searched on github and it shows 5 files only https://github.com/search?q=org%3Arazorpay+CollapsibleBody+_width+NOT+repo%3Arazorpay%2Fblade&type=code

Also its an internal prop so having a minor release with warning note should be fine right?

kamleshchandnani: here: https://app.amplitude.com/analytics/razorpay-mobile/chart/zity1yl/edit/ifsok2i1

kamleshchandnani (approved):

PR #1819, opened 2023-11-14
chore(installation): add note for styled-components supported version

iamakshatjain:
Notifies blade users about the `styled-components` version (v5) supported by blade


saurabhdaware (approved):
Thanks :D

PR #1820, opened 2023-11-16
feat(Rebranding / Typography): Update Typography Components

saurabhdaware:
Stacked on top of #1811 

### Changes

- font-size and line-height changes for Text, Heading, Display, Code
- Remove variant prop from Heading
- Remove Title component
- Add letterSpacing to Display component
- Add weight prop to Display component


packages/blade/.storybook/react/global.css
```diff
@@ -2,3 +2,14 @@
   /* Need this to ensure mdx stories don't break visually */
   box-sizing: border-box;
 }
+
+body h1,
```
kamleshchandnani, line 6: only h1 should be enough right? it applies to h1 globally?

packages/blade/src/utils/storybook/StoryPageWrapper.tsx
```diff
@@ -101,9 +101,9 @@ const StoryPageWrapper = (props: StoryPageWrapperTypes): React.ReactElement => {
   return (
     <BladeProvider themeTokens={paymentTheme}>
       <WithGlobalStyles>
-        <Title size="xlarge" marginBottom="spacing.3">
+        <Display size="small" marginBottom="spacing.3">
```
kamleshchandnani, line 104: so title xlarge and large both point to Display small, wouldn't things look visually break for consumers once they migrate?

packages/blade/src/components/Typography/Text/Text.tsx
```diff
@@ -19,7 +18,7 @@ type TextCommonProps = {
   contrast?: ColorContrastTypes;
   truncateAfterLines?: number;
   children: React.ReactNode;
-  weight?: keyof Theme['typography']['fonts']['weight'];
+  weight?: Extract<BaseTextProps['fontWeight'], 'regular' | 'medium' | 'semibold'>;
```
kamleshchandnani, line 21: why not extract from typography itself?

packages/blade/src/components/Typography/Display/Display.tsx
```diff
@@ -18,6 +18,12 @@ export type DisplayProps = {
    */
   color?: BaseTextProps['color'];
   size?: Extract<BaseTextSizes, 'small' | 'medium' | 'large' | 'xlarge'>;
+  weight?: Extract<BaseTextProps['fontWeight'], 'regular' | 'medium' | 'semibold'>;
+  /**
+   *
+   * @default - 0% when weight is semibold, otherwise -1%
+   */
+  letterSpacing?: BaseTextProps['letterSpacing'];
```
kamleshchandnani, line 26: do we need to expose this to consumer? I thought we'll handle this internally always 🤔

packages/blade/.storybook/react/global.css
saurabhdaware: nope all headings use Tasa Orbiter no. So changed it for all hx tags. This is for Storybook's own styling btw.

packages/blade/src/components/Typography/Text/Text.tsx
saurabhdaware: Text extends on top of BaseText so taking from that is more stable no? image we add more typography font weights but not add it in BaseText. Since we pass down weight from Text to BaseText, there might be something that shows up in type of Text but is not handled on BaseText.

packages/blade/src/components/Typography/Display/Display.tsx
saurabhdaware: Aha right. We don't. Removed it https://razorpay.slack.com/archives/G01B3LQ9H0W/p1701170941397549

packages/blade/src/utils/storybook/StoryPageWrapper.tsx
saurabhdaware: ops nope. find and replace gone wrong. It should be Heading xlarge in other places. Fixed it for this and some other storybook heading size issues

kamleshchandnani (approved):
remove the `body h1`. rest looks good

packages/blade/src/components/BottomSheet/BottomSheet.stories.tsx
```diff
@@ -74,8 +74,8 @@ const Page = (): React.ReactElement => {
           'https://www.figma.com/file/sAdplk2uYnI2ILnDKUxycW/Blade---Banking-Dark?type=design&node-id=16767-706689&t=6FVL3Ha33gwM45Cm-0',
       }}
     >
-      <Title>Usage</Title>
-      <Sandbox editorHeight={600}>
+      <Heading size="large">Usage</Heading>
```
snitin315, line 77: I think the codemod will also need to fall back to `size="large"` if the size isn't explicitly defined.

snitin315 (approved):
LGTM

PR #1821, opened 2023-11-16
fix: set default breakpoint without computing

kamleshchandnani:
This was a nasty bug and can be discovered when we do SSR and open the application on mobile device. There was a class mismatch between server and client leading to mismatch in DOM trees between server and client

```js
const [breakpointAndDevice, setBreakpointAndDevice] = useState(() => {
    const matchedBreakpoint = getMatchedBreakpoint();
    const matchedDeviceType = getMatchedDeviceType(matchedBreakpoint);
    return {
      matchedBreakpoint,
      matchedDeviceType,
    };
  });
```

The above code will compute `matchedBreakpoint: undefined` and `matchedDeviceType: 'desktop'` when run on server and now since the useState is a function it gets evaluated again on the client side and now if the client side is being opened on mobile device the useState will return `matchedBreakpoint: s` and `matchedDeviceType: 'mobile'` so the initial render if it had anything that changes because the screensize eg: typography the DOM tree will be different between server and client


packages/blade/src/utils/useBreakpoint/useBreakpoint.web.test.ts
```diff
@@ -4,15 +4,15 @@ import { useBreakpoint } from './';
...
-  it('should detect mobile and return smallest breakpoint of mobile devices', () => {
+  it('should return default breakpoint and device', () => {
     setupMatchMediaMock({
       matches: true,
     });
 
     const { result } = renderHook(() => useBreakpoint({ breakpoints }));
 
-    expect(result.current.matchedBreakpoint).toBe('base');
-    expect(result.current.matchedDeviceType).toBe('mobile');
```
saurabhdaware, line 15: It should continue to detect mobile and base no? just not detect it on the initial client render 🤔

packages/blade/src/utils/useBreakpoint/useBreakpoint.ts
```diff
@@ -83,13 +83,9 @@ export const useBreakpoint = ({
...
 
-  const [breakpointAndDevice, setBreakpointAndDevice] = useState(() => {
-    const matchedBreakpoint = getMatchedBreakpoint();
-    const matchedDeviceType = getMatchedDeviceType(matchedBreakpoint);
-    return {
-      matchedBreakpoint,
-      matchedDeviceType,
-    };
+  const [breakpointAndDevice, setBreakpointAndDevice] = useState<BreakpointAndDevice>({
+    matchedBreakpoint: undefined,
```
saurabhdaware, line 87: Now that we're not calling those functions here, are we sure they are getting called somewhere in useEffect or useLayoutEffect to make sure correct device is set on load?

saurabhdaware (commented):
Looks good. Can you just check if it is working correctly for scenario where we do want mobile to be detected after initial load.

Also should we call those functions in useIsomorphicLayoutEffect now to detect it a bit earlier before layout shifts?

anuraghazra:
Let's add SSR test for useBreakpoint? 

Not sure why this mismatch wasn't caught in other components. also can you link the slack thread that sumit raised?

kamleshchandnani:
> Let's add SSR test for useBreakpoint?
> 
> Not sure why this mismatch wasn't caught in other components. also can you link the slack thread that sumit raised?

It wasn't caught because the target changes from SSR(desktop because node) to client(mobile). doesn't happen when you open the SSR app on desktop itself. I've mentioned that in the description.

adding SSR test won't help either here. Can't change the breakpoints from node(desktop) to client(mobile) in the jest

packages/blade/src/utils/useBreakpoint/useBreakpoint.web.test.ts
```diff
@@ -4,7 +4,7 @@ import { useBreakpoint } from './';
 import { breakpoints } from '~tokens/global';
 
 describe('useBreakpoint', () => {
-  it('should detect mobile and return smallest breakpoint of mobile devices', () => {
+  it('should return default breakpoint and device', () => {
```
saurabhdaware, line 7: its not returning default right? we're returning mobile and base

saurabhdaware (approved):
Looks good to me. Tiny change in test description I guess?

packages/blade/src/utils/useBreakpoint/useBreakpoint.web.test.ts
kamleshchandnani: oh yes this changed multiple times. will update

PR #1822, opened 2023-11-17
feat(Rebranding / Typography): Blade Internal Component Migrations

saurabhdaware:


packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
kamleshchandnani, line 1: why?

packages/blade/src/components/Badge/__tests__/Badge.web.test.tsx
```diff
@@ -59,7 +59,7 @@ describe('<Badge />', () => {
 
   it('should render Badge with bold font', () => {
     const label = 'LABEL';
-    const { container } = renderWithTheme(<Badge fontWeight="bold">{label}</Badge>);
+    const { container } = renderWithTheme(<Badge fontWeight="semibold">{label}</Badge>);
```
kamleshchandnani, line 62: no snapshots changed? 🤔

packages/blade/src/components/Amount/amountTokens.ts
saurabhdaware: We have not fixed Amount component. We will be picking up during individual component migration since its a bit more complex and needs other rework as well.

Making it ts-nocheck because otherwise the typescript fails and we don't get chromatic published to test out changes.

packages/blade/src/components/Badge/__tests__/Badge.web.test.tsx
saurabhdaware: yes. They are in this branch - https://github.com/razorpay/blade/tree/rebranding/typography-snapshots

I'll merge these changes first to `rebranding/master` and then update snapshots again

kamleshchandnani (approved):

snitin315 (commented):
Let's once run the codemod on Blade codebase itself to make sure we haven't missed any instance?

PR #1840, opened 2023-11-23
feat(rfc-font-loading-strategy): conclude to using variable fonts in both

saurabhdaware:
Since we're using 3 weights on inter now. The static font size becomes almost same as variable font. I think its better to load variable font in this scenario since it will give us more control and size difference is not that big.

### DIFF
--- a/rfcs/2023-11-08-font-loading-strategy.md
+++ b/rfcs/2023-11-08-font-loading-strategy.md
@@ -204,16 +204,16 @@ The performance of variable font vs static font depends on how many weights and 
 These are the weights and styles being used right now
 
 **Tasa Orbiter:** 400, 500, 600
-**Inter:** 400, 600
+**Inter:** 400, 500, 600
 
 **Conclusion**
 
-We'll use TASA Orbiter as variable font and Inter as static font based on which is the lightest version.
+We'll be using variable fonts in both Tasa and Inter based on the size below-
 
-|                  | **Static**                        | **Variable** |
-| ---------------- | --------------------------------- | ------------ |
-| **Inter**        | 70kb \* 2 weights = ✅ **~140kb** | ~200kb       |
-| **TASA Orbiter** | 30kb \* 3 weights = ~90kb         | ✅ **~32kb** |
+|                  | **Static**                 | **Variable**  |
+| ---------------- | -------------------------- | ------------- |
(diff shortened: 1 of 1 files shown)


rfcs/2023-11-08-font-loading-strategy.md
```diff
@@ -195,6 +195,16 @@ Cons from all of the above 🙈
 
 ## Implementation Details
 
+### Font Size Optimization
+
+To optimize on font-size, we are splitting fonts into multiple files.
+
+- inter-latin-rzp.woff2 (alphabets, numbers, common symbols, currency)
```
kamleshchandnani, line 202: can we name them as blade rather than rzp?

rfcs/2023-11-08-font-loading-strategy.md
```diff
@@ -204,16 +214,16 @@ The performance of variable font vs static font depends on how many weights and
...
-We'll use TASA Orbiter as variable font and Inter as static font based on which is the lightest version.
+We'll be using variable fonts in both Tasa and Inter based on the size below-
 
-|                  | **Static**                        | **Variable** |
-| ---------------- | --------------------------------- | ------------ |
-| **Inter**        | 70kb \* 2 weights = ✅ **~140kb** | ~200kb       |
-| **TASA Orbiter** | 30kb \* 3 weights = ~90kb         | ✅ **~32kb** |
+|                  | **Static**                 | **Variable**                   |
+| ---------------- | -------------------------- | ------------------------------ |
+| **Inter**        | 70kb \* 3 weights = ~210kb | ✅ **~215kb (80kb base file)** |
```
kamleshchandnani, line 225: when we say 80kb base file what does that exactly mean? does it mean because of unicode optimisation we'll only load this much or does that mean that we'll load 80kb and then load additional things as well?

saurabhdaware: It means in most scenarios we'll load 80kb but in rare scenarios if we end up using some unicode from other file, we might end up loading more.

rfcs/2023-11-08-font-loading-strategy.md
saurabhdaware: yup better

kamleshchandnani (approved):

PR #1853, opened 2023-11-29
fix: local basic build

saurabhdaware:
Earlier we were pointing to `blade/build` instead of `blade` the package so we could just pick the things from root imports like `blade/components`. But in this setup it was trying to pick React from `node_modules/@razorpay/blade/node_modules` and not the root node_modules

**New Approach**

- Use `../../blade` instead of `../../blade/build` so we don't install build directory but blade package which includes package.json exports so it respects our original exports directories and doesn't rely on paths
- We then remove the node_modules from blade directory in the `postinstall` of example so example doesn't look inside blade's node_modules for React and picks it from root of example

### DIFF
--- a/packages/examples/basic/package.json
+++ b/packages/examples/basic/package.json
@@ -2,14 +2,15 @@
   "private": "true",
   "scripts": {
     "start": "DISABLE_ESLINT_PLUGIN=true react-scripts start",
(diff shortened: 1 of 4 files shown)


snitin315 (approved):
:shipit:

PR #1871, opened 2023-12-07
fix: add stylis plugin name

snitin315:


packages/blade/src/components/BladeProvider/stylisCSSHigherSpecificity.ts
```diff
@@ -9,5 +9,9 @@ export default function createStylisCSSHigherSpecificityPlugin(): StylisPlugin {
     selectors[0] = `${selector.repeat(5)}`;
   };
 
+  Object.defineProperty(stylisCSSHigherSpecificity, 'name', {
+    value: 'stylisCSSHigherSpecificity',
+  });
```
anuraghazra, line 14: This could be a sideeffect: https://github.com/rollup/rollup/issues/1130

Let's merge this but later we should check for bundle size.

anuraghazra (approved):

PR #1892, opened 2023-12-19
fix(Rebranding / tsconfig): fix typechecks and build

saurabhdaware:


packages/blade/src/components/index.ts
```diff
@@ -1,44 +1,14 @@
...
 export * from './Badge';
 export * from './BladeProvider';
 export * from './Box';
-export * from './Card';
 export * from './Button';
 export * from './Button/IconButton';
 export * from './Counter';
-export * from './Checkbox';
-export * from './Carousel';
-export * from './Chip';
```
snitin315, line 12: comment instead?

anuraghazra (approved):

kamleshchandnani (approved):

PR #1895, opened 2023-12-20
feat: setup interaction testing

kamleshchandnani:
This PR does the following things:
1. Setup Interaction testing with Playwright
2. Incorporates carousel fixes from #1854
3. Interaction tests for Carousel
4. Run Interaction tests on CI via `Run Interaction Test` label or trigger it manually from the actions tab
5. Run interaction tests across Chromium, Firefox and Webkit based browsers


packages/blade/.storybook/react/preview.tsx
```diff
@@ -27,6 +28,19 @@ export const parameters = {
...
+    viewports: {
+      ...MINIMAL_VIEWPORTS,
+      iPhone6: {
+        name: 'iPhone 6',
+        styles: {
+          height: '667px',
+          width: '375px',
+        },
+        type: 'mobile',
+      },
```
anuraghazra, line 41: Does the MINIMAL_VIEWPORTS doesn't have any mobile viewport by default?

packages/blade/src/components/Carousel/__tests__/Carousel.test.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const previousButton = getByRole('button', { name: 'Previous Slide' });
+  await userEvent.click(nextButton);
+  await sleep(1000);
+  await expect(onChange).toBeCalledWith(1);
+  await userEvent.click(previousButton);
+  await sleep(1000);
+  await expect(onChange).toBeCalledWith(0);
+  getByText(/Single Flow To Collect And Disburse Payments/)?.scrollIntoView({ behavior: 'smooth' });
+  await sleep(1000);
+  await expect(onChange).toBeCalledWith(3);
```
anuraghazra, line 40: Actually we should also do this, after all the toBeCalledWith are done we should do another check for. 

onChange.toBeCalledTimes(3); 

This will ensure that onChange hasn't been called twice or more than it needs to be. 

This we can add in all the tests.

packages/blade/src/components/Carousel/__tests__/Carousel.test.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+};
+
+export const TestAutoPlay: StoryFn<typeof CarouselComponent> = (props): React.ReactElement => {
+  onChange = jest.fn();
+  return <BasicCarousel {...props} autoPlay visibleItems={2} onChange={onChange} />;
+};
+
+TestAutoPlay.play = async ({ canvasElement }) => {
+  const { getByRole } = within(canvasElement);
+  await sleep(8000);
```
anuraghazra, line 84: Before this we should do `onChange.not.toHaveBeenCalled()`

anuraghazra (dismissed):
LGTM

packages/blade/.storybook/react/preview.tsx
kamleshchandnani: they have weird viewports like iPhone-5 or smaller so i added standard one

PR #1905, opened 2023-12-26
refactor(blade): rebrand carousel

anuraghazra:


packages/blade/src/components/Carousel/NavigationButton/StyledNavigationButton.native.tsx
```diff
@@ -28,6 +31,7 @@ const StyledNavigationButton = ({
       onPress={onClick}
       onPressIn={() => setIsPressed(true)}
       onPressOut={() => setIsPressed(false)}
+      style={castNativeType(theme.elevation.lowRaised)}
```
kamleshchandnani, line 34: we added shadow to nav buttons now? it wasn't there before. is this intended?

packages/blade/src/components/Carousel/types.ts
```diff
@@ -1,14 +1,17 @@
...
-type BackgroundColorString<T extends ColorObjects> = `${T}.background.${DotNotationColorStringToken<
-  Theme['colors'][T]['background']
->}`;
-type BrandColorString = `brand.${DotNotationColorStringToken<Theme['colors']['brand']>}`;
-type OverlayColor = BackgroundColorString<'surface'> | BrandColorString;
+type InteractiveTokens = DotNotationToken<Theme['colors']['interactive']['background']>;
+type FeedbackTokens = DotNotationToken<Theme['colors']['feedback']['background']>;
+type SurfaceTokens = DotNotationToken<Theme['colors']['surface']['background']>;
+type OverlayTokens = DotNotationToken<Theme['colors']['overlay']>;
+type OverlayColor =
```
kamleshchandnani, line 10: why so many options for overlay color?

packages/blade/src/components/Carousel/NavigationButton/StyledNavigationButton.native.tsx
anuraghazra: I think it was missed before.

packages/blade/src/components/Carousel/types.ts
anuraghazra: Yes previously we also exposed all the feedback, surface etc colors.

packages/blade/src/components/Carousel/types.ts
```diff
@@ -1,14 +1,17 @@
 import type { Theme } from '~components/BladeProvider';
 import type { BoxProps } from '~components/Box';
 import type { Platform } from '~utils';
-import type { DotNotationColorStringToken } from '~utils/types';
+import type { DotNotationToken } from '~utils/lodashButBetter/get';
```
chaitanyadeorukhkar, line 4: Why is `DotNotationToken` in `lodashButButter/get`?

chaitanyadeorukhkar:
Why are the quotes on carousel in yellow? They were in blue before.

Before:
Screenshot 2023-12-27 at 3 38 23 PM

Now:
Screenshot 2023-12-27 at 3 38 08 PM

packages/blade/src/components/Carousel/types.ts
anuraghazra: It's a new improved type that I wrote, we will refactor out the older DotNotationColorStringToken things and replace it with this one later.

anuraghazra:
Changed the tokens for it in storybook to `theme.colors.feedback.text.information.subtle`

kamleshchandnani (approved):

PR #1907, opened 2023-12-26
refactor(blade): rebrand progressbar

anuraghazra:


packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -22,11 +22,6 @@ type ProgressBarCommonProps = {
    * Sets aria-label to help users know what the progress bar is for. Default value is the same as the `label` passed.
    */
   accessibilityLabel?: string;
-  /**
-   * Sets the contrast for the progress bar
-   * @default 'low'
-   */
-  contrast?: ColorContrastTypes;
```
kamleshchandnani, line 29: can you add this to the migration doc?

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -105,7 +100,6 @@ const progressBarHeight: Record<NonNullable<ProgressBarCommonProps['size']>, 2 |
 
 const ProgressBar = ({
   accessibilityLabel,
-  contrast = 'low',
   intent,
```
kamleshchandnani, line 103: we wanted to get rid of intent prop right?

packages/blade/src/components/ProgressBar/ProgressBar.tsx
anuraghazra: done

packages/blade/src/components/ProgressBar/ProgressBar.tsx
anuraghazra: No 🤔, intent will still be there right? We want to have colored progressbars also

kamleshchandnani (approved):

PR #1910, opened 2023-12-28
fix: rebranding errors after merge

chaitanyadeorukhkar:


packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -99,9 +99,10 @@ type GridProps = MakeObjectResponsive<
 >;
 
 type ColorObjects = 'feedback' | 'surface' | 'interactive';
-type BackgroundColorString<T extends ColorObjects> = `${T}.background.${DotNotationToken<
-  Theme['colors'][T]['background']
->}`;
+type BackgroundOnlyColorObjects = 'popup' | 'overlay';
```
chaitanyadeorukhkar, line 102: We were only exposing 'feedback' | 'surface' | 'interactive' for background colors but we also need 'popup' | 'overlay'

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -174,6 +175,8 @@ type BaseBoxVisualProps = MakeObjectResponsive<
       | BackgroundColorString<'feedback'>
       | BackgroundColorString<'surface'>
       | BackgroundColorString<'interactive'>
+      | BackgroundColorString<'overlay'>
+      | BackgroundColorString<'popup'>
       | 'transparent';
     // | (string & Record<never, never>);
```
chaitanyadeorukhkar, line 181: Removing this means we don't accept string colors on box. Is that intentional now @saurabhdaware @anuraghazra

kamleshchandnani (approved):

PR #1930, opened 2024-01-05
refactor: rebrand Radio, Checkbox & related common form components

chaitanyadeorukhkar:


packages/blade/src/components/Checkbox/checkboxTokens.ts
```diff
@@ -86,12 +86,12 @@ const checkboxIconColors: CheckboxIconColors = {
...
     background: {
-      checked: 'colors.brand.primary.600',
-      unchecked: 'colors.brand.gray.a50.lowContrast',
+      checked: 'colors.interactive.background.primary.highlighted',
+      unchecked: 'colors.interactive.background.gray.faded',
     },
     border: {
-      checked: 'colors.brand.primary.600',
-      unchecked: 'colors.brand.gray.500.lowContrast',
+      checked: 'colors.interactive.border.background.highlighted', // Intentionally not using border tokens here since we want to match the background color
```
anuraghazra, line 93: wrong token? 

border.background?

packages/blade/src/components/Form/FormHint.tsx
```diff
@@ -1,38 +1,30 @@
...
 import { CheckIcon, InfoIcon } from '~components/Icons';
-import type { BaseTextProps } from '~components/Typography/BaseText/types';
 import { getPlatformType } from '~utils/getPlatformType';
 
 type HintTextProps = {
   icon?: React.ElementType;
   children: string;
   id?: string;
-  color: BaseTextProps['color'];
+  color: TextProps<{ variant: 'caption' }>['color'];
```
anuraghazra, line 15: Let's just keep it as `BaseTextProps['color']`?

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
```diff
@@ -28,8 +29,13 @@ const getHoverStyles = ({
...
 
   return {
-    borderColor: borderColor ? getIn(theme, borderColor) : undefined,
-    backgroundColor: getIn(theme, backgroundColor),
+    borderColor: borderColor
+      ? borderColor === 'transparent'
+        ? 'transparent'
+        : getIn(theme, borderColor)
+      : undefined,
+    backgroundColor:
```
anuraghazra, line 37: We can remove this code and replace this with just `getIn(theme, borderColor)` because in our theme we now have a token called `transparent` which can be accessed via `colors.transparent`

```
    borderColor: getIn(theme, borderColor!),
    backgroundColor: getIn(theme, backgroundColor),
```

And in `packages/blade/src/components/Form/Selector/types.ts`

```diff
- type ColorTokens = `colors.${DotNotationToken<Theme['colors']>}` | 'transparent';
+ type ColorTokens = `colors.${DotNotationToken<Theme['colors']>}`;

```

packages/blade/src/utils/getFocusRingStyles/getFocusRingStyles.d.ts
```diff
@@ -0,0 +1 @@
+export * from './getFocusRingStyles.web';
```
anuraghazra, line 1: We don't require these .d.ts files anymore. 

Just

getFocusRingStyles.native
getFocusRingStyles.web
index.ts // reexport without extension

packages/blade/src/utils/getFocusRingStyles/getFocusRingStyles.web.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { Theme } from '../../components/BladeProvider';
+import { castWebType, makeMotionTime } from '~utils';
+
+function getFocusRingStyles(theme: Theme): CSSProperties {
+  return {
+    outline: `4px solid ${theme.colors.surface.border.primary.muted}`,
+    outlineOffset: '1px',
+    transitionProperty: 'outline-width',
+    transitionDuration: castWebType(makeMotionTime(theme.motion.duration['2xquick'])),
+    transitionTimingFunction: castWebType(theme.motion.easing.standard.effective),
```
anuraghazra, line 11: We have motion for this?

anuraghazra (changes requested):
Mostly lgtm

snitin315 (commented):
Also need to add these components to export.

https://github.com/razorpay/blade/blob/rebranding/master/packages/blade/src/components/index.ts

packages/blade/src/components/Checkbox/checkboxTokens.ts
chaitanyadeorukhkar: Intentional

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
chaitanyadeorukhkar: Nice! This makes things easy!

packages/blade/src/utils/getFocusRingStyles/getFocusRingStyles.web.ts
chaitanyadeorukhkar: This was there only Button motion. I will still confirm

packages/blade/src/components/Form/Selector/SelectorSupportText.tsx
```diff
@@ -1,27 +1,29 @@
...
-      as={isReactNative ? undefined : 'span'}
-      color="surface.text.muted.lowContrast"
-      fontSize={50}
-      lineHeight={50}
-      fontStyle="italic"
-      fontFamily="text"
-    >
-      {children}
-    </BaseText>
+    <Box id={id} display="contents">
```
snitin315, line 17: `display: "contents"` seems to have [partial support](https://caniuse.com/css-display-contents).  Would it cause any a11y issues, ([reference](https://hidde.blog/more-accessible-markup-with-display-contents/#heading-4))?

chaitanyadeorukhkar: > However, there is one caveat: none of the browsers that currently support display: contents expose elements that have the property to the accessibility tree with their original role.


This is problematic only when being used with an element that has an a11y role. In this case there is no a11y role for the box and it should be fine

packages/blade/src/components/Checkbox/checkboxTokens.ts
anuraghazra: I'm saying this token itself is wrong. It doesn't exist. 

image

anuraghazra (approved):
lgtm

packages/blade/src/components/Form/Selector/SelectorSupportText.tsx
snitin315: Cool, it looks like we also need to update the types to accept `contents.`

Screenshot 2024-01-09 at 4 47 06 PM

chaitanyadeorukhkar: Yup have done that too

snitin315 (approved):
LGTM. Good job ⭐

PR #1938, opened 2024-01-09
feat(Rebranding / Dropdown): Dropdown and ActionList rebranding changes

saurabhdaware:
TODO before merging this-

Dropdown relies on following components so the PR checks will fail until we merge those components in rebranding/master
- [x] Checkbox
- [x] TextInput


packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -87,38 +60,34 @@ const _ActionList = ({
   // 1. We don't render the box wrapper styles which includes shadows, padding, border etc
   // 2. to ensure GorhomBottomSheetSectionList works as expected, if we add extra wrappers GorhomBottomSheet won't render the content inside
   // NOTE: That this also means inside BottomSheet, ActionList won't render any ActionListHeader or Footer.
-  return (
-    <ActionListContext.Provider value={actionListContextValue}>
```
chaitanyadeorukhkar, line 91: Removed the provider?

packages/blade/src/components/ActionList/ActionList.tsx
```diff
@@ -159,4 +128,4 @@ const ActionList = assignWithoutSideEffects(React.memo(_ActionList), {
 });
 
 export type { ActionListProps };
-export { ActionList, useActionListContext };
+export { ActionList };
```
chaitanyadeorukhkar, line 131: What if existing consumers were using useActionListContext since its exported?

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -101,9 +100,6 @@ const _Dropdown = ({
 
   useDidUpdate(() => {
     onOpenChange?.(isOpen);
-    if (!isOpen) {
```
chaitanyadeorukhkar, line 104: Why was this removed?

packages/blade/src/components/Dropdown/docs/DropdownWithButton.stories.tsx
```diff
@@ -179,7 +179,7 @@ export const InternalAutoPositioning = (): React.ReactElement => {
         <Dropdown>
           <DropdownButton>Top Right Menu</DropdownButton>
           <DropdownOverlay>
-            <ActionList surfaceLevel={3}>
```
chaitanyadeorukhkar, line 182: So now that we've removed surfaceLevel prop, are we introducing color prop?

How will existing consumers use different surface levels?

chaitanyadeorukhkar: Also, breaking change. We would need a codemod

packages/blade/src/components/Dropdown/types.ts
```diff
@@ -14,27 +14,6 @@ type DropdownProps = {
    */
   onOpenChange?: (isOpen: boolean) => void;
   selectionType?: 'single' | 'multiple';
-  /**
-   * ## Deprecated ⚠️
```
chaitanyadeorukhkar, line 18: Breaking change. We should mention it in the release notes. Should we even do this breaking change now? Too many breaking changes happening with v11 anyway

packages/blade/src/components/ActionList/ActionList.tsx
saurabhdaware: yes. It was only used for passing `surfaceLevel` prop down. That prop is removed and dropdown has static non-configurable color now

packages/blade/src/components/ActionList/ActionList.tsx
saurabhdaware: I guess it was only exported fir tests. Consumers cant really use it since they can't add custom components in ActionList

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: `onDismiss` prop was deprecated as we later added `onOpenChange`. The deprecated prop is removed.

packages/blade/src/components/Dropdown/docs/DropdownWithButton.stories.tsx
saurabhdaware: > How will existing consumers use different surface levels?

Fun fact: In design we removed the surfaceLevel option earlier only. So all designs were always surfaceLevel={2} and this prop was not needed. 

Even if someone is using surfaceLevel={3}, its probably not following design so should be fine to change it to static color. Anyway visually also its just color change and doesn't look broken if someone was using it.

> Also, breaking change. We would need a codemod

Yup we have added it in our codemod already

packages/blade/src/components/Dropdown/types.ts
saurabhdaware: For this also we have codemod.

I don't see harm in removing it since it was deprecated for a long time and Dropdown isn't as heavily used as Text or Input so might just break at few places. We should mention it in manual migration for sure. 

I have added both surfaceLevel and onDimiss changes to blade migration doc, Nitin is adding all changes from that doc to codemod and migration guide.

chaitanyadeorukhkar (approved):

anuraghazra:
Is this expected? 
Selected state of the rejected button is also blue, ideally it should be red-ish?  

image

anuraghazra (approved):

saurabhdaware:
> Selected state of the rejected button is also blue, ideally it should be red-ish?

Yup selected state is blue for all items in current version also. Can discuss and figure out later.

PR #1952, opened 2024-01-11
refactor: rebrand `SpotlightPopoverTour`

snitin315:
- No breaking changes.


anuraghazra (approved):

packages/blade/src/components/SpotlightPopoverTour/docs/examples.ts
```diff
@@ -110,12 +110,12 @@ export const BasicExample = `
         >
           <Box width="100%" display="flex" gap="spacing.4">
             <SpotlightPopoverTourStep name="step-1">
-              <Box padding="spacing.4" backgroundColor="brand.gray.400.lowContrast">
+              <Box padding="spacing.4" backgroundColor="interactive.border.gray.faded">
```
kamleshchandnani, line 113: why are we using interactive color on a box? should be surface right?

snitin315: Hmm, this value is mapped in the codemod. I think we should update this with `surface.border.gray.muted`.

https://github.com/razorpay/blade/blob/b0e37d8d99ec5befc21ce68a3558bc80b9a6523c/packages/blade/codemods/brand-refresh/transformers/colorTokensMapping.json#L45

`interactive.border.gray.faded` & `surface.border.gray.muted` have the same value.

https://github.com/razorpay/blade/blob/b0e37d8d99ec5befc21ce68a3558bc80b9a6523c/packages/blade/src/tokens/theme/bladeTheme.ts#L39

https://github.com/razorpay/blade/blob/b0e37d8d99ec5befc21ce68a3558bc80b9a6523c/packages/blade/src/tokens/theme/bladeTheme.ts#L300

anuraghazra: @snitin315 put this on a TODO somewhere, we will take this up as a separate PR. 
Let's merge this component. This is just an example file. 

// cc @kamleshchandnani

anuraghazra: Are we sure about this token? We are setting border color is `backgroundColor`. 

> And I'm not sure we should also map this to border on the colorMapping.json file - since we don't know if it's being used as border/background or surface.

snitin315: This was previously discussed in https://github.com/razorpay/blade/pull/1883#discussion_r1427744961

So currently, we migrate every token to surface tokens by default via the `colorMapping.json `,  then we try to update these tokens contextually based on usage wherever possible.

https://github.com/razorpay/blade/blob/40dd3046990b3ef32926d0d0cbf234e7bd13980b/packages/blade/codemods/brand-refresh/transformers/index.ts#L113

https://github.com/razorpay/blade/blob/40dd3046990b3ef32926d0d0cbf234e7bd13980b/packages/blade/codemods/brand-refresh/transformers/__tests__/migrate-colors.test.ts#L38


However, the issue with `brand.gray.*` tokens is that not all the tokens have their equivalent in surface tokens. hence we have them mapped to `border.*`.  So we need[ designers inputs here](https://razorpay.slack.com/archives/G01B3LQ9H0W/p1705317155508699?thread_ts=1705314490.473439&cid=G01B3LQ9H0W).

PR #1953, opened 2024-01-11
chore: remove razorsharp from workspace packages

snitin315:
It shouldn't be there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,6 @@
   "workspaces": {
     "packages": [
       "packages/blade",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -9,7 +9,6 @@
   "workspaces": {
     "packages": [
       "packages/blade",
-      "packages/razorsharp",
```
kamleshchandnani, line 12: once we remove this all the eslint and everything needs to be setup inside razorsharp then isn't it?

PR #1954, opened 2024-01-11
feat(tokens): add new tokens

rzpcibot:


packages/blade/src/tokens/theme/bladeTheme.ts
```diff
@@ -13,2652 +13,900 @@ const colors: ColorsWithModes = {
...
-      icon: {
-        blue: {
-          lowContrast: globalColors.chromatic.azure[400],
-          highContrast: globalColors.neutral.blueGrayDark[0],
-        },
+        subtle: globalColors.neutral.blueGrayDark.a100,
+        intense: globalColors.neutral.blueGrayDark.a100,
       },
     },
+    elevation: {
```
saurabhdaware, line 904: Is this expected? These also look different from what we migrated to recently 🤔

PR #1959, opened 2024-01-11
feat(blade): rebranding whitelabel fix

anuraghazra:


packages/blade/src/components/Badge/Badge.tsx
```diff
@@ -13,7 +13,7 @@ import type { StyledPropsBlade } from '~components/Box/styledProps';
 import type { StringChildrenType, TestID } from '~utils/types';
 import { getStringFromReactText } from '~src/utils/getStringChildren';
 import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
-import { isReactNative, makeSize } from '~utils';
+import { isReactNative, makeSize, useTheme } from '~utils';
```
chaitanyadeorukhkar, line 16: why

packages/blade/src/storybook-recipes/ThemePlayground/ThemeSelector.tsx
```diff
@@ -68,7 +72,7 @@ const ColorSelector = ({
 }): React.ReactElement => {
   return (
     <ColorSelection color={color} onClick={onClick} isSelected={isSelected}>
-      {isSelected ? <CheckIcon size="large" color="action.icon.primary.default" /> : null}
+      {isSelected ? <CheckIcon size="large" color="interactive.icon.staticWhite.normal" /> : null}
```
chaitanyadeorukhkar, line 75: use onPrimary here ?

anuraghazra: done

packages/blade/src/components/Badge/Badge.tsx
anuraghazra: removed

chaitanyadeorukhkar:
Lets use the primary color on carousel's quotes like we do on master? Right now it always remains blue
Screenshot 2024-01-12 at 12 58 03 PM

chaitanyadeorukhkar:
Focus ring color is not changing for multiple components like Checkbox, Radio, Accordion, Switch, Chips... can you check this once? I created a common `getFocusRingStyles` function that uses `surface.border.primary.muted` which should ideally change as per the white labeled theme
Screenshot 2024-01-12 at 1 00 25 PM

chaitanyadeorukhkar:
Carousel's button's arrow icons should also follow the primary color, right?
Screenshot 2024-01-12 at 1 03 22 PM

anuraghazra:
> Focus ring color

Now that you mention it, I wonder if we should keep the focus ring color static and not tie it with brand color. 

The focus ring is an fundamental accessibility requirement, if brand color becomes very light then focus ring color won't be visible to the user and could cause accessibility issues. 

https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html

anuraghazra:
> Carousel's button's arrow icons should also follow the primary color, right?

Nope, rebranded carousel nav buttons icons are staticBlack.

anuraghazra:
> Lets use the primary color on carousel's quotes like we do on master? Right now it always remains blue

Done.

chaitanyadeorukhkar:
> > Focus ring color
> 
> Now that you mention it, I wonder if we should keep the focus ring color static and not tie it with brand color.
> 
> The focus ring is an fundamental accessibility requirement, if brand color becomes very light then focus ring color won't be visible to the user and could cause accessibility issues.
> 
> https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html

Hmm. That is fair. I think that we can think about on its own but right now we're using primary color token for the focus ring ideally, it should use the primary color when it changes as well?

chaitanyadeorukhkar:
> > Carousel's button's arrow icons should also follow the primary color, right?
> 
> Nope, rebranded carousel nav buttons icons are staticBlack.

Why are they blue in this screenshot then? 🤔 
Screenshot 2024-01-12 at 1 03 22 PM

chaitanyadeorukhkar:
@anuraghazra Tests are failing. Mostly looks good to me. Just need to check

1. Focus ring color
2. Carousel's blue button

anuraghazra:
> Why are they blue in this screenshot then? 🤔

Wierd, are you checking the correct chromati link? Because it's not possible for it to be blue. 

image

chaitanyadeorukhkar:
> > Why are they blue in this screenshot then? 🤔
> 
> Wierd, are you checking the correct chromati link? Because it's not possible for it to be blue.
> 

Might be the wrong link. Checked again. Looks good

chaitanyadeorukhkar (approved):

anuraghazra:
> Focus ring color

I've standardized all components to use the `getFocusRingColor` function.

PR #1960, opened 2024-01-11
refactor: rebrand input components

chaitanyadeorukhkar:


packages/blade/.storybook/react-native/main.tsx
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   stories: [
-    '../../src/**/!(_KitchenSink)*.stories.?(ts|tsx|js|jsx)',
+    '../../src/**/!(_KitchenSink|*.test)*.stories.?(ts|tsx|js|jsx)',
```
chaitanyadeorukhkar, line 3: Excludes all e2e tests

packages/blade/src/components/Tabs/TabItem.web.tsx
```diff
@@ -128,6 +128,8 @@ const TabItem = ({
     default: 'default',
     hover: 'highlighted',
     focus: 'highlighted',
+    active: 'highlighted',
```
anuraghazra, line 131: Why did we added this active state?  

I changed the `useInteraction` to remove the `active` state since `focus` can just be used as `active` state.

anuraghazra: Also, 

In this file `packages/blade/src/components/Input/BaseInput/AnimatedBaseInputWrapper.web.tsx`

These if checks will never occur, because while we changed the types the useInteraction hook will never return `active` as a state. It only returns `focus`

image

chaitanyadeorukhkar: Makes sense. I think Input had an active state but we can change it with focus and remove active entirely

chaitanyadeorukhkar: Done. I've pushed the changes

anuraghazra:
Is this expected?

When we focus into the Input, There is a border bottom (not the animated border) which suddenly becomes blue, and then on top of that the animated border comes in. 

Previously this was not the case, that static border would stay gray and on top of that gray border the animated blue border would come. 

Rebranded: 


https://github.com/razorpay/blade/assets/35374649/fbdc1894-61c3-44ef-abf0-7b04b2ee321f



Master: 

https://github.com/razorpay/blade/assets/35374649/00ba7e4e-20c5-4702-bb25-d266c8af6034

packages/blade/src/components/Input/BaseInput/baseInputStyles.ts
```diff
@@ -46,32 +46,33 @@ export const getInputBackgroundAndBorderStyles = ({
...
   if (isHovered) {
-    backgroundColor = theme.colors.brand.gray.a100.lowContrast;
+    backgroundColor = theme.colors.interactive.background.gray.highlighted;
   }
 
   // focused state
   if (isFocused) {
-    backgroundColor = theme.colors.brand.primary[300];
+    backgroundColor = theme.colors.interactive.background.primary.faded;
+    borderBottomColor = theme.colors.interactive.background.primary.default; // TODO: Check why this is not a border color
```
chaitanyadeorukhkar, line 60: Checked with design, they're using a background color here since its an animated block and not really a border

packages/blade/src/utils/useInteraction.ts
```diff
@@ -1,7 +1,7 @@
 import type { Dispatch, SetStateAction } from 'react';
 import { useState } from 'react';
 
-export type ActionStates = 'default' | 'hover' | 'focus';
+export type ActionStates = 'default' | 'hover' | 'focus' | 'active' | 'disabled';
```
chaitanyadeorukhkar: I'm not sure why the diff is showing `active` here. I've removed it. Even when I try to edit this file on github, it doesn't show active. Seems like a bug on github
Screenshot 2024-01-12 at 12 41 02 PM

anuraghazra:
There's a double border which is happening 1 gray 1 blue. 

Figma: 
image


Dev: 
image

Interestingly this is happening on master as well.

packages/blade/src/components/Input/BaseInput/StyledBaseInput.web.tsx
```diff
@@ -34,9 +34,8 @@ const getWebInputStyles = (
       ...getTextStyles({
         size: 'medium',
         variant: 'body',
-        type: 'placeholder',
         weight: 'regular',
-        contrast: 'low',
```
anuraghazra, line 39: TODO: Document any prop removals/replacements in the migration doc for codemod development.

anuraghazra (approved):
1 comment, otherwise looks good to me.

saurabhdaware (approved):
Looks good to me

PR #1962, opened 2024-01-12
feat(Dropdown): add E2E tests

saurabhdaware:
- Add basic E2E tests for Dropdown
- Remove unit tests which were e2e tests in disguise and move them to real e2e
- Fixes a bug that was found while writing these e2e tests 🙈 . The dropdown was getting closed even when the state change on consumer end was not called inside `onOpenChange`
- Fixes onClick getting called twice on Dropdown triggers


packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -99,24 +98,22 @@ const _Dropdown = ({
 
   const dropdownBaseId = useId('dropdown');
 
-  useDidUpdate(() => {
-    onOpenChange?.(isOpen);
-    if (!isOpen) {
```
saurabhdaware, line 104: With this logic, we were calling `onOpenChange` as well as we were actually closing and opening dropdown.

Ideally we should only call `onOpenChange` wherever state needs to change and then consumers add 

```jsx
<Dropdown isOpen={isDropdownOpen} onOpenChange={(isOpen) => { setIsDropdownOpen(isOpen) }} />
```

if someone doesn't call state change in onOpenChange, then it shouldn't close the dropdown.

packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -99,24 +98,22 @@ const _Dropdown = ({
...
+  const setIsOpenControlled = (isControlledStateOpen: boolean): void => {
+    onOpenChange?.(isControlledStateOpen);
+    if (!isControlledStateOpen) {
```
anuraghazra, line 103: Can we use the useControllableState hook to manage the controlled/uncontrolled behaviour of `isOpen`?

anuraghazra (approved):
LGTM. 

- Check the tests on Safari/Firefox

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: ouu didn't know about this hook. let me see.

saurabhdaware: done

saurabhdaware:
> LGTM.
>
> Check the tests on Safari/Firefox

done

anuraghazra (approved):
lint failing. else lgtm

chaitanyadeorukhkar (approved):

PR #1963, opened 2024-01-12
chore: add server side pagination with `totalItemCount` & `paginationType`  to Blade `v10`

chaitanyadeorukhkar:
- fix: add totalItemCount to TablePagination
- update type


chaitanyadeorukhkar:
This is released in `v10.24.0`

Issue #1934, opened 2024-01-08
Card : requires `minHeight` prop


iamakshatjain:
Card component so far accepts only a `height` prop, when using height with `100%`, the size of the card changes with or without the data. This might leave some empty real estate around the web page. 
To solve for similar problems, this requires a `minHeight` prop within the Card component to prevent such UX breakage
slack : https://razorpay.slack.com/archives/CMQ3RBHEU/p1704286719871369

AritraLeo:
I'd like to work on this issue but I am so sorry that I can't view your slack page. Please give me a bit more insight about the issue, like the name of the file and so on!

iamakshatjain:
> I'd like to work on this issue but I am so sorry that I can't view your slack page. Please give me a bit more insight about the issue, like the name of the file and so on!

@AritraLeo nothing much on slack, you can checkout this file https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Card/Card.tsx

minHeight and minWidth are not available as props here

AritraLeo:
> > I'd like to work on this issue but I am so sorry that I can't view your slack page. Please give me a bit more insight about the issue, like the name of the file and so on!
> 
> @AritraLeo nothing much on slack, you can checkout this file https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Card/Card.tsx
> 
> minHeight and minWidth are not available as props here

cool thanks for the file path !

AritraLeo:
> > I'd like to work on this issue but I am so sorry that I can't view your slack page. Please give me a bit more insight about the issue, like the name of the file and so on!
> 
> @AritraLeo nothing much on slack, you can checkout this file https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Card/Card.tsx
> 
> minHeight and minWidth are not available as props here

I've gone through the files related to the issue seems like I just have to add minHeight and minWidth props to CardProps 
Like minHeight?: BoxProps['param'] now BoxProps extends LayoutProps that has minHeight and minWidth keys. 

I'll proceed with this after confirmation and please give me the PR message last one was editted because the message was not apt. 

Thanks!

iamakshatjain:
> > > I'd like to work on this issue but I am so sorry that I can't view your slack page. Please give me a bit more insight about the issue, like the name of the file and so on!
> > 
> > 
[...]

PR #1964, opened 2024-01-12
feat(blade): add minHeight prop to CardProps

AritraLeo:
I've added the minHeight prop to CardProps and also to Card component in Card.tsx file. 
path - blade\packages\blade\src\components\Card\Card.tsx
issue - https://github.com/razorpay/blade/issues/1934

fixes https://github.com/razorpay/blade/issues/1934


packages/blade/src/components/Card/Card.tsx
```diff
@@ -83,12 +83,17 @@ export type CardProps = {
    * Sets the height of the card
    */
   height?: BoxProps['height'];
+    /**
+    Stops UI breakage when height is 100% 
+   */
+  minHeight?: BoxProps['minHeight'];
```
iamakshatjain, line 89: 1. Formatting
2. Can update text to maybe just "sets minimum height of the card"
3. Can add min width also

packages/blade/src/components/Card/Card.tsx
iamakshatjain, line 1: You might need to pass these to CardRoot also, try running this on local also

packages/blade/src/components/Card/Card.tsx
AritraLeo: Thanks !
I'll make the changes asap .

packages/blade/src/components/Card/Card.tsx
AritraLeo: Please have a look sir. If there is anything remaining please me know. 
Thanks !

iamakshatjain (approved):

iamakshatjain:
@chaitanyadeorukhkar @kamleshchandnani @saurabhdaware please take a look at the PR

packages/blade/src/components/Card/Card.tsx
```diff
@@ -83,12 +83,21 @@ export type CardProps = {
    * Sets the height of the card
    */
   height?: BoxProps['height'];
+    /**
+    Sets minimum height of the card
+   */
+  minHeight?: BoxProps['minHeight'];
+      /**
+    Sets minimum width of the card
+   */
```
anuraghazra, line 92: Please format the code same as how `width` is formatted. 

```suggestion
  /**
   * Sets minimum width of the card
   */
```

anuraghazra:
@AritraLeo thanks for the PR. 

Can you add a story example for this in [here](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Card/Card.stories.tsx)? 

You can add a new "SameHeightCard" story (refer to the examples)

packages/blade/src/components/Card/Card.tsx
AritraLeo: Ok I'll do that and apologies for not doing the formatting properly this time !

anuraghazra:
> yeah sure I can have a look but the task is actually not very clear to me. Can you please brief me a little bit more ?

You can look at the existing example [here](https://github.com/razorpay/blade/blob/e0caa7e223e13918ba02cfb6520c33d1fa98597c/packages/blade/src/components/Card/Card.stories.tsx#L371-L407). 

In this file you can add another example called `CardWithSameHeight`.

```jsx
const CardWithSameHeightExample: StoryFn<typeof Card> = (): React.ReactElement => {
  // Add 3 Cards with same heights by setting minHeight property to each of them
  return null;
};

export const CardWithSameHeight = CardWithSameHeightExample.bind({});
CardWithSameHeight.parameters = {
  controls: {
    disable: true,
  },
};
```

anuraghazra:
FYI, tried this out locally, this doesn't work correctly. 

You also need to add minHeight prop to `CardSurface` because that's whats adds the background color/elevation etc.

anuraghazra:
Merging this for now, we are releasing v11 soon and we want this change to be released in v10.

anuraghazra (approved):

snitin315 (approved):
LGTM, thanks for contributing.

@kamleshchandnani @chaitanyadeorukhkar need your approval for merging.

AritraLeo:
> FYI, tried this out locally, this doesn't work correctly.
> 
> You also need to add minHeight prop to `CardSurface` because that's whats adds the background color/elevation etc.

Apologies for the inconvenience I assure you it will not happen again I will be cautious.

kamleshchandnani (approved):
Thank you for your contribution 🎉

anuraghazra:
> Apologies for the inconvenience I assure you it will not happen again I will be cautious.

No worries.

AritraLeo:
> > Apologies for the inconvenience I assure you it will not happen again I will be cautious.
> 
> No worries.

I'll try to create a PR for - "SameHeightCard" story

PR #1965, opened 2024-01-12
chore: add `onHover` & `onClick` callback to `TableRow`

chaitanyadeorukhkar:
Makes rows hoverable without them needing to be selectable. Consumers can decide what they want to do on hover.

Ref: [Issue raised on slack](https://razorpay.slack.com/archives/CMQ3RBHEU/p1704712316037749)
Resolves: #1966

### DIFF
--- a/.changeset/quick-students-beam.md
+++ b/.changeset/quick-students-beam.md
@@ -2,4 +2,4 @@
 "@razorpay/blade": patch
 ---
 
-chore: add `onHover` & `onClick` callback to `TableRow`
+feat: add `onHover` & `onClick` callback to `TableRow`
--- a/.changeset/quick-students-beam.md
+++ b/.changeset/quick-students-beam.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+chore: add `onHover` callback to `TableRow`
--- a/.changeset/quick-students-beam.md
+++ b/.changeset/quick-students-beam.md
@@ -2,4 +2,4 @@
 "@razorpay/blade": patch
 ---
 
-chore: add `onHover` callback to `TableRow`
+chore: add `onHover` & `onClick` callback to `TableRow`
--- a/packages/blade/src/components/Table/__tests__/Table.web.test.tsx
+++ b/packages/blade/src/components/Table/__tests__/Table.web.test.tsx
@@ -617,6 +617,45 @@ describe('<Table />', () => {
     expect(getAllByRole('row')[0]).toHaveTextContent('completed');
   });
 
+  it('should call onHover when row is hovered', async () => {
+    const onHover = jest.fn();
+    const user = userEvent.setup();
+    const { getByText } = renderWithTheme(
+      <Table data={{ nodes: nodes.slice(0, 5) }} selectionType="single">
+        {(tableData) => (
+          <>
+            <TableHeader>
+              <TableHeaderRow>
+                <TableHeaderCell>Payment ID</TableHeaderCell>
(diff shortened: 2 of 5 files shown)


packages/blade/src/components/Table/__tests__/Table.web.test.tsx
```diff
@@ -617,6 +617,45 @@ describe('<Table />', () => {
...
+                <TableHeaderCell>Amount</TableHeaderCell>
+                <TableHeaderCell>Status</TableHeaderCell>
+                <TableHeaderCell>Type</TableHeaderCell>
+                <TableHeaderCell>Method</TableHeaderCell>
+                <TableHeaderCell>Name</TableHeaderCell>
+              </TableHeaderRow>
+            </TableHeader>
+            <TableBody>
+              {tableData.map((tableItem, index) => (
+                <TableRow item={tableItem} key={index} onHover={onHover}>
```
saurabhdaware, line 639: should it be `onMouseOver` to make it consistent with web's event handler? that is what we have on Box as well I guess 🤔

chaitanyadeorukhkar: Also we should use `onMouseEnter` & not `onMouseOver`. onMouseOver will keep retriggering for the same row over and over again whenever the mouse enters any child element. We want it to only trigger once for the whole row

anuraghazra (approved):

packages/blade/src/components/Table/__tests__/Table.web.test.tsx
anuraghazra: > why did we not use onMouseOver for card

No particular, reason. There was no discussion on event names in Interactive Card API [Decisions](https://github.com/razorpay/blade/blob/master/packages/blade/src/components/Card/_decisions/interactive-cards.md). 

For Card maybe onHover makes sense since users should not care about the internal implementation on if it's "onMouseEnter" or "onMouseOver". 

Tho if we want to standardise if further to keep it same with web, I don't see an issue with onMouseEnter also.

anuraghazra (approved):
Mostly looks good, just change all the onHover to onMouseEnter

.changeset/quick-students-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+chore: add `onHover` callback to `TableRow`
```
anuraghazra, line 5: onMouseEnter

snitin315:
```suggestion
chore: add `onMouseEnter` callback to `TableRow`
```

packages/blade/src/components/Table/__tests__/Table.web.test.tsx
chaitanyadeorukhkar: Calling it onHover since onMouseEnter is a web event and expects callbacks that are true to `React.MouseEventHandler<HTMLElement>` but we only want to pass `{ item }` as a callback

.changeset/quick-students-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+chore: add `onHover` & `onClick` callback to `TableRow`
```
snitin315, line 5: It should be marked as `feat` as this change impacts consumers.

```suggestion
feat: add `onHover` & `onClick` callback to `TableRow`
```

snitin315 (dismissed):
LGTM

PR #1968, opened 2024-01-15
fix: remove page change check that would cause infinite renders

chaitanyadeorukhkar:


anuraghazra (approved):

packages/blade/src/components/Table/TablePagination.web.tsx
```diff
@@ -212,19 +211,6 @@ const _TablePagination = ({
...
   }, [controlledCurrentPage, currentPage, handlePageChange, onPageChange]);
 
-  if (currentPage > totalPages - 1) {
-    if (!isUndefined(controlledCurrentPage)) {
-      if (__DEV__) {
-        throwBladeError({
-          moduleName: 'TablePagination',
-          message: `Value of 'currentPage' prop cannot be greater than the total pages`,
-        });
-      }
```
snitin315, line 222: Need to remove this test case too

https://github.com/razorpay/blade/blob/021f11fba14baa0819672c3a4ede0bd0b3510af8/packages/blade/src/components/Table/__tests__/Table.web.test.tsx#L810

PR #1973, opened 2024-01-16
refactor(blade): rebrand Table

anuraghazra:


chaitanyadeorukhkar:
@anuraghazra rename the Table with Surface Levels story to Table with Background Colors

packages/blade/src/components/Table/Table.web.tsx
```diff
@@ -121,9 +122,9 @@ const _Table = <Item,>({
   height,
   showStripedRows,
   gridTemplateColumns,
-  surfaceLevel = 2,
   isLoading = false,
   isRefreshing = false,
+  backgroundColor = 'surface.background.gray.intense',
```
chaitanyadeorukhkar, line 127: Should we also add a JS check to ensure no other colors are passed here?

packages/blade/src/components/Table/TableBody.web.tsx
```diff
@@ -263,10 +266,7 @@ const StyledRow = styled(Row)<{ $isSelectable: boolean }>(({ theme, $isSelectabl
           cursor: 'pointer',
         },
       }),
-      '&:focus': {
-        outline: 'none',
-        boxShadow: `0 0 0 2px ${getIn(theme.colors, 'brand.primary.300')}`,
-      },
+      '&:focus': getFocusRingStyles({ theme, negativeOffset: true }),
```
chaitanyadeorukhkar, line 269: What is negativeOffset for?

packages/blade/src/components/Table/TableToolbar.web.tsx
```diff
@@ -96,7 +96,11 @@ const _TableToolbar = ({
       >
         <BaseBox display="flex" alignItems="center" flex={1}>
           <BaseBox>
-            <Text size="medium" weight="semibold">
+            <Text
+              size="medium"
+              weight="medium"
+              color={isSelected ? 'surface.text.gray.normal' : 'surface.text.gray.subtle'}
```
chaitanyadeorukhkar, line 102: Earlier it was always the same color. Why are we changing the color now depending on whether its selected or not?

packages/blade/src/components/Table/__tests__/Table.web.test.tsx
```diff
@@ -387,9 +387,9 @@ describe('<Table />', () => {
     expect(container).toMatchSnapshot();
   });
 
-  it('should render table with surfaceLevel', () => {
+  it('should render table with backgroundColor', () => {
     const { container } = renderWithTheme(
-      <Table data={{ nodes: nodes.slice(0, 2) }} surfaceLevel={3}>
+      <Table data={{ nodes: nodes.slice(0, 2) }} backgroundColor="surface.background.sea.intense">
```
chaitanyadeorukhkar, line 392: We shouldn't allow non gray background

packages/blade/src/components/Table/docs/BasicTable.stories.tsx
```diff
@@ -151,6 +184,18 @@ const TableTemplate: StoryFn<typeof TableComponent> = ({ ...args }) => {
                 </TableRow>
               ))}
             </TableBody>
+            <TableFooter>
+              <TableFooterRow>
+                <TableFooterCell>Total</TableFooterCell>
+                <TableFooterCell>-</TableFooterCell>
+                <TableFooterCell>-</TableFooterCell>
+                <TableFooterCell>-</TableFooterCell>
```
chaitanyadeorukhkar, line 192: You might have to add another TableFooterCell here ifselectionType is multiple. Otherwise there will be a blank cell

packages/blade/src/utils/getFocusRingStyles/getFocusRingStyles.web.test.ts
```diff
@@ -3,7 +3,9 @@ import { getFocusRingStyles } from './getFocusRingStyles';
 
 describe('getFocusRingStyles', () => {
   it('should return correct styles when called with valid parameters', () => {
-    const result = getFocusRingStyles({ ...bladeTheme, colors: bladeTheme.colors.onLight });
+    const result = getFocusRingStyles({
```
chaitanyadeorukhkar, line 6: add test for negativeOffset

packages/blade/src/components/Table/Table.web.tsx
anuraghazra: I don't think that would be necessary. 

The UI itself will look broken if they @ts-ignore and set any different color.

packages/blade/src/utils/getFocusRingStyles/getFocusRingStyles.web.ts
```diff
@@ -1,15 +1,25 @@
...
 
-function getFocusRingStyles(theme: Theme): CSSProperties {
+type GetFocusRingProps = {
+  theme: Theme;
+  negativeOffset?: boolean;
+};
+
+/**
+ * @param props.theme Blade Theme Object
+ * @param props.negativeOffset if set the outline offset will be set to -4px, this is useful
```
chaitanyadeorukhkar, line 12: How were we doing this earlier? I think I was using some CSS workaround only for this. Was it offset only?

packages/blade/src/utils/getFocusRingStyles/getFocusRingStyles.web.ts
```diff
@@ -1,15 +1,25 @@
...
+ * in table component where the outline will get cutoff by the table border
```
chaitanyadeorukhkar, line 13: If this is specific to table, lets add outlineOffset to table only?

packages/blade/src/utils/lodashButBetter/get.ts
```diff
@@ -8,7 +8,7 @@ import type { EasingType } from '~tokens/global/motion';
  * @description Tokenises objects to dot notation strings, eg: `surface.text.normal.lowContrast`
  */
 export type DotNotationToken<_TokenType, TokenType = Omit<_TokenType, 'name'>> = {
-  [K in keyof TokenType]: TokenType[K] extends string | ElevationStyles
+  [K in keyof TokenType]: TokenType[K] extends string | number | ElevationStyles
```
chaitanyadeorukhkar, line 11: Where are we accepting number in dot notation?

packages/blade/src/components/Table/TableBody.web.tsx
anuraghazra: Added comments on the getFocusRingStyles implementation. 

With negative offset: 
image


Without negative offset:
image

packages/blade/src/components/Table/TableToolbar.web.tsx
anuraghazra: Matching figma behaviour, it seems we missed adding this isSelected color change before.

packages/blade/src/components/Table/__tests__/Table.web.test.tsx
anuraghazra: Fixed. TS will throw for consumers.

packages/blade/src/components/Table/docs/BasicTable.stories.tsx
anuraghazra: Added

packages/blade/src/utils/getFocusRingStyles/getFocusRingStyles.web.test.ts
anuraghazra: These tests are pretty useless to be honest. 
We are just snapshot testing them. 

If we update getFocusRingStyles anyways every single interactive element's focus styles will update which are already has snapshots.

packages/blade/src/utils/getFocusRingStyles/getFocusRingStyles.web.ts
anuraghazra: We used `inset` box shadows.

packages/blade/src/utils/getFocusRingStyles/getFocusRingStyles.web.ts
anuraghazra: This can be used in other places also if we face similar focus ring cuttoff issues. 

I would rather keep focus ring styles in a single place than spread it all over the codebase, it would become same as before. 

Also the negativeOffset has to match the outline-width, if we put negativeOffsets in Table styles we will have random `-4px` negative offsets all over the place. 

```
&focus-visible: {
 ...getFocusRingStyles(),
 outlineOffset: `-4px` // Developer has no idea why is this -4px, unless they look inside getFocusRingStyles.
}
```

anuraghazra:
> rename the Table with Surface Levels story to Table with Background Colors

Fixed.

chaitanyadeorukhkar (approved):

PR #1975, opened 2024-01-17
ci: add bundle size report

snitin315:
### PR

- Read package exports & run size-limit on each component export.
- Create json output file named `PRBundleSizeStats.json`
- Compare `PRBundleSizeStats.json` with `baseBundleSizeStats.json` from the master and report the difference on PR via danger-js.

### master

- Read package exports & run size-limit on each component export.
- Create json output file named `baseBundleSizeStats.json`.
- Commit & push the `baseBundleSizeStats.json` file.

### Addition info

- Report - https://github.com/razorpay/blade/pull/1975#issuecomment-1925215322
   - currently base stats is 0 for all components since we don't have data on master.
- Base size 0 means the component is newly added.
- Current Size 0 means the component has been removed.
- [JIRA Story](https://razorpay.atlassian.net/browse/FC-2587)

Screenshot 2024-02-04 at 12 46 52 PM


packages/blade/scripts/generateBundleSizeInfo.js
```diff
@@ -0,0 +1,95 @@
...
+    'getTextProps',
+    'screenReaderStyles',
+    'useActionListContext',
+    'ActionListItemAsset', // This is just an <img> tag
+    // These are same as Badge, Amount, Counter, Link, Text components
+    'CardHeaderAmount',
+    'CardHeaderBadge',
+    'CardHeaderCounter',
+    'CardHeaderLink',
+    'CardHeaderText',
```
saurabhdaware, line 36: Any harm in keeping these? They are wrappers but their size might increase decrease depending on the code we add later no.

snitin315: It just increases the script run time, these components can add approx ~2mins. I believe these wrappers would hardly get updates. They have dev-only checks to verify amount is being used inside card only.
https://github.com/razorpay/blade/blob/8d89eb5365f75880ac3c07ce054d8633c3f05422/packages/blade/src/components/Card/CardHeader.tsx#L57-L61

kamleshchandnani:
instead of ✅  and 🚫 can we use 🟢 and 🔴 to be more clear?

saurabhdaware:
The check seems to take 15 - 17minutes 😢 Is there any alternate way to implement or any way to optimize this?

snitin315:
> The check seems to take 15 - 17 minutes 😢 Is there any alternate way to implement or any way to optimize this?

Yeah, this is the time taken by the size limit package. But this is expected because we have to initialize and bundle a webpack project for each component. Let me try to use m1 MacOS on CI. Otherwise, we can move to our custom implementation in the future but that would be a significant effort.

packages/blade/scripts/generateBundleSizeInfo.js
```diff
@@ -0,0 +1,95 @@
...
+  // Parse the file content to generate an AST
+  const ast = babelParser.parse(fileContent, {
+    sourceType: 'module',
+  });
+
+  // Arrays to store size-limit stats and component names to exclude
+  const sizes = [];
+  const excludedComponents = [
+    'useTheme',
+    'BladeProvider',
```
anuraghazra, line 23: BladeProvider should not be excluded, it's an important entry point for blade and if we accidentally import some third party lib here or mess up treeshaking on this component could affect all other components.

snitin315: I'll include it.

saurabhdaware (approved):

packages/blade/scripts/generateBundleDiff.js
```diff
@@ -0,0 +1,78 @@
...
+  if (bundleDiff.length === 0) {
+    return { diffTable: null };
+  }
+
+  // Calculate the size differences and create a formatted diff table
+  bundleDiff.forEach((component) => {
+    const currentComponent = currentBundleSizeStats.find((stat) => stat.name === component.name);
+    const baseComponent = baseBundleSizeStats.find((stat) => stat.name === component.name);
+
+    if (baseComponent && !currentComponent) {
```
anuraghazra, line 40: Maybe put this in variable: 

const isComponentRemoved = baseComponent && ..

anuraghazra (approved):

packages/blade/scripts/generateBundleDiff.js
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  // Generate a Markdown table for the bundle size differences
+  const diffTable = `
+  | Status | Component | Base Size (kb) | Current Size (kb) | Diff |
+  | --- | --- | --- | --- | --- |
+  ${bundleDiff
+    .map(
+      ({ name, baseSize, prSize, diffSize, isSizeIncreased }) =>
+        `| ${isSizeIncreased ? '🔴' : '🟢'} | ${name} | ${baseSize} | ${prSize} | ${
```
chaitanyadeorukhkar, line 68: Not sure if we should keep red dot for size increase. Size will always keep increasing and we will see a red dot on almost every PR. Lets use Arrows instead? ⬆ ⬇

packages/blade/scripts/generateBundleSizeInfo.js
```diff
@@ -0,0 +1,115 @@
...
+      sizes.push(
+        JSON.parse(
+          jsonLikeString.substring(jsonLikeString.indexOf('[') + 1, jsonLikeString.indexOf(']')),
+        ),
+      );
+    }
+  });
+
+  // Write the gathered size information to the specified file
+  const filename = process.env.BUNDLE_SIZE_STATS_FILENAME || 'PRBundleSizeStats.json';
```
chaitanyadeorukhkar, line 111: Lets run prettier on this file once generated?

packages/blade/scripts/generateBundleSizeInfo.js
```diff
@@ -0,0 +1,115 @@
...
+              gzip: true,
+            },
+          ],
+          null,
+          2,
+        ),
+      );
+
+      // Run size-limit command and capture the output to gather size information
+      const { stdout } = execa.commandSync('yarn size-limit --json');
```
chaitanyadeorukhkar, line 93: The sizes for each component that we recieve here is not the size of the component but the size of the project when using only this component. This would be a misleading stat to post as a comment on the PR. Lets do either
1. Measure the size of the component only (this would mean using something other than size-limit)
2. Measure the size of the component & the size of an empty size-limit project and use the diff as the component size
3. Mention a note in the comment saying that the sizes mentioned in the table are not actual component sizes but rather size of the project when using only this component

chaitanyadeorukhkar (changes requested):

packages/blade/scripts/generateBundleSizeInfo.js
```diff
@@ -0,0 +1,115 @@
...
+      sourceType: 'module',
+    });
+    // Arrays to store size-limit stats and component names to exclude
+    const exportedComponents = [];
+
+    // Traverse the AST to get the component names
+    traverse(ast, {
+      // Get the component name from the export statement
+      ExportSpecifier: ({ node }) => {
+        const componentName = node.exported.name;
```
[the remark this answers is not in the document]
snitin315: Our codebase already follows this convention, we have to provide named exports for our components because we allow all imports from `@razorpay/blade/components`.

packages/blade/scripts/generateBundleSizeInfo.js
snitin315: @chaitanyadeorukhkar I've updated the PR with 2nd approach.

packages/blade/scripts/generateBundleSizeInfo.js
snitin315: I feel prettier is not required, we need to store json just for diff. Someone will hardly read it.

packages/blade/package.json
```diff
@@ -224,11 +227,13 @@
     "chalk": "4.1.1",
     "copyfiles": "2.4.1",
     "cross-env": "7.0.3",
+    "danger": "^11.3.1",
```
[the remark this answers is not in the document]
snitin315: Done.

packages/blade/package.json
```diff
@@ -127,6 +127,7 @@
     "watch:test:react-native": "yarn test:react-native --watch --onlyChanged",
     "chromatic": "npx chromatic",
     "publish-npm": "node ./scripts/publishToNpm.js",
+    "bundle-size": "node ./scripts/generateBundleSizeInfo.js",
```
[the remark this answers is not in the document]
snitin315: Done. We don't need to run the entire build, just the `react:prod` build. Added `generate-bundle-size-info` & `pregenerate-bundle-size-info` (this will automatically run before `generate-bundle-size-info`)

.github/actions/install-dependencies/action.yml
```diff
@@ -25,6 +25,6 @@ runs:
       if: steps.cache.outputs.cache-hit != 'true'
 
     - name: Run postinstall script
-      run: yarn postinstall && yarn workspace @razorpay/blade postinstall
+      run: yarn postinstall
```
chaitanyadeorukhkar, line 28: Why this change?

packages/blade/.gitignore
```diff
@@ -54,6 +54,8 @@ yarn-error.log
 
 # Bundle artifact
 *.jsbundle
+.size-limit.json
+PRBundleSizeStats.json
```
chaitanyadeorukhkar, line 58: Lets use camelCase instead of PascalCase. Its more consistent with our practices
```suggestion
prBundleSizeStats.json
```

packages/blade/scripts/generateBundleDiff.js
```diff
@@ -0,0 +1,92 @@
...
+
+  // Calculate the size differences and create a formatted diff table
+  bundleDiff.forEach((component) => {
+    if (component.name === 'Base') {
+      return;
+    }
+
+    const currentComponent = currentBundleSizeStats.find((stat) => stat.name === component.name);
+    const baseComponent = baseBundleSizeStats.find((stat) => stat.name === component.name);
+    const emptyProjectSize =
```
chaitanyadeorukhkar, line 46: Add comments on why we are doing this. Hard for people to get context otherwise

packages/blade/scripts/generateBundleSizeInfo.js
```diff
@@ -0,0 +1,127 @@
...
+
+    if (exportedComponents.length > 0) {
+      const imports = exportedComponents.join(', ');
+
+      runSizeLimit({ name: imports, importedComponents: `{ ${imports} }` });
+    }
+  });
+
+  // Write the gathered size information to the specified file
+  const filename = process.env.BUNDLE_SIZE_STATS_FILENAME || 'PRBundleSizeStats.json';
```
chaitanyadeorukhkar, line 123:
```suggestion
  const filename = process.env.BUNDLE_SIZE_STATS_FILENAME || 'prBundleSizeStats.json';
```

packages/blade/scripts/generateBundleSizeInfo.js
```diff
@@ -0,0 +1,127 @@
...
+  fs.writeFileSync(path.resolve(__dirname, `../${filename}`), JSON.stringify(sizes));
```
chaitanyadeorukhkar, line 124: Please run prettier 😅 

Its a file on master branch, lets not keep it unreadable

.github/actions/install-dependencies/action.yml
snitin315: postinstall script is no longer required, it was just running patch-package for bundlemon & we no longer depend on bundlemon.

chaitanyadeorukhkar (approved):

snitin315:
Let's see how it goes 🙏🏻

PR #1981, opened 2024-01-18
fix: remaining rebranded stories

chaitanyadeorukhkar:


anuraghazra:
You can revert the bottomsheet changes, I've done that in my PR with some other changes and fixed that scroll getting locked on docs panel issue.

PR #1985, opened 2024-01-19
test(e2e): add popover e2e

anuraghazra:


packages/blade/src/components/Popover/__tests__/Popover.test.stories.tsx
```diff
@@ -0,0 +1,271 @@
...
+      </PopoverComponent>
+      <Button onClick={toggle}>Controlled Show</Button>
+    </Box>
+  );
+};
+
+TestPopoverControlled.play = async () => {
+  const { getByRole, queryByText } = within(document.body);
+  const popoverContent = 'Hello World';
+  const toggleButton = getByRole('button', { name: 'Controlled Show', hidden: true });
```
kamleshchandnani, line 86: why do we need `hidden: true`? this button will always be visible right?

anuraghazra: Popover traps focus when open, that means when popover is open it will make everything else in the body invisible to screen readers.

kamleshchandnani (approved):

PR #1987, opened 2024-01-22
feat(Amount): add new props & remove `prefix`

snitin315:
- Removed `prefix` prop in favor of the new `currencyIndicator` prop.
- Following new props have been added:
  - **`currencyIndicator`**:  Determines the visual representation of the currency (symbol or code).
  - **`isStrikethrough`**: If true, the amount value will have a line through it.
- Updated the codemod & migration guide.


packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -41,24 +41,30 @@ type AmountCommonProps = {
...
    *
    * @default 'INR'
    * */
   currency?: Currency;
+  /**
+   * If true, the amount text will have a line through it.
+   *
+   * @default false
+   */
+  isStrikethrough?: boolean;
```
kamleshchandnani, line 67: nit: wouldn't `shouldStrikethrough` be more suitable?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -112,18 +119,30 @@ const AmountValue = ({
...
+
+        {isStrikethrough && (
+          <BaseBox
+            // @ts-expect-error - intentionally setting the border color to the color prop for this hacky strikethrough
+            borderBottomColor={amountValueColor}
+            borderBottomWidth={type === 'body' ? 'thin' : 'thicker'}
+            borderBottomStyle="solid"
+            position="absolute"
+            width="100%"
+            top="50%"
```
kamleshchandnani, line 143: will top 50% be always correct? we use rems and if we increase or decrease fontsize of browser then I hope the place of line doesn't changes?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -99,11 +106,11 @@ const AmountValue = ({
...
     const decimal = value.split('.')[1];
 
-    // Native does not support alignItems of Text inside a div, insted we need to wrap is in a Text
-    const AmountWrapper = getPlatformType() === 'react-native' ? Text : React.Fragment;
+    // Native does not support alignItems of Text inside a div, instead we need to wrap is in a Text
+    const AmountWrapper = isReactNative ? BaseText : BaseBox;
 
     return (
-      <AmountWrapper>
+      <AmountWrapper position="relative">
```
chaitanyadeorukhkar, line 113: Amount is a Typography component. Can we avoid adding a div to it?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -274,24 +313,40 @@ const _Amount = ({
...
-          {currencyPrefix}
-        </BaseText>
+        {currencyPosition === 'left' && (
+          <BaseText
+            marginRight="spacing.1"
+            fontWeight={weight}
+            fontSize={currencyFontSize}
+            color={amountValueColor}
+            as={isReactNative ? undefined : 'span'}
+            opacity={isAffixSubtle ? 0.5 : 1}
```
chaitanyadeorukhkar, line 323: We're using opacity instead of a color? Is that the correct usage? What if the amount component is placed on a non-white background. It will change actual color depending on the background then. Can you check with design once?

packages/blade/src/components/Typography/BaseText/BaseText.web.tsx
```diff
@@ -84,6 +87,7 @@ export const BaseText = ({
       textAlign={textAlign}
       numberOfLines={truncateAfterLines}
       wordBreak={wordBreak}
+      opacity={opacity}
```
chaitanyadeorukhkar, line 90: I don't think we should expose opacity. This will make base text colors unpredictable and will always rely on the background color. Lets speak with the design team once to confirm this

chaitanyadeorukhkar:
it should strikethrough across the whole amount right? So the rupee symbol should also have a strikethrough? But seems like its only the numbers that have it right now

Screenshot 2024-01-23 at 4 32 28 PM

packages/blade/src/components/Amount/Amount.tsx
snitin315: This is needed for the strikethrough. Maybe we can change it to `Text` here, but we need at least one `div` below for the strike-through.

snitin315:
>it should strikethrough across the whole amount right? So the rupee symbol should also have a strikethrough? But seems like its only the numbers that have it right now

@chaitanyadeorukhkar This is as per the designs.

Screenshot 2024-01-23 at 4 39 33 PM

packages/blade/src/components/Collapsible/__tests__/__snapshots__/Collapsible.native.test.tsx.snap
```diff
@@ -342,19 +344,14 @@ exports[`<Collapsible /> should render with CollapsibleButton 1`] = `
...
-                      lineHeight={100}
+                      data-blade-component="base-text"
+                      position="relative"
                       style={
                         [
                           {
                             "color": "hsla(212, 39%, 16%, 1)",
                             "fontFamily": "Inter",
-                            "fontSize": 14,
+                            "fontSize": 16,
```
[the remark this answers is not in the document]
snitin315: I think because we updated the component from `Text` to `BaseText`. There were some ts issues with `Text`, I'll check once again.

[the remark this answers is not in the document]
snitin315: Resolved.

packages/blade/src/components/Typography/BaseText/BaseText.web.tsx
snitin315: As consulted with design, we have now increased the opacity to 64%.

packages/blade/src/components/Amount/Amount.tsx
snitin315: I'm fine with either `isStrikethrough` or `hasStrikethrough `.

packages/blade/src/components/Amount/Amount.tsx
snitin315: Yes, 50% will always be in the middle of the amount value.

snitin315:
> > it should strikethrough across the whole amount right? So the rupee symbol should also have a strikethrough? But seems like its only the numbers that have it right now
> 
> @chaitanyadeorukhkar This is as per the designs.


We have updated it to strike through whole component (including currency-symbol/code) after consulting with the designers.

packages/blade/src/components/Amount/amountTokens.ts
```diff
@@ -220,11 +220,19 @@ const getCurrencyAbbreviations = (currency: Currency): CurrencyAbbreviation[] =>
   ];
 };
 
+const currencyPositionMapping: Record<string, string> = {
+  DZD: 'right',
+  BHD: 'right',
+  OMR: 'right',
+  KWD: 'right',
+};
```
anuraghazra, line 228: Will these get filled with more currencies later?

snitin315: Yes.

anuraghazra (approved):

PR #1988, opened 2024-01-22
test: add tooltip e2e

anuraghazra:


kamleshchandnani:
overall lgtm except the custom interactive trigger which shows tooltip in the center
image

anuraghazra:
Done

packages/blade/src/components/Tooltip/__tests__/Tooltip.test.stories.tsx
```diff
@@ -96,6 +96,10 @@ const CustomTrigger = React.forwardRef<
   ) => {
     return (
       <BaseBox
+        width="80px"
+        height="80px"
```
kamleshchandnani, line 100: why pass manually? this would cause issue when consumer use it the way you were using before right? also why use basebox? we can use `Box` right?

anuraghazra: This is just an example showing how to pass proper props like `onMouseOver` `onMouseHover` `ref` `tabIndex` to properly create a trigger for tooltip. The styling hardly matters in this example.

anuraghazra: Plus Box doesn't even support onBlur onFocus etc props.

kamleshchandnani: Sure but what i'm trying to ask is if consumer will create custom tooltip they'll have to pass these things as well right?

anuraghazra: Yes. events are required for users to create custom triggers. This is already well documented in the main story example

https://blade.razorpay.com/?path=/story/components-tooltip--with-custom-trigger

kamleshchandnani (approved):

PR #1990, opened 2024-01-23
docs(blade): Toast API decisions

anuraghazra:
## [Formatted Document](https://github.com/razorpay/blade/blob/anu/toast-api/packages/blade/src/components/Toast/_decisions/decisions.md)


packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+- Auto dismissable / manual dismissable
+- Positioning
+- `informational` or `promotional` types
+
+## Anatomy
+
+<img src="./toast-anatomy.png" width="100%" alt="Toast anatomy" />
+
+## Simple Usage
+
```
kamleshchandnani, line 24: add information about
* imperative API and why
* Explicitly mention we are using 3rd party lib as the base

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+      Do payment
+    </Button>
+  );
+};
+```
+
+<img src="./example-promotional-toast.png" width="380" alt="Promotional toast example" />
+
+
+## Motion
```
kamleshchandnani, line 218: motion is still pending?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+
+- TBD
+
+## Accessibility
+
+- Toast components will follow the WAI-ARIA guidelines for [alert](https://www.w3.org/WAI/ARIA/apg/patterns/alert/examples/alert/). For error toast we will use the `alert` role and for informational toast we will use the `status` role.
```
kamleshchandnani, line 224: can you mention the exact attributes? also the alert role will be applied on toast or the parent container can you mention that too?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+const HomePage = () => {
+  const { showToast } = useToast();
+
+  return (
+    <Button
+      onClick={() => {
+        showToast({
+          type: 'informational',
+          color: 'success',
+          content: 'Payment Successful',
```
kamleshchandnani, line 37: can you put jsx here to demonstrate the example from anatomy with bold text? although its mentioned in props but looking at example gives more context

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+
+  /**
+   * If the type is `promotional`, the content will be `React.ReactNode`
+   */
+  content: string | React.ReactNode;
+
+  /**
+   * Can be used to render an icon
+   */
+  leading?: IconComponent;
```
kamleshchandnani, line 97: since we don't have trailing counterpart and we know that icon will always be leading wouldn't it be better to name it just `icon`? so it will also be consistent with rest of the components?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
+# Toast 🏷️
+
+Toasts are used to provide feedback to the user after an important action has been performed. 
+Toasts can also be used to provide feedback to the user when a system event occurs, such as when a file is saved.
+
+<img width="426" alt="image" src="./toast-thumbnail.png" />
```
kamleshchandnani, line 6: add image of promotional toast as well

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+## Features
+
+- Stackable
+- Auto dismissable / manual dismissable
+- Positioning
+- `informational` or `promotional` types
+
+## Anatomy
+
+<img src="./toast-anatomy.png" width="100%" alt="Toast anatomy" />
```
kamleshchandnani, line 21: show promotional type here as well

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+   */
+  leading?: IconComponent;
+
+  /**
+   * If true, the toast will be dismissed after few seconds
+   * 
+   * Duration for promotional toast is 6s
+   * Duration for informational toast is 4s
+   * 
+   * @default false
```
kamleshchandnani, line 105: the default shall be true IMO? because majority of the times you want this distracting pice to be autohidden?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+   */
+  autoDismiss?: boolean;
+
+  /**
+   * Called when the toast is dismissed or duration runs out
+   */
+  onDismissButtonClick?: () => void;
```
kamleshchandnani, line 112:
```suggestion
  onDismiss?: () => void;
```

to avoid confusion?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+    onClick?: () => void;
+    isLoading? boolean;
+  }
+
+  /**
+   * Forwarded to react-hot-toast
+   * 
+   * This can be used to programatically update toasts by providing an id
+   */
+  id?: string;
```
kamleshchandnani, line 128: can we call it toastId to not confuse it with the HTML id attribute? or react-hot-toast considers this as an HTML id attribute only?

Afraid that people might hack the UI using id

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+
+  /**
+   * Primary action of toast
+   */
+  action?: {
+    text: string;
+    onClick?: () => void;
+    isLoading? boolean;
+  }
+
```
kamleshchandnani, line 122: what about position it will always be bottom left?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+}
+```
+
+### useToast
+
+The useToast hook will return few modified methods from [react-hot-toast](https://react-hot-toast.com/docs/toast): 
+
+```ts
+type useToastReturnType = {
+  /**
```
kamleshchandnani, line 138: we shall also give an option to know the current id of the toast that is opened?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+  return (
+    <Button
+      onClick={() => {
+        showToast({
+          type: 'promotional',
+          content: (
+            <Box>
+              <Heading>Payment Successful</Heading>
+              <Text>Amount: ₹100</Text>
+            </Box>
```
kamleshchandnani, line 200: we don't need to put heading and description structure for promotional type?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+#### Removing a toast
+
+react-hot-toast provides this functionality, for more info see [react-hot-toast docs](https://react-hot-toast.com/docs/toast#dismiss-toast-programmatically)
+
+```jsx
+import { BladeProvider, ToastContainer, useToast } from "@razorpay/blade/components"
+
+const Example = () => {
+  const toastId = React.useRef(null);
+  const { showToast, dismissToast } = useToast();
```
kamleshchandnani, line 163: toast will have a context at the app level right? can't see that in the example?

packages/blade/src/components/Toast/_decisions/decisions.md
chaitanyadeorukhkar: +1. Lets just call it `icon` similar to how we do it for button

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+    >
+      Show Toast
+    </Button>
+  );
+};
+
+const App = () => {
+  return (
+    <BladeProvider>
+      <ToastContainer position="bottom-left" />
```
chaitanyadeorukhkar, line 60: Will there be multiple toast containers in an app?

Can there not be a single ToastProvider that is part of BladeProvider and `useToast` can decide the toast position?

packages/blade/src/components/Toast/_decisions/decisions.md
chaitanyadeorukhkar: +1 we should autodismiss by default

packages/blade/src/components/Toast/_decisions/decisions.md
chaitanyadeorukhkar: Maybe we need to have `onDismissClicked` & `onAutoDismiss`?

packages/blade/src/components/Toast/_decisions/decisions.md
chaitanyadeorukhkar: As I can see on design its toast stacking one above the other (not on top of each other). 

When toasts cross a threshold and start fading away, is that container scrollable? Can we scroll to older toasts that are faded away but not dismissed?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,251 @@
...
+
+- Toast components will follow the WAI-ARIA guidelines for [alert](https://www.w3.org/WAI/ARIA/apg/patterns/alert/examples/alert/). For error toast we will use the `alert` role and for informational toast we will use the `status` role.
+
+
+## References
+
+- https://react-hot-toast.com/docs
+- https://chakra-ui.com/docs/components/toast/usage
+
+## Open Questions
```
chaitanyadeorukhkar, line 232: Unclear on how the `ToastContainer` plays a role here and what happens if there are multiple `ToastContainers` with multiple positions?

chaitanyadeorukhkar: Why can't it be baked with BladeProvider? (other than bundle size bloat)

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: Basically this will be same a11y as our Alert component. Just a role prop. 

And anyways the react-hot-toast will handle the accessibility props out of the box and we will just spread the props.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: In promotional toast the leading can also become an asset, so `leading` maybe more flexible naming here. 

image

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: Okay, makes sense. 

Discussed with RK, for informational the autoDismiss we can keep true, and for promotional we can keep it false.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: Roast position can only be change globally via the ToastContainer, and not on the individual toasts.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: "current" can also be multiple toasts, we can give an option like: 

const { toasts } = useToasts();

but not sure about the usecase.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: No the promotional has a slot, it can have product images & other things, it's not just title+description.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: Maybe, I'm not sure right now actually, so react-hot-toast works in a way that it doesn't need ReactContext. 
In the current POC that I've done that also didn't need a toast context.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: There can only be 1. 

> Can there not be a single ToastProvider that is part of BladeProvider and useToast can decide the toast position?

That should be possible, we can keep the ToastContainer inside BladeProvider but the thing is we also may need to pass some props to ToastContainer, how should we handle that?

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: ToastContainer is the container where all the toasts are rendered inside: 

image

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: anatomy is the same. leading, button, close button & content.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: There is no JSX for the Toast component itself. 

the `<Toast />` component won't be exposed to the consumer, only the API for creating new toast. eg `showToast` `dismissToast` etc

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: Added

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: Yeah maybe we need this separated out.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: done.

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Simple Usage
+
+The blade toasts will use [react-hot-toast](https://react-hot-toast.com/) under the hood with a similar imperative API to show, dismiss and create new toasts without needing for consumer to handle positional or stacking logic. 
+
+```jsx
+import { BladeProvider, ToastContainer, useToast } from "@razorpay/blade/components"
+
+const HomePage = () => {
+  const { showToast } = useToast();
```
parichay28, line 31: `react-hot-toast` supports [imperative api](https://react-hot-toast.com/docs) which is independent of hooks, will blade support that? Also with the hook based api, how do we plan to support class components?

packages/blade/src/components/Toast/_decisions/decisions.md
parichay28: perhaps single onDismiss with type = 'auto' | 'onclick'?

packages/blade/src/components/Toast/_decisions/decisions.md
parichay28: Current toast message in X allows different positions per toast. Will check with x design team if this is needed going fwd

packages/blade/src/components/Toast/_decisions/decisions.md
chaitanyadeorukhkar: What do we mean by asset? It can be an image or text or video? Any restrictions or is it a slot?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+
+  /**
+   * @default `neutral`
+   */
+  color: 'neutral' | 'positive' | 'negative' | 'warning' | 'information'
+
+  /**
+   * If the type is `promotional`, the content will be `React.ReactNode`
+   */
+  content: string | React.ReactNode;
```
nashcheez, line 94: Even in informational toast I think Design concluded that we could have bold and italics text in the content, is that not true?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+   */
+  autoDismiss?: boolean;
+
+  /**
+   * Duration after which the toast will be dismissed (in ms)
+   * 
+   * Duration for promotional toast is 8s
+   * Duration for informational toast is 4s
+   */
+  duration?: 8000 | 4000;
```
nashcheez, line 115: Can a promotional toast be live for 4s and an informational toast be up for 8s?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+
+## References
+
+- https://react-hot-toast.com/docs
+- https://chakra-ui.com/docs/components/toast/usage
+
+## Open Questions
+
+- Q. What should be the default duration for auto dismissable toasts?
+  - A. 4s for informational toasts and 6s for promotional toasts
```
nashcheez, line 243: Should be a design call based on the durations that are prevalent today. Why use three different durations - 4, 6 & 8? why can't we limit it to any 2?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+
+- Q. Should we call it `onDismissButtonClick` or `onDismiss`? 
+  
+- Q. Should the dismiss handler be called even when the toast is auto dismissed? Or should we have different handlers for auto dismiss and manual dismiss? (eg: `onAutoDismiss` `onDismissButtonClick`)
+
+- Q. In the `useToast` hook should we call the returned functions `showToast`/`dismissToast` or `show`/`dismiss`?
+
+If we call them `show` & `dismiss` consumer can do this and might look more cleaner:
```
nashcheez, line 251: +1 for `show` and `dismiss`

packages/blade/src/components/Toast/_decisions/decisions.md
snitin315: `onAutoDismiss` & `onManualDismiss`

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+- https://chakra-ui.com/docs/components/toast/usage
+
+## Open Questions
+
+- Q. What should be the default duration for auto dismissable toasts?
+  - A. 4s for informational toasts and 6s for promotional toasts
+
+- Q. Should we call it `onDismissButtonClick` or `onDismiss`? 
+  
+- Q. Should the dismiss handler be called even when the toast is auto dismissed? Or should we have different handlers for auto dismiss and manual dismiss? (eg: `onAutoDismiss` `onDismissButtonClick`)
```
snitin315, line 247: `onAutoDismiss` & `onManualDismiss`.

packages/blade/src/components/Toast/_decisions/decisions.md
snitin315: It can be, for example in the case of failure/error messages, we may want it to show the error for 8s.

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+### useToast
+
+The useToast hook will return few modified methods from [react-hot-toast](https://react-hot-toast.com/docs/toast): 
+
+```ts
+type useToastReturnType = {
+  /**
+   * @returns id of the toast
+   */
+  showToast: (toast: Toast) => string;
```
snitin315, line 149: Just for clarity, initially, I interpreted it as the component. From the examples, the param is options. 
```suggestion
  showToast: (toast: ToastOptions) => string;
```

packages/blade/src/components/Toast/_decisions/decisions.md
snitin315: Also, does the counting stop when we hover over the toast?

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+
+- https://react-hot-toast.com/docs
+- https://chakra-ui.com/docs/components/toast/usage
+
+## Open Questions
+
+- Q. What should be the default duration for auto dismissable toasts?
+  - A. 4s for informational toasts and 6s for promotional toasts
+
+- Q. Should we call it `onDismissButtonClick` or `onDismiss`?
```
snitin315, line 245: `onManualDismiss`

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+```jsx
+const App = () => {
+  const toast = useToast();
+
+  toast.show(); // <-- they will also get auto complete when writing `toast.`
+  toast.dismiss();
+}
+```
+
+- Q. In design we are restricting the Toast position to be bottom-left. In that case should we also do the same or should we allow all the positions & set the default to bottom-left? (If we allow all the positions, we will have to add some additional logic to handle stacking/animation of toasts coming from top instead of bottom)
```
snitin315, line 262: We can initially implement a bottom-left position only and later add more positions as use cases come.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: Don't have any plans to support both APIs, currently only exposing it via a hook which wraps the imperative API. 

If there's a consumer ask for a separate imperative API we can provide that in future.

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Design
+
+- [Toast - Figma Design](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade-DSL?type=design&node-id=7665-27414&mode=design&t=UNInCMmP1iFCu9je-0)
+
+## Features
+
+- Stackable
+- Auto dismissable / manual dismissable
+- Positioning
```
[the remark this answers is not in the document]
anuraghazra: Yeah, will remove this.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: I'll get this checked, in that case we can accept React.ReactNode all the time.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: We will set defaults, but product can override the duration based on their usecase.

anuraghazra: > Also, does the counting stop when we hover over the toast?

Yes

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: We are limiting to 2 only & this is defined by design side only. 

4s for informational toasts and 6s for promotional toasts.

packages/blade/src/components/Toast/_decisions/decisions.md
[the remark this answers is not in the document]
anuraghazra: Mentioned already for promotional & informational toast. 4 & 8s

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: No restrictions, it's a slot. we should atleast keep the prop name generic to cater to future usecase. 

Same thing we did for BottomSheetHeader too.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: Changed to react.reactNode

packages/blade/src/components/Toast/_decisions/decisions.md
parichay28: This is resolved ([ref](https://razorpay.slack.com/archives/CMQ3RBHEU/p1707453288701379?thread_ts=1706686131.017819&cid=CMQ3RBHEU))

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: calling it show, dismiss

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: Stacking is resolved.

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,266 @@
...
+}
+```
+
+## Props
+
+### ToastContainer
+
+```ts
+type ToastContainer = {
+  position: 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';
```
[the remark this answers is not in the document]
anuraghazra: bottom-left. This prop is removed now.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: Going ahead with onDismissButtonClick, let's not have onManual, onAutoDismiss because there is no controlled state that users can manage for toasts so they don't really need to know if a toast is auto dismissed or not.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: We are only adding bottom-left.

packages/blade/src/components/Toast/_decisions/decisions.md
anuraghazra: added

saurabhdaware:
Can you check these 2 comments? they are related to API decision

- https://github.com/razorpay/blade/pull/2000#discussion_r1485772469
- https://github.com/razorpay/blade/pull/2000#discussion_r1485772557

packages/blade/src/components/Toast/_decisions/decisions.md
kamleshchandnani: maybe we can have icon and asset as a naming then?

anuraghazra: Two props? Will cause impossible states. 

<Toast icon={InfoIcon} asset={CircleIcon} />

anuraghazra:
> Can you check these 2 comments? they are related to API decision

Responded.

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Open Questions
+
+- Q. What should be the default duration for auto dismissable toasts?
+  - A. 4s for informational toasts and 8s for promotional toasts
+
+- Q. Should we call it `onDismissButtonClick` or `onDismiss`? 
+  
+- Q. Should the dismiss handler be called even when the toast is auto dismissed? Or should we have different handlers for auto dismiss and manual dismiss? (eg: `onAutoDismiss` `onDismissButtonClick`)
+  - A. We will call it `onDismissButtonClick` and it will only be called when the user clicks on the dismiss button.
```
kamleshchandnani, line 245: if product needs to track the event on auto dismiss then how can consumer handle that??

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,267 @@
...
+### useToast
+
+The useToast hook will return few modified methods from [react-hot-toast](https://react-hot-toast.com/docs/toast): 
+
+```ts
+type useToastReturnType = {
+  /**
+   * @returns id of the toast
+   */
+  show: (toast: Toast) => string;
```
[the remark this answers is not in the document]
anuraghazra: Id of the toast just being created via this show function.

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Anatomy
+
+### Informational Toast
+
+<img src="./toast-anatomy.png" width="100%" alt="Toast anatomy" />
+
+### Promotional Toast
+
+<img src="./example-promotional-toast.png" width="380" alt="Promotional toast example" />
```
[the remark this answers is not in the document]
anuraghazra: > add anatomy wala screenshot here
Design hasn't created anatomy illustration for promo toast. // @saurav12 

> primarily because icon is asset in promotional
There is no difference of anatomy between info/promo toast. Both are 1:1 

- Leading: icon or assset
- Content: JSX element (to put bold/italic text inside info)
- Action Button
- Dismiss Button

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,267 @@
...
+type ToastProps = {
+  /**
+   * @default `informational`
+   */
+  type?: 'informational' | 'promotional';
+
+  /**
+   * @default `neutral`
+   */
+  color?: 'neutral' | 'positive' | 'negative' | 'warning' | 'information'
```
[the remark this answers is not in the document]
anuraghazra: We just ignore this prop in case of promo.

packages/blade/src/components/Toast/_decisions/decisions.md
```diff
@@ -0,0 +1,267 @@
...
+   */
+  leading?: React.ReactElement;
+
+  /**
+   * If true, the toast will be dismissed after few seconds
+   * 
+   * @default true - for informational toast
+   * @default false - for promotional toast
+   */
+  autoDismiss?: boolean;
```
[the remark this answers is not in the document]
anuraghazra: That would be a product decision. We do not have a defined usecase from design side.

packages/blade/src/components/Toast/_decisions/decisions.md
kamleshchandnani: you can do multiple things like for eg: a user was shown a toast and they navigate to a different page then when the toast is dismissed you can track the URL of the page. this way you can identify a pattern that after performing some actions what are the pages a user navigates to and get some insights around user behavior

anuraghazra: > a user was shown a toast and they navigate to a different page then when the toast is dismissed you can track the URL of the page

That is possible. 

`onDismissButtonClick={() => naigateToPage()}`

But don't think it will be a very good idea for toast to automatically navigate the page after dismiss. Navigation should happen after action.

anuraghazra: FYI, 

react-hot-toast doesn't have a [callback](https://github.com/timolins/react-hot-toast/issues/174) that it calls when a toast gets auto dismissed, so it's not even possible to have an `onAutoDismiss` method. 

There is a https://github.com/timolins/react-hot-toast/issues/63#issuecomment-1344772935 but doing that means you need to keep track of each and every toast's timer manually, too messy.

kamleshchandnani (approved):

PR #2003, opened 2024-02-02
feat: i18nify SDK integrated for `Amount` component

tarun-khanna:
## Description

This PR integrates i18nify (Atlas) SDK in Blade's Amount Component.  

## Changes
1. **Modified internal logic within Amount component to format number by utilising i18nify APIs instead.** 
1.1 Amount formatting (commas and spaces) will be now driven by locale state set in i18nify SDK. Earlier, Blade supported only 2 types for formatting `en-IN` and `en-US`. If currency was INR, it chose `en-IN` otherwise `en-US`.

| Before            | After              |
| ------------- | ------------- |
| Screenshot 2024-02-02 at 2 21 18 PM | Screenshot 2024-02-02 at 2 21 53 PM  |

1.2. CurrencyList to be now retrieved from getCurrencyList API of i18nify. 
Screenshot 2024-02-02 at 2 25 10 PM

1.3. Denomination symbol will now be localized based on the locale set. Earlier, the abbreviations were localized only for India and US. If the currency was INR, it localized it to K, lac, Crores otherwise K, M, B..

| Before            | After              |
| ------------- | ------------- |
image | image | 

1.4. Denomination position will also be decided by i18nify based on the locale state set in the SDK. Earlier, denomination symbol was simply placed opposite to the placement of currency symbol.

| Before            | After              |
| ------------- | ------------- |
Screenshot 2024-02-02 at 2 26 44 PM | Screenshot 2024-02-02 at 2 27 29 PM |

1.5. Currency symbol will also be based on the user's locale and not the native language to give better readability to that user. 

| Before            | After              |
| ------------- | ------------- |
Screenshot 2024-02-02 at 2 52 13 PM | Screenshot 2024-02-02 at 2 52 29 PM |

2. **Updated Unit test cases for new contracts of functions that have changed.** 

3. **Added a new story to display Amount in different locales**
image
image

_**Note => We are currently exploring the appropriate short form for denoting 1000 in the Indian locale—whether it should be 1K or 1T. Once a decision is reached, I will add the necessary adjustments in this PR before merging.**_
Update => We have concluded to keep it T as it is a standard norm and followed by other global applications too. For more details, refer [here](https://docs.google.com/document/d/1ImrtEX3H_QQp4Zb06Dousmmu1hOdsb9lUQPbrP_3YzA/edit#heading=h.nvwaidgytqdu). 

**References** 
1. [I18nify Tech spec](https://docs.google.com/document/d/1jr4uoI19-qvgohuN1fosv2lSrmINhxXkZVRHzvd9WBo/edit#heading=h.h91088eh91av)
2. [i18nify repo](https://github.com/razorpay/i18nify)
3. [Slack thread](https://razorpay.slack.com/archives/C05RBSKC6KD/p1705482465156459)
5. [Case study on Stripe](https://docs.google.com/document/d/1ImrtEX3H_QQp4Zb06Dousmmu1hOdsb9lUQPbrP_3YzA/edit#heading=h.6jn10buryynj)

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/package.json
```diff
@@ -283,7 +284,8 @@
     "react-native-svg": "^12.3.0",
     "react-native-gesture-handler": "^2.9.0",
     "@gorhom/bottom-sheet": "^4.4.6",
-    "@gorhom/portal": "^1.0.14"
+    "@gorhom/portal": "^1.0.14",
+    "@razorpay/i18nify-js": "^1.3.1"
```
snitin315, line 288: Is this a mandatory peer dep? Does this work with react-native too? If not, we may have to make it optional under the `peerDependenciesMeta` key.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -1,14 +1,9 @@
...
-import {
-  normalAmountSizes,
-  getCurrencyAbbreviations,
-  currencyIndicatorMapping,
-  subtleFontSizes,
-  amountLineHeights,
-  currencyPositionMapping,
-} from './amountTokens';
+import type { CurrencyCodeType } from '@razorpay/i18nify-js';
+import { formatNumber, formatNumberByParts } from '@razorpay/i18nify-js';
```
snitin315, line 4: This file is common for both native & web, so any package imported has to be compatible on both platforms.

packages/blade/package.json
tarun-khanna: Yes it's a mandatory peer dep and it works with react-native too. 
Check this [section](https://docs.google.com/document/d/1jr4uoI19-qvgohuN1fosv2lSrmINhxXkZVRHzvd9WBo/edit#heading=h.84rlxo6w3k3e) in tech spec for an integration example on react-native.

packages/blade/src/components/Amount/Amount.tsx
tarun-khanna: yep, It's compatible.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -143,83 +139,78 @@ const AmountValue = ({
       color={amountValueColor}
       lineHeight={amountLineHeights[type][size]}
     >
-      {value}
+      {amount.formatted}
```
snitin315, line 142: Earlier `<Amount value={123456.789} suffix="humanize">` was rendering `123.46K`, but now it renders only `123K` ( or `1.2L`). Is this formatting expected? not showing decimals for `K` but showing it in case of `L`.

snitin315 (commented):
Can you also add the following:
- A story that showcase i18n support for multiple locales.
- Changeset with upgrade guide

snitin315:
Are these changes backward compatible with the old Amount component? For instance, if a project is using the Amount component which hasn't integrated the i18nify SDK, would it cause any issues? Is just installing the peerDep `@razorpay/i18nify-js` enough to upgrade or they will need some additional configuration steps as well?

packages/blade/src/components/Amount/Amount.tsx
tarun-khanna: @snitin315 It will render `123.46K` for `en-US` locale but `1.2L` for `en-IN` locale. 
The decimals are omitted based on the locale and rounding off of digits.

tarun-khanna:
> Thanks for the PR, overall looks good. Can you also add the following:
> 
> * A story that showcases i18n support for multiple locales.
> * Changeset with upgrade guide

Have added the story and changeset with guide. @snitin315 
image

.changeset/purple-flies-end.md
```diff
@@ -0,0 +1,13 @@
+---
+"@razorpay/blade": minor
```
snitin315, line 2: Ideally, this should be treated as a breaking change since users need manual intervention (installing peerDep) to upgrade.

We can make an exception for this PR, to avoid a major release. I'll discuss this with the team.

cc @anurag-rzp @saurabhdaware @kamleshchandnani @chaitanyadeorukhkar

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -217,3 +223,98 @@ StrikeThrough.args = {
...
+    country: 'France',
+    locale: 'fr-FR',
+  },
+  {
+    country: 'Germany',
+    locale: 'de-DE',
+  },
+];
+
+const I18nAmountWrapper = (args: AmountProps) => {
```
snitin315, line 251: Need to fix lint here.

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -217,3 +223,98 @@ StrikeThrough.args = {
...
+        </DropdownOverlay>
+      </Dropdown>
+    </>
+  );
+};
+
+const I18nAmountTemplate: StoryFn<typeof AmountComponent> = (args) => {
+  return (
+    <I18nProvider>
+      <BaseBox justifyContent="flex-start" maxHeight="300px" overflowY="auto">
```
snitin315, line 301: I think here you can set a minHeight to ensure the dropdown gets sufficient space to show options.

snitin315:
@tarun-khanna Also you need to update snapshots, you can run `yarn test:react -u` & `yarn test:react-native -u` to update them.

packages/blade/src/components/Amount/Amount.stories.tsx
tarun-khanna: Fixed.

packages/blade/src/components/Amount/Amount.stories.tsx
tarun-khanna: It looks a bit odd but does the job for now. @snitin315 , have updated.
Screenshot 2024-02-06 at 10 13 47 PM

snitin315:
@tarun-khanna There are still some react-native snapshots failing, can you update them too?

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -217,3 +223,98 @@ StrikeThrough.args = {
   isStrikethrough: true,
 };
 StrikeThrough.storyName = 'Strike Through';
+
+// TODO: Replace below with i18nify getDefaultLocales API
```
saurabhdaware, line 227: is it yet to be implemented?

.changeset/purple-flies-end.md
kamleshchandnani: We can do it as minor, should be fine

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -143,83 +139,77 @@ const AmountValue = ({
...
+ * if suffix === 'decimals' => {
+    "formatted": "12,500.45",
+    "integer": "12,500",
+    "decimal": ".",
+    "fraction": "45",
+    "isPrefixSymbol": false,
+    "rawParts": [{"type": "integer","value": "12"},{"type": "group","value": ","},{"type": "integer","value": "500"},{"type": "decimal","value": "."},{"type": "fraction","value": "45"}]
+}
+ * else if suffix === 'humanize' => { formatted: "1.2T" }
+ * else => { formatted: "1,23,456" }
```
saurabhdaware, line 165: Shouldn't `currency` prop also internally handle default locale?

E.g. if I write 

```js
<Amount currency="INR" value={120000} /> 
```

It should print `INR 1,20,000` right?

Currently it prints `INR 120,000` unless locale is set correctly as per currency

.changeset/purple-flies-end.md
saurabhdaware: minor should be fine I believe since its just additional installation step and shouldn't require consumers to change anything in code as such. And even if it breaks, it breaks on development itself

tarun-khanna:
> @tarun-khanna There are still some react-native snapshots failing, can you update them too?

My bad, skipped these. Updated now @snitin315

snitin315 (dismissed):
LGTM, great job ⭐. Thank you for contributing and bearing with us for all of these back-and-forth discussions.

Moving this to L2 review.

packages/blade/src/components/Amount/Amount.tsx
tarun-khanna: @saurabhdaware 
No, the currency shouldn't define the default locale or formatting of the amount string. 
It has to be driven from user's browser locale settings. We also conducted a case study on Stripe for the same, you can find the document over [here](https://docs.google.com/document/d/1ImrtEX3H_QQp4Zb06Dousmmu1hOdsb9lUQPbrP_3YzA/edit#heading=h.6jn10buryynj). (attached in PR desc too)

Your locale in your browser must be `en`, `en-US` or `en-GB` which is why you're seeing amount being printed as `INR 120,000`. You can check the settings from here chrome://settings/languages#lang

packages/blade/src/components/Amount/Amount.stories.tsx
tarun-khanna: Yes this is supposed to be implemented in the upcoming sprints. As this is just for storybook, we can update this once we have it ready in `i18nify`.

packages/blade/src/components/Amount/Amount.tsx
snitin315: A currency can have multiple locale, it reads the locale from your browser, in your case it must be `en-US` instead of `en-IN`.

Screenshot 2024-02-07 at 3 43 23 PM

saurabhdaware: > A currency can have multiple locale, it reads the locale from your browser, in your case it must be en-US instead of en-IN.

Yes yes I meant, Do we know that all Amount instances on page should always be formatted on page-level or browser-configuration-level locale? or should Amount itself also have locale configuration to have consistent formatting

snitin315: > Amount itself also have locale configuration to have consistent formatting

The locale should be set via the i18n SDK via the consumers, if they don't set it will use browsers default which I believe is fine.

> In places like Pricing page or Landing pages, we might look for a consistent formatting of numbers according to currency no irrespective of the page language?

For such cases, the locale can be set by the consumers, right?

tarun-khanna: +1 to @snitin315 here. 
@saurabhdaware The consumers have the option to set the locale state and all of our APIs will internally pull that locale to drive formatting. We fallback to browser settings incase they don't set it. 

Giving an additional `locale` prop to Amount component was ruled out in our initial discussions because of 2 primary reasons 
1. It increases the complexity of props in Amount component. This data should rather come from a global state. 
2. Multiple Amount instances can then have their own unique formatting which will not be consistent with the page. 

lmk if this is clear, else we can jump over a short call and close this. 
// @kamleshchandnani

saurabhdaware: Makes sense. If a usecase comes where we want different Amount instances on page to be differently formatted, then we can come back and explore locale prop option.

saurabhdaware (approved):
Looks good to me. Thanks for the contributions 🥳

.changeset/purple-flies-end.md
```diff
@@ -0,0 +1,13 @@
...
+
+feat: Added internationalization in Amount component via i18nify.
+References
+- **i18nify-js:** https://www.npmjs.com/package/@razorpay/i18nify-js
+- **i18nify-react:** https://www.npmjs.com/package/@razorpay/i18nify-react
+
+How to update ? 
+1. Install i18nify peer dependency `yarn add @razorpay/i18nify-js`
+2. _[Optional]_: Install i18nify-react peer dependency to manage state effectively `yarn add @razorpay/i18nify-react`
+3. Install latest Blade `yarn add @razorpay/blade@latest`
```
saurabhdaware, line 13: Should we ask consumers to wrap their apps with I18nProvider here? otherwise all their existing Amount components will start getting formatted as their browser config no?

saurabhdaware (changes requested):
Oh 1 last doubt about changelog

.changeset/purple-flies-end.md
snitin315: We should add a note about this

tarun-khanna: IMO, existing Amount components should default back to browser settings only. 
But, I've added a brief context and guide in the changelog for developers who wish to maintain a fixed locale or if they provide users the option to change page locale. 

We should probably talk about `i18nify` and `i18nify-react` in installation steps in the Blade documnetation as well. 

_Note -> JFYI, `I18nProvider` (i18nify-react) is an optional feature. Nothing would change if developers wrap their application inside this provider but do not set any state._

.changeset/purple-flies-end.md
```diff
@@ -0,0 +1,23 @@
...
+- **i18nify-js:** https://www.npmjs.com/package/@razorpay/i18nify-js
+- **i18nify-react:** https://www.npmjs.com/package/@razorpay/i18nify-react
+
+**What changes ?**
+
+1. The `<Amount />` component will now automatically format numbers based on the user's browser locale. For example, `<Amount value={123456.789} currency="INR">` will render `₹1,23,456.79` for browsers with the `en-IN` default locale, whereas it will render `₹123,456.79` for browsers with the `en-US` locale.
+
+2. If you want to enable users to change the locale of your page, add the `@razorpay/i18nify-react` package and wrap your app inside the `I18nProvider`. Utilize the `setI18nState` utility to modify the locale. Please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react) for more details.
+
+3. Additionally, if you prefer to maintain a fixed locale for your page and amount component, simply enclose your app within `<I18nProvider data={{locale: 'locale-you-want'}}>..` . Please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react) for more details.
```
snitin315, line 17:
```suggestion
3. Additionally, if you prefer to maintain a fixed locale for your page and amount component, enclose your app within `<I18nProvider data={{locale: 'locale-you-want'}}>..`. For more details, please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react).
```

.changeset/purple-flies-end.md
```diff
@@ -0,0 +1,23 @@
...
+References**
+
+- **i18nify-js:** https://www.npmjs.com/package/@razorpay/i18nify-js
+- **i18nify-react:** https://www.npmjs.com/package/@razorpay/i18nify-react
+
+**What changes ?**
+
+1. The `<Amount />` component will now automatically format numbers based on the user's browser locale. For example, `<Amount value={123456.789} currency="INR">` will render `₹1,23,456.79` for browsers with the `en-IN` default locale, whereas it will render `₹123,456.79` for browsers with the `en-US` locale.
+
+2. If you want to enable users to change the locale of your page, add the `@razorpay/i18nify-react` package and wrap your app inside the `I18nProvider`. Utilize the `setI18nState` utility to modify the locale. Please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react) for more details.
```
snitin315, line 15:
```suggestion
2. If you want to enable users to change the locale of your page, add the `@razorpay/i18nify-react` package and wrap your app inside the `I18nProvider`. Utilize the `setI18nState` utility to modify the locale. For more details, please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react).
```

snitin315 (dismissed):
Thank you for your contribution ❤️

saurabhdaware:
Hi, we're just discussing if we should release it as major or minor change. Will update you by end of the day.

.changeset/purple-flies-end.md
saurabhdaware: > IMO, existing Amount components should default back to browser settings only.

The issue is the existing Amount components. For example, if someone is displaying `Rs 10L` right now. It will turn into `Rs 1M` without a breaking change. And Rs 10L -> Rs 1M is a big enough change for product and design.

Internally also we checked in team and most people have en-US set on their browser

image

.changeset/purple-flies-end.md
```diff
@@ -0,0 +1,23 @@
...
+1. The `<Amount />` component will now automatically format numbers based on the user's browser locale. For example, `<Amount value={123456.789} currency="INR">` will render `₹1,23,456.79` for browsers with the `en-IN` default locale, whereas it will render `₹123,456.79` for browsers with the `en-US` locale.
+
+2. If you want to enable users to change the locale of your page, add the `@razorpay/i18nify-react` package and wrap your app inside the `I18nProvider`. Utilize the `setI18nState` utility to modify the locale. For more details, please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react).
+
+3. Additionally, if you prefer to maintain a fixed locale for your page and amount component, enclose your app within `<I18nProvider data={{locale: 'locale-you-want'}}>..`. For more details, please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react).
+
+**How to update ?**
+
+1. Install i18nify peer dependency `yarn add @razorpay/i18nify-js`
+2. _[Optional]_: Install i18nify-react peer dependency to manage state effectively `yarn add @razorpay/i18nify-react`
```
saurabhdaware, line 22: We were discussing about breaking change internally. We decided to have this as minor release and add its changelog into v11 https://github.com/razorpay/blade/blob/master/packages/blade/upgrade-v11.md#amount because we very recently had a breaking release which people are yet to migrate to.

I think we'll have `@razorpay/i18nify-react` as mandatory dependency and ask consumers to add a `I18nProvider` on their end with expected locale. 

A bit risky to rely on browser locale for existing Amount components since our default so far has been en-IN behaviour so things might break for product if something like this changes - https://github.com/razorpay/blade/pull/2003/files#r1482641498

saurabhdaware (commented):
Hi things look good from code end. On Changelog, we are making `i18nify-react` required and consumers should add a I18nProvider wrapper with their expected locale set to make sure things don't visually change for existing consumers of Amount. Added more details in comment below

We'll add migration guide to our v11 changelog since it was recently released and people are yet to migrate to it.

.changeset/purple-flies-end.md
tarun-khanna: @saurabhdaware What are the next steps here ? 
I think `@razorpay/i18nify-react` should be just added as mandatory step in migration guide to v11 but shouldn't be added as a mandatory peer dependecy for new users.

.changeset/purple-flies-end.md
tarun-khanna: Agreed, makes sense.

.changeset/purple-flies-end.md
saurabhdaware: > I think @razorpay/i18nify-react should be just added as mandatory step in migration guide to v11 but shouldn't be added as a mandatory peer dependecy for new users.

Yes makes sense

snitin315: > shouldn't be added as a mandatory peer dependecy for new users.

regarding this, should we explicitly mark `@razorpay/i18nify-react` as optional via peerDependenciesMeta?

packages/blade/upgrade-v11.md
```diff
@@ -96,6 +96,47 @@ Only use this if you're unable to run the codemod described above.
...
+
+1. The `<Amount />` component will now automatically format numbers based on the user's browser locale. For example, `<Amount value={123456.789} currency="INR">` will render `₹1,23,456.79` for browsers with the `en-IN` default locale, whereas it will render `₹123,456.79` for browsers with the `en-US` locale.
+
+2. If you want to enable users to change the locale of your page, add the `@razorpay/i18nify-react` package and wrap your app inside the `I18nProvider`. Utilize the `setI18nState` utility to modify the locale. For more details, please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react).
+
+3. Additionally, if you prefer to maintain a fixed locale for your page and amount component, enclose your app within `<I18nProvider initData={{locale: 'locale-you-want'}}>..`. For more details, please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react).
+
+Example with `@razorpay/i18nify-react`
+
+```
```
saurabhdaware, line 109:
```suggestion
```jsx
```

packages/blade/upgrade-v11.md
```diff
@@ -96,6 +96,47 @@ Only use this if you're unable to run the codemod described above.
...
 ### Amount
 
+- **Amount component is now internationalized via [@razorpay/i18nify-js](https://www.npmjs.com/package/@razorpay/i18nify-js).**
+
+1. The `<Amount />` component will now automatically format numbers based on the user's browser locale. For example, `<Amount value={123456.789} currency="INR">` will render `₹1,23,456.79` for browsers with the `en-IN` default locale, whereas it will render `₹123,456.79` for browsers with the `en-US` locale.
+
+2. If you want to enable users to change the locale of your page, add the `@razorpay/i18nify-react` package and wrap your app inside the `I18nProvider`. Utilize the `setI18nState` utility to modify the locale. For more details, please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react).
+
+3. Additionally, if you prefer to maintain a fixed locale for your page and amount component, enclose your app within `<I18nProvider initData={{locale: 'locale-you-want'}}>..`. For more details, please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react).
+
```
snitin315, line 106: Users will come to this section when the codemod fails, so the ideally we need to mention this change under the `Watch Out for Limitations & Edge Cases` section well. You can just add a single point about i18n support and link to this section.

tarun-khanna: done, can you check.

saurabhdaware (dismissed):
Looks good to me

snitin315 (dismissed):

.changeset/purple-flies-end.md
```diff
@@ -0,0 +1,23 @@
...
+
+**What changes ?**
+
+1. The `<Amount />` component will now automatically format numbers based on the user's browser locale. For example, `<Amount value={123456.789} currency="INR">` will render `₹1,23,456.79` for browsers with the `en-IN` default locale, whereas it will render `₹123,456.79` for browsers with the `en-US` locale.
+
+2. If you want to enable users to change the locale of your page, add the `@razorpay/i18nify-react` package and wrap your app inside the `I18nProvider`. Utilize the `setI18nState` utility to modify the locale. For more details, please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react).
+
+3. Additionally, if you prefer to maintain a fixed locale for your page and amount component, enclose your app within `<I18nProvider initData={{locale: 'locale-you-want'}}>..`. For more details, please refer to the [documentation](https://www.npmjs.com/package/@razorpay/i18nify-react).
+
+**How to update ?**
```
kamleshchandnani, line 19: we shall also update the [installation guide](https://blade.razorpay.com/?path=/docs/guides-installation--docs) with installation of i18nify. its located at `packages/blade/docs/guides/Installation.stories.mdx`

kamleshchandnani:
the stories page for amount shall also be updated with the i18nify installation guides and some information.

kamleshchandnani:
@tarun-khanna how would I as a user know that the `currency-indicator` prop is tied to user's browser locale? for eg: my default locale is `en-GB` and when I toggle this property from storybook to `currency-code` and select currency as `USD` it shows me `US$123` which is correct but confusing to me as a dev as I was expecting `$`.
Screenshot 2024-02-09 at 12 51 45 PM

To fixx this we shall do 2 things
* Put notes in the amount output and show the browser's defualt locale in the storybook.
* In the `currencyIndicator` prop we shall add this description to the TS types that this is tied to locale set via i18nify and it defaults to the browser's default(and print the browser's default locale in the table as well within the description)

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -217,3 +223,98 @@ StrikeThrough.args = {
...
+        <SelectInput label="Select locale" />
+        <DropdownOverlay>
+          <ActionList>
+            {localeList.map((item) => (
+              <ActionListItem
+                key={item.locale}
+                title={`${item.country}(${item.locale})`}
+                value={item.locale}
+                onClick={({ name }) => {
+                  setI18nState?.({ locale: name });
```
kamleshchandnani, line 287: @tarun-khanna this is setting the context at a global amount's level ideally we shall scope it within this story itself?

Why?
I opened the amount story page i.e default `Docs` page and saw the currency in some default locale then I navigated to the locale story and was playing around with it and then went back to the Docs page of amount and it was pointing to the locale I set under the locale story which was misleading as there's no information mentioned about that

packages/blade/package.json
```diff
@@ -269,7 +269,9 @@
     "typescript": "4.9",
     "typescript-transform-paths": "3.4.6",
     "@types/body-scroll-lock": "3.1.0",
-    "ramda": "0.29.1"
+    "ramda": "0.29.1",
+    "@razorpay/i18nify-js": "^1.4.0",
```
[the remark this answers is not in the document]
tarun-khanna: No, it not should not install `i18nify-js` automatically if someone install `i18nify-react`.  
As we want to maintain a single copy of `i18nify-js` in consumer apps for state management, `i18nify-js` is added as a peer dependency even for `i18nify-react`. So you will always have to install `i18nify-js` separately if you want `i18nify-react`. 

Here for Blade, we needed `I18nProvider` and associated context state, so I had to install `i18nify-react` as a separate dev dependency but not peer dependency. 
Additionally, the end consumers won't lead to a out of sync issue, as these packages are released always in sync in our mono-repo setup. If a consumer installs a latest copy of `i18nify-react`, they will have to update latest version of `i18nify-js` as well.

A similar discussion happened with Blade team on this [thread](https://github.com/razorpay/blade/pull/2003#discussion_r1482729329) too.

packages/blade/package.json
```diff
@@ -283,7 +285,8 @@
     "react-native-svg": "^12.3.0",
     "react-native-gesture-handler": "^2.9.0",
     "@gorhom/bottom-sheet": "^4.4.6",
-    "@gorhom/portal": "^1.0.14"
+    "@gorhom/portal": "^1.0.14",
+    "@razorpay/i18nify-js": "^1.4.0"
```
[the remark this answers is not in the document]
tarun-khanna: No, the mandatory peer dependency for Blade components to function properly is only `i18nify-js`. For instance, the Amount component relies on the `formatNumber` API provided by `i18nify-js`. 
The addition and use of `i18nify-react` is only necessary if you intend to manage state within your consumer app or maintain a fixed locale on the page.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -1,14 +1,9 @@
...
-import type { AmountTypeProps, Currency } from './amountTokens';
-import {
-  normalAmountSizes,
-  getCurrencyAbbreviations,
-  currencyIndicatorMapping,
-  subtleFontSizes,
-  amountLineHeights,
-  currencyPositionMapping,
-} from './amountTokens';
+import type { CurrencyCodeType } from '@razorpay/i18nify-js/currency';
```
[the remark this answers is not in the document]
tarun-khanna: To use `i18nify-react`, you always have to install `i18nify-js` as it is a core package which maintains state. 
It is added as a peer dependency as we want to keep a single copy of `i18nify-js` in the consumer app.

.changeset/purple-flies-end.md
tarun-khanna: Sure, I was planning to send this in a separate PR, but we can incorporate these changes into this one. 
I'll need your assist to finalize the content and format for inclusion here. @kamleshchandnani

tarun-khanna:
> @tarun-khanna the stories page for amount shall also be updated with the i18nify installation guides and some information. Right the information is outdated on this page
> 
>
@kamleshchandnani Can we connect on this ? Would like to know your thoughts on how to showcase this in the Amount story.

tarun-khanna:
> To fix this we shall do 2 things
> 
> * Put notes in the amount output and show the browser's defualt locale in the storybook.
> * In the `currencyIndicator` prop we shall add this description to the TS types that this is tied to locale set via i18nify and it defaults to the browser's default(and print the browser's default locale in the table as well within the description)

Makes sense @kamleshchandnani , Are you looking to send these changes in this PR itself or can we send this separately ?

packages/blade/src/components/Amount/Amount.stories.tsx
tarun-khanna: Setting a state by I18nProvider's `setI18nState` will always set it globally for `i18nify-js`. We decided to architect our library in this fashion so to deal with out of sync state issues between sidebar, header, body and other elements as discussed in our earlier meetings. 

To deal with this, 
1. We can introduce an internal `locale` prop hidden to consumers.
2. Then on locale dropdown click, We don't have to call `setI18nState` but can maintain an internal locale state which will be passed to the Amount instance used in this story.
3. We can then pass `locale` prop to i18nify-js `formatNumber` api which accepts this locale param for overriding global state. 

Thoughts ? @kamleshchandnani

.changeset/purple-flies-end.md
```diff
@@ -0,0 +1,23 @@
...
+
+1. Install i18nify peer dependency `yarn add @razorpay/i18nify-js`
```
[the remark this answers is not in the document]
tarun-khanna: Sure, this was written from Blade's peer dependency perspective but have updated now.

packages/blade/package.json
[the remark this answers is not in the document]
kamleshchandnani: >If a consumer installs a latest copy of i18nify-react, they will have to update latest version of i18nify-js as well.

this is the issue i see. how will we prevent consumers from updating the one and not the other?

packages/blade/src/components/Amount/Amount.stories.tsx
kamleshchandnani: Okay i'm saying keep it simple from demonstration point of view because right now there's no co-relation to the end consumer that changes in one story is affecting other. what I'm suggesting is wrap another i18n context around just the locale story so the state changes is only affected to that story and not goes all the stories of amount component.

kamleshchandnani:
> > To fix this we shall do 2 things
> > 
> > * Put notes in the amount output and show the browser's defualt locale in the storybook.
> > * In the `currencyIndicator` prop we shall add this description to the TS types that this is tied to locale set via i18nify and it defaults to the browser's default(and print the browser's default locale in the table as well within the description)
> 
> Makes sense @kamleshchandnani , Are you looking to send these changes in this PR itself or can we send this separately ?

yes since everything its related directly to amount component

.changeset/purple-flies-end.md
tarun-khanna: updated.

packages/blade/src/components/Amount/Amount.stories.tsx
tarun-khanna: As discussed earlier, there is no finalised approach here that will resolve this issue so we will be addressing this issue in the next PR. The current one can still go as it is. 
Attaching MOM [here](https://razorpay.slack.com/archives/C05RBSKC6KD/p1707485708148249?thread_ts=1705482465.156459&cid=C05RBSKC6KD)

cc @kamleshchandnani

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -17,13 +17,14 @@ import { ActionList, ActionListItem } from '~components/ActionList';
...
 import { Divider } from '~components/Divider';
+import { Link } from '~components/Link';
 
 const Page = (): React.ReactElement => {
   return (
     <StoryPageWrapper
       componentName="Amount"
       componentDescription="Amounts are used to show small amount of color coded metadata, which are ideal for getting user attention."
-      note="This component only displays the provided value in the specified currency, it does not perform any currency conversion."
+      note="This component only displays the provided value in the specified currency with the formatting capabilities enabled by @razorpay/i18nify-js (https://www.npmjs.com/package/@razorpay/i18nify-js), it does not perform any currency conversion."
```
kamleshchandnani, line 27:
```suggestion
      note="This component only displays the provided value in the specified currency with the formatting capabilities enabled by @razorpay/i18nify-react, it does not perform any currency conversion."
```

packages/blade/src/components/Amount/Amount.stories.tsx
```diff
@@ -37,6 +38,22 @@ const Page = (): React.ReactElement => {
         export default App;
         `}
       </Sandbox>
+      <Box marginBottom="spacing.10">
+        <Text>
+          The Amount component automatically formats numbers based on the user's browser locale
```
kamleshchandnani, line 43: you can use the propname `propsDescription` on the StoryPageWrapper component and put this information there you can remove links and only mention `i18nify-react`. ignore the js since that will go away with the approach we discussed last time

packages/blade/docs/guides/Installation.stories.mdx
```diff
@@ -40,47 +40,62 @@ Before you install the package, make sure that you have performed the following
...
   > **Note**
   >
   > Currently, blade only supports styled-components v5 only
 
-   ```shell
-   yarn add @razorpay/blade styled-components@5.3.11
-   ```
-2. Follow [this guide](#-install-fonts) to install the fonts.
+```shell
+yarn add @razorpay/blade styled-components@5.3.11
```
kamleshchandnani, line 52: move this shell block before styled components and put all the dependencies including i18nify in one shell block itself as we have done in React native section below and then you can use bullt points to describe all the dependencies together

kamleshchandnani (changes requested):

packages/blade/docs/guides/Installation.stories.mdx
tarun-khanna: Done

packages/blade/src/components/Amount/Amount.stories.tsx
tarun-khanna: Done 
Screenshot 2024-02-12 at 3 05 56 PM

tarun-khanna: Done
Screenshot 2024-02-12 at 3 31 47 PM

kamleshchandnani (approved):

PR #2008, opened 2024-02-07
feat: coverage plugin variables upgrade

kamleshchandnani:
## Description

This PR upgrades the coverage plugin to accommodate brand visual upgrades
* Color styles are now moved to variables and the coverage is measured with new color variables
* Typography styles are updated with the new styles
* Recommend using relevant Blade component if Frame/Rectangle is being used as a custom component
* Check if rectangle/box uses `surface/border` and `surface/background` colors respectively
* The coverage is now mandated to 90% as acceptable, anything below this will be marked as red

In addition, this PR also updates token publisher plugin to generate code syntax in dev mode for component tokens

## Changes

image
image

## Additional Information

NA

## Component Checklist

NA


saurav12 (approved):
@kamleshchandnani I checked the plugin by running it on local Figma. It works like a charm ✨

PR #2009, opened 2024-02-08
docs(Drawer): add API decision

saurabhdaware:
### [✨  Pretty Doc ✨ ](https://github.com/razorpay/blade/blob/drawer/api-decision/packages/blade/src/components/Drawer/_decisions/decisions.md)


packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,147 @@
...
+   * Dismiss handler
+   */
+  onDismiss: () => void;
+
+  /**
+   * Show or hide overlay.
+   *
+   * Also decides if clicking outside on overlay closes the drawer or not
+   */
+  showOverlay?: boolean;
```
chaitanyadeorukhkar, line 88: So overlay is optional? Won't that look weird? What is the use case?

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,147 @@
+# Drawer
+
+A drawer is a panel that slides in mostly from right side of the screen over the existing content in the viewport. It helps in providing additional details or context and can also be used to promote product features or new products.
+
+<img width="500px" src="./2024-02-07-20-04-07.png" alt="Drawer Figma Skeleton">
+
+## Design
```
chaitanyadeorukhkar, line 7: Will Drawer be absolute positioned always? What if it needs to be used as a left/right navigation drawer?

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   * Dismiss handler
+   */
+  onDismiss: () => void;
+
+  /**
+   * Show or hide overlay.
+   *
+   * Also decides if clicking outside on overlay closes the drawer or not
```
chaitanyadeorukhkar, line 86: Wont there ever be a possibility where overlay isn't needed but click outside still needs to close the drawer?

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: It will be fixed positioned on top right. It is always positioned like navigation drawer

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: no I did ask this to designers. If someone wants clicking outside to close drawer, they should always have overlay

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,147 @@
...
+};
+```
+
+Other supporting wrapper components for trailing and leading space-
+
+- DrawerHeaderBadge
+- DrawerHeaderIcon
+- DrawerHeaderAsset
+- DrawerHeaderLink
+- DrawerHeaderButton
```
chaitanyadeorukhkar, line 131: We keep creating these components for multiple components... Modal, Bottomsheet, Drawer. We might need to look into changing these into just `HeaderBadge`, `HeaderIcon`, etc. This would keep adding unnecessary repetitive components.

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,147 @@
...
+
+_No alternate APIs were considered because Drawer is closer to Modal on overall meaning and API perspective so made sense to go with API that is closer to Modal, also all DS Drawer components I referenced have similar API_
+
+## Accessibility
+
+- **Aria Attributes:** Drawer will have `aria-modal="true"` and `role="dialog"` and will be treated as modal for voiceover users.
+- **Focus Handling:** Ensure Close Icon is focussable
```
chaitanyadeorukhkar, line 138: Like Modal & Bottomsheet, should we also expose i`nitialFocusRef`?

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,147 @@
+# Drawer
+
+A drawer is a panel that slides in mostly from right side of the screen over the existing content in the viewport. It helps in providing additional details or context and can also be used to promote product features or new products.
+
+<img width="500px" src="./2024-02-07-20-04-07.png" alt="Drawer Figma Skeleton">
+
+## Design
+
+- [Drawer - Figma Design](https://www.figma.com/file/jubmQL9Z8V7881ayUD95ps/Blade-DSL?node-id=78667%3A66663&mode=dev)
```
chaitanyadeorukhkar, line 9: Does the drawer have a fixed width? Will the width change based on the content? If it is fixed width, why do we not have size options?

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,147 @@
...
+</details>
+
+### Props
+
+```ts
+type DrawerProps = {
+  /**
+   * Controlled state of drawer open or not
+   */
+  isOpen: boolean;
```
chaitanyadeorukhkar, line 76: What happens if multiple drawers are opened together? I'm guessing same behaviour as Modal?

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: We have usecases where Drawer is used for Announcements. In those cases, users should still be able to interact with dashboard items even with Drawer open so we cant show overlay or hide drawer on clicking outside.

image

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: Yes similar but only 2 drawers can be kept open at once from design-side. And the 2nd drawer has a back button instead of close button. 

Let me add the stacking behaviour to this API doc I didn't mention it in this

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: It currently has a fixed max-width of 420px and inside that it takes width based on the content.

Let me check with design again once if we want to go with max-width or defined size because in Modal I see its defined size

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: true. `HeaderBadge` is a good idea. Lets discuss this with team once in next standup?

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: Yes I think we should. Drawer will also have similar usecases like Modal and BottomSheet so requirement will anyway come

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: Just checked with RK. It is fixed width of 420px on desktop always. We will only support one size right now

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,147 @@
+# Drawer
+
+A drawer is a panel that slides in mostly from right side of the screen over the existing content in the viewport. It helps in providing additional details or context and can also be used to promote product features or new products.
```
snitin315, line 3:
```suggestion
A drawer is a panel that slides in mostly from the right side of the screen over the existing content in the viewport. It helps in providing additional details or context and can also be used to promote product features or new products.
```

packages/blade/src/components/Drawer/_decisions/decisions.md
snitin315: So we are setting a default position for Drawers on the top-right across all products? Like we set `bottom-left` for toasts?

packages/blade/src/components/Drawer/_decisions/decisions.md
snitin315: > Also decides if clicking outside on overlay closes the drawer or not

I feel we should have a separate prop for deciding this, it shouldn't be dependant on the presence of overlay.

packages/blade/src/components/Drawer/_decisions/decisions.md
snitin315: How are these components different from normal `Badge`/`Icon`? For example, I see `_CardHeaderBadge` is same as `Badge`. Why can't we use the`Badge` component directly? 

https://github.com/razorpay/blade/blob/ba5ba9ba719d94448b0c7c9faae0c35f83d5551a/packages/blade/src/components/Card/CardHeader.tsx#L48

packages/blade/src/components/Drawer/_decisions/decisions.md
snitin315: Add keyboard handling details, e.g, `Esc` should close the drawer.

packages/blade/src/components/Drawer/_decisions/decisions.md
snitin315: when overlay is disabled, does the area outside the drawer remain interactive?

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: yup

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: you mean for a usecase where someone doesn't want to show overlay but still want to close Drawer on clicking outside? I can confirm this once with design if this should be possible. Ideally overlay should be present in most cases except few exceptional cases where outer items have to be interactive

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: These subcomponents usually have defined defaults that are different from defaults of actual component. E.g.

https://github.com/razorpay/blade/blob/ba5ba9ba719d94448b0c7c9faae0c35f83d5551a/packages/blade/src/components/ActionList/ActionListItem.tsx#L214

However in some examples like CardHeaderBadge, the defaults of wrapper are same as defaults of actual component. There it is just for consistency in API otherwise some components will be like `Badge` while some be like `CardHeaderText` etc

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: yes

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: done

packages/blade/src/components/Drawer/_decisions/decisions.md
nashcheez: I would also vote for the separation of the logic to dismiss the overlay items on click outside on a different prop. If Modal and BottomSheet have a prop tomorrow to disable dismiss on click outside on the overlay the same prop should be extended over here.

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,178 @@
...
+   * children node.
+   *
+   * Supports DrawerHeader and DrawerBody
+   */
+  children: React.ReactNode;
+
+  /**
+   * Override z-index of Drawer.
+   *
+   * @default 1002
```
nashcheez, line 105: @saurav12 Can design take up a decision on the alignment of all our default z-indices for our overlay components? Imo it should be Modal < Drawer < BottomSheet < Popover < Tooltip. We have the `zIndex` property to override for each component but by default the system should make a recommendation.

saurav12: Sure, makes sense. We will document these in the Figma itself.

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,178 @@
...
+    trailing={
+      <>
+        <Link />
+        <Button />
+      </>
+    }
+  />
+  <DrawerBody>
+    <Slot />
+  </DrawerBody>
```
snitin315, line 33: No Footer?

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,178 @@
...
+   * Show or hide overlay.
+   *
+   * Also decides if clicking outside on overlay closes the drawer or not
+   */
+  showOverlay?: boolean;
+
+  /**
+   * Initial focus reference element
+   */
+  initialFocusRef?: React.MutableRefObject<any>;
```
snitin315, line 93: What is the use-case of this prop?

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,178 @@
...
+- DrawerHeaderBadge
+- DrawerHeaderIcon
+- DrawerHeaderAsset
+
+_No alternate APIs were considered because Drawer is closer to Modal on overall meaning and API perspective so made sense to go with API that is closer to Modal, also all DS Drawer components I referenced have similar API_
+
+## Drawer Stacking
+
+- Only 2 Drawers can be stacked on top of each other
+- 2nd Drawer always has overlay independent of `showOverlay` prop
```
snitin315, line 160: So, the `showOverlay` prop is redundant for the 2nd drawer? Why so? If I don't want to show overlay in 1st drawer because important information is present on the left then that should hold for the 2nd drawer as well.

So I feel, if we are giving control of showing overlay it should be either on both drawers or none. WDYT?

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,178 @@
...
+- 2nd Drawer always has back button instead of close button. Clicking on back button closes the 2nd drawer.
+
+<img width="500px" src="./2024-02-12-11-01-32.png" />
```
snitin315, line 163: Can you add a code example for stacked drawers?

packages/blade/src/components/Drawer/_decisions/decisions.md
chaitanyadeorukhkar: There are use-cases where drawer is used for navigation on left side in a lot of apps. Won't that ever be a use-case for us?

chaitanyadeorukhkar: Example: 
image

packages/blade/src/components/Drawer/_decisions/decisions.md
chaitanyadeorukhkar: > fixed width of 420px
Max width would be 420px right?

packages/blade/src/components/Drawer/_decisions/decisions.md
chaitanyadeorukhkar: Would we need a separate callback for the back button?

packages/blade/src/components/Drawer/_decisions/decisions.md
chaitanyadeorukhkar: Agreed. Better to have a separate prop since we might be adding the same to Modal & BottomSheet too

packages/blade/src/components/Drawer/_decisions/decisions.md
chaitanyadeorukhkar: > does the area outside the drawer remain interactive

We should ideally control this with a new prop like `shouldDismissOnClickOutside`

packages/blade/src/components/Drawer/_decisions/decisions.md
chaitanyadeorukhkar: +1

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,178 @@
...
+- DrawerHeaderIcon
+- DrawerHeaderAsset
+
+_No alternate APIs were considered because Drawer is closer to Modal on overall meaning and API perspective so made sense to go with API that is closer to Modal, also all DS Drawer components I referenced have similar API_
+
+## Drawer Stacking
+
+- Only 2 Drawers can be stacked on top of each other
+- 2nd Drawer always has overlay independent of `showOverlay` prop
+- 2nd Drawer always has back button instead of close button. Clicking on back button closes the 2nd drawer.
```
chaitanyadeorukhkar, line 161: Why though? a back button only makes sense when the navigation is _within_ the drawer and not another drawer on top of it.

If we are doing this, then we should also have a back button on 2nd Modal? Same logic would be applicable there. I think we shouldn't have a back button in either cases.

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: > There are use-cases where drawer is used for navigation on left side in a lot of apps. Won't that ever be a use-case for us?

Nope. So for desktop, our navigation bar itself is on left so there won't be anything opening from that side.

On mobile too we have drawers opening from right side only. Plus navigation component is something that we might tackle separately. This navigation component will use our Drawer internally on mobile. So at that time we can relook at this.

Anyway even if usecase comes up, its just addition of new prop so not much effort and no breaking change

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: Modal and BottomSheet would anyway have different prop than Drawer in this particular case no?

E.g. in Modal and BottomSheet, we always want to show overlay so there will never be a `showOverlay` prop. Whereas in Drawer, we definitely need to have `showOverlay` prop. 

If we're having that prop and we know interactions outside have to be consistent with display of overlay, wouldn't adding new prop be extra overhead for consumers here?

Because if we add 2 props, it creates impossible states of 
```jsx
<>
  <Drawer
    showOverlay={true}
    shouldDismissOnOutsideClick={false}
  />
  <Drawer
    showOverlay={false}
    shouldDismissOnOutsideClick={true}
  />
</>
``` 

These scenarios should throw error in 2 props case because its not correct usage from UX perspective

saurabhdaware: Oh anyway just discussed this with design - https://razorpay.slack.com/archives/G01B3LQ9H0W/p1707808893726369

Usecase that we have on Modal is something that we would have it here as well. I'll make changes to add 2 props.

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: nope. Drawer is always 420px irrespective of the content.

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: back button is just dismiss of 2nd Drawer. Consumer knows which is their second drawer so they can use that `onDismiss` itself.

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: nope

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: If they want to keep initial focus to be on some particular button inside of Drawer. By default focus goes to close button.

We have same prop on Modal and BottomSheet

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: > So, the showOverlay prop is redundant for the 2nd drawer? Why so? If I don't want to show overlay in 1st drawer because important information is present on the left then that should hold for the 2nd drawer as well.

For example in X, 1st Drawer is used for showing information based on the interaction on left side. But 2nd Drawer does not have that usecase.

Plus overlay on 2nd Drawer is also there to denote the stacking. If there's 2 Drawers open without overlay, there is no direct way to tell user that there is another drawer behind this drawer.

Also in most cases, Drawer should always have overlay. There are just few exceptions for 1st Drawer which is why we need to provide overlay prop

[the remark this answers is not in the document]
saurabhdaware: > Won't this be counterintuitive?

True but don't see any other way to implement this. We do want to show overlay on 2nd drawer.

https://github.com/razorpay/blade/pull/2009#discussion_r1487372260

> How will we communicate this to our devs? 

We can throw error. Also its enforced from design side so its not possible to implement something like this on design as the UI itself is slightly different for 2nd drawer.

packages/blade/src/components/Drawer/_decisions/decisions.md
saurabhdaware: In case of Modal, the size of the modal would also let people know about modal behind it no? If we put close button on 2nd Drawer, because all Drawers are same size and similarly placed, it feels like clicking close will close everything. But instead we only go back to the first Drawer in the stack.

Basically we would need something to let consumer know that this Drawer is stacked. Back button and overlay was our way to communicate that. But we can discuss this once with RK maybe?

saurabhdaware: Started discussion here - https://razorpay.slack.com/archives/G01B3LQ9H0W/p1707813910309619?thread_ts=1707803235.144969&cid=G01B3LQ9H0W

Adding it to open questions

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,174 @@
...
+   * Dismiss handler
+   */
+  onDismiss: () => void;
+
+  /**
+   * Show or hide overlay.
+   *
+   * Also decides if clicking outside on overlay closes the drawer or not
+   */
+  showOverlay?: boolean;
```
Rozamo, line 85: Should it be `shouldShowOverlay`?

packages/blade/src/components/Drawer/_decisions/decisions.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Proposed API
+
+```jsx
+<Drawer
+  isOpen={}
+  onDismiss={() => {}}
+  showOverlay
+>
+  <DrawerHeader
```
Rozamo, line 19: Will this be fixed while scrolling?

saurabhdaware: Good question. Header scrolls right now. Let me confirm with designer once.

saurabhdaware: Confirmed. It should be fixed on top similar to how we have on Modal

chaitanyadeorukhkar (approved):

snitin315 (approved):
:shipit:

PR #2010, opened 2024-02-08
docs(FileUpload): add API decisions

snitin315:
### [Pretty Markdown Doc](https://github.com/razorpay/blade/blob/docs/file-upload-api-descisions/packages/blade/src/components/FileUpload/_decisions/decisions.md)

This PR addresses the decisions made for the FileUpload component API.
  - **API Specification:** Detailed the structure, functionality, and visual representation of the FileUpload component API.
  - **Examples:** Provided illustrative examples showcasing various use cases, including controlled and uncontrolled behaviors, single and multiple file selections.


packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,373 @@
...
+   * The percentage of file upload completion.
+   */
+  percent?: number;
+  /**
+   * Text indicating an error state
+   */
+  errorText?: string;
+};
+
+type FileList = File[];
```
anuraghazra, line 68: Name FileList, File something different. Otherwise they will conflict with the global FileLIst interfaces/APIs 

https://developer.mozilla.org/en-US/docs/Web/API/FileList

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,373 @@
...
+   */
+  onRemove?: ({ removedFile, fileList }: { removedFile: File; fileList: FileList }) => void;
+  /**
+   * Callback function executed when files are dropped into the upload area
+   */
+  onDrop?: (event: React.DragEvent) => void;
+  /**
+   * State indicating whether there is an error in the FileUpload component
+   */
+  validationState?: 'none' | 'error';
```
anuraghazra, line 136: I can see that individual File has a `status` too, what happens when 1 file is set to error but the Root validationState prop is set to `none`?

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: Yeah, Will name it `BladeFile` & `BladeFileList`.

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,373 @@
...
+
+#### Single File selection
+
+```jsx
+import React, { useState } from 'react';
+import axios from 'axios';
+import { Box, FileUpload } from '@razorpay/blade/components';
+
+const UncontrolledSingleFileUploadForm = () => {
+  const [file, setFile] = useState();
```
anuraghazra, line 174: How is this uncontrolled when I'm actually managing the `file` state? 

Ideally we should not have this and only have handleSubmit? 

```
handleSubmit = (e) => {
  const files = e.target.files;
  // do something
}
```

snitin315: We aren't passing it to the component, we don't need this to attach/select/drop the files. Users need this information when they upload files.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: Actually it would be 

```
interface BladeFile extends File {
  /**
   * The unique identifier of the file.
   */
  id: string;
  /**
   * The file's upload status.
   */
  status?: 'uploading' | 'success' | 'error';
  /**
   * The percentage of file upload completion.
   */
  percent?: number;
  /**
   * Text indicating an error state
   */
  errorText?: string;
}
```

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: Maybe I can update the example using ref

snitin315: Updated.

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,359 @@
...
+   */
+  onRemove?: (removedFile: BladeFile) => void;
+  /**
+   * Callback function executed when files are dropped into the upload area
+   */
+  onDrop?: (event: React.DragEvent) => void;
+  /**
+   * State indicating whether there is an error in the FileUpload component
+   */
+  validationState?: 'none' | 'error';
```
saurabhdaware, line 125: do we not have success state like we have in other inputs?

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,359 @@
...
+    } catch (error) {
+      // Handle errors
+      console.error('File upload failed:', error.message);
+    }
+  };
+
+  return (
+    <Box>
+      <form onSubmit={handleSubmit}>
+        <FileUpload
```
saurabhdaware, line 194: This file list is always rendered below input? any usecase where it needs to be rendered at different place?
image

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Direct Upload with Progress Display (Controlled)
+
+https://github.com/razorpay/blade/assets/46647141/cb690625-011b-4c20-b610-9af5f0c5e077
+
+## Examples
+
+Here are a few illustrative examples showcasing the utilization of the `FileUpload` component:
+
+### Uncontrolled Usage
```
saurabhdaware, line 153: if consumer wants to show progress, they have to go with controlled usage?

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,359 @@
...
+   */
+  maxSize?: number;
+  /**
+   * Callback function triggered when files are selected
+   */
+  onChange?: ({ name, fileList }: { name: string; fileList: BladeFileList }) => void;
+  /**
+   * Callback function triggered when the preview icon is clicked
+   */
+  onPreview?: (previewedFile: BladeFile) => void;
```
saurabhdaware, line 113: what happens when preview is clicked? like why is it clickable?

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: Yes.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: @saurabhdaware For example, you uploaded a file and now you want to check if you have uploaded/selected the correct file you can click on the preview button to open & see it.

For example, onPreview can be used in the following after the image has been uploaded:

https://github.com/razorpay/blade/assets/46647141/31162e42-5347-4468-b493-5187d85dcacf


There are two ways how we can approach it:

- Don't show the preview icon by default, show it only when `onPreview` is provided. 
- Always show the preview icon by providing a default behavior of opening the file in a new tab which is overridden by the `onPreview` prop.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: Not as of now, if we give users control over this we won't be able to standardize it.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: No, the success state is indicated by the attached FileItem. Because if it's successful items would have been successfully attached / uploaded.

Screenshot 2024-02-13 at 4 36 27 PM

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: I can add this as an open question too

snitin315: Thinking more on this, we can't predict the default behavior it could be different for different file types. So, I'm inclined on the 1st behavior mentioned above.

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,398 @@
...
+  /**
+   * Position of the label relative to the file upload area. Desktop only prop. Default value on mobile will be 'top'
+   *
+   * @default 'top'
+   */
+  labelPosition?: 'top' | 'left';
+  /**
+   * Defines the selection behavior of the FileUpload component
+   */
+  selectionType?: 'single' | 'multiple';
```
Rozamo, line 69: Is there a scope for more types? Can we convert it to a boolean prop? For example `isMultiple`.

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,398 @@
...
+   */
+  defaultFileList?: BladeFileList;
+  /**
+   * List of files that have been selected/uploaded, useful when the component is controlled
+   */
+  fileList?: BladeFileList;
+  /**
+   * Limit the number of files that can be uploaded
+   */
+  maxCount?: number;
```
Rozamo, line 105: Can we use only this to toggle between single and multiple file selection?

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,398 @@
...
+  /**
+   * Limit the size of the uploaded files (in bytes)
+   */
+  maxSize?: number;
```
Rozamo, line 109: For multiple files does it limit total size or per file?

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,398 @@
...
+   */
+  validationState?: 'none' | 'error';
+  /**
+   * Additional text providing assistance or guidance
+   */
+  helpText?: string;
+  /**
+   * Text indicating an error state
+   */
+  errorText?: string;
```
Rozamo, line 137: When will this error text be used? Seeing that there is the same prop in `BladeFile`.

snitin315: Explained in https://github.com/razorpay/blade/pull/2010#discussion_r1485802508

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: This is per file, I'll update the description to make it clear.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: That is not an ideal behavior for a component. Also, this prop will take effect only when `selectionType="multiple"`.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: Keeping it consistent with other component's API. [Ref. ChipGroup's API](https://blade.razorpay.com/?path=/docs/components-chip-chipgroup--docs)

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,398 @@
...
+   */
+  id: string;
+  /**
+   * The file's upload status.
+   */
+  status?: 'uploading' | 'success' | 'error';
+  /**
+   * The percentage of file upload completion.
+   */
+  percent?: number;
```
chaitanyadeorukhkar, line 50: Instead of `percent` should we use `progressValue`, `progressMin`, & `progressMax`? You will need to pass these to Progress bar component anyway and the way we accept "progress" related props can be similar here as well as in Progress bar

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,398 @@
...
+  /**
+   * Text indicating an error state
+   */
+  errorText?: string;
+}
+
+type BladeFileList = BladeFile[];
+
+type FileUploadCommonProps = {
```
chaitanyadeorukhkar, line 59: What callback would the user get when an upload is cancelled midway?

packages/blade/src/components/FileUpload/_decisions/decisions.md
chaitanyadeorukhkar, line 69: We are not "selecting" anything here. Rename this to something more contextual 
```suggestion
  uploadType?: 'single' | 'multiple';
```

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,398 @@
...
+  /**
+   * Defines the selection behavior of the FileUpload component
+   */
+  selectionType?: 'single' | 'multiple';
+  /**
+   * File types that can be accepted. See [input's accept attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept)
+   *
+   * Usage: accept=".jpg, .png, .pdf", accept="image/*", accept="image/png, image/jpeg, application/pdf"
+   */
+  accept?: string;
```
chaitanyadeorukhkar, line 75: Why is this string? Can we use a proper TS type here?

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,398 @@
...
+   */
+  necessityIndicator?: 'required' | 'optional' | 'none';
+  /**
+   * The name of the file upload input, [useful in form submissions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name)
+   */
+  name?: string;
+  /**
+   * Default list of files that have been uploaded, useful when the component is uncontrolled
+   */
+  defaultFileList?: BladeFileList;
```
chaitanyadeorukhkar, line 97: Do we really need this? When will there be a case where we load a default set of list?

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,398 @@
...
+   */
+  maxSize?: number;
+  /**
+   * Callback function triggered when files are selected
+   */
+  onChange?: ({ name, fileList }: { name?: string; fileList: BladeFileList }) => void;
+  /**
+   * Callback function triggered when the preview icon is clicked
+   */
+  onPreview?: ({ previewedFile }: { previewedFile: File }) => void;
```
chaitanyadeorukhkar, line 117: Why do we need to call it previewedFile?
```suggestion
  onPreview?: ({ file }: { file: File }) => void;
```

packages/blade/src/components/FileUpload/_decisions/decisions.md
```diff
@@ -0,0 +1,398 @@
...
+  /**
+   * Callback function triggered when a file is removed
+   */
+  onRemove?: ({ removedFile }: { removedFile: File }) => void;
```
chaitanyadeorukhkar, line 121:
```suggestion
  onRemove?: ({ file }: { file: File }) => void;
```

packages/blade/src/components/FileUpload/_decisions/decisions.md
chaitanyadeorukhkar, line 50: We might not always have the percent, we might need to rely on `indeterminate` state as well. Lets add `isProgressIndeterminate` as well

chaitanyadeorukhkar (changes requested):

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: It's [an HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept) that accepts a string. We can't use any other type.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: This could be useful if a form is populating data from local storage. But I think it can be achieved via `fileList` too. We can add this prop later if any use case come.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: We can call it `file`.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: So we actually don't need `isProgressIndeterminate ` prop, we can drive it  from `status` & `percent`.

https://github.com/razorpay/blade/blob/1bca91e6a78c5df6bda15c3044f3042002dbfc7a/packages/blade/src/components/FileUpload/FileUploadItem.tsx#L98-L100

snitin315: I don't think we need `progressMin` & `progressMax` it would always be 0 & 100 for a file being uploaded. We can rename `percent` to `uploadPercent` maybe.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: As per the current API, `onRemove` will be triggered when `CloseIcon` or `TrashIcon` is clicked. But users may want to use different logic so we should introduce a new`onCancel` prop. Otherwise they would need to rely on `file.status` which may not be ideal.

Screenshot 2024-02-20 at 2 44 56 PM

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: Discussed with Anurag, and removed this prop.

packages/blade/src/components/FileUpload/_decisions/decisions.md
chaitanyadeorukhkar: Yeah. Either call it `uploadPercent` or `ProgressPercent`

packages/blade/src/components/FileUpload/_decisions/decisions.md
chaitanyadeorukhkar: `onRemove` theek hai na?

So upload can get cancelled 2 ways
1. User cancels it - we call `onRemove` because user can cancel it only by clicking the X button
2. Internet/Server Failure - what will we call here?

packages/blade/src/components/FileUpload/_decisions/decisions.md
chaitanyadeorukhkar: You need to do `isIndeterminate={isUndefined(percent)}`

percent will be 0 or 1 as valid values and at that time Boolean(0) will be false

snitin315: correct.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: >onRemove theek hai na?

So we trigger `onRemove` when either CloseIcon Or TrashIcon is clicked, right?

snitin315: Thinking more no this, I feel we should have 2 different props `onRemove` & `onCancel` because users may want different behavior when the file is uploading or fails to upload and when it has been successfully uploaded.

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: Updated

packages/blade/src/components/FileUpload/_decisions/decisions.md
saurabhdaware: Wouldn't server failures be handled on user's end wherever they are calling the API? How will we call that onCancel?

```js
try {
  // fetch api call
} catch (err) {
 // this is server failure
}
```

snitin315:
```js
try {
  // fetch api call
} catch (err) {
 // file.status = "error"
// file.errorText = error.message
}
```

After this the file is displayed in error state:

Screenshot 2024-02-27 at 11 36 05 AM

Now clicking the close icon triggers `onCancel`. Also trigger when clicking the close icon when uploading files:
Screenshot 2024-02-27 at 11 37 30 AM

@saurabhdaware

saurabhdaware: no I mean is it us who call API internally in our component? or do consumers call it?

snitin315: >no I mean is it us who call API internally in our component? or do consumers call it?

@saurabhdaware Consumers have control of the submit logic, check examples I've added in this doc for more info

packages/blade/src/components/FileUpload/_decisions/decisions.md
chaitanyadeorukhkar, line 50: This doesn't seem updated here. I thought we were calling it `uploadPercent`?

chaitanyadeorukhkar (approved):

packages/blade/src/components/FileUpload/_decisions/decisions.md
snitin315: Yes, forgot to push changes.

snitin315: Fixed in https://github.com/razorpay/blade/pull/2010/commits/ee2f27b58c04157db4a60ff89223bac28a2772cd

chaitanyadeorukhkar (approved):

packages/blade/src/components/FileUpload/_decisions/decisions.md
saurabhdaware: so clicking on close icon always calls `onCancel` right? should we call it `onCloseButtonClick`, `onDismiss` in that case similar to our other components?

Because to me, cancel implies that uploading got cancelled but in case where we show error and show close icon, we're not exactly cancelling the upload request no? upload request just failed the validation and the error was dismissed by consumer 🤔  
image

snitin315: Yeah, `onDismiss` is more suitable.

saurabhdaware (approved):

PR #2014, opened 2024-02-13
fix: i18nify-js local dependency added in sandboxes

tarun-khanna:
## Description
As `@razorpay/i18nify-js` dependency was not added in sandbox environment, the sandbox was breaking with the error "Could not find @razorpay/i18nify-js dependency". 
Have added a fix for this issue in this PR.

Slack thread => 
https://razorpay.slack.com/archives/C01H13RTF8V/p1707796443539579 

Issue =>
image

Fix => 
Screenshot 2024-02-13 at 7 34 42 PM

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/.changeset/three-humans-divide.md
+++ b/.changeset/three-humans-divide.md
@@ -1,5 +1,0 @@
----
-"@razorpay/blade": patch
----
-
-fix: i18nify-js local dependency added in sandboxes
--- a/.changeset/three-humans-divide.md
+++ b/.changeset/three-humans-divide.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix: i18nify-js local dependency added in sandboxes
--- a/packages/blade/src/utils/storybook/Sandbox/baseCode.ts
+++ b/packages/blade/src/utils/storybook/Sandbox/baseCode.ts
@@ -59,6 +59,7 @@ export const getReactScriptsJSDependencies = (): Dependencies => {
       'styled-components': packageJson.peerDependencies['styled-components'],
       '@emotion/react': '11.11.1',
       '@table-library/react-table-library': '4.1.7',
+      '@razorpay/i18nify-js': '1.4.0',
     },
   };
 };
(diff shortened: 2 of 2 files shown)


saurabhdaware (dismissed):
Looks good to me. Thanks!

packages/blade/src/utils/storybook/Sandbox/baseCode.ts
```diff
@@ -74,6 +75,7 @@ const getViteReactTSDependencies = (): Dependencies => {
       'styled-components': packageJson.peerDependencies['styled-components'],
       '@emotion/react': '11.11.1',
       '@table-library/react-table-library': '4.1.7',
+      '@razorpay/i18nify-js': '1.4.0',
```
kamleshchandnani, line 78: we need `@razorpay/i18nify-react` too right?

.changeset/three-humans-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
```
kamleshchandnani, line 2: we don't need to do release for this since this is a fix in documentation and that gets published for all master merged. so can you remove this changeset?

saurabhdaware: removed it

packages/blade/src/utils/storybook/Sandbox/baseCode.ts
saurabhdaware: Can we separate it into different PR for `@razorpay/i18nify-react` along with its provider and usage in Amount component example? 

Examples are breaking currently so we can merge this first to unblock consumers

tarun-khanna: `@razorpay/i18nify-react` is not required for Blade to work right now. 
It's just an optional peer dependency for state management. 

If we decide to adopt our discussed approach, only then it will be required. @kamleshchandnani

kamleshchandnani (approved):

PR #2017, opened 2024-02-15
fix(Dropdown): remove scrollbar on tag slot and clicks on icon

saurabhdaware:
## Description
Fixes 2 issues -
1. Hides scrollbar from tag slot ([reference - Internal Slack](https://razorpay.slack.com/archives/CMQ3RBHEU/p1707983918459799?thread_ts=1707983890.941129&cid=CMQ3RBHEU))
2. Click on DropdownLink and DropdownButton icon was not working which is fixed ([referrence](https://twitter.com/championswimmer/status/1758053933799690424?s=46&t=z3Gl4H0P-0tu0p7yvK1HbQ))

## Changes

- The clicks on icon of DropdownLink and DropdownButton were not working because when we click these icons, we trigger `setIsOpen(true)` and switch the icon from chevron down to chevron up. The icon that is clicked now no longer exists because of switch in icons in React DOM and DOM so our `dropdown.contains(target)` check (which checks if outside item is clicked was getting false)
- This is fixed by adding `document.body.contains` check. Which checks if item clicked exists in the DOM or not before marking it as outside click

## Additional Information

https://github.com/razorpay/blade/assets/30949385/359fa2fd-15c7-434b-a1d3-a869b69b7f7c

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Dropdown/Dropdown.tsx
+++ b/packages/blade/src/components/Dropdown/Dropdown.tsx
@@ -227,7 +227,10 @@ const _Dropdown = ({
           return;
         }
 
-        const isOutsideClick = !dropdown.contains(target) && !isTagDismissedRef.current?.value;
+        const isOutsideClick =
+          !dropdown.contains(target) &&
+          !isTagDismissedRef.current?.value &&
+          document.body.contains(target);
 
         const isDropdownOpenState = isDropdownOpenRef.current;
         if (isOutsideClick && isDropdownOpenState) {
--- a/.changeset/tiny-cooks-relate.md
+++ b/.changeset/tiny-cooks-relate.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fix(Dropdown): remove scrollbar on tag slot and allow clicks on DropdownLink and DropdownButton chevron icons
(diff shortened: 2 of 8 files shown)


packages/blade/src/components/Dropdown/Dropdown.tsx
```diff
@@ -227,7 +227,10 @@ const _Dropdown = ({
           return;
         }
 
-        const isOutsideClick = !dropdown.contains(target) && !isTagDismissedRef.current?.value;
+        const isOutsideClick =
+          !dropdown.contains(target) &&
+          !isTagDismissedRef.current?.value &&
+          document.body.contains(target);
```
saurabhdaware, line 233: isOutsideClick was getting true when icon was clicked because when you click ChevronDown icon, it opens dropdown and switches to ChevronUp icon. So ChevronDown icon is the `target` here which did not exist in DOM after click so `dropdown.contains(target)` was returning true.

Now we also check if target exists in DOM with `document.body.contains(target)`

anuraghazra: How was this working before?

anuraghazra (approved):

packages/blade/src/components/Dropdown/Dropdown.tsx
saurabhdaware: No clue lmao. I think the sequence of state setting changed after `useControllableState` implementation. Earlier it must've been calling set state twice but maybe in this sequence 
- Close dropdown
- Open dropdown

But now it seems to call
- Open dropdown
- Close dropdown

so its closing the dropdown immediately after open when icon of DropdownLink is clicked

kamleshchandnani (approved):

PR #2028, opened 2024-02-20
feat(blade): add `Breadcrumb` component

anuraghazra:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Breadcrumb/types.ts
```diff
@@ -0,0 +1,64 @@
...
+   * aria-label for breadcrumb
+   */
+  accessibilityLabel?: string;
+};
+
+type BreadcrumbItemProps = {
+  /**
+   * Href of the BreadcrumbItem
+   */
+  href: string;
```
saurabhdaware, line 36: Should we also add `target`?

packages/blade/src/components/Breadcrumb/Breadcrumb.web.tsx
```diff
@@ -0,0 +1,75 @@
...
+      color={color === 'white' ? 'surface.text.staticWhite.muted' : 'surface.text.gray.muted'}
+      size={size}
+      weight="medium"
+    >
+      /
+    </Text>
+  );
+};
+
+const listStyle = { listStyle: 'none' };
```
saurabhdaware, line 25:
```suggestion
const listStyleNone = { listStyle: 'none' };
```

nit

packages/blade/src/components/Breadcrumb/Breadcrumb.web.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+const Breadcrumb = ({
+  size = 'medium',
+  color = 'primary',
+  showLastSeparator = false,
+  accessibilityLabel = 'Breadcrumb',
```
saurabhdaware, line 31: we shouldn't have default label no? text will be announced by default I believe

packages/blade/src/components/Breadcrumb/Breadcrumb.web.tsx
```diff
@@ -0,0 +1,75 @@
...
+              key={index}
+              as="li"
+              display="flex"
+              alignItems="center"
+              gap="spacing.3"
+              {...makeAccessible({
+                current: (child as React.ReactElement)?.props?.isCurrentPage ? 'page' : undefined,
+              })}
+            >
+              {React.cloneElement(child as React.ReactElement, { _size: size, _color: color })}
```
saurabhdaware, line 60: Why not use context? this feels hacky no?

packages/blade/src/components/Breadcrumb/types.ts
anuraghazra: I don't think that should be needed. 

A breadcrumb shows a user journey, so it's unlikely that you will by default want to open a breadcrumb link on a new tab, if user needs to open a link on a new tab they can `ctrl + click` the link

packages/blade/src/components/Breadcrumb/Breadcrumb.web.tsx
anuraghazra: We should have default. 

A page should ideally have 1 breadcrumb only, thus it should be announced as "Breadcrumb" - This pattern is set on [APG](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/) too.

packages/blade/src/components/Breadcrumb/Breadcrumb.web.tsx
anuraghazra: yyyeeahh could've used Context, too lazy to write the Context wala code this was easier since there's just 1 component and 1 level deep 😆

packages/blade/src/components/Breadcrumb/Breadcrumb.web.tsx
```diff
@@ -0,0 +1,82 @@
...
+  color = 'primary',
+  showLastSeparator = false,
+  accessibilityLabel = 'Breadcrumb',
+  children,
+  ...styledProps
+}: BreadcrumbProps): React.ReactElement => {
+  const contextValue = React.useMemo(() => ({ size, color }), [size, color]);
+
+  return (
+    <BaseBox as="nav" {...getStyledProps(styledProps)}>
```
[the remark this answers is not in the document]
anuraghazra: Added

chaitanyadeorukhkar:
@anuraghazra Keep home icon only on the first item in this story? This looks weird
Screenshot 2024-02-23 at 3 04 14 PM

packages/blade/src/components/Breadcrumb/Breadcrumb.web.tsx
```diff
@@ -0,0 +1,88 @@
...
+};
+
+const listStyleNone = { listStyle: 'none' };
+
+const Breadcrumb = ({
+  size = 'medium',
+  color = 'primary',
+  showLastSeparator = false,
+  accessibilityLabel = 'Breadcrumb',
+  children,
```
chaitanyadeorukhkar, line 34: add testID also?

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
```diff
@@ -78,6 +78,7 @@ type BaseLinkCommonProps = {
    * The title of the link which is displayed as a tooltip. This is a web only prop and has no effect on react-native.
    */
   htmlTitle?: string;
+  opacity?: number;
```
chaitanyadeorukhkar, line 81: Why are changing the opacity? Can't we use a gray color that is lighter instead? Our color already has an alpha color and on top of that we're changing it more. This just means that our colors aren't sufficing our needs if we need to use opacity?

packages/blade/src/components/Breadcrumb/Breadcrumb.web.tsx
anuraghazra: 🤔 not needed, people can do getByRole(nav)

packages/blade/src/components/Breadcrumb/BreadcrumbItem.web.tsx
```diff
@@ -0,0 +1,54 @@
...
+          color={isWhite ? 'surface.text.staticWhite.normal' : 'surface.text.gray.normal'}
+        >
+          {children}
+        </Text>
+      </BaseBox>
+    );
+  }
+
+  return (
+    <BaseLink
```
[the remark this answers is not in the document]
anuraghazra: href will already be passed by consumer, rel is not needed. 

rel is only needed if we had target, to add `rel="noreferrer noopener"`

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
anuraghazra: Added same comments on figma, to be discussed with RK on monday.

packages/blade/src/components/Breadcrumb/types.ts
```diff
@@ -0,0 +1,64 @@
...
+   * Whether the BreadcrumbItem is the current page,
+   * Sets the aria-current attribute to `page`
+   *
+   * @default false
+   */
+  isCurrentPage?: boolean;
+  /**
+   * Content of the BreadcrumbItem
+   */
+  children?: string;
```
saurabhdaware, line 53: we should use that ReactText type here no that we use on other components? This will break
```jsx
<BreadcrumpItem>Home {someVariable}</BreadcrumpItem>
```

anuraghazra: done

packages/blade/src/components/Breadcrumb/Breadcrumb.web.tsx
snitin315: does that work with cypress/playwright too?

packages/blade/src/components/Link/BaseLink/BaseLink.tsx
chaitanyadeorukhkar: Is this discussion done? Can we move away from opacity?

packages/blade/src/components/Breadcrumb/Breadcrumb.web.tsx
anuraghazra: yes, https://playwright.dev/docs/locators

saurabhdaware (dismissed):

.changeset/five-cows-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": minor
+---
+
+feat(blade): breadcrumb web implementation
```
chaitanyadeorukhkar, line 5:
```suggestion
feat(blade): add Breadcrumb component
```

chaitanyadeorukhkar (approved):

PR #2061, opened 2024-03-04
docs(ButtonGroup): add API decisions

snitin315:
### [Pretty Markdown Doc](https://github.com/razorpay/blade/blob/docs/button-group-api/packages/blade/src/components/ButtonGroup/_decisions/decisions.md)

This PR addresses the decisions made for the ButtonGroup component API.
  -  Detailed the structure, functionality, and visual representation of the ButtonGroup component API.
  -  Provided illustrative examples showcasing various use cases
  - [FC-2709](https://razorpay.atlassian.net/browse/FC-2709)

[FC-2709]: https://razorpay.atlassian.net/browse/FC-2709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


anuraghazra (approved):

packages/blade/src/components/ButtonGroup/_decisions/decisions.md
```diff
@@ -0,0 +1,138 @@
...
+      </ButtonGroup>
+    );
+  };
+  ```
+
+- Buttons inside a `ButtonGroup` can be used to trigger a dropdown.
+
+  > [!IMPORTANT]
+  >
+  > We will introduce a new `placement?:| "bottom-end" | "bottom-start" | "left-end" | "left-start" | "right-end" | "right-start" | "top-end" | "top-start"` prop in the `DropdownOverlay` component to control the placement of the dropdown.
```
chaitanyadeorukhkar, line 91: Can't this be auto handled similar to how dropdown opens up right now depending on available space?

// @saurabhdaware

snitin315: @chaitanyadeorukhkar We discussed this yesterday in office, It's auto handled when there is no space in left/right/top/bottom. But in the case of a button group we want to open it in a certain place even if there's space available.

Screenshot 2024-03-06 at 10 56 20 AM

packages/blade/src/components/ButtonGroup/_decisions/decisions.md
```diff
@@ -0,0 +1,138 @@
...
+    ActionList,
+    ActionListItem,
+    ChevronDownIcon,
+  } from '@razorpay/blade/components';
+
+  const App = () => {
+    return (
+      <ButtonGroup>
+        <Button icon={PlusIcon}>Payout</Button>
+        <Dropdown>
```
chaitanyadeorukhkar, line 111: So button group will allow Button & Dropdown components? What if someone uses  some weird dropdown trigger like SelectInput? Are we also adding a check that will restrict only using `DropdownButton` inside `Dropdown`?

snitin315: >So button group will allow Button & Dropdown components? What if someone uses some weird dropdown trigger like SelectInput?

Consumers control the children's content, however, they won't pass any random thing unless it comes from the design. 

>Are we also adding a check that will restrict only using DropdownButton inside Dropdown?

We can add checks, but it may increase the complexity because `DropdownButton`  is nested inside `Dropdown`. I'll need to explore the feasibility.

snitin315: We can add the check it would look something like below:
```js
      if (__DEV__) {
            // throw error if child is not a button or dropdown with button trigger
            if (
              child?.type?.componentId !== 'Button' &&
              !(
                child?.type?.componentId === 'Dropdown' &&
                child.props.children[0].type.componentId === 'DropdownButton'
              )
            ) {
              console.error(
                `ButtonGroup only accepts button or dropdown elements as children.
              );
            }
          }
```

packages/blade/src/components/ButtonGroup/_decisions/decisions.md
saurabhdaware: Yup as Nitin mentioned. We basically do auto-handle it depending on space, but in this case we want to change the default position itself.

Currently the default position is always bottom-left but we need more options in button group

packages/blade/src/components/ButtonGroup/_decisions/decisions.md
saurabhdaware, line 111: Do we also need to support something like CollapsibleButton and other Button sub-components?

snitin315: I feel those aren't needed. We will evaluate them if any use case comes up.

saurabhdaware (approved):

PR #2073, opened 2024-03-11
feat: `OTPInput` component redesign

chaitanyadeorukhkar:
## Description
OTPInput Redesign 

## Changes

- Add size prop
- Update font family & size being used
- Update visuals

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


anuraghazra:
Why is the mask square shaped? 
image

expected design?

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -309,6 +309,11 @@ type BaseInputCommonProps = FormInputLabelProps &
      * ```
      */
     trailingButton?: React.ReactElement<LinkProps>;
+    /**
+     * Whether to use Text or Heading component for Input text
+     * @default text
+     **/
+    valueComponentType?: 'text' | 'heading';
```
anuraghazra, line 316: Don't have much context on this, what exactly was the usecase for this?

packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
```diff
@@ -181,7 +202,12 @@ export default {
 } as Meta<OTPInputProps>;
 
 const OTPInputTemplate: StoryFn<typeof OTPInputComponent> = ({ ...args }) => {
-  return <OTPInputComponent {...args} />;
+  const maxWidth = args.otpLength === 4 ? '376px' : '568px';
```
anuraghazra, line 205: So the OTPInput itself doesn't have MaxWidth is it?

packages/blade/src/components/Typography/Heading/getHeadingStyles.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const getHeadingStyles = ({
+  weight,
+  size,
+  theme,
+  color,
+}: Pick<HeadingProps, 'weight' | 'size' | 'color'> & {
+  theme: Theme;
+}): CSSObject => {
+  return getBaseTextStyles({ ...getHeadingProps({ weight, size, color }), theme });
```
anuraghazra, line 15: nice

anuraghazra (approved):

packages/blade/src/components/Input/OTPInput/OTPInput.stories.tsx
chaitanyadeorukhkar: Can't find it in ts so guessing nahi hai. We can add it if someone raises it though

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
chaitanyadeorukhkar: So for displaying value of the input, we sometimes need it to have the styles of our "Text" component and other times styles of our "Heading" component. Text & Heading differ in sizes as well as font families. OTPInput needs the input value to be shown as the same styles as Heading component whereas other input components need the same style as Text component

Issue #1978, opened 2024-01-17
Bug(Table): Clicking on multiselect checkbox the row doesn't get selected.


anuraghazra:
Raised by https://razorpay.slack.com/archives/G01B3LQ9H0W/p1705485717711829?thread_ts=1705483552.271549&cid=G01B3LQ9H0W

> If I click on the row it gets selected which is fine. But if I click on the checkbox, the row doesn’t get selected.

anuraghazra:
> joining your slack channel

Sorry it's private organization slack, only razorpay employees can join. 

Here's a video explaining the issue: 

https://github.com/razorpay/blade/assets/35374649/fc415c62-7057-409c-8b8b-1ce40c1a2193

amandal97:
### Premise 📄
The issue will only occur if the following are true:
1. Both  `Checkbox` and `Row` are clickable
2. `Blade Checkbox` is used for selecting rows

[...]

PR #2076, opened 2024-03-12
fix(Table): selection toggle for multiselect table

amandal97:
## Description

For a Multiselect Table Component, selection does not work if we click on the checkbox. However, it works if we click on the row itself. This PR is a short term solution to fix the problem.

Provided the issue link below for more details.

## Changes

Stop event propagation while clicking on the multiselect checkbox

## Additional Information

https://github.com/razorpay/blade/issues/1978

## Component Checklist

- [ ] Update Component Status Page
- [ ] Perform Manual Testing in Other Browsers
- [ ] Add KitchenSink Story
- [ ] Add Interaction Tests (if applicable)
- [ ] Add changeset

PR #2077, opened 2024-03-12
fix(Checkbox): stop unnecessary event delegation in checkbox

amandal97:
## Description

Event Delegation occurs twice even if the Checkbox is clicked once. This can cause issues while implementing analytics as a single click might infer that the checkbox has been clicked twice. Also, due to this, table multi-select feature not working if we click on the checkbox.

Provided the issue link below for more details.

## Changes

Stop propagation from all the child components of the `label` wrapper except the `input` element.

## Additional Information

https://github.com/razorpay/blade/issues/1978

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Checkbox/Checkbox.tsx
```diff
@@ -241,13 +241,18 @@ const _Checkbox: React.ForwardRefRenderFunction<BladeElementRef, CheckboxProps>
               isNegative={_hasError}
             />
             {children ? (
-              <SelectorTitle size={_size} isDisabled={_isDisabled}>
-                {children}
-              </SelectorTitle>
+              <BaseBox onClick={(e) => e.stopPropagation()}>
+                <SelectorTitle size={_size} isDisabled={_isDisabled}>
```
anuraghazra, line 245: Is it possible to have onClick on SelectorTitle itself instead of introducing a BaseBox wrapper?

saurabhdaware:
I think #2076 was merged which was related so closing this one since its been stale

PR #2081, opened 2024-03-15
feat: `SelectInput` component redesign

chaitanyadeorukhkar:
## Description

- `SelectInput` Redesign 
- `Autocomplete` Redesign

## Changes

- Add size prop
- Update visuals

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Input/BaseInput/baseInputTokens.ts
```diff
@@ -19,23 +19,19 @@ export const baseInputHeight: Record<
 };
 
 /**
- * 37px (36px height + 1px bottom line height)
+ * medium - 37px (36px height + 1px bottom line height)
+ * large - 49px (48px height + 1px bottom line height)
  */
 export const baseInputWrapperMinHeight = {
   medium: (baseInputHeight.medium as number) + BASEINPUT_BOTTOM_LINE_HEIGHT,
```
saurabhdaware, line 26: we can remove the `BASEINPUT_BOTTOM_LINE_HEIGHT` variable and its usage now since there is no bottom line in new inputs

chaitanyadeorukhkar: Done!

saurabhdaware (approved):

PR #2087, opened 2024-03-18
tests(AutoComplete): add Interaction tests for AutoComplete

saurabhdaware:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


snitin315 (approved):

packages/blade/src/components/Dropdown/useDropdown.ts
```diff
@@ -329,7 +329,7 @@ const useDropdown = (): UseDropdownReturnValue => {
 
       const filteredIndexes = filteredValues
         .map((filteredValue) => options.findIndex((option) => option.value === filteredValue))
-        .sort();
+        .sort((a, b) => a - b);
```
anuraghazra, line 332: why?

anuraghazra (approved):

packages/blade/src/components/Dropdown/useDropdown.ts
saurabhdaware: It is expected to sort indexes which are numbers and js does weird things with numbers - 
image


So in AutoComplete, when we add more than 10 items in search, the focus was jumping incorrectly to 10th item after 1st

PR #2112, opened 2024-04-03
fix: npmrc registry url

anuraghazra:
## Changes

- Forced npm registry to use the default npm registry instead of our local machine's razorpay scoped registry. (Since we are not using any private packages)
- Fixed yarn resolutions for react types. Since we had to regenerate the yarn.lock things got messed up

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/package.json
```diff
@@ -306,6 +306,7 @@
     "@storybook/**/react": "18.2.0",
     "react-dom": "18.2.0",
     "react": "18.2.0",
+    "@types/react": "17.0.38",
```
snitin315, line 309: We should target upgrading types to v18

.npmrc
```diff
@@ -0,0 +1 @@
+@razorpay:registry=https://registry.npmjs.org/
```
chaitanyadeorukhkar, line 1: Isn't the default npmjs registry?

packages/blade/package.json
```diff
@@ -250,7 +250,7 @@
     "react-native": "0.72.3",
     "react-native-gesture-handler": "2.9.0",
     "react-native-reanimated": "3.4.2",
-    "react-native-svg": "12.3.0",
+    "react-native-svg": "15.1.0",
```
chaitanyadeorukhkar, line 253: Why?

anuraghazra: v12 types are not compatible with react18 types. 

+ we and most of our consumers are using react-native >v70, thus according to this [table](https://github.com/software-mansion/react-native-svg?tab=readme-ov-file#supported-react-native-versions) we should be using v15 of rnsvg.

.github/workflows/blade-pr-title-check.yml
```diff
@@ -2,7 +2,7 @@ name: Blade PR Title Check
 
 on:
   pull_request_target:
-    types: [open, edited, synchronize]
+    types: [opened, edited, synchronize]
```
saurabhdaware, line 5: This was incorrect all along? how was it working 😄 ?

package.json
```diff
@@ -29,10 +29,7 @@
     "@babel/eslint-parser": "7.21.8",
     "@changesets/cli": "2.24.1",
     "@types/jest": "26.0.23",
-    "@types/react": "17.0.4",
-    "@types/react-native": "0.64.4",
-    "@types/styled-components": "5.1.25",
-    "@types/styled-components-react-native": "5.1.3",
+    "@types/react": "18.2.24",
```
saurabhdaware, line 32: styled-components, react-native types not needed?

saurabhdaware: Oh just saw they are added in blade's package.json. So why is types/react needed here but types/styled-components are in blade's package.json?

packages/blade/src/components/ActionList/ActionListBox.native.tsx
```diff
@@ -23,17 +23,17 @@ const _ActionListBox = React.forwardRef<SectionList, ActionListBoxProps>(
   ({ sectionData, actionListItemWrapperRole, isMultiSelectable, isInBottomSheet }, ref) => {
     const { footerHeight, setContentHeight } = useBottomSheetContext();
 
-    const renderActionListItem = React.useCallback(({ item }) => {
+    const renderActionListItem = React.useCallback(({ item }: any) => {
```
saurabhdaware, line 26: wouldn't this make `renderActionListItem`'s type any as well? should it be React.ReactElement instead?

saurabhdaware (commented):
Looks good overall. Can you check if we can do something about those `any`s 🙈

kamleshchandnani (approved):

PR #2117, opened 2024-04-04
fix: additional parameters added to retain fraction digits in humanize Amount figure

### LINKED ISSUE #2115: Amount Component: Rounding values to nearest thousand for 10k - 100k
When using `humanize` suffix with`<Amount/>`, the component is rounding the numbers to the nearest thousand.

```
<Amount
  suffix="humanize"
  value={99500}
/>
```

Examples: 

1. 99500 -> 100K
2. 12500.45 -> 13k
3. 55500 -> 56k

Screenshots from the docs:
Screenshot 2024-04-04 at 12 23 57 PM
Screenshot 2024-04-04 at 12 24 32 PM
Screenshot 2024-04-04 at 12 24 44 PM

tarun-khanna:
Fixes #2115 

## Description
Passing below extra paramters to i18nify library to retain fraction digits in compact formatted value. 
```
minimumFractionDigits: 1,
trailingZeroDisplay: 'stripIfInteger',
```

For more detail on these attributes, refer to the [Intl API documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat). 

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/.changeset/tiny-beds-switch.md
+++ b/.changeset/tiny-beds-switch.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+[feat]: additional parameters added to retain fraction digits in humanize Amount figure
--- a/.changeset/tiny-beds-switch.md
+++ b/.changeset/tiny-beds-switch.md
@@ -2,4 +2,4 @@
 "@razorpay/blade": patch
 ---
 
-[feat]: additional parameters added to retain fraction digits in humanize Amount figure
+fix: additional parameters added to retain fraction digits in humanize Amount figure
--- a/packages/blade/src/components/Amount/Amount.tsx
+++ b/packages/blade/src/components/Amount/Amount.tsx
@@ -190,6 +190,7 @@ export const formatAmountWithSuffix = ({
           intlOptions: {
             notation: 'compact',
             minimumFractionDigits: 1,
+            maximumFractionDigits: 2,
             trailingZeroDisplay: 'stripIfInteger',
           },
         });
--- a/packages/blade/src/components/Amount/Amount.tsx
+++ b/packages/blade/src/components/Amount/Amount.tsx
@@ -189,6 +189,8 @@ export const formatAmountWithSuffix = ({
         const formatted = formatNumber(value, {
           intlOptions: {
             notation: 'compact',
+            minimumFractionDigits: 1,
+            trailingZeroDisplay: 'stripIfInteger',
           },
         });
         return {
--- a/packages/blade/src/components/Amount/Amount.tsx
+++ b/packages/blade/src/components/Amount/Amount.tsx
@@ -189,7 +189,6 @@ export const formatAmountWithSuffix = ({
         const formatted = formatNumber(value, {
           intlOptions: {
             notation: 'compact',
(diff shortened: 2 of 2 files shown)


snitin315 (commented):
We need to update snaps. I see  `1T` has also changed to `1.0T`, is it possible to omit `.0`?

mihir-karbelkar-rp:
Shall we add a test case for this? From what I can see is that we had our own [implementation](https://github.com/razorpay/blade/blob/d447772dc639f461e931a0d2ac680db0ff49581e/packages/blade/src/components/Amount/Amount.tsx#L176) of humanized format and then we switched to Intl and that broke an existing functionality which is why I think adding a test covering this case for this component will help.

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -189,6 +189,8 @@ export const formatAmountWithSuffix = ({
         const formatted = formatNumber(value, {
           intlOptions: {
             notation: 'compact',
+            minimumFractionDigits: 1,
```
snitin315, line 192: do we have a prop to control max fraction digits too?

tarun-khanna:
> We need to update snaps. I see `1T` has also changed to `1.0T`, is it possible to omit `.0`?

Not sure snaps have to be updated here. I ran `yarn test:react -u` & `yarn test:react-native -u` but could see snaps changing only for phone number component. 
Also, 1000 is getting resolved to 1T only 🙃 ... refer below ss 
Screenshot 2024-04-04 at 7 04 24 PM

packages/blade/src/components/Amount/Amount.tsx
tarun-khanna: We have maxFractionDigits, but do we need it ? 
What was the number that you were trying here ? @snitin315

snitin315: Let's set the max digits as `2`. You can [try with `2566.780`.](https://61c19ee8d3d282003ac1d81c-cpfmlxnxou.chromatic.com/?path=/story/components-amount--amount&args=value:2566.78;type:display;suffix:humanize)

snitin315:
>Not sure snaps have to be updated here. I ran yarn test:react -u & yarn test:react-native -u but could see snaps changing only for phone number component.

@tarun-khanna Yes, you need to commit & push those updated snapshots. [CI is failing](https://github.com/razorpay/blade/actions/runs/8551668637/job/23431155295?pr=2117)

Screenshot 2024-04-05 at 11 55 30 AM

.changeset/tiny-beds-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+[feat]: additional parameters added to retain fraction digits in humanize Amount figure
```
snitin315, line 5:
```suggestion
fix: additional parameters added to retain fraction digits in humanize Amount figure
```

snitin315 (approved):
LGTM. Thank you ⭐

PR #2122, opened 2024-04-08
feat: add Circular Progress

snitin315:
#### Current API
- `<ProgressBar variant="meter | progress">`

#### Ideal New API

- `<ProgressBar type="meter | progress" variant="linear | circular" >`

#### Proposed API to Avoid Breaking Change

- `<ProgressBar type="meter | progress" variant="linear | circular | meter | progress">`
- If `type` prop is provided throw an error when the `variant` is `meter` or `progress.`
- Provide a codemod to migrate to the new API.
- Eventually, remove support for the old API & move to the Ideal API in the next major release.

## Related Stacked PRs

- Tests - #2126 
- Codemod - #2127 

## Component Checklist

- [x] Add KitchenSink Story
- [x] Add changeset


packages/blade/src/components/ProgressBar/CircularProgressBar.web.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const sqSize = circularProgressSizeTokens[size].size;
+
+  // SVG centers the stroke width on the radius, subtract out so circle fits in square
+  const radius = (sqSize - strokeWidth) / 2;
+  // Enclose circle in a circumscribing square
+  const viewBox = `0 0 ${sqSize} ${sqSize}`;
+  // Arc length at 100% coverage is the circle circumference
+  const dashArray = radius * Math.PI * 2;
+  // Scale 100% coverage overlay with the actual percent
+  const dashOffset = dashArray - (dashArray * progressPercent) / 100;
```
anuraghazra, line 103: I think the same math is being used in both RN/Web. Let's move this to a shared util?

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -34,7 +40,7 @@ type ProgressBarCommonProps = {
    * Sets the size of the progress bar.
    * @default 'small'
```
anuraghazra, line 41: We should add a small note about the runtime check we have added. 

Like which variant cannot be used with which size.

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -113,13 +122,36 @@ const ProgressBar = ({
...
+      throwBladeError({
+        moduleName: 'ProgressBar',
+        message: `Cannot set 'isIndeterminate' when 'type' is 'meter'.`,
+      });
+    }
+
+    if (progressVariant === 'circular' && isIndeterminate) {
+      throwBladeError({
+        moduleName: 'ProgressBar',
+        message: `Cannot set 'isIndeterminate' when 'variant' is 'circular'.`,
```
anuraghazra, line 140: So we aren't sufficing the spinner usecase with Circular progressbar is it?

packages/blade/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -113,13 +122,36 @@ const ProgressBar = ({
...
+    }
+
+    if (progressVariant === 'linear' && size === 'large') {
+      throwBladeError({
+        moduleName: 'ProgressBar',
+        message: `Large size isn't available when 'variant' is 'linear'.`,
+      });
+    }
+
+    if (type && variant && (variant === 'progress' || variant === 'meter')) {
```
anuraghazra, line 151:
```suggestion
    if (type && (variant === 'progress' || variant === 'meter')) {
```

packages/blade/src/components/ProgressBar/progressBarTokens.ts
```diff
@@ -15,4 +15,22 @@ const pulseAnimation = {
   backgroundColor: 'white',
 };
 
-export { indeterminateAnimation, pulseAnimation };
+const circularProgressSizeTokens = {
+  small: {
+    size: 24,
+    strokeWidth: 3,
```
anuraghazra, line 21: these values are pixels or spacing tokens or sizing tokens?

anuraghazra (changes requested):
Mostly lgtm, minor comments.

packages/blade/src/components/ProgressBar/ProgressBar.tsx
snitin315: Yes. The Spinner will be a separate component.

snitin315:
> @snitin315 can you check this once? this is also present on master. Not sure if this layout shift is intended or not.

@anuraghazra I'll check this separately.

packages/blade/src/components/ProgressBar/CircularProgressBar.native.tsx
saurabhdaware, line 1: https://61c19ee8d3d282003ac1d81c-qbudppbfos.chromatic.com/?path=/story/components-progressbar--circular-progress&args=value:36;type:meter;isIndeterminate:!false;showPercentage:!true

The percentage is not showing up even with the showPercentage: true option in meter. Is that expected? There might be usecases of showing percentage in middle in circular meter on dashboard no?

packages/blade/src/components/ProgressBar/ProgressBar.stories.tsx
saurabhdaware, line 1: Lets remove the variant meter and progress from storybooks table?

packages/blade/src/components/ProgressBar/CircularProgressBar.native.tsx
snitin315: It's intended. We don't shoe percentage in the `meter` variant. https://blade.razorpay.com/?path=/story/components-progressbar--progress-bar-meter-variant&args=showPercentage:!true

packages/blade/src/components/ProgressBar/ProgressBar.stories.tsx
snitin315: It's using `autodocs` so it's auto-generated, we will need to manually define all props otherwise. I think it's fine to keep it as is now since it is supported.

packages/blade/src/components/ProgressBar/CircularProgressBar.web.tsx
snitin315: done

anuraghazra (approved):

anuraghazra (approved):

PR #2123, opened 2024-04-09
feat: setup webdriverio

anuraghazra:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Input/PhoneNumberInput/__tests__/PhoneNumberInput.e2e.stories.tsx
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
saurabhdaware, line 1: Do we want to write separate stories for E2E? what if test on stories that we write for documentation? This might also force us to cover more cases in documentation itself

packages/blade/tsconfig.json
```diff
@@ -10,6 +10,14 @@
       "~tokens/*": ["src/tokens/*"],
       "~src/*": ["src/*"]
     },
+    "types": [
```
saurabhdaware, line 13: why do we have to define it like this? the import from package would resolve these types no?

packages/blade/wdio.bstack.conf.ts
```diff
@@ -0,0 +1,92 @@
...
+    {
+      browserName: 'Firefox',
+      'bstack:options': {
+        browserVersion: 'latest',
+        os: 'OS X',
+        osVersion: 'Sonoma',
+        ...commonCapabilities,
+      },
+    },
+    // Skipping safari for now, our storybook is broken in safari desktop :(
```
saurabhdaware, line 44: I am able to open it in my laptop 😕

packages/blade/wdio.bstack.conf.ts
```diff
@@ -0,0 +1,92 @@
...
+        osVersion: '13.0',
+        ...commonCapabilities,
+      },
+    },
+  ],
+};
+
+const config = { ...baseConfig, ...bstackConfig };
+
+// Code to support common capabilities
```
saurabhdaware, line 86: Common capabilities of? also expected to be commented out?

packages/blade/tsconfig.json
anuraghazra: As part of their [TS](https://webdriver.io/docs/typescript/) setup.

webdriver.io injects globals into the tests so we need these types to come up and modify the `expect` etc functions.

packages/blade/wdio.bstack.conf.ts
anuraghazra: Removed.

packages/blade/src/components/Input/PhoneNumberInput/__tests__/PhoneNumberInput.e2e.stories.tsx
anuraghazra: We can, but E2Es has very specific stories too, it might make the stories confusing if we just keep these as part of main stories.

packages/blade/wdio.bstack.conf.ts
anuraghazra: blade.razorpay.com? 
What's your safari version? I'm using v15.4

image

packages/blade/src/components/Input/PhoneNumberInput/__tests__/PhoneNumberInput.e2e.stories.tsx
saurabhdaware: My concern is there's too many stories to write now ;__; 

What if we keep the stories that we don't want to expose to consumer as Internal?

In most cases the stories we write for consumers would work for testing and for the cases where it doesn't, we can make them internal and then test on them?

saurabhdaware:
@anuraghazra we can close this also? keep the branch so we can re-open

PR #2124, opened 2024-04-12
docs(StepGroup): API Decision

saurabhdaware:
## [✨ Formatted Document ✨](https://github.com/razorpay/blade/blob/stepper/step-0/packages/blade/src/components/StepGroup/_decisions/decisions.md)


packages/blade/src/components/StepGroup/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Interactive StepGroup
+
+Interactive StepGroup can be used to add clickable items in StepGroup. When StepItem has `onClick` or `href` prop, it becomes interactive.
+
+<img width="200px" src="./2024-04-12-15-16-25.png" alt="" />
+
+```jsx
+const InteractiveStepGroup = () => {
+  const [selectedIndex, setSelectedIndex] = React.useState(-1);
```
anuraghazra, line 88: How to handle interactive + nested step group?

packages/blade/src/components/StepGroup/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+## Props
+
+### StepGroup
+
+| **Props**   | **Description**          | **Type**                   | **Default Value** |
+| ----------- | ------------------------ | -------------------------- | ----------------- |
+| size        | Size of StepGroup        | 'medium' \| 'large'        | 'medium'          |
+| orientation | Orientation of StepGroup | 'horizontal' \| 'vertical' | 'vertical'        |
+
+### StepItem
```
anuraghazra, line 20: Where's children for StepItem? 

Slot ka content kaha jayga?

packages/blade/src/components/StepGroup/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+- [React Stepper - MUI](https://mui.com/material-ui/react-stepper/)
+- [Stepper - ZenDesk](https://garden.zendesk.com/components/stepper)
+
+All follow a similar compound API.
+
+## Open Questions
+
+- ### Should we add `isInteractive` prop?
+
+  In the proposed API, I have proposed that we can turn item into interactive or static based on whether it has `onClick` or `href` or nothing. Is it intuitive enough? or should we add more explict prop called `isInteractive` like we have in design
```
anuraghazra, line 180: I think this is Okay, similar thing we are doing on Card too. where adding href/onClick makes the card interactive without an explicit prop.

packages/blade/src/components/StepGroup/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+
+- ### Alternative to `leading`
+
+  Currently I have proposed `leading` prop where we can add Icon or indicator. Although in horizontal orientation, its not exactly "leading". It comes on top. Its also not very equivalent to leading we have in other components.
```
anuraghazra, line 184: what happens to trailing on horizontal variant?

packages/blade/src/components/StepGroup/_decisions/decisions.md
saurabhdaware: this index is anyway consumer-side only so they can either just give 1, 2, 3, 4, indexes in sequential manner indepdent of nesting and handle

Or they can keep state like `useState('1-2')` to represent second item in first StepGroup. Either way eventually they just have to give `isSelected` to selected item somehow

packages/blade/src/components/StepGroup/_decisions/decisions.md
saurabhdaware: There is no trailing in horizontal

packages/blade/src/components/StepGroup/_decisions/decisions.md
```diff
@@ -0,0 +1,219 @@
...
+  An approach that MUI takes is mix of above 2 approaches -
+
+  ```jsx
+  <StepGroup selected={selectedIndex}>
+    <StepItem onClick={() => setSelecteIndex(0)} />
+    <StepItem onClick={() => setSelecteIndex(1)} />
+  </StepGroup>
+  ```
+
+  Q. Should we go with 1st, 2nd, or 3rd approach?
```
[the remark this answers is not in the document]
saurabhdaware: Yup I am also leaning towards 1st one because of flexibility. I don't think so multiple selections is a usecase but that additional condition seems fair.

Also biggest reason is if we go with 3rd or 2nd, we have to define how we want to handle nested + interactive. So might have to define some strange index like `selected="1-2"`

Which is not the concern in 1st - https://github.com/razorpay/blade/pull/2124#discussion_r1562425878

packages/blade/src/components/StepGroup/_decisions/decisions.md
```diff
@@ -0,0 +1,228 @@
...
+| onClick      | StepItem's click handler. Turns StepItem into interactive item and render it as button tag        | () => void                                                         | undefined                             |
+| children     | Slot                                                                                              | ReactNode                                                          | undefined                             |
+
+### StepItemIcon
+
+| **Props** | **Description** | **Type**                                                                        | **Default Value** |
+| --------- | --------------- | ------------------------------------------------------------------------------- | ----------------- |
+| icon      | icon component  | IconComponent                                                                   |                   |
+| color     | color of icon   | 'positive' \| 'negative' \| 'neutral' \| 'notice' \| 'information' \| 'primary' | 'neutral'         |
```
snitin315, line 41: Default isn't `primary`?

packages/blade/src/components/StepGroup/_decisions/decisions.md
```diff
@@ -0,0 +1,228 @@
...
+  <StepItem
+    leading={<StepItemIcon icon={CheckIcon} color="positive" />}
+    title="Disputes Raised"
+    timestamp="Thu, 11th Oct23 | 12:00pm"
+    stepProgress="full"
+    trailing={<Badge>Label</Badge>}
+  />
+  {/* Nested StepGroup */}
+  <StepGroup>
+    <StepItem
```
snitin315, line 133: So we need to follow Nested `StepGroup` pattern to show indentation. Is it possible to handle indentation via a prop? Something like:

```js
<StepGroup>
  <StepItem />
  <StepItem />
  <StepItem parentIndex="1" />
  <StepItem parentIndex="1" />
  <StepItem />
</StepGroup>
```

packages/blade/src/components/StepGroup/_decisions/decisions.md
saurabhdaware: nope. Primary is blue color so having it as default will make it seem like everything is important

packages/blade/src/components/StepGroup/_decisions/decisions.md
saurabhdaware: yup. But we follow this structure for nesting in List component and common structure for nesting in compound components.

snitin315: Makes sense then.

snitin315 (approved):
LGTM

PR #2134, opened 2024-04-19
test(StepGroup): Tests for StepGroup

saurabhdaware:
## Component Checklist

- [x] Add KitchenSink Story
- [x] Add changeset


packages/blade/src/components/StepGroup/StepLine.web.tsx
```diff
@@ -24,7 +25,7 @@ const useDottedLineStyles = ({ isHorizontal }: { isHorizontal?: boolean } = {}):
   </svg>
   `;
 
-  const backgroundURL = `url(data:image/svg+xml;base64,${window.btoa(svgString)})`;
+  const backgroundURL = `url(data:image/svg+xml;base64,${encode(svgString)})`;
```
saurabhdaware, line 28: need this library because we don't have any server build so we can't separate out `import Buffer from 'buffer'` which is needed to base64 conversion in node into separate build. Checks dont work because import itself needs to be differet.

This library handles that.

packages/blade/src/components/StepGroup/__tests__/StepGroup.web.test.tsx
```diff
@@ -0,0 +1,88 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+/* eslint-disable @typescript-eslint/no-unsafe-return */
```
snitin315, line 3: Are these needed? I'm unsure about the `no-unused-vars` rule.

packages/blade/src/components/Amount/__tests__/__snapshots__/Amount.web.test.tsx.snap
```diff
@@ -1934,36 +1934,6 @@ exports[`<Amount /> should render isStrikethrough={true} 1`] = `
   opacity: 0.64;
 }
 
-@media screen and (min-width:320px) {
-  .c5.c5.c5.c5.c5 {
-    border-bottom-style: solid;
-  }
-}
```
snitin315, line 1941: Did we change anything in Amount or Box?

snitin315 (dismissed):
Overall looks good, minor comments.

packages/blade/src/components/Amount/__tests__/__snapshots__/Amount.web.test.tsx.snap
saurabhdaware: yup we changed it in Box. We were adding borderBottomStyle by default earlier which was making it add it to all media queries.

Now we only add it to given breakpoint so it removed the additional code from media queries

packages/blade/src/components/StepGroup/__tests__/StepGroup.web.test.tsx
saurabhdaware: ops nope. none of these are needed. I copy-pasted base test code from somewhere so it came with that 😆 Removed them all

packages/blade/src/components/StepGroup/StepLine.web.tsx
saurabhdaware: we can but we'll also have to make changes in our build step. We handle this in universe but in blade repo we're only handling web and native

anuraghazra (approved):
LGTM

chaitanyadeorukhkar (approved):

snitin315 (approved):

PR #2139, opened 2024-04-23
docs: SearchInput API

snitin315:
### [Formatted Document](https://github.com/razorpay/blade/blob/docs/search-input-api/packages/blade/src/components/Input/SearchInput/_decisions/decisions.md)


saurabhdaware (commented):
How are we handling Global Search usecase from figma?

image

snitin315:
>How are we handling Global Search usecase from figma?

@saurabhdaware It's nothing but a search input inside a Modal. We will be just exposing an input field and users can perform any action based on this field's value. We are not enforcing any patterns with the release of this component.

packages/blade/src/components/Input/SearchInput/_decisions/decisions.md
```diff
@@ -0,0 +1,184 @@
...
+    'Payouts',
+  ];
+  const popularItems = ['Transactions', 'Settlements'];
+
+  return (
+    <Dropdown>
+      <SearchInput
+        label="Search"
+        onChange={({ value }) => setSearchTerm(value)}
+        placeholder="Search here"
```
saurabhdaware, line 95: Have we explored non-compound API here?

I see that items are being set from array only so we need data in array.

In SelectInput and AutoComplete, we went with this compound API because consumers can keep their data inside `<ActionListItem />` individually, add sections, descriptions, trailing text etc

If keeping data in anyway a requirement for SearchInput, can we just have-

```
<SearchInput
  label="Something"
  data={[

  ]}
/>
```

... type of API and internally use Dropdown component?

packages/blade/src/components/Input/SearchInput/_decisions/decisions.md
```diff
@@ -0,0 +1,184 @@
...
+                </TableRow>
+              ))}
+          </TableBody>
+        </>
+      )}
+    </Table>
+  );
+};
+```
+
```
saurabhdaware, line 176: Add references to how other design systems handle their SearchInput if there are any?

saurabhdaware:
> @saurabhdaware It's nothing but a search input inside a Modal. We will be just exposing an input field and users can perform any action based on this field's value. We are not enforcing any patterns with the release of this component.

Can you add an example of this in API decision? not sure if I understood correctly. Does that mean that SearchInput is just TextInput with SearchIcon? How do we render that list inside the Modal? is it custom?

packages/blade/src/components/Input/SearchInput/_decisions/decisions.md
saurabhdaware: Discussed on slack. SearchInput is small wrapper on top of TextInput. It also has modal and pattern-like usecases so better to not tie it to Dropdown.

snitin315:
>  Does that mean that SearchInput is just TextInput with SearchIcon? 

Yes.

> How do we render that list inside the Modal? is it custom?

Modal content will be handled by the consumers, this is just an input field we are providing.

packages/blade/src/components/Input/SearchInput/_decisions/decisions.md
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * Callback that is called when the clear button is clicked.
+   */
+  onClearButtonClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
+};
+```
+
+## Usage
+
+### With Dropdown
```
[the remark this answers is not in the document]
snitin315: We are not handling any patterns, we will be just exposing an input field. Consumers can use it anywhere in there UI

packages/blade/src/components/Input/SearchInput/_decisions/decisions.md
```diff
@@ -0,0 +1,184 @@
+# SearchInput
+
+The Search Input component enables users to input text and find relevant information within a diverse range of data or choices.
```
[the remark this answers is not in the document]
snitin315: Good Question.

anuraghazra (approved):

PR #2143, opened 2024-04-30
fix: tab item button type

grvsharma1810:
## Description

Since, `button` element is used as an undelying element for `TabItem`, If the `Tab` component is wrapped inside the `form`, clicking on `TabItem` is triggering form `onSubmit` event which should not be the case. Hence changing the type of `button` to `button` for `TabItem` to prevent this.

## Changes

Change type of `button` as `button` for `TabItem`

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


anuraghazra (dismissed):

anuraghazra:
@grvsharma1810 can you update the snapshots?

anuraghazra (approved):

PR #2149, opened 2024-05-02
docs(SideNav): API Decision

saurabhdaware:
#### First Class Tooltip support as suggested by @anuraghazra 

```diff
- <Tooltip content="Something">
-   <SideNavLink />
- </Tooltip>
+ <SideNavLink tooltip={{content: 'Something'}}>
```

#### as prop on SideNavLink as suggested by @kamleshchandnani 

```diff
- <SideNav routerLink={NavLink}>
-  <SideNavLink />
- </SideNav>
+ <SideNav>
+  <SideNavLink as={NavLink} />
+ </SideNav>
```

#### Active state won't be handled internally

Consumers have to explicitly mark item as Active based on their router state. Checkout formatted doc for more info

```diff
<SideNavLink
   as={NavLink}
+  isActive={true}
/>
```

#### SideNavItem component addition

Instead of asking consumers to build custom item with Box, we give paddings and default styles pre-configured in SideNavItem. We can use it for Test Mode usecase now

```jsx
<SideNavItem
  leading={}
  trailing={}
  title=""
/>
```

#### SideNavBody component addition

We needed to add SideNavBody to add scroll in the items

```diff
<SideNav>
-   <SideNavLink />
-   <SideNavLink />
+  <SideNavBody>
+    <SideNavLink />
+    <SideNavLink />
+  </SideNavBody>
</SideNav>
  
```


packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+- [Stackblitz - One Nav Router POC](https://stackblitz.com/edit/one-nav-poc-router?file=App.tsx,NavSubComponents.tsx,package.json,SideNav.tsx)
+
+## API
+
+```jsx
+import { NavLink } from 'react-router-dom';
+
+<SideNav routerLink={NavLink}>
+  {/* L1 Items */}
+  <SideNavLink title="Home" icon={HomeIcon} href="/" />
```
kamleshchandnani, line 21: how can we make a particular nav item selected by default? also, it's not just default if someone lands on a URL say one merchant sharing it with another/our support team when we land on that URL the nav item shall be selected by default so shall we also have controlled API?

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+<td><img src="navlink-eg-3.png" width="200px" /></td>
+</tr>
+
+<tr>
+
+<td>
+
+<!-- prettier-ignore -->
+```jsx
+<Tooltip content="Action Name (Cmd + P)">
```
kamleshchandnani, line 251: is it also possible to have an info icon and show the tooltip on that? can we add an example of that?

kamleshchandnani: also shall we not have tooltip on the entire nav link and have on info icon? because a11y?

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+<table>
+<tr>
+<td>
+
+```jsx
+<SideNav>
+  {/* L1 Items */}
+  <SideNavLink title="L1 Item" />
+
+  <SideNavLink title="L2 Trigger">
```
kamleshchandnani, line 284: what if we just have NavLink and provide `onClick` and `href` as 2 props and if devs want nested levels they can use onClick and if its href that's a navigation? its the same terminology as link as button and link as nav

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+| children  | children slot. Accepts SideNavLink as children        | JSX      |                   |
+
+<table>
+<tr>
+<td>
+
+```jsx
+<SideNav>
+  {/* L1 Items */}
+  <SideNavLink title="L1 Item" />
```
kamleshchandnani, line 282: how are we deciding when to open the sub nav(L2) and when to expand it(L3)? using context?

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+
+</table>
+
+### SideNavLevel
+
+Nested SideNavLevel components create new levels. This can be used to create L1 - L2 - L3 levels in your navbar
+
+| **Props** | **Description**                                       | **Type** | **Default Value** |
+| --------- | ----------------------------------------------------- | -------- | ----------------- |
+| title     | Only applicable in L2. Becomes the title of the Level | string   |                   |
```
kamleshchandnani, line 272: we need to give a way for product teams to track analytics for every sub-component of Nav being clicked.

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+</tr>
+</table>
+
+### SideNavSection
+
+| **Props**       | **Description**                                              | **Type** | **Default Value** |
+| --------------- | ------------------------------------------------------------ | -------- | ----------------- |
+| title           | title of the section                                         | string   |                   |
+| maxVisibleItems | Number of items visible (rest go inside +x more collapsible) | number   | undefined         |
+| children        | Children slot. For SideNavLink children items                | JSX      |                   |
```
kamleshchandnani, line 314: when a user clicks the +13 more we need to give them a trigger to track analytics

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+
+<!-- prettier-ignore -->
+```jsx
+<Tooltip content="Action Name (Cmd + P)">
+  <SideNavLink
+    icon={LayoutIcon}
+    title="L1 Item Name"
+    href="/new-item-link"
+  />
+</Tooltip>
```
anuraghazra, line 257: This will cause a lot of issues while iterating over the nested items, because now you have to map over the Tooltip and allow it as part of the composite element.

anuraghazra: We can instead have this as a first class feature. 

`<SideNavLink tooltip={{ content: "Tooltip content" }} />

saurabhdaware: > This will cause a lot of issues while iterating over the nested items, because now you have to map over the Tooltip and allow it as part of the composite element.

Trying out approach that doesn't require iterating over items

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: As discussed in office, even if we support adding isSelected or isActive prop, it will only select item and not change the route itself. With the current approach, they can move to their selected route and that will select item

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: I think it will be there on overall item (similar to our Tooltip with Button example). Because tooltip represents what happens on click of item.

Anurag suggested having tooltip inside SideNavLink component as first-class feature so will explore that

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: Currently clicking item also changes the route so need href there

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: yup using context

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: onClick on SideNavLink should be able to cover most usecases 🤔

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: yup right. Will add this

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+
+```jsx
+import { NavLink } from 'react-router-dom';
+
+<SideNav routerLink={NavLink}>{/* children */}</SideNav>;
+```
+
+### SideNavLink
+
+| **Props**   | **Description**                                                                                                              | **Type**                                                            | **Default Value**                                            |
```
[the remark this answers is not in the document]
saurabhdaware: no

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+  icon={PayoutIcon}
+  title="Create Payouts"
+  href="/payouts"
+  trailing={
+    <Button
+      href="/payouts/create"
+      icon={PlusIcon}
+      size="xsmall"
+      variant="tertiary"
+    />
```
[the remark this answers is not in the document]
saurabhdaware: I think both trailing and item itself can have tooltip so we can keep the one that we have below. I'll add tooltip example on trailing

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: > `<SideNavLink tooltip={{ content: "Tooltip content" }} />

This makes sense. Lets go with this

packages/blade/src/components/SideNav/_decisions/decisions.md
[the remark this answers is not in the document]
saurabhdaware: yes. And currently only button is accepted in trailing slot

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,545 @@
...
+        as={NavLink} 
+        title="Settings" 
+        icon={UserIcon} 
+        // sets the link as active 
+        isActive={true} 
+        href="/accounts/settings" 
+      />
+      <SideNavLink as={NavLink} title="Edit" icon={UserIcon} href="/accounts/settings">
+        {/* L3 */}
+        <SideNavLevel>
```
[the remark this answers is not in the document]
saurabhdaware: > do we really need to have sidelevel here? this kind of is leading to confusion because I can't have title on level 3

I missed removing `title` prop from SideNavLevel in this example. Had removed it from props table. So yup we won't have `title` prop in both L2 and L3.

I kept SideNavLevel here because it then follows the same structure in L2 and L3 so people can recursively loop through the object and render items without any additional logic

> Also, the SideNavLink /accounts/settings you won't navigate right when you're on L2 and want to access L3 items because in this case the Edit its just acting like a collapsible item?

This is true. Not sure what is alternative. Currently I have kept `href` optional so if you don't have href value here, it opens collapsible without navigation. But if you have href, then it navigates and opens collapsible at the same time (kind of like https://react.dev/reference/react-dom/client)

packages/blade/src/components/SideNav/_decisions/decisions.md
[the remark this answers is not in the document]
chaitanyadeorukhkar: We should mention somewhere in documentation and here in prop table that trailing is only visible on hover

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,545 @@
...
+/>
+```
+
+</td>
+<td><img src="navlink-eg-tooltip.png" width="300px" /></td>
+</tr>
+
+</table>
+
+### SideNavLevel
```
chaitanyadeorukhkar, line 336: I hope we will handle the edge case where people try to add L4 sidenav level and throw some meaningful error

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+</td>
+</tr>
+</table>
+
+### SideNavSection
+
+| **Props**       | **Description**                                              | **Type** | **Default Value** |
+| --------------- | ------------------------------------------------------------ | -------- | ----------------- |
+| title           | title of the section                                         | string   |                   |
+| maxVisibleItems | Number of items visible (rest go inside +x more collapsible) | number   | undefined         |
```
vinilvasani: if maxVisibleItems is not provided will it be in expanded state by default?

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+
+<td>
+<img width="300px" src="navfooter-eg.png" />
+</td>
+</tr>
+</table>
+
+## Accessibility
+
+1. All items should be accessible by `TAB`. Including going between levels L1, L2, L3
```
vinilvasani, line 465: when the SideNav mounts will the initially active item be in focus by default?

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+2. We should use Blade's `SkipNav` utility to provide option of skipping nav and going to content
```
vinilvasani, line 466: what is the SkipNav utility can you share more context about it?

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,545 @@
...
+  ##### Conclusion
+
+  We are going with `as={NavLink}`
+
+- #### Why `isActive` prop? Why not handle active state internally
+
+  Earlier we started the API by saying that marking of the link as active based on route will be handled internally in blade component. There are some implemetation constraints when we go with this approach of handling active link internally
+
+  1. We don't have access to hooks like `useLocation` or utilities like `matchPath` inside Blade since blade is library indepedent of the routing logic
+  2. Although React Router's NavLink automatically adds `aria-current="page"` to active link which can be used to change colors in CSS, it doesn't give us access to handling state internally for our L1 -> L2 navigations [Here's the related proposal I created in React Router's Repo](https://github.com/remix-run/react-router/discussions/11543).
```
vinilvasani, line 526: NavLink internally maintains the isActive based on route defined right? Can we not make use of that?


> it doesn't give us access to handling state internally for our L1 -> L2 navigations

can you please elaborate on this a little?

vinilvasani (commented):
The API decision doc is nicely put together and was fun to read through 🚀 
Have dropped a few comments

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+        },
+      ],
+    },
+  ]}
+/>
+```
+
+#### Why not config-driven API?
+
+1. Config-Driven API requires you to have a low-level compound API anyways since compound APIs give higher flexibility into adding leading, trailing, descriptions, and other items when needed.
```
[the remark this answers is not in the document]
saurabhdaware: Lets say you want to add leading, trailing, descriptions, etc to items. With config-driven API we'll have to keep on adding that to the config schema. So only things that we allow in our config schema are valid rest are invalid.

With compound API, on lower level you can put any JSX and that will be rendered. So if any new usecase comes, in a lot of scenarios we might be able to write custom JSX that allows adding that usecase without opting out of entire SideNav.

```jsx
<SideNav>
  <Box>Random other text</Box>
</SideNav>
```

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+
+| **Props**  | **Description**                                                              | **Type**             | **Default Value** |
+| ---------- | ---------------------------------------------------------------------------- | -------------------- | ----------------- |
+| routerLink | Prop for passing NavLink of React Router                                     | NavLinkComponentType |                   |
+| children   | children slot of SideNav, accepts SideNavLink, SideNavSection, SideNavFooter | JSX                  |                   |
+
+```jsx
+import { NavLink } from 'react-router-dom';
+
+<SideNav routerLink={NavLink}>{/* children */}</SideNav>;
```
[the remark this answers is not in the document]
saurabhdaware: We have `onClick` on items so that can be used for analytics. E.g. If onClick on L2 trigger item is called then its L2 open event for analytics

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+| rel         | anchor tag rel attribute [rel - MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#rel)          | AnchorRelType                                                       | target === ' \_blank ' ? ' noreferrer noopener ' : undefined |
+| onClick     | Click handler on item                                                                                                        | (e: React.MouseEvent) => void                                       |                                                              |
+| icon        | Blade's Icon Component                                                                                                       | IconComponent                                                       |                                                              |
+| trailing    | Trailing Slot of Item. Can be used for adding Quick Shortcut Button, Trailing Text                                           | JSX                                                                 |                                                              |
```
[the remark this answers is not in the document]
saurabhdaware: Added the screenshots in table below. titleSuffix will come next to title and trailing will be on the right. However depending on usecase and how it looks, size of title, etc. Designers can take call whether they really want to show both or if just one is fine

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+| titleSuffix | Slot after the title to add Badge, Counter                                                                                   | JSX                                                                 |                                                              |
```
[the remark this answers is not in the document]
saurabhdaware: `titleSuffix` is the standard name we follow in components like AccordionHeader, Card, DropdownHeader, etc to make API more predictable across blade so following the same thing here

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+| ref         | Forwards the given ref. This also allows it to be a trigger for Tooltip                                                      | idk man ref types are hard.<br/> I'll figure out while implementing |                                                              |
```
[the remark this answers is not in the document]
saurabhdaware: I have removed it. I think for some reason your review went into older version of API 🤔

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+#### Why not this API?
+
+1. While its the easiest to implement considering how close it is to final DOM structure 🙈, it is complex in understanding and the backend schema we have has nested L1, L2 JSON. So its more complex to loop through a data like that and render this structure
+2. When you go from L1 -> L2, SideNav slides from one nav level to other where this API makes sense. But when you go from L2 -> L3, L3 becomes a collapsed menu where nested API makes sense.
+
+</details>
+
+## Props
+
+### SideNav
```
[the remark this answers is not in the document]
saurabhdaware: We tried that but failed. Have added reasoning in open questions

packages/blade/src/components/SideNav/_decisions/decisions.md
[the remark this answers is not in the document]
saurabhdaware: image
When one of the item inside is marked active, its parent is also marked active in L1

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: yup in that case all items are always visible and that "Show More" link itself won't exist

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: no the focus will be controlled on page level so assuming the top-nav would have first focus and then sidenav

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: > NavLink internally maintains the isActive based on route defined right? Can we not make use of that?

It doesn't give any `isActive` in JS / React state. It adds aria-current="page" to link which can be used for CSS styling but because we have multiple menus where lets say you want to open L2 from L1, you need a way to know which item is Active in the react state (Added a bit more context in that React Router's proposal)

> NavLink internally maintains the isActive based on route defined right

Basically the issue is its "only" internally maintaining that active state inside React Router's NavLink. Its not giving us a way to access that active state in our component's state

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: We have this utility in blade - https://blade.razorpay.com/?path=/docs/components-accessibility-skipnav--docs

So if you start tabbing, you see a Skip to content link which you can click to skip navigation and go to content directly

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,418 @@
...
+  <SideNavLink
+    title="Create Payouts"
+    trailing={<Button icon={PlusIcon} variant="tertiary" />}
+    icon={HomeIcon}
+    href="/create-payouts"
+  />
+
+  <SideNavLink title="Accounts" icon={AccountsIcon} href="/accounts/profile">
+    {/* L2 */}
+    <SideNavLevel title="Accounts">
```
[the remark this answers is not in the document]
saurabhdaware: From implementation-point, we can. Earlier we discussed that having SideNavLevel just gives an easier differentiator into understanding that its new level. Nested SideNavLink does't make it that clear on what happens to the nested items. 

Moreover the Heading of SideNavLevel is also part of this so having a new container makes it clear that its not just items that you're rendering
image

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,623 @@
...
+    />
+    <SideNavLink 
+      as={NavLink}
+      icon={GearIcon}
+      href="/settings" 
+      title="Settings" 
+    >
+      {/* L2 SideNavLink items */}
+    </SideNavLink>
+  </SideNavFooter>
```
anuraghazra, line 529: How will this actually work? 

Because SideNavLink will have it's own styling (hover, bg etc) and you are putting it inside SideNavItem which will also have it's own styling.

packages/blade/src/components/SideNav/_decisions/decisions.md
```diff
@@ -0,0 +1,623 @@
...
+      href="/create-payouts"
+    />
+
+    <SideNavLink 
+      as={NavLink} 
+      title="Accounts" 
+      icon={AccountsIcon} 
+      // sets the submenu as active
+      isActive={true}
+      href="/accounts"
```
anuraghazra, line 39: Why does a submenu has an href? 

Shouldn't the href of the first children of this submenu be auto selected?

saurabhdaware: I can't tell which is the first element without doing `React.Children.map` which we can't do because people will create wrapper NavLink which handles react router state

saurabhdaware: It still works. Just L2 menu won't have any selection and it will take you to submenu's href

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: Basically I pass `level` in context, then in next context I pass `level + 1` so it gives the nesting of context providers from that I can tell if its L2 or L3

packages/blade/src/components/SideNav/_decisions/decisions.md
saurabhdaware: yup added in implementation

anuraghazra (approved):

PR #2155, opened 2024-05-06
fix: i18nify-react version to resolve duplicate i18nifyJs copies

tarun-khanna:
## Description
1. i18nify-react version upgraded to resolve duplicate i18nifyJs copies 
2. new unit tests added over Amount component to test output in diff locale state set by i18nify-react

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Amount/__tests__/Amount.web.test.tsx
```diff
@@ -194,4 +204,19 @@ describe('<Amount />', () => {
...
     });
   });
+
+  AMOUNT_SUFFIX_TEST_SET.forEach((item) => {
+    it(`should render ${item.output} in Amount for value:${item.value} & suffix:${item.suffix}`, () => {
+      const { getByTestId } = renderWithTheme(
+        <I18nAmountWrapper
+          value={item.value}
+          suffix={item.suffix}
+          testID="amount-test"
```
kamleshchandnani, line 214: why are we using test id? we shall use get by text isn't it? that's how we validate how will the text be read by our users

packages/blade/src/components/Amount/__tests__/mock.ts
```diff
@@ -0,0 +1,15 @@
+import type { AmountProps } from '../Amount';
+
+export const AMOUNT_SUFFIX_TEST_SET: {
+  value: number;
+  suffix: AmountProps['suffix'];
+  output: string;
+  locale?: string;
+}[] = [
+  { value: 1000000.22, suffix: 'humanize', output: '1 Mio.', locale: 'de-DE' },
```
kamleshchandnani, line 9: shouldn't we also have currency symbols in these mocks?

anuraghazra (dismissed):

packages/blade/src/components/Amount/__tests__/Amount.web.test.tsx
tarun-khanna: `getByText` can't be used here as the text is divided across multiple elements. For eg,
```
<p>$</p>
<p>1000</p>
```
Therefore, getByTestId was used from parent to find if there's any text content that matches the output

packages/blade/src/components/Amount/__tests__/mock.ts
tarun-khanna: Currency symbol was little redundant here as formatting was the core part being driven by locale and was being checked already. But I've still included it now in the test case suite.

packages/blade/src/components/Amount/__tests__/mock.ts
```diff
@@ -0,0 +1,15 @@
+import type { AmountProps } from '../Amount';
+
+export const AMOUNT_SUFFIX_TEST_SET: {
+  value: number;
+  suffix: AmountProps['suffix'];
+  output: string;
+  locale?: string;
+}[] = [
+  { value: 1000000.22, suffix: 'humanize', output: '1 Mio.₹', locale: 'de-DE' },
+  { value: 1000000.0, suffix: 'decimals', output: '₹1,000,000.00', locale: 'en-US' },
```
anuraghazra, line 10: Add the en-IN case also since it was part of the reported bug.

packages/blade/src/components/Amount/__tests__/mock.ts
```diff
@@ -0,0 +1,15 @@
...
+  { value: 10000000, suffix: 'none', output: '₹10,000,000', locale: 'en-US' },
+  { value: 2.07, suffix: 'decimals', output: '2,07₹', locale: 'fr-FR' },
+  { value: 2.077, suffix: 'decimals', output: '₹2.08' },
```
anuraghazra, line 13: See if it's possible to add a test case for: when the locale isn't set and browser locale is different. Ideally it should auto detect/fallback to browser locale. 

https://stackoverflow.com/questions/52868727/how-to-mock-window-navigator-language-using-jest

packages/blade/src/components/Amount/__tests__/Amount.web.test.tsx
anuraghazra: Either way works. 

getByText(item.output) vs getByTestId("amount.test").toHaveText(item.output)

kamleshchandnani: yes but would prefer to refrain from using test ids as with these tests and RTL we are trying to mimic our end users

packages/blade/src/components/Amount/__tests__/mock.ts
tarun-khanna: @anuraghazra these tests are more related to i18nify's functional logic and have been covered extensively there. 
Check [here](https://github.com/razorpay/i18nify/blob/master/packages/i18nify-js/src/modules/.internal/utils/__tests__/getLocale.test.ts) for more details 

This is not required here IMO.

anuraghazra: Okay, should be fine if you folks have them on your end.

anuraghazra (approved):

packages/blade/src/components/Amount/__tests__/Amount.web.test.tsx
kamleshchandnani: @tarun-khanna differences: https://stackoverflow.com/a/65001155

approving this for now but the above justifies the reason

kamleshchandnani (approved):

Issue #2132, opened 2024-04-19
helpText in Radio component has red color


AnkitaBagale:
helpText in Radio component has red color, it should be muted or subtle.
<img width="577" alt="image" src="https://github.com/razorpay/blade/assets/69760792/2b1845b5-fa71-48c5-a564-09507b0834c3">

```
<RadioGroup
                label=""
                name="beneficiaryInfo"
                size="medium"
                onChange={handleBeneficiaryInfoChange}
                value={formState.beneficiaryInfo}
              >
                <Radio value={beneficiaryInfoMap.details} helpText="Bank details/UPI ID/Wallet no.">
                  Beneficiary details
                </Radio>
                <Radio value={beneficiaryInfoMap.id} helpText="Works for existing contacts only">
                  RazorpayX Fund account ID
                </Radio>
</RadioGroup>
```

anuraghazra:
🤔 Weird this should not happen. 

Can you link to the PR/Code for this?

AnkitaBagale:
@anuraghazra code PR. this is happening in v11 dark mode, working fine in v10
https://github.com/razorpay/x/pull/7658

anuraghazra:
@AnkitaBagale I've raised a fix for this https://github.com/razorpay/blade/pull/2175

Seems like while rebranding we accidentally added isNegative prop to helpText

PR #2175, opened 2024-05-14
fix(blade): radio helptext being negative

anuraghazra:
## Description
fixes: https://github.com/razorpay/blade/issues/2132

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Radio/Radio.tsx
```diff
@@ -74,7 +74,7 @@ const _Radio: React.ForwardRefRenderFunction<BladeElementRef, RadioProps> = (
   const _isDisabled = isDisabled ?? groupProps?.isDisabled;
   const _isRequired = groupProps?.isRequired || groupProps?.necessityIndicator === 'required';
   const name = groupProps?.name;
-  const showHelpText = !hasError && helpText;
+  const showHelpText = helpText;
```
snitin315, line 77: We have similar logic in other components too. Do we need to update all of them?

https://github.com/search?q=repo%3Arazorpay%2Fblade%20showHelpText&type=code

anuraghazra: I've changed on Checkbox/Radio. 

FileUpload I'm not sure because it doesn't have nested items like Radio/RadioGroup. 

Chips me we might need. We can check.

snitin315 (approved):

PR #2160, opened 2024-05-07
fix(DropdownOverlay): use FloatingPortal to portal dropdown outside sticky context

saurabhdaware:
## Description

Fixes https://razorpay.slack.com/archives/C044FVB7UQL/p1714650255488099

## Component Checklist

- [ ] Update Component Status Page
- [ ] Perform Manual Testing in Other Browsers
- [ ] Add KitchenSink Story
- [ ] Add Interaction Tests (if applicable)
- [ ] Add changeset

PR #2187, opened 2024-05-22
fix: Dropdown interaction tests

saurabhdaware:
## Description

Tests started failing after #2160. Fixed them here

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Dropdown/DropdownHeaderFooter.tsx
```diff
@@ -73,21 +73,6 @@ const _DropdownFooter = ({ children, testID }: DropdownFooter): React.ReactEleme
     <BaseBox
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       ref={footerRef as any}
-      {...(isReactNative()
```
anuraghazra, line 76: What's happening here? Why are we deleting this?

saurabhdaware: One of the piece of code that was taking care of not consider footer actions as outside click when focus was used.

No need of this now

anuraghazra (approved):

PR #2191, opened 2024-05-23
chore: display added component data

snitin315:
Show newly added component data in PR.

### DIFF
--- a/packages/blade/scripts/generateBundleDiff.js
+++ b/packages/blade/scripts/generateBundleDiff.js
@@ -21,10 +21,10 @@ const generateBundleDiff = async () => {
 
   // Filter the components that don't have the same size in the base and current bundle
   if (baseBundleSizeStats.length > 0) {
-    bundleDiff = baseBundleSizeStats.filter(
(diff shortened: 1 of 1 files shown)


packages/blade/scripts/generateBundleDiff.js
```diff
@@ -21,10 +21,10 @@ const generateBundleDiff = async () => {
 
   // Filter the components that don't have the same size in the base and current bundle
   if (baseBundleSizeStats.length > 0) {
-    bundleDiff = baseBundleSizeStats.filter(
-      ({ size: baseSize, name: baseName }) =>
-        !currentBundleSizeStats.some(
-          ({ size: currentSize, name: currentName }) =>
+    bundleDiff = currentBundleSizeStats.filter(
```
anuraghazra, line 24: Only var renames?

snitin315: now we are filtering the `currentBundleSizeStats` array, previously it was `baseBundleSizeStats` where DatePicker didn't exist.

PR #2212, opened 2024-05-30
docs(Menu): API Decision

saurabhdaware:
### [✨ Formatted Document ✨](https://github.com/razorpay/blade/blob/menu/api/packages/blade/src/components/Menu/_decisions/decisions.md)


packages/blade/src/components/Menu/_decisions/decisions.md
```diff
@@ -0,0 +1,308 @@
...
+## Design
+
+- [Figma - Menu](https://www.figma.com/design/jubmQL9Z8V7881ayUD95ps/Blade-DSL?node-id=90082-41948&m=dev)
+
+## API
+
+```jsx
+<Menu>
+  {/* Any interactive item which has onClick, onMouseDown exposed */}
+  <Button>Open Menu</Button>
```
anuraghazra, line 18: Let's define what needs to be exposed for a functional Menu Trigger. 

- Events (onClick, onMouseOver)
- Accessibility props (aria-haspopup, aria-controls)
- ref for floating positioning etc etc

packages/blade/src/components/Menu/_decisions/decisions.md
```diff
@@ -0,0 +1,308 @@
...
+
+  /**
+   * Open controlled state
+   */
+  isOpen?: boolean;
+
+  /**
+   * On Menu open change callback
+   */
+  onOpenChange?: (isOpen: boolean) => void;
```
anuraghazra, line 75:
```suggestion
  onOpenChange?: ({ isOpen }: { isOpen: boolean }) => void;
```

For future proofing.

packages/blade/src/components/Menu/_decisions/decisions.md
```diff
@@ -0,0 +1,308 @@
...
+
+## Props
+
+### Menu
+
+```ts
+type MenuProps = {
+  /**
+   * First children is trigger and second children is MenuOverlay
+   **/
```
anuraghazra, line 64: Seems outdated.

packages/blade/src/components/Menu/_decisions/decisions.md
```diff
@@ -0,0 +1,308 @@
...
+
+  /**
+   * disabled state of item
+   */
+  isDisabled?: boolean;
+
+  /**
+   * Color of item. set to negative for dangerous actions like Delete, Remove, etc
+   */
+  color?: Extract<FeedbackColors, 'negative'>;
```
anuraghazra, line 159: This type doesn't make sense? Extracting negative from FeedbackColors means color will always be negative?

packages/blade/src/components/Menu/_decisions/decisions.md
```diff
@@ -0,0 +1,308 @@
...
+</Menu>
+```
+
+### SubMenu
+
+```jsx
+<Menu>
+  <Button>Edit</Button>
+  <MenuOverlay>
+    <MenuItem title="Profile" />
```
anuraghazra, line 203: I remember we discussed around having a wrapper to consolidate the MenuItems in a section which would make it easy for you to add negative margins? 

```
<MenuOverlay>
  // any jsx
  <MenuList>
    <Menu />
  </MenuList>
  // any jsx
</MenuOverlay>
```

packages/blade/src/components/Menu/_decisions/decisions.md
```diff
@@ -0,0 +1,308 @@
...
+## API
+
+```jsx
+<Menu>
+  {/* Any interactive item which has onClick, onMouseDown exposed */}
+  <Button>Open Menu</Button>
+  <MenuOverlay>
+    <MenuHeader title="" />
+    <MenuBody>
+      {/* All of this is a slot */}
```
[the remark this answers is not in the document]
saurabhdaware: yup

packages/blade/src/components/Menu/_decisions/decisions.md
```diff
@@ -0,0 +1,372 @@
...
+#### 3. Two Different Usecases
+
+Normally you would see design-systems having 2 components. 1 for SelectInput, AutoComplete, etc and 2nd for Menu, ContextMenu, ActionMenu, etc
+
+Examples
+
+- [Primer SelectPanel](https://primer.style/components/selectpanel/react/alpha), [Primer ActionMenu](https://primer.style/components/action-menu/react/beta)
+- [Ariakit Select](https://ariakit.org/components/select), [Ariakit Combobox](https://ariakit.org/components/combobox), [Ariakit Menu](https://ariakit.org/components/menu)
+
+  > [!Note]
```
snitin315, line 277:
```suggestion
  > [!NOTE]
```

packages/blade/src/components/Menu/_decisions/decisions.md
```diff
@@ -0,0 +1,372 @@
...
+      <Avatar name="Saurabh Daware" />
+      <Text>Saurabh Daware</Text>
+    </Box>
+    <MenuDivider />
+    <MenuItem title="Undo" onClick={() => console.log('Undo')} />
+    <MenuItem title="Redo" isDisabled />
+    <MenuItem title="Cut" />
+    {/* Sub Menu */}
+    <Menu>
+      <MenuItem title="Copy" />
```
snitin315, line 32: What if we use a button/trigger here? Would it throw an error?

packages/blade/src/components/Menu/_decisions/decisions.md
```diff
@@ -0,0 +1,372 @@
...
+- What do custom components inside DropdownOverlay do? they can't be selected so shouldn't have selection logic code.
+- How do you build more flexible menus with custom items? Dropdown relies on items inside of it to be selectable.
+
+For these reasons, we are going ahead with Menu component which will be for more flexible, clickable menus that are not selectable.
+
+## Is this a breaking change for Dropdown consumers?
+
+**No**. Everything that works currently will continue to work. We will deprecate the DropdownButton, DropdownLink usage and recommend moving to Menu for those usecase but not break the usage of DropdownButton or DropdownLink.
+
+How migration from deprecated API to recommended API will look like-
```
snitin315, line 301: Should we provide a codemod for this migration?

packages/blade/src/components/Menu/_decisions/decisions.md
```diff
@@ -0,0 +1,372 @@
...
+    </Menu>
+  </MenuOverlay>
+</Menu>
+```
+
+### Custom Trigger
+
+```jsx
+<Menu>
+  <CustomComponentTrigger />
```
snitin315, line 221: Can this be used with our tooltip too? [Ref example](https://ariakit.org/examples/menu-tooltip)

packages/blade/src/components/Menu/_decisions/decisions.md
saurabhdaware: Nope. It will use that button is submenu trigger. Menus are flexible so there might be usecase of some custom component triggering submenu

packages/blade/src/components/Menu/_decisions/decisions.md
saurabhdaware: We're not breaking it right now. If we decide to remove the API, we can provide a codemod then

packages/blade/src/components/Menu/_decisions/decisions.md
saurabhdaware: Good catch. Fixed it in office. It can be used with tooltip now by wrapping trigger with Tooltip component

packages/blade/src/components/Menu/_decisions/decisions.md
saurabhdaware: Added it to doc

chaitanyadeorukhkar (approved):

snitin315 (approved):

packages/blade/src/components/Menu/_decisions/decisions.md
saurabhdaware: Discussed in office that day. MenuList can't isolate the list navigation in Menu's case and for negative margins, I added MenuDivider component instead so it handles the styling otherwise it was too much of styling just to get end-to-end divider.

Also exposed bleed prop on Divider but we need to know the padding for us to add correct negative margins and position absolute was not working as expected because then it depends on parent position relative

PR #2223, opened 2024-06-10
fix: datepicker zIndex issue and expose zIndex prop

### LINKED ISSUE #2214: Date Picker component not opening
There's a weird issue in date picker. calendar does not open on first load. Issue goes away if we refresh the page once and does not occur again during entire session. 

But issue happens again for next session.

Marking it as P0, as breaks the user experience.

https://github.com/razorpay/blade/assets/69760792/20e4d99d-10d4-4c72-abdf-f623e42d5c39

anuraghazra:
## Description

closes #2214 

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/DatePicker/DatePicker.web.tsx
+++ b/packages/blade/src/components/DatePicker/DatePicker.web.tsx
@@ -33,6 +33,7 @@ import { logger } from '~utils/logger';
 import type { StyledPropsBlade } from '~components/Box/styledProps';
 import { getStyledProps } from '~components/Box/styledProps';
 import { metaAttribute, MetaConstants } from '~utils/metaAttribute';
+import { componentZIndices } from '~utils/componentZIndices';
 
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/DatePicker/DatePicker.web.tsx
```diff
@@ -60,6 +61,7 @@ const DatePicker = <Type extends DateSelectionType = 'single'>({
   defaultPicker = 'day',
   picker,
   onPickerChange,
+  zIndex = componentZIndices.popover,
```
saurabhdaware, line 64: shouldn't we add different zIndex here?

saurabhdaware: Like its not popover no?

anuraghazra: > Like its not popover no?

Why not? It is a popover.

PR #2229, opened 2024-06-13
feat(Menu): add Menu implementation

saurabhdaware:
## Description

Menu Implimentation.

For other PRs, check description of #2244 

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Avatar/Avatar.web.tsx
```diff
@@ -20,24 +21,38 @@ const getInitials = (name: string): string => {
...
-  // Image Props
-  src,
-  alt,
-  srcSet,
-  crossOrigin,
-  referrerPolicy,
-  testID,
-  ...styledProps
-}: AvatarProps): ReactElement => {
+const _Avatar: React.ForwardRefRenderFunction<BladeElementRef, AvatarProps> = (
```
chaitanyadeorukhkar, line 24: Does this mean that we have to add these handlers for every component that could be a dropdown trigger? Even for custom components that users may have?

packages/blade/src/components/BaseMenu/BaseMenuItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+    marginTop: makeSize(props.theme.spacing[1]),
+    marginBottom: makeSize(props.theme.spacing[1]),
+    textDecoration: 'none',
+    cursor: 'pointer',
+    width: '100%',
+    [`@media ${getMediaQuery({ min: props.theme.breakpoints.m })}`]: {
+      padding: makeSize(props.theme.spacing[3]),
+    },
+    '&:focus-visible': getFocusRingStyles({ theme: props.theme }),
+    '&:hover:not([aria-disabled=true]), &.has-submenu-open': {
```
chaitanyadeorukhkar, line 31: Why are we checking for aria-disabled? Lets instead check for the disabled prop directly?

packages/blade/src/components/BaseMenu/BaseMenuItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+      backgroundColor:
+        props.color === 'negative'
+          ? props.theme.colors.interactive.background.negative.faded
+          : props.theme.colors.interactive.background.gray.default,
+    },
+    '&[aria-selected=true]': {
```
chaitanyadeorukhkar, line 37: Lets check for selected prop directly?

packages/blade/src/components/BaseMenu/BaseMenuItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+          {selectionType === 'multiple' ? (
+            // Adding aria-hidden because the listbox item in multiselect in itself explains the behaviour so announcing checkbox is unneccesary and just a nice UI tweak for us
+            <BaseBox
+              pointerEvents="none"
+              paddingRight="spacing.2"
+              {...makeAccessible({
+                hidden: true,
+              })}
+            >
+              <Checkbox isChecked={isSelected} tabIndex={-1} isDisabled={isDisabled}>
```
chaitanyadeorukhkar, line 122: Why do we need a checkbox?

packages/blade/src/components/BaseMenu/BaseMenuItem.tsx
```diff
@@ -0,0 +1,168 @@
...
+} as const;
+
+const menuItemDescriptionColor = {
+  default: 'interactive.text.gray.muted',
+  disabled: 'interactive.text.gray.disabled',
+} as const;
+
+const _BaseMenuItem: React.ForwardRefRenderFunction<BladeElementRef, BaseMenuItemProps> = (
+  {
+    title,
```
chaitanyadeorukhkar, line 76: On figma I can see isLoading state as well. Lets add that?

packages/blade/src/components/BaseMenu/index.ts
chaitanyadeorukhkar, line 1: This file is empty? No re-exports?

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -43,6 +43,32 @@ type ButtonCommonProps = {
...
+   */
+  'aria-haspopup'?: 'menu';
+  /**
+   * It is exposed for internal usage with menu.
+   *
+   * @private
+   */
+  role?: 'menuitem';
+  tabIndex?: BaseButtonProps['tabIndex'];
+  id?: BaseButtonProps['id'];
```
chaitanyadeorukhkar, line 71: People will target the button by id. Lets not expose this please 🙈

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -43,6 +43,32 @@ type ButtonCommonProps = {
...
+   * @private
+   */
+  'aria-haspopup'?: 'menu';
+  /**
+   * It is exposed for internal usage with menu.
+   *
+   * @private
+   */
+  role?: 'menuitem';
+  tabIndex?: BaseButtonProps['tabIndex'];
```
chaitanyadeorukhkar, line 70: Our button will get very polluted with all these props. Lets just use BaseButton for internal usage?

packages/blade/src/components/Menu/VisualSubComponents/MenuDivider.web.tsx
```diff
@@ -0,0 +1,8 @@
+import type { StyledPropsBlade } from '~components/Box/styledProps';
+import { Divider } from '~components/Divider';
+
+const MenuDivider = (styledProps: StyledPropsBlade): React.ReactElement => {
+  return <Divider marginLeft="-8px" marginY="spacing.1" marginRight="-8px" {...styledProps} />;
```
chaitanyadeorukhkar, line 5: use spacing tokens and add a negation before it?

packages/blade/src/components/Menu/Menu.web.tsx
chaitanyadeorukhkar, line 1: Add .native files as well?

packages/blade/src/components/Avatar/Avatar.web.tsx
saurabhdaware: yup. So custom components are likely to have it already since they won't be restrictive but for our internal components we do need them.

We have already added them in Link, Button for Tooltip so just needed to add in Avatar

packages/blade/src/components/BaseMenu/BaseMenuItem.tsx
saurabhdaware: Been preferring the CSS selectors over JS checks. One check won't make much difference but seems like a good practice.

packages/blade/src/components/BaseMenu/BaseMenuItem.tsx
saurabhdaware: https://github.com/razorpay/blade/pull/2229#discussion_r1639283892

packages/blade/src/components/BaseMenu/BaseMenuItem.tsx
saurabhdaware: Been trying to have this component that covers both MenuItem and ActionList styling. Will use this in ActionListItem as well where we need checkbox

packages/blade/src/components/BaseMenu/index.ts
saurabhdaware: ops missed it

packages/blade/src/components/Button/Button/Button.tsx
saurabhdaware: Its not really internal usage. The consumers will import and add Button on their end

saurabhdaware: They don't have to explicitly these props because we clone component and pass them internally. But the component that consumers add is still Button itself

packages/blade/src/components/BaseMenu/index.ts
saurabhdaware: fixed

packages/blade/src/components/BaseMenu/BaseMenuItem.tsx
saurabhdaware: Did the refactor to use this in ActionListItem as well #2232

packages/blade/src/components/BaseMenu/BaseMenuItem.tsx
saurabhdaware: Didn't see loading state anywhere in figma 🤔

packages/blade/src/components/Button/Button/Button.tsx
saurabhdaware: true. The floating UI does pass an id to trigger components. But things seem to work fine without that also so maybe we can skip this 🤔

packages/blade/src/components/Menu/VisualSubComponents/MenuDivider.web.tsx
saurabhdaware: done

packages/blade/src/components/Menu/Menu.web.tsx
saurabhdaware: Will add with documentation and tests

packages/blade/src/components/Button/Button/Button.tsx
saurabhdaware: Removed id

packages/blade/src/components/Menu/Menu.web.tsx
saurabhdaware: Added in documentation PR #2236

packages/blade/src/components/Avatar/Avatar.web.tsx
```diff
@@ -76,14 +103,20 @@ const _Avatar = ({
...
+        </AvatarButton>
+      );
     }
 
     return <AvatarButton {...commonButtonProps} icon={icon ?? UserIcon} />;
   };
 
   return (
     <StyledAvatar
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
anuraghazra, line 119: unused

packages/blade/src/components/Avatar/Avatar.web.tsx
```diff
@@ -57,13 +72,25 @@ const _Avatar = ({
...
+    onTouchEnd,
   };
 
   const getChildrenToRender = (): React.ReactElement => {
     if (src) {
       return (
         <AvatarButton
           {...commonButtonProps}
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          ref={ref as any}
```
anuraghazra, line 93: try `ref as never`

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -104,13 +133,17 @@ const _Button: React.ForwardRefRenderFunction<BladeElementRef, ButtonProps> = (
...
       {...getStyledProps(rest)}
+      id={id}
       ref={ref}
       href={href}
       target={target}
       rel={rel}
       accessibilityProps={{
         label: accessibilityLabel,
         describedBy: rest['aria-describedby'],
+        expanded: rest['aria-expanded'] === 'true',
```
anuraghazra, line 144: This will condition will not work if consumer passes the prop like an actual boolean 

```
<Button aria-expanded />
<Button aria-expanded={true} />
```

Since the type of aria props are `Booleanish`

```ts
type Booleanish = boolean | "true" | "false";
```

anuraghazra: I think you can just pass it has is without checking for `=== true` part that should work.

packages/blade/src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+  argTypes: {
+    ...getStyledPropsArgTypes(),
+    trigger: {
+      table: {
+        disable: true,
+      },
+    },
+  },
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+} as Meta<any>;
```
anuraghazra, line 26: Pass prop to Meta

packages/blade/src/components/Menu/VisualSubComponents/MenuDivider.web.tsx
```diff
@@ -0,0 +1,11 @@
+import { getDividerMarginTokens } from '../tokens';
+import type { StyledPropsBlade } from '~components/Box/styledProps';
+import { Divider } from '~components/Divider';
+import { useTheme } from '~utils';
+
+const MenuDivider = (styledProps: StyledPropsBlade): React.ReactElement => {
```
anuraghazra, line 6: Why do we need a MenuDivider component? Can we use the Divider component directly?

packages/blade/src/components/Menu/useMenu.ts
```diff
@@ -0,0 +1,183 @@
...
+    dismiss,
+    listNavigation,
+  ]);
+
+  const { isMounted, styles: floatingTransitionStyles } = useTransitionStyles(context, {
+    duration: theme.motion.duration.quick,
+    initial: ({ side }) => {
+      let transform: string;
+      const transitionOffset = makeSize(OVERLAY_TRANSITION_OFFSET);
+      switch (side) {
```
anuraghazra, line 105: nit: use ifelse or object mapper. 

Switch case is making the code bit longer than it needs to be.

packages/blade/src/components/Menu/types.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   * disabled state of item
+   */
+  isDisabled?: boolean;
+
+  /**
+   * Color of item. set to negative for dangerous actions like Delete, Remove, etc
+   */
+  color?: Extract<FeedbackColors, 'negative'>;
```
anuraghazra, line 104: color will always have to be negative with this type.

packages/blade/src/components/Button/Button/Button.tsx
```diff
@@ -43,6 +43,32 @@ type ButtonCommonProps = {
...
+   *
+   * @private
+   */
+  'aria-haspopup'?: 'menu';
+  /**
+   * It is exposed for internal usage with menu.
+   *
+   * @private
+   */
+  role?: 'menuitem';
```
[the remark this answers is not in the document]
saurabhdaware: I was taking menuitem or undefined but I think we can just accept all roles as well 🤔

packages/blade/src/components/Button/Button/Button.tsx
saurabhdaware: oh yes I changed this in the other stacked PR so its resolved

packages/blade/src/components/Menu/types.ts
saurabhdaware: It can be negative or undefined

anuraghazra: What is undefined mapping to? Should it be neutral or negative in that case then?

packages/blade/src/components/Menu/Menu.stories.tsx
saurabhdaware: will fix it in docs PR where  I wrote documentation

packages/blade/src/components/Menu/VisualSubComponents/MenuDivider.web.tsx
saurabhdaware: No we have to add exact negative margin for end-to-end divider

packages/blade/src/components/Menu/useMenu.ts
saurabhdaware: makes sense. done in upcoming commit

packages/blade/src/components/Menu/types.ts
saurabhdaware: It maps to one of the surface.text.* color so its not neutral either

anuraghazra (approved):

PR #1144, opened 2023-04-27
feat: add type module [test] [ignore]

saurabhdaware:
This makes things work on Vite

PR #2231, opened 2024-06-16
feat: add type module to package.json

saurabhdaware:
Fork of #1144 with latest master

Want to use blade in my React Nexus demo so putting some changes here to make it work with vite


saurabhdaware:
version: https://pkg.csb.dev/razorpay/blade/commit/e5a681ae/@razorpay/blade

snitin315:
>Want to use blade in my React Nexus demo so putting some changes here to make it work with vite

Why make your demo complicated 😛?

saurabhdaware:
Was able to do this without adding type module to blade by marking blade as noExternal in vite config. Demo - https://github.com/saurabhdaware/react-nexus-final-final-performance-demo

PR #2236, opened 2024-06-18
docs(Menu): add documentation and tests

saurabhdaware:
## Description

- Documentation
- Tests

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


snitin315 (commented):
One of the test is failing. Rest LGTM.
Screenshot 2024-06-21 at 7 49 12 AM

saurabhdaware:
Removed the flaky test

packages/blade/src/components/Menu/Menu.web.tsx
```diff
@@ -90,21 +93,23 @@ const MenuTree = ({
...
-                modal={false}
-                initialFocus={-1}
-                returnFocus={!isNested}
-              >
-                {overlayWithFloatingProps}
-              </FloatingFocusManager>
-            </FloatingPortal>
-          )}
-        </FloatingList>
+        <BaseBox {...metaAttribute({ name: MetaConstants.Menu, testID })}>
```
anuraghazra, line 96: Adding another wrapper may cause issues with integration with other inline components & introduce CSS issues since now in the final html the `trigger` will not be present as how it's rendered in JSX but rather with a Box wrapper. 

We can probably add the Menu meta constants to the Overlay since anyways the trigger itself will not be the "menu" but rather can be any component (eg: button)

packages/blade/src/components/Menu/docs/CustomMenu.web.tsx
```diff
@@ -0,0 +1,143 @@
...
+    color: props.theme.colors.surface.text.gray.normal,
+    backgroundColor: props.isTransparent
+      ? props.theme.colors.transparent
+      : props.theme.colors.interactive.background.gray.faded,
+    borderRadius: props.isTransparent ? '0px' : makeSize(props.theme.border.radius.medium),
+    borderBottomWidth: makeBorderSize(props.theme.border.width.thick),
+    borderBottomStyle: 'solid',
+    borderBottomColor: props.theme.colors.transparent,
+    cursor: 'pointer',
+    '&:hover, &.menu-open': {
```
anuraghazra, line 32: How will consumer know about this internal class called `.menu-open`?? 

Can we use aria attributes (eg: aria-expanded or something) to target the open state instead since that will be more "generic" and standard way to do it. 

Other libraries use the similar method, and some like radix also expose data attributes for the states for styling, eg `[data-open]`

saurabhdaware: Good idea. Let me check

saurabhdaware: Nice suggestion! Removed a lot of code

anuraghazra (approved):

saurabhdaware:
Some strange things happening with tests and ts. Probably because its not in sync with master. Will fix them after merging the other PRs to integration branch

snitin315 (approved):

PR #2237, opened 2024-06-18
docs: add TopNav API

anuraghazra:
## [View Formatted](https://github.com/razorpay/blade/blob/anu/topnav-api/packages/blade/src/components/TopNav/_decisions/decisions.md)


packages/blade/src/components/NavigationBar/_decisions/decisions.md
```diff
@@ -0,0 +1,261 @@
...
+- [HeaderNav - BaseWeb DS / Uber](https://baseweb.design/components/header-navigation/#basic-link1)
+- [TopNav - Spectrum / Adobe](https://opensource.adobe.com/spectrum-web-components/components/top-nav/)
+
+## Open Questions
+
+### Dev
+
+1. How will the Menu component integrate with the BottomSheet?
+2. How will the TabNav component handle the chevron icon for the Menu component?
+3. Should we call this component `TopNav`/`TopNavigationBar` instead of `NavigationBar`?
```
saurabhdaware, line 256: TopNav might help if we're also going to create BottomNav.

If we can have BottomNav inside of this itself, then `NavBar` might be good?

packages/blade/src/components/NavigationBar/_decisions/decisions.md
```diff
@@ -0,0 +1,261 @@
+# NavigationBar - One Top Navigation Bar
+
+The navigation bar is positioned at the top of the screen that provides quick access to different products, search & user profile. 
+
+<img src="./top-nav-bar.png" width="100%" />
```
saurabhdaware, line 5: What is that heavy vertical border in this image next to Razorpay logo?

packages/blade/src/components/NavigationBar/_decisions/decisions.md
```diff
@@ -0,0 +1,261 @@
...
+  </NavigationBarBrand>
+  <NavigationBarContent>
+    <TabNav>
+      <TabNavItem as={RouterLink} isSelected href="/home" icon={HomeIcon} />
+      <TabNavItem as={RouterLink} href="/money">Money</TabNavItem>
+      <TabNavItem as={RouterLink} href="/payroll">Payroll</TabNavItem>
+      <Menu>
+        <TabNavItem trailingIcon={ChevronDown}>Payroll</TabNavItem>
+        <MenuOverlay>
+          { /* Other Menu Items */ }
```
saurabhdaware, line 43: Do we not want control on how the items inside this menu should look like?

packages/blade/src/components/NavigationBar/_decisions/decisions.md
```diff
@@ -0,0 +1,261 @@
...
+
+```jsx
+const WithMenu = () => {
+  const [isOpen, setOpen] = React.useState(false);
+  const [selected, setSelected] = React.useState<string | null>(null);
+
+  return (
+    <TabNav>
+      <TabNavItem as={RouterLink} isSelected href="/home" icon={HomeIcon} />
+      <Menu interactionType="hover">
```
saurabhdaware, line 171:
```suggestion
      <Menu openInteraction="hover">
```

You don't read my APIs. -1 for Anurag

packages/blade/src/components/NavigationBar/_decisions/decisions.md
```diff
@@ -0,0 +1,261 @@
...
+    )
+  }
+
+  return (
+    // ... desktop navigation bar jsx
+  )
+}
+```
+
+## Accessibility
```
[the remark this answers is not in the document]
anuraghazra: Yes will add that.

packages/blade/src/components/NavigationBar/_decisions/decisions.md
anuraghazra: Yeah, because that would require a new component to be built and exported from blade, rather it will be better if we let consumer handle this with Menu. Product requirement of that "others" menu can also change in future so best to keep it a bit flexible.

packages/blade/src/components/NavigationBar/_decisions/decisions.md
```diff
@@ -0,0 +1,261 @@
...
+
+### NavigationBarContent
+
+The container for the main navigation items.
+
+### NavigationBarActions
+
+The container for the right-aligned actions like search, buttons, and user profile.
+
+### TabNav
```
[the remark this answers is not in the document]
anuraghazra: > When would they use TabNav outside of TopNav?

Some usecases similar to [github](https://primer.style/components/tab-nav).

But it may or may not be used. I don't see a real benefit also on preventing them to do so.

packages/blade/src/components/NavigationBar/_decisions/decisions.md
anuraghazra: Bug while exporting the figma image.

packages/blade/src/components/NavigationBar/_decisions/decisions.md
anuraghazra: Yeah BottomNav will be a thing, so maybe TopNav is a better option. Can get opinion of others too.

packages/blade/src/components/TopNav/_decisions/decisions.md
```diff
@@ -0,0 +1,291 @@
...
+**Q:** So where will the product navigation items go? 
+**Ans:** The product navigation items will be moved to the bottom of the screen, but this will be part of a separate component called `BottomNavigation` which we will ship in phase 2.
+
+Mobile Navbar example:
+
+```jsx
+const Dashboard = () => {
+  const { isMobile } = useBreakpoint();
+
+  if (isMobile) {
```
sudhakar-selva, line 229: suggestion: can we toggle the layout based on css media queries instead? I understand this is common pattern and offers some benefits(especially for CSR only apps), however this may not work well for SSR applications. 

Given majority of our apps uses CSR, this may not be a concern for now, just calling it out.

anuraghazra: This is a contrived example for demonstration purposes, so I've just used isMobile prop. 


And this will be handle on consumer side, they can choose to render the layout however they like. via isMobile or via the `Box` responsive props. And whatever will be controlled from blade internally (eg: size of left nav) will use responsive media props only.

anuraghazra: Will probably update this example just in case other people copy paste it from here and use it as a pattern.

packages/blade/src/components/TopNav/_decisions/decisions.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Open Questions
+
+### Dev
+
+1. How will the Menu component integrate with the BottomSheet?
+2. How will the TabNav component handle the chevron icon for the Menu component?
+3. Should we call this component `TopNav`/`TopNavigationBar` instead of `NavigationBar`?
+   1. We will call it `TopNav` since we will also have a `BottomNav` in the future.
+4. For the responsive behavior, should we just hide the `TopNavBrand` and `TopNavContent` with CSS media queries? instead of asking consumers to do conditional rendering?
```
snitin315, line 287: We should handle the responsiveness ourselves.

anuraghazra: That's a bit tricker to do, because of the nature of the TopNav. 

The TopNav is quite flexible and mostly provides a skeleton of UI layout where consumers can add logo, content, footer actions etc. (if you remove the TabNav)

For example, the consumer is in charge of rendering the SearchInput in the footer so we cannot remove it from our end, consumer will use the responsive styled props to remove it on mobile devices. 

Tho few things that we do: 

1. We hide TopNavBrand internally on mobile because there is not enough space at top to show logo/actions + content all at once
2. We adjust padding/margins accordingly to ensure in mobile all the footer actions/content align properly. 
3. Expose responsive props in TopNav components so consumers can dynamically add responsive styles.

saurabhdaware (approved):

PR #2272, opened 2024-06-26
test: add topnav tests

anuraghazra:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/TopNav/__tests__/TabNav.test.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const { getByRole } = within(canvasElement);
+  const homeTab = getByRole('link', { name: 'Home' });
+  const payrollTab = getByRole('link', { name: 'Payroll' });
+  const paymentsTab = getByRole('link', { name: 'Payments' });
+  await expect(homeTab).toHaveAttribute('href', '/home');
+  await expect(payrollTab).toHaveAttribute('href', '/payroll');
+  await expect(paymentsTab).toHaveAttribute('href', '/payments');
+};
+
+export const TestOverflow: StoryFn<typeof TabNav> = (): React.ReactElement => {
```
saurabhdaware, line 32: no test to check whether clicking item makes it active or not?

saurabhdaware (approved):

packages/blade/src/components/TopNav/__tests__/TabNav.test.stories.tsx
anuraghazra: The selection logic doesn't really work depending on "click" but rather it works same as SideNavLink, isActive gets selected based on page URL. 

The tab itself doesn't do any state management.

snitin315 (approved):
LGTM

Issue #2271, opened 2024-06-26
Table: Checkbox click is showing incorrect selction


AnkitaBagale:
onSelectionChange is called with incorrect values on page change.

https://stackblitz.com/edit/mytgza-efebsi?file=App.tsx

https://github.com/razorpay/blade/assets/69760792/cdd9ac97-e456-4902-963a-1ad57a2267d8

snitin315:
@AnkitaBagale It's working as expected in the [Basic Table story](http://localhost:9009/?path=/story/components-table--table&args=selectionType:multiple). I believe there is some other issue.

Also `x items selected` is the default selected title so you can just remove the `selectedItem` logic in your example.

AnkitaBagale:
@snitin315 
If I remove selectedItem logic, I won't be able to access selectedItems ids.

anuraghazra:
This is fixed in #2282 

For server side paginated data or partial data it is now recommended to use the `selectedIds` value exposed from `onSelectionChange` function. 

```ts
<Table onSelectionChange={({ selectedIds }) => {
  setSelectedIds(selectedIds)
}} />
```

PR #2282, opened 2024-06-28
fix: table multiselect with pagination bug

anuraghazra:
## Description

Fixes: https://github.com/razorpay/blade/issues/2271

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Table/types.ts
+++ b/packages/blade/src/components/Table/types.ts
@@ -86,7 +86,19 @@ type TableProps<Item> = {
     values,
     selectedIds,
   }: {
+    /**
+     * Note: on server side paginated data, this prop will only contain the selected rows on the current page.
+     *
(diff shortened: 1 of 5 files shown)


packages/blade/src/components/Table/types.ts
```diff
@@ -82,7 +82,13 @@ type TableProps<Item> = {
    * The onSelectionChange prop is a function that is called when the selection changes.
    * The function is called with an object that has a values property that is an array of the selected rows.
    **/
-  onSelectionChange?: ({ values }: { values: TableNode<Item>[] }) => void;
+  onSelectionChange?: ({
+    values,
+    selectedIds,
```
chaitanyadeorukhkar, line 87: Why do we need both?

anuraghazra: It'll be a breaking change if we remove the values

chaitanyadeorukhkar (approved):

Issue #2273, opened 2024-06-27
Table row selected state


AnkitaBagale:
Issue: Table row when clicked, I want to open details view but the row checkbox should not get selected. and Row should be in selected state as details view is open.

> Row should be in selected state as details view is open.

Can we add isSelected prop on row?

> the row checkbox should not get selected

Can we expose events in onClick handlers so that consumer can stop propagation if needed, or is there any way we can provide this experience?

Existing table:
![image](https://github.com/razorpay/blade/assets/69760792/9535fbca-2b45-495b-90ac-8e678680f993)

Blade table:

![image (1)](https://github.com/razorpay/blade/assets/69760792/8a67191a-faeb-440c-bdc2-740f729fc35c)

anuraghazra:
This is fixed in #2283 

We've added a new prop called `multiSelectTrigger` which you can set to `checkbox` to achieve the desired behaviour.

PR #2283, opened 2024-06-28
feat: add multiselection-mode prop 

anuraghazra:
## Description

Enables rows to be clickable as a whole or only on the checkbox.

Fixes: https://github.com/razorpay/blade/issues/2273

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Table/types.ts
+++ b/packages/blade/src/components/Table/types.ts
@@ -73,6 +73,11 @@ type TableProps<Item> = {
    */
   data: TableData<Item>;
   /**
+   * Selection mode determines how the table rows can be selected.
+   * @default 'row'
+   **/
+  multiSelectionMode?: 'checkbox' | 'row';
+  /**
    * The selectionType prop determines the type of selection that is allowed on the table.
    * The selectionType prop can be 'none', 'single' or 'multiple'.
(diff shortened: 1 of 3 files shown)


packages/blade/src/components/Table/types.ts
```diff
@@ -72,6 +72,11 @@ type TableProps<Item> = {
    * The object should have an id property that is a unique identifier for the row.
    */
   data: TableData<Item>;
+  /**
+   * Selection mode determines how the table rows can be selected.
+   * @default 'row'
+   **/
+  multiSelectionMode?: 'checkbox' | 'row';
```
chaitanyadeorukhkar, line 79: Should we call this `multiSelectTrigger?: 'checkbox' | 'row'`

chaitanyadeorukhkar (approved):
Suggested prop name change. Otherwise LGTM

packages/blade/src/components/Table/types.ts
anuraghazra: Sounds good to me.

chaitanyadeorukhkar (dismissed):

chaitanyadeorukhkar (approved):

PR #2295, opened 2024-07-09
feat(TableEditableCell): refactor TableEditableCell and add TableEditableDropdownCell

chaitanyadeorukhkar:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


chaitanyadeorukhkar:
Getting some error on the story @saurabhdaware 
Screenshot 2024-07-12 at 3 38 38 PM

chaitanyadeorukhkar:
@saurabhdaware you also need to hide the error/success/help text on table. Similar to how we hide it for input
Screenshot 2024-07-12 at 3 44 20 PM

chaitanyadeorukhkar:
Alignment gets weird for the last element if you squeeze the UI for some reason
Screenshot 2024-07-12 at 3 46 45 PM

saurabhdaware:
Lol I messed up the `isInsideTable` check. While testing I had defined it true explicitly so missed testing the check itself. Fixing it.

packages/blade/src/components/Table/index.ts
```diff
@@ -5,3 +5,4 @@ export * from './TableFooter';
 export * from './TablePagination';
 export * from './TableToolbar';
 export * from './types';
+export { TableEditableCell } from './TableEditableCell';
```
chaitanyadeorukhkar, line 8: We also need to export TableEditableDropdownCell

.changeset/tame-stingrays-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
```
chaitanyadeorukhkar, line 2:
```suggestion
"@razorpay/blade": minor
```

packages/blade/src/components/Input/DropdownInputTriggers/AutoComplete.stories.tsx
```diff
@@ -274,7 +274,7 @@ export const InternalAutoCompleteUncontrolled = (): React.ReactElement => {
   return (
     <Box maxWidth="200px">
       <Dropdown selectionType="multiple">
-        <AutoComplete maxRows="single" label="City" />
+        <SelectInput maxRows="single" label="City" size="large" />
```
chaitanyadeorukhkar, line 277: Using select input in autcomplete story?

packages/blade/src/components/Table/TableEditableCell.web.tsx
```diff
@@ -0,0 +1,174 @@
...
+      </StyledEditableCell>
+    </TableEditableCellContext.Provider>
+  );
+};
+
+const TableEditableCell = assignWithoutSideEffects(_TableEditableCell, {
+  componentId: ComponentIds.TableEditableCell,
+});
+
+export { TableEditableCell, TableEditableDropdownCell };
```
chaitanyadeorukhkar, line 174: Since we are keeping them in the same place, won't this cause a problem with bundle size? If I am not using dropdown cell but only need editable cell, this would end up importing dependencies for dropdown as well, right?

packages/blade/src/components/Table/__tests__/Table.web.test.tsx
```diff
@@ -2,11 +2,12 @@ import userEvent from '@testing-library/user-event';
 import { useState } from 'react';
 import { fireEvent, waitFor } from '@testing-library/react';
 import { Table } from '../Table';
-import { TableBody, TableCell, TableEditableCell, TableRow } from '../TableBody';
+import { TableBody, TableCell, TableRow } from '../TableBody';
 import { TableFooter, TableFooterCell, TableFooterRow } from '../TableFooter';
 import { TableHeader, TableHeaderCell, TableHeaderRow } from '../TableHeader';
 import { TableToolbar } from '../TableToolbar';
 import { TablePagination } from '../TablePagination';
+import { TableEditableCell } from '../TableEditableCell.web';
```
chaitanyadeorukhkar, line 10:
```suggestion
import { TableEditableCell } from '../TableEditableCell';
```

packages/blade/src/components/Table/TableBody.web.tsx
```diff
@@ -152,7 +147,7 @@ const TableBody = assignWithoutSideEffects(_TableBody, {
   componentId: ComponentIds.TableBody,
 });
 
-const StyledCell = styled(Cell)<{
+export const StyledCell = styled(Cell)<{
```
chaitanyadeorukhkar, line 150: This will expose StyledCell and CellWrapper to our consumers too since we are exporting from here and also doing `export * from './TableBody' in index.ts

packages/blade/src/components/Table/index.ts
```diff
@@ -5,3 +5,4 @@ export * from './TableFooter';
 export * from './TablePagination';
 export * from './TableToolbar';
 export * from './types';
+export { TableEditableCell, TableEditableDropdownCell } from './TableEditableCell';
```
chaitanyadeorukhkar, line 8: We should do named exports for the above exports too. I think we may be exposing a lot of unneeded internal components too like CellWrapper and StyledCell

packages/blade/src/components/Table/TableEditableCell.web.tsx
saurabhdaware: Nope the webpack will follow the imports and usage so it won't import things that are not used in that component

PR #2303, opened 2024-07-10
fix: table footer background color

### LINKED ISSUE #2279: For table (with paginationType="server") footer, background color is missing.
For table (with paginationType="server") footer, background color is missing.

Actual - 
image

Expected - 
image

Sumitmaithani:
## Description
Fixes #2279 

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Table/TablePagination.web.tsx
+++ b/packages/blade/src/components/Table/TablePagination.web.tsx
@@ -141,7 +141,6 @@ const _TablePagination = ({
   currentPage: controlledCurrentPage,
   onPageChange,
   onPageSizeChange,
-  backgroundColor = tablePagination.backgroundColor,
   defaultPageSize = tablePagination.defaultPageSize,
   showPageSizePicker = true,
   showPageNumberSelector = false,
@@ -156,6 +155,7 @@ const _TablePagination = ({
     totalItems,
     setPaginationRowSize,
     setPaginationType,
+    backgroundColor,
   } = useTableContext();
   const [currentPageSize, setCurrentPageSize] = React.useState<number>(defaultPageSize);
   const [currentPage, setCurrentPage] = React.useState<number>(
--- a/packages/blade/src/components/Table/TablePagination.web.tsx
+++ b/packages/blade/src/components/Table/TablePagination.web.tsx
@@ -141,6 +141,7 @@ const _TablePagination = ({
   currentPage: controlledCurrentPage,
   onPageChange,
   onPageSizeChange,
+  backgroundColor = tablePagination.backgroundColor,
   defaultPageSize = tablePagination.defaultPageSize,
   showPageSizePicker = true,
   showPageNumberSelector = false,
@@ -242,7 +243,7 @@ const _TablePagination = ({
       display="flex"
       flexDirection="row"
       padding={tablePagination.padding}
-      backgroundColor={tablePagination.backgroundColor}
+      backgroundColor={backgroundColor}
     >
       {showLabel && !onMobile && (
         <BaseBox display="flex" justifyContent="center" alignItems="center">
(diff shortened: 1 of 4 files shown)


Sumitmaithani:
Hi @snitin315 @saurabhdaware can you please review this PR?

packages/blade/src/components/Table/types.ts
```diff
@@ -354,6 +354,11 @@ type TablePaginationCommonProps = {
    * @default false
    */
   showLabel?: boolean;
+  /**
+   * The background color of the component.
+   * @default transparent
+   */
+  backgroundColor?: string;
```
saurabhdaware, line 361: Hi @Sumitmaithani , thanks for the PR.

So the issue is not that the backgroundColor prop is missing. Its expected that we set backgroundColor internally without taking it as a prop.

The issue is that the backgroundColor is not being set when the `paginationType="server"` prop is there.

You can replicate the issue by setting that prop on TablePagination component.

```jsx
<Table
          data={data}
          pagination={
            <TablePagination paginationType="server" showPageSizePicker showPageNumberSelector />
          }
        >
```

Sumitmaithani: Hi @saurabhdaware,

Thank you for helping me understand the issue. I am new to this codebase. From my understanding, I tried to reproduce the issue here: [https://stackblitz.com/edit/hhwvwi-wm3jcu?file=App.tsx](https://stackblitz.com/edit/hhwvwi-wm3jcu?file=App.tsx). It appears that the TablePagination component has a default transparent background color, regardless of whether we pass paginationType="server".

Could you clarify what it means when we set the background color internally without taking it as a prop? Does that mean we override its CSS? I apologize if I am misunderstanding or bothering you.

packages/blade/src/components/Table/TablePagination.web.tsx
```diff
@@ -242,7 +243,7 @@ const _TablePagination = ({
       display="flex"
       flexDirection="row"
       padding={tablePagination.padding}
-      backgroundColor={tablePagination.backgroundColor}
+      backgroundColor={backgroundColor}
```
chaitanyadeorukhkar, line 246: Take the backgroundColor from `useTableContext` Similar to how we have done it on TableToolbar.web.tsx

Sumitmaithani: Thanks @chaitanyadeorukhkar 🙌, Done.

packages/blade/src/components/Table/types.ts
saurabhdaware: So we by default want to set the backgroundColor. We dont want to accept that as prop from consumer.

You can check how its being set in TableHeader. I think right now we're setting the background color as transparent for footer but it should be same as header so I think we just need to change the background color

Sumitmaithani: Yes, I have fixed it as @chaitanyadeorukhkar mentioned. We are now taking the background color from `useTableContext`.

chaitanyadeorukhkar (approved):

packages/blade/src/components/Table/TablePagination.web.tsx
```diff
@@ -242,7 +243,7 @@ const _TablePagination = ({
       display="flex"
       flexDirection="row"
       padding={tablePagination.padding}
-      backgroundColor={tablePagination.backgroundColor}
```
saurabhdaware, line 245: Small change request: Can we also remove `backgroundColor` key from `tablePagination` object if its not being used anymore?

Sumitmaithani: how I can miss this 🤦. Thanks @saurabhdaware  🙌, done.

saurabhdaware (approved):
Yay! Thanks for your contribution 🥳

PR #2320, opened 2024-07-24
feat(Amount): Negative Amount support

saurabhdaware:
## Changes

- Upgrades i18nify library
- Removes formatNumber and adds formatNumberByParts to support negative sign in amount
- Fixes i18nify and phonenumber input compatibility issue of invalid range

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


saurabhdaware:
Pausing on this for some time since it got a little large task. Will pick it up as proper task in next sprint

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -19,6 +19,43 @@ import { Text } from '~components/Typography';
 import { opacity } from '~tokens/global';
 import type { FontFamily } from '~tokens/global';
 
+/**
+ * Pollyfill function to get around the node 18 error
+ *
+ * This function is maintained by i18nify team. Reach out to them for any change regarding this.
```
anuraghazra, line 25: Should this come from i18nify library itself then instead of being in blade repo?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -19,6 +19,43 @@ import { Text } from '~components/Typography';
...
+    parts.rawParts = parts.rawParts.filter(({ type }) => type !== 'decimal' && type !== 'fraction');
+  }
+
+  return parts;
+};
+
+/**
+ * Wrapper that uses pollyfill of i18nify team
+ */
+const pollyfilledFormatNumberByParts: typeof formatNumberByParts = (value, options) => {
```
anuraghazra, line 49: Also since this is a polyfill are we conditionally calling it or are we always calling it regardless. 

eg: checking for [formatToParts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts) existence

packages/blade/src/components/Amount/Amount.tsx
saurabhdaware: I did suggest them this but they are saying they'll just document it on their end in documentation and consumers will have to add it. Can discuss it later separately as well. Wanted to get it resolved earlier for negative amount thing. But yup overall I too feel this should be in their repo

packages/blade/src/components/Amount/Amount.tsx
saurabhdaware: I've put conditions inside this function itself. So we can always call this instead of `formatNumberByParts`. formatToParts is supported in most major browsers we target so its fine to not check that

anuraghazra (approved):

PR #2336, opened 2024-08-22
docs(motion-presets): Motion Refresh / Presets RFC 2024

saurabhdaware:
### [✨ Preview ✨](https://github.com/razorpay/blade/blob/rfc/motion-presets/rfcs/2024-08-21-motion-presets.md)


rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,161 @@
...
+- What use cases does it support?
+- What is the expected outcome?
+
+Try to focus on explaining the motivation so that if this RFC is not accepted, the motivation could be used to develop alternative solutions. In other words, try to list down the constraints you are trying to solve without coupling them too closely to the solution you have in mind.
+
+# Detailed Design
+
+## Library Comparison
+
+### Goals of Ideal Library
```
chaitanyadeorukhkar, line 40: Our goals should also tie in with design goals... evaluate libraries to see if something we want to do on design whether or not we can do it on dev. Good example you've added is morph animations... do we want to do morph now or later as per design?

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,161 @@
...
+| **React Router Page Transition**         | ✅                                                                                                                                            | ❌ (No native support but can be implemented) | ✅                                     | ❌ (No native support but can be implemented) |
+| **Morph Animations / Layout Animations** | ✅ [Framer Motion POC](#morph-animations-with-framer-motion)                                                                                  | ❌                                            | ✅                                     | ❌                                            |
+| **Small bundle size**                    | ❌ (4.6kb core + 40kb features that can be lazy loaded)                                                                                       | ✅ (4kb )                                     | ❌ (26kb core + features)              | ✅ (0kb)                                      |
```
chaitanyadeorukhkar, line 60: Lets detail out the bundle size for framer motion. How will we load the core module only? In which cases will we need rest of the bundles?

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,161 @@
...
+
+<p align="right">
+<a href="https://codesandbox.io/p/sandbox/framer-motion-hardware-accelarated-animations-forked-7qjyg9">
+  <img alt="Edit GSAP: Hardware Accelarated Animations " src="https://codesandbox.io/static/img/play-codesandbox.svg">
+</a>
+</p>
+
+### Morph / Layout Animations with Framer Motion
+
+https://github.com/user-attachments/assets/ad3d4a23-c3b9-4980-a051-a0f44e7224dc
```
chaitanyadeorukhkar, line 90: Try doing the morph animation with blade components?

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,161 @@
...
+| **License** (Preferrably free to use)    | ✅ MIT                                                                                                                                        | ✅ MIT                                        | ❌ (Commercial License + Paid Plugins) | ✅ No License                                 |
+| **Hardware Accelarated Animations**      | ✅ (Hybrid - WAAPI for some transformations with fallback to JS) [Hardware Accelarated POC](#hardware-accelarated-motion-using-framer-motion) | ✅ (Built on WAAPI)                           | ❌                                     | ✅                                            |
+| **Easy to implement complex animations** | ✅ (Declarative API)                                                                                                                          | ❌                                            | ❌                                     | ❌                                            |
```
chaitanyadeorukhkar, line 57: Motion one seems super light-weight. What is the difficulty in implementing animations with it? Can you add an example of an animation with framer vs with motion one?

Since we will be exposing presets to our end consumers won't it make sense to use a light-weight library within blade and abstract out complex animations within the presets?

rfcs/2024-08-21-motion-presets.md
saurabhdaware: Yup I've added morph and page transitions based on design expectations. We're doing morph now only. 

Other things that are on design are simple opacity, slide, fade, etc transitions which are possible in every library

rfcs/2024-08-21-motion-presets.md
saurabhdaware: Yup will do that POC with API decision now

rfcs/2024-08-21-motion-presets.md
saurabhdaware: Will add more details here. Majority of things come from this article https://motion.dev/blog/should-i-use-framer-motion-or-motion-one (Find "Design goals" heading)

Overall its a low-level imperative API so requires us to call `animate()` function in vanilla JS and write react wrappers on top of it (which is something that we can do with CSS as well)

It lacks support for some important features like layout animations, page transitions, which we're expecting to do. So that will require custom implementation from scratch. It would've worked if we only had opacity, fade, transform transitions but we're doing more than that . If we implement such things from scratch, we'll anyway reach closer to framer-motion minified bundle.

There are projects like modular onboarding, frontend-website which are already using framer-motion and we have't seen load-time issues there. frontend-website had runtime laggy animation issues which are solved in their new update.

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,523 @@
...
+
+# Alternatives
+
+- Alternative libraries and native CSS solution is compared in [Library Comparison Section](#library-comparison)
+- Other alternative is to let consumers do animations
+  - Since there are less high level primitives available and it has led to inconsistent motion across products, we prefer to simplify building animations while giving out consistency
+
+# Adoption strategy
+
+- We plan to target 1 project this quarter (Q2) to get motion adopted
```
kamleshchandnani, line 483:
```suggestion
- We plan to target 1 project this quarter (Q3) to get motion adopted
```
??

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,523 @@
...
+- The new projects that are built, should be built with motion presets on design and dev
+- The earlier project that we have should use motion presets when they redesign / revamp
+
+# How do we educate people?
```
kamleshchandnani, line 487: lets add how will we get this implemented on Figma  for easing the hand-off as well here?

saurabhdaware: Yup I'll update it here. We're still figuring out the right approach. We discussed few things yesterday, RK will just confirm it once if things are working fine or not

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,523 @@
...
+- [API Decision for Motion Presets](#api-decisions)
+- [Libarary to use for Animations](#library-comparison)
+
+## API Decisions
+
+> [!NOTE]
+>
+> The API decisions here are only there to give some basic idea on the structure and usage. More accurate props will be updated here later once they are finalised in design.
+
+### Entry / Exit Animation Presets
```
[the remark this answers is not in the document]
saurabhdaware: Oh yes I was of similar opinion and I discussed this with design but what we realised is its valid case where card on some page might move up whereas card in some other page might fade in. We can't at one place define how section or component appears without knowing context.

From tech perspective I was also thinking of API like `<Card animate>` but in that case framer motion will always end up in bundle irrespective of whether animations are used or not since we'll have to define the node as motion node.

What I was also thinking is, once we have overall idea of animation patterns. Similar to how we have components and how we'll have patterns, we can have motion components and motion patterns (or incorporate motion in some our component patterns itself)

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,523 @@
...
+## Library Comparison
+
+### Goals of Ideal Library
+
+1. License (Preferrably free to use)
+2. Hardware Accelarated Animations (Using CSS or [WAAPI](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API))
+3. Easy to implement complex animations
+4. React Router page transition support
+5. Morph Animations / Layout Animations
+6. Small bundle-size
```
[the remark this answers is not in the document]
saurabhdaware: I think I can add as a note for react native here. Not sure if it'll make sense to add it as a factor in the library comparison table because no web animation library really supports react native so the library exploration for web and native will be different.

I can add a note saying how we use react-native-reanimated already and some of these presets can be built there. I saw that react-native-reanimated already supports a good amount of these presets - https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/entering-exiting-animations

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,523 @@
...
+- Allows presetting animation properties on blade
+
+**Cons:**
+
+- Inconsistent with other motion presets so not very intuitive
+- Comparitively more verbose than suggested API
+
+</details>
+
+#### AnimateInteractions
```
anuraghazra, line 269: What does AnimateInteractions do internally? Can you provide an overview?

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,523 @@
...
+  AnimateInteractions, 
+  Scale 
+} from '@razorpay/blade/components';
+
+<AnimateInteractions>
+  <Card>
+    <CardBody>
+      <Scale>
+        <img src="./rajorpay.jpeg" />
+      </Scale>
```
anuraghazra, line 319: This feels a bit "Magic" with AnimateInteraction automatically making the Scale working. 

What if I want the Scale to trigger based upon some other state or timer? 

Eg: Scale an icon button to make attract attention after a set duration.

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,523 @@
...
+<Stagger isVisible={showCards}>
+  <Fade>
+    <Box />
+  </Fade>
+  <Fade>
+    <Box />
+  </Fade>
+  <Fade>
+    <Box />
+  </Fade>
```
anuraghazra, line 360: What if I mix Scale, Move etc?

rfcs/2024-08-21-motion-presets.md
anuraghazra: And in here what is AnimateInteraction doing? It will create another wrapper with hover events attached? 

Define behaviour of:
1. What happens on focusing the AnimateInteractions? should it have focus? 
2. How will it work in conjunction with shouldScaleOnHover prop of Card
3. AnimateInteractions can be inside Card Body?

rfcs/2024-08-21-motion-presets.md
anuraghazra: We should try to replicate the Switch animation and see how our primitives plays out. 

Switch has a 2 stage animation

Click Hold (Elongate the circle) -> Release (De-elongate the circle + move to the opposite side) 

Scale, transformX both will have to animate at once.

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,523 @@
...
+- [References](#references)
+
+# Summary
+
+# Basic Example
+
+```jsx
+import { Fade, Card, CardBody } from '@razorpay/components/blade';
+
+<Fade isVisible={showCard}>
```
anuraghazra, line 29: How do we pass duration tokens etc?

rfcs/2024-08-21-motion-presets.md
saurabhdaware: It creates motion.div with whileHover, whileTap, whileFocus etc (depending on which interactions you want. Taking it as a prop), it defines the name of the variant of its child.

You can check this POC - https://codesandbox.io/p/sandbox/framer-motion-hover-presets-poc-ghlcpl

The Card component in this example is basically the AnimateInteractions

rfcs/2024-08-21-motion-presets.md
saurabhdaware: works

rfcs/2024-08-21-motion-presets.md
saurabhdaware: we can't. Which is why we have preset so that we can define durations and easing on preset level and make it consistent

rfcs/2024-08-21-motion-presets.md
saurabhdaware: > This feels a bit "Magic" with AnimateInteraction automatically making the Scale working.

True true. I didn't know how the props would be defined while writing this. I feel this feels less magic (this is the final code after implementation) -

### Scale when parent is hovered
```jsx
<AnimateInteraction motionTriggers={['hover']}>
  <SomeOtherCompoennt>
    <Scale><ComponentToScale /></Scale>
  </SomeOtherComponents>
</AnimateInteraction>
```

### Scale when parent is tapped
```jsx
<AnimateInteraction motionTriggers={['tap']}>
  <SomeOtherCompoennt>
    <Scale><ComponentToScale /></Scale>
  </SomeOtherComponents>
</AnimateInteraction>
```


### Scale when the component itself is hovered
```jsx
<Scale motionTriggers={['hover']}><ComponentToScale /></Scale>
```

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,555 @@
...
+   * @default inout
+   */
+  variant: 'in' | 'out' | 'inout';
+
+  /**
+   * What should trigger the motion
+   *
+   * @default ['mount']
+   */
+  motionTriggers: ('mount' | 'hover' | 'tap' | 'inView')[];
```
anuraghazra, line 97: Why do we need to define tap and inview triggers? 

Consumers can do this on their end also like so: 

```
const isInView = useIsInView()

<Fade isVisible={isInView} />
```

Also if we are supporting tap, then we need to support click handlers too?

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,555 @@
...
+<!-- prettier-ignore -->
+```jsx
+import { 
+  AnimateInteractions, 
+  Scale 
+} from '@razorpay/blade/components';
+
+<AnimateInteractions motionTriggers={['hover']}>
+  <Card>
+    <CardBody>
```
anuraghazra, line 301: Will AnimateInteraction create a new div wrapper? or is this a enhancer component?

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,555 @@
...
+      <Text>Fade In/Out Card</Text>
+    </CardBody>
+  </Card>
+</Fade>;
+```
+
+```ts
+type FadeProps = {
```
anuraghazra, line 84: Should we provide atleast the `delay` prop to every primitive? 

So that consumers can choose to delay the animation run by x duration?

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,555 @@
...
+| **React Router Page Transition**         | ✅                                                                                                                                            | ❌ (No native support but can be implemented) | ✅                                     | ❌ (No native support but can be implemented) |
+| **Morph Animations / Layout Animations** | ✅ [Framer Motion POC](#morph-animations-with-framer-motion)                                                                                  | ❌                                            | ✅                                     | ❌                                            |
+| **Small bundle size**                    | ❌ (4.6kb core + 15kb (for base animations) + 10kb (if Morph preset is used))                                                                 | ✅ (4kb )                                     | ❌ (26kb core + features)              | ✅ (0kb)                                      |
```
anuraghazra, line 409: We should evaluate LazyMotion to see if we can shave off few bytes: 

https://www.framer.com/motion/lazy-motion/

rfcs/2024-08-21-motion-presets.md
```diff
@@ -0,0 +1,555 @@
...
+
+#### App.tsx
+
+```jsx
+// Make sure to return the specific export containing the feature bundle.
+const loadFeatures = () => import('./features').then((res) => res.default);
+
+// This animation will run when loadFeatures resolves.
+function App({ children }) {
+  return <LazyMotion features={loadFeatures}>{children}</LazyMotion>;
```
anuraghazra, line 501: Will consumer do this or we do this? 

We will anyways have to use the framer-motion's provider right because blade components itself may also use the motion primitives. 

But if we already have the provider and as a library we can't really do lazy import on our end then we don't really have choice other than to use the full framer motion right?

saurabhdaware: Consumer will do it. It will be required peer dependency.

We will use it in our components. Sort of like i18nify.

rfcs/2024-08-21-motion-presets.md
saurabhdaware: I added all the triggers here which are natively supported by framer-motion so API is simpler

https://www.framer.com/motion/scroll-animations/#scroll-triggered-animations

> Also if we are supporting tap, then we need to support click handlers too?

Nope. Click handlers can be added to children directly no if they support it? This is purely for animation component. It shouldn't be used to add any logic.

E.g. I can just add click handlers on children if needed for some logic

```jsx
<Fade>
  <Card onClick={() => } />
</Fade>
```

rfcs/2024-08-21-motion-presets.md
saurabhdaware: I had to create a div wrapper because something was not working. Lemme check again and see if it can be enhancer component

rfcs/2024-08-21-motion-presets.md
saurabhdaware: These props are not finalized. Only structure is. Will add props depending on discussion with design

rfcs/2024-08-21-motion-presets.md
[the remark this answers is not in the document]
saurabhdaware: I added our discussion points and why we're going this way in open questions section along with conclusion

rfcs/2024-08-21-motion-presets.md
jaywalker21: On the same note, should we have tokens for `duration` if we don't allow explicit values? Certain use cases might need animations to run for different times. I understand it won't be an issue for looped ones, but for non-looped ones, this will help

saurabhdaware: Atleast from design-side we're saying that the enter and exit durations should have same duration always. Do you know any usecase where a single duration won't work? we can also add this prop in future if usecase comes since this anyways won't be a breaking change then.

rfcs/2024-08-21-motion-presets.md
Rozamo: > define durations and easing on preset level
Do you mean when we define the tokens?

saurabhdaware: No like internally durations are defined for each preset. E.g Fade has certain given duration that cannot be changed on design as well

PR #2341, opened 2024-08-28
feat(ActionList): add ActionListItemAvatar component for leading prop

saurabhdaware:
## Description

Support for Avatar in ActionListItem leading

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/ActionList/ActionListItem.tsx
+++ b/packages/blade/src/components/ActionList/ActionListItem.tsx
@@ -25,7 +25,7 @@ import { Box } from '~components/Box';
 import { dropdownComponentIds } from '~components/Dropdown/dropdownComponentIds';
 import { BaseMenuItem, useBaseMenuItem } from '~components/BaseMenu';
 import { Checkbox } from '~components/Checkbox';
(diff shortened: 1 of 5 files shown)


packages/blade/src/components/ActionList/ActionListItem.tsx
```diff
@@ -210,6 +212,14 @@ const ActionListItemBadgeGroup = assignWithoutSideEffects(_ActionListItemBadgeGr
   componentId: componentIds.ActionListItemBadgeGroup,
 });
 
+const _ActionListItemAvatar = (avatarProps: Omit<AvatarProps, 'size' | ''>): React.ReactElement => {
+  return <Avatar size="xsmall" {...avatarProps} />;
```
anuraghazra, line 216: Consumer cannot change size?

anuraghazra (approved):

packages/blade/src/components/ActionList/ActionListItem.tsx
saurabhdaware: yup checked with RK. size is always same

PR #2344, opened 2024-08-28
fix: hint-text wrapping and alignment

anuraghazra:
## Description

Fixes: https://razorpay.atlassian.net/browse/DSSUP-69

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Form/FormHint.tsx
+++ b/packages/blade/src/components/Form/FormHint.tsx
@@ -6,11 +6,13 @@ import { hintIconSize, hintMarginTop, hintTextSize } from './formTokens';
 import type { TextProps } from '~components/Typography/Text';
 import { Text } from '~components/Typography/Text';
 import BaseBox from '~components/Box/BaseBox';
+import type { IconComponent } from '~components/Icons';
 import { CheckIcon, InfoIcon } from '~components/Icons';
 import { getPlatformType } from '~utils/getPlatformType';
+import { Box } from '~components/Box';
 
 type HintTextProps = {
-  icon?: React.ElementType;
+  icon?: IconComponent;
   children: string;
   id?: string;
   color: TextProps<{ variant: 'caption' }>['color'];
@@ -23,12 +25,17 @@ const HintText = ({ icon: Icon, children, id, color, size }: HintTextProps): Rea
   return (
     <BaseBox marginTop={hintMarginTop[size]} id={id}>
       <FormHintWrapper>
-        {Icon ? <Icon /> : null}
+        {Icon ? (
+          <Box flexShrink={0} marginTop="spacing.1">
+            <Icon />
+          </Box>
+        ) : null}
         <Text
           as={isReactNative ? undefined : 'span'}
           color={color}
           size={hintTextSize[size]}
           variant="caption"
+          wordBreak="break-word"
         >
           {children}
         </Text>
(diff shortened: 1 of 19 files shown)


packages/blade/src/components/Form/FormHint.tsx
```diff
@@ -23,12 +25,17 @@ const HintText = ({ icon: Icon, children, id, color, size }: HintTextProps): Rea
   return (
     <BaseBox marginTop={hintMarginTop[size]} id={id}>
       <FormHintWrapper>
-        {Icon ? <Icon /> : null}
+        {Icon ? (
+          <Box flexShrink={0} marginTop="spacing.1">
+            <Icon />
+          </Box>
```
saurabhdaware, line 31: what is spacing.1? add comment maybe?

packages/blade/src/components/Form/FormHint.tsx
```diff
@@ -79,16 +86,18 @@ export type FormHintProps = {
 
 const Icons = {
   error: ({ size }: { size: 'small' | 'medium' | 'large' }): ReactElement => (
-    <>
-      <InfoIcon color="feedback.icon.negative.intense" size={hintIconSize[size]} />
-      <BaseBox marginRight="spacing.2" />
-    </>
+    <InfoIcon
+      display={'block' as never}
```
saurabhdaware, line 90: why never? this code is used in react native also no?

packages/blade/src/components/Form/FormHintWrapper.web.tsx
```diff
@@ -3,15 +3,7 @@ import BaseBox from '~components/Box/BaseBox';
...
-    <BaseBox
-      as="span"
-      display="flex"
-      flexDirection="row"
-      alignItems="center"
-      style={{
-        wordBreak: 'break-all',
-      }}
-    >
+    <BaseBox as="span" display="flex" flexDirection="row" alignItems="flex-start" gap="spacing.2">
```
saurabhdaware, line 6: gap wasn't there earlier. is this adding some extra spacing?

anuraghazra: It was there. https://github.com/razorpay/blade/blob/b9d4e13725382b07f791ad3b8bf0980e9ae64c7b/packages/blade/src/components/Form/FormHint.tsx#L90

We used spacer components before cuz RN didn't support gap prop

packages/blade/src/components/Form/FormHint.tsx
anuraghazra: I think it'll just be ignored on RN. 

We do the same on other places as well

https://github.com/razorpay/blade/blob/601cedc8ccf07984a2a2a88b92ca3eb8d28b1f84/packages/blade/src/components/Card/Card.tsx#L204

anuraghazra: Our RN setup is totally broken as of now, I do not currently have bandwidth to check and confirm if this works or not.

packages/blade/src/components/Form/FormHint.tsx
anuraghazra: done

packages/blade/src/components/Form/FormHint.tsx
saurabhdaware: lol makes sense. yolo

packages/blade/src/components/Form/FormHint.tsx
saurabhdaware: lies

saurabhdaware (approved):

PR #2346, opened 2024-08-28
fix: support overriding tooltip interactive wrapper display prop

anuraghazra:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.web.tsx
+++ b/packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.web.tsx
@@ -12,7 +12,7 @@ const TooltipInteractiveWrapper = React.forwardRef<HTMLDivElement, Omit<BaseBoxP
       <BaseBox
         ref={ref}
         tabIndex={-1}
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/Tooltip/TooltipInteractiveWrapper.web.tsx
```diff
@@ -1,22 +1,26 @@
...
 
-  return {
-    display: 'inline-block',
-  };
-});
+    return (
+      <BaseBox
+        ref={ref}
+        tabIndex={-1}
+        display={props.display || 'inline-block'}
```
saurabhdaware, line 15:
```suggestion
        display="inline-block"
```

This is not needed right? since we're destructuring props below

anuraghazra: 🤔 right..

saurabhdaware (approved):

saurabhdaware (approved):

PR #2354, opened 2024-09-06
feat(Table): add `defaultSelectedIds` prop

chaitanyadeorukhkar:
## Description

- Add `defaultSelectedIds` to Table component
- Fixes https://razorpay.atlassian.net/browse/DSSUP-90

## Component Checklist

- [x] Add KitchenSink Story
- [x] Add changeset

### DIFF
--- a/packages/blade/src/components/Table/__tests__/__snapshots__/Table.web.test.tsx.snap
+++ b/packages/blade/src/components/Table/__tests__/__snapshots__/Table.web.test.tsx.snap
@@ -501,6 +501,8 @@ exports[`<Table /> should render table 1`] = `
         data-table-library_body=""
       >
         <tr
+          aria-multiselectable="false"
+          aria-selected="false"
           class="tr tr-body c11 clickable css-1rzx1gz-getRowContainerStyle-Row"
           data-blade-component="table-row"
           data-table-library_tr-body=""
@@ -665,6 +667,8 @@ exports[`<Table /> should render table 1`] = `
           </td>
         </tr>
         <tr
+          aria-multiselectable="false"
+          aria-selected="false"
           class="tr tr-body c11 clickable css-1rzx1gz-getRowContainerStyle-Row"
           data-blade-component="table-row"
           data-table-library_tr-body=""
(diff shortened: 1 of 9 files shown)


packages/blade/src/components/Table/__tests__/__snapshots__/Table.web.test.tsx.snap
```diff
@@ -1157,6 +1165,8 @@ exports[`<Table /> should render table 1`] = `
           </td>
         </tr>
         <tr
+          aria-multiselectable="false"
```
saurabhdaware, line 1168: aria-multiselectable will go on the parent and aria-selected will be there on child right 🤔 ?

saurabhdaware (approved):
Looks good to me. One comment regarding a11y

packages/blade/src/components/Table/docs/BasicTable.stories.tsx
```diff
@@ -119,6 +119,8 @@ const TableTemplate: StoryFn<typeof TableComponent> = ({ ...args }) => {
       <TableComponent
         {...args}
         data={data}
+        defaultSelectedIds={['1', '3']}
+        onSelectionChange={console.log}
```
saurabhdaware, line 123: I don't see this reflected in story 🤔 

https://61c19ee8d3d282003ac1d81c-pghwdoruck.chromatic.com/?path=/story/components-table--table&args=selectionType:multiple

saurabhdaware: oh got it got it. It works when you select "multiple" then refresh the page

chaitanyadeorukhkar: Yeah since its `default`, it'll only take the values on first render

packages/blade/src/components/Table/__tests__/__snapshots__/Table.web.test.tsx.snap
chaitanyadeorukhkar: Yeah, this should go on table instead

chaitanyadeorukhkar: Updated

saurabhdaware (approved):

PR #2356, opened 2024-09-10
feat(motion.ts): add refreshed tokens and internal components migration

saurabhdaware:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


anuraghazra:
2 queries: 

1. Did we sync up with RK about the animations being slow?
2. Did we check the blast radius of changing token names?

anuraghazra (approved):

saurabhdaware:
> Did we sync up with RK about the animations being slow?

Yup we changed durations for some components

> Did we check the blast radius of changing token names?

Yes there are few (~40 files) in razorpay which use these. we've written codemod

PR #2357, opened 2024-09-10
fix: AccordionItemHeader to take full width

Akshay090:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


anuraghazra:
@Akshay090 please update snapshots and add changesets.

packages/blade/src/components/BaseHeaderFooter/BaseHeader.tsx
```diff
@@ -363,7 +363,7 @@ const _BaseHeader = ({
           alignItems="center"
           justifyContent="space-between"
         >
-          <Box>{children}</Box>
+          {children}
```
saurabhdaware, line 366: I think we can try giving the existing Box a `flex={1}` or `width="100%"`

From what I remember, this exists because if in some case you pass multiple children component instead of one component, they will get rendered side-by-side because outer Box to this is a flex

saurabhdaware (commented):
Can you update the snapshots?

saurabhdaware (approved):

PR #2360, opened 2024-09-17
feat(motion): add initial presets code

saurabhdaware:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/.storybook/react/preview.tsx
```diff
@@ -180,15 +181,18 @@ export const decorators = [
...
-          key={`${context.globals.themeTokenName}-${context.globals.colorScheme}`}
-          themeTokens={getThemeTokens()}
-          colorScheme={context.globals.colorScheme}
-        >
-          <StoryCanvas context={context}>
-            <Story />
-          </StoryCanvas>
-        </BladeProvider>
+        {/* strict in LazyMotion will make sure we don't use excessive `motion` component in blade components and instead use light weight `m` */}
+        <LazyMotion strict features={domMax}>
```
anuraghazra, line 185: Does the consumer need to add the Motion Provider?

If so, this would be a breaking change, correct? Assuming we start using Framer Motion internally for our components as well.

Consider this scenario:

- The consumer is on version 11.37.0.
- In version 11.38.0, we introduce Framer Motion in the BottomSheet component.
- Since the consumer hasn't added the Motion Provider at the root of their app, the BottomSheet component in Blade would stop working.

This means simply running `npm i` to upgrade the Blade version wouldn’t be sufficient in this case.

packages/blade/src/components/AnimateInteractions/useFocusWithin.ts
```diff
@@ -0,0 +1,35 @@
...
+export function useFocusWithin<T extends HTMLElement>(
+  ref: React.RefObject<T>,
+  { onFocusWithin, onBlurWithin }: FocusWithinHandlers,
+) {
+  useEffect(() => {
+    const element = ref.current;
+    if (!element) return;
+
+    const handleFocusIn = () => {
+      if (onFocusWithin) onFocusWithin();
```
anuraghazra, line 17:
```suggestion
      onFocusWithin?.();
```

packages/blade/src/components/AnimateInteractions/useFocusWithin.ts
```diff
@@ -0,0 +1,35 @@
...
+    };
+
+    const handleFocusOut = (event: FocusEvent) => {
+      // Ensure that focus is not still within the container
+      if (element && !element.contains(event.relatedTarget as Node)) {
+        if (onBlurWithin) onBlurWithin();
```
anuraghazra, line 23:
```suggestion
        onBlurWithin?.();
```

packages/blade/src/components/BaseMotion/BaseMotionBox.tsx
```diff
@@ -0,0 +1,105 @@
...
+        prevProps.animate = conditionalAnimate;
+        return prevProps;
+      }
+
+      prevProps[motionTriggersArrayToGesturePropsMap[currentTrigger]] = 'animate';
+
+      return prevProps;
+    }, {});
+
+    // When component is rendered inside stagger, we remove the initial, animate, exit props
```
anuraghazra, line 65: This comment talks about "when rendered inside stagger" but i don't see any checks or boolean flags which lets you know if you are inside stagger, what are we doing here?

packages/blade/src/components/BaseMotion/BaseMotionBox.tsx
```diff
@@ -0,0 +1,105 @@
...
+  }: BaseMotionBoxProps,
+  ref: React.Ref<BladeElementRef>,
+) => {
+  const animationVariables = useAnimationVariables({
+    type,
+    motionTriggers,
+    conditionalAnimate,
+  });
+
+  console.log({ motionVariants, animationVariables, rest });
```
anuraghazra, line 94:
```suggestion
```

packages/blade/src/components/BaseMotion/types.ts
```diff
@@ -0,0 +1,68 @@
+import type { AnimationControls, Variant } from 'motion/react';
+
+type MotionTriggerEntryExitType = 'mount' | 'inView' | 'focus' | 'on-animate-interactions';
```
anuraghazra, line 3: why is "inView" in camel case and others hyphenated?

packages/blade/src/utils/cssBezierToMotionFn.ts
```diff
@@ -0,0 +1,24 @@
+const cssBezierToMotionFn = <
```
anuraghazra, line 1: Bit misleading name, it says "bezier to motion fn" but returns an array. :p 

CSSBezierToArray

packages/blade/src/utils/getMotionRefs.ts
```diff
@@ -0,0 +1,17 @@
+import type React from 'react';
+import type { MotionMetaProp } from '~components/BaseMotion';
+import type { BladeElementRef } from './types';
+
+type MotionRefsType = {
+  ref: React.Ref<BladeElementRef>;
+} & MotionMetaProp;
+
+const getOuterMotionRef = ({ _motionMeta, ref }: MotionRefsType): any => {
```
anuraghazra, line 9: let's add some details/docs and reasoning for these two functions, right now someone looking at this code have no idea what is `isEnhanced` why it's there, why we are using `innerRef` where did it come from etc etc. 

Let's properly document these

packages/blade/src/components/Tag/Tag.tsx
```diff
@@ -28,44 +29,20 @@ const FocussableTag = styled(BaseBox)<{ _isVirtuallyFocused: TagProps['_isVirtua
...
-  icon: Icon,
-  onDismiss,
-  children,
-  isDisabled,
-  testID,
-  _isVirtuallyFocused,
-  _isTagInsideInput,
-  ...styledProps
-}: TagProps): React.ReactElement | null => {
+const _Tag = (
```
anuraghazra, line 32: Removed jsdoc?

packages/blade/src/components/Scale/Scale.web.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { BaseMotionEnhancerBox } from '~components/BaseMotion';
+import type { MotionVariantsType } from '~components/BaseMotion';
+import { makeSecondsDuration } from '~utils/makeSecondsDuration';
+import { cssBezierToMotionFn } from '~utils/cssBezierToMotionFn';
+import { castWebType, useTheme } from '~utils';
+import type { ScaleProps } from './types';
+
+export const Scale = ({
+  children,
+  isHighlighted,
```
anuraghazra, line 11: what does isHighlighted prop do?

packages/blade/.storybook/react/preview.tsx
saurabhdaware: Yup we're making breaking change

packages/blade/src/components/Scale/Scale.web.tsx
saurabhdaware: in context of scale, `isHighlighted={true}` will scale the item up

packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -326,5 +330,7 @@ const Alert = ({
   );
 };
 
+const Alert = forwardRef(_Alert);
```
anuraghazra, line 333: Should we use that `assignWithoutSideEffects` and add displayNames to these components?

saurabhdaware: I checked on storybook and they do seem to have correct display names there so don't think so its needed for other things

anuraghazra (approved):

PR #2388, opened 2024-10-25
docs(BottomNav): add BottomNav API

saurabhdaware:
### [✨ Pretty Doc ✨](https://github.com/razorpay/blade/blob/bottomnav/api/packages/blade/src/components/BottomNav/_decisions/decisions.md)


packages/blade/src/components/BottomNav/_decisions/decisions.md
```diff
@@ -0,0 +1,130 @@
...
+
+<!-- prettier-ignore -->
+```jsx
+import { NavLink } from 'react-router-dom';
+
+<BottomNav>
+  <BottomNavItem 
+    as={NavLink} 
+    isActive={true}
+    title="Payments"
```
anuraghazra, line 21: We can keep this API same as the TabNavItem? 

We use `children` instead of `label`

saurabhdaware: Children we can leave out if we need slot in future no? This one is similar to SideNavLink, ActionListItem, MenuItem, AccordionItem etc.

I think overall what we're following on these item level components is to use `title` as default and use `children` when you want to override it to add slot

anuraghazra: makes sense, changed topnav to have title.

kamleshchandnani (approved):

PR #2395, opened 2024-10-28
feat(BottomNav): Implementation, Documentation, Tests

saurabhdaware:
## Description

- feat: BottomNav component
- Docs and Tests

## Component Checklist

- [x] Add KitchenSink Story
- [x] Add changeset


packages/blade/src/components/BottomNav/_decisions/decisions.md
```diff
@@ -0,0 +1,130 @@
+# BottomNav
```
anuraghazra, line 1: Decisions are also here?

packages/blade/src/components/BottomNav/types.ts
```diff
@@ -0,0 +1,65 @@
...
+   * Active state of the BottomNavItem.
+   *
+   * Use this to set the current page's active state using react router
+   */
+  isActive?: boolean;
+
+  /**
+   * onClick handler of BottomNavItem
+   */
+  onClick?: () => void;
```
anuraghazra, line 62: Expose onClick `event` and also expose `ref` from the BottomNavItem. `(event) => void`

React router depends on ref to inject few attributes into DOM, and ReactRouterV6 requires the onClick's event.

packages/blade/src/components/BottomNav/BottomNav.web.tsx
```diff
@@ -0,0 +1,153 @@
...
+    backgroundColor: 'transparent',
+    border: 'none',
+    paddingLeft: makeSpace(props.theme.spacing[0]),
+    paddingRight: makeSpace(props.theme.spacing[0]),
+    '&[aria-current="page"]': {
+      color: props.theme.colors.interactive.text.primary.subtle,
+    },
+    '&:focus-visible': {
+      ...getFocusRingStyles({ theme: props.theme }),
+    },
```
anuraghazra, line 109:
```suggestion
    '&:focus-visible': getFocusRingStyles({ theme: props.theme }),
```

packages/blade/src/components/BottomNav/types.ts
```diff
@@ -0,0 +1,65 @@
...
+   * Accepts icon component from blade
+   */
+  icon: IconComponent;
+
+  /**
+   * href property of link
+   *
+   * maps to `to` property when react router is being used
+   */
+  href?: string;
```
anuraghazra, line 38: add link related props like `target`

packages/blade/src/components/BottomNav/BottomNav.web.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+const StyledBottomNavItem = styled(BaseBox)<{ to?: string }>((props) => {
+  return {
+    textDecoration: 'none',
+    color: props.theme.colors.interactive.text.gray.subtle,
+    backgroundColor: 'transparent',
+    border: 'none',
+    paddingLeft: makeSpace(props.theme.spacing[0]),
+    paddingRight: makeSpace(props.theme.spacing[0]),
+    '&[aria-current="page"]': {
```
anuraghazra, line 104: No transition or animations when switching active tab? won't it look abrupt?

packages/blade/src/components/BottomNav/_decisions/decisions.md
saurabhdaware: forgot to stack it on top of earlier PR

packages/blade/src/components/BottomNav/types.ts
saurabhdaware: We're using `as={NavLink}` for React Router so it won't be used for routing but yup I'll expose event either ways

packages/blade/src/components/BottomNav/types.ts
saurabhdaware: I did think of it during API decision. I feel we shouldn't because BottomNav shouldn't open links in new window

packages/blade/src/components/BottomNav/BottomNav.web.tsx
saurabhdaware: Hmm yes I can check with RK once he is back. Will probably take it as enhancement if we end up going this route. I can add same transitions as SideNavLink meanwhile

packages/blade/src/components/BottomNav/types.ts
anuraghazra: Sure, 

Ideally our components should be composable with any method. for example react router v6 expects composition via their hook and passing props like ref etc. 

https://reactrouter.com/en/main/upgrading/v5#remove-link-component-prop

packages/blade/src/components/BottomNav/types.ts
saurabhdaware: makes sense

saurabhdaware: done

anuraghazra (approved):

anuraghazra (approved):

anuraghazra (approved):

PR #2401, opened 2024-11-04
fix(blade):  Date Picker does not fire onInput event [DSSUP-127]

tewarig:
## Description

Date picker component does not fire change or input event when the value is changed or committed. Expectation is to behave like an input field with type as date. 

## Changes
This pr - 
- [x] adds onInput prop in Datepicker
- [x] add mode presets in Datepicker with Presets example 

## Component Checklist

- [ ] Add KitchenSink Story
- [x] Add changeset

### DIFF
--- a/packages/blade/src/components/DatePicker/DatePicker.stories.tsx
+++ b/packages/blade/src/components/DatePicker/DatePicker.stories.tsx
@@ -35,6 +35,7 @@ export default {
     value: baseProp,
     isOpen: baseProp,
     onChange: baseProp,
+    onInput: baseProp,
     selectionType: baseProp,
     presets: baseProp,
     minDate: baseProp,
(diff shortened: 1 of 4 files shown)


packages/blade/src/components/DatePicker/DatePicker.stories.tsx
```diff
@@ -132,13 +136,25 @@ SingleDatePicker.storyName = 'SingleDatePicker';
 SingleDatePicker.args = {
   label: 'Select a date',
   selectionType: 'single',
+  onInput: (e) => {
```
saurabhdaware, line 139: I think what they wanted was, if they do `document.querySelector('input[type="date"]').addEventListener('input', () => {}` then callback should get called (@anuraghazra can you confirm?)

Our onInput and onChange wont dispatch the native DOM events

PR #2402, opened 2024-11-04
Fix(blade) : File upload does not fire change input when file is dropped

tewarig:
## Description

FileUpload component does not fire `change` or `input `event when the file is dropped. 

## Changes

- Add onInput as  prop in FileUpload
- now, file Input fires onChange , onInput when a file is dropped.

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/FileUpload/FileUpload.web.tsx
+++ b/packages/blade/src/components/FileUpload/FileUpload.web.tsx
@@ -38,6 +38,7 @@ const _FileUpload: React.ForwardRefRenderFunction<BladeElementRef, FileUploadPro
     onReupload,
     onDismiss,
     onDrop,
+    onInput,
     isDisabled,
     isRequired,
     necessityIndicator,
@@ -146,6 +147,17 @@ const _FileUpload: React.ForwardRefRenderFunction<BladeElementRef, FileUploadPro
     setIsActive(false);
   };
 
+  const handleFileChangeEvents = ({
+    fileList,
+    name,
+  }: {
+    fileList: BladeFileList;
+    name?: string;
+  }): void => {
+    onChange?.({ name, fileList });
+    onInput?.({ name, fileList });
+  };
+
   const handleDrop = (event: React.DragEvent): void => {
     event.preventDefault();
     setIsActive(false);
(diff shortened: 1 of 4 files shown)


packages/blade/src/components/FileUpload/types.ts
```diff
@@ -70,9 +70,13 @@ type FileUploadCommonProps = {
...
   maxSize?: number;
   /**
-   * Callback function triggered when files are selected
+   * Callback function triggered when files are selected or when a file is dropped into the upload area
    */
   onChange?: ({ name, fileList }: { name?: string; fileList: BladeFileList }) => void;
+  /**
+   * Callback function triggered when files are selected or when a file is dropped into the upload area
+   */
+  onInput?: ({ name, fileList }: { name?: string; fileList: BladeFileList }) => void;
```
snitin315, line 79: This seems same as `onChange `, are we planning to deprecate the `onChange` prop in favor of `onInput`? If so, we need to update the description accordingly.

tewarig: in case of <input type="file" /> onChange and onInput have same behaviour.
https://codepen.io/gauravTRazorpay/pen/wvVxMPM  but currently fileUpload don't have onInput prop. so FileUpload component does not fire  onInput event when the file is dropped.  we need to trigger native input events in file upload.

packages/blade/src/components/FileUpload/FileUpload.web.tsx
```diff
@@ -146,6 +147,17 @@ const _FileUpload: React.ForwardRefRenderFunction<BladeElementRef, FileUploadPro
...
 
+  const handleFileChangeEvents = ({
+    fileList,
+    name,
+  }: {
+    fileList: BladeFileList;
+    name?: string;
+  }): void => {
+    onChange?.({ name, fileList });
+    onInput?.({ name, fileList });
```
saurabhdaware, line 158: https://github.com/razorpay/blade/pull/2401/files#r1830421650

Check this once with Anurag

tewarig: checked closing these PR

tewarig: checked closing these PR

PR #2404, opened 2024-11-05
feat(blade): add controlled props to carousel

anuraghazra:
## Description

Fixes: https://razorpay.atlassian.net/browse/DSSUP-164

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Carousel/Carousel.stories.tsx
+++ b/packages/blade/src/components/Carousel/Carousel.stories.tsx
@@ -14,6 +14,7 @@ import { isReactNative } from '~utils';
 import { List, ListItem } from '~components/List';
 import { Link } from '~components/Link';
 import { useTheme } from '~components/BladeProvider';
+import { Button } from '~components/Button';
 
 const Page = (): React.ReactElement => {
   return (
@@ -475,6 +476,9 @@ export const Controlled: StoryFn<typeof CarouselComponent> = (props) => {
         Setting <Code>activeSlide</Code> & <Code>onChange</Code> you can control the active slide
         and use the carousel in a controlled way. Here the active slide is {activeSlide}
       </Text>
+      <Button marginY="spacing.4" size="small" onClick={() => setActiveSlide(2)}>
+        Go to slide #3
+      </Button>
       <CarouselExample
         {...props}
         activeSlide={activeSlide}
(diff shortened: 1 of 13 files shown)


packages/blade/src/components/Carousel/Carousel.stories.tsx
```diff
@@ -436,6 +438,70 @@ AutoPlay.argTypes = {
...
+      <Text marginY="spacing.5">
+        Setting <Code>activeSlide</Code> & <Code>onChange</Code> you can control the active slide
+        and use the carousel in a controlled way. Here the active slide is {activeSlide}
+      </Text>
+      <Button marginY="spacing.4" size="small" onClick={() => setActiveSlide(2)}>
+        Go to slide #3
+      </Button>
+      <CarouselExample
+        {...props}
+        activeSlide={activeSlide}
```
saurabhdaware, line 484: Should we still call this prop `value`? E.g. in select input, we set selected items using prop `value`.

`activeSlide` + `onChange` combination feels a little strange as opposed to `activeSlide` + `onActiveSlideChange`, `value` + `onChange`

packages/blade/src/components/Carousel/Carousel.web.tsx
```diff
@@ -454,21 +446,33 @@ const Carousel = ({
     },
   );
 
+  // set initial active slide on mount
+  useIsomorphicLayoutEffect(() => {
+    if (!id) return;
+    goToSlideIndex(activeSlide);
+    scrollToSlide(activeSlide, false);
+  }, [id]);
```
saurabhdaware, line 454: when does the id change? wouldn't it always stay same 🤔

anuraghazra: useId hook returns null on first render, thus we only want to run it once id is actually present.

packages/blade/src/components/Carousel/Carousel.stories.tsx
anuraghazra: Sure it sounds a bit strange because we initially called the onChange as `onChange` instead of something else like `onSlideChange` 

But value is also strange in context of carousel, in input it makes sense because input's have "values" but in carousel there no "value" as such, so activeSlide may be better in this case.

saurabhdaware: hmm yeah its not input so makes sense I think

saurabhdaware (approved):
Looks good. Krdo merge

PR #2408, opened 2024-11-06
Fix: trigger native events

tewarig:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/FileUpload/FileUpload.web.tsx
+++ b/packages/blade/src/components/FileUpload/FileUpload.web.tsx
@@ -26,6 +26,7 @@ import { makeAccessible } from '~utils/makeAccessible';
 import { formHintLeftLabelMarginLeft } from '~components/Input/BaseInput/baseInputTokens';
 import { useMergeRefs } from '~utils/useMergeRefs';
 import { useControllableState } from '~utils/useControllable';
+import { fireNativeEvent } from '~utils/fireNativeEvent';
 
 const _FileUpload: React.ForwardRefRenderFunction<BladeElementRef, FileUploadProps> = (
   {
@@ -158,6 +159,7 @@ const _FileUpload: React.ForwardRefRenderFunction<BladeElementRef, FileUploadPro
     if (!hasValidationErrors) {
       handleFilesChange(droppedFiles);
       onDrop?.({ name, fileList: allFiles });
+      fireNativeEvent(inputRef, allFiles, ['change', 'input']);
     }
   };
 
@@ -169,6 +171,7 @@ const _FileUpload: React.ForwardRefRenderFunction<BladeElementRef, FileUploadPro
 
     if (!hasValidationErrors) {
       handleFilesChange(inputFiles);
+      fireNativeEvent(inputRef, inputFiles, ['change', 'input']);
       onChange?.({ name, fileList: allFiles });
     }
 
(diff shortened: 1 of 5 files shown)


packages/blade/src/utils/fireNativeEvent/fireNativeEvent.web.ts
```diff
@@ -0,0 +1,30 @@
...
+  eventTypes: Array<'change' | 'input'>,
+): void => {
+  if (ref) {
+    eventTypes.forEach((eventType) => {
+      const event = new Event(eventType, { bubbles: true });
+      ref.current?.dispatchEvent(event);
+    });
+  } else {
+    console.warn('ref is not defined');
+  }
```
anuraghazra, line 29: I think it's fine if we remove the warning, since if this for some reason stops working it won't really break the whole component. 

We can just do

```
if (!ref) return;
```

packages/blade/src/utils/fireNativeEvent/index.ts
```diff
@@ -0,0 +1 @@
+export * from './fireNativeEvent.web';
```
anuraghazra, line 1: This may break rn builds. 

To support both rn & web seamlessly we can structure it like so: 

We can also add another file specific for react-native `fireNativeEvent.native.ts` 

And inside of the rn implementation, you can just throw error that fireNativeEvent is only supported on web.


And then you will just be able to export 
```
export * from './fireNativeEvent';
```

packages/blade/src/components/FileUpload/FileUpload.web.tsx
```diff
@@ -158,6 +159,7 @@ const _FileUpload: React.ForwardRefRenderFunction<BladeElementRef, FileUploadPro
     if (!hasValidationErrors) {
       handleFilesChange(droppedFiles);
       onDrop?.({ name, fileList: allFiles });
+      fireNativeEvent(inputRef, ['change', 'input']);
```
anuraghazra, line 162: Check the normal DOM inputs if we actually fire both of these or not. I think they do but still confirm once.

packages/blade/src/components/Input/DropdownInputTriggers/BaseDropdownInputTrigger.tsx
```diff
@@ -176,6 +167,7 @@ const _BaseDropdownInputTrigger = (
     defaultValue: props.defaultValue,
     syncInputValueWithSelection: props.syncInputValueWithSelection,
     isSelectInput: props.isSelectInput,
+    triggererRef,
```
anuraghazra, line 170: @saurabhdaware can you check this once, not sure ye kaha se agaya

anuraghazra (changes requested):
Mostly LGTM, Just add tests

packages/blade/src/components/FileUpload/FileUpload.web.tsx
tewarig: checked for ```  <input type="file" /> ``` they do fire both change and  input .

packages/blade/src/components/Input/DropdownInputTriggers/BaseDropdownInputTrigger.tsx
tewarig: this to i have added  in -
``` packages/blade/src/components/Input/DropdownInputTriggers/types.ts ```

anuraghazra:
Can you check for the failing tests

tewarig:
@anuraghazra changes done

anuraghazra (approved):

PR #2418, opened 2024-11-15
fix: dropdown not opening if user clicks on dropdown backdrop

tewarig:
## Description

bug - 
https://github.com/user-attachments/assets/a135e5a2-e283-466e-b58d-76785ed0e10b

RCA -
let's say currently we have a structure like - 
<Box>
 <Input/>
<TriggerElement/>
</Box> 
so we have added onClick listner to input and triggerElement and if user clicks on backdrop of tiggerElement if triggerelement does not takes full space. thier is nothing to fire. this was causing the issue.
image

## Changes
image
added an additional prop onBackdropClick , which is added on the wrapper so that when ever use clicks on backdrop we ca decide what need to be done.

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
+++ b/packages/blade/src/components/Input/BaseInput/BaseInput.tsx
@@ -326,7 +326,10 @@ type BaseInputCommonProps = FormInputLabelProps &
      * @default true
      **/
     isTableInputCell?: boolean;
-
+    /**
+     * Callback to be invoked when the backdrop is clicked
+     */
+    onBackdropClick?: () => void;
     /**
      * Hides the form hints and shows them as tooltip of trailing
      */
@@ -825,6 +828,7 @@ const _BaseInput: React.ForwardRefRenderFunction<BladeElementRef, BaseInputProps
     valueComponentType = 'text',
     isTableInputCell = false,
     showHintsAsTooltip = false,
+    onBackdropClick,
     ...styledProps
   },
   ref,
@@ -971,6 +975,7 @@ const _BaseInput: React.ForwardRefRenderFunction<BladeElementRef, BaseInputProps
               if (!isReactNative) {
                 inputRef.current?.focus();
               }
+              onBackdropClick?.();
             }}
             isTableInputCell={isTableInputCell}
           >
--- a/packages/blade/src/components/Input/DropdownInputTriggers/BaseDropdownInputTrigger.tsx
+++ b/packages/blade/src/components/Input/DropdownInputTriggers/BaseDropdownInputTrigger.tsx
@@ -316,6 +316,7 @@ const _BaseDropdownInputTrigger = (
       onChange={props.isSelectInput ? undefined : props.onInputValueChange}
       onKeyDown={props.onTriggerKeydown}
       size={props.size}
+      onBackdropClick={onTriggerClick}
       trailingInteractionElement={
         isAutoCompleteInHeader || (isInsideTableEditableCell && !isValidationStateNone) ? null : (
           <InputChevronIcon


packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -326,7 +326,10 @@ type BaseInputCommonProps = FormInputLabelProps &
      * @default true
      **/
     isTableInputCell?: boolean;
-
+    /**
+     * Callback to be invoked when the backdrop is clicked
+     */
+    onBackdropClick?: () => void;
```
saurabhdaware, line 332: Should we call this `onInputWrapperClick`? that way its clear on which element we're calling it. Backdrop is a bit misleading because in other places we use it in the context of faded overlay when modal / dropdown opens

saurabhdaware:
I think it stopped working 🙈  

https://github.com/user-attachments/assets/43d75bcb-a2f7-454f-9798-cf551ba2af01

packages/blade/src/components/Input/BaseInput/BaseInputVisuals.tsx
```diff
@@ -329,7 +329,13 @@ export const BaseInputVisuals = ({
             alignItems="stretch"
             alignSelf="stretch"
           >
-            {trailingInteractionElement}
+            <BaseBox
+              onClick={(e) => {
+                e.stopPropagation();
```
saurabhdaware, line 334: With this if we click the eye icon in password input, I'm guessing the focus won't come to input 🤔 

Can we add `onTrailingInteractionElementClick` and add click handler to overall container including that padding space?

Then we can remove the click handler on icon and use that prop itself for handling clicks on icon

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -326,7 +330,10 @@ type BaseInputCommonProps = FormInputLabelProps &
      * @default true
      **/
     isTableInputCell?: boolean;
-
+    /**
+     * Callback to be invoked when the backdrop is clicked
+     */
+    onInputWrapperClick?: (event: React.MouseEvent<HTMLInputElement>) => void;
```
saurabhdaware, line 336: Do we need this prop anymore? The trailingInteractionElement will handle the clicking and input already has onClick for rest of the part right?

tewarig: onTrailingInteractionElementClick will only work within the wrapper. If even the TrailingInteractionElement wrapper does not occupy the full space, an onInputWrapper click is still required.
image

tewarig: update: we decided to go with 
image

packages/blade/src/components/Input/BaseInput/BaseInputVisuals.tsx
```diff
@@ -314,24 +317,30 @@ export const BaseInputVisuals = ({
     );
   }
 
+  if (hasTrailingVisuals && hasTrailingInteractionElement) {
```
saurabhdaware, line 320: This block ends up early returning the jsx unlike the previous implementation. So when there is trailingInteractionElement + trailing visual, it seems to only render trailingInteractionElement and skip the code that comes after this.

You can see this in this example where when showClearButton is set to true, and you start typing, it removes the trailing visual button https://61c19ee8d3d282003ac1d81c-kxotdzpypj.chromatic.com/?path=/story/components-input-textinput--text-input-with-trailing-button&args=showClearButton:!true&globals=showInternalComponents:!true


First is behaviour in this PR, and second is current production behaviour

https://github.com/user-attachments/assets/b3d8c57c-2423-4d8c-96ed-b7e50e30b106

packages/blade/src/components/Input/BaseInput/BaseInputVisuals.tsx
```diff
@@ -314,24 +317,30 @@ export const BaseInputVisuals = ({
...
+          {trailingInteractionElement}
+        </BaseBox>
+      </BaseBox>
+    );
+  }
+
   if (hasTrailingVisuals) {
     return (
       <BaseBox alignSelf="stretch" alignItems="stretch" {...getVisualContainerStyles()}>
-        {hasTrailingInteractionElement ? (
```
saurabhdaware, line 320: see if you can keep this code and make minimal changes with conditions. baseinput has too many gotchas because its used in all inputs 😢

tewarig:
for testing - 
https://codesandbox.io/p/sandbox/razorpay-blade-basic-forked-3przv7?file=%2Fpackage.json%3A13%2C25-13%2C91

saurabhdaware (dismissed):

PR #2428, opened 2024-11-28
docs(Motion Presets): add motion presets documentation

saurabhdaware:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Morph/Morph.stories.tsx
```diff
@@ -69,3 +81,105 @@ const MorphTemplate: StoryFn<typeof Morph> = (args) => {
...
+  const [showPage, setShowPage] = React.useState(false);
+
+  return (
+    <Box minHeight="400px" height="100%">
+      <Button marginBottom="spacing.5" onClick={() => setShowPage(!showPage)}>
+        Toggle Morph
+      </Button>
+      <AnimatePresence>
+        {showPage ? (
+          <Morph layoutId="box-shape">
```
anuraghazra, line 145: Btw we should be able to morph between any N number of components right? 

This would be important so we can do things like this: https://codesandbox.io/p/sandbox/4uph8w

saurabhdaware: yup this is possible with morph

anuraghazra (approved):

PR #2434, opened 2024-12-02
fix: speed up stackblitz

saurabhdaware:
## Description

Speeds up stackblitz examples 

## Changes

- Move from node template to javascript template of stackblitz as it support autocomplete now
- Remove typescript from our examples as it throws runtime errors in javascript template

### DIFF
--- a/packages/blade/.storybook/react-native/Storybook.tsx
+++ b/packages/blade/.storybook/react-native/Storybook.tsx
@@ -8,7 +8,7 @@ import './storybook.requires';
 
 import { name as appName } from '../../app.json';
 
-const App = (): React.ReactElement => {
+const App = () => {
   const Storybook = getStorybookUI({
     shouldPersistSelection: true,
     // keeping in comments becuase this is not documented properly in the docs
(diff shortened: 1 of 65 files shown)


tewarig (commented):
I tried running fix/broken-stackblitz locally, and everything seems to work fine. However, in the Storybook preview URL, it always shows the error: Can't find packages: react, react-dom, @razorpay/blade.  Could you please check onces  🤔 
Also, can we update the StackBlitz SDK to version 1.11.0? Currently, we are using version 1.9.0.
Everything else seems fine to me.

saurabhdaware:
The installation not working in PRs seem to be some issue with codesandbox package registry. I am unable to access it outside of stackblitz either

image

tewarig:
> razorpay/blade: basic

working now

saurabhdaware:
I added a check to ensure it only adds import if code doesn't have `import React` already.

> Additionally, we need to verify that React is not being imported unnecessarily in other examples.

We need it in every example actually because this is a `javascript` template and requires React to compile JSX

tewarig (approved):

PR #2443, opened 2024-12-10
fix: recommend `framer-motion` name installation instead of `motion/react`

saurabhdaware:
## Description

So due the name change of framer-motion -> motion. There's several gotchas that we end up with on consumer depending on bundlers and package installers they are using. E.g.
- pnpm fails when you install subdependency because it has flat directory structure so there we have to provide extra custom config that resolves the correct package. But it works fine in yarn and npm
- vite is resolving motion and framer-motion as separate instances for some reasons so provider of motion + animation from framer-motion is not working correctly
- I tried to mark motion as external package in build but that too starts throwing other babel issues

There are configs and ways to solve these for each bundler and package installer but that makes our upgrade guide confusing and things can break with those hacky setups

So for now to simplify setups, I've changed docs to recommend framer-motion installation only as I can see them shipping new versions there. Since v11 is there under framer-motion name as well we should be good for some time now

## Changes

Changes instances from `motion` to `framer-motion` in docs and stackblitz

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/docs/guides/Installation.stories.mdx
+++ b/packages/blade/docs/guides/Installation.stories.mdx
@@ -93,7 +93,7 @@ Before you install the package, make sure that you have performed the following 
         <Text>Blade has a few peer dependencies that you may already have installed in your project. If so, you can skip adding them again.</Text>
 
         ```shell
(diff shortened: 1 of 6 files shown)


saurabhdaware:
@anuraghazra I am merging this into integration branch. You can add any comment / approval there.

PR #2450, opened 2024-12-24
fix(blade): fallback for useInsertionEffect

anuraghazra:
## Description

Fixes DSSUP-178

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/utils/useCallbackRef.ts
+++ b/packages/blade/src/utils/useCallbackRef.ts
@@ -1,8 +1,13 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable @typescript-eslint/explicit-function-return-type */
-import { useCallback, useInsertionEffect, useRef } from 'react';
+import React, { useCallback, useRef } from 'react';
 import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';
 
-const useInsertionEffectFallback = useInsertionEffect || useIsomorphicLayoutEffect;
+// Prevent webpack from importing this:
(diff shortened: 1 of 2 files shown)


kamleshchandnani (approved):

packages/blade/src/utils/useCallbackRef.ts
```diff
@@ -1,5 +1,8 @@
 /* eslint-disable @typescript-eslint/explicit-function-return-type */
 import { useCallback, useInsertionEffect, useRef } from 'react';
+import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';
+
+const useInsertionEffectFallback = useInsertionEffect || useIsomorphicLayoutEffect;
```
saurabhdaware, line 5: Wouldn't react 17 fail with useInsertionEffect not defined or no import error instead of it being undefined / falsy? 

Like does the function exist in react 17 but has falsy value?

anuraghazra: I'll check once, with React17, otherwise we can just do React.useInsertionEffect instead of importing it.

anuraghazra: Tested, worked in Vite failed in Webpack. 

Fixed it in https://github.com/razorpay/blade/pull/2450/commits/4ea8acf3b52bd86549c695b5d22f673cfee4a9ac

saurabhdaware (approved):
Looks good. Add changeset

saurabhdaware (approved):

PR #2453, opened 2024-12-26
docs(ListView): API Decision

saurabhdaware:
## Formatted Documents

- [Pattern Terminologies](https://github.com/razorpay/blade/blob/patterns/rfc/packages/blade/src/components/ListView/_decisions/pattern-terminology.md)
- [ListView Pattern API Decision](https://github.com/razorpay/blade/blob/patterns/rfc/packages/blade/src/components/ListView/_decisions/decisions.md)


packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,481 @@
...
+**Pros:**
+
+- Layout can be controlled on design-system end which means in future if we decide to change layout of the list view pattern, consumers will only have to upgrade version
+- Responsive layout can also be defined on design-system layer which other people might miss
+- Consistent layout across all pages
+- Functionality such as hide / show filters, placement of search, etc comes out of the box
+- Less verbose compared to API without layout component
+
+**Cons:**
+
```
chaitanyadeorukhkar, line 59: Is it really a con? Lets check with design if this will ever be needed to move things around within quick filters

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,481 @@
...
+- Too strict with very little flexibility
+- Adds up more things in tables which can make table difficult to maintain and will contain bundle of filters even when they are not used
+
+</details>
+
+## Enhancements / Components
+
+- Enhancements:
+  - Searchable Dropdown
+  - InputGroup (TBD)
```
chaitanyadeorukhkar, line 245: Where is this needed?

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,481 @@
...
+    </DropdownHeader>
+    <ActionList>
+      <ActionListItem />
+      <ActionListItem />
+      <ActionListItem />
+    </ActionList>
+</Dropdown>
+```
+
+### FilterChipGroup
```
chaitanyadeorukhkar, line 269: There will be different types of filter chips, right? Will this API suffice for all the types?

chaitanyadeorukhkar, line 269: How is the naming convention on Figma for this? Are we also call this FilterChipGroup & FilterChip?

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,481 @@
...
+        onClick: () => {},
+      },
+      {
+        title: 'Pending',
+        onClick: () => {},
+      },
+    ]}
+    onSearch={({ value, searchType }) => {
+      /* filter on search */
+    }}
```
anuraghazra, line 398: image


@saurabhdaware I believe we should keep dev side a bit more flexible with slots. 

For example in many places in X we don't need search but rather we have some action buttons. 

We can keep the QuickFilters and SearchGroup into separate slots which can be independently be used. 


```
      <ListView> // internally it will have justify-content: space-between
        <ListViewQuickFilters
          filters={[
            {
              title: 'All',
              onClick: () => {},
            },
            {
              title: 'Pending',
              onClick: () => {},
            },
          ]}
        />
        <Box>
          <ListViewFilterToggle />
          <ListViewSearchGroup />
        </Box>
      </ListView>
```

not exactly but something like this ^ this way I can use QuickFilters independently of ListView. 

Or 

```
      <ListView> // internally it will have justify-content: space-between
        <ListViewFilters>
          <QuickFilters
            filters={[
              {
                value: 'all',
                title: 'All',
                onClick: () => {},
              },
              {
                value: 'pending',
                title: 'Pending',
                onClick: () => {},
              },
            ]}
          />
          <ListViewFiltersActions> // control the gap also
            <ListViewFilterToggle />
            <ListViewSearchGroup />
          </ListViewFiltersActions>
        </ListViewFilters>
      </ListView>
```

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,481 @@
...
+```
+
+##### ListViewFilters
+
+<img src="image-4.png" width="600px" />
+
+```ts
+type QuickFilter = {
+  title: string;
+  onClick: (e: React.MouseEvent) => void;
```
anuraghazra, line 423: We will also need onChange, value & controlled/uncontrolled state for QuickFilters. 

Because QuickFilters will have the same semantics of RadioGroup it can also change on keyboard events not just click. 

Plus this will also be a requirement if we want this to work independently as i've mentioned [here](https://github.com/razorpay/blade/pull/2453/files#r1912960927)

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,481 @@
...
+      {
+        title: 'All',
+        onClick: () => {},
+      },
+      {
+        title: 'Pending',
+        onClick: () => {},
+      },
+    ]}
+    onSearchChange={() => {
```
anuraghazra, line 28: We need to make sure consumers are able to externally control all aspects of quickfilters/search etc. 

Quick filters, search, and dropdowns may be controlled or have initial values set via URL query parameters.

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,481 @@
...
+          value="2d"
+          onClick={({ name, value }) => setDuration(name)}
+        />
+        <ActionListItem
+          title="1 week"
+          value="1w"
+          onClick={({ name, value }) => setDuration(name)}
+        />
+        <ActionListItem
+          title="1 month"
```
anuraghazra, line 298: Wouldn’t this get quite cumbersome to write all these JSX even when we just have 2-3 filter chips, not to mention most production apps will have 8-12 chips

Are there any alternatives we can approach? 🤔 Since we already know the specific inputs involved, could we consider a more streamlined, config-driven, and focused API instead?

<FilterChipInput type="select-dropdown" options={[]} />

saurabhdaware (commented):
Ignore the irrelevant comments. Old comments. I forgot to submit these.

packages/blade/src/components/ListView/_decisions/decisions.md
saurabhdaware: Yup it might come in as we discussed earlier on call. Turning 2 individual inputs to InputGroup, moving items to other side, etc might come in future

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,481 @@
...
+
+- Too strict with very little flexibility
+- Adds up more things in tables which can make table difficult to maintain and will contain bundle of filters even when they are not used
+
+</details>
+
+## Enhancements / Components
+
+- Enhancements:
+  - Searchable Dropdown
```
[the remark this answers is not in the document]
saurabhdaware: Yup. It's AutoComplete inside DropdownOverlay instead of AutoComplete as trigger for Dropdown.

Something that we can use inside PhoneNumberInput as well to search for countries inside the Dropdown

packages/blade/src/components/ListView/_decisions/decisions.md
saurabhdaware: It was needed for search but we decided to deprioritise it and do it later
image

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,481 @@
...
+  - InputGroup (TBD)
+- New Components
```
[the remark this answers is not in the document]
saurabhdaware: Discussed it earlier in call. Also InputGroup is deprioritised and will be done later.

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,481 @@
...
+      {
+        title: 'Pending',
+        onClick: () => {},
+      },
+    ]}
+    onSearch={({ value, searchType }) => {
+      /* filter on search */
+    }}
+  >
+    <FilterGroup />
```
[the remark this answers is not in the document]
saurabhdaware: It's FilterChipGroup component. Will be fixed in next commit

packages/blade/src/components/ListView/_decisions/decisions.md
saurabhdaware: We discussed it that day. It'll make API a bit verbose and we still want to control the overall layout of this section on DS-side otherwise any change in position of items will require consumer migration. We can control it with some leading trailing type of component like we have in Card but they too will require migration if the terms Leading and Trailing become meaningless with new change.

packages/blade/src/components/ListView/_decisions/decisions.md
saurabhdaware: - Search will be optional
- Counter, Badge, etc

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,508 @@
...
+  >
+    <FilterChipGroup></FilterChipGroup>
+  </ListViewFilters>
+  <Table />
+</ListView>
+```
+
+#### Props
+
+##### ListView
```
chaitanyadeorukhkar, line 382: Weren't we going to make the search optional? Can't see any prop to hide search

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,508 @@
...
+- ### Do we want to control layout or not?
+
+  Currently proposed API assumes that we are controlling which item is placed where and thus proposes a Layout Component of ListView. I've mentioned the pros and cons of layout component. The API and alternate api is proposed above.
+
+  #### Tradeoffs of Layout Component
+
+  - If we go with Layout component, then the tradeoff is the educational effort of teaching consumer which part of the UI comes directly from blade as layout component.
+  - If we go without Layout component, tradeoff is not being able to update the layout changes on consumer if something updates in dev.
+
+- ### InputGroup
```
chaitanyadeorukhkar, line 499: Isn't this resolved now? We concluded to not do input group right?

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,508 @@
...
+
+  Currently we're calling it ListView because the terminology was used in Razorpay although this terminology is not really used anywhere outside of Razorpay and can cause confusion to new people who join.
+
+  Alternate Options
+
+  - ListView
+  - FilterView
+  - TableFilterView
+
+- ### Do we want to control layout or not?
```
chaitanyadeorukhkar, line 490: This is resolved right?

packages/blade/src/components/ListView/_decisions/decisions.md
saurabhdaware: Not adding `onSearch` would remove the search. I'll add a note. Don't think so we need another prop right? 🤔

packages/blade/src/components/ListView/_decisions/decisions.md
```diff
@@ -0,0 +1,520 @@
...
+      <Menu>
+        <FilterChip />
+        <MenuOverlay>
+          <TextInput />
+        </MenuOverlay>
+      </Menu>
+    </FilterChipGroup>
+  </ListViewFilters>
+  <Table data={{ nodes: filteredNodes }} />
+</ListView>
```
inaseem, line 43: Hi @saurabhdaware , if possible please also mention how states such as loading, error states are handled in list views. Would it be handled out of the box by the ListView pattern, or would the consumers will need to implement it on their own.

saurabhdaware: Oh good question. It will have to be handled on consumer-side since blade itself won't have context of your data fetching.

But I think we might have to define it on figma for designers. Let me check with our design team.

tewarig (approved):

PR #2464, opened 2025-01-08
fix: fileUpload onRemove clashes with onChange [FC-1234]

tewarig:
## Description

 fileUpload component's onRemove behavior clashes with onChange.

I believe onChange should be triggered when onRemove is called, but it should provide the latest value instead of returning outdated values. Currently, we are using onRemove to handle file removal. However, ideally, this should be managed by onChange, similar to how an input's onChange event reflects both the addition and removal of characters in a string.

## Additional Information

this pr should go with a major release. 
## Component Checklist

- [x] Add KitchenSink Story
- [x] Add changeset

### DIFF
--- a/.changeset/modern-geckos-relate.md
+++ b/.changeset/modern-geckos-relate.md
@@ -1,0 +1,5 @@
+---
+'@razorpay/blade': patch
+---
+
+fix: fileupload on remove clashes with onchage
--- a/.changeset/modern-geckos-relate.md
+++ b/.changeset/modern-geckos-relate.md
@@ -2,4 +2,4 @@
 '@razorpay/blade': patch
 ---
 
-fix: fileupload on remove clashes with onchage
+fix(FileUpload): fileupload on remove clashes with onchage
--- a/.changeset/modern-geckos-relate.md
+++ b/.changeset/modern-geckos-relate.md
@@ -3,3 +3,7 @@
 ---
 
 fix(FileUpload): fileupload on remove clashes with onchage
+
+> [!NOTE]
+>
+> Previously, onChange did not provide an updated value when a file was removed. Now, when a file is removed from FileUpload, onChange will return the updated value.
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/FileUpload/FileUpload.web.tsx
```diff
@@ -1,4 +1,5 @@
 import { useState, useCallback, useMemo, useRef, forwardRef } from 'react';
+import { flushSync } from 'react-dom';
```
anuraghazra, line 2: Is flushSync available in React17? If not we won't be able to use it as it will fail on consumer end.

tewarig: Yes, flushSync was initially added in React 16. I was able to find the PR: https://github.com/facebook/react/pull/10225. However, I think the team didn’t document it.

I have tested this branch on React 17. Here is the sandbox link:
https://codesandbox.io/p/sandbox/blade-testing-template-ky8wfs

saurabhdaware:
To confirm, the change post-this would be that onChange will get triggered with latest value after item is removed right? does anything change in arguments we pass to onRemove or arguments passed to onChange in other scenarios?

tewarig:
> To confirm, the change post-this would be that onChange will get triggered with latest value after item is removed right? does anything change in arguments we pass to onRemove or arguments passed to onChange in other scenarios?

yes. now onChange will  get triggered with latest value after removing the item.
no. regarding arguments nothing have changed

packages/blade/src/components/FileUpload/FileUpload.web.tsx
anuraghazra: Broken csb link ^

tewarig: @anuraghazra can you try it now? 
https://codesandbox.io/p/sandbox/blade-testing-template-ky8wfs

.changeset/modern-geckos-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@razorpay/blade': patch
+---
+
+fix: fileupload on remove clashes with onchage
```
saurabhdaware, line 5:
```suggestion
fix(FileUpload): fileupload on remove clashes with onchage
```


Also can you add a note here for consumers on what exactly we changed in behaviour?

packages/blade/src/components/FileUpload/FileUpload.web.tsx
anuraghazra: Cool

anuraghazra (dismissed):

.changeset/modern-geckos-relate.md
tewarig: @saurabhdaware note added

anuraghazra (approved):

Issue #2470, opened 2025-01-13
DatePicker: Month Picker


Akshat-Maheshwari:
**Chevron down** is not working in **Month picker**. <br />
<img width="363" alt="Screenshot 2025-01-13 at 7 39 34 PM" src="https://github.com/user-attachments/assets/45531c02-2265-4b0f-bdeb-16d029916528" />

Issue #2469, opened 2025-01-13
DatePicker: Month/Year Picker


Akshat-Maheshwari:
### Current: 
1. Month Picker: Display is a Date, for example: 01/(mon)/(year). <img width="693" alt="Screenshot 2025-01-13 at 7 27 03 PM" src="https://github.com/user-attachments/assets/85a46b18-ad95-4b69-bbdc-38431f98d04d" />
2. Year Picker: Display is a Date, for example: 01/01/(year). <img width="694" alt="Screenshot 2025-01-13 at 7 28 23 PM" src="https://github.com/user-attachments/assets/1b5d081e-36de-42ff-b1c2-dfc0cefb0e9d" />

 <br/>

### Expected:
1. Month Picker: Display should be month name. <img width="696" alt="Screenshot 2025-01-13 at 7 18 51 PM" src="https://github.com/user-attachments/assets/9ea83266-febe-4701-95ef-d899980b9e02" />
2. Year Picker:  Display should be only year  <img width="699" alt="Screenshot 2025-01-13 at 7 19 14 PM" src="https://github.com/user-attachments/assets/acd5365f-afc1-4fa8-9205-e3495afd7b48" />

### Reason:

The current format is misleading. 
For example, I have a download button which downloads all the reports for the selected month, so when I select February, the date picker displays 01/02/2025, which suggests I only need the report for that specific day, when in fact, I need the report for the entire month of February.

PR #2472, opened 2025-01-14
feat: add dateformat and placeholder prop for datepicker input [DSSUP-199]

tewarig:
## Description

- Added dateFormat and placeholder props:
The dateFormat prop allows custom formatting for the date displayed in the input.
The placeholder prop enables customization of the placeholder text, which defaults to the date format. This is useful since the placeholder text might differ from the date format or could be unclear to users.
- Removed the chevron icon when the picker prop is used.

## Additional Information
this fixes - 
https://github.com/razorpay/blade/issues/2469 
https://github.com/razorpay/blade/issues/2470 

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/DatePicker/CalendarHeader.web.tsx
+++ b/packages/blade/src/components/DatePicker/CalendarHeader.web.tsx
@@ -14,6 +14,7 @@ type CalendarHeaderProps = {
   isRange: boolean;
   date: DateValue | DatesRangeValue;
   pickerType: PickerType;
+  showLevelChangeLink?: boolean;
   onNextMonth: () => void;
   onPreviousMonth: () => void;
   onNextYear: () => void;
@@ -21,6 +22,35 @@ type CalendarHeaderProps = {
   onNextDecade: () => void;
   onPreviousDecade: () => void;
   onLevelChange: (level: MantineCalendarLevel) => void;
+};
+const CalendarLevelIndicator = ({
+  onClick,
+  showLevelChangeLink,
+  accessibilityLabel,
+  text,
+}: {
+  onClick: () => void;
+  showLevelChangeLink?: boolean;
+  accessibilityLabel: string;
+  text: string;
+}): React.ReactElement => {
+  return showLevelChangeLink ? (
+    <Link
+      onClick={onClick}
+      size="large"
+      variant="button"
+      color="neutral"
+      iconPosition="right"
+      icon={ChevronDownIcon}
+      accessibilityLabel={accessibilityLabel}
+    >
+      {text}
+    </Link>
+  ) : (
+    <Text size="large" weight="medium" color="interactive.text.neutral.normal">
+      {text}
+    </Text>
+  );
 };
 
 const CalendarHeader = ({
@@ -34,7 +64,10 @@ const CalendarHeader = ({
   onPreviousYear,
   onPreviousDecade,
   onLevelChange,
-}: CalendarHeaderProps): React.ReactElement => {
+  showLevelChangeLink,
+}: CalendarHeaderProps & {
+  showLevelChangeLink?: boolean;
+}): React.ReactElement => {
   const { i18nState } = useI18nContext();
   const locale = convertIntlToDayjsLocale(i18nState?.locale ?? 'en-IN');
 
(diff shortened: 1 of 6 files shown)


packages/blade/src/components/DatePicker/DateInput.web.tsx
```diff
@@ -92,12 +92,13 @@ const _DatePickerInput = (
     successText,
     errorText,
     helpText,
+    dateFormat,
+    placeholder,
```
saurabhdaware, line 96: Its fine to not expose placeholder right? dateFormat can have its own defined placeholder I'm guessing 🤔

packages/blade/src/components/DatePicker/types.ts
```diff
@@ -162,6 +162,14 @@ type DatePickerProps<Type extends DateSelectionType> = Omit<
      */
     onApply?: Type extends 'single' ? (value: DateValue) => void : (value: DatesRangeValue) => void;
     labelPosition?: BaseInputProps['labelPosition'];
+    /**
+     * Sets the date format to be displayed in the input field.
+     */
+    dateFormat?: string;
```
saurabhdaware, line 168: can we have more explit types like `'DD/MM/YYYY' | 'MMM'` etc?

packages/blade/src/components/DatePicker/DateInput.web.tsx
tewarig: I think we should expose the placeholder.
If someone wants to change it from "DD/MM/YYYY" to something like "Select Date," it should be allowed.
developers may want to modify the placeholder to suit their needs.
We should allow developers to modify the placeholder because having a predefined one could impose unnecessary constraints.

packages/blade/src/components/DatePicker/types.ts
tewarig: We can expose limited options here in dateFormat, but I think using a string is more ideal. What if the user wants to display the date like 'yyyy/mm/dd'? Using a string here allows more flexibility in how the date is formatted. Let me know if we should define an explicit type. 

There is one thing, though maybe I should change the default type based on the picker prop as well.

packages/blade/src/components/DatePicker/DateInput.web.tsx
anuraghazra: Yeah fine. just keep the default as it was before.

packages/blade/src/components/DatePicker/DateInput.web.tsx
```diff
@@ -109,7 +110,7 @@ const _DatePickerInput = (
   if (selectionType == 'single') {
     const dateValue = getFormattedDate({
       date,
-      format,
+      format: dateFormat,
```
anuraghazra, line 113: Can we just call this `format`?

<DatePicker dateFormat="" /> `date` sounds a bit redundant

tewarig: sure

packages/blade/src/components/DatePicker/DateInput.web.tsx
saurabhdaware: If we set placeholder to "Select Date" how would the end-user know which format they are supposed to enter date in 🤔 

Works to have placeholder for now since we'll have good default anyways. Can discuss it later.

packages/blade/src/components/DatePicker/types.ts
saurabhdaware: But in this case the consumer wouldn't get autocomplete on which formats are possible no? they'll have to read documentation for that.

> What if the user wants to display the date like 'yyyy/mm/dd'?

Also, this is something that we should avoid exposing. If one datepicker on one product is dd/mm/yyyy, other is mm/dd/yyyy, and other is yyyy/mm/dd the consumer will have different experiences of typing dates in all products so this is something that we can put guardrails on.

IMO we can support `dd/mm/yyyy | mmm | mmmm | yyyy`. This should cover most usecases while keeping the experience consistent.

> There is one thing, though maybe I should change the default type based on the picker prop as well.

We can if we want but its fine even if we skip this since I don't see anyone using day picker just to show month. That experience will look visibly broken so consumers won't do it most likely

packages/blade/src/components/DatePicker/DatePicker.web.tsx
```diff
@@ -83,6 +85,37 @@ const DatePicker = <Type extends DateSelectionType = 'single'>({
       onPickerChange?.(picker);
     },
   });
+  const finalFormat = React.useMemo(() => {
+    if (format) {
+      return format;
+    }
+    if (picker === 'day') {
+      return 'DD';
```
saurabhdaware, line 93: @anuraghazra Shouldn't day picker also be `dd/mm/yyyy` by default? just the date number won't be useful in most scenarios no?

Like I would normally use day picker when I want people to only choose a day of given month but I would still want user to know which month they've chosen right?

tewarig: okay, will change this

packages/blade/src/components/DatePicker/types.ts
tewarig: added explit formats

packages/blade/src/components/DatePicker/DateInput.web.tsx
tewarig: okay leaving placeholder for now. with default value DD/MM/YYYY.

packages/blade/src/components/DatePicker/DatePicker.web.tsx
tewarig: changed

saurabhdaware (dismissed):

anuraghazra (approved):
Mostly LGTM

saurabhdaware (approved):

packages/blade/src/components/DatePicker/CalendarHeader.web.tsx
```diff
@@ -135,34 +168,21 @@ const CalendarHeader = ({
       ) : (
         <Box display="flex" gap="spacing.5" alignItems="center">
           {pickerType === 'day' && (
-            <Link
-              onClick={() => {
-                onLevelChange('year');
-              }}
```
tewarig, line 141: need to change this

PR #2488, opened 2025-01-22
feat(SideNav): add description and onLevelChange callback

saurabhdaware:
## Description

- Add description for L2 items
- Add `onVisibleLevelChange` callback

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/SideNav/types.ts
+++ b/packages/blade/src/components/SideNav/types.ts
@@ -34,7 +34,7 @@ type SideNavProps = {
    * - Select the active link changes between L1 and L2 which can collapse or expand the L1
    * - When you hover / unhover L1 in collapsed state which can temporarily expand the L1
    */
-  onLevelChange?: ({ visibleLevel }: { visibleLevel: number }) => void;
+  onVisibleLevelChange?: ({ visibleLevel }: { visibleLevel: number }) => void;
 
   /**
    * Banner slot for usecases like adding Activation Panel
--- a/packages/blade/src/components/SideNav/types.ts
+++ b/packages/blade/src/components/SideNav/types.ts
@@ -27,9 +27,14 @@ type SideNavProps = {
    */
   onDismiss?: DrawerProps['onDismiss'];
 
+  /**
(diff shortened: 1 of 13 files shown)


tewarig:
Here, onLevelChange is being called twice when I am on the same level.

When I am on level 2, I hover over level 1 items, and the level gets changed, triggering onLevelChange. Then I click on a level 1 item. Although the level has not changed, onLevelChange is still being called.

https://github.com/user-attachments/assets/7f451046-fa41-4862-a065-b77c069fb702

packages/blade/src/components/SideNav/types.ts
```diff
@@ -42,6 +51,13 @@ type SideNavLinkProps = {
    */
   title: string;
 
+  /**
+   * description of the Link
+   *
+   * **Note**: Only applicable for L2 items
+   */
+  description?: string;
```
tewarig, line 59: can we have a check for this ? we can throw warning if description is provided to L2 items ?

saurabhdaware: makes sense. We can throw dev-only error. I was avoiding production error because if backend API mistakenly sends description then we don't want the sidenav to error out completely.

saurabhdaware:
> When I am on level 2, I hover over level 1 items, and the level gets changed, triggering onLevelChange. Then I click on a level 1 item. Although the level has not changed, onLevelChange is still being called.

So I was kind of thinking onLevelChange as callback that also gets triggered when the active level changes. So in future we can have argument of `activeLevel` alongside `visibleLevel`. For the scenario in video, the activeLevel changes when you click the item in L1.

But yes without the activeLevel argument it looks confusing. What do you think should we add `activeLevel` argument right now or should we only cover visibility change as activeLevel change can also be derived from which item is selected.

saurabhdaware:
@tewarig Renamed the prop `onVisibleLevelChange` and calling it only once now when L1 is hovered and clicked

anuraghazra (dismissed):

anuraghazra (approved):

tewarig (approved):

PR #2497, opened 2025-01-27
chore: memoized otp input and added example for using onOTPFilled

tewarig:
## Description

This PR fixes the issue with the OTPInput component where the onOTPFilled callback was causing infinite re-renders.

To avoid re-renders, we can wrap OTPInput in React.memo. I have implemented this. Additionally, the function passed to onOTPFilled should be wrapped in useCallback. I have added an example for this in the documentation.

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Input/OTPInput/OTPInput.tsx
+++ b/packages/blade/src/components/Input/OTPInput/OTPInput.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useImperativeHandle, useState } from 'react';
+import React, { memo, useEffect, useImperativeHandle, useState } from 'react';
 import type { BaseInputProps } from '../BaseInput';
 import { BaseInput } from '../BaseInput';
 import { getHintType } from '../BaseInput/BaseInput';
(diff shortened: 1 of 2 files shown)


.changeset/empty-comics-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@razorpay/blade': major
```
saurabhdaware, line 2: major :o

saurabhdaware (commented):
@tewarig what is this PR? were there some perf issues?

tewarig:
> @tewarig what is this PR? were there some perf issues?


it was related to this

https://razorpay.atlassian.net/browse/DSSUP-190?atlOrigin=eyJpIjoiZjQ4MTllNGQ3YzkyNGI1MWJmMTAyNTE2NmYxMGVjZTEiLCJwIjoiaiJ9

packages/blade/src/components/Input/OTPInput/OTPInput.tsx
```diff
@@ -419,7 +419,7 @@ const _OTPInput: React.ForwardRefRenderFunction<HTMLInputElement[], OTPInputProp
   );
 };
 
-const OTPInput = React.forwardRef<HTMLInputElement[], OTPInputProps>(_OTPInput);
+const OTPInput = memo(React.forwardRef<HTMLInputElement[], OTPInputProps>(_OTPInput));
```
saurabhdaware, line 422: What is the cause of re-renders though? Shouldn't it not re-render independent of whether we have memo or not?

PR #2508, opened 2025-01-29
chore: update chatMessage api decision

tewarig:
## Description

feat: chat bubble api decision #2508

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/ChatMessage/_decision.md
```diff
@@ -0,0 +1,85 @@
...
+<ChatMessage cardBody={<SomeComponent/>} />
+
+
+
+// Markdown
+<ChatMessage markdown="# this is markdown" />
+```
+
+## Open Questions
+
```
saurabhdaware, line 84: So normally we keep the questions in open questions and add conclusion and discussion points based on discussion so that we have reasoning in place if we want to find it later

packages/blade/src/components/ChatMessage/_decision.md
```diff
@@ -0,0 +1,85 @@
+# ChatMessage Decisions
```
saurabhdaware, line 1: You can get the format from other API decision

packages/blade/src/components/ChatMessage/_decision.md
```diff
@@ -0,0 +1,85 @@
...
+
```
saurabhdaware, line 2: You can add annotations image from figma cover of this component like we do in other API decisions

packages/blade/src/components/ChatMessage/_decision.md
```diff
@@ -0,0 +1,85 @@
...
+- [Design](#design)
+- [ChatMessage Component](#ChatMessage-component)
+  - [ChatMessage API](#ChatMessage-api)
+- [Open Questions](#open-questions)
+
+## Design
+
+[Figma Link](https://www.figma.com/design/jubmQL9Z8V7881ayUD95ps/Blade-DSL?node-id=100413-32686&t=n9A7LztwEkIsly3v-0)
+
+## ChatMessage Component
```
saurabhdaware, line 12: We usually keep Usage section first to give idea of the API and then add props table and then more examples of usage

packages/blade/src/components/ChatMessage/_decision.md
```diff
@@ -0,0 +1,85 @@
...
+</Move>
+
+// with card
+<ChatMessage cardBody={<SomeComponent/>} />
+
+
+
+// Markdown
+<ChatMessage markdown="# this is markdown" />
+```
```
saurabhdaware, line 81: We can later add pros and cons and why we went with that children API

saurabhdaware (approved):
The API looks good to me. 

Some comments on doc format but lets pick them up after FTX. You can create a github issue to keep track so that we can fix it later.

packages/blade/src/components/ChatMessage/_decision.md
```diff
@@ -0,0 +1,85 @@
...
+| senderType          | 'self' or 'other'                              | self   | No       | we will add  different styles based on this bubble                              |
+| isLoading              | Boolean                                     | false   | No       | If the message is loading, we will add a loading animation to the chat bubble                                                                                   |
+| validationState                | 'error' or 'none'                              | none   | No       |  if validation state is error we will show error decoration and message|
+| errorText              | String                                      | null    | No       | If the message is an error message, we will show the error message in the chat bubble                                                                                   |
+| onClick           | Function                                    | null    | No       | callback to be called when ever component is clicked.                                                                                                                          |
```
snitin315, line 25: Does it need to be interactive/clickable? If we render a Card inside the slot, it would support its own `onClick` prop.

packages/blade/src/components/ChatMessage/_decision.md
```diff
@@ -0,0 +1,85 @@
...
+| footerActions        | ReactNode | null    | No       | if this is passed we will render this at the end of chat bubble                               |
```
snitin315, line 26: Can you add a sample usage for this `prop` as well? Usually we follow compound API for rendering footer/headers.

packages/blade/src/components/ChatMessage/_decision.md
```diff
@@ -0,0 +1,85 @@
...
+<Move>
+  <ChatMessage senderType="self">Demo Text</ChatMessage>
+</Move>
+
+// with error
+<ChatMessage validationState="error" errorText="Error Message">Demo Text</ChatMessage>
+
+// with card
+<ChatMessage><Card></Card></ChatMessage>
+
```
snitin315, line 60: How about this API?
```jsx
// normal string text
<ChatMessage>string message</ChatMessage>

// With a complex body like Mardown/Card
<ChatMessage>
  <ChatMessageBody>
     <Markdown>  Demo Text </Markdown>
     <Card></Card>
  </ChatMessageBody>
  <ChatMessageFooter>
   <Button> Get STarted</Button>
  </ChatMessageFooter>

</ChatMessage>

```

packages/blade/src/components/ChatMessage/_decision.md
```diff
@@ -0,0 +1,85 @@
...
+| errorText              | String                                      | null    | No       | If the message is an error message, we will show the error message in the chat bubble                                                                                   |
+| onClick           | Function                                    | null    | No       | callback to be called when ever component is clicked.                                                                                                                          |
+| footerActions        | ReactNode | null    | No       | if this is passed we will render this at the end of chat bubble                               |
+| children               |  'ReactNode' or 'string'                          | null    | yes      | The children that will be rendered inside the chat bubble. can be react node or a string                                                                        |
+| leading             | ReactNode                                   | null    | No       | will be displayed only in case  if message is other and also, will contain animation by default                                   |
```
snitin315, line 28: I think we need a separate prop for just controlling typing animation as well, because otherwise if the component re-mounts/re-renders, the text will animate once again.


https://github.com/user-attachments/assets/815da708-e573-43ae-9c31-56a03b7a0c50

tewarig: typing animation will be controlled  by user.

packages/blade/src/components/ChatMessage/_decision.md
tewarig: Yes, in case of an error, we need to support the onClick prop since children accepts JSX. It should support its own onClick, but in that case, we need to stop event bubbling.

packages/blade/src/components/ChatMessage/_decision.md
tewarig: We would just be adding an animation to the leading.

packages/blade/src/components/ChatMessage/_decision.md
```diff
@@ -0,0 +1,85 @@
...
+
+## ChatMessage Component
+
+This will be our main component that will be used to render the chat bubble.
+
+## ChatMessage API
+
+| Prop                   | Type                                        | Default | Required | Description                                                                                                                                                     |
+| ---------------------- | ------------------------------------------- | ------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| messageType          | 'last' or  'default'                                | default   | No       | If the message is the last message in the chat and if this prop is enabled we will add decorations messageBubble                                                |
```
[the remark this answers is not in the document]
tewarig: update it to - 
If  messageType is  lastmessage in the chat we will add different styles in chat message. currently we have different borderRadius if messageType is last or default                                                |

packages/blade/src/components/ChatMessage/_decision.md
tewarig: added - 
  <ChatMessage
     footerActions={
       <ChipGroup>
         <Chip value="value1">value1</Chip>
         <Chip value="value2">value2</Chip>
       </ChipGroup>
     }
   >
   demo message
   </ChatMessage>

packages/blade/src/components/ChatMessage/_decision.md
tewarig: updated to - 
It will be displayed only if the message is 'other' and will also contain an spin animation by default.

anuraghazra:
Why is this still open?

tewarig:
> Why is this still open?

need to format few things  🤔

anuraghazra (approved):

PR #2513, opened 2025-01-31
feat(blade): chat message component

tewarig:
## Description

feat: add chat bubble component 

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/ChatBubble/ChatBubble.web.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
+import type { BladeElementRef } from '~utils/types';
+
+const _ChatBubble = (
+  {
+    messageType = 'default',
+    senderType = 'self',
+    isLoading = false,
+    validationState = 'none',
+    errorText = '',
```
anuraghazra, line 17:
```suggestion
    errorText,
```

packages/blade/src/components/ChatBubble/ChatBubble.web.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { SelfMessageBubble } from './SelfMessageBubble';
+import { DefaultMessageBubble } from './DefaultMessageBubble';
+import type { ChatBubbleProps } from './types';
+import { Text } from '~components/Typography';
+import BaseBox from '~components/Box/BaseBox';
+import { getStringFromReactText } from '~utils/getStringChildren';
+import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
+import type { BladeElementRef } from '~utils/types';
+
+const _ChatBubble = (
```
anuraghazra, line 11: Are the types correct? Let's follow the same convention here like we do in other places: 

React.ForwardRefRenderFunction<BladeElementRef, ChatBubbleProps>

packages/blade/src/components/ChatBubble/ChatBubble.web.tsx
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  return (
+    <BaseBox {...props} ref={ref as never}>
+      {senderType === 'self' ? (
+        <SelfMessageBubble
+          isError={validationState === 'error'}
+          onClick={onClick}
+          errorText={errorText}
+          children={childrenToRender()}
```
anuraghazra, line 55: Check once: May cause unnecessary rerenders

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
```diff
@@ -0,0 +1,66 @@
...
+  errorText = 'Message not sent. Tap to retry.',
+  onClick,
+  messageType,
+}: {
+  children: React.ReactNode | string;
+  isError?: boolean;
+  errorText?: string;
+  onClick?: () => void;
+  messageType: 'last' | 'default';
+}): React.ReactElement => {
```
anuraghazra, line 18: type SelfMessageBubbleProps = {} 

And also keep internal component's types similar to the main component such that we can just do `SelfMessageBubbleProps = Pick<CommonProps, 'children' | 'validationState' | 'errorText'....>`

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <BaseBox
+      display="flex"
+      gap="10px"
+      flexDirection="column"
+      onClick={onClick}
```
anuraghazra, line 24: when onClick is present we should change the chat bubble to become a button for a11y reasons.

packages/blade/src/components/ChatBubble/types.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { StyledPropsBlade } from '~components/Box/styledProps';
+import type { DataAnalyticsAttribute, StringChildrenType, TestID } from '~utils/types';
+
+type CommonChatBubbleProps = {
+  isLoading?: boolean;
+  validationState?: 'error' | 'none';
+  errorText?: string;
+  onClick?: () => void;
+  footerActions?: React.ReactNode;
+  children?: React.ReactNode | StringChildrenType;
```
anuraghazra, line 11: If we are adding React.ReactNode do we need to explicitly mention StringChildrenType?

packages/blade/src/components/ChatBubble/types.ts
```diff
@@ -0,0 +1,42 @@
...
+type DefaultMessageBubbleProps = {
+  children: React.ReactNode | string;
+  leading?: React.ReactNode;
+  isLoading?: boolean;
+  onClick?: () => void;
+};
+
+type SenderChatBubbleProps = CommonChatBubbleProps & {
+  senderType: 'self';
+  messageType: 'default' | 'last';
```
anuraghazra, line 25: Add jsdoc comments for inline documentation in these props. 

Like what is the default value of messageType, how to use footerActions etc, see other components to get better idea.

packages/blade/src/components/ChatBubble/__tests__/ChatBubble.web.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { ChatBubble } from '../ChatBubble.web';
+import renderWithTheme from '~utils/testing/renderWithTheme.web';
+import { RayIcon } from '~components/Icons';
+import { Card, CardBody } from '~components/Card';
+import { Box } from '~components/Box';
+import { Text } from '~components/Typography';
+import { Radio, RadioGroup } from '~components/Radio';
+
+describe('<ChatBubble/>', () => {
+  it('should render last message correctly', () => {
```
anuraghazra, line 11: add onClick, validation message etc tests

packages/blade/src/components/ChatBubble/ChatBubble.web.tsx
```diff
@@ -0,0 +1,75 @@
...
+        />
+      )}
+      {footerActions}
+    </BaseBox>
+  );
+};
+
+const ChatBubble = assignWithoutSideEffects(React.forwardRef(_ChatBubble), {
+  displayName: 'ChatBubble',
+  componentId: 'ChatBubble',
```
anuraghazra, line 73: Extract out the componetnId into `MetaConstants` and we generally use kebab case for componentIds

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
```diff
@@ -0,0 +1,66 @@
...
+          isError ? 'feedback.background.negative.subtle' : 'surface.background.primary.subtle'
+        }
+        padding="spacing.4"
+        borderTopLeftRadius="large"
+        borderTopRightRadius="large"
+        borderBottomLeftRadius="large"
+        borderBottomRightRadius={messageType === 'last' ? 'none' : 'large'}
+        width="fit-content"
+        height="auto"
+        word-break="break-word"
```
anuraghazra, line 39: ??
```suggestion
        wordBreak="break-word"
```

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
```diff
@@ -0,0 +1,66 @@
...
+      {isError && (
+        <BaseBox
+          maxWidth="240px"
+          display="flex"
+          justifyContent="flex-start"
+          alignItems="center"
+          gap="4px"
+        >
+          <AlertCircleIcon size="small" color="feedback.icon.negative.intense" />
+          <Text
```
anuraghazra, line 52: Use the existing `HintText` if possible. components/Form/FormHint

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <BaseBox
+      display="flex"
+      gap="10px"
+      flexDirection="column"
+      onClick={onClick}
+      cursor={isError ? 'pointer' : 'default'}
+    >
+      <BaseBox
+        maxWidth="240px"
```
anuraghazra, line 28: is this maxWidth set from design side? Shouldn't this be controlled by the consumer and our chatbubble be fluid to accommodate any UI?

packages/blade/src/components/ChatBubble/DefaultMessageBubble.web.tsx
```diff
@@ -0,0 +1,26 @@
...
+import BaseBox from '~components/Box/BaseBox';
+
+const DefaultMessageBubble = ({
+  children,
+  leading,
+  isLoading,
+  onClick,
+}: DefaultMessageBubbleProps): React.ReactElement => {
+  return (
+    <BaseBox maxWidth="296px" onClick={onClick}>
```
anuraghazra, line 13: same: button if onClick

packages/blade/src/components/ChatBubble/Rotate.web.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children,
+  animate,
+}: {
+  children: React.ReactElement;
+  animate?: boolean;
+}): React.ReactElement => {
+  const { theme } = useTheme();
+
+  return (
+    <LazyMotion features={domAnimation}>
```
anuraghazra, line 18: @saurabhdaware now that blade internally uses motion.div, do we need to put this LazyMotion on top level blade provider or can we skip and expect consumers to add motion providers?

saurabhdaware: https://github.com/razorpay/blade/blob/master/packages/blade/docs/migration-docs/upgrade-v12.md

LazyMotion itself is added on consumer-level only so ideally we should remove it from here. I've already added it to our Storybook so our examples should work fine without this

packages/blade/src/components/ChatBubble/index.ts
```diff
@@ -0,0 +1 @@
+export * from './ChatBubble';
```
saurabhdaware, line 1: Only export the expected components and types instead of exporting all from here

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
tewarig: i think we don't need this.

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
tewarig: Yes, maxWidth is set from the design for now. It should be, but currently, Bubble should only be used in specific UIs

packages/blade/src/components/ChatBubble/ChatBubble.web.tsx
```diff
@@ -0,0 +1,75 @@
...
+          errorText={errorText}
+          children={childrenToRender()}
+          messageType={messageType}
+        />
+      ) : (
+        <DefaultMessageBubble
+          children={childrenToRender()}
+          leading={leading}
+          onClick={onClick}
+          isLoading={isLoading}
```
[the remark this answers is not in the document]
tewarig: their won't be error in case of DefaultMessageBubble , even if their is error it will be handled in message reply, like The server is busy. Please try again later

packages/blade/src/components/ChatBubble/DefaultMessageBubble.web.tsx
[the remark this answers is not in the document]
tewarig: i think we don't need 296px onces. it was for wrapper. 
chatMessage have a set width of 256px

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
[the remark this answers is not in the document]
tewarig: no, i used wrong width before this... checked in design it should be 240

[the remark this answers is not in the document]
tewarig: for default message - 
Screenshot 2025-02-04 at 10 12 12 AM
for self message 
Screenshot 2025-02-04 at 10 12 31 AM

packages/blade/src/components/ChatBubble/ChatBubble.web.tsx
```diff
@@ -0,0 +1,75 @@
...
+    footerActions,
+    children,
+    leading,
+    loadingText,
+    ...props
+  }: ChatBubbleProps,
+  ref: React.Ref<BladeElementRef>,
+): React.ReactElement => {
+  const childrenToRender = (): React.ReactElement => {
+    // their can be a case where childrens are passed like  "{' '} some text" so we need to check if children is string or not
```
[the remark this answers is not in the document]
tewarig: I am already using getStringFromReactText inside childrenToRender, but before using it, I need to check if the children are strings or not.

packages/blade/src/components/ChatBubble/types.ts
tewarig: seems to work fine. removing it

packages/blade/src/components/ChatBubble/ChatBubble.web.tsx
```diff
@@ -0,0 +1,75 @@
...
+        </Text>
+      );
+    }
+    return children as React.ReactElement;
+  };
+
+  return (
+    <BaseBox {...props} ref={ref as never}>
+      {senderType === 'self' ? (
+        <SelfMessageBubble
```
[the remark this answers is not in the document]
tewarig: No, I don't think there is much common code between them. In the case of a self-message, we are using a bubble, and in the case of a default message, we are using an avatar with text. If both were bubbles, then there would be a lot of common code between them.

packages/blade/src/components/ChatMessage/docs/ChatMessage.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+            </Box>
+          </CardBody>
+        </Card>
+      </ChatMessage>
+    </Box>
+  );
+};
+
+export const ChatMessageBody = ChatMessageBodyTemplates.bind({});
+ChatMessageBody.storyName = 'Chat Message Body';
```
snitin315, line 158: Can you add a story for `footerActions` Prop as well?

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
tewarig: done.

packages/blade/src/components/ChatMessage/docs/ChatMessage.stories.tsx
tewarig: done

packages/blade/src/components/ChatBubble/__tests__/ChatBubble.web.test.tsx
tewarig: added

packages/blade/src/components/ChatBubble/Rotate.web.tsx
tewarig: removed lazyMotion

packages/blade/src/components/ChatBubble/ChatBubble.web.tsx
tewarig: seems fine to me.. refactored a bit

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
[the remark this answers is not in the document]
saurabhdaware: @tewarig can you check with design if this is expected? what is the reason to have 2 different max-widths?

packages/blade/src/components/ChatMessage/token.ts
```diff
@@ -0,0 +1,20 @@
...
+  self: {
+    maxWidth: '240px',
+    padding: 'spacing.4',
+    borderTopLeftRadius: 'large',
+    borderTopRightRadius: 'large',
+    borderBottomLeftRadius: 'large',
+    borderBottomRightRadiusForLastMessage: 'none',
+    borderBottomRightRadius: 'large',
+    backgroundColor: {
+      default: 'surface.background.primary.subtle',
```
saurabhdaware, line 14: nice 🔥

packages/blade/src/components/ChatMessage/__tests__/ChatBubble.web.test.tsx
saurabhdaware, line 1: I think some filenames are old ones

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
anuraghazra: I don't think adding maxWidth is a good idea, we did the same thing with Accordion and it backfired causing a lot of issues because: 

Once you add maxWidth removing it in future becomes kind of a styling breakage for consumers. 

It should be fluid and expose maxWidth so consumers can add it.

packages/blade/src/components/ChatMessage/docs/ChatMessage.stories.tsx
```diff
@@ -0,0 +1,270 @@
...
+  parameters: {
+    docs: {
+      page: Page,
+    },
+  },
+} as Meta<ChatMessageProps>;
+
+const ChatMessageTemplate: StoryFn<typeof ChatMessage> = () => {
+  return (
+    <ChatMessage senderType="self" messageType="last">
```
saurabhdaware, line 56:
```suggestion
const ChatMessageTemplate: StoryFn<typeof ChatMessage> = (args) => {
  return (
    <ChatMessage {...args}>
```

This plus
```jsx
export const Default = ChatMessageTemplate.bind({});
Default.args = {
  senderType: 'self',
  messageType: 'last'
}
```

This will make the storybook arguments work
image

packages/blade/src/components/ChatMessage/types.ts
```diff
@@ -0,0 +1,95 @@
...
+   * children prop is used to show content in chat message.
+   * can be string or react node.
+   * */
+  children?: React.ReactNode;
+  /**
+   * SenderType prop is used to show chat message as self or other.
+   * self: chat message will be shown as self. you can show error state and footer actions.
+   * other: chat message will be shown as other. you can show loading state and leading icon.
+   */
+  SenderType?: 'self' | 'other';
```
anuraghazra, line 47:
```suggestion
  senderType?: 'self' | 'other';
```

packages/blade/src/components/ChatBubble/SelfMessageBubble.web.tsx
saurabhdaware: It should still have a default maxWidth and then expose to consumers right? otherwise the messages will overflow till the end of chatbox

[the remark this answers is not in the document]
tewarig: checked with design . We need two different max-width values, but we are removing max-width from the token. We will allow consumers to pass it instead

tewarig: checked with design, we can remove maxWidth, allow consumers to pass it.

anuraghazra: > It should still have a default maxWidth and then expose to consumers right?

No. That's what caused the [unintuitive](https://razorpay.slack.com/archives/CMQ3RBHEU/p1738575481446369?thread_ts=1738575045.916759&cid=CMQ3RBHEU) issues that consumers faced. 


> It should still have a default maxWidth and then expose to consumers right? otherwise the messages will overflow till the end of chatbox

<Box width="500px">
  <ChatBubble />
</Box>

ChatBubble should fluidly take it's parent's width.

snitin315:
footerActions are not aligned correctly by default.

Screenshot 2025-02-06 at 9 23 21 AM

tewarig:
> footerActions are not aligned correctly by default like children.
> 
> Screenshot 2025-02-06 at 9 23 21 AM

you can follow this story, https://61c19ee8d3d282003ac1d81c-lbirdazznk.chromatic.com/?path=/story/components-chatmessage--chat-message-with-footer-actions 

you need to wrap footer actions in box . then you can wrap them as you want.

saurabhdaware (approved):

saurabhdaware (approved):

PR #2514, opened 2025-01-31
feat(DropdownIconButton): add support for DropdownIconButton and tooltip

saurabhdaware:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Dropdown/DropdownButton.tsx
```diff
@@ -42,46 +46,48 @@ const _DropdownButton = ({
   return (
     // Using BaseButton here to avoid exporting onBlur and onKeyDown from Button
     // If in future we decide to export onBlur and onKeyDown on Button, this can be replaced with Button
```
anuraghazra, line 48: Why in future, should we expose them? Seems fine to expose really.

packages/blade/src/components/Dropdown/DropdownButton.tsx
```diff
@@ -42,46 +46,48 @@ const _DropdownButton = ({
...
+          onClick?.(e as any);
+        }}
+        onBlur={(e) => {
+          // With button trigger, there is no "value" as such. It's just clickable items
+          // Setting it for web fails it on native typecheck and vice versa
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion, @typescript-eslint/no-explicit-any
+          onBlur?.(e as any);
+        }}
+        onKeyDown={(e) => {
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion, @typescript-eslint/no-explicit-any
```
anuraghazra, line 83: Globally ignore this rule for this file.

packages/blade/src/components/Dropdown/DropdownIconButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { TooltipifyComponent } from '~utils/TooltipifyComponent';
+
+type DropdownIconButtonProps = Omit<IconButtonProps, 'onClick'> & {
+  onBlur?: BaseButtonProps['onBlur'];
+  onKeyDown?: BaseButtonProps['onKeyDown'];
+  onClick?: IconButtonProps['onClick'];
+  tooltip?: TooltipifyComponentProps['tooltip'];
+};
+
+const _DropdownIconButton = ({
```
anuraghazra, line 19: I wish Dropdown worked like Menu and didn't require extra wrapper components to be exposed 😞

packages/blade/src/components/Dropdown/DropdownIconButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+        accessibilityProps={{
+          label: accessibilityLabel,
+          hasPopup: getActionListContainerRole(hasFooterAction, 'DropdownIconButton'),
+          expanded: isOpen,
+          controls: `${dropdownBaseId}-actionlist`,
+          activeDescendant: activeIndex >= 0 ? `${dropdownBaseId}-${activeIndex}` : undefined,
+        }}
+        onClick={(e) => {
+          onTriggerClick();
+          // Setting it for web fails it on native typecheck and vice versa
```
anuraghazra, line 62: Ignore for the whole file

packages/blade/src/components/Dropdown/docs/DropdownWithButton.stories.tsx
```diff
@@ -271,4 +284,53 @@ InternalLinkDropdown.parameters = {
...
+  const [isDropdownOpen, setIsDropdownOpen] = React.useState(false);
+
+  return (
+    <Box padding="spacing.10">
+      <Dropdown onOpenChange={setIsDropdownOpen} isOpen={isDropdownOpen}>
+        <DropdownIconButton
+          icon={BoxIcon}
+          tooltip={{ content: 'Check status' }}
+          accessibilityLabel="Status Dropdown"
+        />
```
anuraghazra, line 298: Do we need to expose a tooltip prop? 
It feels a bit off from an API perspective since I’d expect the component to support seamless composition and work directly with the Tooltip component - especially since it's a compound component. Something like:

```ts
<Dropdown>
  <Tooltip><DropdownButton /></Tooltip>
</Dropdown>
```

snitin315: +1 to Anurag.

packages/blade/src/components/Dropdown/DropdownButton.tsx
saurabhdaware: Oh actually I realised we'll also have to expose accessibility props so its not just these handlers. I'll update the comment but we can probably do it later. Need to figure out if we should expose all accessibility props or somehow limited props with limited values, etc.

packages/blade/src/components/Dropdown/DropdownIconButton.tsx
saurabhdaware: Yes. I feel at some point, we should just expose isSelected on MenuItem and move all these usecases to Menu and keep Dropdown as Select and AutoComplete only.

packages/blade/src/components/Dropdown/docs/DropdownWithButton.stories.tsx
saurabhdaware: That is what I was thinking too but it breaks too many of Dropdown's assumptions because now Dropdown can't figure out what children it has.

One way is we go the nested route and find the eventual children of Tooltip but that feels too hacky and delicate so in this case I'm following what we followed in SideNavLink. There too we had similar issue and went with tooltip as a prop

packages/blade/src/components/Dropdown/DropdownIconButton.tsx
saurabhdaware: done

packages/blade/src/components/Dropdown/docs/DropdownWithButton.stories.tsx
saurabhdaware: Discussed with Anurag. We can follow what we follow on Menu. So Tooltip can be wrapped around the Dropdown component itself with InteractiveWrapper-

```jsx
<Tooltip content="Check status">
  <TooltipInteractiveWrapper>
    <Dropdown>
      <DropdownIconButton />
      <DropdownOverlay>
      </DropdownOverlay>
    </Dropdown>
  </TooltipInteractiveWrapper>
</Tooltip>
```

packages/blade/src/utils/TooltipifyComponent.tsx
```diff
@@ -0,0 +1,43 @@
...
+ *     <button>Click me</button>
+ *    </TooltipifyComponent>
+ *   )
+ *
+ * }
+ * ```
+ *
+ * This adds a tooltip to the button when `tooltip` prop is passed. Otherwise renders the children as is
+ */
+const TooltipifyComponent = ({
```
anuraghazra, line 27: Is this exposed to consumers?

anuraghazra (approved):

packages/blade/src/utils/TooltipifyComponent.tsx
saurabhdaware: nope internal util. Using it in SideNavLink and we might need this if we decide to add tooltip to ActionListItem

PR #2521, opened 2025-02-04
fix(FileUpload): revert to not calling onChange on file remove

saurabhdaware:
## Description

Issue Context: https://razorpay.slack.com/archives/CMQ3RBHEU/p1738568779980839?thread_ts=1738140484.133379&cid=CMQ3RBHEU

Timeline:

In 11.34.1: We did not call `onChange` on removing of file. Only `onRemove` was called
In 11:36.2: We added dispatchEvent call which started calling onChange on onRemove
In 12.4.0: We released a fix thinking onChange call was expected behaviour and we just updated the state value for it

So for consumers upgrading from 11.34.1 -> 12.4.0, it becomes breaking change because onChange was not previously called

### Why is onChange getting called on dispatchEvent?

Hmm this is wild. So React apparently treats dispatchEvent with input type="text" differently than it treats input type="file"

Reproduction: https://codesandbox.io/p/sandbox/friendly-ishizaka-yk7mm3

I've added more details in issue here - https://github.com/facebook/react/issues/32300

## Changes

- For now, I am removing the dispatchEvent from FileUpload to unblock the dashboard team. We'll have to add it again though for analytics sdk
- I am also removing flushSync changes because we added it to fix onChange arguments which is not needed now

New behaviour: We'll revert back to behaviour of 11.34.1 where onChange is not called on removing file

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/FileUpload/FileUpload.web.tsx
+++ b/packages/blade/src/components/FileUpload/FileUpload.web.tsx
@@ -1,5 +1,4 @@
 import { useState, useCallback, useMemo, useRef, forwardRef } from 'react';
-import { flushSync } from 'react-dom';
 import type { FileUploadProps, BladeFile, BladeFileList } from './types';
 import { StyledFileUploadWrapper } from './StyledFileUploadWrapper';
 import {
@@ -29,7 +28,6 @@ import { useMergeRefs } from '~utils/useMergeRefs';
 import { useControllableState } from '~utils/useControllable';
 import { getInnerMotionRef, getOuterMotionRef } from '~utils/getMotionRefs';
 import { makeAnalyticsAttribute } from '~utils/makeAnalyticsAttribute';
-import { fireNativeEvent } from '~utils/fireNativeEvent';
 
 const _FileUpload: React.ForwardRefRenderFunction<BladeElementRef, FileUploadProps> = (
   {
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/FileUpload/FileUpload.web.tsx
```diff
@@ -377,11 +372,8 @@ const _FileUpload: React.ForwardRefRenderFunction<BladeElementRef, FileUploadPro
               size={size}
               onRemove={() => {
                 const newFiles = selectedFiles.filter(({ id }) => id !== file.id);
-                flushSync(() => {
```
anuraghazra, line 380: We are also removing flushSync?

anuraghazra (approved):

packages/blade/src/components/FileUpload/FileUpload.web.tsx
saurabhdaware: Yup we added it to fix the onChange args right? now we are not calling onChange on remove

saurabhdaware:
I verified the changes. Seems to work fine. Lets merge

anuraghazra (approved):

tewarig (approved):

PR #2524, opened 2025-02-04
feat: OTPInput field value clear behaviour changes

kabilansakthivelu:
## Description

In OTPInput, when a value is cleared in one of the input field, persisting the focus in the same input field instead of shifting the focus to previous input field

### Current behaviour:
When user wants to update a particular input field value in OTPInput component, the focus gets shifted to previous input field on clearing the value and user have to shift to the corresponding input field again to enter the updated value

https://github.com/user-attachments/assets/c10a9de6-d959-4d10-8f7b-3d2d19da4965

### Changes made:
When `Backspace` or `Delete` key is pressed on an input field in OTPInput component,
- if value existed, then it will be cleared and focus will be persisted on the same input field allowing user to update
- if value is empty, then focus will be shifted to previous input field and its value will be cleared

https://github.com/user-attachments/assets/53caae98-9b12-4878-8643-a2a7fb2c5099

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/.changeset/eighty-doors-yell.md
+++ b/.changeset/eighty-doors-yell.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+feat: Improve OTPInput field value clear behaviour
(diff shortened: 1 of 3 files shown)


saurabhdaware:
Hi, thanks for the PR. This looks great. I'll just get it reviewed from our design team and update here.

saurabhdaware (approved):

anuraghazra (dismissed):

anuraghazra (approved):

PR #2532, opened 2025-02-07
feat(TextArea): expose onKeydown on TextArea

saurabhdaware:
## Description

- add `numberOfLines={1}`
- expose `onKeyDown` on TextArea. 

onKeyDown will be exposed with these arg type-
```
onKeyDown={(args) => 
  args.name
  args.value
  args.event
}
```

Currently other handlers of TextArea expose name and value so it feels inconsistent to have one handler that exposes just event. So using the same type that we internally also use where we append event on top of existing name and value

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Input/TextArea/TextArea.tsx
+++ b/packages/blade/src/components/Input/TextArea/TextArea.tsx
@@ -21,6 +21,7 @@ import type {
   DataAnalyticsAttribute,
 } from '~utils/types';
 import { hintMarginTop } from '~components/Form/formTokens';
+import type { FormInputOnKeyDownEvent } from '~components/Form/FormTypes';
 
 type TextAreaCommonProps = Pick<
   BaseInputProps,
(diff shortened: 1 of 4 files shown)


anuraghazra (dismissed):

packages/blade/src/components/Input/TextArea/TextArea.tsx
```diff
@@ -244,6 +246,7 @@ const _TextArea: React.ForwardRefRenderFunction<BladeElementRef, TextAreaProps>
       }}
       onKeyDown={(e) => {
         handleTaggedInputKeydown(e);
+        onKeyDown?.(e);
```
anuraghazra, line 249: What's e here? {name, value, event}???

saurabhdaware: > {name, value, event}???

I'm fixing it. I just realised its `{name, key, event}`. We can make it `{ name, value, event }`

snitin315 (approved):

anuraghazra (approved):

snitin315 (approved):

PR #2536, opened 2025-02-09
fix(blade):  chipgroup grid alignment

tewarig:
## Description

ref - https://razorpay.slack.com/archives/CMQ3RBHEU/p1738752395317139 

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


anuraghazra:
We should ideally do this, and let consumer handle the layout.

```jsx
      <ChipGroupComponent
        selectionType="single"
        label="Select a gift card with value (custom chipGroupContainerLayout)"
      >
        <Box display="grid" gridTemplateColumns="1fr 1fr 1fr">
          {chipArray.map((chip, index) => (
            <ChipComponent key={index} value={chip.value}>
              {chip.label}
            </ChipComponent>
          ))}
        </Box>
      </ChipGroupComponent>
```

Only challenge is internally we are applying some default margins to each Chip component, we need to somehow make it work with the above example + keep the existing behaviour intact. 

Maybe we could do something like: 

```
            <BaseBox
              display="flex"
              flexDirection="row"
              flexWrap="wrap"
              rowGap={chipGroupGapTokens[size].bottom}
              columnGap={chipGroupGapTokens[size].right}
              marginBottom={chipGroupGapTokens[size].bottom}
            >
              {React.Children.map(children, (child, index) => {
                return <BaseBox key={index}>{child}</BaseBox>; // maybe we can remove the box if needed
              })}
            </BaseBox>
```

tewarig:
> We should ideally do this, and let consumer handle the layout.
> 
> ```js
>       <ChipGroupComponent
>         selectionType="single"
>         label="Select a gift card with value (custom chipGroupContainerLayout)"
>       >
>         <Box display="grid" gridTemplateColumns="1fr 1fr 1fr">
>           {chipArray.map((chip, index) => (
>             <ChipComponent key={index} value={chip.value}>
>               {chip.label}
>             </ChipComponent>
>           ))}
>         </Box>
>       </ChipGroupComponent>
> ```
> 
> Only challenge is internally we are applying some default margins to each Chip component, we need to somehow make it work with the above example + keep the existing behaviour intact.
> 
> Maybe we could do something like:
> 
> ```js
>             <BaseBox
>               display="flex"
>               flexDirection="row"
>               flexWrap="wrap"
>               rowGap={chipGroupGapTokens[size].bottom}
>               columnGap={chipGroupGapTokens[size].right}
>               marginBottom={chipGroupGapTokens[size].bottom}
>             >
>               {React.Children.map(children, (child, index) => {
>                 return <BaseBox key={index}>{child}</BaseBox>; // maybe we can remove the box if needed
>               })}
>             </BaseBox>
> ```

thanks for suggesting this... i have implemented this... the api looks like 
```
  <ChipGroupComponent
        selectionType="single"
        label="Select a gift card with value (custom chipGroupContainerLayout)"
      >
        <Box
          display="grid"
          gridTemplateColumns="repeat(3, minmax(0,120px))"
          gridTemplateRows="repeat(3, minmax(0,30px))"
          gap="spacing.3"
        >
          {chipArray.map((chip, index) => (
            <ChipComponent key={index} value={chip.value} width="120px">
              {chip.label}
            </ChipComponent>
          ))}
        </Box>
      </ChipGroupComponent>
```

also regarding that margin ... i think we don't need to do anything .. we can get children's from wrapper and render them. 
```
   {React.Children.map(getChipWrapperChildren(), (child, index) => {
                return (
                  <BaseBox
                    key={index}
                    marginBottom={chipGroupGapTokens[size].bottom}
                    marginRight={chipGroupGapTokens[size].right}
                  >
                    {child}
                  </BaseBox>
                );
              })}
```

packages/blade/src/components/Chip/ChipGroup.tsx
```diff
@@ -68,6 +70,24 @@ const _ChipGroup = (
       });
     }
   }
+  const isWrapperLayoutControlled =
+    typeof children === 'object' && getComponentId(children) === 'Box';
```
anuraghazra, line 74:
```
            <BaseBox
              display="flex"
              flexDirection="row"
              flexWrap="wrap"
              rowGap={chipGroupGapTokens[size].bottom}
              columnGap={chipGroupGapTokens[size].right}
              marginBottom={chipGroupGapTokens[size].bottom}
            >
              {React.Children.map(children, (child, index) => {
                return <BaseBox key={index}>{child}</BaseBox>; // maybe we can remove the box if needed
              })}
            </BaseBox>
```

This code i shared, should just work without doing any of these checks/conditionals.

tewarig: Yes, but I think we should add a wrapper and check for styles since consumers can pass props like background color or margin and padding, which should not be supported.

anuraghazra: We should not do arbitrary checks if it's absolutely not necessary. If they add background etc anyways it will look odd. 

Your above code will fail in many cases like so: 

```ts
const SomeCustomWrapperBox = styled(Box)``

<ChipGroup>
  <SomeCustomWrapperBox>
    ...
  </SomeCustomWrapperBox>
</ChipGroup>
``

File: packages/blade/src/components/Chip/utils.ts [hunk 2]
```diff
@@ -0,0 +1,57 @@
+// ref - /blade/packages/blade/src/components/Box/styledProps/getStyledProps.ts
```
anuraghazra, line 1: Remove this

packages/blade/src/components/Chip/ChipGroup.tsx
```diff
@@ -97,17 +97,7 @@ const _ChipGroup = (
...
-                  <BaseBox
-                    key={index}
-                    marginBottom={chipGroupGapTokens[size].bottom}
-                    marginRight={chipGroupGapTokens[size].right}
-                  >
-                    {child}
-                  </BaseBox>
-                );
-              })}
+              {React.Children.map(children, (child) => child)}
```
anuraghazra, line 100: Again, we need to keep the row gap etc props to make sure we are not changing the existing spacing that was there PLUS we let consumers handle the layout if they want. 

```
            <BaseBox
              display="flex"
              flexDirection="row"
              flexWrap="wrap"
              // Keep existing spacing that was there
              rowGap={chipGroupGapTokens[size].bottom}
              columnGap={chipGroupGapTokens[size].right}
              marginBottom={chipGroupGapTokens[size].bottom}
            >
              {React.Children.map(children, (child, index) => {
                return <BaseBox key={index}>{child}</BaseBox>;
              })}
            </BaseBox>
```

tewarig: https://github.com/razorpay/blade/blob/98296fb7f8da7dfbb9a9db44e5b7e0fa957ddf51/packages/blade/src/components/Chip/Chip.tsx 
Moved that to Chip; consumers can modify it from there.
Adding spacing on the wrapper won't work since this spacing needs to be applied to Chip.

anuraghazra: @tewarig reusable components should not have margins, they are bit [harmful](https://mxstbr.com/thoughts/margin). If in future we want to expose Chip as a standalone component (eg: Clickable only chip) then it will break component encapsulation. 

> Adding spacing on the wrapper won't work since this spacing needs to be applied to Chip.
Thus the BaseBox i've added on my example.

packages/blade/src/components/Chip/ChipGroup.tsx
```diff
@@ -96,18 +96,15 @@ const _ChipGroup = (
...
-              })}
+            <BaseBox
+              display="flex"
+              flexDirection="row"
+              flexWrap="wrap"
+              rowGap={chipGroupGapTokens[size].bottom}
+              columnGap={chipGroupGapTokens[size].right}
+              marginBottom={chipGroupGapTokens[size].bottom}
+            >
+              {React.Children.map(children, (child) => child)}
```
anuraghazra, line 107: Probably can just be `{children}`

packages/blade/src/components/Chip/ChipGroup.stories.tsx
```diff
@@ -727,3 +727,80 @@ chipRef.parameters = {
...
+        label="Select a gift card with value (custom chipGroupContainerLayout)"
+      >
+        <Box
+          display="grid"
+          gridTemplateColumns="repeat(3, minmax(0,120px))"
+          gridTemplateRows="repeat(3, minmax(0,30px))"
+          gap="spacing.3"
+        >
+          {chipArray.map((chip, index) => (
+            <ChipComponent key={index} value={chip.value} width="120px">
```
anuraghazra, line 798: 🤔 We shouldn't need to add explicit pixel values in ChipComponent. 

`<ChipComponent key={index} value={chip.value} width="100%" />`

I would assume this would work and the chip should automatically stretch/flex to available space. 


Not to mention explicit pixel values will break the layout if there is no space: 

image

packages/blade/src/components/Box/styledProps/getStyledProps.ts
```diff
@@ -144,5 +144,5 @@ const getStyledProps = (props: Record<string, any>): StyledPropsBlade => {
   return removeUndefinedStyledProps(makeStyledProps(props));
 };
 
-export type { StyledPropsBlade };
+export type { StyledPropsBlade, StyledPropsInputType };
```
saurabhdaware, line 147: where is this used?

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
```diff
@@ -332,5 +332,5 @@ type BoxRefType = Platform.Select<{
   native: View;
 }>;
 
-export type { BaseBoxProps, BoxProps, BoxRefType, StyledPropsBlade, FlexboxProps };
+export type { BaseBoxProps, BoxProps, BoxRefType, StyledPropsBlade, FlexboxProps, GridProps };
```
saurabhdaware, line 335: where is this used?

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -39,7 +39,7 @@ type OnChange = ({
 }) => void;
 
 const _Chip: React.ForwardRefRenderFunction<BladeElementRef, ChipProps> = (
-  { isDisabled, value, children, icon: Icon, color, testID, _motionMeta, ...rest },
+  { isDisabled, value, children, icon: Icon, color, testID, _motionMeta, width, ...rest },
```
saurabhdaware, line 42: we should expose `minWidth`, `maxWidth` as well right if we're exposing `width`?

packages/blade/src/components/Box/styledProps/getStyledProps.ts
tewarig: removed

packages/blade/src/components/Box/BaseBox/types/propsTypes.ts
tewarig: removed

packages/blade/src/components/Chip/AnimatedChip.web.tsx
```diff
@@ -14,7 +14,9 @@ const AnimatedChip = styled(BaseBox)<AnimatedChipProps>((props) => {
   );
   return {
     ...getAnimatedChipStyles(props),
-    width: 'fit-content',
+    width: props?.width ? (props.width as string | number) : 'fit-content',
+    maxWidth: props?.maxWidth as string | number,
+    minWidth: props?.minWidth as string | number,
```
saurabhdaware, line 19: why is `as string  | number` required here? isn't styled-components by default accepting string and number?

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -158,9 +170,20 @@ const _Chip: React.ForwardRefRenderFunction<BladeElementRef, ChipProps> = (
         onKeyDown={handleKeyboardPressedIn}
         onKeyUp={handleKeyboardPressedOut}
         inputProps={isReactNative() ? inputProps : {}}
-        style={{ cursor: _isDisabled ? 'not-allowed' : 'pointer' }}
+        style={{
+          cursor: _isDisabled ? 'not-allowed' : 'pointer',
+          width: width as string | number,
+          maxWidth: maxWidth as string | number,
+          minWidth: minWidth as string | number,
```
saurabhdaware, line 177: why is `as string | number` required here?

saurabhdaware, line 177: This will break if I do `width="spacing.3"` no?

packages/blade/src/components/Chip/ChipGroup.tsx
```diff
@@ -96,18 +96,15 @@ const _ChipGroup = (
...
-              })}
+            <BaseBox
+              display="flex"
+              flexDirection="row"
+              flexWrap="wrap"
+              rowGap={chipGroupGapTokens[size].bottom}
+              columnGap={chipGroupGapTokens[size].right}
+              marginBottom={chipGroupGapTokens[size].bottom}
+            >
+              {children}
```
saurabhdaware, line 107: nice 🏅

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -158,9 +170,20 @@ const _Chip: React.ForwardRefRenderFunction<BladeElementRef, ChipProps> = (
...
+        }}
       >
-        <BaseBox display="flex" flexDirection="column">
+        <BaseBox
+          display="flex"
+          flexDirection="column"
+          width={width}
+          maxWidth={maxWidth}
+          minWidth={minWidth}
```
saurabhdaware, line 185: Why are we passing width to multiple childs in this component? Shouldn't only parent component have width and child components take parent's width 🤔

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -158,9 +177,12 @@ const _Chip: React.ForwardRefRenderFunction<BladeElementRef, ChipProps> = (
         onKeyDown={handleKeyboardPressedIn}
         onKeyUp={handleKeyboardPressedOut}
         inputProps={isReactNative() ? inputProps : {}}
-        style={{ cursor: _isDisabled ? 'not-allowed' : 'pointer' }}
+        style={{
+          cursor: _isDisabled ? 'not-allowed' : 'pointer',
+          ...selectorLabelStyles,
```
saurabhdaware, line 182: Sorry for one more comment 🙈 But shouldn't we be adding width styles to outermost element? In this the basebox that we have? and let Chip take its width?

That way you also won't have to do `getBaseBoxStyles` hack since its BaseBox itself where you can directly pass width props.

Currently it feels too aggresive to add styles in style tag since they take very high priority. Plus I'm guessing it will break when I do `width={{ base: 'spacing.3', m: 'spacing.4' }}` since style tag won't support media query

packages/blade/src/components/Chip/ChipGroup.stories.tsx
```diff
@@ -727,3 +727,82 @@ chipRef.parameters = {
...
+      value: '200000',
+      label: '₹200000',
+    },
+  ];
+
+  return (
+    <Box gap="spacing.3" display="flex" flexDirection="column">
+      <ChipGroupComponent
+        selectionType="single"
+        label="Select a gift card with value (without custom chipGroupContainerLayout)"
```
saurabhdaware, line 779:
```suggestion
        label="Select a gift card with value (without custom chipGroupContainerLayout)"
```

what is chipGroupContainerLayout?

saurabhdaware (approved):
Looks good to me. Minor doc comment

packages/blade/src/components/Chip/types.ts
```diff
@@ -226,6 +239,9 @@ type AnimatedChipProps = {
   isDesktop?: boolean;
   theme: Theme;
   children: React.ReactNode;
+  width?: BaseBoxProps['width'];
+  maxWidth?: BaseBoxProps['maxWidth'];
+  minWidth?: BaseBoxProps['minWidth'];
```
anuraghazra, line 244: Let's use BoxProps instead of BaseBoxProps for the props which are publicly exposed.

saurabhdaware (approved):

PR #2540, opened 2025-02-13
fix: prevent duplicate tag addition

kabilansakthivelu:
## Description

- In `TaggedInput`, added a check to avoid duplicate tag addition
- Since in uncontrolled tagged input component, where the `onChange` will be called from `useControllableState` hook, wrapped the `onTagChange` for controlled component inside `else` branch to avoid getting called twice for uncontrolled component

#### Current behaviour:
- In `TaggedInput`, the `key` prop is constructed for each `tag` using [key={`${tagName}-${tagIndex}`}](https://github.com/razorpay/blade/blob/fb661be3e4870107910eb56d080981f35e7d25f3/packages/blade/src/components/Tag/getTagsGroup.tsx#L14)
- In case if duplicate tags are added, then the key for each tag gets reset on removal which is causing the other duplicated tag (which gets previously deleted tag index) non visible as it gets `isTagVisible` prop value as false and few other duplicated tags as non-removable

https://github.com/user-attachments/assets/96fcf325-4c13-47b5-8126-312ff47b6855

#### Changes made:
- Added a check to avoid duplicate tag entry
- Wrapped the `onTagChange` invocation part inside `else` branch to avoid getting called twice incase of `uncontrolled` tagged input as it will be called once through `useControllableState` hook

https://github.com/user-attachments/assets/1e818e7f-f05d-4a90-a459-7136369862cf

Kindly suggest whether we need to prevent the duplicate tag addition or do we need to remove the already existing tag of same input value on hitting [Enter]

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Input/BaseInput/useTaggedInput.ts
```diff
@@ -78,11 +77,11 @@ const useTaggedInput = ({
         activeTagIndex,
         isDisabled,
         onDismiss: ({ tagIndex }) => {
-          console.log('dismiss', { tagIndex });
           if (!isTagsControlled) {
             setTagsValue(() => getNewTagsArray(tagIndex));
+          } else {
+            onTagChange?.({ tags: getNewTagsArray(tagIndex) });
           }
-          onTagChange?.({ tags: getNewTagsArray(tagIndex) });
```
saurabhdaware, line 85: Wouldn't this mean that onTagChange won't be called when tags are uncontrolled? we still want to call onTagChange in those scenarios right?

packages/blade/src/components/Input/BaseInput/useTaggedInput.ts
```diff
@@ -146,20 +145,23 @@ const useTaggedInput = ({
     const inputValue = isControlledValue ? value?.trim() : inputValueUncontrolled.trim();
     if (e.key === 'Enter' || e.key === ',') {
       e.event.preventDefault?.(); // we don't want textarea to treat enter as line break in tagged inputs
-      if (inputValue) {
+      const isDuplicateTag = currentTags?.includes(inputValue);
```
saurabhdaware, line 148: Consumers should handle this no? from UX, if our end-user tries to add duplicate tag it should not add that tag and show validation error with message something like "tag already exist" which should be handled on consumer-side 🤔 . It shouldn't magically not add tag if user is trying to add

kabilansakthivelu: understood. thanks for the clarification. removing the duplicate tag check added

saurabhdaware (approved):

packages/blade/src/components/Input/BaseInput/useTaggedInput.ts
kabilansakthivelu: Thank you Saurabh

PR #2544, opened 2025-02-13
fix: button group loading 

tewarig:
## Description

this fixes - https://razorpay.atlassian.net/browse/DSSUP-245 

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/ButtonGroup/StyledButtonGroup.tsx
+++ b/packages/blade/src/components/ButtonGroup/StyledButtonGroup.tsx
@@ -2,33 +2,37 @@ import styled from 'styled-components';
 import type { StyledButtonGroupProps } from './types';
 import BaseBox from '~components/Box/BaseBox';
 import { makeBorderSize } from '~utils';
-import { getBackgroundColorToken } from '~components/Button/BaseButton/BaseButton';
-import getIn from '~utils/lodashButBetter/get';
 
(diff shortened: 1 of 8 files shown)


packages/blade/src/components/ButtonGroup/StyledButtonGroup.tsx
```diff
@@ -2,33 +2,38 @@ import styled from 'styled-components';
...
+        },
+        'button[role="button"]:not(:first-child):not(:last-child)': {
+          borderLeft: 'none',
+          borderRight: 'none',
+        },
+        'button[role="button"]:last-child': {
+          borderLeft: 'none',
+          borderTopRightRadius: makeBorderSize(theme.border.radius.medium),
+          borderBottomRightRadius: makeBorderSize(theme.border.radius.medium),
+        },
```
saurabhdaware, line 35: How is this related to loading? 🤔

tewarig: Earlier, we added a border to the Button Group container and removed the borders from all individual buttons.
However, when we enabled the loading state for a button, it looked like this:

Screenshot 2025-02-14 at 3 58 11 PM
I think the ideal way to handle this was to modify the border-radius of the buttons. So, I refactored the Button Group code a bit and also updated how the border is handled for the secondary variant.

saurabhdaware (approved):

PR #2567, opened 2025-03-09
fix(blade): focus on selected date by default in datepicker

tewarig:
## Description

[ref](https://razorpay.slack.com/archives/CMQ3RBHEU/p1741088046734839)
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/DatePicker/Calendar.web.tsx
+++ b/packages/blade/src/components/DatePicker/Calendar.web.tsx
@@ -70,10 +70,11 @@ const Calendar = <Type extends DateSelectionType>({
     if (_date) {
       return _date;
     }
-    if (Array.isArray(oldValue) && oldValue[0]) {
+    const isRangeSelection = Array.isArray(oldValue);
+    if (isRangeSelection && oldValue[0]) {
       return oldValue[0];
     }
-    if (!Array.isArray(oldValue) && oldValue) {
+    if (!isRangeSelection && oldValue) {
       return oldValue;
     }
     return shiftTimezone('add', new Date());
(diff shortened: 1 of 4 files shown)


packages/blade/src/components/DatePicker/Calendar.web.tsx
```diff
@@ -64,7 +66,18 @@ const Calendar = <Type extends DateSelectionType>({
 
   const dateContext = useDatesContext();
   const isMobile = useIsMobile();
-  const currentDate = _date ?? shiftTimezone('add', new Date());
+  const currentDate = React.useMemo(() => {
```
anuraghazra, line 69: Can you explain whats happening here? the `date` prop as is should work?

tewarig: We are already using dateProp. Currently, _date is passed to Date.
Whenever the user clicks on the next month/year, we update _date, and the date picker focuses on _date.
I modified this logic slightly:
If _date exists, focus on it.
Otherwise, focus on oldValue.
If oldValue is also unavailable, focus on currentDate.

anuraghazra: what is oldValue?

tewarig: It's the last applied value in the date picker.
Whenever you click/select any value on the date picker, its value changes. But if you click on Cancel, we revert back to the older value using oldValue. When you click on Apply, we save the applied value in oldValue to retrieve it next time.

anuraghazra (approved):
Mostly LGTM, Add interaction tests

anuraghazra (approved):

PR #2586, opened 2025-03-27
feat(StepGroup): support collapsible inside of StepGroup

saurabhdaware:
## Description

Add Collapsible StepGroup for QuickView

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


tewarig:
collapsible inside of StepGroup does not looks good in case of  horizontal orientation
https://github.com/user-attachments/assets/6a5050e7-e450-4809-9000-80e4c319c08e

saurabhdaware:
> The collapsible element inside the StepGroup does not look good when the orientation is horizontal.

True. I think we can throw error if collapsible is used inside orientation horizontal 🤔. What do you think?

tewarig:
> > The collapsible element inside the StepGroup does not look good when the orientation is horizontal.
> 
> True. I think we can throw error if collapsible is used inside orientation horizontal 🤔. What do you think?

That can work. We can mention in the documentation that horizontal orientation is not supported yet and ping design onces regarding this.

tewarig (approved):

PR #2610, opened 2025-04-13
fix: list view improvements

tewarig:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/ListView/ListViewFilters.web.tsx
+++ b/packages/blade/src/components/ListView/ListViewFilters.web.tsx
@@ -88,6 +88,7 @@ const ListViewFilters = ({
   const searchNameValue = searchName || searchId;
   const isMobile = useIsMobile();
   const { theme } = useTheme();
+  const showSearchInput = onSearchChange || onSearchClear || searchValuePlaceholder || searchName;
 
   return (
     <ListViewFiltersProvider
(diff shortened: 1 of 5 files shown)


packages/blade/src/components/ListView/ListViewFilters.web.tsx
```diff
@@ -88,6 +88,7 @@ const ListViewFilters = ({
   const searchNameValue = searchName || searchId;
   const isMobile = useIsMobile();
   const { theme } = useTheme();
+  const showSearchInput = onSearchChange || onSearchClear || searchValuePlaceholder || searchName;
```
saurabhdaware, line 91: Oh I missed this earlier but why do we need `onSearchClear` prop? wouldn't onSearchChange with empty value be considered as search clear?

saurabhdaware (commented):
This PR looks good. I added one comment on `onSearchClear` prop. Sorry missed seeing it earlier 🙈

packages/blade/src/components/ListView/ListViewFilters.web.tsx
tewarig: yes that can be considered as well. will remove this .

tewarig: yes we can use that well . will remove onSearchClear.

tewarig: onSearchClear does not get called when the value is cleared.
You can check this example: https://stackblitz.com/edit/akftnedq?file=App.tsx
We cannot remove onSearchClear.

saurabhdaware (approved):

saurabhdaware (approved):

PR #2613, opened 2025-04-17
fix(blade): broken padding logic in action list item 

tewarig:
## Description

fix(blade): broken padding logic in action list item 

RCA: 

I refactored the leading logic from:

```tsx
leading={
   <BaseBox
            pointerEvents="none"
            // Adding aria-hidden because the listbox item in multiselect in itself explains the behaviour so announcing checkbox is unneccesary and just a nice UI tweak for us
            {...makeAccessible({
              hidden: true,
            })}
          >
            <Checkbox isChecked={isSelected} tabIndex={-1} isDisabled={props.isDisabled}>
              {/* 
      Checkbox requires children. Didn't want to make it optional because its helpful for consumers
      But for this case in particular, we just want to use Text separately so that we can control spacing and color and keep it consistent with non-multiselect dropdowns
    */}
              {null}
            </Checkbox>
          </BaseBox>
}
```
to:

```tsx
leading={
  <BaseMenuLeadingItem
    key={`${dropdownBaseId}-${props._index}-leading-${isSelected}`}
    isSelected={isSelected}
    leading={props.leading}
    selectionType={selectionType}
    isDisabled={props.isDisabled}
  />
}
```
Even though BaseMenuLeadingItem returns null, 
it is still a valid React component. 
for padding We have logic like ```paddingLeft={leading ? spacing.3 : spacing.0} ```, 
and since leading is now always a component (not null or undefined),
 the padding was being applied regardless—causing the issue.

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/BottomSheet/__tests__/__snapshots__/BottomSheet.web.test.tsx.snap
+++ b/packages/blade/src/components/BottomSheet/__tests__/__snapshots__/BottomSheet.web.test.tsx.snap
@@ -142,7 +142,7 @@ exports[`<BottomSheet /> should compose with DropdownButton 1`] = `
   -ms-flex-direction: column;
   flex-direction: column;
   padding-right: 8px;
-  padding-left: 8px;
+  padding-left: 0px;
 }
 
 .c22.c22.c22.c22.c22 {
--- a/.changeset/ten-buttons-accept.md
+++ b/.changeset/ten-buttons-accept.md
@@ -1,0 +1,5 @@
+---
+'@razorpay/blade': patch
+---
+
+fix(blade): actionlistitem padding logic broken
--- a/packages/blade/src/components/ActionList/ActionListItem.tsx
+++ b/packages/blade/src/components/ActionList/ActionListItem.tsx
@@ -284,35 +284,28 @@ const ActionListItemText = assignWithoutSideEffects(_ActionListItemText, {
 
 const BaseMenuLeadingItem = ({
   isSelected,
-  leading,
-  selectionType,
   isDisabled,
 }: {
   isSelected?: boolean;
-  leading?: React.ReactNode;
-  selectionType: string | undefined;
   isDisabled?: boolean;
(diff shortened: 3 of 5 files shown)


packages/blade/src/components/BottomSheet/__tests__/__snapshots__/BottomSheet.web.test.tsx.snap
```diff
@@ -142,7 +142,7 @@ exports[`<BottomSheet /> should compose with DropdownButton 1`] = `
   -ms-flex-direction: column;
   flex-direction: column;
   padding-right: 8px;
-  padding-left: 8px;
+  padding-left: 0px;
```
saurabhdaware, line 145: Is this expected? Wouldn't this visually look broken on BottomSheet, Dropdown, etc?

saurabhdaware (commented):
Can you just visually check if all table stories and basic bottomsheet, dropdown, autocomplete things are working?

packages/blade/src/components/BottomSheet/__tests__/__snapshots__/BottomSheet.web.test.tsx.snap
tewarig: "Yes... I think during the list view, there were many changes, so I missed this part.

tewarig:
> Can you just visually check if all table stories and basic bottomsheet, dropdown, autocomplete things are working?

checked... things seems to be fine.

packages/blade/src/components/BottomSheet/__tests__/__snapshots__/BottomSheet.web.test.tsx.snap
saurabhdaware: ohhh got it. So it was 0px before then became 8px during list view and that we're fixing now. Understood.

saurabhdaware (approved):

PR #2620, opened 2025-04-23
feat: full size modal

tewarig:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Modal/docs/FullScreenModal.stories.tsx
saurabhdaware, line 1: why separate story file?

packages/blade/src/components/Modal/docs/SimpleModal.stories.tsx
saurabhdaware, line 1: I think here you can just do 

```jsx
const FullScreenModal = ModalTemplate.bind({})
FullScreenModal.args = {
  size: 'full'
}
```

This should do the job right?

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -31,41 +32,52 @@ import { componentZIndices } from '~utils/componentZIndices';
...
-    transform: translate(-50%, -50%) scale(0.9) translateY(20px);
-  }
-  to {
-    opacity: 1;
-    transform: translate(-50%, -50%) scale(1) translateY(0px);
-  }
-`;
+const ModalContent = styled(BaseBox)<{ isVisible: boolean; size: NonNullable<ModalProps['size']> }>(
+  ({ isVisible, theme, size }) => {
+    const centerTransform = size === 'full' ? 'none' : 'translate(-50%, -50%)';
```
saurabhdaware, line 37: Is this animation defined from design?

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -152,18 +164,25 @@ const Modal = ({
...
+                maxWidth={size === 'full' ? '100%' : makeSize(modalMaxWidth[size])}
                 minWidth={makeSize(modalMinWidth)}
-                maxHeight={modalMaxHeight}
-                width={`calc(100vw - ${makeSize(modalResponsiveScreenGap)})`}
+                maxHeight={modalMaxHeight[size]}
+                width={
+                  size === 'full'
+                    ? `calc(100vw - ${makeSize(modalMargin[size])} - ${makeSize(
+                        modalMargin[size],
+                      )})`
```
saurabhdaware, line 174:
```suggestion
                    ? `calc(100vw - ${makeSize(modalMargin[size] * 2)},
                      )})`
```

Same thing?

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -152,18 +164,25 @@ const Modal = ({
...
+                    : `calc(100vw - ${makeSize(modalResponsiveScreenGap)})`
+                }
                 borderRadius={modalBorderRadius}
                 backgroundColor="popup.background.subtle"
                 display="flex"
                 flexDirection="column"
-                top="50%"
-                left="50%"
+                top={size === 'full' ? makeSize(modalMargin[size]) : '50%'}
+                left={size === 'full' ? makeSize(modalMargin[size]) : '50%'}
```
saurabhdaware, line 182: Why do we have to define different value for full? wouldn't it still be center aligned just full width?

saurabhdaware:
image

borderRadius on bottom left is not working 🤔

packages/blade/src/components/Modal/Modal.web.tsx
tewarig: nope. i have kept is same for normal and full size modal. 

https://github.com/user-attachments/assets/d93cb4d9-7e00-45f5-81d2-c877c708b801

but we don't need translate(-50%, -50%) .. in case of full page modal. since it will already at center. 
also had a discussion with RK. this animation looks weird for full page modal will update it to fade in and fade out

packages/blade/src/components/Modal/ModalBody.web.tsx
```diff
@@ -18,11 +19,13 @@ type ModalBodyProps = {
    * - Docs: https://blade.razorpay.com/?path=/docs/tokens-spacing--docs
    */
   padding?: Extract<SpacingValueType, 'spacing.0' | 'spacing.6'>;
+  height?: BoxProps['height'];
 } & DataAnalyticsAttribute;
 
 const _ModalBody = ({
   children,
   padding = 'spacing.6',
+  height,
```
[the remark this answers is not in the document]
tewarig: The Modal body should have a height prop, as in the case of a full Modal, we need the Modal body to take up 100% of the available height inside the Modal.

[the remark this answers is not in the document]
saurabhdaware: Ohhh what happens if I do size="full" and not define height here?

Can we not define height="100%" for all ModalBody?

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -31,41 +33,75 @@ import { componentZIndices } from '~utils/componentZIndices';
...
+      from {
+        opacity: 0;
+        transform: translate(-50%, -50%) scale(0.9) translateY(20px);
+      }
+      to {
+        opacity: 1;
+        transform: translate(-50%, -50%) scale(1) translateY(0px);
+      }
+    `;
+
```
saurabhdaware, line 48: This change is not needed anymore right? we can move keyframes outside and revert to previous state

tewarig:
> Few comments. Rest looks good:
> 
> * [feat: full size modal #2620 (comment)](https://github.com/razorpay/blade/pull/2620#discussion_r2067891031)
> * [feat: full size modal #2620 (comment)](https://github.com/razorpay/blade/pull/2620#discussion_r2067892759)
> * There is a strange flash when modal opens. Can you check the animations again? this is `0.5x` video:
> 
>  animation-flash.mov

found the issue

Uploading Screen Recording 2025-04-30 at 5.44.09 PM.mov…

tewarig:
> chore: update css

seems like same issue exists in master in case of nomal modal 

https://github.com/user-attachments/assets/52ede700-64cf-487e-b3a1-04fb870c218e
Screenshot 2025-04-30 at 8 44 16 PM

tewarig:
@saurabhdaware let me know if u can pick this animation bug in another pr ?

packages/blade/src/components/Modal/Modal.web.tsx
```diff
@@ -1,8 +1,9 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import React, { useEffect } from 'react';
-import styled, { css, keyframes } from 'styled-components';
+import styled, { css } from 'styled-components';
 import { FloatingFocusManager, FloatingPortal, useFloating } from '@floating-ui/react';
 import usePresence from 'use-presence';
+import { m } from 'framer-motion';
```
saurabhdaware, line 6: nice

saurabhdaware: Our modal seems to lag a bit too much now with this though. Its a bit too visible 😭 . Lets go back to previous CSS animation and lets live with that flash of content instead?

https://github.com/user-attachments/assets/b9884a77-5dfa-4e41-bfc4-fb8e08cc21a0

tewarig: @saurabhdaware reverted changes

saurabhdaware (approved):

PR #2624, opened 2025-04-25
feat: remove universal-base64 library

saurabhdaware:
## Description

Removes the `universal-base64` library and adds custom base64 encoding function.

The library used `"browser"` key in package.json which required special handling in bundler. Old projects like admin dashboard don't have setup to handle that and anyways that library was fairly small.

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/StepGroup/StepLine.web.tsx
+++ b/packages/blade/src/components/StepGroup/StepLine.web.tsx
@@ -1,5 +1,4 @@
 /* eslint-disable @typescript-eslint/restrict-plus-operands */
-import { encode } from 'universal-base64';
 import type { StepItemProps } from './types';
 import { StepItemIndicator } from './StepItemMarker';
 import { useStepGroup } from './StepGroupContext';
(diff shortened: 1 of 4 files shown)


packages/blade/src/components/StepGroup/StepLine.web.tsx
```diff
@@ -16,6 +15,25 @@ import BaseBox from '~components/Box/BaseBox';
...
+    return window.btoa(svgString);
+  } catch (error: unknown) {
+    throwBladeError({
+      message: `Failed to encode SVG string: ${error}`,
+      moduleName: 'StepLine',
+    });
+
+    return '';
+  }
+};
```
tewarig, line 36: can we the original code and attribute the author ? 
 https://github.com/blakeembrey/universal-base64/blob/master/src/browser.ts
 
 this code looks fine ... but might have some edge cases

tewarig: we can skip this as well.. since it's used for only one

tewarig (approved):

packages/blade/src/components/StepGroup/StepLine.web.tsx
saurabhdaware: So I checked for the edge cases. `window.btoa` fails on emojis and characters outside of latin1.

Since we're using this on static string defined by us, its fine to not cover those edge cases. I tried and it works on our strings. 

Anyways if we're using this on some dynamic string later, we'll need a sanitise function to avoid XSS and plain encoding won't work

PR #2633, opened 2025-05-01
fix: Accessibility issues

Shubhdeep12:
## Description

Fixed some accessibility issues found while going through the  docs
- Avatar: to add children wrapper tags(`p` and `h5`) with avatar only when there is a children element.
- Card: `aria-pressed` should only be added to the button, not the anchor tag.
- Checkbox
  - To have `role="alert"` in errorText 
  - To set `aria-checked="mixed"` in case of inderminate state

## Component Checklist

- [ ] Add KitchenSink Story
- [x] Add changeset


anuraghazra (approved):

anuraghazra:
Thanks for the PR @Shubhdeep12 

I've reverted the role=alert for HintText since role=alert is a bit intrusive if multiple inputs are present and it will degrade the screen reader experience. Our errorText nodes are already associated with IDs and bound to the `<input>` so it should be okay.

Shubhdeep12:
Yeah makes sense. 
Thanks!

PR #2635, opened 2025-05-02
feat(blade-coverage-extension): add Accessibility tab in chrome extension

saurabhdaware:
## Description

- Add esbuild setup for background_script
- Add accessibility tab in chrome extension
- Add accessibility coverage script

What is not covered-

- Did not make the accessibility coverage script reusable yet. Its a bit tricky because we'll probably have to release some new package that can be used in both app-metrics server and this extension. I don't think so moving them to one repo makes sense because app-metrics will have to be in frontend-core-tools and this coverage extension will be open-source 

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


anuraghazra (approved):
Mostly looks good, hope we've tested this manually locally.

anuraghazra (approved):

PR #2636, opened 2025-05-02
fix(StepGroup): extra space because of trailing property

abhay-rzp:
## Description

The `trailing` prop in the StepItem Component rendered extra space because of Flex Properties which caused the title and description to not be visually center aligned
## Changes
Box Component for Trailing renders only when the trailing property is present

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


swapnil-kr1 (approved):

packages/blade/src/components/StepGroup/StepItem.web.tsx
```diff
@@ -164,7 +164,7 @@ const _StepItem = ({
           {description}
         </Text>
       </Box>
-      <Box>{trailing}</Box>
+      {trailing ? <Box>{trailing}</Box> : null}
```
swapnil-kr1, line 167: Current {trailing ? <Box>{trailing}</Box> : null} works fine.

Small nit: since trailing is a React element and we don’t need to render anything when it’s falsy, you could simplify to:
{trailing && <Box>{trailing}</Box>}

Feel free to keep it as is or switch to the && form if you prefer the terser syntax — either works.

saurabhdaware: should be fine I guess. Will merge this PR for now

saurabhdaware:
Thanks @abhay-rzp for the contribution 👯

packages/blade/src/components/StepGroup/StepItem.web.tsx
abhay-rzp: @swapnil-kr1 I followed this convention because eslint complained for having `&&` operator for conditional rendering in the `Merchant Dashboard`, was not aware if it can be used here without the linter errors 😅.

PR #2639, opened 2025-05-05
ci: mcp server release infra

kamleshchandnani:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


scripts/publishToNpm.js
saurabhdaware, line 1: so we can remove our earlier publishToNpm files?

packages/blade/README.md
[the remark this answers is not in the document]
kamleshchandnani: we were previously copying this readme to blade directory when publishing to npm to show the readme on npm page. now since we are doing the publishing generally i just copied it here. ideally it should be inside `pacakeges/blade` only but since we have a monorepo and we want to show this in GitHub's readme as well we put it on root directory. too much round and round 🙈

scripts/publishToNpm.js
kamleshchandnani: yes. oops i forgot to commit. let me do that

PR #2664, opened 2025-05-09
feat: preview window

tewarig:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/package.json
+++ b/packages/blade/package.json
@@ -147,7 +147,8 @@
     "@mantine/dates": "6.0.21",
     "@mantine/hooks": "6.0.21",
     "dayjs": "1.11.10",
-    "react-window": "1.8.11"
+    "react-window": "1.8.11",
+    "react-zoom-pan-pinch": "3.7.0"
   },
   "devDependencies": {
     "http-server": "14.1.1",
--- a/packages/blade/src/components/PreviewWindow/docs/PreviewWindow.stories.tsx
+++ b/packages/blade/src/components/PreviewWindow/docs/PreviewWindow.stories.tsx
@@ -1,0 +1,70 @@
+import type { StoryFn, Meta } from '@storybook/react';
+import React from 'react';
+import { PreviewBody, PreviewHeader, PreviewWindow, PreviewFooter } from '../PreviewWindow';
+import type { PreviewWindowProps } from '../types';
+import { Heading } from '~components/Typography/Heading';
+import { Sandbox } from '~utils/storybook/Sandbox';
+import StoryPageWrapper from '~utils/storybook/StoryPageWrapper';
+import { getStyledPropsArgTypes } from '~components/Box/BaseBox/storybookArgTypes';
+import { Box } from '~components/Box';
+
+const Page = (): React.ReactElement => {
+  return (
+    <StoryPageWrapper
+      componentName="PreviewWindow"
+      componentDescription="A Preview Window is a component that is used to preivew a file"
+      apiDecisionLink={null}
+      figmaURL="https://www.figma.com/design/jubmQL9Z8V7881ayUD95ps/Blade-DSL?node-id=100413-32686&t=n9A7LztwEkIsly3v-0"
+    >
+      <Heading size="large">Usage</Heading>
+      <Sandbox showConsole>
+        {`
+        import {  } from '@razorpay/blade/components';
+        
+        function App() {
+          return (
+            <ChatMessage > Hi, from ray! </ChatMessage>
+          )
+        }
+
+        export default App;
+      `}
+      </Sandbox>
+    </StoryPageWrapper>
+  );
+};
+
+export default {
+  title: 'Components/PreviewWindow',
+  component: PreviewWindow,
+  tags: ['autodocs'],
+  argTypes: {
+    ...getStyledPropsArgTypes(),
+  },
+  parameters: {
+    docs: {
+      page: Page,
+    },
+  },
+} as Meta<PreviewWindowProps>;
+
+const PreviewWindowTemplate: StoryFn<typeof PreviewWindow> = (args) => {
+  return (
+    <PreviewWindow>
+      <Box>
+        <PreviewHeader title="Preview " />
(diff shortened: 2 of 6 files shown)


packages/blade/package.json
```diff
@@ -287,7 +288,8 @@
     "@razorpay/i18nify-react": "4.0.12",
     "plop": "3.1.1",
     "node-plop": "0.32.0",
-    "svgson": "5.3.1"
+    "svgson": "5.3.1",
+    "react-pdf": "^9.2.1"
```
saurabhdaware, line 292:
```suggestion
    "react-pdf": "9.2.1"
```

packages/blade/src/utils/componentZIndices.ts
```diff
@@ -9,4 +9,5 @@ export const componentZIndices = {
   popover: 1100,
   tooltip: 1100,
   topnav: 100,
+  previewPanel: 1000,
```
saurabhdaware, line 12: Should this be a bit higher? 1100? because you would have preview panel in modal and if you full screen from modal, it might in some cases end up opening behind the modal since both have 1000 zIndex.

packages/blade/src/components/PreviewWindow/docs/PreviewWindow.stories.tsx
```diff
@@ -0,0 +1,333 @@
...
+  parameters: {
+    docs: {
+      page: Page,
+    },
+  },
+} as Meta<PreviewWindowProps>;
+
+const PreviewWindowTemplate: StoryFn<typeof PreviewWindow> = () => {
+  return (
+    <PreviewWindow>
```
saurabhdaware, line 74: Should the component just be called `<Preview>`? `PreviewWindow` sounds like its some nested component unless we name inner components `PreviewWindowHeader` etc.

Preview sounds fine I guess?

packages/blade/src/components/PreviewWindow/PreviewWindow.web.tsx
```diff
@@ -0,0 +1,339 @@
...
+    onDragChange?.({ x: positionX, y: positionY });
+  };
+
+  const handleFullScreen = useCallback(() => {
+    if (!document.fullscreenElement) {
+      containerRef.current?.requestFullscreen().catch((err) => {
+        console.error(`Error attempting to enable fullscreen: ${err.message}`);
+      });
+      setIsFullScreen(true);
+      onFullScreenProp?.();
```
saurabhdaware: no need to wait for promise here?

packages/blade/src/components/PreviewWindow/PreviewWindow.web.tsx
```diff
@@ -0,0 +1,339 @@
...
+    } else {
+      document.exitFullscreen().catch((err) => {
+        console.error(`Error attempting to exit fullscreen: ${err.message}`);
+      });
+      setIsFullScreen(false);
+      onFullScreenProp?.();
```
saurabhdaware, line 249: same here

packages/blade/src/components/PreviewWindow/PreviewWindow.web.tsx
```diff
@@ -0,0 +1,339 @@
...
+    }
+  }, [onFullScreenProp]);
+
+  // Handle keyboard shortcut for fullscreen
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent): void => {
+      if ((event.metaKey || event.ctrlKey) && event.key === 'f') {
```
saurabhdaware, line 256: nice!

packages/blade/src/components/PreviewWindow/PreviewWindow.web.tsx
```diff
@@ -0,0 +1,339 @@
...
+  }, []);
+
+  // filter out preview header, preview body, preview footer separately using componentId
+  const previewHeader = React.Children.toArray(children)
+    .filter((child) => getComponentId(child as React.ReactElement) === MetaConstants.PreviewHeader)
+    .map((child) => {
+      if (React.isValidElement(child)) {
+        return React.cloneElement(child, {
+          // @ts-expect-error
+          _onFullScreen: handleFullScreen,
```
saurabhdaware, line 287: Either way is fine but if you want you can also use the same PreviewWindowContext for keeping onFullScreen as well right?

packages/blade/src/utils/componentZIndices.ts
swapnil-kr1: +1

packages/blade/src/components/PreviewWindow/PreviewWindow.web.tsx
tewarig: will go with this way only..

saurabhdaware (dismissed):

packages/blade/src/components/Preview/Preview.web.tsx
```diff
@@ -0,0 +1,369 @@
...
+      } catch (err: unknown) {
+        handleFullScreenError(err);
+      }
+    }
+  }, [onFullScreenProp]);
+
+  // Handle keyboard shortcut for fullscreen
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent): void => {
+      if ((event.metaKey || event.ctrlKey) && event.key === 'f') {
```
anuraghazra, line 284: Why are we hijacking cmd+f (search in page) shortcut?

tewarig: i think we should just keep it "F" , since youtube also have that as a shortcut

saurabhdaware: Right now the shortcut seems to be on the `document`. Doesn't that mean it will end up full-screening the preview while I am typing something in the input 🙈?

I don't think so we should have shortcut key for full screening. In case of youtube or vimeo it makes sense because the media is the focus for those sites 🤔

saurabhdaware (approved):

PR #2665, opened 2025-05-09
feat(blade-mcp): tweaks for non-dev personas

saurabhdaware:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade-mcp/src/server.ts
```diff
@@ -130,7 +130,7 @@ try {
 
             1. Call create_blade_cursor_rules to create the cursor rules.
             2. Run \`npm install --legacy-peer-deps && npm install @razorpay/blade@latest --legacy-peer-deps\` to install the latest blade.
-            3. Always start the development server using \`npm run dev\` before starting to code
+            3. Always start the development server using \`npm run dev\` (if its not already running) before starting to code
```
kamleshchandnani, line 133: @saurabhdaware can we add the node version check and throw the error when creating the base project as discussed?

PR #2680, opened 2025-05-16
feat(Card): improve card knowledgebase

saurabhdaware:
## Description

This makes the Card component aware of what all is allowed and not allowed

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade-mcp/knowledgebase/Card.md
+++ b/packages/blade-mcp/knowledgebase/Card.md
@@ -6,7 +6,7 @@ Card
 
 Cards are containers that group related content and actions on a single topic. They help separate content into distinct sections, making interfaces easier to scan and understand. Cards support various layouts with customizable headers, bodies, and footers, enabling consistent presentation of information while providing clear interaction points for users.
 
-**Important Note:**
+## Important Constraints
 
 - `Card` component only accepts `CardHeader`, `CardBody`, `CardFooter` components as children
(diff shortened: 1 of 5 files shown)


packages/blade-mcp/knowledgebase/Card.md
```diff
@@ -6,6 +6,14 @@ Card
 
 Cards are containers that group related content and actions on a single topic. They help separate content into distinct sections, making interfaces easier to scan and understand. Cards support various layouts with customizable headers, bodies, and footers, enabling consistent presentation of information while providing clear interaction points for users.
 
+**Important Note:**
```
kamleshchandnani, line 9: Maybe change the heading to Rules or Constraints? so it becomes generic and we'll use this pattern for all other component too?

saurabhdaware: Makes sense. I'll keep the "important" since that is the thing I feel that does the trick 😆

saurabhdaware: done

anuraghazra: Many components may have "Important Notes" let's call this "Card Rules & Constraints" or something. So it's easier to prompt LLM to follow these sections like: 

```
- Remember to follow "Rules & Constraints" section for each component. 
```

saurabhdaware: yes yes changed it already

PR #2683, opened 2025-05-16
feat: add mcp error monitoring

kamleshchandnani:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/.changeset/perfect-files-help.md
+++ b/.changeset/perfect-files-help.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade-mcp": patch
+---
+
+feat: add mcp error monitoring
--- a/.changeset/perfect-files-help.md
+++ b/.changeset/perfect-files-help.md
@@ -1,5 +1,0 @@
----
-"@razorpay/blade-mcp": patch
----
-
-feat: add mcp error monitoring
--- a/packages/blade-mcp/src/replaceEnv.js
+++ b/packages/blade-mcp/src/replaceEnv.js
@@ -10,6 +10,8 @@ const __dirname = path.dirname(__filename);
 
 // Define the path to the server.js file
 const SERVER_JS_PATH = path.resolve(__dirname, '../dist/server.js');
+// Define the path to the utils.js file
+const UTILS_JS_PATH = path.resolve(__dirname, '../dist/utils.js');
 
 // Load env variables from .env file (if exists) or use process.env
 dotenv.config();
@@ -23,7 +25,7 @@ async function replaceEnvironmentVariables() {
     // Replace NODE_ENV
     await replaceInFile({
       files: SERVER_JS_PATH,
-      from: /process\.env\.NODE_ENV\s*\|\|\s*['"]development['"]/g,
+      from: /process\.env\.NODE_ENV\s*\?\?\s*['"]development['"]/g,
       to: `'${NODE_ENV}'`,
     });
 
@@ -32,6 +34,14 @@ async function replaceEnvironmentVariables() {
       files: SERVER_JS_PATH,
       from: /process\.env\.BLADE_MCP_SENTRY_DSN/g,
       to: `'${BLADE_MCP_SENTRY_DSN}'`,
+    });
+
+    // Replace BLADE_SEGMENT_KEY
+    const BLADE_SEGMENT_KEY = process.env.BLADE_SEGMENT_KEY || '';
+    await replaceInFile({
+      files: UTILS_JS_PATH,
+      from: /process\.env\.BLADE_SEGMENT_KEY/g,
+      to: `'${BLADE_SEGMENT_KEY}'`,
     });
   } catch (error) {
     console.error('Error during environment variables replacement:', error);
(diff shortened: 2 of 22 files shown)


packages/blade-mcp/src/replaceEnv.js
```diff
@@ -0,0 +1,52 @@
...
+dotenv.config();
+
+// Use environment variables with defaults
+const NODE_ENV = process.env.NODE_ENV;
+const BLADE_MCP_SENTRY_DSN = process.env.BLADE_MCP_SENTRY_DSN || '';
+
+async function replaceEnvironmentVariables() {
+  try {
+    // Replace NODE_ENV
+    await replaceInFile({
```
saurabhdaware, line 26: interesting

packages/blade-mcp/src/replaceEnv.js
```diff
@@ -0,0 +1,52 @@
...
+
+    // Replace SENTRY_DSN
+    await replaceInFile({
+      files: SERVER_JS_PATH,
+      from: /process\.env\.BLADE_MCP_SENTRY_DSN/g,
+      to: `'${BLADE_MCP_SENTRY_DSN}'`,
+    });
+
+    // Replace BLADE_SEGMENT_KEY
+    const BLADE_SEGMENT_KEY = process.env.BLADE_SEGMENT_KEY || '';
```
saurabhdaware, line 40: In production mode while publishing package on NPM or GitHub, where do we define these keys? Also are we ok with these keys going public? cause they'll be in the bundle of non-razorpay people as well right?

packages/blade-mcp/src/utils.ts
```diff
@@ -48,12 +53,142 @@ const getBladeComponentsList = (): string[] => {
...
   }
 
   return bladeComponentsList;
 };
 
+const handleError = ({
+  toolName,
+  error,
+  customErrorMessage = '',
+}: {
```
saurabhdaware, line 68:
```suggestion
const handleMCPError = ({
  toolName,
  shouldCaptureError,
  errorMessage = '',
}: {
```

Will this make it more readable?

```jsx
// return error but don't capture it
handleMCPError({ toolName: 'get_blade_component_docs', shouldCaptureError: false, errorMessage: 'No cursor rules' }) 

// return error and also capture it
handleMCPError({ toolName: 'get_blade_component_docs', shouldCaptureError: true, errorMessage: error }) 
```

I was a bit confused on which errors are going to sentry and which ones are part of our workflow

packages/blade-mcp/src/utils.ts
```diff
@@ -48,12 +53,142 @@ const getBladeComponentsList = (): string[] => {
...
+          : customErrorMessage,
+      },
+    ],
+  };
+};
+
+/**
+ * Get MAC addresses from network interfaces
+ */
+const getMacAddresses = (): string[] => {
```
saurabhdaware, line 95: nice

packages/blade-mcp/src/tools/createBladeCursorRules.ts
```diff
@@ -2,59 +2,83 @@ import { join } from 'path';
...
-  writeFileSync(ruleFilePath, ruleFileTemplateContent);
+    writeFileSync(ruleFilePath, ruleFileTemplateContent);
 
-  return {
-    content: [
-      {
-        type: 'text',
-        text: `Blade cursor rules created at: ${ruleFilePath}. Cursor Rules Version: ${CURSOR_RULES_VERSION}`,
+    sendAnalytics({
+      eventName: 'Blade MCP Tool Called',
```
saurabhdaware, line 56: we can move this string to variable as well. Someone might mess up the spacing or spellings

saurabhdaware (approved):

packages/blade-mcp/src/replaceEnv.js
```diff
@@ -0,0 +1,42 @@
...
+
+// Define the path to the server.js file
+const SERVER_JS_PATH = path.resolve(__dirname, '../dist/server.js');
+
+// Load env variables from .env file (if exists) or use process.env
+dotenv.config();
+
+// Use environment variables with defaults
+const NODE_ENV = process.env.NODE_ENV;
+const BLADE_MCP_SENTRY_DSN = process.env.BLADE_MCP_SENTRY_DSN || '';
```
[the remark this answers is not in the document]
kamleshchandnani: the thing is replacements are in different files both from and to.
for node_env I do it in script command since we don't want node_env to be prod always even when we are in dev mode, second some env variables are in server.ts and some are in utils

packages/blade-mcp/src/replaceEnv.js
kamleshchandnani: I have added in GH action as secrets. Yeah should be fine because even on all our client apps these keys are public in js bundles. 🙈

saurabhdaware: oh right makes sense.

.changeset/perfect-files-help.md
saurabhdaware, line 1: 2 changesets in this PR. Lets remove this one?

saurabhdaware (dismissed):

saurabhdaware (approved):

PR #2687, opened 2025-05-20
feat(blade-mcp): add runtime check fixes on cursor

saurabhdaware:
## Description

- Add `vite-plugin-terminal`
- Add error boundary

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade-mcp/src/tools/createNewBladeProject.ts
```diff
@@ -57,6 +57,7 @@ const createNewBladeProjectCallback: ToolCallback<typeof createNewBladeProjectSc
             2. Run \`npm install --legacy-peer-deps && npm install @razorpay/blade@latest --legacy-peer-deps\` to install the latest blade.
             3. Always start the development server using \`npm run dev\` (if its not already running) before starting to code. Do not start it again after writing code.
             4. Start coding in App.tsx. use get_blade_docs to get information about the components.
+            5. If you see any errors on terminal, fix them.
```
kamleshchandnani, line 60: does this help?

packages/blade-mcp/base-blade-template/src/ErrorFallback.tsx
```diff
@@ -0,0 +1,57 @@
...
+              text: isCopied ? 'Copied' : 'Copy Error',
+              onClick: () => {
+                const prompt = `Fix the following runtime error: ${error.message}\n\nstack: ${error.stack}`;
+                navigator.clipboard.writeText(prompt);
+                setIsCopied(true);
+              },
+            },
+          }}
+        />
+        <Text>You can ask AI to "Fix runtime errors" or copy paste this error in chat</Text>
```
kamleshchandnani, line 51: do we need this since our copy error already has it? maybe copy CTA should be enough to make it easy? we can just mention copy the error in the chat?

packages/blade-mcp/base-blade-template/package.json
```diff
@@ -10,7 +10,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@razorpay/blade": "^12.25.0",
+    "@razorpay/blade": "^12.28.0",
```
kamleshchandnani, line 13: can we use "latest" here? so it picks latest always? and probably we can remove that from our cli flag as well

packages/blade-mcp/src/tools/createNewBladeProject.ts
saurabhdaware: nope 🤣

packages/blade-mcp/base-blade-template/src/ErrorFallback.tsx
saurabhdaware: It doesn't lead to much maintainence anyways so more of "good to have". Also simpler to fix errors this way as you don't have to pass any context of error message and all.

packages/blade-mcp/base-blade-template/package.json
saurabhdaware: "latest" might end up fetching the latest version on consumer side as well after the app is built so if we ship breaking change, it might break all apps that have `latest` in their package.json 🤔

packages/blade-mcp/src/tools/createNewBladeProject.ts
saurabhdaware: removed it

packages/blade-mcp/base-blade-template/package.json
kamleshchandnani: that issue will also happen since we have added the `--latest` flag to new project creation cli as well? also this will only happen for people creating new projects so no breaking changes as such? since its new

saurabhdaware: No no so when you run `npm install @razorpay/blade@latest` you install the latest version at that particular point in time when the command is running but your package.json would still have numbered version only `^12.31.0` for example.


But if you put "latest" in package.json itself, it means you're telling to install latest blade anytime the installation runs. So it can end up updating the version unexpectedly. 

package-lock.json does avoid unexpected updates but still the intent that you convey from package.json is to use latest blade version everytime you run `npm install`

saurabhdaware: Separately running `npm install @razorpay/blade@latest` also ensures the latest version while scaffolding blade project. So it also goes and updates the latest version in package-lock.json

Vs if you just keep it in package.json, pacakge-lock will have a defined version which will keep on getting installed everytime unless you explicitly update package-lock in our templates

kamleshchandnani: Hmm that would help mostly in cases of preventing major version bumps right? like it won't allow going from `^12.31.0` to `13.0.0`?

kamleshchandnani (approved):

PR #2697, opened 2025-05-21
feat: optimize blade for POS [experimental]

saurabhdaware:
## Description

- Related thread: https://razorpay.slack.com/archives/C06N18S6JNR/p1747663095269229?thread_ts=1747385088.256059&cid=C06N18S6JNR

## What was happening in this case?

- CardRoot component had a state update on touch of it (which is normally used for scale on hover on interactive cards)
- So if you have multiple nested cards among each other, the state update is triggered on card root for each card.

## Changes

- I have added a check that only adds these touch handlers if the card is interactive card and has shouldScaleOnHover prop

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/tokens/theme/__tests__/__snapshots__/createTheme.web.test.tsx.snap
```diff
@@ -1212,10 +1212,25 @@ exports[`createTheme should create a theme with the correct brand colors 3`] = `
...
+  transition-timing-function: cubic-bezier(0.3,0,0.2,1);
+}
+
+button:active .c1.c1.c1.c1.c1 {
+  -webkit-transform: scale(0.95);
+  -ms-transform: scale(0.95);
+  transform: scale(0.95);
+}
+
+a:active .c1.c1.c1.c1.c1 {
```
Vish7-77, line 1230: duplication usage of css detected, 
button:active .c1.c1.c1.c1.c1,a:active .c1.c1.c1.c1.c1 , use same css, dont need two blocks to define

PR #2717, opened 2025-05-26
feat(knowledgebase): support different types of documentations

saurabhdaware:
### New Behaviour

- AI can now figure out which pattern to use based on natural language.

  image

  image

- It can answer general questions about white labelling, usage, etc and also add white labelling
  
  image

## Changes

- Refactored utils to be directory instead of files
- We now have 3 tools and 3 different types of knowledgebases - `components`, `patterns`, `general` with 3 respective tool calls for each.
- When cursor feels that it can use a pattern, it makes the pattern docs tool call, and later makes the components tool call based on the components used in that pattern

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade-mcp/knowledgebase/general/index.md
+++ b/packages/blade-mcp/knowledgebase/general/index.md
@@ -8,7 +8,7 @@ The Usage guide covers the core fundamentals of implementing Blade in your appli
 
 ## AvailableIcons
 
-A comprehensive reference of all 356+ icons available in the Blade design system. This includes usage examples showing how to import icons, use them in components with icon props, and implement them as standalone components.
+A reference of which all icons are available in the blade design system. This includes usage examples showing how to import icons, use them in components with icon props, and implement them as standalone components.
 
 ## WhiteLabelling
 


packages/blade-mcp/knowledgebase/components/prompt.txt
```diff
@@ -7,4 +7,4 @@ Scan this Storybook file, generate a markdown file with below mentioned sections
...
...nent. Do not ever use comments `{/* some comment *}` or headings to merge variations in one example. Split examples whenever necessary in their own sections with heading and short description of what each example does. All the examples must show the import statements too required for creating that example. Use Box component only wherever its required and necessary. Focus on comprehensive usage rather than isolated prop demonstrations. Create examples that focuses more on this components use cases and use other components only as supporting components whenever required. This will be fed to LLMs in future when the user ask to build something that involves this component. This example will just act as the blueprint and then referring the types and prompt of user the LLMs will be able to create the required example prompted by a user. Do not pick components that start with `_something` those are private components and never exposed to consumer and meant for internal purposes. All the imports that start with `~/something` should be replaced with named exports of `@razorpay/blade/something` eg: `~components/*` shall be replaced with named absolute imports of `@razorpay/blade/components`.
```
anuraghazra, line 8: How and where are these prompt.txt files are used?

packages/blade-mcp/knowledgebase/general/AvailableIcons.md
```diff
@@ -0,0 +1,416 @@
...
+- WifiIcon
+- WifiOffIcon
+- WindIcon
+- WorldwideIcon
+- XCircleIcon
+- XSquareIcon
+- YoutubeIcon
+- ZapIcon
+- ZoomInIcon
+- ZoomOutIcon
```
tewarig, line 416: Not related to this, but we should modify  `generateIcons.mjs` to update this every time an icon is added.

saurabhdaware: Oh right that's also a good idea. I was thinking of adding some build script before blade-mcp's deployment that reads the latest icons and adds it here. Will explore this later

PR #2736, opened 2025-06-03
feat(InfoGroup): add InfoGroup component and docs

saurabhdaware:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/InfoGroup/InfoGroup.web.tsx
```diff
@@ -0,0 +1,466 @@
...
+      : `repeat(min(4, ${React.Children.count(children)}), 1fr)`;
+
+  // Use provided gridTemplateColumns or default based on itemOrientation
+  const templateColumns = gridTemplateColumns ?? defaultGridTemplateColumns;
+
+  return (
+    <InfoGroupContext.Provider value={contextValue}>
+      <BaseBox
+        ref={ref as never}
+        as="dl"
```
anuraghazra, line 411: Do we have other related `as` props for `dt` and `dd`?? If not we should either remove this and use normal spans or add related tags in other components.

saurabhdaware: yes yes I've added the dt and dd as well. Following the same structure as MDN example: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dd

packages/blade/src/components/InfoGroup/types.ts
```diff
@@ -0,0 +1,123 @@
...
+   * Truncates text after specified number of lines
+   * @default undefined
+   */
+  truncateAfterLines?: number;
+};
+
+export type TitleCollectionProps = {
+  children: string | React.ReactNode;
+  titleWeight: 'medium' | 'semibold';
+  titleColor: 'surface.text.gray.muted' | 'surface.text.gray.subtle';
```
anuraghazra, line 104: kinda error prone, maybe we can do `Extract<TextColors, 'surface.text.gray.muted' | 'surface.text.gray.subtle'>`

It will still not give us error but atleast we will know if the tokens are wrong. 

image


We should refactor all our Extract calls to `StrongExtract` 

```ts
type StrongExtract<T, E extends T> = T extends E ? T : never;

type K = StrongExtract<TextColors, 'gives error'>
```

packages/blade/src/components/Icons/useIconProps/iconSizeMap.ts
```diff
@@ -0,0 +1,12 @@
+import { size } from '~tokens/global';
```
anuraghazra, line 1: why put it on a different file?

packages/blade/src/components/InfoGroup/types.ts
saurabhdaware: We can use `Pick` no?. I'll add

swapnil-kr1 (approved):

PR #2740, opened 2025-06-04
feat(accordion): UI enhancement

swapnil-kr1:
## Description

UI refinements for Accordion

## Changes

* Added divider for expanded accordion
* Divider hides on hover
* Changed hover color
* Removed background color for expanded state
* Aligned all trailing, leading icons and numbers to top
* Leading Icon (Slot):
  - With sub-header: leading icon top-aligned, trailing icon top-aligned
  - Without sub-header: both leading and trailing icons center-aligned

## Additional Information
* Added divider for expanded accordion and * Removed background color for expanded state

Design:
Screenshot 2025-06-11 at 1 59 36 AM
Main:
Screenshot 2025-06-11 at 1 59 52 AM
Changed to: 
Screenshot 2025-06-11 at 2 00 35 AM

* Divider hides on hover And * Changed hover color
Design:
Screenshot 2025-06-11 at 2 03 00 AM

Main:
Screenshot 2025-06-11 at 2 03 08 AM

Changed to:
Screenshot 2025-06-11 at 2 03 13 AM

* Aligned all trailing, leading icons and numbers to top
Design:
Screenshot 2025-06-11 at 2 05 06 AM
Main already satisfies so no change required:
Screenshot 2025-06-11 at 2 08 23 AM

* Leading Icon (Slot):
  - With sub-header: leading icon top-aligned, trailing icon top-aligned
  - Without sub-header: both leading and trailing icons center-aligned
  Design:
Screenshot 2025-06-11 at 2 09 46 AM

  Main:
  
Screenshot 2025-06-11 at 2 11 43 AM

  Changed to:
  
Screenshot 2025-06-11 at 2 12 14 AM

  
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/BaseHeaderFooter/BaseHeader.tsx
```diff
@@ -299,12 +307,25 @@ const _BaseHeader = ({
...
   trailingInteractionElement,
   backgroundImage,
+  isAccordionContext = false,
   ...rest
 }: BaseHeaderProps): React.ReactElement => {
   const validatedTrailingComponent = useTrailingRestriction({ trailing, size });
   const shouldWrapTitle = titleSuffix && trailing && showBackButton && showCloseButton;
   const hasOnlyChildren = children && !(title || subtitle || titleSuffix || leading);
 
+  const isLeadingNumberOrIcon = React.useMemo(() => {
```
anuraghazra, line 317: Can you show an example on why do we need to calculate / know these specifics?

🤔

swapnil-kr1: Slack link: https://razorpay.slack.com/archives/C01H13RTF8V/p1749031551761339?thread_ts=1748604804.427729&cid=C01H13RTF8V

Case wise:
1. For number, icon : all trailing, leading are aligned to top: we need align flex-start
2. For slot: 
If sub-header is present → leading icon(slot) is top-aligned, trailing stays top-aligned : we need flex-start
If no sub-header → both leading icon(slot) and trailing icons are centre-aligned: we need center

So, to handle this dynamically, we need a function isLeadingNumberOrIcon to determine whether the user has passed a number, icon or slot as the leading prop — because alignment logic depends on the type of content passed.

packages/blade/src/components/Accordion/commonStyles.ts
```diff
@@ -22,7 +22,7 @@ const getBackgroundColor = ({
 
   if (isExpanded) {
     if (isActive) {
-      return gray.fadedHighlighted;
+      return gray.faded;
     }
```
tewarig, line 26: do we even need this if statement now ?

packages/blade/src/components/Dropdown/__tests__/__snapshots__/Dropdown.ssr.test.tsx.snap
```diff
@@ -488,6 +488,26 @@ exports[`<Dropdown /> should render dropdown and make it visible on click 2`] =
...
+}
+
+@media screen and (min-width:1024px) {
+
+}
+
+@media screen and (min-width:1200px) {
+
+}
+
```
tewarig, line 510: not sure but can you check why these empty breakpoint styles are being added ? 🤔

packages/blade/src/components/Accordion/commonStyles.ts
swapnil-kr1: Nice catch, I have optimised the function itself

packages/blade/src/components/Dropdown/__tests__/__snapshots__/Dropdown.ssr.test.tsx.snap
saurabhdaware: unrelated issue but strange. We have a check to not add breakpoint if css value is not present inside https://github.com/razorpay/blade/blob/16a8f55c53c7229d169cb4b6c67005d9beee7ec5/packages/blade/src/components/Box/BaseBox/baseBoxStyles.ts#L312

Also have tests on that - https://github.com/razorpay/blade/blob/16a8f55c53c7229d169cb4b6c67005d9beee7ec5/packages/blade/src/components/Box/__tests__/baseBoxStyles.web.test.ts#L38

saurabhdaware: > However, Object.fromEntries() still processes these empty arrays, resulting in empty media queries

Ohhh

saurabhdaware: But yeah anyways. Lets solve this separately later.

swapnil-kr1: Yes, Will look into this separately

swapnil-kr1: The issue isn't that Object.fromEntries() creates { undefined: undefined } - that would actually be better because it would get filtered out during CSS generation.
The real issue is more subtle:
When shouldAddBreakpoint() returns false, we return [] (empty array). But Object.fromEntries() expects key-value pairs. Here's what happens:
// Current problematic flow:
const entries = [
  ['@media screen and (min-width: 768px)', { display: 'block' }], // Valid
  [], // Empty array from shouldAddBreakpoint() === false
  ['@media screen and (min-width: 1024px)', { color: 'red' }], // Valid
];

Object.fromEntries(entries);
// This tries to destructure [] as [key, value] where both become undefined
The problem is that styled-components still processes these malformed entries and somehow generates empty CSS blocks. The exact mechanism depends on how styled-components handles undefined keys in the CSS object.

packages/blade/src/components/BaseHeaderFooter/BaseHeader.tsx
```diff
@@ -452,7 +475,12 @@ const _BaseHeader = ({
           ) : null}
         </BaseBox>
       </BaseBox>
-      {showDivider ? <Divider /> : null}
+      {showDivider ? (
+        <Divider
+          thickness={isAccordionContext ? 'thinner' : undefined}
+          marginX={isAccordionContext ? 'spacing.5' : 'spacing.0'}
```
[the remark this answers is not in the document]
swapnil-kr1: Done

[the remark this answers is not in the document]
saurabhdaware: Don't see things changed. Have you pushed the changes?

saurabhdaware (approved):

PR #2718, opened 2025-05-26
feat: add dropdown in search & textInput

tewarig:
## Component Checklist

- [ ] Update Component Status Page
- [ ] Perform Manual Testing in Other Browsers
- [ ] Add KitchenSink Story
- [ ] Add Interaction Tests (if applicable)
- [ ] Add changeset

PR #2761, opened 2025-06-11
fix: leading and trailing in text input

tewarig:
## Description

 leading / trailing was only accepting dropdown, in case of TextInput,  
 logic for icon was wrong in  https://github.com/razorpay/blade/pull/2718 . 
 plus leading/trailing should accept any React children. this pr fixes it.
 plus added some examples to storybook

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


swapnil-kr1 (approved):

packages/blade/src/components/Input/TextInput/TextInput.tsx
```diff
@@ -241,16 +241,19 @@ const _TextInput: React.ForwardRefRenderFunction<BladeElementRef, TextInputProps
 
   const trailingDropdown =
     trailing && getComponentId(trailing as React.ReactElement) === 'Dropdown' ? trailing : null;
-
+  // we need to look into name of component and check if it 's and icon or a dropdown
   const _leadingIcon: IconComponent | undefined =
```
anuraghazra, line 245: odd, how was it working before? I don't think we have componentId set in icons as "Icon"

packages/blade/src/components/Input/TextInput/TextInput.tsx
```diff
@@ -340,10 +342,21 @@ const _TextInput: React.ForwardRefRenderFunction<BladeElementRef, TextInputProps
       );
     }
 
+    if (showClearButton && hasTrailingInteractionElement) {
```
anuraghazra, line 345: test this combination out manually

packages/blade/src/components/Input/TextInput/TextInput.tsx
```diff
@@ -340,10 +342,21 @@ const _TextInput: React.ForwardRefRenderFunction<BladeElementRef, TextInputProps
...
+      return (
+        <BaseBox display="flex" gap="spacing.3">
```
anuraghazra, line 347: Let's actually add unit tests for these if not already present

tewarig: https://github.com/razorpay/blade/issues/2763

PR #2773, opened 2025-06-16
fix: label postion

tewarig:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Form/FormLabel.tsx
+++ b/packages/blade/src/components/Form/FormLabel.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import {
   labelLeftMarginRight,
   labelMarginBottom,
+  labelMarginBottomInChipGroup,
   labelOptionalIndicatorTextSize,
   labelTextSize,
   labelWidth,
@@ -28,6 +29,10 @@ type CommonProps = {
    * @default medium
    */
   size?: 'small' | 'medium' | 'large';
+  /**
+   * Name of the component that the label is inside
+   */
+  componentName?: 'chip-group' | 'checkbox-group' | 'radio-group' | 'file-upload';
 };
 
 type LabelProps = CommonProps & {
@@ -66,6 +71,7 @@ const FormLabel = ({
   id,
   htmlFor,
   size = 'medium',
+  componentName,
 }: FormLabelProps): React.ReactElement => {
   const { theme } = useTheme();
   const { matchedDeviceType } = useBreakpoint({ breakpoints: theme.breakpoints });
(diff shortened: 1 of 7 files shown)


swapnil-kr1 (dismissed):

packages/blade/src/components/Form/Selector/SelectorGroupField.tsx
```diff
@@ -42,6 +42,7 @@ const SelectorGroupField = ({
         role,
         labelledBy,
       })}
+      alignItems={componentName === 'chip-group' && isLabelLeftPositioned ? 'center' : undefined}
```
saurabhdaware, line 45: This will make the label center aligned when there are more chips right? I think that is not expected here 🤔 

## Current Prod

image

## Post this PR

image

packages/blade/src/components/Form/FormLabel.tsx
```diff
@@ -158,9 +175,7 @@ const FormLabel = ({
       id={id}
       {...metaAttribute({ name: MetaConstants.FormLabel })}
     >
-      <BaseBox marginBottom={isLabelLeftPositioned ? 'spacing.0' : labelMarginBottom[size]}>
-        {textNode}
-      </BaseBox>
+      <BaseBox marginBottom={textNodeContainerMarginBottom()}>{textNode}</BaseBox>
```
saurabhdaware, line 178: shouldn't the marginBottom be 0 always in case of labelLeftPositioned independent of ChipGroup? This will only push the text slightly upwards but won't necessarily align it right?

packages/blade/src/components/Form/FormLabel.tsx
```diff
@@ -66,6 +72,7 @@ const FormLabel = ({
   id,
   htmlFor,
   size = 'medium',
+  componentName,
```
saurabhdaware, line 75: Instead of componentName we can take `labelMarginBottom` type of prop so that we don't end up with too many component-specific checks here. Otherwise it becomes sort of a magic abstraction.

saurabhdaware: I am guessing we need `labelMarginTop` instead 🤔

tewarig:
Have fixed now it looks like - 

- PROD 
Screenshot 2025-07-03 at 11 54 32 PM

- DEV
Screenshot 2025-07-03 at 11 54 55 PM

- DEV
Screenshot 2025-07-03 at 11 58 10 PM

- PROD
Screenshot 2025-07-03 at 11 59 32 PM

saurabhdaware (approved):

PR #2774, opened 2025-06-16
fix: console error fix for BottomSheet and update margin for TableEditableDropdownCell

kabilansakthivelu:
#### BottomSheet:
- When `BottomSheet` component was mounted with `isOpen={false}`, the `scrollRef` was not updated with an actual reference to DOM element even after `BottomSheetBody` was rendered into DOM later

#### TableEditableDropdownCell:
- The field margin was not passed with the updated value for `rowDensity` - `comfortable` prop type

#### BottomSheet:
- Added `scrollRef` to `useEffect` dependencies array, to update the `lockTarget` with the updated value
- with this, we could avoid passing `null` to `enableBodyScroll` as we will be getting the updated `ref` element even when `BottomSheetBody` component was rendered later

#### TableEditableDropdownCell:
- Consumed `getEditableInputMargin` util which will return the margin for the `Dropdown` element based on `rowDensity` prop

#### TableEditableDropdownCell - before margin computation:

Screenshot 2025-06-16 at 12 15 37

#### TableEditableDropdownCell - after margin computation:

Screenshot 2025-06-16 at 12 18 26

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
+++ b/packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
@@ -443,13 +443,14 @@ const _BottomSheet = ({
   // Disable body scroll lock when the component is unmounted forcefully
   React.useEffect(() => {
     const lockTarget = scrollRef.current;
-    return function () {
+    return () => {
       if (lockTarget) {
         enableBodyScroll(lockTarget);
       }
     };
   // when BottomSheet is mounted with isOpen={false}, then BottomSheetBody does not set scrollRef
-  // so, we added scrollRef to dependencies to ensure that we update lockTarget when scrollRef is updated
+  // so, we added scrollRef to dependencies array to ensure that we update lockTarget when scrollRef is updated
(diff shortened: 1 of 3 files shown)


packages/blade/src/components/BottomSheet/BottomSheet.web.tsx
```diff
@@ -442,11 +442,16 @@ const _BottomSheet = ({
...
-    const lockTarget = scrollRef.current!;
+    const lockTarget = scrollRef.current;
     return () => {
-      enableBodyScroll(lockTarget);
+      if (lockTarget) {
+        enableBodyScroll(lockTarget);
+      }
     };
-  }, []);
+    // when BottomSheet is mounted with isOpen={false}, then BottomSheetBody does not set scrollRef
```
anuraghazra, line 451: Odd, can you share a small snippet on what was the previous behaviour / if any error that we were getting?

anuraghazra (approved):

PR #2840, opened 2025-07-14
feat(table): grouping

swapnil-kr1:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


tewarig:
this seems to be a bug

https://github.com/user-attachments/assets/536dd258-94fd-4d15-8bdc-4821760c5239

minor - you need to check for these white lines 
Screenshot 2025-07-17 at 1 20 11 AM

tewarig:
also can you add one disabled cell example ?

tewarig:
related to table in general ... should we add tests for these ? 
since this seems like an major change

saurabhdaware (dismissed):
Looks good

saurabhdaware:
https://github.com/user-attachments/assets/925c1cb9-fa04-455a-b5f2-7c86d0cde038

I think when you select group, and deselect an item from it, the group should go to intermediate state right?

swapnil-kr1:
Regarding the comment on the table grouping PR, I tried to solve it but noticed that the React Table library exhibits the same behavior. I couldn’t find a straightforward solution for this. I also explored some custom logic, but implementing it would make the logic quite complex.
We can discuss this further after the stand-up.
Here’s the reference link: https://react-table-library.com/?path=/docs/mixing-features-tree-select--expand-tree-on-row-click-select-on-checkbox

saurabhdaware (dismissed):

tewarig (dismissed):

saurabhdaware (approved):

PR #2841, opened 2025-07-16
fix(ListView): gradient visibility and scroll in list view filters

swapnil-kr1:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/ListView/ListViewFilters.web.tsx
+++ b/packages/blade/src/components/ListView/ListViewFilters.web.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useRef } from 'react';
 import { AnimatePresence, m } from 'framer-motion';
 import styled from 'styled-components';
 import type { ListViewFilterProps, ListViewSelectedFiltersType } from './types';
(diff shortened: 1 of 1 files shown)


tewarig:
can you add an example as well. with many filters.. maybe modify current one

anuraghazra:
Few things: 

1. When reaching the end the gradient overflow is still there, ideally we want that to be removed? 

image

1.1. Color of the gradient needs to match the background color so it blends properly, meaning we may need a new prop. 

2. When scrolled to right, in the left side there also should be a gradient? 
image

3. When selecting anything, the scroll resets 

https://github.com/user-attachments/assets/be4e14f5-f868-4dda-94e7-d14699d5dd55


4. Discuss with RK if we even need the gradient overlay, maybe just having the QuickFilters cut off is enough to indicate there is a scroll - anyways this is an edgecase and we recommend not to have these many items. We did this in many components where we realized that gradient overlay is unnecessarily complicating things so we removed them alltogether.

packages/blade/src/components/ListView/ListViewFilters.web.tsx
```diff
@@ -138,39 +150,19 @@ const ListViewFilters = ({
...
             width={getFilterContainerWidth()}
-            marginRight={isMobile ? 'spacing.2' : 'spacing.0'}
+            marginRight="spacing.3"
           >
             <StyledQuickFilterContainer
-              overflow={isMobile ? 'scroll' : 'visible'}
-              width={isMobile ? '100%' : 'auto'}
+              overflow="scroll"
+              width="100%"
               ref={(node) => {
```
anuraghazra, line 158:
```suggestion
              ref={containerRef}
```

packages/blade/src/components/ListView/ListViewFilters.web.tsx
```diff
@@ -93,20 +72,53 @@ const ListViewFilters = ({
...
+
+    const handleScroll = (): void => {
+      scrollPositionRef.current = container.scrollLeft;
+    };
+
+    container.addEventListener('scroll', handleScroll);
+    return (): void => container.removeEventListener('scroll', handleScroll);
+  }, []);
+
+  useIsomorphicLayoutEffect(() => {
```
anuraghazra, line 94: This seems like a hack to fix a fundamental issue that we have maybe in our check/radio boxes or in our quickfilter implementation. 

Why do we need to reposition scroll on simply a selection change of a quickfilter? Let's see whats the root cause is.

swapnil-kr1: Done

anuraghazra (approved):

PR #2871, opened 2025-08-05
feat: add Area , Bar , line and Donut Api decision

tewarig:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Charts/AreaChart/_decision.md
```diff
@@ -0,0 +1,369 @@
...
+### 4.1 Hybrid API (Prop & Config Driven)
+
+This approach combines component props for high-level control with configuration objects for detailed specifications.
+
+```ts
+// Import statement
+import { AreaChart } from '@razorpay/blade/charts';
+
+const chartData = [
+  { month: 'Jan', teamA: 2450, teamB: 2600 },
```
swapnil-kr1, line 188: Not a comment, just curious: what happens if the user passes { month: 'Jan', teamA: 'invalid', teamB: null }?

packages/blade/src/components/Charts/AreaChart/_decision.md
```diff
@@ -0,0 +1,369 @@
...
+
+/**
+ * Configuration for a single area on the chart.
+ */
+interface AreaDefinition {
+  dataKey: string;
+  name: string;
+  stroke: string;
+  fill: string;
+  stackId?: string | number;
```
swapnil-kr1, line 262: Lets include Accessibility props also

packages/blade/src/components/Charts/AreaChart/_decision.md
```diff
@@ -0,0 +1,369 @@
...
+interface YAxisConfigItem {
+  value: number;
+  label: string;
+}
+
+/**
+ * The main interface for all props accepted by the AreaChart component.
+ */
+interface CustomAreaChartProps {
+  data: ChartData[];
```
swapnil-kr1, line 286: Is there any limit on data size, or will it handle larger datasets gracefully — e.g., 1k, 10k, or 100k points?

packages/blade/src/components/Charts/AreaChart/_decision.md
```diff
@@ -0,0 +1,369 @@
...
+ * Configuration for a single area on the chart.
+ */
+interface AreaDefinition {
+  dataKey: string;
+  name: string;
+  stroke: string;
+  fill: string;
+  stackId?: string | number;
+}
+
```
swapnil-kr1, line 264: What about animations are we controlling them internally, or should we expose a prop to allow different animation types?
We might need to check with the design team if this hasn’t been discussed yet.

packages/blade/src/components/Charts/BarChart/_decision.md
```diff
@@ -0,0 +1,301 @@
...
+1. ### [TinyBarChart](https://recharts.org/en-US/examples/TinyBarChart)
+
+   ![](./images/tiny.png)
+   Simple bar graph with responsive container and BarChart.  
+2. [SimpleBarChart](https://recharts.org/en-US/examples/SimpleBarChart)
+
+   ![](./images/normal.png)
+   Bar Chart with other chart components. Need to discuss ActiveBar here..  With [Rama Krushna Behera](mailto:rama.behera@razorpay.com) . should users be able to change it or not. 
+
+3. [StackedBarChart](https://recharts.org/en-US/examples/StackedBarChart)
```
swapnil-kr1, line 40: Items 1, 2, and 3 have different styling and spacing, let’s make them consistent.

packages/blade/src/components/Charts/BarChart/_decision.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+#### **TypeScript Interface**
+
+```ts
+interface BarDefinition {
+  dataKey: string;
+  name: string;
+  color: BladeColorToken;
+  stackId?: string;
```
swapnil-kr1, line 232: Lets include Accessibility props also

packages/blade/src/components/Charts/BarChart/_decision.md
```diff
@@ -0,0 +1,301 @@
...
+  barConfig: BarDefinition[];
+  xAxisDataKey: string;
+  layout?: 'vertical' | 'horizontal';
+  stacked?: boolean;
+  yAxisLabel?: string;
+  xAxisLabel?: string;
+  showGrid?: boolean;
+  showTooltip?: boolean;
+  showLegend?: boolean;
+  // ... other high-level props
```
swapnil-kr1, line 246: Should we also expose an onHover prop?

packages/blade/src/components/Charts/AreaChart/_decision.md
```diff
@@ -0,0 +1,369 @@
...
+/**
+ * Configuration for a single area on the chart.
+ */
+interface AreaDefinition {
+  dataKey: string;
+  name: string;
+  stroke: string;
+  fill: string;
+  stackId?: string | number;
+}
```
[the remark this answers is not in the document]
tewarig: i think that should be handled by consumer.

packages/blade/src/components/Charts/AreaChart/_decision.md
tewarig: so we won't be exposing animations to consumers. they would be handled internally.

packages/blade/src/components/Charts/BarChart/_decision.md
tewarig: don't think we have need to expose it

swapnil-kr1 (dismissed):

tewarig:
> Should we merge all the Charts APIs into single Charts API decision?
> 
> Because reasoning to go with certain API is same. Only the props table is different so we can mention it in that single API decision itself

we can do that

packages/blade/src/components/Charts/_decisions/decisions.md
```diff
@@ -0,0 +1,1229 @@
...
+       <Line
+         dataKey="teamA"
+         type="solid"
+         color="theme.charts.grey"
+       />
+     </LineChart>
+   </ResponsiveContainer>
+
+ // ForeCast line chart 
+  <ResponsiveContainer width="100%" height="100%">
```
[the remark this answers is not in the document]
tewarig: Yes, a ResponsiveContainer is necessary.

The reason is that you cannot pass percentage-based strings like width="100%" or height="100%" directly to the chart components (e.g., LineChart, AreaChart). These components require a specific pixel value (a number) to render correctly. Therefore, this approach will not work for making charts responsive.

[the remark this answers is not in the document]
tewarig: or we can have a `isResponsive ` prop which will add this wrapper .. and if user want to modify height / width they can pass it in height / width prop of LineChart, AreaChart etc.

saurabhdaware (approved):
Are we calling our PieChart, DonutChart?

PR #2906, opened 2025-08-20
feat: phone number input validation on input and paste events

jaywalker21:
## Description

This PR adds validation to disallow non-numeric inputs in `PhoneNumberInput` component.
It also ensures that the same is handled when any text is pasted into the input.

## Changes
- Adds `onPaste` event support for `BaseInput` component for web via `handleOnPaste` handler
- Validates the character pressed using `onKeyDown` event and ensures that only `numeric` characters are allowed
- Validates the text pasted using `onPaste` event and ensures that text only contains `numbers`

## Additional Information

- Future scope - If number is pasted with dialling code, the event dispatched can be about updating `diallingCode` and `number`. This is not covered within the scope of this PR

## Component Checklist

- [x] Add KitchenSink Story
- [x] Add changeset

### DIFF
--- a/packages/blade/src/components/Input/PhoneNumberInput/PhoneNumberInput.web.tsx
+++ b/packages/blade/src/components/Input/PhoneNumberInput/PhoneNumberInput.web.tsx
@@ -18,6 +18,29 @@ import type { BladeElementRef } from '~utils/types';
 import { CloseIcon } from '~components/Icons';
 import { MetaConstants } from '~utils/metaAttribute';
 import { useControllableState } from '~utils/useControllable';
+import type { FormInputOnKeyDownEvent, FormInputOnPasteEvent } from '~components/Form/FormTypes';
+
+const NUMBERS_ONLY_REGEX = /^\d+$/;
+
+export const isNumericInput = (input: string): boolean => {
+  return NUMBERS_ONLY_REGEX.test(input);
+};
+
+export function validateNumericInput(event: FormInputOnKeyDownEvent['event']): void {
+  const { key, ctrlKey, metaKey } = event;
+
+  // Check if the entered key is a number; if not, prevent the default action
+  const isCharacterKey = key.length === 1;
+
+  // return for non-character keys
+  if (ctrlKey || metaKey || !isCharacterKey) {
+    return;
+  }
+
+  if (!isNumericInput(key)) {
+    event.preventDefault();
+  }
+}
 
 const _PhoneNumberInput: React.ForwardRefRenderFunction<BladeElementRef, PhoneNumberInputProps> = (
   {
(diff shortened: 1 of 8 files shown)


packages/blade/src/components/Input/PhoneNumberInput/PhoneNumberInput.web.tsx
```diff
@@ -121,6 +144,19 @@ const _PhoneNumberInput: React.ForwardRefRenderFunction<BladeElementRef, PhoneNu
...
   }, [allowedCountries, flags]);
 
+  const handleKeyDown = (keyboardEvent: FormInputOnKeyDownEvent): void => {
+    validateNumericInput(keyboardEvent.event);
+  };
+
+  const handlePaste = (pasteEvent: FormInputOnPasteEvent): void => {
+    const pastedText = pasteEvent.value?.clipboardData?.getData('text') ?? '';
+
+    // Only allow pastedText if it only has numeric characters
```
saurabhdaware, line 154: I am not sure if we should stop people from pasting numbers like `+91 9999 999999`. If you copy from whatsapp or a lot of other websites, it often comes with spaces, `-`, or `+`. We can let people paste things as is or we can replace the unsupported characters before pasting and paste the numbers, spaces, hyphens, and plus, etc.

saurabhdaware: You can check the `<input type="number" />` behaviour actually. We can replicate that.

saurabhdaware: Yeah skipping plus and space also fine. Overall its fine if wrongly formatted number is pasted. People can remove the 91 manually. We should just avoid blocking the paste

packages/blade/src/components/Input/PhoneNumberInput/PhoneNumberInput.web.tsx
```diff
@@ -18,6 +18,29 @@ import type { BladeElementRef } from '~utils/types';
...
+  const isCharacterKey = key.length === 1;
+
+  // return for non-character keys
+  if (ctrlKey || metaKey || !isCharacterKey) {
+    return;
+  }
+
+  if (!isNumericInput(key)) {
+    event.preventDefault();
+  }
```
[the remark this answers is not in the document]
jaywalker21: I missed one aspect in your comment. In case of non-numeric input, we need to prevent default handlers for event, so condition will simply be to return in case of `ctrlKey` and `metaKey`.
In case of non-numeric input, we will do `event.preventDefault()`

packages/blade/src/components/Input/PhoneNumberInput/PhoneNumberInput.web.tsx
jaywalker21: Checked what `<input type="number" />` does. It strips the string from all non-numeric inputs, adding the same logic here.

PR #2910, opened 2025-08-22
fix: wrong layout position in case of bottom sheet

tewarig:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
+++ b/packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
@@ -7,7 +7,7 @@ import type { HoverProps, SelectorInputProps } from './types';
 import getIn from '~utils/lodashButBetter/get';
 import type { Theme } from '~components/BladeProvider';
 import { castWebType } from '~utils';
-import { screenReaderStyles } from '~components/VisuallyHidden';
+import { screenReaderStyles, legacyScreenReaderStyles } from '~components/VisuallyHidden';
 import type { BladeElementRef } from '~utils/types';
 import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
 import { makeMotionTime } from '~utils/makeMotionTime';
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
```diff
@@ -37,17 +38,19 @@ const getHoverStyles = ({
...
-      ...getFocusRingStyles({ theme }),
-    },
-    '&:hover + div': {
-      ...getHoverStyles({ theme, isChecked, isDisabled, hasError, hoverTokens }),
-    },
-  }),
-);
+const StyledInput = styled.input<
+  HoverProps & {
+    isInBottomSheet?: boolean;
```
swapnil-kr1, line 43: Instead of naming it isInBottomSheet, maybe we could use a more generic name like showLegacyReaderStyles. That way, if the same issue comes up in other components, we can just reuse the same flag by passing true instead of introducing new variables each time.

packages/blade/src/components/Form/Selector/SelectorInput.web.tsx
```diff
@@ -76,6 +80,7 @@ const _SelectorInput: React.ForwardRefRenderFunction<BladeElementRef, SelectorIn
       hasError={hasError}
       tabIndex={tabIndex}
       hoverTokens={hoverTokens}
+      isInBottomSheet={isInBottomSheet}
```
swapnil-kr1, line 83: We could use useBottomSheetContext at line 73 — otherwise it might lead to prop drilling. At the same time, we can consider updating the prop name being passed so that it’s a bit more scalable.

tewarig: we should change it to 
```  useLegacyScreenReaderStyles={isInBottomSheet}  ```

PR #3036, opened 2025-12-02
CSS STyling migration | Migrated all css logic to blade-core

kashishbehl:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade-core/src/styles/BaseText/baseText.module.css
+++ b/packages/blade-core/src/styles/BaseText/baseText.module.css
@@ -1,3 +1,4 @@
+/* Base styles */
 .base {
   margin: 0;
   padding: 0;
@@ -5,10 +6,22 @@
   letter-spacing: var(--letter-spacing-100);
 }
 
-/* Text decoration color - inherits from color property (currentColor) */
+/* Text decoration variants with nested modifiers */
 .text-decoration-underline,
 .text-decoration-line-through {
   text-decoration-color: currentColor;
+
+  &.text-decoration-underline {
+    text-decoration-line: underline;
+  }
+
+  &.text-decoration-line-through {
+    text-decoration-line: line-through;
+  }
+}
+
+.text-decoration-none {
+  text-decoration-line: none;
 }
 
 /* Font style variants */
@@ -20,20 +33,7 @@
   font-style: italic;
 }
 
-/* Text decoration variants */
-.text-decoration-none {
-  text-decoration-line: none;
-}
-
-.text-decoration-underline {
-  text-decoration-line: underline;
-}
-
-.text-decoration-line-through {
-  text-decoration-line: line-through;
-}
-
-/* Text align variants */
+/* Text align variants - grouped using nesting */
 .text-align-left {
   text-align: left;
 }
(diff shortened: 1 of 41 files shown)


.cursor/rules/svelte-migration.mdc
```diff
@@ -19,6 +19,131 @@ Don't use createEventDispatcher for event handlers like on:click etc. Instead us
...
+
+```
+packages/
+└── blade-svelte/
+    └── src/
+        └── components/
+            ├── Button/
+            │   ├── Button.svelte
+            │   ├── types.ts
+            │   └── button.css
```
saurabhdaware: where is cva config placed in this directory structure?

saurabhdaware: oh wait the structure that is mentioned here seems different from the structure mentioned below in this file 🤔

packages/blade-core/src/styles/BaseText/baseText.module.css
```diff
@@ -0,0 +1,383 @@
...
+.color-current {
+  color: currentColor;
+}
+
+/* Color classes - generated dynamically from color tokens */
+/* Pattern: color-{token-path} where dots are replaced with hyphens */
+/* Example: surface.text.gray.normal -> color-surface-text-gray-normal */
+
+/* Surface Text Colors */
+.color-surface-text-gray-normal { color: var(--surface-text-gray-normal); }
```
saurabhdaware: hmm seems like we do have to define these. Should we define our tokens as global classes in this case? so that if people on consumer-side also want to use these, they can.

We already have semantic meaning to our colors so we will have lesser tokens compared to other utility class frameworks.

Don't see much reasons to keep this only in Text

E.g.
```css
// valid classes ✅ 
.color-surface-text-gray-subtle
.background-surface-background-gray-subtle

// invalid classes 🚫 
.background-surface-text-gray-subtle
```

packages/blade-core/src/styles/BaseText/baseText.ts
```diff
@@ -0,0 +1,193 @@
...
+    numberOfLines:
+      numberOfLines !== undefined && numberOfLines >= 1 && numberOfLines <= 10
+        ? (numberOfLines as 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10)
+        : undefined,
+  };
+
+  const classes = [
+    baseTextStyles(cvaVariants),
+    // Generate color class name dynamically
+    color ? colorTokenToClassName(color) : undefined,
```
saurabhdaware: is there no way to still keep color in CVA without having to define each prop? like is there any way to do

```js
{
  color: (color) => colorTokenToClassName(color)
}
```

packages/blade-svelte/src/components/Typography/Code/Code.svelte
```diff
@@ -37,37 +36,32 @@
...
     name: MetaConstants.Code,
     testID,
   });
 
-  // Combine classes
+  // Combine classes - everything is class-based, no inline styles
+  // styledPropsClasses.inlineStyles is intentionally ignored to maintain pure class-based styling
   const containerClasses = $derived(() => {
     const classes = ['code-container'];
     if (finalIsHighlighted) {
```
saurabhdaware: this can't be in cva?

PR #3042, opened 2025-12-04
Added Button.Svelte

kashishbehl:
## Description
image

image

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade-core/src/styles/Button/button.module.css
+++ b/packages/blade-core/src/styles/Button/button.module.css
@@ -1,4 +1,4 @@
-/* Button Styles */
+/* Button Base Styles */
 .btn {
   display: inline-flex;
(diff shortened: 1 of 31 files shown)


packages/blade-core/src/styles/Button/button.module.css
```diff
@@ -0,0 +1,335 @@
...
+  height: 36px;
+}
+
+.icon-only.large {
+  width: 48px;
+  height: 48px;
+}
+
+/* Color variants - Background colors */
+.color-primary.primary {
```
saurabhdaware: should we follow the `<key>-<value>` naming convention everywhere?

Then this class will become `.color-primary.variant-primary` right?

PR #3053, opened 2025-12-16
fix(mcp/README): mcp install link

saurabhdaware:
## Description

It was using cursor:// link which GitHub does not allow. Changed it to use web link

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/.changeset/large-mayflies-melt.md
+++ b/.changeset/large-mayflies-melt.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade-mcp": patch
+---
+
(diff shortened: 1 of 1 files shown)


.changeset/large-mayflies-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade-mcp": patch
+---
+
+fix(mcp/README): mcp install link
```
tewarig: 🤔 we don't need to add changeset to update readme

PR #3060, opened 2025-12-29
Migrated vite to rollup

kashishbehl:
## Description

1. Migrated vite to rollup
2. Fixed fonts issues

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade-svelte/src/components/Button/BaseButton/BaseButton.svelte
+++ b/packages/blade-svelte/src/components/Button/BaseButton/BaseButton.svelte
@@ -7,10 +7,7 @@
   import { getStyledPropsClasses } from '@razorpay/blade-core/utils';
   import {
     getButtonClasses,
-    buttonContentClass as _buttonContentClass,
-    buttonIconClass as _buttonIconClass,
-    loadingSpinnerClass as _loadingSpinnerClass,
-    loadingClass as _loadingClass,
+    getButtonTemplateClasses,
     getButtonTextColorToken,
     getButtonTextSizes,
     getButtonSpinnerSize,
@@ -18,11 +15,8 @@
     type SpinnerColor,
   } from '@razorpay/blade-core/styles';
 
-  // Workaround for Svelte tree-shaking: reference imported classes to prevent removal
-  const buttonContentClass = _buttonContentClass;
-  const buttonIconClass = _buttonIconClass;
-  const loadingSpinnerClass = _loadingSpinnerClass;
-  const loadingClass = _loadingClass;
+  // Get template classes via function call to prevent Svelte tree-shaking
+  const buttonClasses = getButtonTemplateClasses();
 
   let {
     children,
@@ -197,17 +191,25 @@
   // Extract styled props
   const styledProps = $derived(getStyledPropsClasses(rest));
 
-  // Combine classes
+  // Combine classes for button element
   const combinedClasses = $derived(() => {
     const classes = [
       baseButtonClasses,
-      isLoading ? loadingClass : '',
-      isPressed ? 'pressed' : '',
+      isLoading ? buttonClasses.loading : '',
       'focus-ring-parent',
     ];
     if (styledProps.classes) {
       classes.push(...styledProps.classes);
     }
+    return classes.filter(Boolean).join(' ');
+  });
+
+  // Animated content wrapper classes (for scale animation)
+  const animatedContentClasses = $derived(() => {
+    const classes = [
+      buttonClasses.animatedContent,
+      isPressed ? buttonClasses.pressed : '',
+    ];
     return classes.filter(Boolean).join(' ');
   });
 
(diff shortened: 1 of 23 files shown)


packages/blade-svelte/src/components/index.ts
```diff
@@ -0,0 +1,27 @@
+// Re-export Base components with their original names to ensure they're available
```
kashishbehl: One drawback of this approach is for blade and consumer to be on same svelte version as far as I have observed.

Another approach could be to let consumer compile the svelte

`
module.exports = {
  module: {
    rules: [
      {
        test: /\.svelte$/,
        // ⚠️ Must include node_modules/@razorpay
        include: [
          path.resolve(__dirname, 'src'),
          /node_modules\/@razorpay/  // Add this!
        ],
        use: {
          loader: 'svelte-loader',
          options: { preprocess: sveltePreprocess() }
        }
      }
    ]
  }
};
`

packages/blade-core/src/styles/BaseText/baseText.ts
```diff
@@ -2,7 +2,7 @@ import { cva } from 'class-variance-authority';
 import { utilityClasses } from '../utilities';
 // @ts-expect-error - CSS modules may not have type definitions in build
 import styles from './baseText.module.css';
-import { kebabCase } from '~utils';
+import { kebabCase } from '../../utils/lodashButBetter/kebabCase';
```
saurabhdaware: can we add this alias instead?

packages/blade-core/rollup.config.mjs
```diff
@@ -139,6 +139,10 @@ const getWebConfig = (inputs) => {
         modules: true,
         minimize: process.env.NODE_ENV === 'production',
         sourceMap: true,
+        // Use proper package import instead of relative node_modules path
+        inject(cssVariableName) {
+          return `import styleInject from 'style-inject';\nstyleInject(${cssVariableName});`;
+        },
```
saurabhdaware: what does this do 🤔 ?

packages/blade-svelte/src/components/Button/BaseButton/BaseButton.svelte
```diff
@@ -7,17 +7,23 @@
...
     getButtonSpinnerSize,
     type ActionStatesType as ButtonActionStatesType,
     type SpinnerColor,
   } from '@razorpay/blade-core/styles';
 
+  // Workaround for Svelte tree-shaking: reference imported classes to prevent removal
+  const buttonContentClass = _buttonContentClass;
+  const buttonIconClass = _buttonIconClass;
+  const loadingSpinnerClass = _loadingSpinnerClass;
+  const loadingClass = _loadingClass;
```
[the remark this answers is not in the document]
kashishbehl: Basicall what is happening is while compiling i
```
import {
 ButtonContentClass
} from '@rzorpay/blade-core.styles'
```

Svelte 5 compiler treat this as unused import since it;s reference is only made in the html template and not JS execution script.

import * as styles from '@razorpay/blade-core/styles'; we can do this as well but this would mean there is no treeshaking.

packages/blade-svelte/src/components/index.ts
kashishbehl: My bad not a svelte version issue. The re-export was needed so that when svelte 5 compiles for eg Button component which needs basebutton it is able to find both. Without re-exporting it was compoiling button but only loading the import statement of Basebutton and not the actual component

.github/workflows/release.yml
```diff
@@ -43,10 +43,18 @@ jobs:
...
         run: yarn publish-npm
+      - name: Build Blade Core if not built by changeset
+        if: steps.changesets.outputs.published == 'false'
+        working-directory: packages/blade-core
+        run: yarn build
       - name: Build Blade if not built by changeset
         if: steps.changesets.outputs.published == 'false'
         working-directory: packages/blade
         run: yarn build
+      - name: Build Blade Svelte if not built by changeset
```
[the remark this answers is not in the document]
kashishbehl: Added changes

packages/blade-svelte/src/components/index.ts
saurabhdaware: ohh. Is it common known issue? Like why does svelte not follow the import statements while compiling 🤔 ?

packages/blade-svelte/svelte.config.js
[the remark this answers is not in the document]
kashishbehl: Yes.

saurabhdaware (approved):
Looks good to me. Added one followup comment. 

Can you somewhere maybe add a README on how to use this library? you can add a Installation section in Svelte storybook I guess the way we have in react storybook? 

You can do it in separate PR if you want

packages/blade-svelte/src/components/Button/BaseButton/BaseButton.svelte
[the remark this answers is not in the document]
kashishbehl: Could nto find any open issue on svelte side but solved it with this approach https://github.com/razorpay/blade/pull/3062

PR #3071, opened 2026-01-07
feat: add logic to hide legends for line

tewarig:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade-coverage-extension/background-scripts/blade-coverage-script.js
+++ b/packages/blade-coverage-extension/background-scripts/blade-coverage-script.js
@@ -2,9 +2,9 @@
 const calculateBladeCoverage = (shouldHighlightNodes) => {
   const bladeElementExceptions = [
     // table library adds a div internally which we want to skip
-    '[data-blade-component="table-cell"] > div',
-    '[data-blade-component="table-header-cell"] > div',
-    '[data-blade-component="table-footer-cell"] > div',
+    '[data-blade-component="table-cell"] > div', 
+    '[data-blade-component="table-header-cell"] > div', 
+    '[data-blade-component="table-footer-cell"] > div'
   ];
 
   /**


packages/blade/src/components/Charts/CommonChartComponents/types.ts
```diff
@@ -64,6 +64,26 @@ type ChartTooltipProps = ComponentProps<typeof RechartsTooltip>;
...
+  /**
+   * Callback fired when the selection changes.
+   * Provides the new array of selected dataKeys.
+   */
+  onSelectedDataKeysChange?: (selectedDataKeys: string[]) => void;
+  /**
+   * Callback fired when a legend item is clicked.
+   * Provides the dataKey of the clicked legend item and whether it's currently selected.
+   */
+  onLegendClick?: ({ dataKey, isSelected }: { dataKey: string; isSelected: boolean }) => void;
```
anuraghazra: We dont need this.

packages/blade/src/components/Charts/CommonChartComponents/CommonChartComponents.web.tsx
```diff
@@ -315,14 +358,87 @@ const CustomSquareLegend = (props: {
...
+
+  // Sync selectedDataKeys to context's visibleDataKeys
+  React.useEffect(() => {
+    setVisibleDataKeys?.(selectedDataKeys);
+  }, [selectedDataKeys, setVisibleDataKeys]);
+
+  // Handle toggle
+  const handleToggle = React.useCallback(
+    (dataKey: string) => {
+      // Mark that user has interacted to prevent auto-reset behavior
```
anuraghazra: Whats hapepning here? What do we mean by auto-reset?

packages/blade/src/components/Charts/CommonChartComponents/types.ts
```diff
@@ -82,11 +102,17 @@ type ChartColorToken = ChartsCategoricalColorToken | ChartSequentialColorToken;
 type CommonChartComponentsStateType = {
   dataColorMapping?: DataColorMapping;
   chartName?: ChartName;
+  visibleDataKeys?: Set<string>;
```
anuraghazra: Why called is visible when we are opting for "selected" terminogloy.

packages/blade/src/components/Charts/LineChart/LineChart.stories.tsx
```diff
@@ -411,15 +411,41 @@ export const TinyLineChart: StoryFn<typeof ChartLine> = ({
...
+      if (historicalIsSelected) {
+        // Show both
+        setSelectedDataKeys(['historical', 'forecast']);
+      } else {
+        // Hide both
+        setSelectedDataKeys(newSelectedKeys.filter((key) => key !== 'forecast'));
+      }
+    } else {
+      setSelectedDataKeys(newSelectedKeys);
+    }
```
anuraghazra: Doesn't make much sesne to do all this actually. 

Ideally this should just work since newSelectedKeys will be the new state. 

```
const handleSelectionChange = (newSelectedKeys) => setSelectedDataKeys(newSelectedKeys)
```

packages/blade/src/components/Charts/LineChart/LineChart.stories.tsx
```diff
@@ -411,15 +411,41 @@ export const TinyLineChart: StoryFn<typeof ChartLine> = ({
 };
 
 // Forecast Line Chart Example
+// This example shows how to link multiple lines together so they toggle visibility together
+// when clicking the legend. The forecast line has showLegend={false} but toggles with historical.
 export const ForecastLineChart: StoryFn<typeof ChartLine> = () => {
+  const [selectedDataKeys, setSelectedDataKeys] = React.useState(['historical', 'forecast']);
+
+  const handleSelectionChange = (newSelectedKeys: string[]): void => {
```
anuraghazra: Why are we passing newSelectedKeys directly in the first param? use objects. We discussed this yesterday only.

packages/blade/src/components/Charts/CommonChartComponents/types.ts
tewarig: changed name to selected :)

packages/blade/src/components/Charts/LineChart/LineChart.stories.tsx
tewarig: it won't since in case of forecast charts we have 2 lines which act as one...  

and this is the way to use it.  

in this case we don't have any way to  make these 2 lines connect with one another...

so 
```
const handleSelectionChange = (newSelectedKeys) => setSelectedDataKeys(newSelectedKeys)
```
won't work.
```
   <ChartLine
            dataKey="historical"
            name="Historical Data"
            connectNulls={true}
            color="data.background.categorical.blue.moderate"
          />
          <ChartLine
            dataKey="forecast"
            name="Forecasted Data"
            strokeStyle="dashed"
            connectNulls={true}
            showLegend={false}
            color="data.background.categorical.blue.moderate"
          />
```

packages/blade/src/components/Charts/LineChart/types.ts
```diff
@@ -57,6 +57,10 @@ interface ChartLineProps {
    * @private
    */
   _totalLines?: number;
+  /**
+   * Whether to hide this line (controlled by legend click)
+   */
+  hide?: boolean;
```
saurabhdaware: what is this prop for? we use selectedDataKeys array for hiding and showing right?

saurabhdaware (approved):

PR #3075, opened 2026-01-11
feat: add support for text truncation in chart labels & secondary label

tewarig:
## Description

feat: add support for text truncation

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
+++ b/packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
@@ -167,7 +167,7 @@ const ChartBarWrapper: React.FC<ChartBarWrapperProps & TestID & DataAnalyticsAtt
     barChartModifiedChildrens,
     totalBars,
     dataColorMapping,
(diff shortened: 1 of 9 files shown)


packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
```diff
@@ -201,7 +201,9 @@ const ChartBarWrapper: React.FC<ChartBarWrapperProps & TestID & DataAnalyticsAtt
   }, [children, themeColors]);
 
   return (
-    <CommonChartComponentsContext.Provider value={{ chartName: 'bar', dataColorMapping }}>
+    <CommonChartComponentsContext.Provider
+      value={{ chartName: 'bar', dataColorMapping, chartData: data }}
+    >
```
tewarig: i think we should not use chartData... we can have map of secondayLabelValues.

saurabhdaware (approved):

PR #3081, opened 2026-01-13
feat: make legends for other charts clickable

tewarig:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Charts/AreaChart/types.ts
+++ b/packages/blade/src/components/Charts/AreaChart/types.ts
@@ -45,7 +45,7 @@ type ChartAreaProps = {
    * Whether to hide the area.
    *
    */
-   hide?: boolean;
+  hide?: boolean;
   /**
    * @private
    */
--- a/packages/blade/src/components/Charts/BarChart/types.ts
+++ b/packages/blade/src/components/Charts/BarChart/types.ts
@@ -37,7 +37,7 @@ type ChartBarProps = {
   showLegend?: boolean;
   /**
    * Whether to hide the bar.
-   * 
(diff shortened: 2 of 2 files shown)


packages/blade/src/components/Charts/DonutChart/DonutChart.web.tsx
```diff
@@ -285,6 +293,29 @@ const _ChartDonut: React.FC<ChartDonutProps> = ({
...
+    const map: Record<number, number> = {};
+    let filteredIdx = 0;
+    data.forEach((item, originalIdx) => {
+      if (selectedDataKeys.includes(sanitizeString(item.name as string))) {
+        map[filteredIdx] = originalIdx;
+        filteredIdx++;
+      }
+    });
+    return map;
+  }, [data, selectedDataKeys]);
```
swapnil-kr1: The map logic works, but you're iterating through data twice (once for filteredData, once for the map). Consider computing both in one pass:
const { filteredData, indexMap } = useMemo(() => {
  if (!selectedDataKeys) return { filteredData: data, indexMap: null };
  
  const filteredData: typeof data = [];
  const indexMap: Record<number, number> = {};
  
  data.forEach((item, originalIdx) => {
    if (selectedDataKeys.includes(sanitizeString(item.name as string))) {
      indexMap[filteredData.length] = originalIdx;
      filteredData.push(item);
    }
  });
  
  return { filteredData, indexMap };
}, [data, selectedDataKeys]);

swapnil-kr1 (approved):

PR #3087, opened 2026-01-14
feat: add GenUI SDK

anuraghazra:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/.changeset/nice-laws-thank.md
+++ b/.changeset/nice-laws-thank.md
@@ -1,0 +1,5 @@
+---
+'@razorpay/blade': minor
+---
+
+feat(blade): added GenUI components


packages/blade/src/components/GenUI/GenUIComponents.tsx
```diff
@@ -0,0 +1,1002 @@
...
+  | TableCellAmount
+  | TableCellIndicator
+  | TableCellBadge
+  | TableCellDate;
+
+type TableComponent = GenUIBaseComponent & {
+  component: typeof ComponentType.TABLE;
+  headers?: string[];
+  rows?: TableCellType[][];
+};
```
saurabhdaware: why do we define this schema separately for each component and not do some common `compound component -> flat structure` translation? So the way Card, Accordion, Table, etc gets represented stays same

saurabhdaware (commented):
Not a full review. Will add pending comments in some time

packages/blade/src/components/GenUI/GenUIComponents.tsx
```diff
@@ -0,0 +1,1002 @@
...
+      display="flex"
+      flexDirection="column"
+      alignItems="center"
+      justifyContent="center"
+      height="100%"
+      width="100%"
+    >
+      <Text size="medium" weight="semibold">
+        Error rendering component
+      </Text>
```
abhieshekumar: can this error state be given more freedom, we can check with design on once?

packages/blade/src/components/GenUI/GenUIComponents.tsx
```diff
@@ -0,0 +1,1002 @@
...
+
+const RenderTextComponent = memo(({ content }: TextComponent) => {
+  if (!content) return null;
+  return (
+    <Streamdown
+      controls={{ table: false, code: false, mermaid: false }}
+      rehypePlugins={[]}
+      components={{
+        h1: ({ children }) => (
+          <Heading marginBottom="spacing.4" size="large">
```
abhieshekumar: ques: why fix bottom margins?

packages/blade/src/components/GenUI/GenUIComponents.tsx
```diff
@@ -0,0 +1,1002 @@
...
+      if (isNaN(dateValue.getTime())) {
+        return <Text size="medium">{cell.value}</Text>;
+      }
+      const formattedDate = dateValue.toLocaleDateString('en-IN', {
+        day: '2-digit',
+        month: 'short',
+        year: 'numeric',
+      });
+      const formattedTime = dateValue.toLocaleTimeString('en-IN', {
+        hour: '2-digit',
```
abhieshekumar: we usually get date-time as epochs, and formatting also depends on use-case. Sharing a screenshot of existing table that we have

image

abhieshekumar (changes requested):
one more question : does consume create zod schema for components?

packages/blade/src/components/GenUI/GenUIComponents.tsx
anuraghazra: Can check, for now i'll make this a bit prettier.

packages/blade/src/components/GenUI/GenUIComponents.tsx
anuraghazra: Didn't understood, can you share a demo JSON representation?

packages/blade/src/components/GenUI/GenUIComponents.tsx
```diff
@@ -0,0 +1,1002 @@
+import React, { memo } from 'react';
+import { ErrorBoundary } from 'react-error-boundary';
+import { Streamdown } from 'streamdown';
+import {
```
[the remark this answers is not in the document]
saurabhdaware: > Plus from library side we can't really do lazy loading.

earlier we couldn't but now we're on ESM dist so its possible right?

> Can't use dynamic imports, LLM can arbitrarily return any component.

wouldn't it still work 🤔 ? Like it will import when the render of component is called right?

packages/blade/src/components/GenUI/GenUIComponents.tsx
saurabhdaware: Something like this. Single schema type for representation of any compound component

```jsx
<Parent parentProp="" parentProp2="">
  <Child>
    <NestedChild nestedChildProp="" />
  </Child>
</Parent>
```

```js
{
  componentName: 'Parent',
  parentProp: '',
  parentProp2: '',
  children: [{
    componentName: 'Child',
    children: [
      {
        componentName: 'NestedChild',
        nestedChildProp: ''
      }
    ]
  }]
}
```

packages/blade/src/components/Amount/__tests__/Amount.native.test.tsx
```diff
@@ -125,7 +125,7 @@ describe('<Amount />', () => {
   it('should check if formatAmountWithSuffix is returning the right value for humanize decimals and none', () => {
     setState({ locale: 'en-IN' });
     expect(getAmountByParts({ value: 1000.22, suffix: 'humanize', currency: 'INR' })).toEqual({
-      compact: 'T',
+      compact: 'K',
```
saurabhdaware: expected?

tewarig (dismissed):

abhieshekumar (dismissed):

PR #3088, opened 2026-01-15
Feat/input component foundation

kashishbehl:
## Description
The PR contains changes to add foundational setup for Input components in svelte

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


rohankokane-dev (commented):
## PR Review: Feat/input component foundation

### Summary

This PR adds foundational Form components for the Blade Svelte migration: **FormLabel**, **FormHint**, **CharacterCounter**, plus supporting CSS/tokens in blade-core and two new icons (EyeIcon, EyeOffIcon). The overall structure is reasonable -- CSS modules in blade-core, Svelte components using `$props()` and `$derived`, token re-exports in blade-svelte. However, there are several issues that need to be addressed before merging.

---

### Issues Found

#### HIGH: Incorrect label text `weight` -- diverges from React source
**File:** `packages/blade-svelte/src/components/Form/FormLabel/FormLabel.svelte` (line ~82)

The Svelte component uses `weight="semibold"` for the label text, but the React source (`FormLabel.tsx` line 149) uses `weight="medium"`. This is a visual regression.

```svelte
<!-- Current (wrong) -->
weight="semibold"

<!-- Should be -->
weight="medium"
```

---

#### HIGH: Incorrect label text color logic -- diverges from React source
**File:** `packages/blade-svelte/src/components/Form/FormLabel/FormLabel.svelte`

The Svelte version conditionally sets the label text color:
```ts
const textColor = $derived(
  isLabelLeftPositioned ? 'surface.text.gray.subtle' : 'surface.text.gray.muted'
);
```

But the React source (`FormLabel.tsx` line 147) uses `color="surface.text.gray.subtle"` **unconditionally** for the label text. The `surface.text.gray.muted` color is only used for the `(optional)` indicator. The conditional logic here will make top-positioned labels render with the wrong (muted) color.

---

#### HIGH: Inline styles violate Blade Svelte conventions
**File:** `packages/blade-svelte/src/components/Form/FormLabel/FormLabel.svelte` (lines ~74-75)

The component uses inline `style=` attributes with dynamic values:
```svelte
<div class="label-content" style="display: flex; flex-direction: {isLabelLeftPositioned ? 'column' : 'row'}; ...">
    <div style="display: flex; flex-direction: row; align-items: center; gap: 8px; ">
```

This violates the convention: **"Never use inline styles or `style:` directive"**. These should be CSS classes in the `form.module.css` file in blade-core, potentially using the existing class generation functions. Additionally, `gap: 8px` is a hardcoded pixel value instead of a design token.

---

#### HIGH: `<style>` block in blade-svelte component
**File:** `packages/blade-svelte/src/components/Form/FormLabel/FormLabel.svelte` (lines ~133-149)

The component has a `<style>` block with `.visually-hidden` and `.label-content` classes. Per convention, **CSS goes in `blade-core/src/styles/`, NEVER in `blade-svelte`**. The `visually-hidden` utility already exists in blade-core's CSS (used by Chip, Switch components). Move these to `form.module.css` and consume them via class getters, or use an existing `srOnly` / `VisuallyHidden` utility from blade-core.

---

#### MEDIUM: Missing export registration in `components/index.ts`
**File:** `packages/blade-svelte/src/components/index.ts`

The Form components (FormLabel, FormHint, CharacterCounter) and their types are not registered in the top-level `components/index.ts` barrel file. Per convention: *"Exports: register in both `{Component}/index.ts` and `components/index.ts`"*. Add:
```ts
// Form
export { FormLabel, FormHint, CharacterCounter } from './Form';
export type { FormLabelProps, FormHintProps, CharacterCounterProps } from './Form';
```

---

#### MEDIUM: Missing `role="alert"` or `aria-live` on error/success hint messages
**File:** `packages/blade-svelte/src/components/Form/FormHint/FormHint.svelte`

Error and success hints are dynamically shown/hidden but lack `role="alert"` or `aria-live="polite"` attributes. When validation state changes (e.g., user submits a form and error text appears), screen readers won't announce the new message. The `id` attributes on the hint divs are good for `aria-describedby` linkage, but the dynamic appearance of error/success messages also needs a live region announcement.

---

#### MEDIUM: React source uses `VisuallyHidden` for necessity indicator accessibility text
**File:** `packages/blade-svelte/src/components/Form/FormLabel/FormLabel.svelte`

The React source wraps the necessity indicator text AND `accessibilityText` together in a `<VisuallyHidden>` component (lines 128-133):
```jsx
<VisuallyHidden>
  {necessityIndicator !== 'none' && <Text>{necessityIndicator}</Text>}
  <Text>{accessibilityText}</Text>
</VisuallyHidden>
```

The Svelte version only renders `accessibilityText` in a visually-hidden span, but misses announcing the necessity indicator word ("required" / "optional") to screen readers. This is an accessibility regression.

---

#### MEDIUM: `FormHintProps.type` missing `@default` JSDoc tag
**File:** `packages/blade-svelte/src/components/Form/FormHint/types.ts`

The `type` prop has no `@default` tag. While it's required (no `?`), the JSDoc says "Type of hint to display" but doesn't explain the behavior contract. Consider adding a note that it controls which text variant renders.

---

#### LOW: Duplicate token definitions -- blade-core formTokens use raw numbers instead of `size` token references
**File:** `packages/blade-core/src/styles/Form/formTokens.ts`

The React source imports `size` from `~tokens/global` and uses `size[120]`, `size[176]` for `labelWidth`. The blade-core version uses raw numbers `120` and `176`. While the React `size` token likely maps to the same numbers, using raw values loses the indirection. If the global size tokens ever change, the blade-core values won't update. This is acceptable for now but worth noting for future alignment.

---

#### LOW: Stray blank line added to `Button.stories.svelte`
**File:** `packages/blade-svelte/src/components/Button/Button.stories.svelte`

An extra blank line was introduced. Trivial, but suggests an unintended edit.

---

#### LOW: `formTokens.ts` re-export file in blade-svelte may be unnecessary
**File:** `packages/blade-svelte/src/components/Form/formTokens.ts`

This file just re-exports everything from `@razorpay/blade-core/styles`. If consumers are already importing from blade-core, this indirection adds no value. Consider whether direct imports from `@razorpay/blade-core/styles` in the Svelte components (which FormHint and CharacterCounter already do) is the preferred pattern.

---

### Suggestions

1. **Use CVA** for the conditional `flex-direction` / `align-items` logic in FormLabel instead of inline styles. Add variants like `labelContent--top` / `labelContent--left` to the CSS module.

2. **Consolidate `visually-hidden`** into a shared utility class in blade-core (e.g., `utilities.module.css`) rather than duplicating it across components.

3. **Add a `FormLabel` usage example** in the JSDoc of `Form/index.ts`, per convention: *"Public component `index.ts` must have JSDoc with usage example"*.

4. **Consider adding `metaAttribute` to FormHint** for consistency with FormLabel. The React FormHint doesn't have it, but other Svelte components tend to add meta attributes for testing.

---

### Verdict: **Request Changes**

The core architecture (blade-core CSS, token separation, Svelte 5 runes) is solid, but the **incorrect text weight/color**, **inline styles**, and **CSS in blade-svelte** are blockers that need to be fixed before merging. The accessibility gaps (missing live region, missing necessity indicator a11y text) should also be addressed.

PR #3116, opened 2026-01-22
Feat/text input dependent components

kashishbehl:
## Description

The PR contains the support to add TextInput to Svelte with support of Icons, Divider, prefix/suffix and formatting options

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


rohankokane-dev (commented):
## PR Review: Feat/text input dependent components (#3116)

### Summary

This PR adds several foundational input components for the Blade Svelte migration:
- **blade-core**: Form styles (FormLabel, FormHint, CharacterCounter) and Input styles (BaseInput wrapper, styled input, visuals) with CSS modules and class generator functions
- **blade-svelte**: FormLabel, FormHint, CharacterCounter, BaseInput (with sub-components), TextInput, IconButton, EyeIcon/EyeOffIcon, Divider fix
- Storybook stories for TextInput with formatting examples

Overall this is a substantial, well-structured PR that follows the Blade Svelte architecture well. The CSS-in-blade-core / components-in-blade-svelte split is correct, types are well-documented with JSDoc, and the component layering (BaseInput -> TextInput) mirrors React. There are a handful of issues to address before merging.

---

### Blockers

**1. Merge Conflicts (blocking)**
The PR has merge conflicts with its base branch `feat/input-component-foundation`. The `mergeable` state is `CONFLICTING`. This must be resolved before merge.

**2. Semgrep CI Failure (blocking)**
The `semgrep-cloud-platform/scan` check is failing. Likely a vite vulnerability flagged in `yarn.lock`. This needs investigation and resolution.

**3. PR Title Check Failure (blocking)**
The `PR Title Check` CI job is failing. The PR title `Feat/text input dependent components` likely does not follow the conventional commit format (e.g., `feat(blade-svelte): add text input dependent components`).

---

### Issues Found

**Severity: High**

**H1. `useFormattedInput.ts` uses `$state` and `$effect` runes in a `.ts` file**
`packages/blade-svelte/src/components/Input/TextInput/useFormattedInput.ts`
Lines 76-77 use `$state()` and line 165 uses `$effect()`. Svelte runes only work in `.svelte` or `.svelte.ts` files, not plain `.ts` files. This will cause a runtime error. The `createFormattedInputState()` function appears to be dead code (TextInput.svelte duplicates the logic inline), but having broken exports in a module is still a problem. Either:
- Remove `createFormattedInputState()` entirely since it's unused, keeping only the pure utility functions (`formatValue`, `stripPatternCharacters`, `isUserCharacter`)
- Or rename to `.svelte.ts` if you intend to use the reactive version (though this contradicts the migration convention of keeping utilities in regular `.ts` files)

**H2. Inline styles used extensively -- violates "never use inline styles" convention**
Multiple components use `style=` attributes for layout:
- `BaseInput.svelte`: `style="display: flex; flex-direction: {isLabelLeftPositioned ? 'row' : 'column'}; width: 100%; "` on the container div, `style="display: flex; flex-direction: column; flex: 1; width: {isLabelLeftPositioned ? 'auto' : '100%'}; "` on the input+hint wrapper, `style="display: flex; flex-direction: row; justify-content: space-between; width: 100%; "` on the hint row
- `FormLabel.svelte`: `style="display: flex; flex-direction: {isLabelLeftPositioned ? 'column' : 'row'}; ..."` on the label-content div, and `style="display: flex; flex-direction: row; align-items: center; gap: 8px; "` using a hardcoded 8px value
- `TextInput.svelte`: `style="display: flex; align-items: center; gap: 8px; "` (hardcoded) and `style="margin-top: {hintMarginTop[size]}; margin-right: 4px; "` (hardcoded 4px)

Per Blade Svelte conventions, inline styles should not be used. These should be CSS module classes in `blade-core`. The hardcoded pixel values (`8px`, `4px`) are especially problematic -- use design tokens (`var(--spacing-3)` for 8px, etc.).

**H3. `<style>` block in FormLabel.svelte and IconButton.svelte -- CSS should be in blade-core**
`FormLabel.svelte` has a `<style>` block with `.visually-hidden` and `.label-content` classes. `IconButton.svelte` has a full `<style>` block with `.icon-button` styles including hardcoded fallback values like `var(--blade-border-radius-medium, 4px)` and `var(--blade-color-interactive-background-gray-default, rgba(0, 0, 0, 0.04))`. These CSS definitions should live in `blade-core/src/styles/` as CSS modules, not in the Svelte component files.

**Severity: Medium**

**M1. IconButton uses non-standard CSS variable names**
`IconButton.svelte` references `--blade-border-radius-medium`, `--blade-color-interactive-background-gray-default`, and `--blade-color-interactive-background-primary-faded`. The rest of the codebase uses tokens like `--border-radius-medium`, `--interactive-background-primary-faded` (without the `blade-` prefix). This will cause the fallback values to always be used instead of the theme tokens.

**M2. Missing `@default` JSDoc tags on several props**
- `TextInputProps`: `isDisabled`, `isRequired`, `autoFocus` all have default values in the component but no `@default` in JSDoc
- `BaseInputProps`: `isDisabled`, `isRequired`, `autoFocus`, `hideLabelText`, `hideFormHint` are missing `@default`
- `IconButtonProps`: `isDisabled`, `isHighlighted` are missing `@default`

Per convention, all props with default values must have `@default` tags.

**M3. Event handler JSDoc missing on TextInputProps**
The event handler props (`onChange`, `onFocus`, `onBlur`, `onClick`, `onKeyDown`) have no JSDoc comments. Even a brief description improves DX.

**M4. `BaseInput.svelte` does not guard event handlers behind `isDisabled` check**
The event handlers (`handleChange`, `handleFocus`, `handleBlur`, etc.) do not check `isDisabled` before executing. While the native `<input>` element won't fire events when `disabled`, the wrapper's `handleWrapperClick` and other handlers could still fire. The Blade convention is to always check `isDisabled` before executing event handlers.

**M5. `BaseInputWrapperProps` type not used for BaseInputWrapper.svelte**
`BaseInputWrapper.svelte` defines its own inline `Props` type instead of using the `BaseInputWrapperProps` from `types.ts`. This creates type drift -- the types file exports one definition while the component uses another (which adds `onclick` and `children` that are missing from the exported type).

**Severity: Low**

**L1. `hintMarginTop` token used as inline style value in TextInput**
`TextInput.svelte` line: `style="margin-top: {hintMarginTop[size]}; margin-right: 4px; "` -- the `hintMarginTop` tokens are string values like `'spacing.2'`, not CSS values like `var(--spacing-2)`. This will render as `margin-top: spacing.2` which is invalid CSS.

**L2. Dead code in `useFormattedInput.ts`**
The `createFormattedInputState()` function, `UseFormattedInputOptions` type, and `UseFormattedInputResult` type are never imported by any component. TextInput.svelte implements the formatting logic inline instead. Clean this up to reduce bundle size and confusion.

**L3. `InputVisualsOptions` parameter in `getInputVisualsClasses` only uses `type`**
The `InputVisualsOptions` type has `hasIcon`, `hasText`, `hasInteractionElement`, and `hasOtherTrailingElements` fields, but `getInputVisualsClasses()` only uses `type`. The unused fields are misleading.

**L4. `formTokens.ts` in blade-svelte re-exports everything from blade-core**
`packages/blade-svelte/src/components/Form/formTokens.ts` is just re-exports from `@razorpay/blade-core/styles`. Components could import directly from blade-core. While this is a convenience layer, it's an extra indirection that other Svelte components in the project don't use.

**L5. Component checklist items are all unchecked**
The PR description shows all checklist items (manual testing, kitchen sink story, interaction tests, changeset) as incomplete. Ensure these are addressed before merge.

---

### Suggestions

1. **Consider using CVA** for the inline style logic in BaseInput and FormLabel. The conditional layout styles (flex-direction based on label position, width based on label position) are good candidates for CVA variants or additional CSS module classes.

2. **The `leading`/`trailing` prop type detection pattern in TextInput is fragile**: checking `typeof leading === 'function' && 'name' in leading` to distinguish between an IconComponent and a Snippet is unreliable since both are functions and both can have a `name` property. Consider using separate props or a discriminated union approach.

3. **Template classes functions** (`getFormTemplateClasses()`, `getBaseInputTemplateClasses()`) are defined but never called anywhere. If these are needed to prevent tree-shaking in Svelte, ensure they're actually invoked in the components.

4. **`StyledBaseInput` controlled/uncontrolled mode**: The component tries to support both modes, but the interaction between `bind:value` on the parent's `<StyledBaseInput>` and the internal `displayValue` derived state may lead to double-update issues. Test this carefully with controlled mode where the parent holds the state.

---

### Verdict

**Request Changes** -- The inline styles (H2), CSS-in-component-files (H3), broken rune usage in `.ts` (H1), and the invalid `hintMarginTop` CSS value (L1) need to be fixed before this can be merged. The merge conflicts and CI failures are also blockers. The architectural approach is sound and most of the code is well-written; these are addressable issues.

PR #3126, opened 2026-02-02
feat: export FileUploadItem

abhieshekumar:
## Description

Export FileUploadItem to consumers.

## Changes

FileUpload internally uses FileUploadItem component to display the file which got uploaded. We are making  FileUploadItem available to consumers.

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


.changeset/fifty-doors-join.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
```
saurabhdaware:
```suggestion
"@razorpay/blade": minor
```

packages/blade/src/components/FileUpload/index.ts
```diff
@@ -1 +1,2 @@
 export * from './FileUpload';
+export * from './FileUploadItem';
```
saurabhdaware: just check if its getting exported from /components/index.ts as well

PR #3140, opened 2026-02-05
feat(blade-svelte): add counter component

rohankokane-dev:
## Description

This adds counter component to blade svelte

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade-svelte/src/components/Counter/BaseCounter/BaseCounter.svelte
```diff
@@ -0,0 +1,92 @@
...
+    emphasis = 'subtle',
+    size = 'medium',
+    testID,
+    ...rest
+  }: BaseCounterProps = $props();
+
+  // Format the counter content
+  const content = $derived(() => {
+    if (max !== undefined && value > max) {
+      return `${max}+`;
```
saurabhdaware: **Svelte 5 `$derived` misuse — reactivity bug**

Wrapping the expression in an arrow function makes `content` a stable function reference, not a reactive value. Changes to `value` or `max` from Storybook controls (or at runtime) will **not** trigger re-renders.

The same issue also affects `counterClasses` (line 52) and `combinedClasses` (line 57).

```svelte
// ❌ current — not reactive
const content = $derived(() => {
  if (max !== undefined && value > max) return `${max}+`;
  return `${value}`;
});
// ... template: {content()}

// ✅ fix
const content = $derived(max !== undefined && value > max ? `${max}+` : `${value}`);
// ... template: {content}
```

packages/blade-core/src/styles/Counter/index.ts
```diff
@@ -0,0 +1,13 @@
+export {
+  counterStyles,
+  getCounterClasses,
+  getCounterContentClasses,
+  getCounterTemplateClasses,
+  counterContentClass,
+  counterContentPaddingClass,
```
saurabhdaware, line 6: **Unused exports**

`counterHeight`, `counterHorizontalPadding`, and `getCounterTemplateClasses` are exported here (and re-exported from `blade-core/styles/index.ts`) but are not imported or used anywhere in the Svelte component. If these are not intended for external consumers, they should be removed to keep the public surface clean.

packages/blade-svelte/src/components/Counter/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default as Counter } from './Counter.svelte';
```
saurabhdaware, line 1: **Missing changeset**

The PR checklist item for adding a changeset is unchecked. Please run `yarn changeset` and commit the generated file so this component version bump is captured in the release notes.

saurabhdaware (commented):
Added 3 comments from few of the things that claude was flagging. Can you take a look once?. Rest looks good to me

saurabhdaware (approved):

PR #3172, opened 2026-02-12
feat(blade-svelte): add card component 

rohankokane-dev:
## Description

migrated Card component to svelte
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade-svelte/src/components/Card/CardRoot.svelte
```diff
@@ -0,0 +1,138 @@
...
+    cardRootStyles({
+      borderRadius,
+      shouldScaleOnHover,
+      isPressed: shouldScaleOnHover ? isPressed : false,
+      asLabel: as === 'label',
+    })
+  );
+
+  const styledProps = $derived(getStyledPropsClasses(rest));
+  const combinedClasses = $derived(() => {
```
saurabhdaware: `$derived` wrapping a function creates a derived that returns a closure, not a reactive string. The inner computation may not re-run correctly if references change.

```ts
// Current
const combinedClasses = $derived(() => {
  ...
  return classes.filter(Boolean).join(' ');
});
// used as: class={combinedClasses()}

// Fix
const combinedClasses = $derived.by(() => {
  ...
  return classes.filter(Boolean).join(' ');
});
// used as: class={combinedClasses}
```

packages/blade-svelte/src/components/Card/LinkOverlay.svelte
```diff
@@ -0,0 +1,74 @@
...
+    onFocus?.();
+  }
+
+  function handleBlur(): void {
+    onBlur?.();
+  }
+</script>
+
+{#if as === 'button'}
+  <button
```
saurabhdaware, line 54: Missing `type="button"` on this element. Without it, browsers default to `type="submit"`, which means clicking the card overlay inside a `<form>` will unintentionally submit the form.

```svelte
<button
  type="button"
  class={templateClasses.linkOverlay}
  ...
```

packages/blade-svelte/src/components/Card/CardSurface.svelte
```diff
@@ -0,0 +1,47 @@
...
+    padding = 'spacing.7',
+    borderRadius = 'medium',
+    height,
+    minHeight,
+    overflow,
+    overflowX,
+    overflowY,
+  }: {
+    children: Snippet;
+    backgroundColor?: string;
```
saurabhdaware: `backgroundColor` is typed as `string` but `cardSurfaceStyles()` (CVA) only accepts a specific union of values. Passing an invalid string will silently apply no class with no TypeScript error.

Narrow the type to match what CVA actually accepts:
```ts
import type { CardSurfaceVariants } from '@razorpay/blade-core/styles';

backgroundColor?: CardSurfaceVariants['backgroundColor'];
```

packages/blade-svelte/src/components/Card/CardHeaderText.svelte
```diff
@@ -0,0 +1,21 @@
...
+  import type { Snippet } from 'svelte';
+  import Text from '../Typography/Text/Text.svelte';
+  import { useCardContext } from './CardContext';
+
+  // Verify inside Card
+  useCardContext('CardHeaderText');
+
+  let { children, ...props }: {
+    children: Snippet | string;
+    [key: string]: unknown;
```
saurabhdaware: `[key: string]: unknown` is overly permissive — it bypasses TypeScript's prop checking entirely, letting any prop through silently. Use a specific intersection type or enumerate the known props instead.

packages/blade-svelte/src/components/Card/Card.svelte
saurabhdaware, line 1: Image

This prop was deprecated so maybe we can remove it? We can also remove hover scale animation from other interactive card stories

saurabhdaware:
can you check the failing checks and conflicts?

saurabhdaware (dismissed):

saurabhdaware (approved):

PR #3251, opened 2026-03-03
feat(blade): added reasoning traces in ChatInput

anuraghazra:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/ChatMessage/ChatMessage.web.tsx
```diff
@@ -42,6 +42,10 @@ const _ChatMessage: React.ForwardRefRenderFunction<BladeElementRef, ChatMessageP
     maxWidth,
     thumbnails,
     onThumbnailClick,
+    reasoningTraces,
+    reasoningStatus,
```
saurabhdaware:
```suggestion
    reasoningStatus = 'loading',
```

?

Don't see any handling for undefined internally so should we set default value?

PR #3319, opened 2026-03-28
fix(docs): correct installation steps count from 3 to 4

### LINKED ISSUE #3077: Docs: four steps are listed in the installation section, but the header indicates only three.
Image

Four steps are listed in the installation section, but the header indicates only three.

abdulmusawwir2:
Fixes #3077

Corrected the installation steps count in the documentation from 3 to 4 to match the actual steps listed.

This resolves the mismatch between header and content.

### DIFF
--- a/.changeset/solid-jars-kiss.md
+++ b/.changeset/solid-jars-kiss.md
@@ -1,5 +1,0 @@
----
-'@razorpay/blade': patch
----
-
-fix docs: correct installation steps count from 3 to 4
--- a/.changeset/solid-jars-kiss.md
+++ b/.changeset/solid-jars-kiss.md
@@ -1,0 +1,5 @@
+---
+'@razorpay/blade': patch
+---
+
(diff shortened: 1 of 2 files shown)


anuraghazra (dismissed):

anuraghazra (approved):

Mrsandeep27 (approved):
Verified — the installation guide has 4 tabs (`1. Basic Setup`, `2. Install Fonts`, `3. Setup Motion React`, `4. TypeScript`) so "Follow these 4 steps" matches the actual count. The `3` was stale. Small, correct fix — LGTM.

Minor nit (non-blocking): the line directly above ("Before you install the package, make sure that you have performed the following steps:") refers to the prerequisite accordion, not the tabbed steps below. Since those two "steps" concepts live on the same page, maintainers might eventually want to rephrase the lower sentence to something like "Follow these 4 steps to integrate Blade:" for clarity — but that's a follow-up, not a blocker for this PR.

PR #3333, opened 2026-04-14
feat(blade-mcp): claude code compatibile guidelines

saurabhdaware:
## Description

Changes:

- Moved cursor rules to .agents/skills with symlink for claude code
- Moved parts of skills to its references to optimize the AI context (from 300 lines to 50 lines)
- Added some tests for versioning of skill

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade-mcp/skillTemplate/SKILL.md
saurabhdaware, line 1: same cursor rule. I just moved it to different directory and moved some types into separate references file to reduce the context size

packages/blade-mcp/src/utils/skillUtils.ts
saurabhdaware, line 1: This is same as `cursorRulesUtils.ts` file but for skill now instead

packages/blade-mcp/src/utils/tokens.ts
```diff
@@ -7,7 +7,30 @@ const PROJECT_ROOT_DIRECTORY = join(__dirname, '..', '..');
...
+const SKILL_VERSION_STRING = `version: '${SKILL_VERSION}'`;
+
+const SKILL_TEMPLATE_DIRECTORY = join(PROJECT_ROOT_DIRECTORY, 'skillTemplate');
+const BLADE_SKILL_FILE_PATH = join(SKILL_TEMPLATE_DIRECTORY, 'SKILL.md');
+const SKILL_REFERENCES_DIRECTORY = join(SKILL_TEMPLATE_DIRECTORY, 'references');
+
+const SKILL_FILE_NAME = 'SKILL.md';
+const SKILL_DIRECTORY_NAME = 'ui-code-guidelines';
+
+const CONSUMER_SKILL_RELATIVE_PATH = `.agents/skills/${SKILL_DIRECTORY_NAME}/${SKILL_FILE_NAME}`;
```
swapnil-kr1, line 21: **Bug: Grep pattern won't match the SKILL.md template**

`SKILL_VERSION_STRING` uses single quotes:
```ts
const SKILL_VERSION_STRING = `version: '${SKILL_VERSION}'`;
```

But the grep in `CHECK_SKILL_VERSION_DESCRIPTION` searches for double quotes:
```bash
grep -o 'version: "[0-9.]*"' .agents/skills/ui-code-guidelines/SKILL.md
```

This will never match `version: '1.0.0'` in the actual SKILL.md, so the client will always send `0` as the version fallback — causing a false "skill missing" error on every tool call and prompting the user to recreate the skill every time.

**Fix:** Change the grep to match single quotes:
```bash
grep -o "version: '[0-9.]*'" .agents/skills/ui-code-guidelines/SKILL.md
```

_added via claude_

packages/blade-mcp/skillTemplate/references/styled-props-types.md
```diff
@@ -0,0 +1,159 @@
+# StyledPropsBlade Type Definitions
+
+These are the types supported in components that accept styled props.
+
+```ts
+type SpacingValueType =
+  | `spacing.${keyof Spacing}`
+  | `${string}${'px' | '%' | 'vh' 'vw'}`
```
swapnil-kr1: **Syntax error: Missing `|` operator**

```ts
| `${string}${'px' | '%' | 'vh' 'vw'}`
```

Should be:
```ts
| `${string}${'px' | '%' | 'vh' | 'vw'}`
```

Missing the `|` between `'vh'` and `'vw'`.

_added via claude_

packages/blade-mcp/src/utils/__tests__/skillUtils.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      });
+      const text = result!.content[0].text;
+      expect(text).toContain(outdatedVersion);
+      expect(text).toContain(SKILL_VERSION);
+      expect(text).toContain(mockProjectRootDirectory);
+    });
+
+    it('should return undefined when skill version matches SKILL_VERSION', () => {
+      expect(
+        shouldCreateOrUpdateSkill(SKILL_VERSION, mockProjectRootDirectory, true),
```
swapnil-kr1, line 131: **Test name contradicts the assertion**

The test is named *"should return undefined when skill is up to date"* but actually asserts that the result contains `"Blade skill is outdated"`. The description and the assertion are opposite.

Suggested rename:
```ts
it('should return outdated error when version is "0.0.0"', () => {
```

_added via claude_

packages/blade-mcp/src/tools/createBladeSkill.ts
```diff
@@ -0,0 +1,172 @@
+import { join, basename } from 'path';
```
swapnil-kr1, line 1: **Breaking change: MCP tool renamed without deprecation**

The tool name changed from `create_blade_cursor_rules` → `create_blade_skill`. Any existing client integrations (e.g. Cursor MCP configs, scripts, or plugins) calling the old tool name will get a "tool not found" error.

Consider either:
- Keeping the old tool as a deprecated alias that delegates to the new one
- Or bumping a major version and documenting this as a breaking change in a changeset

_added via claude_

saurabhdaware: Expected. Client won't have to make any change either ways. The index of tools itself will get updated so previous name won't exist in the index

swapnil-kr1 (dismissed):

swapnil-kr1 (approved):

PR #3335, opened 2026-04-15
feat: alert, breadcrumb and avatar

rohankokane-dev:
## Description

Adds 3 components to the blade-svelte package.
1. Alert
2. Breadcrumb
3. Avatar
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


saurabhdaware (commented):
## Review Summary

**Prettier lint error** in `packages/blade-core/src/styles/Avatar/index.ts` line 20 — the multiline `export type {}` block needs to be collapsed to a single line:

```ts
// Current (failing):
export type {
  AvatarWrapperVariants,
  AvatarButtonVariants,
  AvatarGroupVariants,
} from './avatar';

// Fix:
export type { AvatarWrapperVariants, AvatarButtonVariants, AvatarGroupVariants } from './avatar';
```

Run `yarn prettier --write packages/blade-core/src/styles/Avatar/index.ts` to fix.

---

### Code Concerns

**1. `$derived(() => ...)` closure pattern in `BaseLink.svelte`**

```ts
const combinedClasses = $derived(() => { ... return classes.join(' '); });
```

Then used as `class={combinedClasses()}`. This creates a derived that returns a closure rather than a reactive string. Should use `$derived.by(...)` to return the string directly:

```ts
const combinedClasses = $derived.by(() => { ... return classes.join(' '); });
// used as: class={combinedClasses}
```

**2. `AvatarGroup` maxCount uses direct DOM mutation in `$effect`**

The `$effect` that hides children via `el.style.display = 'none'` is fragile — it relies on `groupEl.children` matching 1:1 with Avatar components. If any child ever wraps in a fragment or Svelte wrapper, this breaks silently. A more robust approach would be passing a `hiddenIndices` set down via Svelte context so each Avatar can hide itself reactively.

**3. `test-results/.last-run.json` should not be committed**

This is a local Playwright test artifact. A `.gitignore` entry is added but the file itself appears in the diff. Please remove it from tracking (`git rm --cached test-results/.last-run.json`).

saurabhdaware (dismissed):

saurabhdaware (approved):

PR #3343, opened 2026-04-21
feat: Chip component svelte

rohankokane-dev:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Chip/chipTokens.ts
```diff
@@ -1,11 +1,11 @@
+import type { DotNotationSpacingStringToken } from '~utils/types';
+import { size } from '~tokens/global';
+import type { DurationString, EasingString } from '~tokens/global';
 import type { ChipGroupProps } from './ChipGroup';
```
saurabhdaware, line 1: **[Critical — CI Blocking]** This import reordering is breaking the `Validate Source Code` CI check with 2 `import/order` ESLint errors.

The project's lint rule requires relative imports (`./ChipGroup`, `./types`) to come _before_ external path alias imports like `~utils/types`. The original import order was correct — please revert these changes in this file.

packages/blade-core/src/styles/Chip/chip.ts
```diff
@@ -0,0 +1,254 @@
...
+});
+
+export function getChipInnerClasses(props: ChipInnerVariants): string {
+  return chipInnerCva(props);
+}
+
+/** @deprecated Use getAnimatedChipClasses + getChipInnerClasses instead */
+export function getChipClasses(props: ChipInnerVariants): string {
+  return getChipInnerClasses(props);
+}
```
saurabhdaware: **[Significant]** `getChipClasses` is being added as a new export in a new file but is already marked `@deprecated`. If nothing currently calls it, it shouldn't be added at all. If it's needed for a migration path from existing consumers, the PR description should explain the context.

packages/blade-svelte/src/components/Chip/Chip.svelte
```diff
@@ -0,0 +1,203 @@
...
+  );
+  const fontSize = $derived(textSizes.fontSize[_size]);
+  const lineHeight = $derived(textSizes.lineHeight[_size]);
+  const letterSpacing = $derived(textSizes.letterSpacing[_size]);
+  const iconSize = $derived(iconSizes[_size]);
+
+  const isStringChildren = $derived(typeof children === 'string');
+  const snippetChildren = $derived(
+    !isStringChildren ? (children as Snippet | undefined) : undefined,
+  );
```
saurabhdaware: **[Significant]** This check will always be `false` at runtime. In Svelte, `children` is always a `Snippet` function — text content passed between tags is wrapped in a snippet automatically, never a raw string. This means the `isStringChildren` branch (and the `BaseText` rendering path that depends on it) is dead code.

The snippet path should be the single rendering path for children. Consider simplifying to just render the snippet.

saurabhdaware (changes requested):
/review-pr skill's review

packages/blade-core/src/styles/Chip/chip.ts
saurabhdaware: we can skip the deprecated parts in svelte if there are alternatives

saurabhdaware (dismissed):
Just check the failing checks once. Rest looks good

saurabhdaware (dismissed):

saurabhdaware (approved):

PR #3344, opened 2026-04-21
fix: align blade-core utilities

rohankokane-dev:
## Description

Align blade-core utilities with React tokens (normalize casing, fill missing variants, prune broken)

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade-core/src/styles/utilities.module.css
+++ b/packages/blade-core/src/styles/utilities.module.css
@@ -163,6 +163,62 @@
 .padding-y-spacing-10 { padding-top: var(--spacing-10); padding-bottom: var(--spacing-10); }
 .padding-y-spacing-11 { padding-top: var(--spacing-11); padding-bottom: var(--spacing-11); }
 
+/* Padding Top - Spacing Tokens */
+.padding-top-spacing-0 { padding-top: var(--spacing-0); }
+.padding-top-spacing-1 { padding-top: var(--spacing-1); }
+.padding-top-spacing-2 { padding-top: var(--spacing-2); }
+.padding-top-spacing-3 { padding-top: var(--spacing-3); }
+.padding-top-spacing-4 { padding-top: var(--spacing-4); }
+.padding-top-spacing-5 { padding-top: var(--spacing-5); }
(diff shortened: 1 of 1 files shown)


packages/blade-core/src/styles/utilities.module.css
```diff
@@ -419,6 +483,58 @@
 .color-interactive-text-static-black-muted { color: var(--interactive-text-static-black-muted); }
 .color-interactive-text-static-black-disabled { color: var(--interactive-text-static-black-disabled); }
 
+/* Interactive Icon Colors */
+/* Note: sets `color:` so SVGs using `fill="currentColor"` pick it up. */
+.color-interactive-icon-primary-normal { color: var(--interactive-icon-primary-normal); }
+.color-interactive-icon-primary-subtle { color: var(--interactive-icon-primary-subtle); }
+.color-interactive-icon-primary-muted { color: var(--interactive-icon-primary-muted); }
+.color-interactive-icon-primary-disabled { color: var(--interactive-icon-primary-disabled); }
```
saurabhdaware, line 491: should these not be in icon's CSS file?

saurabhdaware (approved):
Looks good for now but long term I think we should have some objective differentiation on what goes in common global CSS file vs what stays in component-level file. The lines are a bit blurry as of now. E.g. should icon-related tokens be global or in icon's css file, should text-related tokens be global or in text's CSS file, etc


We can do one analysis and refactor at the end to optimize the CSS file a bit. Otherwise the bundle size of global CSS will become too much.


Another thing we can check is if its even worth keeping `var(--iteractive-icon-primary-normal)` type of CSS variables in tokens file or if we should just directly build classes themselves. That will save bundle size a bit. Right now it feel repeatitive.

packages/blade-core/src/styles/utilities.module.css
rohankokane-dev: these are not used by the icons but there wrappers. right now the whole file was made in 1:1 mapping with the token tree in parity with react's. we will discuss on how to consolidate these.

PR #3348, opened 2026-04-24
feat(blade-svelte): Accordion component

rohankokane-dev:
## Description

Migrates the React Accordion component to Svelte 5.

## Changes

- Adds `packages/blade-svelte/src/components/Accordion/`
- Adds `packages/blade-core/src/styles/Accordion/`
- Re-exports Accordion from `packages/blade-svelte/src/components/index.ts`

## Component Checklist

- [ ] Add KitchenSink Story
- [x] Add changeset


packages/blade-svelte/src/components/Accordion/AccordionItem.svelte
```diff
@@ -0,0 +1,63 @@
...
+
+  const isExpanded = $derived(accordionCtx.expandedIndex === itemIndex);
+  const variant = $derived(accordionCtx.variant);
+  const numberOfItems = $derived(accordionCtx.numberOfItems);
+  const isLastItem = $derived(
+    itemIndex !== undefined && itemIndex === numberOfItems - 1,
+  );
+  const showDivider = $derived(!isLastItem || variant === 'transparent');
+
+  const collapsibleBodyId = `accordion-body-${Math.random().toString(36).slice(2, 9)}`;
```
rohankokane-dev: **Bug: `Math.random()` for `collapsibleBodyId` breaks SSR determinism**

This ID is used for `aria-controls` on the button and `id` on the body region. `Math.random()` produces different values on server vs client, causing an SSR hydration mismatch.

Consider using a deterministic counter-based ID (e.g., pass the `itemIndex` through) or use a utility like `crypto.randomUUID()` only if SSR is not a concern. Since Blade is a design system and SSR support is likely expected, a pattern like:

```ts
const collapsibleBodyId = `accordion-body-${itemIndex}`;
```

...would be simpler and deterministic. If multiple Accordions can coexist on the same page, you could add a unique accordion-level ID prefix passed through context.

packages/blade-svelte/src/components/Accordion/Accordion.svelte
```diff
@@ -0,0 +1,101 @@
...
+  let _itemCounter = 0;
+  let _numberOfItems = $state(0);
+
+  // Reset counter before each render cycle
+  $effect.pre(() => {
+    // This runs before children render
+    _itemCounter = 0;
+  });
+
+  const registerItem = (): number => {
```
rohankokane-dev, line 43: **Potential issue: Item registration relies on render-order side effects**

This pattern calls `registerItem()` during component initialization (outside `$effect`) and uses `$effect.pre` to reset the counter. This works as long as:

1. AccordionItems are never conditionally rendered (e.g., `{#if condition}<AccordionItem>...`)
2. AccordionItems are never dynamically reordered

If an item is removed from the middle, the remaining items will re-register with stale indices on the next render cycle because `registerItem()` is called once during init (line 25 of AccordionItem.svelte), not on every render.

For now this matches the React behavior (which also uses a counter), but worth noting as a known limitation in the component docs or as a comment.

packages/blade-svelte/src/components/Accordion/AccordionItemBody.svelte
```diff
@@ -0,0 +1,133 @@
...
+          if (bodyRef) bodyRef.style.display = 'none';
+        });
+      }
+    }
+  };
+
+  const metaAttrs = metaAttribute({ name: MetaConstants.AccordionItemBody });
+  const analyticsAttrs = $derived(makeAnalyticsAttribute(rest));
+  const bodyA11y = $derived(
+    makeAccessible({ role: 'region', hidden: !isExpanded }),
```
rohankokane-dev, line 103: **Accessibility: Body region is missing `aria-labelledby`**

Per WAI-ARIA Accordion pattern, the collapsible region (`role="region"`) should have `aria-labelledby` pointing to the button that controls it. Currently, the button has `aria-controls` pointing to the body, but the body region does not have `aria-labelledby` pointing back to the button.

Suggested fix: Give the accordion button an `id` (e.g., `accordion-button-{itemIndex}`) and add `aria-labelledby` to this region element:

```ts
const bodyA11y = $derived(
  makeAccessible({ role: 'region', hidden: !isExpanded, labelledBy: collapsibleButtonId }),
);
```

Reference: https://www.w3.org/WAI/ARIA/apg/patterns/accordion/

packages/blade-svelte/src/components/Accordion/AccordionItemBody.svelte
```diff
@@ -0,0 +1,133 @@
...
+  bind:this={bodyRef}
+  id={collapsibleBodyId}
+  class={templateClasses.collapsibleContent}
+  ontransitionend={onTransitionEnd}
+  {...bodyA11y}
+  {...metaAttrs}
+  {...analyticsAttrs}
+>
+  <div class={templateClasses.body}>
+    <BaseText
```
rohankokane-dev, line 117: **Nit: Wrapping `BaseText` around all body children forces text styling on non-text content**

When `children` is a Snippet (custom content like forms, cards, etc.), it still gets wrapped in `BaseText` with `color="surface.text.gray.subtle"` and specific font sizing. This means custom slot content inherits text color/size styling that may not be desired.

In the React implementation, string children get wrapped in `<Text>` but non-string `ReactNode` children are rendered directly without a text wrapper.

Consider conditionally wrapping only the string case in `BaseText` and rendering the snippet directly for non-string children.

packages/blade-svelte/src/components/Accordion/AccordionItemHeader.svelte
```diff
@@ -0,0 +1,219 @@
...
+        {/if}
+
+        {#if trailing}
+          <div class={templateClasses.headerTrailing}>
+            {@render trailing()}
+          </div>
+        {/if}
+
+        <div class={chevronClass}>
+          <ChevronDownIcon size="large" color={chevronColor} />
```
rohankokane-dev, line 208: **Nit: `ChevronDownIcon` size is hardcoded to `"large"` regardless of accordion size**

The icon is always rendered with `size="large"`, but when the accordion `size` is `"medium"`, the chevron may appear disproportionately large relative to the header content (medium header slot height is 20px vs large at 28px).

Consider deriving the chevron size from `accordionSize`:

```svelte
<ChevronDownIcon size={accordionSize === 'large' ? 'large' : 'medium'} color={chevronColor} />
```

Check against the React source to confirm whether the original also uses a fixed size or adapts.

packages/blade-svelte/src/components/Accordion/Accordion.svelte
```diff
@@ -0,0 +1,101 @@
...
+  const handleExpandChange = (nextExpandedIndex: number) => {
+    if (controlledExpandedIndex !== undefined) {
+      onExpandChange?.({ expandedIndex: nextExpandedIndex });
+    } else {
+      internalExpandedIndex = nextExpandedIndex;
+      onExpandChange?.({ expandedIndex: nextExpandedIndex });
+    }
+  };
+
+  setAccordionContext(() => ({
```
rohankokane-dev, line 63: **Context reactivity: Props passed to context getter may go stale**

When props like `showNumberPrefix`, `variant`, and `size` change, the context getter function captures the current values at the time `setAccordionContext` is called (component init). Since these are destructured from `$props()`, they are reactive in Svelte 5 and the getter should re-evaluate correctly because it reads them from the closure.

However, verify that child components that call `getAccCtx()` inside a `$derived` block actually re-derive when these parent props change. The pattern looks correct (getter function + `$derived` consumer), but it is worth testing:

1. Toggle `showNumberPrefix` dynamically
2. Switch `variant` from `transparent` to `filled`
3. Switch `size` from `large` to `medium`

If any of these fail to update the children, the context getter pattern needs adjustment.

packages/blade-svelte/src/components/Accordion/AccordionItemBody.svelte
```diff
@@ -0,0 +1,133 @@
...
+  const descriptionFontSize = $derived(accordionSize === 'large' ? 100 : 75);
+  const descriptionLineHeight = $derived(accordionSize === 'large' ? 100 : 75);
+
+  const isStringChildren = $derived(typeof children === 'string');
+  const snippetChildren = $derived(!isStringChildren ? (children as Snippet) : undefined);
+
+  let bodyRef: HTMLDivElement | undefined = $state(undefined);
+  let isInitialRender = true;
+
+  $effect(() => {
```
rohankokane-dev, line 37: **Animation `$effect` cleanup: No cleanup function means potential stale RAF callbacks**

The `$effect` schedules nested `requestAnimationFrame` callbacks but does not return a cleanup function. If the component unmounts (or the effect re-runs) while a RAF is pending, the callback can execute on a detached/undefined `bodyRef`.

The null checks (`if (!bodyRef) return`) mitigate crashes, but the stale callbacks still execute needlessly. Consider storing RAF IDs and canceling them in a cleanup return:

```ts
$effect(() => {
  const expanded = isExpanded;
  // ... setup ...
  let raf1: number, raf2: number;
  // ... raf1 = requestAnimationFrame(() => { ... raf2 = requestAnimationFrame(...) });
  return () => {
    cancelAnimationFrame(raf1);
    cancelAnimationFrame(raf2);
  };
});
```

This is a minor robustness improvement, not a blocker.

rohankokane-dev (commented):
## Accordion Svelte 5 Migration -- Code Review Summary

### Overall Assessment

Solid migration. The component architecture (Accordion > AccordionItem > AccordionItemHeader + AccordionItemBody) maps cleanly from React to Svelte 5. Context pattern using getter functions, CVA-based styling in blade-core, and CSS modules are all well-structured. The stories provide good coverage across variants.

---

### Key Findings

**Must-fix (1):**

1. **`Math.random()` for `collapsibleBodyId` breaks SSR** -- The `aria-controls`/`id` pairing uses a non-deterministic random value. This will cause hydration mismatches in SSR scenarios. Use a counter-based or index-based deterministic ID instead.

**Accessibility (1):**

2. **Missing `aria-labelledby` on the body region** -- Per the WAI-ARIA Accordion pattern, the `role="region"` element should reference the controlling button via `aria-labelledby`. Currently only the forward reference (`aria-controls` on button -> body) exists. The reverse reference is missing.

**Behavioral concerns (1):**

3. **Animation `$effect` does not clean up RAF callbacks** -- The nested `requestAnimationFrame` calls in AccordionItemBody have no cleanup. While the null checks prevent crashes, stale callbacks still fire after unmount. Minor but worth addressing for robustness.

**Design parity (2):**

4. **`BaseText` wraps all body children including Snippets** -- Custom slot content (forms, cards, etc.) inherits text styling. The React version only wraps string children in `<Text>`.

5. **ChevronDownIcon is hardcoded to `size="large"`** -- When accordion `size="medium"`, the chevron may be disproportionate. Verify against React source whether this is intentional.

---

### What looks good

- **Svelte 5 patterns**: Proper use of `$props()`, `$derived`, `$state`, `$effect`, `$effect.pre`, and `@render` snippets. Context uses the getter-function pattern correctly.
- **CVA + CSS modules**: Clean separation of styling concerns in blade-core. Compound variants for filled+position+expanded states are well-modeled.
- **Type safety**: `types.ts` is comprehensive with proper JSDoc. Analytics attribute index signatures are correct.
- **Accessibility fundamentals**: `role="heading"` with `level=3` on the header wrapper, `aria-expanded` and `aria-controls` on the button, `role="region"` on the body, `disabled` attribute on the button -- all present.
- **Keyboard navigation**: Since a native `<button>` is used, Enter/Space toggling works for free.
- **Controlled/uncontrolled pattern**: Clean implementation with `expandedIndex` vs `defaultExpandedIndex`.
- **Stories**: Good coverage -- basic, number prefix, icons, controlled, custom header/body, filled variant composition, header variants, disabled state.

---

### Checklist items from PR template

The PR template shows unchecked items for Component Status Page update, manual browser testing, KitchenSink story, and interaction tests. These should be addressed before merge.

---

### Minor notes

- The `Divider` import in AccordionItemHeader.svelte appears unused in some code paths (it is used inside the expanded divider section, so this is fine).
- The `$effect.pre` counter-reset pattern for item registration is clever but fragile with conditional rendering. This matches the React behavior, so it is acceptable, but worth documenting as a known limitation.
- The `.cursor/` agent/rule changes included in this PR are unrelated tooling updates. Consider splitting them into a separate PR for cleaner review history.

packages/blade-svelte/src/components/Accordion/AccordionItem.svelte
rohankokane-dev: Fixed in 9cc12d937. Replaced `Math.random()` with a module-scoped monotonically-increasing counter — deterministic in component-tree order, SSR-safe, and collision-free.

packages/blade-svelte/src/components/Accordion/Accordion.svelte
rohankokane-dev: Acknowledged. The pattern matches the React Accordion's index-based registration. Will revisit if conditional-rendering edge cases surface.

packages/blade-svelte/src/components/Accordion/AccordionItemBody.svelte
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will add `aria-labelledby` on the body region (pointing to the header button id) in a follow-up to complete the WAI-ARIA accordion bidirectional link.

packages/blade-svelte/src/components/Accordion/AccordionItemBody.svelte
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will narrow the `BaseText` wrapping to string children only (matching React) in a follow-up.

packages/blade-svelte/src/components/Accordion/AccordionItemHeader.svelte
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will check the React source for the intended icon-size mapping and parameterize.

packages/blade-svelte/src/components/Accordion/Accordion.svelte
rohankokane-dev: Acknowledged. The getter pattern + closure access keeps reads live, so the concern is theoretical here. Noting for future awareness.

packages/blade-svelte/src/components/Accordion/AccordionItemBody.svelte
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will return a cleanup that cancels pending RAFs in a follow-up.

saurabhdaware (dismissed):

saurabhdaware (approved):

PR #3349, opened 2026-04-24
feat(ChatInput, FileUploadItem): disabled submit on file error, trash icon in error state, autoscroll, fixed file item width

saurabhdaware:
## Summary

- **ChatInput `isSubmitDisabled`**: Adds `isSubmitDisabled` prop to disable the submit button for reasons beyond empty input (e.g. a required field elsewhere on the page is unfilled, or an async validation is in flight). OR-combined with the internal guard so both the submit button and keyboard Enter are blocked.

- **FileUploadItem trash icon in error state**: Shows a trash/remove icon alongside the Re-upload button in the error state when `onRemove` is provided, letting users discard a failed upload without retrying first. Backward-compatible — omitting `onRemove` shows only the Re-upload link.

- **ChatInput file preview autoscroll**: The file preview row automatically scrolls to the rightmost position whenever files change, so the most recently added file is always visible.

- **ChatInput FileUploadItem fixed width**: Each `FileUploadItem` in the ChatInput file preview row now has a fixed width of `200px` (replacing the previous `minWidth`/`maxWidth`/`flexGrow`/`flexBasis` approach) for a consistent, predictable layout.

## Changes

| File | What changed |
|------|-------------|
| `ChatInput/types.ts` | Add `isSubmitDisabled?: boolean` prop with JSDoc |
| `ChatInput/useChatInput.ts` | Accept prop in `UseChatInputProps`; combine with internal disabled state |
| `ChatInput/ChatInput.web.tsx` | Destructure and forward `isSubmitDisabled`; add `fileScrollRef` + `useEffect` for autoscroll; set fixed `width="200px"` on each `FileUploadItem` |
| `ChatInput/docs/ChatInput.stories.tsx` | Add `SubmitDisabled` story (checkbox gate demo); add `WithManyFiles (Autoscroll)` story |
| `ChatInput/_decisions/decisions.md` | Document fixed 200px width and autoscroll-to-latest behaviour |
| `FileUpload/FileUploadItem.tsx` | Render trash icon + divider in error state when `onRemove` is provided |
| `FileUpload/docs/FileUploadItem.stories.tsx` | Update `ErrorState` to include `onRemove`; add `ErrorStateReuploadOnly` story |

## Test plan

- [ ] `ChatInput` with `isSubmitDisabled={true}` — submit button is disabled even when there is text in the input
- [ ] `ChatInput` with `isSubmitDisabled={true}` — pressing Enter does not submit
- [ ] `ChatInput` with `isSubmitDisabled={false}` (default) — submit button enabled when there is text (unchanged behavior)
- [ ] `FileUploadItem` with `status="error"` and `onRemove` provided — both Re-upload link and trash icon are visible; clicking trash fires `onRemove`
- [ ] `FileUploadItem` with `status="error"` and no `onRemove` — only Re-upload link is visible (backward-compatible)
- [ ] `ChatInput` with an error-status file and `onFileRemove` — trash icon appears in file chip and fires `onFileRemove`
- [ ] `ChatInput` with multiple files — file preview row scrolls to show the latest file when a new file is added
- [ ] Each `FileUploadItem` in the ChatInput preview row is exactly 200px wide regardless of file name length

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


packages/blade/src/components/ChatInput/useChatInput.ts
```diff
@@ -90,7 +90,8 @@ const useChatInput = (
 
   const hasText = textValue.trim().length > 0;
   const hasFiles = files.length > 0;
-  const isSubmitDisabled = !hasText && !hasFiles;
+  const hasErrorFiles = files.some((f) => f.status === 'error');
+  const isSubmitDisabled = (!hasText && !hasFiles) || hasErrorFiles;
```
swapnil-kr1, line 94: Should we also block submit while a file is mid-upload (`status === 'uploading'`)? Today a user can submit while an upload is in flight and `onSubmit` will receive a fileList with a not-yet-ready file. Worth aligning the guard:

```suggestion
  const hasErrorFiles = files.some((f) => f.status === 'error');
  const hasUploadingFiles = files.some((f) => f.status === 'uploading');
  const isSubmitDisabled = (!hasText && !hasFiles) || hasErrorFiles || hasUploadingFiles;
```

swapnil-kr1, line 94: UX concern with the implicit disable: when `hasErrorFiles` flips submit to disabled, the user gets no feedback for *why* Enter no-ops or the submit button greys out — the error indicator lives only on the failed file chip. A user who types a long message and hits Enter will silently lose the action.

Options to consider:
- bubble a tooltip on the disabled submit button (e.g. "Remove or retry the failed file to send"),
- or surface it via the existing `validationState` / `errorText` slot in `ChatInput`.

Also — please document this new contract in `_decisions/decisions.md`. Since v1 of this PR exposed `isSubmitDisabled` as an explicit prop and you reverted to implicit, the rationale should live next to the code so future contributors don't re-litigate it.

packages/blade/src/components/ChatInput/ChatInput.web.tsx
```diff
@@ -154,18 +169,14 @@ const _ChatInput: React.ForwardRefRenderFunction<BladeElementRef, ChatInputProps
...
-                flexBasis={1}
-                minWidth="160px"
-                maxWidth="200px"
-                key={file.id ?? file.name}
-                file={file}
-                onRemove={() => handleFileRemove(file)}
-                onDismiss={() => handleFileDismiss(file)}
-                onReupload={onFileReupload ? () => onFileReupload({ file }) : undefined}
-              />
+              <BaseBox key={file.id ?? file.name} flexShrink={0} width="200px">
```
swapnil-kr1: **`200px` is a magic number.** `_decisions/decisions.md` already documents 200px as the contract; please move it into `chatInputTokens.ts` (e.g. `chatInputFilePreviewItemWidth`) and reference from both places so the source of truth stays singular.

packages/blade/src/components/FileUpload/FileUploadItem.tsx
```diff
@@ -125,6 +125,13 @@ const FileUploadItem = memo(
                 >
                   Re-upload
                 </BaseLink>
+                {onRemove ? (
+                  <IconButton
+                    accessibilityLabel="Remove File"
+                    icon={TrashIcon}
+                    onClick={() => onRemove({ file })}
```
swapnil-kr1: Both the trash icon here (error state) and the trash icon in the success branch use `accessibilityLabel="Remove File"`. For users on AT, when both an error chip and success chips coexist in the same list, the labels become ambiguous. Consider including the file name in the label, e.g. `accessibilityLabel={\`Remove ${file.name}\`}` — same applies to the success-state trash icon and the uploading-state close icon for consistency.

packages/blade/src/components/ChatInput/useChatInput.ts
saurabhdaware: I agree with this actually but that feedback already exists clearly as error state in FileUploadItem so don't think so a separate feedback is needed

saurabhdaware: Done — merged into the existing `hasErrorFiles` check as an OR condition so we don't add a second variable:

```ts
const hasErrorFiles = files.some((f) => f.status === 'error' || f.status === 'uploading');
```

Commit: 6619d4184

saurabhdaware: The visual feedback already lives on the FileUploadItem chip itself (red error state / spinner for uploading) — adding a tooltip or errorText on the submit button would duplicate that signal without adding clarity.

Documented this rationale in `_decisions/decisions.md` under **Submit Disabled Behaviour** so future contributors don't re-litigate it. Commit: 6619d4184

packages/blade/src/components/ChatInput/ChatInput.web.tsx
saurabhdaware: Done — extracted to `chatInputFilePreviewItemWidth` in `chatInputTokens.ts` and referenced from `ChatInput.web.tsx`. Commit: 6619d4184

packages/blade/src/components/FileUpload/FileUploadItem.tsx
saurabhdaware: Done — all icon buttons in `FileUploadItem` now include the file name in the label (`Remove ${name}`, `Preview ${name}`), covering the uploading (close), error (trash), and success (eye/trash) states. Commit: 6619d4184

swapnil-kr1 (dismissed):

swapnil-kr1 (approved):

PR #3361, opened 2026-04-30
feat: migration agents enhacements

rohankokane-dev:
## Description

Native Task Tool — Replaces run-agent CLI with Cursor Task Tool
Parallel component migrations — Plan, Execute, and Verify phases now run across all components simultaneously instead of one at a time
Isolated git worktrees per component — each migration runs in its own branch and directory, so components don't conflict
Failure isolation — a failed component doesn't block the rest of the batch
Per-component PRs — each migrated component gets its own PR opened automatically
agent-base-directory.mdc rule — prevents agents from writing artifacts to the wrong directory when running inside a worktree
Self-heal for misplaced artifacts — orchestrator detects and moves files if an agent resolved paths relative to the wrong base

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


.cursor/commands/migrate-to-svelte.md
```diff
@@ -1,142 +1,442 @@
...
+2. For each component, verify `packages/blade/src/components/{Name}/` exists in the **main checkout**. If missing → drop from batch and warn user.
+3. Check `packages/blade-svelte/src/components/index.ts` in the main checkout. If `{Name}` is already exported → drop from batch with "already migrated".
+4. Capture the **main checkout absolute path** (`pwd`) once. You will use it to construct worktree absolute paths for the Task prompts.
+
+### 0.2 Allocate slots and ports
+
+Number the surviving components `1..N` (in input order).
+
+- **React Storybook port (shared, all slots): `9009`** (the script's default).
+- **Svelte Storybook port (per slot `i`): `6000 + i*100 + 7`** → slot 1 = `6107`, slot 2 = `6207`, …, slot 5 = `6507`.
```
rohankokane-dev, line 89: **Port formula may collide on large batches.** The formula `6000 + i*100 + 7` produces ports 6107, 6207, ..., 6507. But if someone runs a batch of 10+ components, slot 10 = port 7007, slot 60 = port 12007 (above ephemeral range start on some systems). Also, there is no check that the computed port is actually free before assigning it.

Consider adding:
1. A max batch size guard (e.g., cap at 5-8 parallel slots) with a queuing mechanism for larger batches.
2. A port-availability check (`lsof -i :PORT` or `curl -s`) during allocation, with a retry/offset if occupied.

.cursor/commands/migrate-to-svelte.md
```diff
@@ -1,142 +1,442 @@
...
-| Unexpected error in any phase         | Save current state to artifacts, ask user how to proceed |
+## Failure Handling Summary
+
+| Scenario                               | Effect on this component    | Effect on others |
+| -------------------------------------- | --------------------------- | ---------------- |
+| Component already migrated             | Dropped at Step 0           | Continue         |
+| React source not found                 | Dropped at Step 0           | Continue         |
+| Plan agent fails                       | `❌ plan-failed`            | Continue         |
+| Unmigrated dependency, no workaround   | `❌ blocked-by-dep`         | Continue         |
+| Dependency cycle detected              | `❌ dep-cycle` (all in cyc) | Continue         |
```
rohankokane-dev, line 428: **Dependency cycle detection is listed in the failure table but never implemented.** The failure handling summary references `dep-cycle` as a status, and it appears in the Status Vocabulary at the bottom, but Step 2 (Dependency Resolution) only handles two cases: dep-in-batch (`deferred-to-next-batch`) and dep-not-in-batch (`blocked-by-dep`). There is no code path that actually detects or marks a cycle.

If A depends on B and B depends on A and both are in the batch, Step 2 would mark A as deferred (because B is in the batch) and B as deferred (because A is in the batch), leaving both stuck with no actionable resolution.

Either add explicit cycle detection logic in Step 2 or remove `dep-cycle` from the failure table and status vocabulary to avoid confusion.

.cursor/commands/migrate-to-svelte.md
```diff
@@ -1,142 +1,442 @@
...
+# corrupts module resolution. Strip any prior copies before re-cloning.
+# (Worktree is fresh out of `git worktree add` on the happy path — these only fire on retry.)
+rm -rf "$WT/node_modules" \
+       "$WT/packages/blade/node_modules" \
+       "$WT/packages/blade-svelte/node_modules" \
+       "$WT/packages/blade-core/node_modules"
+
+# APFS only: `cp -c` returns non-zero on non-APFS filesystems, so we fall through
+# to `yarn install` per worktree (~30–60s each). Stderr is logged, not swallowed.
+if cp -cR "$(pwd)/node_modules" "$WT/node_modules" 2>>/tmp/blade-migration-clone.log; then
```
rohankokane-dev, line 118: **APFS clone error log appends without rotation.** The clone stderr is appended (`2>>`) to a fixed path `/tmp/blade-migration-clone.log`. Over multiple pipeline runs this log grows unbounded and mixes output from different runs, making it hard to debug a specific failure.

Suggestion: include a timestamp or run-id in the filename, or truncate at the start of each pipeline run:
```bash
CLONE_LOG="/tmp/blade-migration-clone-$(date +%Y%m%d-%H%M%S).log"
```

.cursor/commands/migrate-to-svelte.md
```diff
@@ -1,142 +1,442 @@
...
+   - **Also in this batch** → mark this component `🟡 deferred-to-next-batch` (dep name in `Notes`), drop it from the active set, tell the user to re-run it after the dep's PR merges.
+   - **Not in this batch, no workaround** → mark `❌ blocked-by-dep` (dep name in `Notes`); tell the user to migrate the dep first.
+
+Remaining components proceed independently.
+
+## Step 3: Plan Gate (per-component, sequential)
+
+For each successfully-planned, non-deferred, non-blocked component:
+
+1. **Skip the gate** if the component's tier is `simple`. Auto-approve and continue.
```
rohankokane-dev, line 248: **Auto-approving `simple` tier plans skips human review for an undocumented classification.** Where is `tier` defined and how is it computed? The Plan agent's discovery report template presumably sets it, but the criteria for what qualifies as `simple` vs other tiers are not documented anywhere in this PR. This means the auto-skip behavior is unpredictable.

Consider either:
1. Documenting the tier classification criteria in this file or linking to where it is defined.
2. Making the auto-approve opt-in (e.g., a flag the user passes to the orchestrator) rather than silently skipping review.

.cursor/agents/verify.md
```diff
@@ -11,57 +11,73 @@ You are a Senior UI Engineer. Your job is to ensure the Svelte implementation ac
 > Phase 4: Verify the migrated component through static checks, API parity,
 > visual comparison, and inline fixes. Owns ALL fix strategies.
 
-> You ARE the verifier. Execute these steps inline. Do NOT invoke `run-agent` — that is the orchestrator's role.
+> You ARE the verifier. You own the convergent loop. When parity gaps are found, spawn Execute (patch mode) via the Task tool yourself.
```
rohankokane-dev, line 14: **Verify agent now self-spawns Execute, but there is no iteration cap on the patch loop within a single verify invocation.** The verify agent doc says it returns to Step 1 after spawning a patch, but the orchestrator's Step 5 only sees the terminal status. If the verify agent enters an infinite patch-verify cycle internally (e.g., a flaky build), the orchestrator has no visibility or timeout.

The verify agent has its own `iteration > 6` safety cap (inherited from before), but it would be worth explicitly calling that out here in the orchestrator's Step 5 description so readers understand the blast radius is bounded.

packages/blade-svelte/.storybook/preview.css
```diff
@@ -1,22 +1,39 @@
...
-/* Story container - provides padding and background similar to React StoryCanvas */
-.sb-story {
+#storybook-root {
+  width: 100%;
+  height: 100vh;
+  overflow: auto;
   padding: 2rem;
-  background-color: var(--surface-background-gray-subtle, #f5f5f5);
-  min-height: 100vh;
+  background-color: var(--surface-background-gray-intense);
```
rohankokane-dev, line 19: **Background token changed from `gray-subtle` to `gray-intense` without explanation.** The old CSS used `--surface-background-gray-subtle` (with a fallback to `#f5f5f5`), while the new version uses `--surface-background-gray-intense` with no fallback. The comment says this mirrors React blade's `StoryCanvas` decorator, which is a good justification, but worth double-checking that `gray-intense` is actually the token used in the React Storybook decorator and not `gray-subtle`. A missing fallback value means if the token is undefined, the background disappears entirely.

Consider adding a fallback:
```css
background-color: var(--surface-background-gray-intense, #f5f5f5);
```

.cursor/rules/agent-base-directory.mdc
```diff
@@ -0,0 +1,18 @@
+---
+description: How sub-agents in the Svelte migration pipeline resolve paths relative to a base directory.
+globs:
+alwaysApply: false
```
rohankokane-dev, line 4: **The `agent-base-directory.mdc` rule has `alwaysApply: false`, which means agents will only see it if they explicitly include it.** The plan, execute, and verify agents all have "Read `agent-base-directory.mdc` before starting" in their Include sections, which is correct. But if someone creates a new agent or forgets the include, the path resolution silently falls back to the wrong directory.

Consider setting `alwaysApply: true` since this rule is critical for correctness in any worktree-based workflow, and the cost of applying it unnecessarily is near-zero (it is only 18 lines).

rohankokane-dev (commented):
## Review Summary

This is a substantial and well-thought-out enhancement to the migration pipeline. The shift from sequential `run-agent` CLI calls to parallel Task-tool-based orchestration with isolated git worktrees is a meaningful architectural improvement. The self-heal mechanism for misplaced artifacts shows good defensive thinking. Overall the documentation quality is high and the failure isolation model is solid.

### What works well

- **Worktree isolation model** is well designed. Each component gets its own branch and directory, preventing cross-contamination. The APFS clone optimization for `node_modules` is a pragmatic macOS-specific optimization with a clean `yarn install` fallback.
- **Failure isolation** is excellent. The failure table is comprehensive and the "failed component does not block others" principle is consistently applied.
- **Self-heal for misplaced artifacts** (Step 1.5) is a smart defensive measure against agent CWD confusion.
- **`agent-base-directory.mdc`** is a good addition that addresses a real problem (agents resolving relative paths against wrong CWD).
- **`npm run` to `yarn` corrections** across all agents fix a real inconsistency with the project's package manager.
- **Storybook preview.css/preview.js improvements** are solid -- targeting `#storybook-root` and `.docs-story` with `layout: 'fullscreen'` is the right approach to match React's StoryCanvas decorator.
- **Verify agent now owns the patch loop** (spawns Execute itself via Task tool) -- this is cleaner than the orchestrator mediating every round-trip.

### Issues flagged (see inline comments)

1. **Port allocation has no bounds or availability check** -- large batches can collide with system ports, and occupied ports are not detected.
2. **`dep-cycle` status is documented but never implemented** -- Step 2 would silently deadlock both components in a mutual dependency.
3. **APFS clone log grows unbounded** -- `/tmp/blade-migration-clone.log` is appended without rotation across runs.
4. **`simple` tier auto-approval is undocumented** -- the classification criteria are not defined in this PR.
5. **Verify's internal iteration cap (6) is not mentioned in the orchestrator** -- readers of the orchestrator doc have no visibility into the blast radius.
6. **CSS token changed without fallback** -- `gray-subtle` to `gray-intense` with no fallback value.
7. **`agent-base-directory.mdc` should arguably be `alwaysApply: true`** -- it is critical for correctness and the cost of applying it unnecessarily is near-zero.

### Minor observations (not worth inline comments)

- **PR title typo**: "enhacements" should be "enhancements".
- The hardcoded example ports in the verify agent's URL table (lines 273-274) still show `6007` and `9009` as literal values, which contradicts the parameterized `{SveltePort}` / `{ReactPort}` approach used everywhere else in the same file. These should be `{SveltePort}` and `{ReactPort}` for consistency.
- The orchestrator doc is 442 lines. It is well-structured, but consider extracting the shell script blocks (Step 0.3 APFS clone, Step 6.2 commit/push/PR) into standalone scripts under `.cursor/scripts/` to reduce cognitive load and make them independently testable.

.cursor/commands/migrate-to-svelte.md
rohankokane-dev: Tracking — out of scope for this must-fix pass. Realistic batch sizes (≤5 today) stay below 7000 so this isn't biting in practice. Will add a free-port check + bounds guard in a follow-up.

.cursor/commands/migrate-to-svelte.md
rohankokane-dev: Tracking — out of scope for this must-fix pass. The `dep-cycle` status was added speculatively in the vocabulary; will either implement detection in Step 2 or remove the status in a follow-up.

.cursor/commands/migrate-to-svelte.md
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will switch the log to a per-run filename (or truncate at start) in a follow-up.

.cursor/commands/migrate-to-svelte.md
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will document the `tier` classification criteria in the plan agent's discovery report template in a follow-up.

.cursor/agents/verify.md
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will surface the `iteration > 6` cap in the orchestrator's Step 5 description in a follow-up.

packages/blade-svelte/.storybook/preview.css
rohankokane-dev: Acknowledged. Verified against React's `StoryCanvas` decorator: it does use `surface.background.gray.intense`. Will add a literal hex fallback in a follow-up for safety.

.cursor/rules/agent-base-directory.mdc
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will flip to `alwaysApply: true` in a follow-up since the rule is short and worktree-correctness-critical.

saurabhdaware (dismissed):

saurabhdaware (approved):

PR #3362, opened 2026-04-30
feat(blade-svelte): Skeleton component

rohankokane-dev:
## Description

Migrates the React Skeleton component to Svelte 5.

## Changes

- Adds `packages/blade-svelte/src/components/Skeleton/`
- Adds `packages/blade-core/src/styles/Skeleton/`
- Re-exports Skeleton from `packages/blade-svelte/src/components/index.ts`

## Additional Information

Artifacts (in worktree):

- Discovery report: `.cursor/artifacts/Skeleton/discovery-report.md`
- Migration plan: `.cursor/artifacts/Skeleton/migration-plan.md`
- Verification report: `.cursor/artifacts/Skeleton/verification-report.md`
- Screenshots: `.cursor/artifacts/Skeleton/screenshots/`

## Component Checklist

- [ ] Add KitchenSink Story
- [x] Add changeset


packages/blade-core/src/styles/Skeleton/skeleton.module.css
```diff
@@ -0,0 +1,73 @@
+.skeleton {
+  display: block;
+  opacity: 0;
+  background-color: var(--interactive-background-gray-default);
+  animation-name: skeleton-fade-in, skeleton-pulse;
+  animation-duration: var(--duration-2xgentle), 1320ms;
```
rohankokane-dev, line 6: **Magic number for animation duration**

The `1320ms` here is the sum of `--duration-xmoderate` (360ms) + `--duration-2xgentle` (960ms), matching the React PulseAnimation logic. However, this is a magic number that will silently drift if those token values ever change.

Consider using `calc(var(--duration-xmoderate) + var(--duration-2xgentle))` instead, which stays in sync with the design tokens automatically:

```css
animation-duration: var(--duration-2xgentle), calc(var(--duration-xmoderate) + var(--duration-2xgentle));
```

packages/blade-core/src/styles/Skeleton/skeleton.module.css
```diff
@@ -0,0 +1,73 @@
...
+  animation-delay: 0ms, var(--duration-2xgentle);
+  animation-timing-function: var(--easing-standard), var(--easing-standard);
+  animation-iteration-count: 1, infinite;
+  animation-direction: normal, alternate;
+  animation-fill-mode: forwards;
```
rohankokane-dev, line 11: **Missing `prefers-reduced-motion` media query**

The Skeleton animation runs infinitely with no respect for users who have reduced motion enabled. Consider adding:

```css
@media (prefers-reduced-motion: reduce) {
  .skeleton {
    animation: none;
    opacity: 1;
    background-color: var(--interactive-background-gray-default);
  }
}
```

This is an accessibility best practice for any infinite animation. The React version also lacks this, but the Svelte migration is a good opportunity to add it since this CSS lives in blade-core and would benefit both implementations.

packages/blade-svelte/src/components/Skeleton/Skeleton.svelte
```diff
@@ -0,0 +1,96 @@
...
+  const styledPropsInlineStyle = $derived.by(() => {
+    const entries = Object.entries(styledProps.inlineStyles ?? {});
+    if (entries.length === 0) return undefined;
+    return entries.map(([key, value]) => `${kebabCase(key)}: ${value}`).join('; ');
+  });
+
+  const finalStyle = $derived(combineStyleStrings(ownInlineStyle, styledPropsInlineStyle) || undefined);
+
+  const a11yAttrs = makeAccessible({ hidden: true });
+  const metaAttrs = metaAttribute({ name: MetaConstants.Skeleton, testID });
```
rohankokane-dev, line 86: **`metaAttrs` not reactive to `testID` changes**

`metaAttrs` is computed at initialization time (no `$derived`), so if `testID` changes at runtime, the `data-testid` attribute will not update. Same issue with `a11yAttrs`, though that one uses a static `{ hidden: true }` so it's fine in practice.

For consistency with the reactive pattern used throughout the rest of this component, consider wrapping in `$derived`:

```ts
const metaAttrs = $derived(metaAttribute({ name: MetaConstants.Skeleton, testID }));
```

This ensures the component behaves correctly if consumers conditionally toggle `testID`.

packages/blade-core/src/styles/Skeleton/skeleton.ts
```diff
@@ -0,0 +1,223 @@
...
+      '2xlarge': styles['radius-2xlarge'],
+      max: styles['radius-max'],
+      round: styles['radius-round'],
+    },
+  },
+});
+
+export const skeletonClass = styles.skeleton;
+
+export type SkeletonFlexProps = {
```
rohankokane-dev, line 42: **Unused export: `skeletonClass`**

`skeletonClass` is exported from `skeleton.ts` and re-exported from the barrel `index.ts`, but it's never consumed anywhere in this PR (nor anywhere else in the codebase). The `getSkeletonClasses()` function already returns the base class via the CVA call internally.

If this is intended as a public escape hatch for consumers who need the raw class name, that's fine -- but consider documenting the use case or removing it to keep the public API surface minimal.

packages/blade-svelte/src/components/Skeleton/Skeleton.svelte
```diff
@@ -0,0 +1,96 @@
...
+      flexShrink,
+      flexBasis,
+      order,
+    }),
+  );
+
+  const styledPropsInlineStyle = $derived.by(() => {
+    const entries = Object.entries(styledProps.inlineStyles ?? {});
+    if (entries.length === 0) return undefined;
+    return entries.map(([key, value]) => `${kebabCase(key)}: ${value}`).join('; ');
```
rohankokane-dev, line 80: **Duplicated `styledPropsInlineStyle` logic -- consider consolidating**

This manual kebab-case conversion of `styledProps.inlineStyles` into a CSS string is likely duplicated across other Svelte Blade components. If `combineStyleStrings` already exists as a shared util, consider also adding a shared `inlineStylesToString()` helper in blade-core to avoid reimplementing this pattern in every component.

Not blocking, but worth tracking if this pattern keeps appearing in the Svelte migration.

packages/blade-svelte/src/components/Skeleton/Skeleton.svelte
```diff
@@ -0,0 +1,96 @@
...
+  });
+
+  const finalStyle = $derived(combineStyleStrings(ownInlineStyle, styledPropsInlineStyle) || undefined);
```
rohankokane-dev, line 83: **Line length exceeds 100 chars**

Nit: this line is quite long. Consider breaking it for readability:

```ts
const finalStyle = $derived(
  combineStyleStrings(ownInlineStyle, styledPropsInlineStyle) || undefined,
);
```

packages/blade-svelte/src/components/Skeleton/Skeleton.stories.svelte
```diff
@@ -0,0 +1,541 @@
...
+                  width="30%"
+                  borderRadius="medium"
+                  marginBottom="spacing.3"
+                />
+                <Skeleton height="20px" width="50%" borderRadius="medium" />
+              </div>
+              <Skeleton
+                height="65px"
+                borderRadius="medium"
+                marginBottom="spacing.3"
```
rohankokane-dev, line 200: **Stories: consider extracting duplicated skeleton card layout**

The Complex and CardExample stories share a large amount of duplicated layout markup (the two-card loading state pattern in Complex is copy-pasted between left and right panels). This makes the stories harder to maintain.

Consider extracting the shared skeleton card pattern into a small helper snippet or a `SkeletonCard` fragment within the stories file to reduce duplication and improve readability. Not blocking -- stories are not production code -- but it would make future story updates easier.

rohankokane-dev (commented):
## Skeleton Svelte 5 Migration -- Code Review Summary

**Overall**: Good migration. The component is well-structured, follows Svelte 5 idioms (`$props()`, `$derived`, `$derived.by`), and correctly mirrors the React Skeleton's behavior. The blade-core extraction of styles is clean and will benefit both React and Svelte consumers.

### What looks good

- **Svelte 5 patterns**: Correct use of `$props()` with destructuring + rest, `$derived` for reactive computations, and `$derived.by` for the more complex styled props inline style logic.
- **Architecture**: Clean separation between blade-core (styles, CVA, utility mappings) and blade-svelte (component). The `getSkeletonClasses` / `getSkeletonInlineStyle` split is well thought out.
- **Accessibility**: `aria-hidden="true"` on each skeleton node matches the React implementation and the documented decision (consumers should use `aria-busy` on the container). Stories demonstrate the `aria-busy` pattern correctly.
- **Type safety**: Props interface is well-documented with JSDoc comments and correctly extends `StyledPropsBlade`.
- **Animation fidelity**: The CSS keyframes match the React styled-components implementation (fade-in then infinite pulse with the same timing curve and delays).

### Issues to address

1. **[suggestion] `prefers-reduced-motion`**: The infinite pulse animation has no reduced-motion fallback. This is an accessibility gap. See inline comment for the recommended CSS.

2. **[suggestion] Magic number `1320ms`**: The pulse duration is hardcoded as `1320ms` instead of using `calc(var(--duration-xmoderate) + var(--duration-2xgentle))`. If token values change, this will silently break.

3. **[nit] `metaAttrs` reactivity**: `metaAttribute(...)` is called once at init time, not wrapped in `$derived`. If `testID` ever changes at runtime, the DOM won't update. Low impact since most consumers set testID statically, but worth fixing for correctness.

### Minor / Non-blocking

4. **Unused export `skeletonClass`**: Exported but never consumed. Consider removing or documenting the intended use case.
5. **Duplicated `styledPropsInlineStyle` pattern**: The kebab-case + join logic will likely repeat across every Svelte Blade component. Worth extracting into a shared util at some point.
6. **Stories duplication**: The Complex story has copy-pasted layout markup between left/right panels.

### Checklist items from PR description

The PR notes several unchecked items (Component Status Page, manual testing in other browsers, KitchenSink story). Please ensure those are tracked before merge.

No blocking issues. The component is correct and ready to ship after addressing the suggestions above.

packages/blade-core/src/styles/Skeleton/skeleton.module.css
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will switch to `calc(var(--duration-xmoderate) + var(--duration-2xgentle))` in a follow-up so the duration stays in sync with token changes.

packages/blade-core/src/styles/Skeleton/skeleton.module.css
rohankokane-dev: Tracking — out of scope for this must-fix pass. Good a11y suggestion; will add the reduced-motion media query (and likely backport to React Skeleton CSS too) in a follow-up.

packages/blade-svelte/src/components/Skeleton/Skeleton.svelte
rohankokane-dev: Acknowledged. Matches the existing pattern across blade-svelte components; will revisit holistically rather than per-component.

packages/blade-core/src/styles/Skeleton/skeleton.ts
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will remove (or document an external consumer for) the export in a follow-up.

packages/blade-svelte/src/components/Skeleton/Skeleton.svelte
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will extract a shared util in blade-core in a follow-up once two more components hit the same pattern.

packages/blade-svelte/src/components/Skeleton/Skeleton.svelte
rohankokane-dev: Tracking — nit, out of scope for this must-fix pass.

packages/blade-svelte/src/components/Skeleton/Skeleton.stories.svelte
rohankokane-dev: Tracking — nit, out of scope for this must-fix pass.

saurabhdaware (dismissed):

saurabhdaware (approved):

PR #3363, opened 2026-04-30
feat(blade-svelte): Switch component

rohankokane-dev:
## Description

Migrates the React Switch component to Svelte 5.

## Changes

- Adds `packages/blade-svelte/src/components/Switch/`
- Adds `packages/blade-core/src/styles/Switch/`
- Re-exports Switch from `packages/blade-svelte/src/components/index.ts`

## Additional Information

Artifacts (in worktree):

- Discovery report: `.cursor/artifacts/Switch/discovery-report.md`
- Migration plan: `.cursor/artifacts/Switch/migration-plan.md`
- Verification report: `.cursor/artifacts/Switch/verification-report.md`
- Screenshots: `.cursor/artifacts/Switch/screenshots/`

## Component Checklist

- [ ] Add KitchenSink Story
- [x] Add changeset


packages/blade-svelte/src/components/Switch/Switch.svelte
```diff
@@ -0,0 +1,200 @@
...
+      bind:this={inputEl}
+      class={switchClasses.input}
+      type="checkbox"
+      {id}
+      {name}
+      {value}
+      checked={isCheckedState}
+      disabled={isDisabled}
+      onchange={handleChange}
+      {...a11yAttrs}
```
rohankokane-dev, line 193: **Potential a11y conflict: duplicate `aria-disabled` and `disabled` on the input**

The `a11yAttrs` spread includes `aria-disabled` (from `makeAccessible({ disabled: isDisabled })`), while the input also has the native `disabled={isDisabled}` attribute. This is technically fine for a11y -- the React version does the same -- but note that when `isDisabled` is `false`, this renders `aria-disabled="false"` explicitly on the element. The React snapshot shows this is expected parity, so this is consistent. Just flagging for awareness.

packages/blade-svelte/src/components/Switch/Switch.svelte
```diff
@@ -0,0 +1,200 @@
...
+    ...rest
+  }: SwitchProps = $props();
+
+  let internalChecked = $state(untrack(() => defaultChecked) ?? false);
+
+  const isCheckedState = $derived(
+    isChecked !== undefined ? isChecked : internalChecked,
+  );
+
+  const isControlled = $derived(isChecked !== undefined);
```
rohankokane-dev, line 38: **`isControlled` should be a one-time check, not a `$derived`**

`isControlled` is currently `$derived(isChecked !== undefined)`, which means it re-evaluates every time `isChecked` changes. If a consumer passes `isChecked={undefined}` after previously passing a value (or vice versa), the component will silently switch between controlled/uncontrolled modes mid-lifecycle. This is the same pattern React warns about ('A component is changing an uncontrolled input to be controlled').

Consider computing this once on mount using `untrack`:
```ts
const isControlled = untrack(() => isChecked !== undefined);
```

This matches React's behavior where the controlled/uncontrolled decision is made on initial render. In practice this is unlikely to cause bugs since Blade consumers rarely flip between modes, but it's a correctness improvement worth making.

packages/blade-svelte/src/components/Switch/Switch.svelte
```diff
@@ -0,0 +1,200 @@
...
+  const trackA11yAttrs = makeAccessible({ hidden: true });
+</script>
+
+<div class={wrapperClasses} style={wrapperStyles} {...metaAttrs} {...analyticsAttrs}>
+  <label
+    class={switchClasses.label}
+    data-disabled={isDisabled || undefined}
+    onmousedown={handlePointerPressedIn}
+    onmouseup={handlePointerPressedOut}
+    onmouseout={handlePointerPressedOut}
```
rohankokane-dev, line 176: **`onmouseout` vs `onmouseleave` for pressed state reset**

Using `onmouseout` here will fire when the cursor moves over child elements within the label (it bubbles). This could cause the pressed state to flicker if the label contained additional child elements. While the current structure is simple enough that this is unlikely to cause visible issues, `onmouseleave` would be more correct since it only fires when the pointer actually leaves the element boundary, not when it moves between children.

The React version uses `onMouseOut` as well (line 101 of Switch.tsx), so this is parity. But it might be worth switching both versions to `mouseleave` eventually.

packages/blade-svelte/src/components/Switch/Switch.svelte
```diff
@@ -0,0 +1,200 @@
...
+    {...labelMetaAttrs}
+  >
+    <input
+      bind:this={inputEl}
+      class={switchClasses.input}
+      type="checkbox"
+      {id}
+      {name}
+      {value}
+      checked={isCheckedState}
```
rohankokane-dev, line 190: **`checked` attribute will fight with Svelte's reactivity on the input**

In Svelte, setting `checked={isCheckedState}` as a prop on a native `<input type="checkbox">` only sets the initial value -- Svelte doesn't forcefully re-sync the `checked` DOM property on every render the way React does. Combined with the `onchange` handler that computes `next = !isCheckedState`, this works correctly for uncontrolled mode. But for controlled mode (where the parent supplies `isChecked`), if the parent chooses NOT to update `isChecked` in its `onChange` handler (e.g., to reject a toggle), the DOM checkbox will still visually toggle because the browser changed the checked state on click, and Svelte won't re-set it.

This is the classic Svelte controlled input problem. In Svelte 5, you may want to explicitly force the checked property in the `handleChange` handler:
```ts
function handleChange(event: Event): void {
  if (isDisabled) return;
  const next = !isCheckedState;
  if (!isControlled) {
    internalChecked = next;
  }
  onChange?.({ isChecked: next, value, event });
  // Force re-sync DOM for controlled mode where parent might reject the change
  if (isControlled && inputEl) {
    inputEl.checked = isChecked ?? false;
  }
}
```

Or use a `$effect` to sync. This is only relevant when a consumer wants to 'reject' a toggle in controlled mode, which is uncommon but possible.

packages/blade-core/src/styles/Switch/switch.module.css
```diff
@@ -0,0 +1,222 @@
...
+  width: var(--spacing-4);
+  height: var(--spacing-4);
+}
+
+.thumb.size-medium {
+  width: var(--spacing-5);
+  height: var(--spacing-5);
+}
+
+/*
```
rohankokane-dev, line 110: **Empty CSS rulesets rely on CSS Modules implementation detail**

These intentionally-empty rulesets (`.pressed {}` and `.effective-checked {}`) are anchors to ensure CSS Modules emits scoped class names. This is a clever workaround, but it's fragile -- some CSS minifiers or future CSS Modules implementations may strip empty rulesets. Consider adding a comment-only content or a no-op declaration (`display: inherit;` or similar) to make them minifier-safe. Alternatively, a build-time check or test that verifies these classes exist in the output would guard against silent breakage.

packages/blade-svelte/src/components/Switch/types.ts
```diff
@@ -0,0 +1,74 @@
+import type { StyledPropsBlade } from '@razorpay/blade-core/utils';
+
+/**
+ * Payload passed to the `onChange` callback when the switch toggles.
+ */
+export type SwitchOnChange = (event: { isChecked: boolean; value?: string; event?: Event }) => void;
+
```
rohankokane-dev, line 7: **`SwitchOnChange` event payload: `event` should not be optional in the type**

The type declares `event?: Event` (optional), but the component always passes it:
```ts
onChange?.({ isChecked: next, value, event });
```

The `event` parameter in the callback will always be defined when called from the component. Making it optional in the type misleads consumers into thinking it might be `undefined`. Consider changing to `event: Event` (required) in the type definition, or at minimum documenting why it's optional (e.g., for programmatic toggling scenarios that don't originate from DOM events).

rohankokane-dev (commented):
## Switch Svelte 5 Migration Review

### Overall Assessment

Solid migration. The component follows established blade-svelte patterns well, with proper use of `$props()`, `$state()`, `$derived()`, and `untrack()`. The CSS-in-Module + CVA architecture is well-structured and reusable. The stories are comprehensive with good coverage of controlled/uncontrolled, sizes, disabled states, and the label placement patterns from the Figma guidelines.

### What Looks Good

- **Accessibility fundamentals are correct**: `role="switch"` on the input, `aria-label` via `accessibilityLabel` (required prop -- good), `aria-checked` synced to state, focus-visible ring via CSS adjacent sibling selector, track marked `aria-hidden="true"`, SVG checkmark marked `aria-hidden="true"`.
- **Svelte 5 idioms**: Proper use of `$props()` with destructuring and defaults, `$derived()` for computed values, `$state()` for local state, `untrack()` to prevent re-reading `defaultChecked` after mount.
- **CSS architecture**: Clean separation between blade-core styles (CSS Modules + CVA) and the Svelte component. Design token usage is consistent. Mobile breakpoint overrides at 767px match the React version's breakpoint logic.
- **`getSwitchTemplateClasses()` pattern**: Smart solution to prevent tree-shaking of CSS Module classes that are only referenced in templates.
- **Imperative API**: `export function focus()` with `bind:this` support via `SwitchInstance` type -- clean parity with React's `forwardRef`.

### Issues Raised (inline comments)

1. **[Medium] Controlled mode DOM sync** (line 175): Svelte does not force-re-render checked properties like React does. If a parent rejects a toggle in controlled mode, the DOM checkbox will visually desync. Add an `$effect` or manual sync.

2. **[Low] `isControlled` as `$derived`** (line 38): Should be computed once on mount to prevent mid-lifecycle mode switching.

3. **[Low] `onmouseout` vs `onmouseleave`** (line 161): `mouseout` bubbles through children, `mouseleave` doesn't. Parity with React, but `mouseleave` would be more correct.

4. **[Info] `SwitchOnChange` event optionality** (types.ts line 7): The `event` field is typed as optional but always provided.

5. **[Info] Empty CSS rulesets** (switch.module.css line 110): `.pressed {}` and `.effective-checked {}` anchors could be stripped by aggressive minifiers.

6. **[Info] `aria-disabled` + `disabled` duplication** (line 178): Matches React parity, just noting for awareness.

### Missing Items per Checklist

The PR checklist shows several unchecked items:
- **Component Status Page update**: Not included in this PR.
- **Manual testing in other browsers**: Not confirmed.
- **KitchenSink Story**: Not present (React version has `_KitchenSink.Switch.stories.tsx`).
- **Interaction tests**: None included. Given this is an a11y-sensitive component (keyboard navigation, screen reader state), interaction tests would be valuable.

### Nits (not worth inline comments)

- The `wrapperStyles` derived could use `undefined` as the fallback more cleanly with a ternary rather than `|| undefined`.
- Stories file has some inline styles that could potentially use design tokens more consistently (e.g., `gap: var(--spacing-2)` vs hardcoded values).
- The `capitalize` helper in stories could be imported from a shared util if one exists.

### Verdict

The core implementation is correct and follows Blade patterns well. The controlled-mode DOM sync issue (point 1) is the only item I'd want addressed before merge -- the rest are improvements that can be tackled incrementally.

packages/blade-svelte/src/components/Switch/Switch.svelte
rohankokane-dev: Acknowledged — explicit React parity per the snapshot. No action.

packages/blade-svelte/src/components/Switch/Switch.svelte
rohankokane-dev: Tracking — out of scope for this must-fix pass. The mid-lifecycle uncontrolled ↔ controlled flip is the same anti-pattern React warns about at runtime; will harden in a follow-up.

packages/blade-svelte/src/components/Switch/Switch.svelte
rohankokane-dev: Tracking — out of scope for this must-fix pass. The label has no children that would re-fire `mouseout` today, but `mouseleave` is more correct. Will switch in a follow-up.

packages/blade-svelte/src/components/Switch/Switch.svelte
rohankokane-dev: Fixed in 6146787f4. Re-asserting `inputEl.checked = isCheckedState` after `onChange` returns handles the controlled-rejection case where the parent doesn't update `isChecked` and Svelte's attribute binding doesn't re-sync the DOM.

packages/blade-core/src/styles/Switch/switch.module.css
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will replace the empty anchors with `getSwitchTemplateClasses()` references (the pattern used elsewhere in the file) in a follow-up so we don't depend on minifier behavior.

packages/blade-svelte/src/components/Switch/types.ts
rohankokane-dev: Tracking — out of scope for this must-fix pass. Will tighten `event` to required in the type in a follow-up.

saurabhdaware (dismissed):

saurabhdaware (approved):

PR #3370, opened 2026-05-04
feat(native): flesh out DatePicker, BarChart, and CommonChartComponents

pranaybagrecharzp:
## Description

Builds out the native (React Native) implementations of components that previously shipped only as scaffolds. Driven by the Razorpay merchant mobile app's Blade v12 migration spike, where the prior native stubs were insufficient for the home/payments screens to render correctly.

## Changes

- **`DatePicker/DatePicker.native.tsx`** — Tier-1 native date picker. Full input + calendar wiring built out from the prior scaffold (~530 LOC).
- **`Charts/BarChart/BarChart.native.tsx`** — Tier-1 native bar chart implementation, previously a stub (~470 LOC).
- **`Charts/CommonChartComponents/CommonChartComponents.native.tsx`** + **`index.ts`** + **`tokens.ts`** — shared native helpers and tokens used across the chart components.
- **`utils/storybook/componentStatusData.ts`** — bumps DatePicker / BarChart status rows to reflect the new native coverage.

Web behaviour is unchanged. Each touched file's `.web.tsx` counterpart is untouched.

## Additional Information

- This consolidates work that was originally split across two scratch branches (`feat/datepicker-rn`, `feat/linechart-rn`) plus several rounds of in-progress edits in a sibling clone — the latest, working state of each file lives here.
- Followed by `12.101.2` (the rollup tree-shake fix) — once this lands, the merchant-app team will bump again to pick up the Tier-1 implementations.

## Component Checklist

- [ ] Update Component Status Page — done in this PR (`componentStatusData.ts` row updates)
- [ ] Perform Manual Testing in Other Browsers — n/a (native-only changes; web entry untouched)
- [ ] Add KitchenSink Story — exists for both DatePicker and BarChart
- [ ] Add Interaction Tests (if applicable) — happy to add as follow-up if needed
- [x] Add changeset


packages/blade/src/components/DatePicker/DatePicker.native.tsx
```diff
@@ -1,13 +1,531 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
```
swapnil-kr1: File-level `eslint-disable @typescript-eslint/no-explicit-any` is hiding several real type holes below (lines 283, 288, 337, 354, 443). The public component is declared generic over `Type extends DateSelectionType`, but the body re-casts `props as any` (L283) and `onChange as any` (L288), which means the generic is doing nothing for consumers and any prop typo silently compiles.

Could we:
1. Drop the file-level disable.
2. Type `_DatePicker` properly against `DatePickerProps<Type>` so `value`, `defaultValue`, `onChange`, and `onApply` flow through without `as any`.
3. Use a `Preset` import from `./types` instead of redeclaring `{ label: string; value: (date: Date) => any }` on L354/L443 — the existing `Preset` type already says `value: (date: Date) => DatesRangeValue`.

If there's a specific Mantine-shaped collision making the generic unusable, let's narrow it (a single `as unknown as DatesRangeValue` cast at the apply boundary) rather than disabling the rule for the whole file.

packages/blade/src/components/DatePicker/DatePicker.native.tsx
```diff
@@ -1,13 +1,531 @@
...
+      row = [];
+      if (cursor.isAfter(end, 'day')) break;
+    }
+    cursor = cursor.add(1, 'day');
+  }
+  return cells;
+};
+
+const dayHeaders = (firstDayOfWeek: number): string[] => {
+  const labels = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'];
```
swapnil-kr1: Hardcoded English day-of-week labels. The web `DatePicker` localises via Mantine + `@react-aria/i18n`, and Blade ships `@razorpay/i18nify-react` as a peer dep — merchants in IN/SEA already depend on locale-correct weekday names (e.g. Hindi `रवि/सोम/...`).

At minimum: derive these from `dayjs().localeData().weekdaysShort()` (we already use `dayjs`) so the labels follow whatever locale the host app has set. Same applies to the month title on L146 (`'MMMM YYYY'`) — that one is already fine through dayjs, but please double-check by running the calendar with `dayjs.locale('hi')`.

packages/blade/src/components/DatePicker/DatePicker.native.tsx
```diff
@@ -1,13 +1,531 @@
...
+        paddingY="spacing.3"
+      >
+        <Pressable accessibilityRole="button" onPress={goPrev} hitSlop={8}>
+          <Box
+            paddingX="spacing.3"
+            paddingY="spacing.2"
+            borderRadius="medium"
+            backgroundColor="surface.background.gray.subtle"
+          >
+            <Text size="medium">‹</Text>
```
swapnil-kr1: Using unicode glyphs `‹` / `›` (L170, L183) and `✕` (L409) as chrome. These render inconsistently across iOS/Android system fonts (Android in particular has poor coverage for the single-angle quotation marks) and are unreachable to screen readers as icons.

Please use the existing Blade icon components: `ChevronLeftIcon` / `ChevronRightIcon` for the month navigation and `CloseIcon` for the clear button — they're already available via `~components/Icons` and the rest of the Toast.native.tsx in this same PR uses them correctly.

packages/blade/src/components/DatePicker/DatePicker.native.tsx
```diff
@@ -1,13 +1,531 @@
...
+                ✕
+              </Text>
+            </Pressable>
+          ) : null}
+        </Box>
+      </Pressable>
+
+      <BottomSheet isOpen={openInternal} onDismiss={close}>
+        <BottomSheetHeader
+          title={selectionType === 'range' ? 'Select Date Range' : 'Select Date'}
```
swapnil-kr1: Hardcoded English UI strings here (`'Select Date Range'` / `'Select Date'`), L440 (`'Quick Select'`), L488 (`'Cancel'`), L493 (`'Apply'`), L323 (`'Select date range'` / `'Select date'` as placeholder). None of these are localised.

For strings the consumer can't override, we should source them through `@razorpay/i18nify-react`'s formatter or accept them as props with documented defaults. The latter is probably the smaller change for this PR — see how `BottomSheet`/`Drawer` headers take a `title` prop instead of hardcoding.

packages/blade/src/components/DatePicker/DatePicker.native.tsx
```diff
@@ -1,13 +1,531 @@
...
+ * DatePicker
+ *
+ * Lets the user pick a single date or a range of dates. On native, the picker
+ * renders a tap-to-open trigger; the calendar opens inside a `BottomSheet`.
+ *
+ * Native MVP supports: `single` and `range` selection, presets, min/max date,
+ * controlled/uncontrolled value, controlled/uncontrolled open state, custom
+ * footer slot, clear button, and the `compact`/`default` displayFormat. Month
+ * and year picker variants, locale customisation, and the masked text input
+ * are not yet implemented on native.
```
swapnil-kr1: This docblock notes that `picker`/`defaultPicker`/`onPickerChange`, locale customisation, masked text input, and the form props (`validationState`, `helpText`, `errorText`, `isRequired`, `necessityIndicator`, `labelPosition`) are not yet implemented on native. Two requests:

1. **Runtime feedback.** If a consumer passes `picker="month"` today, the prop is silently dropped — they'll think it's their bug. Please add a `__DEV__` warning via `logger({ type: 'warn', moduleName: 'DatePicker', message: '…not yet supported on native' })` for each unsupported prop received.
2. **Platform parity in the type system.** Either narrow the native types to omit unsupported props (so TS catches it at compile time), or leave the types broad but explicitly call out in JSDoc which props are no-ops on native. Today the type surface is identical to web but the runtime is a strict subset — that's the worst-of-both-worlds.

Also: the BottomSheet-based UX deviates substantially from web's popover. Worth a one-line note in this docblock so consumers aren't surprised.

packages/blade/src/components/DatePicker/DatePicker.native.tsx
```diff
@@ -1,13 +1,531 @@
...
+      <Box>
+        {grid.map((row, rowIdx) => (
+          <View
+            key={`row-${rowIdx}`}
+            style={{ flexDirection: 'row', justifyContent: 'space-around' }}
+          >
+            {row.map((day) => {
+              const sameMonth = dayjs(day).isSame(dayjs(visibleMonth), 'month');
+              const disabled = isDisabled(day);
+              const selectedStart = sameDay(day, start);
```
swapnil-kr1, line 248: `accessibilityState` is not set on the day cell `<Pressable>`. Screen readers (TalkBack/VoiceOver) won't announce "selected" / "disabled" / today's date.

Please add `accessibilityState={{ selected: isSelected, disabled }}` on the Pressable (L218), and consider `accessibilityLabel` enriched with selection state — e.g. `"15 May 2026, selected start of range"`. This is the kind of thing that bites us during the merchant-app a11y audit later.

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
```diff
@@ -1,25 +1,454 @@
...
+ * reads its props (`dataKey`, `name`, `color`, `stackId`) and renders the bars.
+ */
+const ChartBar = assignWithoutSideEffects(_ChartBar, {
+  componentId: componentIds.chartBar,
+});
+
+const ChartBarWrapper: React.FC<ChartBarWrapperProps & TestID & DataAnalyticsAttribute> = ({
+  children,
+  colorTheme = 'categorical',
+  layout = 'horizontal',
```
swapnil-kr1, line 183: `layout = 'horizontal'` is destructured and put on `BarChartContext`, but the SVG rendering below (L297–350) only implements horizontal bars — there's no branch for `layout === 'vertical'`. Consumers passing `<ChartBarWrapper layout="vertical">` will silently get a horizontal chart.

Either implement the vertical branch (parity with web) or, if that's out of scope for this PR, drop `layout` from the native prop type and add a `__DEV__` warning so consumers aren't surprised. Same comment will apply to `colorIntensity` if/when web grows it.

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
```diff
@@ -1,25 +1,454 @@
...
+    } else if (id === commonComponentIds.chartLegend) {
+      slots.hasLegend = true;
+    } else if (id === commonComponentIds.chartCartesianGrid) {
+      slots.hasGrid = true;
+    }
   });
 
-  return <Text>ChartBar is not available for Native mobile apps.</Text>;
+  return slots;
+};
```
swapnil-kr1, line 148: `readChildSlots` only recognises `chartBar`, `chartXAxis`, `chartYAxis`, `chartLegend`, `chartCartesianGrid`. The web BarChart also accepts `<ChartTooltip>` and `<ChartReferenceLine>` (we export them right here in this PR in `CommonChartComponents.native.tsx`), and they'll silently render to nothing on native.

Given this is a fresh native build, two options:
1. Implement at least `ChartReferenceLine` (it's a single SVG `<Line>` and a label — small, high-value); leave tooltip as a documented gap with a DEV warning.
2. Or, at minimum, log a one-shot DEV warning the first time we encounter an unsupported child component id, so RN consumers don't burn a debug cycle wondering why their `<ChartTooltip>` isn't showing up.

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
```diff
@@ -1,25 +1,454 @@
...
+    if (!data.length) return 0;
+    let max = 0;
+    data.forEach((row) => {
+      stackGroups.forEach((group) => {
+        const sum = group.bars.reduce((acc, bar) => acc + (Number(row[bar.dataKey]) || 0), 0);
+        if (sum > max) max = sum;
+      });
+    });
+    return max;
+  }, [data, stackGroups]);
```
swapnil-kr1, line 270: Empty-data path: when `data` is `[]`, `dataMax` is 0, `niceCeil(0) === 1`, and we render an SVG with only axes and no message. Web's BarChart has an empty state — for native parity, please render the standard Blade empty-state Text ("No data to display" or similar, sourceable from `surface.text.gray.muted`) inside the chart area when `data.length === 0`.

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
```diff
@@ -1,25 +1,454 @@
...
+import { getComponentId } from '~utils/isValidAllowedChildren';
+import type { TestID, DataAnalyticsAttribute } from '~utils/types';
+import { metaAttribute } from '~utils/metaAttribute';
+import { makeAnalyticsAttribute } from '~utils/makeAnalyticsAttribute';
 
-const ChartBar = (_prop: ChartBarProps): React.ReactElement => {
-  throwBladeError({
-    message: 'ChartBar is not yet implemented for native',
-    moduleName: 'ChartBar',
+const PLOT_PADDING = { top: 16, right: 16, bottom: 32, left: 40 };
```
swapnil-kr1, line 41: All padding / sizing magic numbers (`PLOT_PADDING`, `LEGEND_HEIGHT`, `TICK_FONT_SIZE`, `AXIS_LABEL_FONT_SIZE`, `SECONDARY_LABEL_FONT_SIZE`) are hardcoded pixels. Two issues:

1. They don't respect `theme.spacing` / `theme.typography` — they'll drift if we ever rebrand or scale typography. Use `theme.spacing[3]`, `theme.spacing[4]` etc., and source font sizes from `theme.typography.fonts.size`.
2. `SECONDARY_LABEL_FONT_SIZE = 9` is below the 11pt minimum we keep for any user-readable text in Blade — even axis ticks. Please bump to at least 10 (and verify the secondary-label use case is still legible on a 360dp-wide phone).

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
```diff
@@ -1,25 +1,454 @@
...
+                    <SvgText
+                      x={plotWidth / 2}
+                      y={plotHeight + padding.bottom - 4}
+                      fontSize={AXIS_LABEL_FONT_SIZE}
+                      fill={tickColor}
+                      textAnchor="middle"
+                    >
+                      {slots.xLabel}
+                    </SvgText>
+                  ) : null}
```
swapnil-kr1, line 620: X-axis title rendering at `y={plotHeight + padding.bottom - 4}` collides with the secondary tick labels (L386) when both are present — `plotHeight + TICK_FONT_SIZE + 4 + SECONDARY_LABEL_FONT_SIZE + 2 = plotHeight + 25` vs `plotHeight + padding.bottom - 4 = plotHeight + 28`. With only 3px between them, anti-aliased text bleeds together.

Either expand `PLOT_PADDING.bottom` when `slots.xLabel && slots.xSecondaryDataKey` are both set, or compute `bottom` dynamically from what we're about to render.

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
```diff
@@ -1,25 +1,454 @@
...
+import React, { useMemo, useState } from 'react';
+import { Svg, Rect, Line, Text as SvgText, G } from 'react-native-svg';
+import { View } from 'react-native';
+import type { LayoutChangeEvent } from 'react-native';
+import { BarChartContext } from './BarChartContext';
+import { componentIds, BAR_CHART_CORNER_RADIUS } from './tokens';
 import type { ChartBarProps, ChartBarWrapperProps } from './types';
-import { Text } from '~components/Typography';
-import { throwBladeError } from '~utils/logger';
+import { useChartsColorTheme, assignDataColorMapping, getHighestColorInRange } from '../utils';
```
swapnil-kr1: `isNumber` is imported (L15) and only used on L323 to guard `barHeight <= 0` — but `Number(row[bar.dataKey]) || 0` on L304 already coerces non-numeric values to 0. The `isNumber(value)` check is dead.

Minor, but flagging because removing it lets you drop the import.

packages/blade/src/components/Charts/CommonChartComponents/CommonChartComponents.native.tsx
```diff
@@ -7,62 +7,45 @@ import type {
...
+const ChartCartesianGrid = assignWithoutSideEffects(_ChartCartesianGrid, {
+  componentId: componentId.chartCartesianGrid,
+});
 
-  return <Text>ChartLegend is not available for Native mobile apps.</Text>;
-};
+const _ChartTooltip = (_props: ChartTooltipProps): React.ReactElement | null => null;
+const ChartTooltip = assignWithoutSideEffects(_ChartTooltip, {
+  componentId: componentId.chartTooltip,
+});
```
swapnil-kr1, line 42: `ChartTooltip` and `ChartReferenceLine` are exported here but neither is read by `BarChart.native.tsx`'s `readChildSlots` — they're effectively dead code on the native target.

Two cleanups:
1. Add a `__DEV__` `logger({ type: 'warn', ... })` inside `_ChartTooltip` and `_ChartReferenceLine` so consumers who declaratively pass them get told they're no-ops. We do this elsewhere in the codebase (`throwBladeError` in `~utils/logger`).
2. If we genuinely don't support these yet, please remove them from the public native export until the chart wrappers actually consume them — better to fail-fast at import than render-and-do-nothing.

packages/blade/src/components/Toast/Toast.native.tsx
```diff
@@ -1,17 +1,147 @@
...
+  }, [isVisible, opacity, translateY]);
+
+  const isPromotional = type === 'promotional';
+  const Icon = Leading ?? iconMap[color];
+
+  const handleDismiss = (): void => {
+    onDismissButtonClick?.({
+      // Native has no MouseEvent; surface a minimal stand-in for type compat.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      event: {} as any,
```
swapnil-kr1: `event: {} as any` is shipping a lie to the callback contract — consumers reading `event.target` will crash at runtime. Two cleaner options:

1. Re-type `onDismissButtonClick` / `action.onClick` on the native side so the `event` field is optional (`event?: GestureResponderEvent`) and pass the real `GestureResponderEvent` from `Pressable`'s `onPress`.
2. If we want a single cross-platform type, make `event` `MouseEvent | GestureResponderEvent` and document the platform-conditional behaviour.

Option 1 is the cheap one and the eslint-disable comment goes away.

packages/blade/src/components/Toast/useToast.native.tsx
```diff
@@ -0,0 +1,117 @@
...
+      clearTimeout(timer);
+      this.timers.delete(id);
+    }
+    this.toasts = this.toasts.map((t) => (t.id === id ? { ...t, visible: false } : t));
+    this.notify();
+    // Remove from list after exit animation completes.
+    setTimeout(() => {
+      this.toasts = this.toasts.filter((t) => t.id !== id);
+      this.notify();
+    }, 250);
```
swapnil-kr1: Magic number `250` for the post-dismiss removal timeout (and there's no cleanup if the store is torn down mid-flight). Two suggestions:

1. Define `const EXIT_DURATION_MS = 200; const REMOVAL_GRACE_MS = 50; setTimeout(..., EXIT_DURATION_MS + REMOVAL_GRACE_MS)` (or import `EXIT_DURATION_MS` from `Toast.native.tsx`) so the magic number is named and lives in one place.
2. Track the removal timer in `this.timers` the same way you track the auto-dismiss timer, so a manual `dismiss(id)` followed quickly by a re-`show` with the same id can't accidentally remove the new toast.

Also: when `show()` returns `''` because we rejected a duplicate promotional toast (L50), callers using the return value as a tracking id can't distinguish rejection from a real toast id. Consider returning `null`/`undefined` and updating the union return type.

packages/blade/src/components/Toast/ToastContainer.native.tsx
```diff
@@ -1,12 +1,40 @@
...
+  );
 };
 
+const styles = StyleSheet.create({
+  container: {
+    position: 'absolute',
+    left: 16,
+    right: 16,
+    bottom: 32,
+  },
```
swapnil-kr1: Hardcoded positioning (`left: 16, right: 16, bottom: 32`) 

Please:
1. Source the 16/32 from `theme.spacing` so it scales with the design system.

packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -21,224 +19,379 @@ import { Divider } from '~components/Divider';
...
-};
-
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
 const getTabs = (node: React.ReactNode): React.ReactElement[] => {
   const children = React.Children.toArray(node);
   const tabList = children.find((child) => getComponentId(child) === 'TabList');
   if (!tabList || !React.isValidElement(tabList)) {
     throw new Error('TabList is required');
```
swapnil-kr1: `throw new Error('TabList is required')` — Blade convention is `throwBladeError({ message, moduleName: 'Tabs' })` from `~utils/logger`. That path strips the error in production builds (so consumers don't ship our message text to end-users) and tags the module for Sentry. Please switch.

packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -21,224 +19,379 @@ import { Divider } from '~components/Divider';
...
+        >
+          {panels.map((panel, index) => {
+            const shouldRender = !isLazy || index === safeIndex;
+            return (
+              <View key={panel.value} style={styles.pagerView}>
+                {shouldRender ? panel.children : null}
+              </View>
+            );
+          })}
+        </PagerView>
```
swapnil-kr1, line 426: `<PagerView>` always renders all panels (or only the selected one when `isLazy`), but it doesn't preserve panel state when switching back. Web `Tabs` keeps panel state mounted by default — please verify this matches web's contract, otherwise consumers will see (e.g.) form inputs reset on tab switch.

packages/blade/package.json
```diff
@@ -311,10 +310,9 @@
     "react-hot-toast": "2.4.1",
     "react-native": "^0.72",
     "react-native-gesture-handler": "^2.9.0",
-    "react-native-pager-view": "^6.2.1",
+    "react-native-pager-view": "^6.2.1 || ^8.0.0",
```
swapnil-kr1, line 313: `react-native-pager-view: "^6.2.1 || ^8.0.0"` — broad peer range that spans a major. pager-view@8 had a few API/native-module renames (some props moved, some events were renamed); the Tabs implementation here uses `setPage`, `initialPage`, `onPageSelected` and `scrollEnabled` which I believe are all stable across 6/7/8, but it would be good to:

1. Confirm by checking pager-view@8's CHANGELOG.
2. Add a runtime sanity check (or at least a comment here) noting which APIs we depend on.

Also: dropping `react-native-tab-view` from peerDeps is a peer-dep removal — for consumers on Blade 12.x who don't yet upgrade pager-view, please double-check the changeset says `patch` is correct (I'd argue this is `minor` because we're widening peer-dep range and dropping a peer).

packages/blade/src/utils/storybook/componentStatusData.ts
```diff
@@ -906,7 +906,7 @@ const componentData: ComponentStatusDataType = [
   {
     name: 'Toast',
     description: 'Toast is a component to show a simple floating messages to your users',
-    platform: 'web',
+    platform: 'all',
```
swapnil-kr1, line 909: Bumping `Toast` to `platform: 'all'` here but:

1. There's no native test for Toast (`__tests__/Toast.native.test.tsx` doesn't exist).
2. The Toast changes aren't called out in either changeset (`feat-native-datepicker-charts.md` only lists DatePicker/BarChart/CommonChartComponents; `fix-tabs-native-react19.md` is Tabs).

Please either add a Toast-specific changeset entry or amend the existing one, and add at minimum a snapshot test for `<Toast>` + an integration test for `toastStore.show/dismiss` deduping promotional toasts.

.changeset/fix-tabs-native-react19.md
```diff
@@ -0,0 +1,30 @@
+---
+'@razorpay/blade': patch
```
swapnil-kr1: `patch` may undersell this change. We're:

1. Removing `react-native-tab-view` from `peerDependencies`.
2. Widening `react-native-pager-view` to `^6.2.1 || ^8.0.0`.
3. Regenerating ~3000 lines of native snapshot.

Peer-dep changes are typically `minor` in our changeset policy (consumers on the old peer need to react). The visual diff for `filled` variant (now using `react-native-reanimated` rather than tab-view's indicator) may also be subtly different. Could you confirm with the release captain whether this should be `minor`?

packages/blade/src/components/DatePicker/DatePicker.native.tsx
pranaybagrecharzp: Fixed in the latest push. Dropped the file-level disable, narrowed the destructure to
`DatePickerNativeProps<DateSelectionType>` so the generic still flows to consumers,
and replaced the inline `{ label: string; value: (date: Date) => any }` with the
shared `Preset` type. The only remaining casts are narrow `as AnyValue` / `as ((value: AnyValue) => void) | undefined`
at the controlled-state and apply boundaries — those collapse the conditional types
to their union and don't escape into `any`.

Side-effect of dropping the disable also caught the +M2 issue I flagged below
(range-mode `innerValue=[null,null]` being truthy lit the clear button against
an empty selection — now gated through a proper `hasValue` memo).

packages/blade/src/components/DatePicker/DatePicker.native.tsx
pranaybagrecharzp: Done — `dayHeaders(firstDayOfWeek, localeName)` now goes through
`dayjs().locale(localeName).localeData().weekdaysShort()`. The `localeName` is read
via `useI18nContext()` from `@razorpay/i18nify-react` and converted with the
existing `convertIntlToDayjsLocale` helper (the same one `BaseDatePicker.web.tsx`
uses), so the native target picks up the same locale resolution as web.

Consumers still need to import the dayjs locale file at app boot
(`import 'dayjs/locale/hi'`) — RN doesn't support dynamic script loading like
web does, so I've documented that requirement in the DatePicker docblock and
fall back to whatever dayjs's default is when the locale isn't loaded (English
by default).

packages/blade/src/components/DatePicker/DatePicker.native.tsx
pranaybagrecharzp: Replaced all three. Calendar header now uses `<ChevronLeftIcon size="medium" color="surface.icon.gray.normal" />`
and `<ChevronRightIcon ... />`; the clear-icon button uses `<CloseIcon size="small" color="surface.icon.gray.muted" />`.
Also added `accessibilityLabel="Previous month"` / `"Next month"` on the chevron
`Pressable`s while I was there.

packages/blade/src/components/DatePicker/DatePicker.native.tsx
pranaybagrecharzp: Took the "accept as props with defaults" path since it composes cleanly with the
i18nify integration from comment #2. Added a `NativeLabelProps` mixin —
`headerLabel`, `applyLabel`, `cancelLabel`, `quickSelectLabel`,
`clearAccessibilityLabel` — all optional with English defaults. Consumers can
now pass strings from their i18n source (we'll wire the merchant-app's existing
i18nify-js strings through these in the consumer-side PR).

Kept `inputPlaceHolder` as the public escape hatch for the trigger placeholder
since it's already in the cross-platform type.

packages/blade/src/components/DatePicker/DatePicker.native.tsx
pranaybagrecharzp: Done. Day cell `<Pressable>` now gets:
- `accessibilityState={{ selected: isSelected, disabled }}`
- `accessibilityLabel="${baseLabel}${stateSuffix}"` where `stateSuffix` is one of
  `, selected start of range` / `, selected end of range` / `, selected` / `, in selected range` / empty.

TalkBack / VoiceOver now announce selection state properly.

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
pranaybagrecharzp: DEV warn + JSDoc for now. Added a one-shot `logger({ type: 'warn', moduleName: 'BarChart', ... })`
when `layout="vertical"` is passed (guarded by a `useRef` so it doesn't spam on
re-render). Kept the prop on the public type so cross-platform consumer code
still compiles — narrowing it native-only would force consumers to write
platform-conditional JSX which is worse ergonomics than a runtime warn.

Vertical layout is a substantial parallel render path (every coord flips
y↔x); home/payments dashboards in the merchant-app are horizontal-only, so
deferring vertical to a focused follow-up PR.

`colorIntensity` — I couldn't find it as a current `ChartBarWrapperProps`
field, so no action needed there.

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
pranaybagrecharzp: Both implemented in this push, not just the reference line:

- **`ChartReferenceLine`** — `readChildSlots` now reads `y` and `label`, renders a dashed
  `<Line stroke="surface.border.gray.normal" strokeDasharray="4 4">` with an
  end-anchored label above it. Horizontal lines (`y` is a number) only;
  vertical reference lines (`x`) emit a DEV warn pointing to the follow-up.

- **`ChartTooltip`** — turned out to be cleaner than expected. Wrapped the SVG
  in a `Pressable`, derived the active group index from the tap's `locationX`,
  and rendered an absolutely-positioned overlay near the tapped bar. The overlay
  shows the x-axis label + color-coded `<dot> name : value` rows for each visible
  bar, with the formatter prop honored (Recharts' `(value, name) => string | [value, name]`
  shape). Tap the same group again to dismiss; tap outside the plot region to dismiss.

Now removed the dead-code DEV warn for ChartTooltip in
`CommonChartComponents.native.tsx` since the marker is consumed.

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
pranaybagrecharzp: Done. When `data.length === 0`, the chart area renders a centered
`<Text color="surface.text.gray.muted">No data to display</Text>` inside a
spacing.6-padded Box and skips the SVG entirely (so axes don't draw for an
empty chart either). Matches the web pattern.

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
pranaybagrecharzp: Partial:

- **9px → 10px.** Bumped `SECONDARY_LABEL_FONT_SIZE` from 9 to 10. Light pushback on
  the "11pt minimum" framing — that's Blade's body-text rule; SVG axis tick
  labels are a different visual context (industry standard is 10–11px for
  primary ticks, ~9–10px for secondary). On a 360dp-wide phone the 10px
  secondary label is legible without crowding the primary tick.
- **Theme tokens for the magic numbers.** Added a comment block explaining why
  the SVG layout constants stay as raw pixels (`react-native-svg` works in
  pixel space, not Blade spacing tokens — and the values approximate the
  spacing scale: 16 ≈ spacing.5, 32 ≈ spacing.8). The legend/tick visuals
  are theme-derived (`tickColor`, `axisLineColor`, `gridColor`) via
  `getIn(theme.colors, ...)`. Constants that don't have a clean theme mapping
  (PLOT_PADDING dimensions, Y_TICK_COUNT) remain as named constants.

If we want full theme-token enforcement, the right move is a follow-up that
splits out a `barChartTokens.ts` like the existing `CommonChartComponents/tokens.ts`.

packages/blade/src/components/Charts/BarChart/BarChart.native.tsx
pranaybagrecharzp: Fixed with dynamic padding:

```ts
const needsExtraBottomPad = Boolean(slots.xLabel && slots.xSecondaryDataKey);
const padding = showAxes
  ? {
      ...PLOT_PADDING,
      bottom: needsExtraBottomPad
        ? PLOT_PADDING.bottom + SECONDARY_LABEL_FONT_SIZE + X_AXIS_TITLE_GAP
        : PLOT_PADDING.bottom,
    }
  : { top: 0, right: 0, bottom: 0, left: 0 };
```

`X_AXIS_TITLE_GAP = 6` gives a 6px buffer above the title plus the bumped 10px
secondary label, which clears the previous 3px crowd.

packages/blade/src/components/Charts/CommonChartComponents/CommonChartComponents.native.tsx
pranaybagrecharzp: Resolved by implementing both in `BarChart.native.tsx` (see #8). The marker
components are now actually consumed via `readChildSlots`, so the dead-code
warning isn't needed and the public exports stay as-is. Other chart wrappers
(Line, Area, etc.) are still stubs on native — when they get fleshed out, they
can pick up the same `readChildSlots` pattern.

packages/blade/src/components/Toast/Toast.native.tsx
pranaybagrecharzp: Fixed in two parts:

1. **Cross-platform type.** Found that `ToastProps.action.onClick`'s `event` field
   was typed as `ButtonProps['onClick']` — an existing typo (the type of the
   handler itself, not a MouseEvent). Web Toast was working around it with
   `event as never`. Corrected the type to
   `React.MouseEvent<HTMLButtonElement>` to match `onDismissButtonClick`.
   Toast.web.tsx still compiles; the `as never` cast on the web side becomes
   unnecessary in a follow-up cleanup.

2. **Native call sites.** Replaced both `{} as any` with a single typed
   stand-in:
   ```ts
   const stubEvent = ({} as unknown) as React.MouseEvent<HTMLButtonElement>;
   ```
   No more file-level eslint-disable on `no-explicit-any`. Couldn't pass a real
   `GestureResponderEvent` through because Blade's `Platform.Select`-branded
   `onClick` rejects a native-only function signature — we'd need to broaden
   the cross-platform type, which I scoped out for a follow-up. The stub is
   typed but `event` is empty (consumers reading native-specific fields get
   `undefined` instead of a runtime crash).

packages/blade/src/components/Toast/useToast.native.tsx
pranaybagrecharzp: Two of three accepted, one push-back:

1. ✅ **Named constants.**
   ```ts
   const EXIT_DURATION_MS = 200;
   const REMOVAL_GRACE_MS = 50;
   const REMOVAL_DELAY_MS = EXIT_DURATION_MS + REMOVAL_GRACE_MS;
   ```
   The 200ms half mirrors `EXIT_DURATION_MS` in `Toast.native.tsx`; the 50ms
   grace covers Animated.View's finish callback variance.

2. ✅ **Track the removal timer in `this.timers`.** Was a real race —
   `dismiss(id)` followed by an immediate `show(id)` could have the pending
   removal nuke the freshly-created toast. Now: `show()` checks for a pending
   timer keyed by the same id and clears it before pushing the new entry;
   `dismiss()` stores the removal timer back in `this.timers` so a subsequent
   `show()` can cancel it.

3. ✋ **Push back on `''` → `null`.** Four pieces of evidence:

   **a) Web already returns `''` for the exact same rejection case** —
   `useToast.web.tsx:36` (unchanged in this PR):
   ```ts
   if (toasts.find((t) => t.type === 'promotional') && props.type === 'promotional') {
     // ... logger.warn ...
     return '';
   }
   ```
   Native is mirroring the existing web behavior; changing it to `null`
   would force web to change too.

   **b) Upstream `react-hot-toast` returns `string`, not `string | null`** —
   `react-hot-toast/dist/index.d.ts`:
   ```ts
   type ToastHandler = (message: Message, options?: ToastOptions) => string;
   custom: ToastHandler;
   ```
   And `useToast.web.tsx:65` returns whatever `toast.custom()` returns →
   `string`. The contract is set by the upstream library Blade chose to
   wrap; diverging here breaks that.

   **c) Cross-platform `UseToastReturn.show` signature is identical** —
   `useToast.web.tsx:10` and `useToast.native.tsx:9` both declare
   `show: (props: ToastProps) => string`. Returning `null` from native
   forces either platform divergence (`show` returns different types per
   platform — bad for consumers writing platform-agnostic code) or a
   `string | null` migration on web too (out of scope for this PR).

   **d) No caller in Blade reads the return value** — every `toast.show(...)`
   site I found (`TextArea.stories.tsx:490`, `Toast.stories.tsx:43`,
   `FormGroup.stories.tsx`, etc.) discards it. The change would silently
   benefit no one.

   **Compromise that does both:** exported a typed sentinel constant
   `TOAST_REJECTED = ''` so callers who care can write
   `id === TOAST_REJECTED` instead of comparing string literals, without
   changing the public API contract. Documented in the store's JSDoc.

Plus a parity fix while I was in the file: `dismiss(undefined)` was dismissing
only the most-recent visible toast. `react-hot-toast.toast.dismiss()` dismisses
all visible toasts — native now matches.

packages/blade/src/components/Toast/ToastContainer.native.tsx
pranaybagrecharzp: Done. Switched from static `StyleSheet.create` to inline-via-`useTheme`:
```tsx
<View
  style={{
    position: 'absolute',
    left: theme.spacing[5],   // 16
    right: theme.spacing[5],  // 16
    bottom: theme.spacing[8], // 32
  }}
  pointerEvents="box-none"
>
```

`ToastContainer` renders once at the app root, so the per-render style cost
is negligible.

packages/blade/src/components/Tabs/Tabs.native.tsx
pranaybagrecharzp: Done. Replaced with:
```ts
throwBladeError({
  message: '<TabList> is required as a direct child of <Tabs>. Wrap <Tab> elements inside <TabList>.',
  moduleName: 'Tabs',
});
return []; // fallback for production where throwBladeError no-ops
```

Small correction on the rationale: `throwBladeError` (per `~utils/logger/logger.ts:16`)
is `__DEV__`-only — it strips entirely in production rather than tagging Sentry
directly. The `[Blade: Tabs]:` prefix helps Sentry grouping if Sentry is wired
up elsewhere, but no automatic integration. Either way, the convention swap is
correct and that's what matters.

Also dropped the three file-level eslint-disables (`no-explicit-any`,
`no-non-null-asserted-optional-chain`, `explicit-function-return-type`) and
properly typed `TabListChildProps` / `TabPanelChildProps` / `TabChildProps`
for the `child.props.*` accesses. The `React.ComponentType<any>` for the
polymorphic `leading` icon prop has a narrow inline disable since that's the
canonical Blade pattern.

packages/blade/src/components/Tabs/Tabs.native.tsx
pranaybagrecharzp: Verified — semantics match. Added an explicit JSDoc note in the code:

```
Panel state preservation: when `isLazy=false` (default), all panels render as
siblings inside PagerView — React preserves the mounted tree across tab
switches, so form inputs, scroll positions, etc. survive. When `isLazy=true`,
off-screen panels render `null` and unmount; state is lost on each switch.
This matches the documented lazy semantics on web.
```

PagerView (both v6 and v8) keeps panel children mounted as React siblings, so
the default mode preserves state. `isLazy=true` is the explicit opt-in to
unmount-on-blur — same contract as web. Also added a comment listing the
PagerView APIs we depend on (`setPage`, `initialPage`, `onPageSelected`,
`scrollEnabled`) and a note that they're stable across the v6 → v8 release line.

Fixed an unrelated +M4 issue while I was in the file: `handleTabPress` was
calling both `setIndex(index)` and `pagerRef.current?.setPage(index)`, with the
latter firing `onPageSelected` → `handlePageSelected` → `setIndex` again —
double state update on every tab. Now only `pagerRef.current?.setPage(index)`
fires; `onPageSelected` is the single source of truth.

packages/blade/package.json
pranaybagrecharzp: Added an inline comment listing the four APIs we consume — `setPage`,
`initialPage`, `onPageSelected`, `scrollEnabled` — and a note that they're
stable across the v6 → v8 release line per the pager-view CHANGELOG. Skipping
the runtime sanity check since it'd be over-engineering for a peer-dep that
the consumer already pinned; if any of those APIs change in a future major
we'd want to bump our peer range explicitly.

.changeset/fix-tabs-native-react19.md
pranaybagrecharzp: Bumped to `minor`, added an explanation block in the changeset:
```
> **Why minor (not patch):** this changeset removes react-native-tab-view from
> peerDependencies and broadens the react-native-pager-view peer range to
> ^6.2.1 || ^8.0.0. Both are peer-dep contract changes that consumers must
> react to, which by convention warrants a minor bump.
```

Worth confirming with the release captain before merge.

packages/blade/src/utils/storybook/componentStatusData.ts
pranaybagrecharzp: Done both:

1. **Changeset amended.** `feat-native-datepicker-charts.md` now explicitly
   lists Toast under what landed in this PR, including the
   `ToastProps.action.onClick` event-type correction (see #13).

2. **Tests added.**
   - `Toast/__tests__/Toast.native.test.tsx` — 4 unit tests against `toastStore`
     covering: id return shape, promotional dedup, `dismiss()` no-arg parity
     with web, and removal-timer cancellation.
   - `Charts/BarChart/__tests__/BarChart.native.test.tsx` — 3 tests: snapshot
     of horizontal grouped bars, empty-state copy, and reference-line render.
   - `DatePicker/__tests__/DatePicker.native.test.tsx` — module-load sanity
     check. Full render-and-interact tests are blocked on a known
     `@gorhom/bottom-sheet` test-runner limitation (needs a mock'd
     ScrollableComponent) — flagging that as a separate follow-up.

   Snapshot rendering of `<Toast>` itself is also blocked — Toast.native uses
   `popup.background.*` tokens which Box's validator rejects on native at
   runtime (`'Currently you can only use transparent, surface.background.*, and
   overlay.* tokens'`). That's a pre-existing issue, not introduced by this PR,
   but worth a follow-up to migrate the Toast background colors to validator-
   accepted tokens so the snapshot test can land.

packages/blade/src/components/DatePicker/DatePicker.native.tsx
pranaybagrecharzp: All three:

1. **DEV warns.** Added a `useEffect` that emits `logger({ type: 'warn', moduleName: 'DatePicker', ... })`
   when any of `picker` / `defaultPicker` / `onPickerChange` are passed. Form
   props (`validationState`, `helpText`, `errorText`, `successText`,
   `isRequired`, `necessityIndicator`, `labelPosition`) — those are now
   **actually supported** on native (see the trigger render path), so no warn
   for them. Master/year-picker variants and masked input remain on the
   roadmap; the warn explicitly names them.

2. **Type narrowing vs JSDoc.** Kept the cross-platform type surface so
   consumers can write platform-agnostic code. Documented every native-only
   gap in the docblock instead (Supported / Not yet supported / Localisation
   sections).

3. **BottomSheet vs popover UX note.** Added to the docblock — explicit call-out
   that tablets/large-screen devices see a full-width sheet instead of an
   anchored popover.

saurabhdaware (commented):
Looks good overall. I've triggered slash to fix lint issues

saurabhdaware (approved):

saurabhdaware:
We can merge it for now now to unblock. 

@shivam-rzpay Maybe few things for us to explore later when we come to these components-
1. We can explore and use some library for charts. Right now this works for BarChart since that is the only component we're adding but maybe it will make sense to just have library that helps us build all charts
2. Explore some libraries for DatePicker. Fairly complex components so libraries might handle things a lot smoother
3. Right now these components are very independent to web. We can see if some common code can be moved to single place and reused elsewhere 

Less priority for now. Can be picked when current P0 things are solved

saurabhdaware:
UT run is remaining in pending for some reason. Will merge and see if its some infra issue or some code issue

PR #3392, opened 2026-05-17
feat(blade): add SankeyChart component

debabratama1ik:
## Summary

Adds a `SankeyChart` component to the Blade design system for visualising flow distributions across stages (payment routing, funnel analysis, budget allocation, etc.).

This PR has been updated to address all engineering review feedback from Swapnil — replacing the original d3-sankey implementation with Recharts' native `<Sankey>` component.

---

## What changed (Swapnil's review → addressed)

| Point | Action |
|-------|--------|
| Use Recharts native `<Sankey>` instead of d3-sankey | ✅ Fully migrated; d3-sankey dep removed |
| Align with Blade chart pattern | ✅ `CommonChartComponentsContext`, `ResponsiveContainer`, Recharts `<Tooltip>` |
| Type color tokens properly | ✅ All color props use `ChartsCategoricalColorToken` |
| Flatten data shape (closer to Recharts format) | ✅ `{ nodes, links }` with string IDs; no more `levels[]` |

---

## API

```tsx
<SankeyChart
  data={{
    nodes: [
      { id: 'total',   name: 'Total' },
      { id: 'upi',     name: 'UPI' },
      { id: 'card',    name: 'Card' },
      { id: 'success', name: 'Successful' },
      { id: 'failed',  name: 'Failed' },
    ],
    links: [
      { source: 'total', target: 'upi',     value: 4000 },
      { source: 'total', target: 'card',    value: 3200 },
      { source: 'upi',   target: 'success', value: 3500 },
      { source: 'upi',   target: 'failed',  value: 500  },
      { source: 'card',  target: 'success', value: 2800 },
      { source: 'card',  target: 'failed',  value: 400  },
    ],
  }}
  labelUnit="txn"
  height={420}
  showTooltip
  showLabels
/>
```

### Props

| Prop | Type | Default | Description |
|------|------|---------|-------------|
| `data` | `{ nodes: SankeyDataNode[]; links: SankeyDataLink[] }` | required | Flat node list + directed links referenced by node id |
| `height` | `number` | `400` | SVG height in px |
| `nodeColorOverride` | `ChartsCategoricalColorToken` | — | Apply a single Blade token to all node bars |
| `linkColorOverride` | `ChartsCategoricalColorToken` | — | Apply a single Blade token to all link ribbons |
| `showTooltip` | `boolean` | `true` | Recharts-managed tooltip on hover |
| `showLabels` | `boolean` | `true` | Label chips to the right of each node |
| `labelUnit` | `string` | — | Unit appended to value in chip (e.g. `"txn"`, `"₹M"`) |
| `onNodeClick` | `(node, index) => void` | — | Fired when a node bar is clicked |
| `onLinkClick` | `(link, index) => void` | — | Fired when a link ribbon is clicked |

---

## Design / visual behaviour

- **Colour palette**: Blade categorical tokens, auto-rotated per node (same pool as other Blade charts)
- **Per-node colour**: set `color` on individual `SankeyDataNode` entries
- **Label chips**: Blade-themed rounded chip to the right of each node showing name + value
- **Hover highlight**: hovered node/link dims unrelated elements; smooth opacity transition
- **Tooltip**: Recharts-managed tooltip (auto edge-flip); Blade `staticBlack` background, `staticWhite` text
- **Mobile**: horizontal scroll wrapper at `min-width: 560px`; no layout collapse on narrow viewports

---

## Implementation notes

- **No d3-sankey**: layout is handled entirely by Recharts `<Sankey>`; d3-sankey removed from deps
- **`CommonChartComponentsContext`**: provides `chartName: 'sankey'` + `dataColorMapping` (DonutChart pattern)
- **`ResponsiveContainer`**: replaces the previous manual ResizeObserver; handles responsive sizing
- **Link ribbons**: custom `renderLink` draws filled Bezier `<path>` using Recharts-provided control points
- **Node bars**: custom `renderNode` draws `<rect>` + SVG label chip

---

## Files changed

| File | Change |
|------|--------|
| `SankeyChart/types.ts` | New flat `SankeyDataNode`, `SankeyDataLink`, `SankeyChartProps` with typed color tokens |
| `SankeyChart/SankeyChart.web.tsx` | Full rewrite using Recharts `<Sankey>` + Blade patterns |
| `SankeyChart/SankeyChart.native.tsx` | Unchanged — throws unsupported error |
| `SankeyChart/tokens.ts` | Unchanged |
| `SankeyChart/SankeyChart.stories.tsx` | Updated all stories to new data shape |
| `SankeyChart/__tests__/SankeyChart.web.test.tsx` | Updated fixtures, snapshots regenerated |
| `CommonChartComponents/types.ts` | Added `'sankey'` to `ChartName` union |
| `packages/blade/package.json` | Removed `d3-sankey` + `@types/d3-sankey` |
| `yarn.lock` | Regenerated without d3-sankey |

---

## Test plan

- [x] `SHARD='' yarn test:react --testPathPattern=SankeyChart --forceExit` → 21/21 pass
- [x] `npx eslint --max-warnings=0 src/components/Charts/SankeyChart/SankeyChart.web.tsx` → clean
- [ ] Storybook: Default, SingleColor, WithoutLabels, CustomNodeColors stories
- [ ] Hover/click interactions on node bars and link ribbons
- [ ] Mobile viewport (< 560px) → horizontal scroll works, no layout collapse
- [ ] Tooltip near chart edges → auto-flips, no clipping


rkdotxyz:
Screenshot 2026-05-18 at 4 32 25 PM
Screenshot 2026-05-18 at 4 32 37 PM

Let's not use these nodes as pill shaped, looks wierd and not as per design language.

rkdotxyz:
Screenshot 2026-05-18 at 4 33 44 PM

This also looks weird. The Spacing and corner radius are off.

rkdotxyz:
Screenshot 2026-05-18 at 4 34 27 PM

Why are we not using the chart Tooltip here?

rkdotxyz:
Screenshot 2026-05-18 at 4 36 22 PM

Why do we need a nodeWidth prop? Should it be something like a size prop?

rkdotxyz:
Screenshot 2026-05-18 at 4 37 55 PM

Why do we need a nodePadding prop?

rkdotxyz:
Screenshot 2026-05-18 at 4 46 04 PM

Component is breaking on mobile devices

packages/blade/src/components/Charts/SankeyChart/SankeyChart.tsx
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * In the Blade monorepo the bundler resolves `.web.tsx` / `.native.tsx`
+ * based on the platform target. This file is the platform-agnostic entry
+ * point — callers import from `'./SankeyChart'` and the correct platform
+ * implementation is selected automatically.
+ *
+ * Web:    SankeyChart.web.tsx  — d3-sankey layout + React SVG
+ * Native: SankeyChart.native.tsx — throwBladeError stub (not yet implemented)
+ */
+export { SankeyChart } from './SankeyChart.web';
```
rkdotxyz: **[Blocking] Platform entry file breaks native resolution**

This file explicitly re-exports from `.web.tsx`, defeating the platform-agnostic purpose. Every other chart component (LineChart, BarChart, AreaChart) has _no_ `ComponentName.tsx` — only `.web.tsx` and `.native.tsx`. The bundler resolves the correct variant because the plain `.tsx` entry doesn't exist.

With this file present, the native build may pick `SankeyChart.tsx` (which points to the web implementation) over `SankeyChart.native.tsx`, causing the native stub to never be reached.

**Fix:** Delete `SankeyChart.tsx`. The `index.ts` already exports from `'./SankeyChart'` and the bundler handles platform resolution — identical to how LineChart/BarChart/AreaChart are set up.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,448 @@
...
+                    {showLabels && (
+                      <g style={{ pointerEvents: 'none' }}>
+                        <rect
+                          x={chipX}
+                          y={chipY}
+                          width={chipW}
+                          height={CHIP_H}
+                          fill={chipBg}
+                          rx={chipRadius}
+                          style={{ filter: `drop-shadow(${chipShadow})` }}
```
rkdotxyz: **[Blocking] `elevation.midRaised` (box-shadow token) used as a CSS `drop-shadow` filter**

`castWebType(theme.elevation.midRaised)` returns a `box-shadow` string: `x y blur spread color`. The CSS `filter: drop-shadow()` function does **not** support a spread value — its syntax is `x y blur color` only.

The snapshot confirms the problem: `drop-shadow(0px 16px 12px 0px hsla(...))` — the `0px` spread is invalid here and browsers will silently discard the entire filter or ignore the spread, producing a different visual than intended.

**Fix:** Either wrap the chip `<rect>` in an HTML element that supports `box-shadow`, or construct the drop-shadow string manually from the shadow's x/y/blur/color components (without the spread).

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,448 @@
...
+  );
+
+  const tooltipBg = theme.colors.surface.icon.staticBlack.normal;
+  const tooltipTextColor = theme.colors.surface.text.staticWhite.normal;
+  const labelNameColor = theme.colors.surface.text.gray.normal;
+  const labelValueColor = theme.colors.surface.text.gray.muted;
+  const chipBg = theme.colors.surface.background.gray.subtle;
+  const chipShadow = castWebType(theme.elevation.midRaised);
+  const chipRadius = theme.border.radius.large;
+  const motionDuration = theme.motion.duration.gentle;
```
rkdotxyz: **Hover transitions at 480ms are too slow**

The PR description states this uses `duration.quick`, but the implementation uses `duration.gentle` (480ms). All hover-state transitions across links and nodes use this value, making interactions feel sluggish.

`duration.gentle` is typically used for layout/enter animations. Hover feedback should use `duration.quick` or `duration.moderate` — consistent with what was documented and with how other interactive chart elements behave.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,448 @@
...
+              left: tooltip.x,
+              top: tooltip.y,
+              background: tooltipBg,
+              color: tooltipTextColor,
+              borderRadius: theme.border.radius.medium,
+              fontSize: theme.typography.fonts.size[75],
+              padding: `${theme.spacing[2]}px ${theme.spacing[3]}px`,
+              pointerEvents: 'none',
+              whiteSpace: 'nowrap',
+              zIndex: 100,
```
rkdotxyz: **Hardcoded `zIndex: 100`**

This magic number should at minimum be a named constant in `tokens.ts` alongside the other opacity constants (`LINK_DEFAULT_OPACITY` etc.), so it's discoverable and changeable from one place. A comment explaining why 100 was chosen (e.g. "must sit above SVG stacking context") would also help future maintainers.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,448 @@
...
+                          style={{ userSelect: 'none', fontFamily }}
+                        >
+                          <tspan
+                            fontWeight={theme.typography.fonts.weight.semibold}
+                            fill={labelNameColor}
+                          >
+                            {node.name}
+                          </tspan>
+                          {labelValue && (
+                            <tspan fill={labelValueColor} dx={6}>
```
rkdotxyz: **Hardcoded `dx={6}` — not on the Blade spacing grid**

All other spacing in this component is derived from Blade tokens (`theme.spacing[3]` = 8px, `theme.spacing[2]` = 4px, etc.). `6` is an arbitrary value that sits off the 4px grid.

Use `theme.spacing[2]` (8px) or `theme.spacing[1]` (4px) instead.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,448 @@
...
+// CHIP_PAD_X → theme.spacing[4] = 12px  (horizontal text padding inside chip)
+// CHIP_GAP  → theme.spacing[3] = 8px   (gap between node bar and chip)
+
+// ─── Inner chart ──────────────────────────────────────────────────────────────
+
+function SankeyChartInner({
+  levels,
+  links,
+  height = 400,
+  nodeWidth = 20,
```
rkdotxyz: **Default prop values mismatch between docs and implementation**

The PR description's props table documents:
- `nodeWidth` default: `14`
- `nodePadding` default: `16`

But the actual function signature has `nodeWidth = 20` and `nodePadding = 10`. Please reconcile the docs with the code (or vice versa) before merge.

packages/blade/src/components/Charts/SankeyChart/tokens.ts
```diff
@@ -0,0 +1,15 @@
...
+ * via `useTheme()` at render time. Only stateless, non-theme constants live here.
+ */
+
+// ── Link opacity states ────────────────────────────────────────────────────────
+export const LINK_DEFAULT_OPACITY = 0.55;
+export const LINK_HOVER_OPACITY = 0.85;
+export const LINK_DIMMED_OPACITY = 0.1;
+
+// ── Node opacity states ───────────────────────────────────────────────────────
+export const NODE_DEFAULT_OPACITY = 1;
```
rkdotxyz, line 20: **`NODE_DEFAULT_OPACITY` is exported but never imported or used anywhere**

Dead export. Either remove it, or use it in `SankeyChart.web.tsx` in place of the hardcoded `1` in `getNodeOpacity` when there is no hover state.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,448 @@
...
+            {/* Links */}
+            <g>
+              {(graph.links as InternalLink[]).map((link, i) => {
+                const path = sankeyLinkHorizontal()(link as never) ?? '';
+                const srcNode = link.source as InternalNode;
+                const stroke = linkColor(flatNodes[srcNode.originalIndex], srcNode.originalIndex);
+
+                return (
+                  <path
+                    key={i}
```
rkdotxyz: **`key={i}` on SVG elements — use stable IDs instead**

Both link `<path>` elements and node `<g>` elements use the array index as key. Nodes have a stable `node.id` string; links can be keyed by `link.from + '-' + link.to`.

Index keys cause full DOM replacement for all subsequent elements on any reorder/update, losing in-progress CSS transition state. Stable keys also help React correctly reconcile opacity animations during hover.

packages/blade/src/components/Charts/SankeyChart/types.ts
```diff
@@ -0,0 +1,45 @@
...
+  links: SankeyFlowLink[];
+  height?: number;
+  nodeWidth?: number;
+  nodePadding?: number;
+  /** Override all nodes with a single Blade color token, e.g. 'interactive.background.primary.default' */
+  nodeColorOverride?: string;
+  /** Override all links with a single Blade color token, e.g. 'data.background.categorical.blue.subtle' */
+  linkColorOverride?: string;
+  showTooltip?: boolean;
+  showLabels?: boolean;
```
rkdotxyz, line 33: **`showLabels` prop description says "floating above" but implementation positions chips to the right**

The PR description states: _"Show label chips floating above each node."_

But the rendering code places chips at `chipX = x1 + CHIP_GAP` — horizontally to the right of the node bar, vertically centered. The description is inconsistent with the implementation. Please align the JSDoc and PR description with the actual behavior, or fix the position to match the Figma spec if "above" was the intended design.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.tsx
rkdotxyz: **[File shouldn't exist] Extra platform entry file not present in any other chart**

No other chart component has a plain `ComponentName.tsx` alongside `.web.tsx` and `.native.tsx`. LineChart, BarChart, AreaChart, and DonutChart all have only the two platform files — the bundler resolves the correct one automatically.

This file's comment claims it's a "platform-agnostic entry point", but it hardcodes the web implementation, which means native consumers that might resolve `SankeyChart.tsx` over `SankeyChart.native.tsx` will silently get the wrong platform build.

**Fix:** Delete this file entirely. `index.ts` already re-exports from `'./SankeyChart'`, and with only `.web.tsx` + `.native.tsx` present, Metro and webpack will pick the right one — exactly as they do for every other chart component.

packages/blade/src/components/Charts/_decisions/sankey-decisions.md
```diff
@@ -0,0 +1,132 @@
+# SankeyChart — API & Architecture Decisions
```
rkdotxyz, line 1: **[Wrong location] Per-component `_decisions/` folder doesn't match the established pattern**

No existing chart has its own `_decisions/` subfolder. The convention in this repo is a single shared decisions directory at the Charts level:

```
Charts/_decisions/
├── decisions.md        ← charts-wide decisions
├── alternativeApi1.md
├── alternativeApi2.md
└── requirement.md
```

Adding a `SankeyChart/_decisions/` introduces a new folder convention that isn't followed by any other chart, making the docs structure inconsistent.

**Fix:** Move this file to `packages/blade/src/components/Charts/_decisions/sankey-decisions.md` (or `sankey.md`). Update the self-referencing link inside the file accordingly.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.native.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * The web implementation relies on SVG, ResizeObserver, and d3-sankey —
+ * none of which are available in React Native's runtime. A native port
+ * would require a React Native SVG library (e.g. react-native-svg) and
+ * a separate layout pass. This is tracked as a future enhancement.
+ */
+import React from 'react';
+import type { SankeyChartProps } from './types';
+import { throwBladeError } from '~utils/logger';
+
+const SankeyChart = (_props: SankeyChartProps): React.ReactElement => {
```
rkdotxyz, line 13: **[Missing test] No `SankeyChart.native.test.tsx` in `__tests__/`**

The `__tests__/` folder only contains `SankeyChart.web.test.tsx`. The native stub here — a `throwBladeError` call — has no test coverage. If this stub is ever accidentally modified (e.g. the error removed or the module renamed), the failure would be silent.

BarChart is the reference: it has both `BarChart.web.test.tsx` and `BarChart.native.test.tsx`.

**Fix:** Add `__tests__/SankeyChart.native.test.tsx` with a test that verifies the error throws:

```tsx
import React from 'react';
import { SankeyChart } from '../SankeyChart.native';
import renderWithTheme from '~utils/testing/renderWithTheme.native';

describe('SankeyChart (native)', () => {
  it('throws a Blade error when rendered on native', () => {
    expect(() =>
      renderWithTheme(<SankeyChart levels={[]} links={[]} />)
    ).toThrow('SankeyChart is not yet implemented for native');
  });
});
```

packages/blade/src/components/Charts/SankeyChart/SankeyChartContext.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { createContext, useContext, useState } from 'react';
```
rkdotxyz: **[File shouldn't exist] Context module with no cross-component consumers**

This file exists to share `hoveredNodeId` / `hoveredLinkId` state — but the only consumer of this context is `SankeyChartInner` itself. There are no sibling or child sub-components that independently reach into it.

Compare with `LineChartContext.tsx`, which exists because recharts renders sub-components (axis, tooltip, legend) as separate tree nodes that need to coordinate. SankeyChart has no such sub-components — all rendering is in one function.

Context adds indirection and a small reconciliation cost for zero benefit here.

**Fix:** Delete this file. Move the two state declarations directly into `SankeyChartInner`, and delete the `SankeyChartProvider` wrapper:

```tsx
// Inside SankeyChartInner — replace useSankeyChartContext() with:
const [hoveredNodeId, setHoveredNodeId] = useState<number | null>(null);
const [hoveredLinkId, setHoveredLinkId] = useState<number | null>(null);
```

The outer `_SankeyChart` wrapper that currently renders `<SankeyChartProvider>` can then just render `<SankeyChartInner>` directly.

debabratama1ik:
Closing in favour of #3398 — moved branch from fork to org repo so CI secrets (bundle size check) work correctly.

PR #3394, opened 2026-05-18
fix(native): blade React Native component fixes and stubs

shivam-rzpay:
## Summary

- Fix styled-components v6 error in `BaseBox.native.tsx`
- Fix `makeBorderSize` string issue with a native-only implementation
- Fix reanimated crash in Popover and Tooltip content wrappers
- Fix badge spacing, chips styling, switch AnimatedThumb, and TextInput label on native
- Fix Tooltip native implementation
- Fix Popover on iOS, Android, and crash on re-render
- Add Avatar, AvatarGroup, and EmptyState native implementations
- Add Spark native stubs (FluidGradient, RzpGlass)
- Add native stubs for web-only components (MotionDashboard, StepperRouter, Rotate, Calendar, CustomMenu, SpinWheel)
- Add StyledDropdownOverlay native stub

## What's excluded

This branch intentionally excludes all storybook-specific changes (rollup config, metro config, storybook mocks, storybook.requires.js) from the `chore/storybook-rn-working` branch. The Toast stub is also excluded since master already ships a complete `useToast.native.tsx` implementation.

## Test plan

- [ ] Build the React Native app and verify no native build errors
- [ ] Verify Avatar, AvatarGroup, EmptyState render correctly on iOS and Android
- [ ] Verify Popover opens/closes correctly on iOS and Android without crashes
- [ ] Verify Tooltip renders correctly on native
- [ ] Verify Switch, Chips, Badge render with correct styling on native
- [ ] Verify TextInput label alignment on native

### DIFF
--- a/packages/blade/src/components/Avatar/Avatar.native.tsx
+++ b/packages/blade/src/components/Avatar/Avatar.native.tsx
@@ -1,5 +1,11 @@
 import React from 'react';
 import { Image, Pressable, Linking } from 'react-native';
+import { metaAttribute, MetaConstants } from '~utils/metaAttribute';
+import { throwBladeError } from '~utils/logger';
+import type { BladeElementRef } from '~utils/types';
+import { makeAnalyticsAttribute } from '~utils/makeAnalyticsAttribute';
+import { makeAccessible } from '~utils/makeAccessible';
+import getIn from '~utils/lodashButBetter/get';
 import type { AvatarProps } from './types';
 import {
   avatarSizeTokens,
@@ -11,23 +17,17 @@ import {
   avatarTextSizeMapping,
 } from './avatarTokens';
 import { useAvatarGroupContext } from './AvatarGroupContext';
+import { getInitials } from './avatarUtils';
 import { getStyledProps } from '~components/Box/styledProps';
-import { metaAttribute, MetaConstants } from '~utils/metaAttribute';
 import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
-import { throwBladeError } from '~utils/logger';
 import { UserIcon } from '~components/Icons';
-import type { BladeElementRef } from '~utils/types';
 import BaseBox from '~components/Box/BaseBox';
 import { getComponentId } from '~utils/isValidAllowedChildren';
-import { makeAnalyticsAttribute } from '~utils/makeAnalyticsAttribute';
 import { useTheme } from '~components/BladeProvider';
 import { Heading, Text } from '~components/Typography';
 import { getTextColorToken } from '~components/Button/BaseButton/BaseButton';
 import type { IconColor } from '~components/Button/BaseButton/types';
 import type { BaseTextProps } from '~components/Typography/BaseText/types';
-import { makeAccessible } from '~utils/makeAccessible';
-import getIn from '~utils/lodashButBetter/get';
-import { getInitials } from './avatarUtils';
 
 // Approximate native line-height for Indicator's internal Text.
 // On native an empty <Text> reserves line-height even with no content, pushing
--- a/packages/blade/src/components/Avatar/Avatar.native.tsx
+++ b/packages/blade/src/components/Avatar/Avatar.native.tsx
@@ -1,14 +1,241 @@
+import React from 'react';
+import { Image, Pressable, Linking } from 'react-native';
 import type { AvatarProps } from './types';
+import {
+  avatarSizeTokens,
+  avatarBorderRadiusTokens,
+  avatarToBottomAddonSize,
+  avatarToIndicatorSize,
+  avatarTopAddonOffsets,
+  avatarIconSizeTokens,
+  avatarTextSizeMapping,
+} from './avatarTokens';
+import { useAvatarGroupContext } from './AvatarGroupContext';
+import { getStyledProps } from '~components/Box/styledProps';
+import { metaAttribute, MetaConstants } from '~utils/metaAttribute';
+import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
 import { throwBladeError } from '~utils/logger';
+import { UserIcon } from '~components/Icons';
+import type { BladeElementRef } from '~utils/types';
+import BaseBox from '~components/Box/BaseBox';
+import { getComponentId } from '~utils/isValidAllowedChildren';
+import { makeAnalyticsAttribute } from '~utils/makeAnalyticsAttribute';
+import { useTheme } from '~components/BladeProvider';
+import { Heading, Text } from '~components/Typography';
+import { getTextColorToken } from '~components/Button/BaseButton/BaseButton';
+import type { IconColor } from '~components/Button/BaseButton/types';
+import type { BaseTextProps } from '~components/Typography/BaseText/types';
+import { makeAccessible } from '~utils/makeAccessible';
+import getIn from '~utils/lodashButBetter/get';
 
-const Avatar = (_props: AvatarProps): React.ReactElement => {
-  throwBladeError({
-    message: 'Avatar is not yet implemented for React Native',
-    moduleName: 'Avatar',
-  });
+// Approximate native line-height for Indicator's internal Text.
+// On native an empty <Text> reserves line-height even with no content, pushing
+// the SVG dot down when it is vertically centred inside the flex row.
+// Subtracting (lineHeight - svgSize) / 2 from the web top-offset compensates.
+const INDICATOR_TEXT_LINE_HEIGHT = 20;
 
-  // eslint-disable-next-line react/jsx-no-useless-fragment
-  return <></>;
+// Subtle-emphasis dot sizes (px) per indicator size — from indicatorDotSizes.
+const indicatorSubtleSvgSize: Record<'small' | 'medium' | 'large', number> = {
+  small: 6,
+  medium: 8,
+  large: 10,
 };
 
+function nativeTopOffset(webTop: string, indicatorSize: 'small' | 'medium' | 'large'): number {
+  const px = parseInt(webTop, 10); // e.g. '2px' → 2
+  const svgSize = indicatorSubtleSvgSize[indicatorSize];
+  const centeringOffset = Math.max(0, (INDICATOR_TEXT_LINE_HEIGHT - svgSize) / 2);
+  return px - centeringOffset;
+}
+
+const getInitials = (name: string): string => {
+  const names = name.trim().toUpperCase().split(' ');
+  if (names.length === 1) return names[0].substring(0, 2);
+  return names[0][0] + names[names.length - 1][0];
+};
+
+const _Avatar: React.ForwardRefRenderFunction<BladeElementRef, AvatarProps> = (
+  {
+    name,
+    color = 'neutral',
+    size = 'medium',
+    variant = 'circle',
+    icon,
+    href,
+    isSelected,
+    bottomAddon: BottomAddon,
+    topAddon,
+    src,
+    alt,
+    testID,
+    onClick,
+    ...rest
+  },
+  ref,
+) => {
+  if (__DEV__) {
+    if (src && !alt && !name) {
+      throwBladeError({
+        moduleName: 'Avatar',
+        message: '"alt" or "name" prop is required when the "src" prop is provided.',
+      });
+    }
+    if (topAddon && getComponentId(topAddon) !== 'Indicator') {
+      throwBladeError({
+        moduleName: 'Avatar',
+        message: 'TopAddon only accepts `Indicator` component.',
+      });
+    }
+  }
+
+  const groupProps = useAvatarGroupContext();
+  const { theme } = useTheme();
+  const avatarSize = groupProps?.size ?? size;
+  const isInteractive = Boolean(onClick || href);
+  const dimension = avatarSizeTokens[avatarSize];
+  const borderRadius =
+    theme.border.radius[
+      variant === 'circle'
+        ? avatarBorderRadiusTokens.circle
+        : avatarBorderRadiusTokens.square[avatarSize]
+    ];
+  const bgColor = getIn(theme.colors, `interactive.background.${color}.faded`, '') as string;
+
+  const iconColor = getTextColorToken({
+    property: 'icon',
+    variant: 'secondary',
+    color,
+    state: 'default',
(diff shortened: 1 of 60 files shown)


packages/blade/src/components/Tooltip/Tooltip.native.tsx
```diff
@@ -42,7 +48,19 @@ const Tooltip = ({
     ],
   });
 
-  const { refs, floatingStyles } = context;
+  const { refs, floatingStyles, update } = context;
```
anuraghazra, line 52: 🤔  Was it not working before this change? What is this fixing? 

---

Because when we developed this component it was working.

packages/blade/src/components/Spark/RzpGlass/RzpGlass.native.tsx
```diff
@@ -0,0 +1,14 @@
+import type { RzpGlassProps } from './types';
```
swapnil-kr1: Missing `import React from 'react'` — the JSX `<Text>` in the return statement needs React in scope (unless the project is configured for the new JSX transform everywhere, but other native stubs in this PR do import React explicitly). Same issue in `FluidGradient.native.tsx`.

packages/blade/src/components/Avatar/Avatar.native.tsx
```diff
@@ -1,14 +1,242 @@
...
+
+const _Avatar: React.ForwardRefRenderFunction<BladeElementRef, AvatarProps> = (
+  {
+    name,
+    color = 'neutral',
+    size = 'medium',
+    variant = 'circle',
+    icon,
+    href,
+    isSelected,
```
swapnil-kr1, line 62: **`getInitials` is duplicated** — the web `Avatar` already has an identical `getInitials` helper. Extract this into a shared `avatarUtils.ts` to avoid divergence. If the web version ever changes initials logic (e.g. hyphenated names, non-Latin scripts), this copy will silently drift.

packages/blade/src/components/Avatar/Avatar.native.tsx
```diff
@@ -1,14 +1,242 @@
...
+const INDICATOR_TEXT_LINE_HEIGHT = 20;
 
-  // eslint-disable-next-line react/jsx-no-useless-fragment
-  return <></>;
+// Subtle-emphasis dot sizes (px) per indicator size — from indicatorDotSizes.
+const indicatorSubtleSvgSize: Record<'small' | 'medium' | 'large', number> = {
+  small: 6,
+  medium: 8,
+  large: 10,
 };
```
swapnil-kr1, line 46: `INDICATOR_TEXT_LINE_HEIGHT = 20` and `indicatorSubtleSvgSize` are hardcoded magic numbers that duplicate values from the Indicator component's internals. If Indicator's sizing tokens change, these will break silently.

Can these be imported from indicator tokens, or at minimum add a `// sourced from indicatorTokens.ts` comment with a note to keep in sync?

packages/blade/src/components/Avatar/Avatar.native.tsx
```diff
@@ -1,14 +1,242 @@
...
+  };
+
+  const indicatorSize = avatarToIndicatorSize[avatarSize];
+  const topOffset = avatarTopAddonOffsets[variant][avatarSize];
+
+  const avatarBody = (
+    <BaseBox
+      {...metaAttribute({ name: MetaConstants.Avatar, testID })}
+      {...getStyledProps(rest)}
+      {...makeAnalyticsAttribute(rest)}
```
swapnil-kr1, line 162: `onClick?.(undefined as never)` — this `as never` cast silences the type system but is a footgun. If a consumer inspects the event argument on native they'll get `undefined` instead of a `GestureResponderEvent`.

Consider:
1. Not calling `onClick` on native (it's typed for web `MouseEvent`), and relying solely on `onPress` from the `Pressable` wrapper.
2. Or at minimum, pass `null` with a proper type annotation so consumers can safely check `if (event)`.

packages/blade/src/components/Avatar/AvatarGroup.native.tsx
```diff
@@ -1,14 +1,124 @@
...
+      >
+        {visible.map((child, i) => (
+          // Each wrapper is RING px larger on every side, filled with opaque intense-gray.
+          // The avatar body sits inside the padding exposing the gray ring as a separator.
+          // Ascending zIndex mirrors web's DOM-order stacking so each later avatar is on top.
+          <View
+            key={i}
+            style={{
+              padding: RING,
+              width: wrapperSize,
```
swapnil-kr1: Using array index as `key` (`key={i}`) — since AvatarGroup children can be reordered or dynamically added/removed, index keys will cause incorrect reconciliation and potential visual glitches (wrong avatar in wrong ring).

Consider: `key={React.isValidElement(child) ? (child.key ?? i) : i}` to preserve any keys already set on the Avatar children.

packages/blade/src/components/Popover/Popover.native.tsx
```diff
@@ -14,6 +15,8 @@ import { PopupArrow } from '~components/PopupArrow';
 import { getFloatingPlacementParts } from '~utils/getFloatingPlacementParts';
 import { componentZIndices } from '~utils/componentZIndices';
 
+const IOS_OFFSET_CORRECTION = 10;
+
```
swapnil-kr1, line 22: `IOS_OFFSET_CORRECTION = 10` — this is an unexplained magic number. What exactly causes the 10px offset on iOS? Status bar? Safe area inset? Portal container positioning?

Without context, someone will remove this during a cleanup and re-introduce the bug. Please add a comment explaining the root cause. Ideally derive it from `SafeAreaInsets` or a measurable value rather than hardcoding — different devices with dynamic islands or no notch will have different offsets.

packages/blade/src/components/Popover/Popover.native.tsx
```diff
@@ -52,9 +58,21 @@ const Popover = ({
...
-  const { refs, floatingStyles } = context;
+  const { refs, floatingStyles, update } = context;
   const [computedSide] = getFloatingPlacementParts(context.placement);
 
+  // Measure the Portal backdrop's window position to correct coordinate mismatch
+  // between measureInWindow (used by floating-ui) and the Portal container's coordinate space
+  const onBackdropLayout = React.useCallback(() => {
+    if (backdropRef.current && 'measureInWindow' in backdropRef.current) {
+      ((backdropRef.current as unknown) as {
+        measureInWindow: (cb: (x: number, y: number, w: number, h: number) => void) => void;
```
swapnil-kr1: This `onBackdropLayout` → `measureInWindow` → `setBackdropOffset` pattern is duplicated verbatim in `Tooltip.native.tsx` (~40 lines of identical logic including the Portal wrapping, backdrop ref, offset state, and position correction).

Extract a shared `useFloatingPortal()` hook that encapsulates:
- `backdropRef` + `backdropOffset` state
- `onBackdropLayout` callback
- the `update()` effect after offset measurement
- the `IOS_OFFSET_CORRECTION` application

This ensures future fixes (e.g. replacing the 50ms timeout) only happen in one place.

packages/blade/src/utils/makeBorderSize/makeBorderSize.native.ts
```diff
@@ -0,0 +1,19 @@
...
+ * call sites remain type-safe — the `as unknown as` casts are only in the implementation
+ * body and are never visible to callers.
+ */
+export function makeBorderSize<T extends number>(size: T): `${T}px`;
+export function makeBorderSize<T extends string>(size: T): T;
+export function makeBorderSize<T extends number | string>(size: T): `${T}px` | T {
+  if (typeof size === 'number') {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return (size as any) as `${T}px`;
+  }
```
swapnil-kr1, line 14: The overload signature says it returns `` `${T}px` `` (string) but the implementation returns `size` (a number) cast through `as any`. **The type lies to callers** — they think they're getting `"8px"` but actually get `8`.

Any code that does string operations on the return (e.g., `parseInt(makeBorderSize(8))`) will silently get `NaN` on native.

Since this is a `.native.ts` file, the return type for the numeric overload should honestly be `number`. The native and web overloads are allowed to have different return types — that's the whole point of platform-specific files.

packages/blade/src/components/Box/BaseBox/BaseBox.native.tsx
```diff
@@ -29,7 +29,8 @@ const StyledBaseBox = styled(View)
   })
   .withConfig({
     shouldForwardProp: (prop, defaultValidator) =>
-      isSupportedOnReactNativeElement(prop) && defaultValidator(prop),
+      isSupportedOnReactNativeElement(prop) &&
+      (typeof defaultValidator === 'function' ? defaultValidator(prop) : true),
```
swapnil-kr1, line 33: Good catch on `defaultValidator` not always being a function in styled-components v6. However, defaulting to `true` means if `defaultValidator` is absent, **all props get forwarded** to the native `View`, which could produce RN yellow-box warnings for unknown props.

Consider defaulting to `false` (don't forward) since `isSupportedOnReactNativeElement` is already the primary filter and should be sufficient.

packages/blade/src/components/Spark/RzpGlass/utils.native.ts
```diff
@@ -0,0 +1,54 @@
...
+  _preset: RzpGlassPreset = 'default',
+  _assetsPath: string = DEFAULT_CDN_PATH,
+): Promise<void> {
+  // no-op on native
+}
+
+// no-op stubs for web-only utilities — not available on native
+async function loadImage(_src: string): Promise<HTMLImageElement> {
+  return Promise.reject(new Error('loadImage is not supported on native'));
+}
```
swapnil-kr1, line 17: `loadImage` returns `Promise<HTMLImageElement>` and `loadVideo` returns `Promise<HTMLVideoElement>` — these are DOM types that don't exist in React Native's type environment. They compile only because `@types/react-native` doesn't actively conflict, which is fragile.

Since these are no-op stubs that always reject, type them as `Promise<never>` instead — both accurate and avoids pulling in DOM types.

packages/blade/src/components/Popover/PopoverContentWrapper.native.tsx
```diff
@@ -50,12 +50,14 @@ const PopoverContentWrapper = React.forwardRef<View, PopoverContentWrapperProps>
...
+    // _defineProperty() which is not a worklet and crashes Reanimated v4 on the UI thread.
     const animatedStyles = useAnimatedStyle(() => {
-      const transform = isHorizontal ? 'translateX' : 'translateY';
       return {
         opacity: opacity.value,
-        transform: [{ [transform]: translate.value }],
+        transform: isHorizontal
+          ? [{ translateX: translate.value }]
+          : [{ translateY: translate.value }],
       };
```
swapnil-kr1, line 61: Good fix on the computed property key crashing Reanimated v4 worklets.

Note: `isHorizontal` is captured from the outer scope but not in the `useAnimatedStyle` dependency array (only `[isVisible]` is listed). Since `isHorizontal` derives from `side` which can change when floating-ui flips placement, this could cause a stale closure. Consider adding `isHorizontal` to the deps array.

packages/blade/src/components/EmptyState/EmptyState.native.tsx
```diff
@@ -1,16 +1,92 @@
...
+            textAlign="center"
+            size={tokens.titleSize}
+            weight="semibold"
+            color="surface.text.gray.subtle"
+          >
+            {title}
+          </Heading>
+        ) : null}
+
+        {description ? (
```
swapnil-kr1, line 53: The gap between title and description is hardcoded to `spacing.1`. The web EmptyState uses `emptyStateSizeTokens[size].gapInsideContentSection` (or similar token) for this same gap. Use the token from `emptyStateTokens` here to stay consistent across platforms:
```tsx
<BaseBox display="flex" flexDirection="column" gap={tokens.gapInsideContentSection} alignItems="center">
```

packages/blade/src/components/Tooltip/Tooltip.native.tsx
shivam-rzpay: The tooltip was rendering, but its position was off by ~40px. For instance, when set to top, the content appeared ~40px lower than expected, overlapping the tooltip's trigger button.

packages/blade/src/components/Avatar/Avatar.native.tsx
shivam-rzpay: done

packages/blade/src/components/Avatar/Avatar.native.tsx
shivam-rzpay: done. Added a note for now

packages/blade/src/components/Avatar/Avatar.native.tsx
shivam-rzpay: done

packages/blade/src/components/Popover/Popover.native.tsx
shivam-rzpay: Portal container rendered by @gorhom/portal has a 10px top offset on iOS relative to the window coordinate system that measureInWindow doesn't account for.
Will add a comment here

packages/blade/src/components/Box/BaseBox/BaseBox.native.tsx
shivam-rzpay: This needs to be set as true for styled-component fix with latest react native versions.

Explanation :-
In styled-components v6, the defaultValidator passed to shouldForwardProp is undefined, not a function.
In v5, it was always a function you could call to check if a prop should be forwarded to the DOM/native element. In v6, that second argument was removed.

Setting the fallback to true means: "when there's no validator to ask, assume the prop is fine to forward." This is safe here because isSupportedOnReactNativeElement(prop) — the first condition in the && already filters out unsupported props. So by the time you reach the fallback, the prop has already been validated as a legitimate React Native prop.

Setting it to false means every prop gets blocked, since isSupportedOnReactNativeElement(prop) && false is always false. No styles reach the View, so the entire layout collapses — blank screen.

anuraghazra (dismissed):

swapnil-kr1 (approved):

PR #3400, opened 2026-05-20
feat: add agent context files for AI context

saurabhdaware:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade-core/CLAUDE.md
+++ b/packages/blade-core/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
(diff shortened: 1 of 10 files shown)


packages/blade-core/CLAUDE.md
```diff
@@ -0,0 +1 @@
+AGENTS.md
```
rohankokane-dev, line 1: Can you confirm if its the correct import syntax or do we need @ prefix? `@AGENTS.md`

rohankokane-dev (approved):

packages/blade-core/CLAUDE.md
saurabhdaware: Its not import. Its a symlink. Github just shows symlink like this in its content

saurabhdaware: image

PR #3401, opened 2026-05-20
feat: add get-agentic-blade-metrics skill and github action cron

saurabhdaware:
## Description

- This adds a skill that can be used for fetching the metrics whenever needed
- This also adds a github action cron job that prints the report everyday at https://github.com/razorpay/blade/wiki/agentic-blade-metrics

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


.github/scripts/fetch-metrics.mjs
```diff
@@ -0,0 +1,110 @@
...
+
+const REPO = "razorpay/blade";
+const AGENT_AUTHORS = new Set(["rzp-slash", "rzp-slash-public"]);
+const AGENT_REVIEWERS = new Set(["rzp-slash", "rzp-slash-public", "rzp-slash-reviewer"]);
+const BOT_COMMENTERS = new Set(["changeset-bot[bot]", "github-actions[bot]", "codesandbox-ci[bot]"]);
+const AUTO_APPROVE_LABEL = "rcore:eligible-for-auto-approval";
+
+function gh(...args) {
+  const result = execSync(`gh ${args.join(" ")}`, { encoding: "utf8", stdio: ["pipe", "pipe", "pipe"] });
+  return result.trim() ? JSON.parse(result) : [];
```
swapnil-kr1, line 21: **[Robustness]** `execSync` will throw on non-zero exit codes (e.g., rate-limit 403, network timeout). Since this runs in a loop over every PR, a single API hiccup kills the entire run.

Suggestion: wrap in try/catch, log a warning, and return `[]` on failure — or at minimum, let the caller decide whether to fail fast vs. skip.

.github/scripts/fetch-metrics.mjs
```diff
@@ -0,0 +1,110 @@
...
+const sinceLabel = formatDate(since);
+const todayLabel = formatDate(new Date());
+
+// Fetch PRs
+const allPRs = gh("pr", "list", "--repo", REPO, "--state", "all",
+  "--json", "number,title,createdAt,labels,author", "--limit", "200");
+
+const prs = allPRs.filter(p => new Date(p.createdAt) >= since);
+const totalPRs = prs.length;
+
```
swapnil-kr1, line 44: **[Bug]** `p.author.login?.replace("app/", "")` — the `login` field from `gh pr list --json author` does NOT have the `app/` prefix. GitHub App authors already appear as just `rzp-slash`. The `.replace("app/", "")` is a no-op here and gives a false sense of handling.

Can you double-check by looking at actual `gh pr list` output for a bot-authored PR? If `author.login` is indeed plain, this replace is dead code and should be removed to avoid confusion.

.github/scripts/fetch-metrics.mjs
```diff
@@ -0,0 +1,110 @@
...
+const humanPRs = totalPRs - agentPRs.length;
+const autoApproved = prs.filter(p => p.labels.some(l => l.name === AUTO_APPROVE_LABEL));
+
+// Fetch comments for each PR
+let totalComments = 0;
+let agentComments = 0;
+let resolvedByAgent = 0;
+
+for (const pr of prs) {
+  const reviewComments = gh("api", `repos/${REPO}/pulls/${pr.number}/comments`, "--paginate");
```
swapnil-kr1, line 55: **[Performance / Rate Limiting]** This makes **2 API calls per PR** (review comments + issue comments), sequentially. For 30 days of data with, say, 80 PRs, that is 160 API requests — easily hitting GitHub's secondary rate limits (900 points/min for `GITHUB_TOKEN`).

Suggestions:
1. Use `gh api` with `--paginate` and `per_page=100` to minimize calls.
2. Consider parallelizing with `Promise.all` (batched, e.g., 5 at a time) to speed up runs.
3. Add a brief `--header "X-GitHub-Api-Version: 2022-11-28"` for forward compatibility.

.github/scripts/push-agentic-metrics-wiki.mjs
```diff
@@ -0,0 +1,44 @@
...
+  console.error("Usage: node push-agentic-metrics-wiki.mjs <markdown-file>");
+  process.exit(1);
+}
+
+const content = readFileSync(markdownFile, "utf8");
+
+const wikiDir = join(tmpdir(), `blade-wiki-${Date.now()}`);
+
+try {
+  execSync(`git clone ${WIKI_REPO} ${wikiDir}`, { stdio: "inherit" });
```
swapnil-kr1, line 30: **[Security]** Hardcoded `WIKI_REPO` URL means the `GH_TOKEN` configured in the workflow (via `insteadOf`) provides push access. This is fine for now, but note that `GITHUB_TOKEN` has **read-only** access to wikis by default — this step will fail silently unless the repo has wiki write permissions enabled for Actions, or a PAT/GitHub App token is used.

Have you tested this end-to-end in a workflow run? If not, worth a manual `workflow_dispatch` test before merging.

.github/scripts/push-agentic-metrics-wiki.mjs
```diff
@@ -0,0 +1,44 @@
...
+
+  writeFileSync(join(wikiDir, WIKI_PAGE), content);
+
+  execSync("git add .", { cwd: wikiDir, stdio: "inherit" });
```
swapnil-kr1, line 34: **[Robustness]** If there are no changes (e.g., re-running on same day with same data), `git commit` will exit non-zero with "nothing to commit". This will crash the script.

Add `--allow-empty` or check for changes before committing:
```js
const status = execSync("git status --porcelain", { cwd: wikiDir, encoding: "utf8" });
if (status.trim()) {
  execSync("git add .", ...);
  execSync(`git commit -m "..."`, ...);
  execSync("git push", ...);
}
```

.github/workflows/blade-agentic-metrics.yml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Fetch metrics (1 day)
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
swapnil-kr1, line 22: **[Reliability]** The 3 `fetch-metrics` steps run sequentially. For 30 days of data, the PR comment fetching loop (2 API calls per PR) could easily take 5-10 minutes and risk hitting rate limits.

Would it make sense to fetch all data once (e.g., 30-day window) and compute 1d/7d/30d slices from that single dataset? This would:
- Cut API calls by ~3x
- Avoid inconsistencies if PRs are opened between the 3 runs
- Make the workflow significantly faster

.github/workflows/blade-agentic-metrics.yml
```diff
@@ -0,0 +1,64 @@
...
+        run: node .github/scripts/fetch-metrics.mjs 7 > /tmp/metrics-7d.md
+
+      - name: Fetch metrics (30 days)
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: node .github/scripts/fetch-metrics.mjs 30 > /tmp/metrics-30d.md
+
+      - name: Combine reports
+        run: |
+          echo "# Agentic Blade Metrics" > /tmp/metrics-combined.md
```
swapnil-kr1, line 37: **[Maintainability]** This heredoc-style combine step is pretty verbose. A simpler approach:

```yaml
- name: Combine reports
  run: |
    { echo "# Agentic Blade Metrics";
      for label in "Last 1 Day:1d" "Last 7 Days:7d" "Last 30 Days:30d"; do
        echo -e "\n---\n\n## ${label%%:*}\n";
        cat /tmp/metrics-${label##*:}.md;
      done
    } > /tmp/metrics-combined.md
```

Not blocking, just a readability suggestion.

swapnil-kr1 (approved):

.github/scripts/fetch-metrics.mjs
saurabhdaware: We actually want a single API hiccup to kill the entire run. Otherwise it will inject wrong data to report

.github/scripts/fetch-metrics.mjs
saurabhdaware: > the login field from gh pr list --json author does NOT have the app/ prefix

It does
image

.github/scripts/push-agentic-metrics-wiki.mjs
saurabhdaware: > Have you tested this end-to-end in a workflow run? If not, worth a manual workflow_dispatch test before merging.

Yes yes wiki is already being pushed to wiki page. You can check it here https://github.com/razorpay/blade/wiki/agentic-blade-metrics

.github/scripts/push-agentic-metrics-wiki.mjs
saurabhdaware: The commit is for wiki. The script itself adds content for it to push so it will always have content to commit

.github/workflows/blade-agentic-metrics.yml
saurabhdaware: The blade does not have as many PRs so we are not really hitting anywhere the API limit. Can do these optimisations if we see issue. Right now its a very basic script just for our internal metrics

Right now the same script is shared between github action and skill so keeping isolated script for every duration just makes the job for skill a bit easy

.github/workflows/blade-agentic-metrics.yml
saurabhdaware: can do this if we decide to add more durations. Current one seems easier to read for me. Just 3 lines vs 1 line code

.github/scripts/fetch-metrics.mjs
saurabhdaware: > Use gh api with --paginate and per_page=100 to minimize calls.

--paginate makes internal API calls which also count towards the API rate limit so this won't really save the API calls 

image


Also github API rate limit is 5000 requests per hour. Blade will need to have 2500 PRs in a month to hit that

PR #3405, opened 2026-05-21
feat(blade-svelte): BottomSheet component

rohankokane-dev:
## Description

Migrates the React BottomSheet component to Svelte 5.

## Changes

- Adds `packages/blade-svelte/src/components/BottomSheet/` — Svelte 5 implementation: `BottomSheet`, `BottomSheetHeader`, `BottomSheetBody`, `BottomSheetFooter`, `BottomSheetBackdrop`, context, stack store, utils, types, and 13 stories.
- Adds `packages/blade-core/src/styles/BottomSheet/` — framework-agnostic CSS module + CVA helpers + constants (`BOTTOM_SHEET_Z_INDEX`, `BOTTOM_SHEET_DEFAULT_SNAP_POINTS`).
- Adds `@use-gesture/vanilla@10.2.24` + `body-scroll-lock-upgrade@1.1.0` to `blade-svelte/package.json` to match the React versions exactly (drag rubberband, velocity-driven momentum, `filterTaps`, scroll lock with `reserveScrollBarGap: true`).
- Re-exports BottomSheet from `packages/blade-svelte/src/components/index.ts`.
- Re-exports CSS classes/constants from `packages/blade-core/src/styles/index.ts`.

## Architecture

Single-layer (matches React — no `BaseBottomSheet.tsx` exists). All logic (drag, snap-point, presence, stack registration, scroll lock) lives in `BottomSheet.svelte`. Sub-components are siblings.

## Stories

All 13 React stories preserved verbatim under `Components/BottomSheet`:

1. Default, 2. With Header Footer, 3. With Dropdown Single Select, 4. With Dropdown Button, 5. With Dropdown Multi Select, 6. With Dropdown Sections Select, 7. Bottom Sheet Stacking, 8. Initial Focus, 9. Zero Padding, 10. Custom Snap Points, 11. With OTP Input, 12. Product Use Case 1, 13. Non-Dismissible BottomSheet.

Stories 3, 4, 5, 6, 10 substitute the unmigrated `<Dropdown><SelectInput><ActionList>` chain with a `<Button>` trigger + native `<button>`-based ActionList placeholder (Plan Decision #3). Story 8 uses native `<input>` instead of unmigrated `TextInput`. Functional parity preserved.

## Deferred (per Plan Decisions, to be revisited when deps migrate)

- Body bullets render via native `<ul><li>` instead of React's `<List><ListItem>` (List not migrated).
- ActionList substitute in 5 stories (ActionList not migrated).
- `bottomSheetAndDropdownGlue` stubbed as a `null` getter (Dropdown not migrated).
- Header `children` slot accepted as passthrough; AutoComplete integration deferred (depends on Dropdown).

## Verification

- `yarn svelte-check` (blade-svelte): 0 errors.
- `yarn build` (blade-core, blade-svelte): 0 errors.
- Visual parity confirmed against React Storybook for Default (open) and With Header Footer (open) stories — pixel-identical surface, backdrop, header (title + counter + subtitle + close X + badge), footer (full-width buttons).
- Drag library parity restored via iteration 2 patch: `@use-gesture/vanilla@10.2.24` with `filterTaps: true` on both grab-handle and scroll-element `DragGesture` instances. Drag handler is a verbatim port of `BottomSheet.web.tsx` lines 248–368.
- Body scroll lock library parity restored: `body-scroll-lock-upgrade@1.1.0` (`disableBodyScroll` + `clearAllBodyScrollLocks`) with `reserveScrollBarGap: true`.

## Component Checklist

- [ ] Add KitchenSink Story
- [x] Add changeset


packages/blade-svelte/src/components/BottomSheet/BottomSheet.svelte
```diff
@@ -0,0 +1,537 @@
...
+  let isMounted = $state(false);
+  let isVisible = $state(false);
+  let unmountTimeoutId: ReturnType<typeof setTimeout> | null = null;
+  /* Allow ~320ms (--duration-moderate 280ms + buffer) for the exit
+   * transition before unmounting. */
+  const UNMOUNT_FALLBACK_MS = 320;
+
+  /* Mirrors React's `preventScrollingRef` — set during drag to gate the
+   * body's scroll/touchmove listeners below. */
+  let preventScrolling = $state(true);
```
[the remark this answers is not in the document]
rohankokane-dev: Not a bug. Initial position auto-picks `snapPoints[0]` or `snapPoints[1]` (lines 112–120) — never `[2]`. With defaults `[0.35, 0.5, 0.85]` the open position is always `< upperSnapPoint`, which is precisely the value the drag handler would set (`preventScrolling = newY < upperSnapPoint`). Initial `true` matches React parity (`preventScrollingRef = useRef(true)`) and is the correct value for the actual initial position. No-op.

packages/blade-svelte/src/components/BottomSheet/BottomSheet.svelte
```diff
@@ -0,0 +1,537 @@
...
+    const node = scrollEl;
+    function preventScrollingHandler(event: Event): void {
+      if (!preventScrolling) return;
+      const target = event.target as Element | null;
+      const isAllowedComponent = target?.closest('[data-allow-scroll]');
+      if (!isAllowedComponent) {
+        event.preventDefault();
+      }
+    }
+    node.addEventListener('scroll', preventScrollingHandler);
```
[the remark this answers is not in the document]
rohankokane-dev: Fixed in f2f0349. Dropped the `scroll` listener (`scroll` is non-cancelable per spec, `preventDefault()` was a no-op) and added `{ passive: false }` to the `touchmove` listener so `preventDefault()` is honored on modern browsers (touchmove on scrollables defaults to passive otherwise). React side has the same dead listener — left as-is for now to keep the migration scope tight.

packages/blade-svelte/src/components/BottomSheet/BottomSheet.svelte
```diff
@@ -0,0 +1,537 @@
...
+  }
+  /* eslint-enable @typescript-eslint/no-explicit-any */
+
+  /* Wire `DragGesture` instances reactively. We attach to the same elements
+   * React binds: the grab handle (with no `args`, so `isContentDragging` is
+   * false) and the body's scroll element (with `isContentDragging: true`).
+   * `from` is a getter so each drag-start picks up the *current* `positionY`. */
+  $effect(() => {
+    if (!grabHandleEl) return undefined;
+    if (!isOnTopOfStack || !isOpen) return undefined;
```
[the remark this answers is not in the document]
rohankokane-dev: Fixed in f2f0349. Both `DragGesture` effect cleanups now reset `isDragging = false` before `gesture.destroy()` so the exit transition isn't suppressed by `transition: none` when `isOpen` flips false mid-drag.

packages/blade-svelte/src/components/BottomSheet/index.ts
```diff
@@ -0,0 +1,59 @@
...
+ *     </BottomSheetFooter>
+ *   {/snippet}
+ * </BottomSheet>
+ * ```
+ */
+export { default as BottomSheet } from './BottomSheet.svelte';
+export { default as BottomSheetHeader } from './BottomSheetHeader.svelte';
+export { default as BottomSheetBody } from './BottomSheetBody.svelte';
+export { default as BottomSheetFooter } from './BottomSheetFooter.svelte';
+export { default as BottomSheetBackdrop } from './BottomSheetBackdrop.svelte';
```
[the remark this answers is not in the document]
rohankokane-dev: Fixed in f2f0349. Removed `BottomSheetBackdrop` and the stack helpers (`bottomSheetStack`, `addBottomSheetToStack`, `removeBottomSheetFromStack`, `getTopOfTheStack`, `getCurrentStackIndexById`) from both `BottomSheet/index.ts` and `components/index.ts`. Public API now matches React: only `BottomSheet`, `BottomSheetHeader`, `BottomSheetBody`, `BottomSheetFooter` are exported.

packages/blade-svelte/src/components/BottomSheet/BottomSheetHeader.svelte
```diff
@@ -0,0 +1,208 @@
...
+   * derivation). */
+  $effect(() => {
+    ctx?.setIsHeaderEmpty(isHeaderEmpty);
+  });
+
+  /* Measure the header height after mount/open so the parent can size its
+   * total content. Mirrors React's `useIsomorphicLayoutEffect` measurement. */
+  $effect(() => {
+    if (!headerEl) return;
+    if (!title && !subtitle) return;
```
saurabhdaware: 🟠 **[MAJOR]** · _code-quality-critique_

**Problem:** The `setHeaderHeight` measurement effect has a guard `if (!title && !subtitle) return;` that skips reporting the header height when the header has only `leading`, `trailing`, `showBackButton`, or `children` slots (no title/subtitle). Since `isHeaderEmpty` correctly accounts for all those props, the parent snap-point math receives `headerHeight = 0` for a non-empty header with only slotted content, causing the sheet to open at the wrong height.

**Suggestion:** Remove the `if (!title && !subtitle) return;` guard from the measurement effect (or broaden it to `if (isHeaderEmpty) return;`) so height is always reported when the header is non-empty.

packages/blade-svelte/src/components/BottomSheet/types.ts
```diff
@@ -0,0 +1,167 @@
...
+export interface BottomSheetBodyProps extends StyledPropsBlade {
+  /** Body content. */
+  children: Snippet;
+
+  /**
+   * Equal padding applied on all sides of the body content. Only `spacing.0`
+   * and `spacing.5` are allowed deliberately.
+   *
+   * @default 'spacing.5'
+   */
```
saurabhdaware, line 131: 🟠 **[MAJOR]** · _api-decision-critique_

**Problem:** The `hasActionList` prop is added to `BottomSheetBodyProps` in the Svelte implementation but does not exist on the React `BottomSheetBodyProps`. This is a Svelte-only workaround that leaks internal layout logic into the public API. React detects the ActionList child automatically via `isValidAllowedChildren` — consumers never pass this flag. Exposing it on the Svelte side creates an API inconsistency between platforms and adds a footgun: consumers who pass `hasActionList={true}` with something other than an ActionList will get incorrect padding silently.

**Suggestion:** Keep `hasActionList` as an internal implementation detail rather than a public prop. If Svelte truly cannot introspect snippet children, document the decision in a code comment inside `BottomSheetBody.svelte` and accept the known divergence — but do not expose it as a public prop.

packages/blade-svelte/src/components/BottomSheet/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+    contentHeight,
+    footerHeight,
+  }: {
+    maxHeight: number;
+    headerHeight: number;
+    footerHeight: number;
+    contentHeight: number;
+  },
+  minHeight?: number,
+): number {
```
saurabhdaware, line 71: 🔵 **[MINOR]** · _code-quality-critique_

**Problem:** `utils.ts` exports a local `rubberbandIfOutOfBounds` that is never imported by any file in this PR — `BottomSheet.svelte` imports it directly from `@use-gesture/vanilla`. The local copy is dead code that duplicates a library function.

**Suggestion:** Remove the `rubberbandIfOutOfBounds` export from `utils.ts`. If callers outside the component need it, they can import from `@use-gesture/vanilla` directly.

packages/blade-svelte/src/components/BottomSheet/types.ts
```diff
@@ -0,0 +1,167 @@
...
+   * default focus is moved to the close button. Svelte callers pass the
+   * element obtained via `bind:this`.
+   *
+   * @default null
+   */
+  initialFocusRef?: HTMLElement | null;
+
+  /**
+   * Sets the z-index of the bottom sheet. When stacking multiple sheets keep
+   * `zIndex` identical across all of them — the stacking context is computed
```
saurabhdaware, line 58: 🔵 **[MINOR]** · _api-decision-critique_

**Problem:** The `initialFocusRef` prop accepts `HTMLElement | null` in the Svelte API, whereas the React API types it as `React.MutableRefObject<any>`. In Svelte, `bind:this` produces `HTMLElement | undefined`, not `null`, so callers may need to initialize with `null` explicitly to satisfy the type.

**Suggestion:** Type as `HTMLElement | null | undefined` to match what `bind:this` naturally produces and to avoid requiring callers to explicitly initialize to `null`.

saurabhdaware (commented):
## ✨ Agentic PR Review ✨

### UI Review

🔗 Storybook: [Preview](https://61c19ee8d3d282003ac1d81c-yvxpnflyld.chromatic.com/)

✅ 6 passed

<details>
<summary>Passing checks (6)</summary>

| Check |
|-------|
| ✅ [BottomSheet Default — open, render, close](https://61c19ee8d3d282003ac1d81c-yvxpnflyld.chromatic.com/svelte/?path=/story/components-bottomsheet--default) |
| ✅ [BottomSheet With Header Footer — rich header slots](https://61c19ee8d3d282003ac1d81c-yvxpnflyld.chromatic.com/svelte/?path=/story/components-bottomsheet--with-header-footer) |
| ✅ [BottomSheet Stacking — open sheet over sheet](https://61c19ee8d3d282003ac1d81c-yvxpnflyld.chromatic.com/svelte/?path=/story/components-bottomsheet--bottom-sheet-stacking) |
| ✅ [BottomSheet Non-Dismissible — locked open](https://61c19ee8d3d282003ac1d81c-yvxpnflyld.chromatic.com/svelte/?path=/story/components-bottomsheet--non-dismissible-bottom-sheet) |
| ✅ [BottomSheet Zero Padding — edge-to-edge body](https://61c19ee8d3d282003ac1d81c-yvxpnflyld.chromatic.com/svelte/?path=/story/components-bottomsheet--zero-padding) |
| ✅ [BottomSheet Product Use Case 1 — back button + disabled action](https://61c19ee8d3d282003ac1d81c-yvxpnflyld.chromatic.com/svelte/?path=/story/components-bottomsheet--product-use-case-1) |

</details>

### CI / Sanity

✅ 13 passed · ❌ 1 failed

| Check | Problem |
|-------|---------|
| [Validate Source Code](https://github.com/razorpay/blade/actions/runs/27541942500/job/81405659928) | Lint Blade Svelte failed. The exports for BottomSheet, BottomSheetHeader, BottomSheetBody, and BottomSheetFooter in packages/blade-svelte/src/components/index.ts use newlines and extra indentation instead of single spaces.<br><br>**Suggestion:** Run `yarn lint --fix` in packages/blade-svelte to auto-fix the formatting, or reformat the export to a single line: `export { BottomSheet, BottomSheetHeader, BottomSheetBody, BottomSheetFooter } from './BottomSheet';` |

<details>
<summary>Passing checks (13)</summary>

| Check |
|-------|
| ✅ [Generate Coverage Report](https://github.com/razorpay/blade/actions/runs/27541942500/job/81406243571) |
| ✅ [label](https://github.com/razorpay/blade/actions/runs/27541942533/job/81405659896) |
| ✅ [Run Tests (1)](https://github.com/razorpay/blade/actions/runs/27541942500/job/81405659957) |
| ✅ [Run Tests (2)](https://github.com/razorpay/blade/actions/runs/27541942500/job/81405660002) |
| ✅ [Run Tests (3)](https://github.com/razorpay/blade/actions/runs/27541942500/job/81405659983) |
| ✅ [Run Tests (4)](https://github.com/razorpay/blade/actions/runs/27541942500/job/81405660353) |
| ✅ [Knowledgebase Lint](https://github.com/razorpay/blade/actions/runs/27541942585/job/81405660088) |
| ✅ [PR Title Check](https://github.com/razorpay/blade/actions/runs/27541941239/job/81405656523) |
| ✅ [Chromatic Deployment](https://github.com/razorpay/blade/actions/runs/27541940054/job/81405652205) |
| ✅ [semgrep-cloud-platform/scan](https://semgrep.dev/orgs/razorpay/projects/3302/scans/180883754) |
| ✅ [Danger](http://danger.systems/js) |
| ✅ [Storybook Publish: blade](https://61c19ee8d3d282003ac1d81c-yvxpnflyld.chromatic.com/) |
| ✅ [ci/codesandbox](https://ci.codesandbox.io/status/razorpay/blade/pr/3405/builds/685608) |

</details>

> ⚠️ Missing changeset bump for @razorpay/blade-core (packages/blade-core was modified)
> ⚠️ Missing changeset bump for @razorpay/blade-svelte (packages/blade-svelte was modified)

packages/blade-svelte/src/components/BottomSheet/types.ts
saurabhdaware: @rzp-slash-public check if comment is valid, if yes, resolve this, else respond here with what is not valid ^^

packages/blade-svelte/src/components/BottomSheet/BottomSheetHeader.svelte
saurabhdaware: @rzp-slash-public check if comment is valid, if yes, resolve this, else respond here with what is not valid ^^

packages/blade-svelte/src/components/BottomSheet/types.ts
rohankokane-dev: Keeping `hasActionList` as a public prop. Svelte snippets are opaque functions — there is no `React.Children.forEach` / `isValidAllowedChildren` analogue, so the body cannot detect an ActionList child the way React does. The CVA compound-variant in `blade-core` still requires this signal, so we expose it as an explicit escape hatch (documented as such in JSDoc). All stories that contained ActionList in React already pass `hasActionList` here, matching the React UX 1:1. Switching to a context-based detection would couple ActionList to BottomSheet and isn't done in this migration.

saurabhdaware (dismissed):

saurabhdaware (approved):

PR #3398, opened 2026-05-20
feat(blade): add SankeyChart component

debabratama1ik:
## Summary

Adds a \`SankeyChart\` component to the Blade design system for visualising flow distributions across stages (payment routing, funnel analysis, budget allocation, etc.).

This PR addresses all engineering review feedback from Swapnil and rkdotdesign.

---

## What changed (Swapnil's review → addressed)

| Point | Action |
|-------|--------|
| Use Recharts native \`<Sankey>\` (no extra deps) | ✅ Fully migrated; Recharts \`<Sankey>\` already a Blade dep at v3.7.0 |
| Align with Blade chart pattern | ✅ \`CommonChartComponentsContext\`, \`ResponsiveContainer\`, Recharts \`<Tooltip>\` |
| Type color tokens properly | ✅ All color props use \`ChartsCategoricalColorToken\` |
| Flatten data shape (closer to Recharts format) | ✅ \`{ nodes, links }\` with string IDs; no more \`levels[]\` |

---

## API

\`\`\`tsx
<SankeyChart
  data={{
    nodes: [
      { id: 'total',   name: 'Total' },
      { id: 'upi',     name: 'UPI' },
      { id: 'card',    name: 'Card' },
      { id: 'success', name: 'Successful' },
      { id: 'failed',  name: 'Failed' },
    ],
    links: [
      { source: 'total', target: 'upi',     value: 4000 },
      { source: 'total', target: 'card',    value: 3200 },
      { source: 'upi',   target: 'success', value: 3500 },
      { source: 'upi',   target: 'failed',  value: 500  },
      { source: 'card',  target: 'success', value: 2800 },
      { source: 'card',  target: 'failed',  value: 400  },
    ],
  }}
  labelUnit="txn"
  height={420}
  showTooltip
  showLabels
/>
\`\`\`

### Props

| Prop | Type | Default | Description |
|------|------|---------|-------------|
| \`data\` | \`{ nodes: SankeyDataNode[]; links: SankeyDataLink[] }\` | required | Flat node list + directed links referenced by node id |
| \`height\` | \`number\` | \`400\` | SVG height in px |
| \`nodeColorOverride\` | \`ChartsCategoricalColorToken\` | — | Apply a single Blade token to all node bars |
| \`linkColorOverride\` | \`ChartsCategoricalColorToken\` | — | Apply a single Blade token to all link ribbons |
| \`showTooltip\` | \`boolean\` | \`true\` | Recharts-managed tooltip on hover |
[...]

PR #3414, opened 2026-05-22
feat(blade): add SankeyChart component

debabratama1ik:
> _Replaces #3398 (moved from fork to internal branch so CI runs without fork approval gates)_

## Summary

Adds a `SankeyChart` component to Blade for visualising flow distributions across stages — payment routing, funnel analysis, budget allocation, etc.

This PR addresses all engineering review feedback from Swapnil (@swapnil-kr1) and rkdotdesign.

---

## API

Follows the **`ChartXxxWrapper + ChartXxx` composition pattern** used by DonutChart, BarChart, LineChart, etc.

```tsx
<ChartSankeyWrapper
  height={420}
  showTooltip
  nodeColorOverride="data.background.categorical.blue.intense"
  linkColorOverride="data.background.categorical.blue.subtle"
>
  <ChartSankey
    data={{
      nodes: [
        { id: 'total',   name: 'Total' },
        { id: 'upi',     name: 'UPI' },
        { id: 'card',    name: 'Card' },
        { id: 'success', name: 'Successful' },
        { id: 'failed',  name: 'Failed' },
      ],
      links: [
        { source: 'total', target: 'upi',     value: 4000 },
        { source: 'total', target: 'card',    value: 3200 },
        { source: 'upi',   target: 'success', value: 3500 },
        { source: 'upi',   target: 'failed',  value: 500  },
        { source: 'card',  target: 'success', value: 2800 },
        { source: 'card',  target: 'failed',  value: 400  },
      ],
    }}
    labelUnit="txn"
    showLabels
    showLabelChip
    showPercentage
  />
</ChartSankeyWrapper>
```

### `ChartSankeyWrapper` props

| Prop | Type | Default | Description |
|------|------|---------|-------------|
| `children` | `React.ReactNode` | required | Must contain a `<ChartSankey>` |
| `height` | `number` | `400` | SVG height in px |
| `width` | `number \| string` | `"100%"` | SVG width — pixel number or CSS string (e.g. `"600px"`). Useful for fixed-width dashboard layouts |
| `showTooltip` | `boolean` | `true` | Recharts-managed tooltip on hover |
| `nodeColorOverride` | `ChartsCategoricalColorToken` | — | Apply one Blade token to all node bars |
| `linkColorOverride` | `ChartsCategoricalColorToken` | — | Apply one Blade token to all link ribbons |
| `testID` | `string` | — | `data-testid` on the root wrapper |

### `ChartSankey` props

| Prop | Type | Default | Description |
|------|------|---------|-------------|
| `data` | `{ nodes: SankeyDataNode[]; links: SankeyDataLink[] }` | required | Flat node list + directed links referenced by node id |
| `showLabels` | `boolean` | `true` | Label chips/text to the right of each node bar |
| `showLabelChip` | `boolean` | `true` | When `true`, labels render as Blade-styled chips. When `false`, renders as plain SVG text |
| `showPercentage` | `boolean` | `true` | Show percentage of total flow alongside value in each label |
| `labelUnit` | `string` | — | Unit appended to value (e.g. `"txn"`, `"₹M"`) |
| `formatValue` | `(value: number) => string` | — | Custom formatter. Default truncates (not rounds) to Indian notation (k/L/Cr) |
| `onNodeClick` | `(node: SankeyDataNode, index: number) => void` | — | Fired when a node bar is clicked |
| `onLinkClick` | `(link: SankeyDataLink, index: number) => void` | — | Fired when a link ribbon is clicked |

---

## Design / visual behaviour

- **Colour palette**: Blade categorical tokens, auto-rotated per node (same pool as all other Blade charts)
- **Per-node colour**: set `color: ChartsCategoricalColorToken` on individual `SankeyDataNode` entries
- **Semantic outcome colours**: nodes named `Successful` use `green.subtle`, `Failed` use `red.subtle` in default stories
- **Label chips**: Blade-themed chip to the right of each node — name + humanized value + percentage. Wraps to two lines only when content exceeds chip max width
- **Plain text labels**: `showLabelChip={false}` renders the same info as plain SVG `<text>` with correct symmetric line spacing
- **Hover highlight**: hovered node/link dims unrelated elements; smooth opacity transition
- **Tooltip**: Recharts-managed tooltip (auto edge-flip); Blade `staticBlack` background, `staticWhite` text
- **Mobile**: horizontal scroll wrapper at `min-width: 560px`; no layout collapse on narrow viewports

---

## Implementation notes

- **No new dependencies** — layout handled entirely by Recharts `<Sankey>` (already a Blade dep at v3.7.0)
- **Composition pattern** — `ChartSankeyWrapper` inspects its `ChartSankey` child via `getComponentId` to extract `data` and build `dataColorMapping`, then provides `CommonChartComponentsContext` + private `SankeyChartContext`. Mirrors DonutChart exactly.
- **`componentIds`** — merged into `tokens.ts` with PascalCase values (`ChartSankey`, `ChartSankeyWrapper`), matching BarChart / LineChart / AreaChart convention
- **Context null guard** — `SankeyChartContext` defaults to `null`; rendering `ChartSankey` outside `ChartSankeyWrapper` throws a descriptive Blade error
- **`calculateTextWidth` extended** — added `skipPadding?: boolean` to `CommonChartComponents/utils.ts`; bypasses MIN_WIDTH floor for accurate per-segment layout. Fully backward-compatible.
- **Label helpers** — `renderNode` delegates to `renderChipLabel` / `renderPlainTextLabel` pure module-level functions
- **`ResponsiveContainer`** — owned by `ChartSankeyWrapper`; `ChartSankey` renders `<Sankey>` directly with injected `width` / `height`
- **Event handler casts** — `onMouseEnter` / `onClick` are cast via `as any`; Recharts Sankey's event prop is `MouseEventHandler<SVGSVGElement> & ((item, type, e) => void)`, an intersection TypeScript cannot satisfy with a plain callback

---

## Files changed

| File | Change |
|------|--------|
| `SankeyChart/types.ts` | `ChartSankeyWrapperProps` (+ `width`), `ChartSankeyProps` (+ `formatValue`), `SankeyDataNode`, `SankeyDataLink` |
| `SankeyChart/SankeyChart.web.tsx` | Full implementation: composition, context, rendering, hover, tooltip, label helpers |
| `SankeyChart/SankeyChart.native.tsx` | Dual `throwBladeError` stubs (`ChartSankeyWrapper` + `ChartSankey`) |
| `SankeyChart/tokens.ts` | Layout constants + `LABEL_CAP_HEIGHT_RATIO` + `componentIds` (no separate `componentIds.ts`) |
| `SankeyChart/index.ts` | Clean re-exports of both components + all types |
| `SankeyChart/SankeyChart.stories.tsx` | 4 stories with `action()` handlers on Default story |
| `SankeyChart/__tests__/SankeyChart.web.test.tsx` | 24 tests + 1 snapshot; correctness-level assertions |
| `SankeyChart/__tests__/SankeyChart.native.test.tsx` | Tests both stubs throw correct Blade errors |
| `CommonChartComponents/utils.ts` | Extended `calculateTextWidth` with `skipPadding` option |
| `Charts/_decisions/decisions.md` | Added §3.7 Sankey Chart; fixed §3.8.1 numbering |

---

## Storybook preview (this PR)

🔗 [SankeyChart stories on Chromatic](https://61c19ee8d3d282003ac1d81c-zpojrpaneu.chromatic.com/?path=/story/components-charts-sankeychart--default-sankey-chart)

| Story | Direct link |
|-------|-------------|
| Default Sankey Chart | [open](https://61c19ee8d3d282003ac1d81c-zpojrpaneu.chromatic.com/?path=/story/components-charts-sankeychart--default-sankey-chart) |
| Single Color | [open](https://61c19ee8d3d282003ac1d81c-zpojrpaneu.chromatic.com/?path=/story/components-charts-sankeychart--single-color-sankey-chart) |
| Without Labels | [open](https://61c19ee8d3d282003ac1d81c-zpojrpaneu.chromatic.com/?path=/story/components-charts-sankeychart--sankey-chart-without-labels) |
| Plain Text Labels | [open](https://61c19ee8d3d282003ac1d81c-zpojrpaneu.chromatic.com/?path=/story/components-charts-sankeychart--sankey-chart-with-plain-text-labels) |

---

## Test plan

- [x] `SHARD='' yarn test:react --testPathPattern=SankeyChart --forceExit` → 24/24 pass, 1 snapshot
- [x] `SHARD='' yarn test:native --testPathPattern=SankeyChart --forceExit` → 2/2 pass
- [x] TypeScript build clean (`tsc -p tsconfig-generate-types.web.json`)
- [x] ESLint clean with `--max-warnings=0` on all changed source files
- [ ] Storybook: Default, SingleColor, WithoutLabels, PlainTextLabels stories render correctly
- [ ] Hover/click interactions on node bars and link ribbons (check Actions panel)
- [ ] Mobile viewport (< 560px) → horizontal scroll, no layout collapse
- [ ] Tooltip near chart edges → auto-flips, no clipping


packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,687 @@
...
+  // Cap-height ratio for Inter — uppercase glyphs occupy ~72% of font-size above the baseline.
+  // Using this to position SVG text baselines explicitly avoids the extra descender-space
+  // that dominantBaseline="central" includes, which causes perceived extra whitespace
+  // above cap-height text (e.g. "PayU 2,700 txn" has no descenders).
+  const capHeightRatio = 0.72;
+
+  // Accurate text width measurement via Canvas API — avoids flat px-per-char estimates
+  // that over/under-shoot on proportional fonts (narrow: i,l,1 / wide: W,M)
+  const measureText = useCallback(
+    (text: string, weight: number | string): number => {
```
swapnil-kr1: `CommonChartComponents/utils.ts:calculateTextWidth` has the same pattern (creates a canvas per call without hoisting). Please re-use

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,687 @@
...
+      </Text>
+    </div>
+  );
+}
+
+// ─── Indian number humanizer ─────────────────────────────────────────────────
+// Truncates (never rounds up) to avoid overstating values.
+// Examples: 2550→2.5k, 17100→17.1k, 124500→1.24L, 1000000→10L, 15000000→1.5Cr
+
+function humanizeIndian(value: number): string {
```
swapnil-kr1, line 119: **`humanizeIndian` hardcodes Indian number formatting (k/L/Cr) in a design system component**

Blade is a design system — teams outside India or non-INR contexts get confusing labels ("1.5Cr" means nothing for USD).

Suggested fix: Accept a `formatValue?: (value: number) => string` prop with `humanizeIndian` as the default:
```ts
formatValue={v => Intl.NumberFormat('en-US', { notation: 'compact' }).format(v)}
```
At minimum, accept a `numberSystem?: 'indian' | 'international'` prop.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,687 @@
+import React, { useCallback, useMemo, useState } from 'react';
+import { Sankey, Tooltip, ResponsiveContainer } from 'recharts';
```
swapnil-kr1: **Consider splitting this 687-line file**

This file handles: tooltip, number formatting, color resolution, node rendering, link rendering, hover state, label chip geometry, events, margin calc, BFS depth, and main component.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,687 @@
...
+    | undefined;
+  if (!item) return null;
+
+  const label = item.name ?? '';
+  const value = item.value != null ? item.value.toLocaleString() : '';
+  const content = `${label}: ${value}${labelUnit ? ` ${labelUnit}` : ''}`;
+
+  return (
+    <div
+      style={{
```
swapnil-kr1, line 95: **Tooltip uses wrong semantic color token**

`theme.colors.surface.icon.staticBlack.normal` is an **icon** color token being used as a **surface** background — use the correct semantic token for tooltip backgrounds.

Also note: while this uses a raw `<div>` wrapper, the existing chart tooltip in `CommonChartComponents.web.tsx` follows the same pattern (raw `<div>` outer wrapper with Blade `Box`/`Text` for inner content), so this is consistent with the codebase. The color token issue is the main thing to fix.

packages/blade/src/components/Charts/SankeyChart/__tests__/SankeyChart.web.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+      <SankeyChart
+        data={{
+          nodes,
+          links: [...links, { source: 'nonexistent', target: 'also-missing', value: 999 }],
+        }}
+      />,
+    );
+    expect(container.querySelector('svg')).toBeTruthy();
+  });
+});
```
swapnil-kr1, line 335: **Test coverage gaps**

Missing:
1. `showPercentage={false}` — untested
2. `showLabelChip={false}` — untested
3. Hover dimming — core UX, no test
4. `linkColorOverride` — untested
5. Per-node `color` override — untested
6. Responsive scroll wrapper — untested

Color test only checks `fills.size === 1` — doesn't verify actual resolved value.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,687 @@
...
+  const fontFamily = theme.typography.fonts.family.text;
+  // Cap-height ratio for Inter — uppercase glyphs occupy ~72% of font-size above the baseline.
+  // Using this to position SVG text baselines explicitly avoids the extra descender-space
+  // that dominantBaseline="central" includes, which causes perceived extra whitespace
+  // above cap-height text (e.g. "PayU 2,700 txn" has no descenders).
+  const capHeightRatio = 0.72;
+
+  // Accurate text width measurement via Canvas API — avoids flat px-per-char estimates
+  // that over/under-shoot on proportional fonts (narrow: i,l,1 / wide: W,M)
+  const measureText = useCallback(
```
swapnil-kr1, line 435: **`measureText` duplicates existing `calculateTextWidth` from CommonChartComponents**

`CommonChartComponents/utils.ts` already has a canvas-based `calculateTextWidth()` utility used by other chart tooltips. This PR creates a second, independent canvas text measurement implementation.

Consider reusing or extending the shared utility rather than maintaining two separate canvas measurement functions with the same underlying approach.

packages/blade/src/components/Charts/_decisions/sankey-decisions.md
swapnil-kr1, line 1: We don’t have separate decision docs for each graph component. Please follow the same pattern we’ve used for the other graph components.

packages/blade/src/components/Charts/_decisions/sankey-decisions.md
```diff
@@ -0,0 +1,140 @@
...
+**Component:** `SankeyChart`  
+**Status:** Approved  
+**Authors:** Debabrata Malik (Product Designer, MoneySaver)  
+**Last updated:** 2026-05-21
+
+---
+
+## References
+
+- **Design spec:** Section 3.7 — Volume Distribution (Sankey Chart) in the MoneySaver Dashboard spec
```
swapnil-kr1: We don’t add these details to the decision doc. Please follow the same pattern we’ve used for the other graphs.

packages/blade/src/components/Charts/_decisions/sankey-decisions.md
```diff
@@ -0,0 +1,140 @@
...
+---
+
+## 3. Component API design
+
+### 3a. Single component vs composition API
+
+BarChart uses a **composition API** (`<ChartBarWrapper>` + `<ChartBar>` + `<ChartXAxis>` …) because recharts itself is compositional and each sub-component maps to a recharts primitive. Since SankeyChart's visual output is a single connected diagram, a composition API would add indirection without benefit.
+
+**Decision:** Single `<SankeyChart data={} />` prop API. This is simpler to use and sufficient for the current use cases.
+
```
swapnil-kr1: **The "single connected diagram" argument doesn't justify breaking the composition pattern**

DonutChart is also a single connected diagram — one pie — yet it still uses composition: `<ChartDonutWrapper>` + `<ChartDonut>` + `<ChartDonutCell>`. So "single visual output = single component API" isn't a principle Blade actually follows.

This is a design system. Consistency across the charts layer matters more than saving one wrapper component. Every Blade chart today follows the composition pattern — `Wrapper` + children (`Tooltip`, `Legend`, series elements). Consumers learn it once and apply it everywhere. A single-prop `<SankeyChart data={} />` breaks that mental model.

It also creates a practical problem: every future customization (custom tooltip content, different label rendering, legend, click handlers) must be added as yet another prop on `<SankeyChart>`, leading to prop bloat. With composition, consumers opt into sub-features by rendering the corresponding child.

At minimum, a `<SankeyChartWrapper>` + composable children (tooltip, legend) would align with the existing pattern. Recharts' `<Sankey>` itself supports composable children (`<Tooltip>`, custom node/link renderers) — so there is a natural composition boundary here.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,711 @@
...
+
+type SankeyChartContextType = {
+  height: number;
+  showTooltip: boolean;
+  nodeColorOverride?: ChartsCategoricalColorToken;
+  linkColorOverride?: ChartsCategoricalColorToken;
+  defaultColorTokens: ChartsCategoricalColorToken[];
+};
+
+const SankeyChartContext = createContext<SankeyChartContextType>({
```
swapnil-kr1: **`SankeyChartContext` default value provides `defaultColorTokens: []` — rendering `ChartSankey` outside `ChartSankeyWrapper` will silently produce index-into-empty-array bugs.**

Consider either:
1. Throwing a descriptive Blade error when `ChartSankey` detects it's outside the wrapper (check for empty `defaultColorTokens`)
2. Using `null` as the default context value and asserting non-null in the consumer

This matches the defensive pattern used by other Blade chart components.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,711 @@
...
+        border: `${theme.border.width.thin}px solid ${theme.colors.surface.border.gray.muted}`,
+        padding: theme.spacing[4],
+        pointerEvents: 'none',
+        whiteSpace: 'nowrap',
+        zIndex: TOOLTIP_Z_INDEX,
+        boxShadow: castWebType(theme.elevation.highRaised),
+      }}
+    >
+      <Text size="small" weight="regular" color="surface.text.staticWhite.normal">
+        {content}
```
swapnil-kr1, line 109: **`humanizeIndian` truncates instead of rounding — is this intentional?**

`Math.floor((value / 1_00_00_000) * 100) / 100` means `14999999` → `1.49Cr` instead of `1.5Cr`. The PR description mentions this is intentional to avoid overstating values — but this can be confusing for users who expect standard rounding.

Suggestion: At minimum, document this behavior prominently in the `formatValue` prop JSDoc so consumers know the default truncates.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,711 @@
...
+    return `${parseFloat(v.toFixed(2))}Cr`;
+  }
+  if (value >= 1_00_000) {
+    const v = Math.floor((value / 1_00_000) * 100) / 100;
+    return `${parseFloat(v.toFixed(2))}L`;
+  }
+  if (value >= 1_000) {
+    const v = Math.floor((value / 1_000) * 10) / 10;
+    return `${parseFloat(v.toFixed(1))}k`;
+  }
```
swapnil-kr1, line 131: **perf: `defaultColorTokens` is filtered inline on every render, creating a new array reference each time.**

The `.filter()` creates a new array on every render, which will cause all downstream `useMemo`/`useCallback` that depend on `defaultColorTokens` to recompute. Wrap this in a `useMemo` or move the filter into the hook.

```ts
const allTokens = useChartsColorTheme({ colorTheme: 'categorical' });
const defaultColorTokens = useMemo(
  () => allTokens.filter((t) => t !== 'data.background.categorical.gray.faint'),
  [allTokens],
);
```

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,711 @@
...
+  // Categorical palette — same filter used in DonutChart (gray.faint is near-white)
+  const defaultColorTokens = useChartsColorTheme({ colorTheme: 'categorical' }).filter(
+    (t) => t !== 'data.background.categorical.gray.faint',
+  );
+
+  // Extract data from ChartSankey child — same pattern ChartDonutWrapper uses
+  // to read colorTheme/radius from ChartDonut props.
+  const data = useMemo(() => {
+    const sankeyChild = React.Children.toArray(children).find(
+      (child) => isValidElement(child) && getComponentId(child) === componentIds.ChartSankey,
```
swapnil-kr1: **`dataColorMapping` has a type cast via `as any` to force the color token type.**

```ts
colorToken: (override ?? undefined) as any,
```

This bypasses type safety. Can you use a proper type assertion or adjust the `DataColorMapping` type to accept `ChartsCategoricalColorToken | undefined`? The `as any` masks potential type mismatches the compiler should catch.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,711 @@
...
+  // Read wrapper-level config from private context
+  const { showTooltip, nodeColorOverride, linkColorOverride, defaultColorTokens } = useContext(
+    SankeyChartContext,
+  );
+
+  // ── Theme tokens ──────────────────────────────────────────────────────────
+  const { theme } = useTheme();
+
+  const resolveColor = useCallback(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
swapnil-kr1, line 419: **`capHeightRatio = 0.72` is a magic number hardcoded for Inter font.**

If Blade ever changes its font family (or consumers use white-labelled themes), this ratio will produce incorrect vertical alignment. Consider moving this to `tokens.ts` with a descriptive name and a comment about how it was derived, so it's flagged during any font migration.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,711 @@
...
+    const incomingCount = new Map<string, number>(data.nodes.map((n) => [n.id, 0]));
+    data.links.forEach((l) => incomingCount.set(l.target, (incomingCount.get(l.target) ?? 0) + 1));
+    const outgoing = new Map<string, string[]>(data.nodes.map((n) => [n.id, []]));
+    data.links.forEach((l) => outgoing.get(l.source)?.push(l.target));
+    const depthOf = new Map<string, number>();
+    const queue = data.nodes.filter((n) => incomingCount.get(n.id) === 0).map((n) => n.id);
+    queue.forEach((id) => depthOf.set(id, 0));
+    for (let i = 0; i < queue.length; i++) {
+      const id = queue[i];
+      const d = depthOf.get(id) ?? 0;
```
swapnil-kr1, line 504: **`renderNode` callback is ~120 lines long with 16+ dependencies.**

This is a very large `useCallback`. Consider extracting the label rendering into smaller helper functions (e.g., `renderChipLabel`, `renderPlainTextLabel`, `renderWrappedLabel`). This would:
1. Make the render logic easier to follow and test individually
2. Reduce the cognitive load when reading the node renderer
3. Make the dependency arrays smaller and more obvious

Look at how other Blade charts keep their custom renderers lean — they typically delegate to smaller composable pieces.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,711 @@
...
+                        x={chipX + CHIP_PAD_X}
+                        y={
+                          chipY +
+                          theme.spacing[3] +
+                          fontSize +
+                          lineGap +
+                          (fontSize * (1 + capHeightRatio)) / 2
+                        }
+                        fill={labelValueColor}
+                      >
```
swapnil-kr1: **perf: `renderLink` uses `data.nodes.indexOf(srcNode)` for color index — O(n) scan per link.**

You already have `nodeIdToIndex` map. Use it instead:

```ts
// Instead of:
defaultColorTokens[data.nodes.indexOf(srcNode) % defaultColorTokens.length]
// Use:
defaultColorTokens[srcNodeIndex % defaultColorTokens.length]
```

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,711 @@
...
+            </g>
+          )}
+        </g>
+      );
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    // CHIP_MAX_WIDTH is a module-level constant — stable, intentionally omitted
+    [
+      data.nodes,
+      nodeColorOverride,
```
swapnil-kr1, line 650: **Event handlers are cast through `any` twice (`onMouseEnter` and `onClick`).**

If Recharts' types are genuinely broken here, add a brief comment explaining *which* Recharts type issue forces the cast so future maintainers know when it can be removed (e.g., after a Recharts type fix). Without the comment, someone will try to remove `as any` and waste time rediscovering the issue.

packages/blade/src/components/Charts/SankeyChart/tokens.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Maximum width for a label chip in px.
+ * Fits most typical node names + humanized value + percentage on one line.
+ * Content wider than (CHIP_MAX_WIDTH - horizontal padding) wraps to two lines.
+ */
+export const CHIP_MAX_WIDTH = 160;
+/** Approximate px-per-character for chip width estimation (12px Inter, weight 600) */
+export const CHIP_PX_PER_CHAR = 6;
+/**
+ * Fixed pixel budget reserved for the humanized value + percentage part of a label chip,
```
swapnil-kr1, line 22: **`CHIP_PX_PER_CHAR` is exported but never referenced anywhere in the codebase.**

If it was replaced by the `measureText` canvas-based approach, remove it to avoid dead code.

packages/blade/src/components/Charts/CommonChartComponents/utils.ts
```diff
@@ -1,16 +1,30 @@
...
+ *                           SankeyChart uses size[75] for node labels.
+ * @param options.fontWeight Override the default font weight (theme.typography.fonts.weight.medium).
+ *                           SankeyChart uses semibold/regular depending on label part.
+ */
+const calculateTextWidth = (
+  text: string,
+  theme: Theme,
+  options?: { fontSize?: number; fontWeight?: number | string; rawMeasure?: boolean },
+): { width: number; displayText: string } => {
   // Create a temporary canvas to measure text
```
swapnil-kr1, line 22: **nit: The `rawMeasure` option name is ambiguous.**

It could mean "measure raw HTML" or "raw/unprocessed result". Consider renaming to `skipPadding` or `exactWidth` which clearly communicates what it does: skip the MIN_WIDTH floor and PADDING_HORIZONTAL addition.

packages/blade/src/components/Charts/SankeyChart/__tests__/SankeyChart.web.test.tsx
```diff
@@ -0,0 +1,322 @@
...
+
+  it('calls onLinkClick with the correct link and index when a link path is clicked', () => {
+    const handleLinkClick = jest.fn();
+    const { container } = renderSankey({}, { onLinkClick: handleLinkClick });
+
+    const paths = container.querySelectorAll('svg path');
+    fireEvent.click(paths[0]);
+
+    expect(handleLinkClick).toHaveBeenCalledTimes(1);
+    expect(handleLinkClick).toHaveBeenCalledWith(
```
swapnil-kr1, line 177: **`onNodeClick` test only validates callback shape, not correctness.**

```ts
expect(handleNodeClick).toHaveBeenCalledWith(
  expect.objectContaining({ id: expect.any(String), name: expect.any(String) }),
  expect.any(Number),
);
```

Since you click `rects[0]`, you know which node that is. Assert the actual `id` and `name` values match the expected node. `expect.any(String)` doesn't validate correctness — it only validates shape.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.stories.tsx
```diff
@@ -0,0 +1,339 @@
...
+
+// Fixed dataset used by non-configurable stories
+const { nodes: paymentNodes, links: paymentLinks } = generateChartData([1, 4, 3, 2]);
+
+// ─── Stories ──────────────────────────────────────────────────────────────────
+
+export const DefaultSankeyChart: StoryFn<StoryProps> = ({
+  height = 420,
+  showTooltip = true,
+  showLabels = true,
```
swapnil-kr1, line 252: **Default story doesn't demonstrate `onNodeClick`/`onLinkClick` interactions.**

Consider adding `action()` handlers from `@storybook/addon-actions` so click events appear in the Actions panel. This helps consumers understand the callback API without reading code.

```ts
import { action } from '@storybook/addon-actions';
// ...
onNodeClick={action('onNodeClick')}
onLinkClick={action('onLinkClick')}
```

packages/blade/src/components/Charts/_decisions/decisions.md
```diff
@@ -657,7 +658,68 @@ import {
```
swapnil-kr1, line 659: **Section numbering collision in the existing doc.**

The new Sankey section is §3.7, but the section that follows ("Axis and Grid Components") was renumbered to §3.8 while its sub-section still says §3.7.1 (ReferenceLine). Should be updated to §3.8.1 for consistency.

.changeset/add-sankey-chart.md
```diff
@@ -0,0 +1,11 @@
+---
+'@razorpay/blade': minor
+---
+
+feat(SankeyChart): add SankeyChart component
+
+Adds a new SankeyChart component for visualising flow and proportional
+relationships between nodes. Built on Recharts' native Sankey component.
+Supports Blade color tokens, hover interactions, tooltips, label chips,
+responsive resize, and click callbacks.
```
swapnil-kr1, line 8: Remove this only:

`feat(SankeyChart): add SankeyChart component`

That should be enough.

packages/blade/src/components/Charts/SankeyChart/types.ts
```diff
@@ -0,0 +1,76 @@
...
+   */
+  width?: number | string;
+  /** Show a tooltip when hovering over a node or link ribbon. Default: true */
+  showTooltip?: boolean;
+  /** Override all node bar colors with a single Blade token */
+  nodeColorOverride?: ChartsCategoricalColorToken;
+  /** Override all link ribbon colors with a single Blade token */
+  linkColorOverride?: ChartsCategoricalColorToken;
+} & TestID &
+  DataAnalyticsAttribute;
```
swapnil-kr1: **`ChartSankeyWrapperProps` should extend `BoxProps` like all other chart wrappers**

All existing chart wrappers extend `BoxProps`:
- `ChartDonutWrapperProps` → `& BoxProps` (DonutChart/types.ts:63)
- `ChartBarWrapperProps` → `& BoxProps`
- `ChartLineWrapperProps` → `& BoxProps`
- `ChartAreaWrapperProps` → `& BoxProps`

SankeyChart defines a standalone interface with only `children`, `height`, `width`, `showTooltip`, color overrides, and `testID`. Consumers lose access to layout props like `margin`, `padding`, and the standard `width`/`height` string contract from BoxProps.

**Suggested fix:**
```ts
type ChartSankeyWrapperProps = {
  children: React.ReactElement;
  showTooltip?: boolean;
  nodeColorOverride?: ChartsCategoricalColorToken;
  linkColorOverride?: ChartsCategoricalColorToken;
} & TestID & DataAnalyticsAttribute & BoxProps;
```

Remove the explicit `height` (number) and `width` (number | string) props — they should come from BoxProps as strings (e.g. `"400px"`, `"100%"`), matching every other chart wrapper.

packages/blade/src/components/Charts/SankeyChart/types.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export type ChartSankeyProps = {
+  /** Flat node list + directed flow connections, referenced by node id */
+  data: {
+    nodes: SankeyDataNode[];
+    links: SankeyDataLink[];
+  };
```
swapnil-kr1, line 54: **Missing `colorTheme` prop — all other chart components have it**

Every existing chart component accepts a `colorTheme` prop (currently only `categorical` is supported, with `sequential` planned):
- `ChartDonutProps` → `colorTheme?: ColorTheme` (DonutChart/types.ts:45)
- `ChartBarWrapperProps` → `colorTheme`
- `ChartLineWrapperProps` → `colorTheme`
- `ChartAreaWrapperProps` → `colorTheme`

Sankey hardcodes `useChartsColorTheme({ colorTheme: 'categorical' })` (SankeyChart.web.tsx:138) without exposing this as a prop. When sequential color themes ship, SankeyChart will be the odd one out.

Either add `colorTheme?: ColorTheme` on `ChartSankeyWrapperProps` or `ChartSankeyProps` (DonutChart puts it on the inner component) and thread it through to `useChartsColorTheme`.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,834 @@
...
+  const value = item.value != null ? item.value.toLocaleString() : '';
+  const content = `${label}: ${value}${labelUnit ? ` ${labelUnit}` : ''}`;
+
+  return (
+    <div
+      style={{
+        // surface.icon.staticBlack.normal is the token used by CommonChartComponents tooltip.
+        // The icon→surface semantic mismatch is a known issue to fix system-wide separately.
+        backgroundColor: theme.colors.surface.icon.staticBlack.normal,
+        borderRadius: theme.border.radius.large,
```
swapnil-kr1, line 99: **Tooltip is fully custom instead of using the shared `ChartTooltip` / `CommonChartComponentsContext` pattern**

In Bar/Line/Area charts, tooltip rendering goes through the shared `<ChartTooltip>` component (from `CommonChartComponents`) which uses `CommonChartComponentsContext` for consistent color resolution, styling, and behavior across all charts.

DonutChart disables Recharts tooltip entirely (`tooltipType="none"`) but still wires through the shared context.

SankeyChart builds its own `SankeyTooltipContent` component from scratch (lines 74–100), with:
- Its own background color token resolution
- Its own border/shadow styling
- Its own content formatting
- A `showTooltip` boolean on the wrapper instead of a composable `<ChartTooltip>` child

This means if tooltip styling is updated system-wide (e.g. new background token, new shadow, new border radius), SankeyChart's tooltip won't get the update.

**Suggested approach:** Either reuse `<ChartTooltip>` with a custom content renderer, or at minimum extract the tooltip styling tokens from a shared source so they stay in sync.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
```diff
@@ -0,0 +1,834 @@
...
+// plain callback. This module-level alias is cast through `any` at the call site.
+type SankeyEventHandler = (item: NodeProps | LinkProps, type: 'node' | 'link') => void;
+
+// ─── Tooltip content ──────────────────────────────────────────────────────────
+
+type SankeyTooltipContentProps = {
+  active?: boolean;
+  payload?: TooltipContentProps<number, string>['payload'];
+  labelUnit?: string;
+};
```
swapnil-kr1: Lets move this interface to type.ts file

packages/blade/src/components/Charts/SankeyChart/SankeyChart.web.tsx
debabratama1ik: Partially addressed — extracted `renderChipLabel` and `renderPlainTextLabel` as pure module-level helpers, which reduced `renderNode` from ~120 lines to ~50 lines. The file is still ~700 lines total, but each section is a single cohesive responsibility (context, hover state, tooltip, humanizer, wrapper, label helpers, inner component). Splitting further (e.g. a separate `SankeyTooltipContent.tsx`) would scatter context that is easier to follow in one place for a chart component of this size. Happy to split further if you feel strongly — just flagging the trade-off.

saurabhdaware (approved):
Just reviewed the UI and Public API since L1 is already done. Looks good to me. Merging this

PR #3421, opened 2026-05-25
fix(native): transparency issue in tooltip for native

shivam-rzpay:
## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/.changeset/long-hoops-write.md
+++ b/.changeset/long-hoops-write.md
@@ -2,4 +2,4 @@
 "@razorpay/blade": patch
 ---
 
-fixed transparency issue in tooltip for native
+fix(native): transparency issue in tooltip for native
--- a/.changeset/long-hoops-write.md
+++ b/.changeset/long-hoops-write.md
@@ -1,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fixed transparency issue in tooltip for native
--- a/packages/blade/src/components/PopupArrow/PopupArrow.native.tsx
+++ b/packages/blade/src/components/PopupArrow/PopupArrow.native.tsx
@@ -101,7 +101,12 @@ const PopupArrow = React.forwardRef<SvgProps, PopupArrowProps>(
           viewBox={`0 0 ${width} ${width}`}
           styles={newStyles}
         >
-          <Path fill="none" stroke={strokeColor} strokeWidth={`${strokeWidth}px`} d={dValue} />
+          {strokeColor ? (
+            <Path fill="none" stroke={strokeColor} strokeWidth={`${strokeWidth}px`} d={dValue} />
+          ) : null}
+          {/* Opaque backing: fillColor is semi-transparent and relies on backdropFilter on web.
+              On native we render the surface color beneath so content doesn't bleed through. */}
+          <Path fill={theme.colors.surface.background.gray.intense} stroke="none" d={dValue} />
           <Path fill={fillColor} stroke="none" d={dValue} />
         </StyledSvg>
       </View>
--- a/packages/blade/src/components/Tooltip/Tooltip.native.tsx
+++ b/packages/blade/src/components/Tooltip/Tooltip.native.tsx
@@ -125,7 +125,9 @@ const Tooltip = ({
                   width={ARROW_WIDTH}
                   height={ARROW_HEIGHT}
                   fillColor={theme.colors.popup.background.gray.intense}
-                  strokeColor={theme.colors.popup.border.gray.intense}
+                  strokeColor={
+                    colorScheme === 'dark' ? theme.colors.popup.border.gray.intense : undefined
+                  }
                 />
               }
             >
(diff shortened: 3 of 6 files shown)


packages/blade/src/components/Tooltip/TooltipContentWrapper.native.tsx
```diff
@@ -59,8 +60,49 @@ const TooltipContentWrapper = React.forwardRef<View, TooltipContentWrapperProps>
...
+        {layout && (
+          <View
+            style={{
+              position: 'absolute',
+              left: layout.x,
+              top: layout.y,
+              width: layout.width,
+              height: layout.height,
+              backgroundColor: theme.colors.surface.background.gray.intense,
+              borderRadius: theme.border.radius.medium,
```
swapnil-kr1, line 109: **Hardcoded `borderRadius` on backing View**: The backing View uses `theme.border.radius.medium`, but `StyledTooltipContentWrapper` gets its border radius from `getTooltipContentWrapperStyles()`. If that function ever returns a different radius (e.g. based on size prop or theme variant), the backing and the tooltip will mismatch.

Suggestion: extract the border radius from the same `styles` object that `StyledTooltipContentWrapper` uses to keep them in sync.

```tsx
borderRadius: styles.borderRadius ?? theme.border.radius.medium,
```

packages/blade/src/components/Tooltip/Tooltip.native.tsx
```diff
@@ -125,7 +125,9 @@ const Tooltip = ({
                   width={ARROW_WIDTH}
                   height={ARROW_HEIGHT}
                   fillColor={theme.colors.popup.background.gray.intense}
-                  strokeColor={theme.colors.popup.border.gray.intense}
+                  strokeColor={
+                    colorScheme === 'dark' ? theme.colors.popup.border.gray.intense : undefined
```
swapnil-kr1, line 129: **`strokeColor` conditional on `colorScheme`**: Setting `strokeColor` to `undefined` for light mode removes the arrow border entirely in light theme. Is this intentional, or is this a workaround for a specific visual issue (e.g. the stroke looking bad against the opaque backing)?

If its a workaround, a comment explaining *why* light mode skips the stroke would help future maintainers.

packages/blade/src/components/PopupArrow/PopupArrow.native.tsx
```diff
@@ -101,7 +101,12 @@ const PopupArrow = React.forwardRef<SvgProps, PopupArrowProps>(
           viewBox={`0 0 ${width} ${width}`}
           styles={newStyles}
         >
-          <Path fill="none" stroke={strokeColor} strokeWidth={`${strokeWidth}px`} d={dValue} />
+          {strokeColor ? (
+            <Path fill="none" stroke={strokeColor} strokeWidth={`${strokeWidth}px`} d={dValue} />
```
swapnil-kr1, line 105: **Null check on `strokeColor` may hide bugs**: The original code always rendered the stroke path. Now, if `strokeColor` is falsy (not just `undefined` but also empty string, `null`, etc.), the stroke is silently skipped. Consider using a more explicit check or add a brief comment explaining why skipping rendering for all falsy values is intended.

.changeset/long-hoops-write.md
```diff
@@ -0,0 +1,5 @@
+---
+"@razorpay/blade": patch
+---
+
+fixed transparency issue in tooltip for native
```
swapnil-kr1: Can we use:
feat(native): syntax

packages/blade/src/components/Tooltip/Tooltip.native.tsx
shivam-rzpay: yes this is intentional, we dont want the border in light mode at all.

.changeset/long-hoops-write.md
shivam-rzpay: ack

packages/blade/src/components/PopupArrow/PopupArrow.native.tsx
shivam-rzpay: the strokeColor prop is typed as string | undefined and only ever receives either a theme color string or undefined from **Tooltip.native.tsx**. An empty string or null is impossible from our usage.

packages/blade/src/components/Tooltip/TooltipContentWrapper.native.tsx
shivam-rzpay: The suggestion to pull it from styles won't work here since styles contains floating position data, not computed styled values.

This comment added
Short-term workaround: RN doesn't support backdropFilter so we render an opaque sibling View behind the tooltip to prevent content bleeding through the semi-transparent popup.background.gray.intense background.

TODO: revisit this when upgrading React Native — newer versions may support backdropFilter natively, making this backing View unnecessary.
            
NOTE: borderRadius must stay in sync with getTooltipContentWrapperStyles — if the radius token changes there, update it here too.

swapnil-kr1 (approved):

PR #3428, opened 2026-05-25
feat(native): Added BottomNav component in React Native

shivam-rzpay:
## Component Checklist

- [ ] Add KitchenSink Story
- [x] Add changeset


packages/blade/src/components/BottomNav/BottomNav.native.tsx
```diff
@@ -1,23 +1,196 @@
...
+    if (__DEV__) {
+      logger({
+        message: `Could not open the link "href=${href}"`,
+        moduleName: 'BottomNavItem',
+        type: 'warn',
+      });
+    }
+  }
+};
+
```
swapnil-kr1, line 34: **Concern — `styled-components/native` usage**: Other native components in Blade (e.g., Tabs) tend to use `StyleSheet.create()` or inline styles rather than `styled-components/native`. Is there a reason for choosing styled-components here? Consider aligning with existing native component patterns for consistency, unless the project is standardizing on styled-components for native.

packages/blade/src/components/BottomNav/BottomNav.native.tsx
```diff
@@ -1,23 +1,196 @@
...
+    </StyledBottomNav>
+  );
 };
 
-const BottomNavItem = (_props: BottomNavItemProps): React.ReactElement => {
-  throwBladeError({
-    message: 'BottomNavItem is not yet implemented for native',
-    moduleName: 'BottomNavItem',
-  });
+const BottomNav = assignWithoutSideEffects(React.forwardRef(_BottomNav), {
```
swapnil-kr1, line 130: **Issue — `onClick` type mismatch**: The types define `onClick` as `Platform.Select<{ web: React.MouseEventHandler; native: undefined }>`, meaning `onClick` should be `undefined` on native. But here it's accepted and invoked with a `GestureResponderEvent`. This breaks the type contract.

Consider either:
1. Adding an `onPress` prop to the native types via `Platform.Select`, or
2. Updating `BottomNavItemProps` to include a shared callback type that works on both platforms.

The `@ts-expect-error` comment is a symptom of this type mismatch.

packages/blade/src/components/BottomNav/BottomNav.native.tsx
```diff
@@ -1,23 +1,196 @@
...
+  return (
+    <StyledBottomNav
+      ref={ref as never}
+      backgroundColor={theme.colors.surface.background.gray.intense}
+      borderTopWidth={theme.border.width.thin}
+      borderTopColor={theme.colors.surface.border.gray.muted}
+      paddingHorizontal={theme.spacing[2]}
+      style={[shadowStyle, { zIndex, paddingBottom: insets.bottom }]}
+      {...makeAccessible({ role: 'navigation' })}
+      {...metaAttribute({
```
swapnil-kr1, line 119: **Concern — Unused destructured props**: `rel`, `target`, and `as` are destructured as `_rel`, `_target`, `_as` and never used.

- `as` could be useful for React Navigation integration — is that planned?
- `rel` and `target` have no native equivalent.

If intentionally unsupported on native, consider adding a `__DEV__` warning when they're passed, so consumers know these props are no-ops. Silent prop-swallowing can cause debugging headaches.

packages/blade/src/components/BottomNav/BottomNav.native.tsx
```diff
@@ -1,23 +1,196 @@
...
+const BottomNavItem = ({
+  title,
+  href,
+  rel: _rel,
+  target: _target,
+  as: _as,
+  isActive,
+  onClick,
+  icon: Icon,
+  testID,
```
swapnil-kr1, line 143: **Accessibility gap**: The web version sets `role="tab"` and `aria-current="page"` for active items. Here you have `makeAccessible({ current: isActive ? 'page' : undefined })` which handles `aria-current`, but the Pressable has no explicit `accessibilityRole`. Consider adding `role: 'tab'` to the `makeAccessible` call for screen reader parity with web.

packages/blade/src/components/BottomNav/BottomNav.native.tsx
```diff
@@ -1,23 +1,196 @@
...
-const BottomNav = (_props: BottomNavProps): React.ReactElement => {
-  throwBladeError({
-    message: 'BottomNav is not yet implemented for native',
-    moduleName: 'BottomNav',
-  });
+const openURL = async (href: string): Promise<void> => {
+  try {
+    const canOpen = await Linking.canOpenURL(href);
+    if (canOpen) {
+      await Linking.openURL(href);
```
swapnil-kr1, line 22: **Minor — `Linking.canOpenURL` gotcha**: On iOS, `canOpenURL` requires the scheme to be whitelisted in `Info.plist` under `LSApplicationQueriesSchemes`. If not listed, it returns `false` even for valid URLs. Consider falling back to `Linking.openURL` directly (which throws on failure instead of silently failing), or documenting this requirement for consumers.

packages/blade/src/components/BottomNav/BottomNav.native.tsx
shivam-rzpay: styled-components/native is the established pattern across Blade's native components — BaseButton (StyledBaseButton.native.tsx), BottomSheet, Tabs and BaseBox itself all use styled(View) / styled(Pressable).

packages/blade/src/components/BottomNav/BottomNav.native.tsx
shivam-rzpay: Done. Resolved

packages/blade/src/components/BottomNav/BottomNav.native.tsx
shivam-rzpay: Done

packages/blade/src/components/BottomNav/BottomNav.native.tsx
shivam-rzpay: done

packages/blade/src/components/BottomNav/BottomNav.native.tsx
shivam-rzpay: This is consistent with the existing Blade pattern — StyledBaseButton.native.tsx and StyledBaseLink.native.tsx use the exact same canOpenURL → openURL flow without special-casing the LSApplicationQueriesSchemes requirement. 
The href prop on BottomNavItem is intended for deep links and in-app URLs (e.g. myapp://route) rather than arbitrary external URLs, so consumers already control the scheme registration.

swapnil-kr1 (approved):

PR #3429, opened 2026-05-25
feat(native): Added Skeleton Loader for React Native

shivam-rzpay:
The Skeleton component's native pulse animation was broken due to two issues:

1. HSLA color format incompatibility - Reanimated's interpolateColor does not support hsla() strings. Added hslaToRgba converter to transform theme colors to rgba() before interpolation.

2. Animated styled-component conflict - Animated.createAnimatedComponent wrapping a styled-component (BaseBox) prevented animated styles from being applied. Switched to BaseBox for layout with an inner Animated.View for the pulse animation.

Also aligned easing with web (standard instead of emphasized) and added proper fade-in delay before pulse starts.

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade/src/components/Skeleton/PulseAnimation.native.tsx
+++ b/packages/blade/src/components/Skeleton/PulseAnimation.native.tsx
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/restrict-plus-operands */
 import Animated, {
   cancelAnimation,
-  interpolateColor,
+  interpolate,
   useAnimatedStyle,
   useSharedValue,
   withDelay,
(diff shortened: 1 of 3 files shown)


packages/blade/src/components/Skeleton/PulseAnimation.native.tsx
```diff
@@ -2,81 +2,80 @@
...
-    };
-    progress.value = withRepeat(
-      withSequence(
-        withTiming(0, pulsatingAnimationTimingConfig),
-        withTiming(1, pulsatingAnimationTimingConfig),
-      ),
-      -1,
-      true,
+    fadeIn.value = withTiming(1, { duration: fadeInDuration, easing });
+
```
swapnil-kr1, line 55: **Concern: opacity-only animation may not match the intended color transition**

The old code used `interpolateColor` to transition between `gray.default` and `gray.highlighted` — a full color interpolation. The new approach sets a static `backgroundColor: grayDefault.base` and only animates **opacity** between the two alpha values.

This only produces the correct visual if the two theme colors differ **only** in their alpha channel and have the same h/s/l. If `gray.default` and `gray.highlighted` differ in lightness or saturation (which is common for skeleton shimmer effects), this will look wrong — the pulse will fade in/out rather than shifting between two distinct grays.

Can you confirm the two theme colors only differ in alpha? If they differ in h/s/l, you may need to layer two `Animated.View`s (one per color) and crossfade their opacities, or convert to rgba and use a working color interpolation approach.

shivam-rzpay: Yes they differ only in alpha. The reason for  static backgroundColor: grayDefault.base is because react native reanimated does not support direct hsla color, hence the the parser as well. 
Have confirmed it that the only difference between these two is alpha

swapnil-kr1 (approved):

PR #3434, opened 2026-05-28
feat(native): added phone number input component in native

shivam-rzpay:
## Component Checklist

- [ ] Add KitchenSink Story
- [x] Add changeset


swapnil-kr1 (commented):
## Review: `feat(native): added phone number input component in native`

Good progress on bringing PhoneNumberInput to native — the overall structure mirrors the web implementation well, which is great for maintainability. I have several concerns that should be addressed before merging.

---

### 🔴 Must Fix

**1. `react-native-svg` is not listed as a dependency**
`SvgUri` is imported from `react-native-svg` in both `ActionListItemAsset.native.tsx` and `CountrySelector.native.tsx`, but it's not added to `package.json` (neither as `dependencies` nor `peerDependencies`). This will break for consumers who don't already have it installed. It should likely be a `peerDependency` since RN apps typically manage their own native module versions.

**2. ActionListItemAsset type change is a breaking change**
The `src` prop type changed from `string | ImageSourcePropType` to `string | { uri: string }`. This drops support for `require('./image.png')` which returns a number in RN. Any consumer currently using local image assets with `ActionListItemAsset` on native will break. The changeset is marked as `minor`, but this is a **breaking** change. Either:
- Keep backward compatibility by supporting both `Image` and `SvgUri` based on the src type, or
- Mark this as a major/breaking change in the changeset.

**3. Hardcoded `id="phone-number-input"`**
In `PhoneNumberInput.native.tsx`, the `id` is hardcoded. If a consumer renders multiple `PhoneNumberInput` instances on the same screen, they'll share the same id. The web version likely has the same issue, but this should still be addressed — consider using `useId()` or deriving from `name` prop.

**4. No search/filter in the country BottomSheet**
The web version leverages `Dropdown` with `AutoComplete` for country selection, giving users a search experience. The native version renders a flat `ActionList` inside a `BottomSheet` with 200+ countries and no search capability. This is a significant UX regression on native. Consider adding a search input in the BottomSheetHeader or using a filterable list pattern.

---

### 🟡 Should Fix

**5. `@formatjs` polyfills add bundle weight**
Two new production dependencies (`@formatjs/intl-displaynames`, `@formatjs/intl-locale`) are added. These polyfills pull in transitive deps (`intl-getcanonicallocales`, `intl-supportedvaluesof`, `fast-memoize`). For a design system library, bundle size discipline matters. Questions:
- Is `Intl.DisplayNames` not available in the Hermes engine versions you target? If it is, you can skip the polyfill.
- If the polyfill is needed, should it be a `peerDependency` so consumers can control the version, or at minimum documented as required?

**6. Redundant conditional in BottomSheet rendering**
```tsx
{isBottomSheetOpen ? (
  <BottomSheet isOpen={isBottomSheetOpen} ...>
    <ActionList>
      {(isBottomSheetOpen ? countryData : []).map(...)}
    </ActionList>
  </BottomSheet>
) : null}
```
The inner `isBottomSheetOpen ? countryData : []` is always `countryData` here since the outer conditional already gates on `isBottomSheetOpen`. Remove the inner ternary.

**7. Missing native-specific tests**
No test file was added for the native PhoneNumberInput. While interaction tests are checked in the PR checklist, there's no evidence of native-specific test coverage. The existing snapshot tests only cover web. Consider adding at least basic render + country selection tests for native.

---

### 🟢 Nits / Minor

**8. PR description is empty** — The Description, Changes, and Additional Information sections are blank. For a feature of this scope, please fill these out with context on design decisions (why BottomSheet over a different pattern, etc.), screenshots/recordings of the native component, and any known limitations.

**9. Unchecked checklist items** — "Update Component Status Page", "Perform Manual Testing in Other Browsers", and "Add KitchenSink Story" are unchecked. If not applicable, note why. If applicable, please complete.

**10. `flexDirection: 'row'` addition to BaseMenuItem** — This is correct for native (where RN defaults to column), but adding it to the shared `.tsx` file means it also affects web snapshots. The snapshot diffs look harmless (web flex default is row anyway), but confirm this doesn't cause any visual regression on web ActionList/Menu items.

---

### Summary

The core implementation is solid and follows the web version's patterns. The main blockers are: the undeclared `react-native-svg` dependency, the breaking type change in `ActionListItemAsset`, and the missing search UX in the country selector. Addressing these will make this ready for a second pass.

packages/blade/src/components/ActionList/ActionListItemAsset.native.tsx
```diff
@@ -1,34 +1,27 @@
...
 
 type ActionListItemAssetProps = {
   /**
-   * Source of the image.
-   *
-   * Can either be a string URI or `require('/local/image')` in React Native
+   * Source of the image (SVG URI string).
    */
-  src: string | ImageSourcePropType;
+  src: string | { uri: string };
```
swapnil-kr1: 🔴 **This is a breaking change**

The `src` prop type changed from `string | ImageSourcePropType` to `string | { uri: string }`. This drops support for `require('./image.png')` (which returns a `number` in RN). Any consumer currently passing a local image asset to `ActionListItemAsset` on native will get a type error and runtime breakage.

Options:
1. Keep backward compat by detecting the src type and falling back to `<Image>` for non-string sources.
2. If you intentionally want to drop `Image` support, mark the changeset as a **major/breaking** change.

packages/blade/src/components/Input/PhoneNumberInput/CountrySelector.native.tsx
```diff
@@ -1,12 +1,136 @@
-import { throwBladeError } from '~utils/logger';
+import '@formatjs/intl-locale/polyfill';
```
swapnil-kr1: 🟡 **`@formatjs` polyfills add significant bundle weight**

Have we checked the diff of bundle weight

packages/blade/src/components/Input/PhoneNumberInput/CountrySelector.native.tsx
```diff
@@ -1,12 +1,136 @@
...
+          />
+        </BaseBox>
+      </TouchableOpacity>
+
+      {isBottomSheetOpen ? (
+        <BottomSheet isOpen={isBottomSheetOpen} onDismiss={() => setIsBottomSheetOpen(false)}>
+          <BottomSheetHeader title="Select A Country" />
+          <BottomSheetBody>
+            <ActionList isVirtualized={true}>
+              {(isBottomSheetOpen ? countryData : []).map((country) => (
```
swapnil-kr1: 🟡 **Redundant conditional**

This inner `isBottomSheetOpen ? countryData : []` is always `countryData` because the outer conditional on line 106 already gates on `isBottomSheetOpen`. Remove the inner ternary:

```suggestion
              {countryData.map((country) => (
```

packages/blade/src/components/Input/PhoneNumberInput/PhoneNumberInput.native.tsx
```diff
@@ -1,13 +1,222 @@
...
+      name,
+      value: currentValueRef.current,
+    });
+    inputRef.current?.focus();
+  };
+
+  return (
+    <BaseInput
+      ref={mergedRef}
+      id="phone-number-input"
```
swapnil-kr1: 🔴 **Hardcoded `id` will collide with multiple instances**

If a consumer renders two `PhoneNumberInput` components on the same screen, they'll share the same `id="phone-number-input"`. This can cause accessibility issues and unexpected behavior.

Consider using `React.useId()` or deriving the id from the `name` prop:
```tsx
id={name ?? `phone-number-input-${React.useId()}`}
```

packages/blade/src/components/BaseMenu/BaseMenuItem/BaseMenuItem.tsx
```diff
@@ -69,7 +69,13 @@ const _BaseMenuItem: React.ForwardRefRenderFunction<BladeElementRef, BaseMenuIte
         {children ? (
           children
         ) : (
-          <Box display="flex" alignItems="flex-start" width="100%" justifyContent="flex-start">
+          <Box
+            display="flex"
+            flexDirection="row"
```
swapnil-kr1, line 74: 🟢 **Confirm no web visual regression from this change**

Adding `flexDirection="row"` to this shared `.tsx` file is correct for native (where RN defaults to column), but it also changes the web output — as visible in the snapshot diffs adding `flex-direction: row` CSS. While web flex default is `row` anyway (so it should be a no-op visually), please confirm via Chromatic that no existing ActionList/Menu items are visually affected.

packages/blade/src/components/ActionList/ActionListItemAsset.native.tsx
shivam-rzpay: done

packages/blade/src/components/Input/PhoneNumberInput/CountrySelector.native.tsx
shivam-rzpay: Yes, and as discussed this increase in size will be temporary until we move to latest version of react native
 
 Web: 0 KB increase
  - The polyfill imports are in PhoneNumberInput.native.tsx — this file is never included in web builds (rollup builds .web.tsx for web).
  
  Native: ~200 KB raw / ~54 KB gzipped increase to the consumer app
  - The consumer app installs @formatjs as a transitive dependency of blade. Metro bundles it into the app's JS bundle. This cost is paid only once regardless of how many PhoneNumberInputs are rendered, and only if PhoneNumberInput is actually imported.

packages/blade/src/components/Input/PhoneNumberInput/CountrySelector.native.tsx
shivam-rzpay: done

packages/blade/src/components/Input/PhoneNumberInput/PhoneNumberInput.native.tsx
shivam-rzpay: done

swapnil-kr1 (dismissed):

swapnil-kr1 (approved):

PR #3436, opened 2026-05-29
fix(native): add expanding focus ring animation for all inputs

rohitr-raz:
## Description
before:

Screenshot 2026-05-29 at 4 38 58 PM

after:

https://github.com/user-attachments/assets/05c6c5e9-9fcb-447e-9d88-7805916d4db5

## Changes

Animates borderWidth and position outward (0 → 4px) on focus . Also thickens the border from thin to thick on focus.

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/Input/BaseInput/AnimatedBaseInputWrapper.native.tsx
```diff
@@ -118,53 +121,102 @@ const _AnimatedBaseInputWrapper: React.ForwardRefRenderFunction<
...
-        animatedBorderAndBackgroundStyle,
-      ]}
-      isDropdownTrigger={isDropdownTrigger}
-      numberOfLines={numberOfLines}
-      setShowAllTagsWithAnimation={setShowAllTagsWithAnimation}
-      {...rest}
-    >
-      {children}
-    </StyledBaseInputWrapper>
+    // Outer wrapper has no overflow clipping so the focus ring overlay can extend beyond input edges
```
saurabhdaware: Do you think it might be useful if we create a FocusRing component and wrap all this logic there and use that component here? will make the code more readable + we can use it again at other places if needed

packages/blade/src/components/Input/BaseInput/StyledBaseInput.native.tsx
```diff
@@ -142,14 +142,26 @@ const getRNInputStyles = (
...
+    ...(props.isTextArea || props.hasTags
+      ? {
+          height: getInputHeight({
+            isTextArea: props.isTextArea,
+            hasTags: props.hasTags,
+            numberOfLines: props.numberOfLines,
+            isDropdownTrigger: props.isDropdownTrigger,
+            size: props.$size,
+          }),
+        }
```
saurabhdaware, line 163: can you check if it shouldn't follow some logic like - `isTextArea || hasTags ? {/* set height with getInputHeight */} : {/* set default props */}`


Usually textarea and hasTags follow the same logic though not sure in this case. Right now we seem to have 2 blocks of if-else checks

rohitr-raz: one block would suffice but there's a regular text input in tag where we use it to create tags right , so if we don't add the padding top/bottom: 0 condition to it the text would look kind of squished . If there was only one block of if-else check then text input in tag would take default padding which is 12px and it won't look good

rohitr-raz: So if we add 2 blocks of if-else checks then hasTags would take both the padding condition and the getInputHeight as well

packages/blade/src/components/Input/BaseInput/AnimatedBaseInputWrapper.native.tsx
rohitr-raz: Agree on the readable part , but could you give some examples where we might need focus ring other than the input component??

saurabhdaware: I see TextArea excluded from logic here. I am assuming that's because the focus ring of TextArea is elsewhere so that one would have same focus ring as TextInput.

All our components on web have focus rings but in native I think it won't be needed in all. Few scenarios I can think of are -

- IconButton with tooltip might have focus ring. 
- Also Dropdown, Menu triggers etc might have (I don't think so we have rings there right now but just in case the usecase comes)

rohitr-raz: the exclusion of text area is regarding fixed baseInputHeight , because we would want to change the height if the lines keep increasing , the focus ring part for text area is the same and included with the rest of the Input types

Image

packages/blade/src/components/Input/BaseInput/AnimatedBaseInputWrapper.native.tsx
```diff
@@ -118,53 +121,102 @@ const _AnimatedBaseInputWrapper: React.ForwardRefRenderFunction<
...
+
+  const focusRingStyle = useAnimatedStyle(() => ({
+    top: -ringWidth.value,
+    left: -ringWidth.value,
+    right: -ringWidth.value,
+    bottom: -ringWidth.value,
+    borderRadius: inputBorderRadius + ringWidth.value,
+    borderWidth: ringWidth.value,
+  }));
+
```
swapnil-kr1: Can we move all of this logic into the native-specific file instead? We already have the native file created, even though it's currently empty.

It would be better to follow the same pattern we use for web, for example:
https://github.com/razorpay/blade/blob/master/packages/blade/src/utils/getFocusRingStyles/getFocusRingStyles.web.test.ts

packages/blade/src/components/Input/BaseInput/AnimatedBaseInputWrapper.native.tsx
```diff
@@ -118,53 +121,102 @@ const _AnimatedBaseInputWrapper: React.ForwardRefRenderFunction<
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ref={ref as any}
+        style={[
+          isDropdownTrigger && !isTextArea
+            ? {
+                ...maxHeightStyleObject,
+                ...animatedStyleObject,
+              }
+            : !isTextArea
+            ? { height: baseInputHeight[rest.size] }
```
swapnil-kr1, line 176: **Question**: The previous code only applied `maxHeightStyleObject` and `animatedStyleObject` for dropdown triggers. Now there's a new branch `!isTextArea ? { height: baseInputHeight[rest.size] }` — this sets a fixed height on all non-textarea, non-dropdown inputs. Was this intentional? Could this break inputs that previously relied on auto-sizing (e.g., inputs with tags that aren't dropdown triggers)? Please confirm.

rohitr-raz: !isTextArea ? { height: baseInputHeight[rest.size] }  this line was added to specify the height of the non-text area , non-dropdown inputs , even though intended if its height is content-driven (floating), the outer wrapper's height is also floating, making the top/left/right/bottom offsets on the ring unreliable.
Setting height: baseInputHeight[rest.size] explicitly on StyledBaseInputWrapper pins the outer wrapper's height to a
 known value, so the ring's -ringWidth offsets are always relative to a determinate box.

Regarding the tags that aren't dropdown triggers , any input that holds tags are always set with isDropdownTrigger=true.
 TextInput with isTaggedInput={true} explicitly passes isDropdownTrigger={isTaggedInput} to BaseInput.

packages/blade/src/components/Input/BaseInput/AnimatedBaseInputWrapper.native.tsx
rohitr-raz: done , moved it into the native file and used that focus ring wrapper component here.

swapnil-kr1 (commented):
Reply to this comment https://github.com/razorpay/blade/pull/3436/changes#r3393066315 else LGTM

swapnil-kr1 (approved):

PR #3451, opened 2026-06-05
fix(blade-svelte): adds definite loader and avatargroup to button

rohankokane-dev:
Invert the definite-loader layers so the button base stays its normal color while a rest/disabled-colored cover recedes over it. This removes the blink that occurred when the progress overlay unmounted at completion (the base was previously the disabled color and transitioned to normal).

The disabled token is translucent, so the receding cover is composited over an opaque surface backing to reproduce the true disabled appearance and keep the sweep visible against the normal-colored base.

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


rohankokane-dev:
## Code Review

### Overview

This PR adds two new features to `blade-svelte`'s Button:
1. **Definite loader** — a left-to-right CSS progress fill that recedes over a configurable `loadingTimer` ms, using an "inverted-layer" model to eliminate a previous blink on completion.
2. **AvatarGroup inside Button** — avatars rendered after button text, constrained to `large` size buttons.

The indefinite loader is also reworked from a `BaseSpinner` to a pure-CSS 3-dot animation.

---

### What's Good

- **Inverted-layer model is clever and correct.** Keeping the button base at its normal color and using a receding cover for the "disabled" state neatly eliminates the blink-on-completion bug. The explanation in comments is thorough.
- **Reduced-motion support** is handled correctly — both `dots-loader` and `progress-overlay-fill` animations are frozen under `prefers-reduced-motion: reduce`.
- **Accessibility is solid** — `aria-busy` is exposed during loading, and a visually-hidden live region announces start/stop state changes with `aria-live="polite"`.
- **`definiteLoadingKey` keying pattern** is a clean way to allow re-triggering the definite loader by changing `loadingTimer` without needing an explicit reset prop.
- **Types are well-documented** with JSDoc across `ButtonLoadingType`, `ButtonAvatar`, and all new props.
- **Stories coverage** is good — variants, sizes, colors, avatar group, and the edge case of avatars being ignored on sub-large sizes are all covered.

---

### Issues & Suggestions

**1. `definiteLoadingKey` does not reset when `loadingTimer` changes to the same value**

The key is `"definite:${loadingTimer}"`. If a consumer triggers loading, it completes, and then tries to trigger loading again with the same `loadingTimer` value, `definiteLoadingKey` stays the same and equals `completedDefiniteLoadingKey`, so `isDefiniteLoading` stays `false`. The button won't start loading again.

Consider incorporating a monotonically increasing counter or a separate boolean trigger prop to allow re-runs with identical durations.

**2. Live-region announcement only tracks `isLoading` (indefinite), not definite loading**

```ts
$effect(() => {
  const loading = isLoading;  // only indefinite loading
  ...
});
```

The `$effect` only reacts to `isLoading`, so there is no announcement when a definite loader starts or completes. The `onLoadingComplete` callback fires, but the live region stays silent for definite loading. Consider also announcing when `isDefiniteLoading` transitions.

**3. `.definite-loading` specificity hack is a maintenance risk**

```css
.definite-loading.definite-loading.definite-loading[disabled] { ... }
```

Tripling the class to win specificity over `[disabled]` rules works but is fragile — any future change to the specificity of the disabled rules could silently break this. A comment explains why, which helps, but it may be worth exploring a `data-` attribute selector or a dedicated loading-state CSS custom property as a more explicit override mechanism.

**4. `progress-overlay-fill` hardcoded surface backing color**

```css
.progress-overlay-fill {
  background-color: var(--surface-background-gray-intense);
  ...
}
```

This compositing approach is correct for reproducing the true disabled appearance and is well-commented. However, `--surface-background-gray-intense` is hardcoded to the page background assumption. If a button is rendered on a non-default surface (e.g., inside a card, modal, or colored container), the backing color will mismatch and the receding cover won't look right.

**5. Missing `loadingSpinnerClass` export cleanup**

`loadingSpinnerClass` is still exported from `button.ts` and re-exported in `index.ts` / `styles/index.ts`, but `BaseSpinner` is removed from `BaseButton.svelte` and the spinner class is no longer used. The export itself is harmless but leaves dead surface area in the public API.

**6. `ButtonAvatar.name` and `ButtonAvatar.alt` are both optional — no required identifier**

```ts
export type ButtonAvatar = {
  name?: string;
  src?: string;
  alt?: string;
};
```

In the template, the `{#each}` key falls back to `index` when both `src` and `name` are absent:
```svelte
{#each avatars as avatar, index (avatar.src ?? avatar.name ?? index)}
```

Two avatars with no `name` or `src` will be keyed by index, which can cause incorrect reconciliation if the array is reordered. Requiring at least one of `name` or `src` would be safer.

**7. PR checklist items are unchecked**

The component checklist is entirely unchecked — no changeset, no mention of cross-browser testing, no KitchenSink story, no interaction tests. A changeset is particularly important here since both `@razorpay/blade-core` and `@razorpay/blade-svelte` have public API additions.

---

### Minor Nits

- `sampleAvatars` in stories uses `https://picsum.photos/200` and `https://picsum.photos/300` which return different images on each load — use stable seeds (e.g. `picsum.photos/seed/abc/200`) for deterministic story snapshots.
- The `blade-button-live-region` class is defined in a scoped `<style>` block inside the component, while the rest of the button styles live in `button.module.css`. Minor consistency gap worth noting.

---

### Summary

The core implementation is solid — the inverted-layer model is well-reasoned, accessibility is properly handled, and the CSS-only loader approach is a good fit for Svelte. The main concerns are around the re-triggering limitation of `definiteLoadingKey`, the hardcoded surface backing color, and the missing changeset + checklist items.

saurabhdaware:
image

saurabhdaware (approved):

PR #3457, opened 2026-06-07
fix(rn): spark upgrade RN easy layout fixes

rohitr-raz:
## Summary

- Add missing `flexDirection="row"` to flex containers in `Avatar.stories`, `BaseMenuItem`, `TextInput`, and `SearchInput` — React Native defaults to `column` unlike web CSS which defaults to `row`
- Add `alignSelf="flex-start"` to `PopoverContent` footer wrapper so the button doesn't stretch full-width on native

## Changes

| File | Fix |
|------|-----|
| `Avatar/docs/Avatar.stories.tsx` | `flexDirection="row"` on all horizontal flex containers |
| `BaseMenu/BaseMenuItem/BaseMenuItem.tsx` | `flexDirection="row"` on leading icon + text row |
| `Input/TextInput/TextInput.tsx` | `flexDirection="row"` on clear button + divider rows |
| `Input/SearchInput/SearchInput.tsx` | `flexDirection="row"` on clear button + divider row |
| `Popover/PopoverContent.tsx` | `alignSelf="flex-start"` on footer wrapper |

## Why

React Native's `flexDirection` defaults to `"column"`, unlike web CSS which defaults to `"row"`. Components that relied on the web default were rendering stacked vertically on native instead of side-by-side.

## Test plan

- [ ] Avatar stories render avatars horizontally in RN Storybook
- [ ] Menu items show icon + text side by side on native
- [ ] TextInput / SearchInput clear button renders inline with divider
- [ ] Popover footer button does not stretch full width on native


packages/blade/src/components/Popover/PopoverContent.tsx
```diff
@@ -84,7 +84,7 @@ const PopoverContent = React.forwardRef<HTMLDivElement, PopoverContentProps>(
             <PopoverHeader title={title} titleLeading={titleLeading} />
             <BaseBox>{children}</BaseBox>
           </BaseBox>
-          {footer ? <BaseBox>{footer}</BaseBox> : null}
+          {footer ? <BaseBox alignSelf="flex-start">{footer}</BaseBox> : null}
```
swapnil-kr1: **Question**: `alignSelf="flex-start"` prevents the footer from stretching full-width. But isn't full-width footer a common pattern (e.g., a footer with a full-width CTA button)? This change means a footer button will only be as wide as its content on native. Was this validated against all existing Popover usages? If there's a Popover with a full-width action button, this would break it.

rohitr-raz: Yes it has been validated against all existing Popover usages , No existing Popover usage has a full-width CTA , so maybe if someone needs full-width in the future they can control that from their own component, not by relying on the wrapper stretching.

swapnil-kr1: Have you checked this on Web as well? Could you please verify for diff scenarios like:

* One action button
* Two action buttons inside footer

Just want to make sure the layout behaves correctly across both use cases.

swapnil-kr1: Same Bot also commented: https://github.com/razorpay/blade/pull/3457#discussion_r3404824632

swapnil-kr1 (dismissed):

swapnil-kr1 (approved):

PR #3458, opened 2026-06-07
feat(blade-react): update styling of components

rkdotxyz:
### Accordion
- `variant=filled` now uses the same elevated surface treatment as Card (elevation shadow, gradient overlays, inset border) via the shared `getSurfaceStyles()` utility
- Created platform-specific `AccordionSurface.web.tsx` and `AccordionSurface.native.tsx` following the Card pattern
- Removed horizontal margin on the expanded-state divider so it stretches edge-to-edge
- Added an AvatarGroup example to the "Accordion Item Header Variants" story

### Alert
- Added `primary` color support with both `subtle` and `intense` emphasis
- Introduced `AlertColors` type extending `FeedbackColors` with `'primary'`
- Added icon mapping and token handling for the primary color variant
- Added tests and snapshots for primary color rendering

### Avatar
- Added `density` prop to `AvatarGroup` (`compact` | `comfortable` | `normal`)
- Updated `AvatarGroup` for both web and native with density-based spacing tokens
- Extended `avatarTokens` with density spacing values
- Updated `StyledAvatarGroup` to support the new density prop

### Collapsible
- Added new `CollapsibleText` component (web + native) for collapsible text content
- Updated `Collapsible` types with new props for text collapsing behavior
- Added stories and tests for `CollapsibleText`
- Registered new component ID in `metaConstants`


packages/blade/src/components/Alert/types.ts
```diff
@@ -3,10 +3,12 @@ import type { ReactNode } from 'react';
 import type { FeedbackColors, SubtleOrIntense } from '~tokens/theme/theme';
 import type { StyledPropsBlade } from '~components/Box/styledProps';
```
swapnil-kr1, line 5: **[usecase-critique] major**

Adding `primary` to Alert colors conflates brand identity with feedback semantics. Alert is designed around feedback intents (positive/negative/notice/information/neutral) which communicate system status. `primary` is a brand color, not a feedback state — it does not tell the user whether something succeeded, failed, or needs attention.

**Suggestion:** Consider whether this use case is better served by the existing `information` color (which already uses a blue token) or by a different component entirely. If there is a legitimate design need distinct from `information`, document the semantic difference — when should a consumer choose `primary` over `information`?

packages/blade/src/components/Collapsible/CollapsibleText.web.tsx
```diff
@@ -0,0 +1,67 @@
+import type { ReactElement } from 'react';
```
swapnil-kr1: **[usecase-critique] major**

`CollapsibleText` is functionally near-identical to `CollapsibleLink`. Both render clickable text with a chevron icon that toggles the collapsible body. The only differences are: (1) plain Text instead of BaseLink, and (2) a div with `role=button` instead of a semantic link.

**Suggestion:** Consider extending `CollapsibleLink` with a variant prop (e.g. `variant="text"`) that renders plain text styling instead of link styling, rather than introducing an entirely new component. This reduces API surface and avoids consumers needing to decide between `CollapsibleLink` and `CollapsibleText`.

packages/blade/src/components/Avatar/avatarTokens.ts
```diff
@@ -117,6 +117,23 @@ const avatarTopAddonOffsets = {
   },
 } as const;
```
swapnil-kr1, line 118: **[api-decision-critique] major**

The `avatarGroupDensityOverlapTokens` object defines tokens for `normal` and `comfortable` but is missing the `compact` density variant. The type allows `compact | normal | comfortable` but the token map only has two entries. The code guards with `density === 'compact' ? dimension / 2 : avatarGroupDensityOverlapTokens[density][size]` so it works, but the asymmetry is fragile.

**Suggestion:** Add a `compact` key to `avatarGroupDensityOverlapTokens` with appropriate overlap values (dimension/2 for each size) so the token map is exhaustive and the code doesn't need a special-case check.

packages/blade/src/components/Accordion/AccordionSurface.native.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from 'styled-components/native';
+import BaseBox from '~components/Box/BaseBox';
+import { makeBorderSize } from '~utils';
+
+type AccordionSurfaceProps = {
```
swapnil-kr1, line 7: **[code-quality-critique] major**

The native `AccordionSurface` accepts a `colorScheme` prop in its type but does not use it in the styled-component logic. The parent passes `colorScheme={colorScheme}` but it's dead code on native. The web version uses it with `getSurfaceStyles()` but native just applies a basic border.

**Suggestion:** Either remove the `colorScheme` prop from the native type definition, or implement equivalent color-scheme-aware styling for native. At minimum, document why the native version intentionally differs.

packages/blade/src/components/Alert/Alert.tsx
```diff
@@ -195,6 +197,8 @@ const _Alert = (
         color={
           emphasis === 'intense'
             ? 'surface.icon.staticWhite.normal'
+            : color === 'primary'
```
swapnil-kr1: **[code-quality-critique] major**

The icon color logic is now a nested 3-level ternary that's hard to read. Also, `color='primary'` with `emphasis='intense'` gets `staticWhite` (from the first branch) — verify the primary intense background is dark enough for white icon contrast.

**Suggestion:** Extract the icon color logic into a helper function with explicit cases for each color+emphasis combination for readability and maintainability.

packages/blade/src/components/Collapsible/CollapsibleText.web.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <BaseBox
+      display="flex"
+      flexDirection="row"
+      alignItems="center"
+      justifyContent="space-between"
+      width="100%"
+      cursor={isDisabled ? 'not-allowed' : 'pointer'}
+      opacity={isDisabled ? 0.4 : 1}
+      onClick={toggleIsExpanded}
+      {...makeAccessible({
```
swapnil-kr1, line 55: **[code-quality-critique] minor**

The component uses `onClick` on a `BaseBox` (div) with `role='button'` but does not handle keyboard interaction. Enter/Space key presses won't trigger the toggle, violating accessibility expectations for elements with `role="button"`.

**Suggestion:** Either use an actual `<button>` element (styled appropriately) or add `onKeyDown` handling that triggers `toggleIsExpanded` on Enter and Space key presses, and add `tabIndex={0}` to make it focusable.

packages/blade/src/components/Collapsible/CollapsibleText.native.tsx
```diff
@@ -0,0 +1,71 @@
...
+        flexDirection="row"
+        alignItems="center"
+        justifyContent="space-between"
+        width="100%"
+        opacity={isDisabled ? 0.4 : 1}
+        {...getStyledProps(rest)}
+        {...makeAnalyticsAttribute(rest)}
+      >
+        <Text size={size} weight={weight} color={color}>
+          {children}
```
swapnil-kr1, line 58: **[code-quality-critique] minor**

`CollapsibleChevronIcon` uses `color='currentColor'` on React Native. `currentColor` is a CSS concept that doesn't exist in React Native — it will likely render as default color or cause a warning.

**Suggestion:** Use an explicit color token for the chevron icon on native instead of `currentColor`.

packages/blade/src/components/Avatar/avatarTokens.ts
```diff
@@ -117,6 +117,23 @@ const avatarTopAddonOffsets = {
...
 
+const avatarGroupDensityOverlapTokens = {
+  normal: {
+    xsmall: 14,
```
swapnil-kr1, line 129: **[usecase-critique] minor**

The `normal` density has `xsmall: 14` while all other sizes are `16`. This inconsistency seems unintentional — is it a typo?

swapnil-kr1 (commented):
## PR Review Summary

CI is green (tests pass, lint passes), but Chromatic UI review and baseline acceptance are pending (68 visual changes). Key concerns flagged inline below.

packages/blade/src/components/Alert/types.ts
rkdotxyz: This is an intentional design decision for checkout flows. `primary` is semantically distinct from `information` — it represents brand-emphasis contexts (e.g., checkout confirmation) where the intent is promotional/brand-affirmative rather than informational-neutral. The existing `information` color carries a "system status" connotation that does not fit brand CTAs.

Keeping as-is. Happy to add a code comment documenting this semantic distinction if helpful.

packages/blade/src/components/Collapsible/CollapsibleText.web.tsx
rkdotxyz: Intentional separation — `CollapsibleText` and `CollapsibleLink` have different semantic roles:

- `CollapsibleLink` renders a `<a>` (via BaseLink) with link semantics — appropriate when the trigger text looks/behaves like a navigational link
- `CollapsibleText` renders a `<div role="button">` — appropriate when the trigger is plain text acting as an action toggle

Merging them into one component with a variant prop would conflate navigation semantics with action semantics in the underlying DOM. Keeping as separate components.

(Also addressed the a11y gap flagged in comment #6 — added keyboard handling for the button role.)

packages/blade/src/components/Avatar/avatarTokens.ts
rkdotxyz: Intentional design call — `xsmall: 14` at `normal` density provides slightly tighter overlap for the smallest avatar size to avoid disproportionate spacing. At 20px diameter, a 16px overlap would leave only 4px visible per avatar which is too little; 14px leaves 6px visible, maintaining legibility of the avatar content.

packages/blade/src/components/Avatar/avatarTokens.ts
rkdotxyz: Good catch — added `compact` key to `avatarGroupDensityOverlapTokens` with `size / 2` values and removed the special-case ternary from both web and native AvatarGroup.

packages/blade/src/components/Accordion/AccordionSurface.native.tsx
rkdotxyz: Added a comment explaining the gap — the prop is kept in the type (parent always passes it) but native intentionally does not implement color scheme styling yet. Removing it from the type would cause a TS error at the call site.

packages/blade/src/components/Alert/Alert.tsx
rkdotxyz: Extracted into `getAlertIconColor(color, emphasis)` helper in `styles.ts` — explicit cases for each combination, no nested ternaries.

packages/blade/src/components/Collapsible/CollapsibleText.web.tsx
rkdotxyz: Fixed — added `onKeyDown` handler for Enter/Space and `tabIndex={0}` (disabled → -1) on the web component.

packages/blade/src/components/Collapsible/CollapsibleText.native.tsx
rkdotxyz: Fixed — replaced `currentColor` with `interactive.icon.gray.muted` token on native.

swapnil-kr1 (approved):

PR #3459, opened 2026-06-07
feat(Input): update input sizing, success border, and add validationTextPlacement prop

rkdotxyz:
## Summary

- **Height updates**: Increased `xsmall` input height from 26px → 28px and `small` from 30px → 32px
- **Success border fix**: Changed success `validationState` border from green (`interactive.border.positive.default`) to gray (`interactive.border.gray.default`) with thin width, so the focused state (blue primary border) takes priority — matching other variants
- **New `validationTextPlacement` prop**: Adds `'outside' | 'inside'` option to all input components via `FormInputValidationProps`
  - `'outside'` (default): Current behavior — validation text renders below the input
  - `'inside'`: Renders error/success text inside the input's trailing area with appropriate colors (green for success, red for error) and size-responsive typography
  - `helpText` still renders externally when placement is `'inside'`
  - Error state retains the thick red border regardless of placement


packages/blade/src/components/Form/FormTypes.ts
```diff
@@ -64,4 +64,10 @@ export type FormInputValidationProps = {
    *
    */
   validationState?: 'success' | 'error' | 'none';
+  /**
+   * Placement of the validation text relative to the input
+   *
+   * @default 'outside'
+   */
+  validationTextPlacement?: 'outside' | 'inside';
 };
```
swapnil-kr1, line 73: **[major]** `api-decision-critique`

**Problem:** New `validationTextPlacement` prop overlaps semantically with existing `showHintsAsTooltip` boolean prop on BaseInput. Both control where validation/hint text renders relative to the input. Their interaction is unclear — what happens if both `showHintsAsTooltip={true}` and `validationTextPlacement='inside'` are set?

**Suggestion:** Consider unifying into a single prop (e.g. `validationTextPlacement: 'outside' | 'inside' | 'tooltip'`) or document/enforce mutual exclusivity. At minimum, add runtime validation that prevents both being set simultaneously.

packages/blade/src/components/Input/BaseInput/baseInputStyles.ts
```diff
@@ -116,11 +116,11 @@ export const getInputBackgroundAndBorderStyles = ({
       : getIn(theme.colors, baseInputBorderColor[baseInputState]);
   borderWidth = getIn(theme.border.width, baseInputBorderWidth[baseInputState]);
 
-  if (!isTableInputCell && validationState && validationState !== 'none') {
+  if (!isTableInputCell && validationState === 'error') {
```
swapnil-kr1, line 119: **[major]** `code-quality-critique`

**Problem:** The second branch `else if (validationState && validationState !== 'none' && validationState !== 'success')` now only triggers for 'error' in table input cells (borderless). Since the only remaining valid value is 'error', this is equivalent to `validationState === 'error'`, making it redundant with the first `if` block. For table input cells, the success background color (`baseInputBorderlessBackgroundColor.success`) is now never applied — potentially an unintended regression.

**Suggestion:** Confirm whether table input cells should still show a success background color. If yes, the condition change breaks table cell success styling. If no, simplify the else-if to `else if (validationState === 'error')` for clarity.

packages/blade/src/components/Input/BaseInput/BaseInputVisuals.tsx
```diff
@@ -405,6 +429,22 @@ export const BaseInputVisuals = ({
             </Text>
```
swapnil-kr1, line 442: **[major]** `usecase-critique`

**Problem:** The `validationTextPlacement='inside'` renders validation text in the trailing area, but BaseInput already has `trailingIcon`, `trailingButton`, and `suffix` props that occupy the same space. What happens when a consumer uses both `trailingIcon` and `validationTextPlacement='inside'` simultaneously? The text renders alongside trailing icons with no conflict handling.

**Suggestion:** Document the conflict resolution when trailing visuals compete for space, or add logic to hide the trailing icon when validation text is shown inside. At minimum, add a `__DEV__` warning when both are set.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
```diff
@@ -1024,10 +1025,12 @@ const _BaseInput: React.ForwardRefRenderFunction<BladeElementRef, BaseInputProps
     activeDescendant,
   });
```
swapnil-kr1, line 1027: **[major]** `code-quality-critique`

**Problem:** When `validationTextPlacement` is 'inside', `errorText` and `successText` are suppressed from FormHint, but the accessibility IDs (`errorTextId`, `successTextId`) are still passed to the input's `aria-describedby`. The inside validation text in `BaseInputVisuals` doesn't receive these IDs — breaking screen reader association.

**Suggestion:** Pass the appropriate `id` attribute (`errorTextId` or `successTextId`) to the inside validation text element in `BaseInputVisuals` so that `aria-describedby` on the input correctly references the visible validation message.

packages/blade/src/components/Input/BaseInput/baseInputTokens.ts
```diff
@@ -13,8 +13,8 @@ export const baseInputHeight: Record<
   NonNullable<BaseInputProps['size']>,
   typeof size[keyof typeof size]
 > = {
-  xsmall: size['26'],
-  small: size['30'],
+  xsmall: size['28'],
```
swapnil-kr1, line 16: **[minor]** `api-decision-critique`

**Problem:** Height changes for `xsmall` (26px→28px) and `small` (30px→32px) are visual breaking changes affecting all existing input consumers, unrelated to the new `validationTextPlacement` prop. Bundling unrelated breaking visual changes in a feature PR makes it harder to bisect regressions.

**Suggestion:** Consider splitting size token changes into a separate PR to isolate the blast radius.

packages/blade/src/components/Input/BaseInput/baseInputTokens.ts
```diff
@@ -37,7 +37,7 @@ export const baseInputBorderColor = {
   focused: 'interactive.border.primary.default',
   disabled: 'interactive.border.gray.disabled',
   error: 'interactive.border.negative.default',
-  success: 'interactive.border.positive.default',
+  success: 'interactive.border.gray.default',
```
swapnil-kr1, line 40: **[minor]** `api-decision-critique`

**Problem:** The success border behavior change (green thick → gray thin) is a visual breaking change for all existing consumers using `validationState='success'`. This is bundled into the same PR but is unrelated to `validationTextPlacement`.

**Suggestion:** Consider splitting the success border change into a separate PR or documenting it as a breaking visual change, since existing consumers may rely on the green border as a visual signal.

packages/blade/src/components/Input/BaseInput/BaseInputVisuals.tsx
```diff
@@ -66,6 +67,19 @@ const textSize = {
   large: 'large',
 } as const;
 
+const validationTextSize = {
+  xsmall: 'xsmall',
+  small: 'small',
```
swapnil-kr1, line 74: **[minor]** `code-quality-critique`

**Problem:** The `validationTextColor` map includes a `'none'` entry (`surface.text.gray.subtle`) that is dead code — the ternary already guards against `validationState === 'none'`, so this key can never be reached.

**Suggestion:** Remove the `'none'` entry from `validationTextColor` since it can never be used in practice, or add a comment clarifying it's a TypeScript exhaustiveness safety net.

packages/blade/src/components/Input/BaseInput/baseInputTokens.ts
rkdotxyz: Keeping as-is. These sizing adjustments (xsmall 26→28px, small 30→32px) are part of the same design spec driving the validationTextPlacement feature — they ensure the inside validation text has adequate vertical space. Splitting into a separate PR adds coordination overhead without meaningful blast-radius isolation since both changes ship together in the same minor. Will document as a visual change in the changeset.

packages/blade/src/components/Input/BaseInput/baseInputTokens.ts
rkdotxyz: Keeping as-is. The success border change (green thick → gray thin) is an intentional design decision that accompanies the new validationTextPlacement prop — success state now communicates via inline text rather than a prominent border. This is a coordinated visual update from the same design spec, and splitting it would create an intermediate state that doesn't match any design milestone. Will ensure it's documented as a visual breaking change in release notes.

packages/blade/src/components/Form/FormTypes.ts
rkdotxyz: This is already handled — there is a `__DEV__` warning when both props are set simultaneously (`BaseInput.tsx` ~line 1044):

```js
if (showHintsAsTooltip && isValidationTextInside) {
  logger({
    message: 'showHintsAsTooltip and validationTextPlacement="inside" should not be used together. validationTextPlacement="inside" will take precedence.',
    moduleName: 'Input',
    type: 'warn',
  });
}
```

Unifying into a single prop would be a breaking change for existing consumers of `showHintsAsTooltip`, so keeping both with a runtime warning for now. Deprecation/unification can be considered in a future major version.

packages/blade/src/components/Input/BaseInput/baseInputStyles.ts
rkdotxyz: Good callout on clarity. The behavior change (table cells only showing error background, not success) is intentional per the design spec. Adding a clarifying comment above the condition to make this explicit.

packages/blade/src/components/Input/BaseInput/BaseInputVisuals.tsx
rkdotxyz: This is already handled — there is a `__DEV__` warning when both are set (`BaseInputVisuals.tsx` ~line 332):

```js
if (hasInsideValidationText && (hasTrailingIcon || hasTrailingButton || hasSuffix)) {
  logger({
    message: 'Using validationTextPlacement="inside" with trailingIcon, trailingButton, or suffix may cause layout conflicts. Consider using validationTextPlacement="outside" when trailing visuals are present.',
    moduleName: 'BaseInput',
    type: 'warn',
  });
}
```

Hiding trailing visuals automatically would be a breaking change for consumers relying on them, so a dev warning is the non-breaking approach here.

packages/blade/src/components/Input/BaseInput/BaseInput.tsx
rkdotxyz: This is already fixed in the PR. The inside validation text element receives the correct accessibility ID (`BaseInputVisuals.tsx` ~line 442):

```jsx
<BaseBox
  id={validationState === 'error' ? errorTextId : successTextId}
  ...
>
```

Both `errorTextId` and `successTextId` are passed down to `BaseInputVisuals` from `BaseInput.tsx` (~line 1237-1238). So `aria-describedby` on the input correctly references the visible validation text regardless of placement.

packages/blade/src/components/Input/BaseInput/BaseInputVisuals.tsx
rkdotxyz: Keeping the entry with an inline comment as a TypeScript exhaustiveness safety net:

```ts
none: 'surface.text.gray.subtle', // TypeScript exhaustiveness — never reached at runtime
```

This ensures the `validationTextColor[validationState]` indexing remains type-safe without needing a type assertion. Removing it would require a cast or non-null assertion at the usage site.

swapnil-kr1 (approved):

PR #3460, opened 2026-06-07
feat(FileUploadItem): add small size variant

rkdotxyz:
## Summary
- Adds a `'small'` size variant (36px height) to `FileUploadItem` for standalone usage
- Single-line layout: 20x20 file icon (no format text), filename + file size baseline-aligned, compact action buttons
- Small variant uses 8px border radius, no drop shadow/elevation, and `surface.border.gray.subtle` border color


packages/blade/src/components/FileUpload/FileUploadItem.tsx
```diff
@@ -55,117 +60,214 @@ const FileUploadItem = memo(
...
-                  />
-                )}
+        <BaseBox
+          width="100%"
+          display="flex"
+          flexDirection="column"
+          justifyContent={isSmall ? 'center' : undefined}
+        >
+          {isSmall ? (
+            <BaseBox
```
swapnil-kr1, line 77: **[major]** The small variant's success state does not render a `CheckCircleIcon`, unlike medium/large variants. Is this intentional by design, or should a small-sized check icon be shown to give users visual confirmation of a successful upload?

If intentional, might be worth a code comment noting the design decision.

packages/blade/src/components/FileUpload/FileUploadItem.tsx
```diff
@@ -55,117 +60,214 @@ const FileUploadItem = memo(
...
+              display="flex"
+              flexDirection="row"
+              alignItems="center"
```
swapnil-kr1, line 80: **[minor]** `paddingTop="spacing.2"` looks like a manual vertical alignment nudge. This may break if the icon or text size changes. Consider using `alignSelf="center"` on the icon wrapper for more robust alignment.

packages/blade/src/components/FileUpload/FileUploadItemIcon.tsx
```diff
@@ -6,16 +6,33 @@ import { useTheme } from '~utils';
...
 import { colors as globalColors } from '~tokens/global';
 
-type FileUploadItemIconProps = { fileName: string; uploadStatus: BladeFile['status'] };
+type FileUploadItemIconProps = {
+  fileName: string;
+  uploadStatus: BladeFile['status'];
+  iconSize?: 'small' | 'default';
+};
 
 const FileUploadItemIcon: React.ComponentType<FileUploadItemIconProps> = ({
```
swapnil-kr1, line 15: **[minor]** The small error icon uses `viewBox="0 0 38 38"` (square) while the small non-error icon uses `viewBox="0 0 38 39"` (slightly taller). This mismatch means the icon's aspect ratio subtly changes between error and non-error states at the same 20×20 rendered size. Should these be consistent?

packages/blade/src/components/FileUpload/types.ts
```diff
@@ -180,8 +180,14 @@ type FileUploadProps = (FileUploadPropsWithA11yLabel | FileUploadPropsWithLabel)
...
   FileUploadProps,
-  'onPreview' | 'onRemove' | 'onDismiss' | 'onReupload' | 'size'
+  'onPreview' | 'onRemove' | 'onDismiss' | 'onReupload'
 > & {
+  /**
+   * Size of the FileUploadItem component
+   *
+   * @default 'medium'
+   */
+  size?: 'small' | 'medium' | 'large' | 'variable';
```
swapnil-kr1: **[minor]** The `'variable'` option in `size` is meaningful when `FileUploadItem` is rendered inside `FileUpload` (where it controls drop area text, action buttons, height/width). For standalone usage, `'variable'` is confusing — consumers won't know what it does.

Consider narrowing the public type to `'small' | 'medium' | 'large'` and keeping `'variable'` as an internal-only type passed by the parent `FileUpload`.

packages/blade/src/components/FileUpload/FileUploadItem.tsx
```diff
@@ -55,117 +60,214 @@ const FileUploadItem = memo(
...
-                  <CheckCircleIcon
-                    size="medium"
-                    marginRight="spacing.2"
-                    color="interactive.icon.positive.normal"
-                  />
-                )}
+        <BaseBox
+          width="100%"
+          display="flex"
+          flexDirection="column"
```
swapnil-kr1, line 72: **[minor · maintainability]** The small branch duplicates ~80 lines of action button logic (uploading → close, error → retry+remove, success → preview+remove) that closely mirrors the medium/large branch below. Future bug fixes or new actions will need to be applied in both places.

Consider extracting a shared `<FileUploadItemActions size={...} status={...} ... />` component to DRY this up.

packages/blade/src/components/FileUpload/FileUploadItem.tsx
rkdotxyz: This is intentional by design. At 36px height, the compact layout prioritizes file name and action buttons over status iconography. Adding a check icon here would crowd the row. Will add a code comment noting this decision.

packages/blade/src/components/FileUpload/FileUploadItem.tsx
rkdotxyz: Keeping as-is — the `paddingTop="spacing.2"` is intentional for visual balance with the text baseline at this icon size. Will add a code comment clarifying this.

packages/blade/src/components/FileUpload/FileUploadItemIcon.tsx
rkdotxyz: Keeping as-is — the viewBox difference (`38x38` for error vs `38x39` for non-error) is intentional. The error icon artwork is natively square while the file icon is slightly taller. The mismatch at 20x20 rendered size provides visual balance. Will add a code comment noting this.

packages/blade/src/components/FileUpload/FileUploadItem.tsx
rkdotxyz: Acknowledged, but keeping as-is for now. The small variant has subtly different button sizes (`size="small"`), layout (`flexDirection="row"` inline vs column offsets), and fewer elements (no progress bar, no check icon) — so a shared component would need enough conditional branches that the abstraction may not reduce complexity meaningfully. If a fourth size variant ever comes along, this becomes worth extracting.

packages/blade/src/components/FileUpload/types.ts
rkdotxyz: Good catch — will fix this. Narrowing the public type to `'small' | 'medium' | 'large'` and adding an internal `InternalFileUploadItemSize` type that includes `'variable'` for the parent `FileUpload` to use.

PR #3463, opened 2026-06-08
fix(rn): tabs neutral underline color and scrollable panels

rohitr-raz:
## Summary
- Changes tab underline indicator color from `interactive.border.primary.default` (blue) to `interactive.border.neutral.highlighted` (neutral/black), matching the web implementation
- Wraps each tab panel's content in a `ScrollView` with `nestedScrollEnabled` so panel content scrolls vertically, with `nestedScrollEnabled` ensuring proper gesture coordination on Android inside `PagerView`

## Test plan
- [ ] Open Tabs stories in RN storybook
- [ ] Verify selected tab underline is neutral/black (not blue) for bordered and borderless variants
- [ ] Verify tab panel content scrolls smoothly when content overflows vertically
- [ ] Verify horizontal swipe between tabs still works with nested scroll
- [ ] Test on both iOS and Android


packages/blade/src/components/Tabs/Tabs.native.tsx
```diff
@@ -419,7 +419,15 @@ const _Tabs = (
             const shouldRender = !isLazy || index === safeIndex;
             return (
               <View key={panel.value} style={styles.pagerView}>
-                {shouldRender ? panel.children : null}
+                {shouldRender ? (
+                  <ScrollView
+                    nestedScrollEnabled
+                    showsVerticalScrollIndicator={false}
```
swapnil-kr1: **Concern — forced ScrollView may break consumer layouts**
Rationale: behind this change why we are introducing ScrollView?

rohitr-raz: Without the ScrollView wrapper, tab panel content that's taller than the screen would be clipped with no way to scroll.

swapnil-kr1: What if a consumer is already passing a `<ScrollView>` as the child? In that case we could end up with nested `<ScrollView>` components.

Can we add a check and only wrap with a `<ScrollView>` when the content isn't already inside one?

PR #3469, opened 2026-06-09
feat(Timestamp): add Timestamp component

debabratama1ik:
## Context

Sharing the API decisions doc for alignment before this gets reviewed as a full implementation. Happy to revise props, naming, or behaviour based on feedback — nothing is locked in.

The decisions doc was written using the `write-api-decision` skill from the Blade repo.

📄 **Decisions doc:** [`packages/blade/src/components/Timestamp/_decisions/decisions.md`](packages/blade/src/components/Timestamp/_decisions/decisions.md)

---

## What problem does this solve?

The dashboard currently has 50+ different date/time format strings across 100+ files in 15+ apps — every team solved this independently. The same transaction date renders 4 different ways depending on which page you're on.

## Proposed API

```jsx
// Relative (activity feeds, audit logs)
<Timestamp value={date} format="relative" />
// → "5m ago", auto-updates, tooltip shows full datetime on hover

// Default (transaction tables)
<Timestamp value={date} />
// → "21 Mar, 9:20 AM"

// Full absolute (order detail drawers)
<Timestamp value={date} format="dateTime" dateStyle="full" />
// → "21 March 2026, 9:20 AM"

// Developer / log surface
<Timestamp value={date} format="dateTime" dateStyle="numeric" hourCycle="24h" precision="second" />
// → "21/03/2026, 09:20:45"
```

## What's in this PR

- `_decisions/decisions.md` — API decision document ← **start here**
- Full implementation, Storybook stories, web/native/SSR tests, and changeset are included but treat these as a working draft — open to changes based on API feedback

## Questions for the Blade team

- Does the `format` + `dateStyle` split make sense, or would you prefer a different shape?
- Any prop names that clash with existing Blade conventions?
- Anything missing from the decisions doc?


packages/blade/src/components/Timestamp/_decisions/decisions.md
```diff
@@ -0,0 +1,366 @@
...
+type TimestampBodyProps = {
+  /** @default 'body' */
+  type?: 'body';
+  /** @default 'medium' */
+  size?: 'xsmall' | 'small' | 'medium' | 'large';
+  weight?: 'regular' | 'medium' | 'semibold';
+};
+
+type TimestampHeadingProps = {
+  type: 'heading';
```
saurabhdaware: do we have a usecase of showing the timestamp in a heading and display size?

PR #3477, opened 2026-06-09
fix(rn): fix dark mode card collapse and add scrollable card story

rohitr-raz:
## Summary

- **BladeProvider nested collapse fix**: `GestureHandlerRootView` had `flex: 1` which caused height collapse when `BladeProvider` was nested inside an auto-height flex column (e.g. dark mode card in stories). Now detects nesting via `ThemeContext` and skips `flex: 1` for nested providers.
- **StoryScrollView utility**: Added platform-specific `StoryScrollView` component (`.tsx` = React Fragment for web, `.native.tsx` = `ScrollView` for RN) so the Card story can wrap both light and dark cards in a scroll container without breaking the web storybook bundle.
- **Card story**: `CardTemplate` now wraps content in `StoryScrollView` so both light and dark cards are visible and scrollable on device.

##Description:

before:

Screenshot 2026-06-09 at 3 52 17 PM

after:

https://github.com/user-attachments/assets/a4c9af45-540e-4893-b1cc-c60e74f41ad3

## Test plan

- [ ] Open Card → CardExample story on RN — both light and dark cards should render fully
- [ ] Dark mode card should show all content (header, body, footer) without collapsing
- [ ] Story should be scrollable to reach the dark card below the light card
- [ ] Open web storybook — Card story should render without errors (no react-native bundle issues)
- [ ] Other stories unaffected (preview.tsx unchanged)

### DIFF
--- a/packages/blade/src/components/BladeProvider/BladeProvider.native.tsx
+++ b/packages/blade/src/components/BladeProvider/BladeProvider.native.tsx
@@ -1,5 +1,5 @@
 import type { ReactElement } from 'react';
-import { useContext } from 'react';
+import { createContext, useContext } from 'react';
 import { ThemeProvider as StyledComponentThemeProvider } from 'styled-components';
 import { PortalHost, PortalProvider } from '@gorhom/portal';
 import { GestureHandlerRootView } from 'react-native-gesture-handler';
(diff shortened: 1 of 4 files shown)


packages/blade/src/components/BladeProvider/BladeProvider.native.tsx
```diff
@@ -16,10 +17,12 @@ const BladeProvider = ({
   colorScheme: initialColorScheme,
   children,
 }: BladeProviderProps): ReactElement => {
+  const existingTheme = useContext(ThemeContext);
+  const isNested = existingTheme.theme !== null;
   const { theme, themeContextValue } = useBladeProvider({ initialColorScheme, themeTokens });
```
shivam-rzpay, line 26: Can you please explain why this is needed ? is it just for stories or for real component as well ?

packages/blade/src/utils/storybook/StoryScrollView.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
shivam-rzpay, line 1: why this file was needed ?

shivam-rzpay (changes requested):

packages/blade/src/components/BladeProvider/BladeProvider.native.tsx
rohitr-raz: The change detects nesting via ThemeContext (existingTheme.theme !== null means we're already inside a BladeProvider). We are nesting  a second BladeProvider inside the card to create a dark-mode. If we're already inside a BladeProvider (via ThemeContext) , skip flex: 1 on the inner GestureHandlerRootView , because if the nested GestureHandlerRootView also had flex: 1 , now that it's inside a parent that has no fixed height (cards height is auto/content-sized), so flex: 1 has nothing to expand into so it collapses to 0 height and the dark card's content disappears which was the issue happening in our case

packages/blade/src/utils/storybook/StoryScrollView.tsx
rohitr-raz: the dark mode card was getting clipped as it was crossing through screen size, so to completely see both of the cards had to add a scroll option to the story

packages/blade/src/components/BladeProvider/BladeProvider.native.tsx
shivam-rzpay: can you check how web does it ?

rohitr-raz: The web version simply doesn't have GestureHandlerRootView, so it has no nesting concern. The native check exists just to avoid GestureHandlerRootView forcing flex: 1 on a nested provider.

rohitr-raz: On web, nesting BladeProvider is completely layout-neutral — it's just ThemeContext.Provider and styled-components  ThemeProvider, which are pure React context with zero visual/layout impact. So wrapping a dark card in a nested BladeProvider on web has no side effects at all.

shivam-rzpay (commented):
Can you ask @swapnil-kr1 to skim it once

PR #3484, opened 2026-06-10
fix(rn): fix Popover animation flash and story flex layout

rohitr-raz:
## Summary

- Use `styled(Animated.View)` as the base for `PopoverContentWrapper` instead of `styled(BaseBox)` so opacity is co-located with `elevation` on the same View. On Android, elevation renders its shadow independently of a parent `Animated.View`'s opacity — this caused the shadow/box to flash visible before the content faded in. Co-locating them fixes the flash.
- Add `flexDirection="row"` to the settlement rows in the Popover story so label and amount sit side by side instead of stacking vertically on React Native.

##Description-

before:

https://github.com/user-attachments/assets/98976f23-e4d7-4e04-80fe-4e098def94df

after:

https://github.com/user-attachments/assets/6cca48b2-1227-4c26-9d28-dbb619de20a9

##Test plan

- [ ] Open Popover on Android — shadow and content should fade in together, no white box flash
- [ ] Open Popover on iOS — animation should look the same as before
- [ ] Popover story shows "Gross Settlements / ₹5,000", "Deductions / ₹250", "Net Settlements / ₹4,750" in a single row each

### DIFF
--- a/packages/blade/src/components/Popover/PopoverContentWrapper.native.tsx
+++ b/packages/blade/src/components/Popover/PopoverContentWrapper.native.tsx
@@ -6,20 +6,21 @@ import type { EasingFn } from 'react-native-reanimated';
 import Animated, {
   useAnimatedStyle,
   useSharedValue,
-  withDelay,
   withTiming,
 } from 'react-native-reanimated';
 import React from 'react';
 import type { View } from 'react-native';
 import { getPopoverContentWrapperStyles } from './getPopoverContentWrapperStyles';
 import type { PopoverContentWrapperProps } from './types';
-import BaseBox from '~components/Box/BaseBox';
 import { useTheme } from '~components/BladeProvider';
 import { size } from '~tokens/global';
 import type { ColorSchemeNames } from '~tokens/theme';
 import { castNativeType } from '~utils';
 
-const StyledPopoverContentWrapper = styled(BaseBox)<{
+// Use styled(Animated.View) so opacity animates on the same View that has elevation.
+// On Android, elevation renders its shadow independently of a parent Animated.View's
+// opacity — co-locating opacity and elevation on the same View fixes the flash.
+const StyledPopoverContentWrapper = styled(Animated.View)<{
   collapse?: boolean;
   styles: CSSProperties;
   isMobile: boolean;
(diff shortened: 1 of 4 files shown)


packages/blade/src/components/Popover/PopoverContentWrapper.native.tsx
```diff
@@ -45,29 +42,34 @@ const PopoverContentWrapper = React.forwardRef<View, PopoverContentWrapperProps>
 
     React.useEffect(() => {
       const timings = { easing, duration };
-      opacity.value = withDelay(duration, withTiming(isVisible ? 1 : 0, timings));
-      translate.value = withDelay(duration, withTiming(isVisible ? 0 : offset, timings));
+      opacity.value = withTiming(isVisible ? 1 : 0, timings);
+      translate.value = withTiming(isVisible ? 0 : offset, timings);
```
shivam-rzpay, line 49: The original code had a deliberate delay equal to duration (the quick motion token) before starting the animation. This was likely there so the popover positioning could settle before animating in. Removing it fixes the flash but may introduce a different issue — the popover might now animate from an incorrect position if floating-ui hasn't finished measuring/positioning yet.

Can you please share ios video as well ? with different popover sizes

packages/blade/src/components/Popover/PopoverContentWrapper.native.tsx
```diff
@@ -3,23 +3,20 @@
...
 import { useTheme } from '~components/BladeProvider';
 import { size } from '~tokens/global';
 import type { ColorSchemeNames } from '~tokens/theme';
 import { castNativeType } from '~utils';
 
-const StyledPopoverContentWrapper = styled(BaseBox)<{
+// Use styled(Animated.View) so opacity animates on the same View that has elevation.
+// On Android, elevation renders its shadow independently of a parent Animated.View's
+// opacity — co-locating opacity and elevation on the same View fixes the flash.
+const StyledPopoverContentWrapper = styled(Animated.View)<{
```
shivam-rzpay: The insight about Android elevation rendering its shadow independently of a parent's opacity makes sense.
  
One thing I wanted to flag — switching from styled(BaseBox) to styled(Animated.View) might unintentionally drop some of BaseBox's native functionality. Specifically, BaseBox.native.tsx resolves the elevation prop into actual shadow styles via getElevationValue and merges them into style. 
With a raw Animated.View, the elevation={20} prop might become a no-op and the shadow could silently disappear.

A possible alternative that keeps everything intact:

  const AnimatedBaseBox = Animated.createAnimatedComponent(BaseBox);
  const StyledPopoverContentWrapper = styled(AnimatedBaseBox)<{ ... }>(...);

This way you get BaseBox's prop handling and animated opacity on the same View that carries elevation

shivam-rzpay (changes requested):

packages/blade/src/components/Popover/PopoverContentWrapper.native.tsx
rohitr-raz: done

shivam-rzpay (approved):
LGTM

swapnil-kr1 (approved):

PR #3485, opened 2026-06-10
fix(rn): display tags in a horizontal row in Tag story

rohitr-raz:
## Summary

- Added `flexDirection="row"` and `flexWrap="wrap"` to the tags container in the `ControlledTags` story so tags render side by side instead of stacking vertically on React Native.

## Test plan

- [ ] Add multiple tags in the ControlledTags story on React Native — tags should appear in a horizontal row and wrap to the next line when full

### DIFF
--- a/packages/blade/src/components/Tag/Tag.stories.tsx
+++ b/packages/blade/src/components/Tag/Tag.stories.tsx
@@ -135,7 +135,7 @@ export const ControlledTags = (props: TagProps): React.ReactElement => {
 
   return (
     <Box>
-      <Box paddingY="spacing.4">
+      <Box paddingY="spacing.4" display="flex" flexDirection="row" flexWrap="wrap">
         {tags.map((tagName) => (
           <Tag
             key={tagName}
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/Tag/Tag.stories.tsx
```diff
@@ -135,7 +135,7 @@ export const ControlledTags = (props: TagProps): React.ReactElement => {
 
   return (
     <Box>
-      <Box paddingY="spacing.4">
+      <Box paddingY="spacing.4" display="flex" flexDirection="row" flexWrap="wrap">
```
shivam-rzpay, line 138: can you check if it affects web stories ? If yes, please add a react native check

shivam-rzpay (changes requested):

packages/blade/src/components/Tag/Tag.stories.tsx
rohitr-raz: I have checked the web stories , it isn't getting affected

shivam-rzpay (approved):
LGTM

swapnil-kr1 (approved):

PR #3486, opened 2026-06-10
fix(rn): fix double border on selected ChipGroup items

rohitr-raz:
## Summary

- On React Native, both `StyledChipWrapper` and `StyledAnimatedChip` were rendering a border when a chip was selected, causing a visible double border effect
- On web, the double border is intentional (outer ring + inner chip border) and works correctly due to CSS defaults
- On native, both borders render with no offset, causing an ugly stacked double border

##Description:

before:

Screenshot 2026-06-09 at 6 13 24 PM

after:

Screenshot 2026-06-16 at 7 06 45 PM

## Fix

- Suppressed the inner border (`borderWidth: 0`) on `StyledAnimatedChip` on native — `StyledChipWrapper` owns the border on native
- Doubled `StyledChipWrapper`'s `borderWidth` to `2px` to maintain visual weight similar to web's double border
- Adjusted inner `borderRadius` on `StyledAnimatedChip` to account for the thicker outer border, preventing corner clipping

## Test plan

- [ ] Selected chip shows a single clean border on native
- [ ] Unselected chips are unaffected
- [ ] All chip sizes (xsmall, small, medium, large) look correct
- [ ] Corner radius matches correctly with no clipping


packages/blade/src/components/Chip/__tests__/__snapshots__/Chip.native.test.tsx.snap
```diff
@@ -434,8 +435,8 @@ exports[`<Chip /> should prioritize Chip's isDisabled prop instead of ChipGroup
                           "alignSelf": "center",
                           "backgroundColor": "hsla(0, 0%, 100%, 1)",
                           "borderColor": "hsla(206, 10%, 29%, 0.18)",
-                          "borderRadius": 8,
-                          "borderWidth": 1,
+                          "borderRadius": 6,
+                          "borderWidth": 0,
                           "display": "flex",
```
shivam-rzpay, line 441: There is no border width now when the chip is not selected. Can you please check it once and compare it with web ? When chip is selected, it looks fine but border is not visible when chip is in unselected state

shivam-rzpay (changes requested):

packages/blade/src/components/Chip/__tests__/__snapshots__/Chip.native.test.tsx.snap
rohitr-raz: Screenshot 2026-06-15 at 6 59 21 PM

resolved the issue

packages/blade/src/components/Chip/AnimatedChip.native.tsx
```diff
@@ -1,29 +1,44 @@
...
+    // (getAnimatedChipStyles default) would clip both edges and eat the leading icon's padding.
+    // flex-start clips only the trailing text, keeping the icon's inset.
+    justifyContent: 'flex-start',
+    // The selected fill (interactive.background.*.faded) is semi-transparent and is painted by
+    // StyledChipWrapper — same as web. Keep this outer layer transparent so the alpha isn't
+    // applied twice and the blue doesn't render darker than web.
+    backgroundColor: 'transparent',
+    // When selected, medium & large own a single thicker border (the wrapper border is
+    // suppressed): 1.5px for medium, 2px for large. xsmall/small and all unchecked chips
+    // keep the default thin 1px outline.
```
shivam-rzpay, line 26: is this same for web as well ?

packages/blade/src/components/Chip/Chip.tsx
```diff
@@ -204,7 +205,7 @@ const _Chip: React.ForwardRefRenderFunction<BladeElementRef, ChipProps> = (
                 size={_size}
                 display="flex"
                 flexDirection="row"
-                justifyContent="center"
+                justifyContent={isReactNative() ? 'flex-start' : 'center'}
```
shivam-rzpay, line 208: can you test it once on different text lengths on native

packages/blade/src/components/Chip/StyledChipWrapper.native.tsx
```diff
@@ -9,10 +9,14 @@ const StyledChipWrapper = styled(BaseBox)<StyledChipWrapperProps>(
   ({ theme, borderColor, isChecked, size = 'small' }) => {
     const borderRadius = chipBorderRadiusTokens[size];
     const outerRadius = theme.border.radius[borderRadius];
+    // medium & large show a single border owned by AnimatedChip; drop the wrapper's own
+    // border so the two don't stack into a double. xsmall/small keep their original border.
```
shivam-rzpay, line 13: how does web handles it ?

shivam-rzpay (changes requested):

packages/blade/src/components/Chip/AnimatedChip.native.tsx
rohitr-raz: No — that size-scaling is native-only. On web the selected state is an intentional double border: the outer AnimatedChip ring + the inner StyledChipWrapper ring, both thin (1px), every size — CSS stacks them cleanly. RN can't stack two borders without them merging into an offsetless thick double, so on native medium/large collapse to a single border scaled to 1.5/2px to match web's visual weight. xsmall/small keep the double, like web.

packages/blade/src/components/Chip/Chip.tsx
rohitr-raz: Verified in the All Sizes story across short ("Wallet on My App") and long, truncating ("Offer discounts, Pay Later  & EMI options") labels: short labels render unchanged; long labels truncate the trailing text while the leading icon keeps its inset.

packages/blade/src/components/Chip/StyledChipWrapper.native.tsx
rohitr-raz: StyledChipWrapper.web keeps its border (borderColor: isChecked ? … : transparent), so web deliberately renders the  double border (outer + inner). On native that double stacks badly, so medium/large drop the wrapper  border and let AnimatedChip own a single scaled border; xsmall/small keep the wrapper border since the double reads fine there.

shivam-rzpay (approved):
lgtm. Ask @swapnil-kr1 to skim it once

packages/blade/src/components/Chip/AnimatedChip.native.tsx
```diff
@@ -1,29 +1,44 @@
...
+    ...(props.isChecked &&
+      size === 'medium' && { borderWidth: getIn(theme, 'border.width.thick') as number }),
+    ...(props.isChecked &&
+      size === 'large' && { borderWidth: getIn(theme, 'border.width.thicker') as number }),
```
swapnil-kr1, line 30: for size: small what is the default borderWidth

packages/blade/src/components/Chip/StyledChipWrapper.native.tsx
```diff
@@ -9,10 +9,14 @@ const StyledChipWrapper = styled(BaseBox)<StyledChipWrapperProps>(
...
+    const isMediumOrLarge = size === 'medium' || size === 'large';
     const outerBorderWidth = getIn(theme, 'border.width.thin') as number;
     return {
       display: 'flex',
-      borderColor: isChecked ? getIn(theme.colors, borderColor) : 'transparent',
+      borderColor: isChecked && !isMediumOrLarge ? getIn(theme.colors, borderColor) : 'transparent',
+      ...(isMediumOrLarge && { borderWidth: 0 }),
```
swapnil-kr1, line 19: Why we havent handle for small size

packages/blade/src/components/Chip/AnimatedChip.native.tsx
rohitr-raz: The default `borderWidth` for small comes from `getAnimatedChipStyles`, which sets `borderWidth: thin` (1px) for all sizes. Since neither the `medium` nor `large` branch runs for small/xsmall, the base `thin` (1px) value is kept as-is. So a selected small chip still has a 1px border, same as unselected.

rohitr-raz: for size: small the borderWidth is thin(1px)

packages/blade/src/components/Chip/StyledChipWrapper.native.tsx
rohitr-raz: The double border visual issue was only manifesting for medium and large sizes — small and xsmall were rendering correctly without any fix. So the `isMediumOrLarge` guard is a targeted fix for only the affected sizes.

swapnil-kr1 (approved):

PR #3487, opened 2026-06-10
fix(rn): make overflow cards scrollable in Card story

rohitr-raz:
## Summary
- Wraps content in `overflow="auto"` and `overflowY="scroll"` card variants with a bounded `Box` + `StoryScrollView` on React Native
- On RN, `overflow="auto"` and `overflowY="scroll"` are invalid values — content would overflow freely without a bounded container
- Web behavior is completely unchanged (`isReactNative()` returns false, Box props are `undefined`, `StoryScrollView` is a passthrough)

##Description:
before:
Screenshot 2026-06-10 at 10 40 09 PM

after:

https://github.com/user-attachments/assets/8a93a306-dfd2-4438-9f14-a626102f7924

## Test plan
- [ ] Verify `overflow="auto"` card scrolls within its bounds on RN
- [ ] Verify `overflowY="scroll"` card scrolls within its bounds on RN
- [ ] Verify both cards look and scroll correctly on web (no regression)
- [ ] Verify `overflow="hidden"` card still clips content correctly on both platforms

### DIFF
--- a/packages/blade/src/components/Card/Card.stories.tsx
+++ b/packages/blade/src/components/Card/Card.stories.tsx
@@ -644,7 +644,10 @@ const CardWithOverflowExample = (): React.ReactElement => {
           <CardHeaderLeading title="Scrollable Content" />
         </CardHeader>
         <CardBody>
-          <Box height={isReactNative() ? '120px' : undefined} overflow={isReactNative() ? 'hidden' : undefined}>
+          <Box
+            height={isReactNative() ? '120px' : undefined}
+            overflow={isReactNative() ? 'hidden' : undefined}
+          >
             <StoryScrollView>
               <Text>
                 Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem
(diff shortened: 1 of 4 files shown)


packages/blade/src/components/Card/Card.stories.tsx
```diff
@@ -683,15 +693,23 @@ const CardWithOverflowExample = (): React.ReactElement => {
...
-            Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum
-            has been the industry's standard dummy text ever since the 1500s, when an unknown
-            printer took a galley of type and scrambled it to make a type specimen book. It has
-            survived not only five centuries, but also the leap into electronic typesetting,
-            remaining essentially unchanged. It was popularised in the 1960s with the release of
-            Letraset sheets containing Lorem Ipsum passages, and more recently with desktop
-            publishing software like Aldus PageMaker including versions of Lorem Ipsum.
-          </Text>
+          <Box
+            height={isReactNative() ? '120px' : undefined}
```
shivam-rzpay, line 697: why specific height ? will it affect if the height remains same and content grows beyond this ?

shivam-rzpay (changes requested):

packages/blade/src/components/Card/Card.stories.tsx
rohitr-raz: since RN doesn't support overflow: auto/scroll on views like web does, I used a ScrollView inside a fixed-height container to simulate it , if content grows beyond 120px then scroll option is there so ig its fine

shivam-rzpay (dismissed):
LGTM

swapnil-kr1 (approved):

swapnil-kr1 (dismissed):

swapnil-kr1 (approved):

PR #3488, opened 2026-06-10
fix(rn): baseline-align and prevent clipping of Amount currency/decimal

rohitr-raz:
## Summary
Fixes the React Native (Android) rendering of the `Amount` component, where the ₹ currency symbol and the decimal (`.79`) were clipped at large display sizes and visually misaligned with the integer.

- Switch the native layout from `alignItems: 'flex-end'` (aligns view bottoms, which misplaces smaller glyphs) to `alignItems: 'baseline'`, matching the existing web path.
- Add `includeFontPadding: false` to the text elements so Android reports an accurate baseline (its default font padding skews baseline alignment).
- Give the currency symbol and decimal their own proportional `lineHeight` (their own font-size token rather than the integer's larger one) — enough room to avoid clipping large display-size glyphs without shifting the baseline.

Web rendering is unchanged; all new values are gated behind `isReactNative`.

##Description:

before:
Screenshot 2026-06-10 at 10 38 18 PM

after:
Screenshot 2026-06-11 at 1 02 01 AM

## Test plan
- [ ] Verify on Android across all `type`/`size` combinations (body, heading, display) that ₹, the integer, and `.79` sit on one baseline.
- [ ] Verify nothing clips at the top/bottom, especially `display` medium/large/xlarge.
- [ ] Confirm web rendering is visually unchanged.


shivam-rzpay:
I think the clipping part is only for storybook. Can you check in a react native app and use different sizes for amount ?

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -20,7 +20,6 @@ import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
 import { throwBladeError } from '~utils/logger';
```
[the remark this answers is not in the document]
rohitr-raz: done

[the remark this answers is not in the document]
rohitr-raz: done

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -20,13 +20,19 @@ import { assignWithoutSideEffects } from '~utils/assignWithoutSideEffects';
...
 import { objectKeysWithType } from '~utils/objectKeysWithType';
 import { BaseText } from '~components/Typography/BaseText';
-import { Text } from '~components/Typography';
 import { opacity } from '~tokens/global';
 import type { FontFamily, FontSize } from '~tokens/global';
 import { makeAnalyticsAttribute } from '~utils/makeAnalyticsAttribute';
 import { useTheme } from '~components/BladeProvider';
 import { useBreakpoint } from '~utils/useBreakpoint';
 
+const IS_REACT_NATIVE = getPlatformType() === 'react-native';
```
shivam-rzpay: we already have a util for this import { isReactNative } from '~utils';

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -154,22 +160,57 @@ const AmountValue = ({
...
+            {amount.decimal}
+            {amount.fraction}
+          </BaseText>
+        </BaseBox>
+      );
+    }
 
     return (
-      <AmountWrapper>
+      <React.Fragment>
```
shivam-rzpay: why do you need to this change ? I think we are already adding react native condition above

shivam-rzpay (changes requested):

packages/blade/src/components/Amount/Amount.tsx
rohitr-raz: done

packages/blade/src/components/Amount/Amount.tsx
rohitr-raz: we added an if (isReactNative()) { return ...; } block above that handles the RN case entirely and returns early. So the code that follows is only ever reached on web — meaning AmountWrapper would always resolve to
React.Fragment. We replaced it with <React.Fragment> directly to make that explicit and remove the dead dynamic component.

shivam-rzpay: if not needed, then we can remove this change. Keep the web part as is i.e. in <AmountWrapper>

packages/blade/src/components/Amount/Amount.tsx
```diff
@@ -415,10 +457,11 @@ const _Amount = (
             marginRight="spacing.1"
             fontWeight={weight}
             fontSize={currencyFontProps.fontSize}
+            lineHeight={isReactNative() ? currencyFontProps.fontSize : undefined}
```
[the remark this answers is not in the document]
shivam-rzpay: @rohitr-raz check this comment once ?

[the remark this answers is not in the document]
rohitr-raz: done

packages/blade/src/components/Amount/Amount.tsx
rohitr-raz: done

shivam-rzpay (approved):
lgtm

swapnil-kr1 (approved):

PR #3492, opened 2026-06-11
fix(rn): fix ListItem text overflow on React Native

rohitr-raz:
## Description

On React Native, the content box inside `ListItemContentChildren` was expanding to its intrinsic width and overflowing off the right edge of the screen. Adding `flex=1` constrains it to the available width so text wraps correctly.

This change is inside a `getPlatformType() === 'react-native'` guard so web is unaffected.

Before:
Screenshot 2026-06-11 at 1 29 21 PM

After:

Screenshot 2026-06-11 at 1 23 53 PM

## Changes

- `ListItem.tsx`: add `flex={1}` to the `BaseBox` wrapper in the React Native branch of `ListItemContentChildren`

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset


packages/blade/src/components/List/ListItem.tsx
```diff
@@ -75,7 +75,7 @@ const ListItemContentChildren = ({
     As a workaround, we wrap individual strings in their own <Text /> and handle alignment with a parent <View> (BaseBox).
    */
   return getPlatformType() === 'react-native' ? (
-    <BaseBox display="flex" flexDirection="row" flexWrap="wrap">
+    <BaseBox display="flex" flexDirection="row" flexWrap="wrap" flex={1}>
```
shivam-rzpay, line 78: it can affect web. Please cross check

packages/blade/src/components/Dropdown/FilterChipGroupContext.native.tsx
```diff
@@ -0,0 +1,19 @@
...
+  setClearFilterCallbackTriggerer: () => {},
+});
+const FilterChipGroupProvider = FilterChipGroupContext.Provider;
+
+const useFilterChipGroupContext = (): FilterChipGroupContextType => {
+  const context = React.useContext(FilterChipGroupContext);
+  return context;
+};
+
+export { useFilterChipGroupContext, FilterChipGroupProvider };
```
shivam-rzpay: Check if we can reuse the context from web. The core logic will be same I guess

shivam-rzpay (changes requested):

packages/blade/src/components/List/ListItem.tsx
rohitr-raz: it is inside the react native gate , so I don't think it will affect the web.

shivam-rzpay (approved):
LGTM

swapnil-kr1 (approved):

PR #3502, opened 2026-06-11
feat(ChipGroup): add isFullWidth prop

rkdotxyz:
## Summary
- Adds `isFullWidth?: boolean` prop to `ChipGroup` (default `false`)
- When enabled, chips distribute equally in a single row with no wrap — segmented-control style
- Follows the same pattern as `ButtonGroup`'s `isFullWidth` (container `width: 100%` + child `flex: 1`)
- Zero breaking changes — existing behavior is preserved by default

## Changes
- `types.ts` — added prop to `ChipGroupCommonProps` and `ChipGroupContextType`
- `useChipGroup.ts` — accepts and passes `isFullWidth` through context
- `ChipGroup.tsx` — applies `width: 100%` + `flexWrap: nowrap` when enabled
- `Chip.tsx` — reads context, sets `flex: 1` and removes maxWidth cap when fullWidth

## Usage

```tsx
<ChipGroup isFullWidth selectionType="single" label="Plan">
  <Chip value="basic">Basic</Chip>
  <Chip value="standard">Standard</Chip>
  <Chip value="premium">Premium</Chip>
</ChipGroup>
```

## Test plan
- [x] TypeScript compiles cleanly (web + native)
- [x] All 46 existing + new Chip tests pass
- [x] Snapshot tests capture correct flex/nowrap styles
- [x] Selection behavior works correctly with isFullWidth


packages/blade/src/components/Chip/types.ts
```diff
@@ -150,6 +150,14 @@ type ChipGroupCommonProps = {
    * @default "primary"
    */
   color?: 'primary' | 'positive' | 'negative';
+  /**
```
saurabhdaware: 🟠 **[MAJOR]** · _code-quality-critique_

**Problem:** The JSDoc says 'Only available with selectionType=single', but there is no type-level or runtime enforcement. Passing isFullWidth=true with selectionType='multiple' compiles cleanly and silently produces inconsistent behavior.

**Suggestion:** Add a dev-time invariant in useChipGroup or ChipGroup.tsx (using throwBladeError or console.error in non-production) when isFullWidth && selectionType !== 'single'. Alternatively, enforce via a discriminated union type so the constraint is compile-time.

PR #3504, opened 2026-06-12
feat(Button): add loadingType prop for loading state variants

rkdotxyz:
## Summary
- Adds a new `loadingType` prop to `Button` and `BaseButton` components
- Currently supports `'indefinite'` (default) — the existing spinner-based loading indicator
- Sets up the extensibility point for future loading types (e.g. determinate progress)

## Changes
- `Button.tsx`: Added `loadingType` prop to `ButtonCommonProps`, destructured with default `'indefinite'`, passed through to `BaseButton`
- `BaseButton.tsx`: Added `loadingType` to `BaseButtonCommonProps`, gated spinner rendering on `loadingType === 'indefinite'`

## Test plan
- [ ] Existing loading state behavior unchanged (spinner shows when `isLoading={true}`)
- [ ] `loadingType="indefinite"` is equivalent to current behavior
- [ ] TypeScript compiles without new errors
- [ ] Snapshot tests pass with `yarn test:react Button`

### DIFF
--- a/packages/blade/src/components/Button/BaseButton/BaseButton.tsx
+++ b/packages/blade/src/components/Button/BaseButton/BaseButton.tsx
@@ -77,6 +77,12 @@ type BaseButtonCommonProps = {
   }>;
   type?: 'button' | 'reset' | 'submit';
   isLoading?: boolean;
+  /**
+   * Determines the type of loading indicator displayed when `isLoading` is true.
+   *
+   * @default 'indefinite'
+   */
+  loadingType?: 'indefinite';
   accessibilityProps?: Partial<AccessibilityProps>;
   variant?: 'primary' | 'secondary' | 'tertiary';
   color?:
@@ -393,6 +399,7 @@ const _BaseButton: React.ForwardRefRenderFunction<BladeElementRef, BaseButtonPro
     isDisabled = false,
     isFullWidth = false,
     isLoading = false,
+    loadingType = 'indefinite',
     onClick,
     onBlur,
     onKeyDown,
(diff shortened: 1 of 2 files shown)


packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -77,6 +77,12 @@ type BaseButtonCommonProps = {
   }>;
   type?: 'button' | 'reset' | 'submit';
   isLoading?: boolean;
+  /**
+   * Determines the type of loading indicator displayed when `isLoading` is true.
+   *
+   * @default 'indefinite'
+   */
+  loadingType?: 'indefinite';
```
saurabhdaware, line 85: @rzp-slash-public can you check if this prop name and value name is consistent across different components?

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
```diff
@@ -594,7 +601,7 @@ const _BaseButton: React.ForwardRefRenderFunction<BladeElementRef, BaseButtonPro
         motionEasing={motionEasing}
         isPressed={isPressed}
       >
-        {isLoading ? (
+        {isLoading && loadingType === 'indefinite' ? (
```
saurabhdaware, line 604: @rzp-slash-public does the implementation of how loadingType="indefinite" exists along with its visuals already implemented in BaseButton? I don't see it in this PR

packages/blade/src/components/Button/BaseButton/BaseButton.tsx
saurabhdaware, line 1: @rzp-slash-public add appropriate changeset for this PR

PR #3523, opened 2026-06-15
chore: setup React Native Storybook v10

shivam-rzpay:
## Summary

- Migrates React Native Storybook from v7 to v10 using `@storybook/react-native@10.2.3`
- Uses `withStorybook()` Metro wrapper for automatic story generation and module resolution
- Adds Hermes compatibility patches for Unicode property escapes (`\p{Lu}`, `\p{ID_Start}`) via `scripts/patch-storybook-hermes.js`
- Patch runs automatically before metro starts (`yarn start:android` / `yarn start:ios`)
- Tested and working on both Android emulator and iOS simulator

## Key changes

- `metro.config.js` — `withStorybook()` integration + mocks for web-only packages
- `.storybook/react-native/Storybook.tsx` — v10 API (`view.getStorybookUI()`)
- `.storybook/react-native/main.tsx` — story exclusion pattern for web-only stories
- `scripts/patch-storybook-hermes.js` — patches Unicode regex in es-toolkit and jsdoc-type-pratt-parser
- `package.json` — auto-runs patch before metro start

## How to run

```bash
cd packages/blade
yarn start:android  # or yarn start:ios
```

## Test plan

- [x] Metro bundle compiles without errors
- [x] Android: app launches, Storybook UI renders, stories work
- [x] iOS: app launches, Storybook UI renders, stories work
- [x] Patch script handles all Unicode regex issues automatically


packages/blade/ios/boost_1_76_0.tar.bz2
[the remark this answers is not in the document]
shivam-rzpay: resolved

packages/blade/metro.config.js
```diff
@@ -21,8 +50,50 @@ const config = {
...
+      if (moduleName === 'react-pdf' || moduleName.startsWith('react-pdf/')) {
+        return {
+          type: 'sourceFile',
+          filePath: path.resolve(mocksDir, 'react-pdf.js'),
+        };
+      }
+      if (moduleName === 'pdfjs-dist' || moduleName.startsWith('pdfjs-dist/')) {
+        return { type: 'empty' };
+      }
+      if (moduleName === 'storybook-react-router') {
```
swapnil-kr1, line 76: **The es-toolkit mock only exports `isEqual` and `mergeWith`.** If Storybook or any other dependency imports additional utilities from `es-toolkit` (e.g., `debounce`, `throttle`, `cloneDeep`), this will silently resolve to an object missing those exports, causing runtime `undefined is not a function` errors that are hard to debug.

Consider adding a `Proxy` that logs a warning for unmocked exports, or document prominently which es-toolkit consumers were audited.

packages/blade/.storybook/react-native/mocks/es-toolkit.js
```diff
@@ -0,0 +1,47 @@
...
+// Hermes in RN 0.72 doesn't support Unicode property escapes (\p{Lu}) used in es-toolkit/string.
+// Only mergeWith and isEqual are needed by @storybook/react-native-ui-common.
+
+function isEqual(a, b) {
+  if (a === b) return true;
+  if (a == null || b == null) return false;
+  if (typeof a !== typeof b) return false;
+  if (typeof a !== 'object') return false;
+
+  const keysA = Object.keys(a);
```
swapnil-kr1, line 16: **`isEqual` doesn't handle arrays, Date, RegExp, Map, Set, or circular references.** This is fine if the only consumer is Storybook's internal equality check on simple config objects, but if any story uses `isEqual` on arrays or dates, it will return incorrect results (arrays will be compared by numeric keys which may work accidentally, but `Date`/`RegExp` will fail).

At minimum, add an `Array.isArray` branch: `if (Array.isArray(a) && Array.isArray(b)) { return a.length === b.length && a.every((v, i) => isEqual(v, b[i])); }`

packages/blade/.storybook/react-native/main.tsx
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   stories: [
-    '../../src/**/!(_KitchenSink|*.test)*.stories.?(ts|tsx|js|jsx)',
+    '../../src/**/!(_KitchenSink|*.test|MotionRecipes|BottomNav|ChatInput|DatePicker|DetailedView|DrawerExamples|Fade|ListView|ListViewFigmaExample|Menu|Morph|Move|Settings|SideNav|Slide|RazorSense|Stagger|TableEditableCellAPI|TableFigmaExample|TableStripedRows|Toast|Typography|ChatMessage|CreationView|TopNav|LightBox|Preview)*.stories.?(ts|tsx|js|jsx)',
```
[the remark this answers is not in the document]
swapnil-kr1: @shivam-rzpay Can we use like *.native.stories

packages/blade/metro.config.js
shivam-rzpay: Tried wrapping it in a Proxy that throws for unmocked exports, but Metro's module system internally probes `$$typeof`, `__esModule`, and other symbol properties during resolution — causing the Proxy to throw before the app even loads. Since only `mergeWith` and `isEqual` are used by `@storybook/react-native-ui-common` (verified), keeping it as a plain export. Any missing export would surface as a clear `undefined is not a function` at runtime anyway.

packages/blade/.storybook/react-native/mocks/es-toolkit.js
shivam-rzpay: Fixed — added `Array.isArray` handling with recursive comparison.

packages/blade/metro.config.js
```diff
@@ -7,10 +10,36 @@ const { getDefaultConfig, mergeConfig } = require('@react-native/metro-config');
...
+  const pkgJsonPath = path.resolve(nodeModulesDir, pkgName, 'package.json');
+  if (!fs.existsSync(pkgJsonPath)) return null;
+
+  const pkgJson = JSON.parse(fs.readFileSync(pkgJsonPath, 'utf8'));
+  const exports = pkgJson.exports;
+  if (!exports || !exports[subpath]) return null;
+
+  const target = exports[subpath];
+  const resolved = typeof target === 'string' ? target : target.default || target.import;
+  if (!resolved) return null;
```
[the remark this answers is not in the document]
shivam-rzpay: Fixed — added a `Map` cache so each `package.json` is read and parsed only once per Metro bundle pass.

packages/blade/scripts/patch-storybook-hermes.js
```diff
@@ -0,0 +1,139 @@
...
+// Patch identifierStartRegex from jsdoc-type-pratt-parser (uses \p{ID_Start}, \p{Hex_Digit})
+let idFiles = [];
+for (const nodeModules of [localNodeModules, rootNodeModules]) {
+  if (!fs.existsSync(nodeModules)) continue;
+  try {
+    const found = execSync(
+      `grep -rl "identifierStartRegex = new RegExp" "${nodeModules}" --include="*.js" --include="*.mjs" 2>/dev/null || true`,
+      { encoding: 'utf8', maxBuffer: 10 * 1024 * 1024 }
+    ).trim().split('\n').filter(Boolean);
+    idFiles.push(...found);
```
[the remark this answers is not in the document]
shivam-rzpay: Fixed — removed `\-` from the character class.

shivam-rzpay:
Good suggestion! However, doing this now would require renaming 50+ existing story files to *.native.stories.tsx (or creating separate native-only story files). Keeping the exclusion list for now to unblock this setup — we will move to a convention-based approach as a follow-up.

packages/blade/.storybook/react-native/main.tsx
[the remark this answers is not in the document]
anuraghazra: Yes let's make this more generic.

packages/blade/.storybook/react-native/storybook.requires.ts
```diff
@@ -0,0 +1,66 @@
...
+import "@storybook/addon-ondevice-notes/register";
+import "@storybook/addon-ondevice-controls/register";
+import "@storybook/addon-ondevice-backgrounds/register";
+import "@storybook/addon-ondevice-actions/register";
+
+const normalizedStories = [
+  {
+    titlePrefix: "",
+    directory: "./src",
+    files: "**/!(_KitchenSink|*.test|MotionRecipes|BottomNav|ChatInput|DatePicker|DetailedView|DrawerExamples|Fade|ListView|ListViewFigmaExample|Menu|Morph|Move|Settings|SideNav|Slide|RazorSense|Stagger|TableEditableCellAPI|TableFigmaExample|TableStripedRows|Toast|Typography|ChatMessage|CreationView|TopNav|LightBox|Preview)*.stories.?(ts|tsx|js|jsx)",
```
anuraghazra, line 13: make this generic matcher or reuse it in a constant somehwre, importPathMatcher is also using this i think.

packages/blade/scripts/patch-storybook-hermes.js
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env node
+// Patches es-toolkit Unicode property escapes (\p{Lu}) with ASCII equivalents.
+// Hermes in RN 0.72 doesn't support Unicode property escapes in RegExp.
```
anuraghazra, line 3: is this a storybook specific thing? let add reference to the migration guide or any guide where you picked this up.

packages/blade/metro.config.js
```diff
@@ -7,10 +10,45 @@ const { getDefaultConfig, mergeConfig } = require('@react-native/metro-config');
  * @type {import('metro-config').MetroConfig}
  */
 
-// eslint-disable-next-line import/no-extraneous-dependencies
+const mocksDir = path.resolve(__dirname, '.storybook/react-native/mocks');
+const nodeModulesDir = path.resolve(__dirname, 'node_modules');
+
+// Metro 0.76 doesn't fully support package.json "exports" for subpath imports.
```
anuraghazra, line 16: I thought newer version of metro bundlers do support it? What's the effort of upgrading to latest?

packages/blade/.storybook/react-native/storybook.requires.ts
shivam-rzpay: Extracted the exclusion list into a `WEB_ONLY_STORIES` array constant in `main.tsx`. The glob is now built dynamically from that array, so there's a single place to add/remove web-only stories. `storybook.requires.ts` is auto-generated by `withStorybook()` at metro start from this config, so it stays in sync automatically.

packages/blade/scripts/patch-storybook-hermes.js
shivam-rzpay: Added references at the top of the script explaining what it patches (es-toolkit `CASE_SPLIT_PATTERN` and jsdoc-type-pratt-parser `identifierStartRegex`), why (Hermes in RN 0.72 lacks Unicode property escape support), and why upgrading RN isn't feasible (RN 0.76+ requires React 19, which breaks blade's React 18 peer dependency).

packages/blade/metro.config.js
shivam-rzpay: upgrading RN requires React 19 — RN 0.74 needs it for the new architecture internals, and RN 0.76+ explicitly depends on it. Since blade's peer dependency is React 18, we can't upgrade without a breaking change. Explored this during the setup — RN 0.74/0.76/0.85 all hit this wall.

anuraghazra (dismissed):

swapnil-kr1 (approved):

PR #3539, opened 2026-06-16
feat(native): add native shadow overlay to match web 3D styling

shivam-rzpay:
## Summary
- Adds `ButtonShadowOverlay.native.tsx` that simulates CSS inset box-shadows on native using SVG stroked Rects with masks
- Uses stroke on rounded Rects so shadows naturally follow border-radius contour (matching web's inset box-shadow behavior)
- Applies linear gradient masks for smooth side tapering
- Primary buttons get: top highlight, bottom shadow, border ring, and radial gradient
- Secondary/tertiary buttons get: bottom shadow and border ring only
- Disabled buttons render no overlay

## Test plan
- [ ] Verify primary button on iOS matches web's 3D appearance (highlight at top, darker bottom, radial glow)
- [ ] Verify secondary/tertiary buttons get appropriate shadow treatment
- [ ] Verify disabled buttons show no overlay
- [ ] Verify button press animation still works (background darkens, overlay remains)
- [ ] Verify no regression on web (ButtonShadowOverlay.web.tsx is a null stub)

### DIFF
--- a/packages/blade/src/components/Button/BaseButton/ButtonShadowOverlay.native.tsx
+++ b/packages/blade/src/components/Button/BaseButton/ButtonShadowOverlay.native.tsx
@@ -1,6 +1,13 @@
 import React from 'react';
 import { View, StyleSheet } from 'react-native';
-import Svg, { Defs, RadialGradient, Stop, Rect, ClipPath } from 'react-native-svg';
+import Svg, {
+  Defs,
+  RadialGradient,
+  LinearGradient,
+  Stop,
+  Rect,
+  Mask,
+} from 'react-native-svg';
 
 type ButtonShadowOverlayProps = {
   borderRadius: number;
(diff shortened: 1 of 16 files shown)


packages/blade/src/components/Button/BaseButton/ButtonShadowOverlay.native.tsx
```diff
@@ -0,0 +1,133 @@
...
+        </Mask>
+        <Mask id="bottomMask">
+          <Rect x="0" y="80%" width="100%" height="20%" fill="url(#bottomFade)" />
+        </Mask>
+        {showGradient ? (
+          <RadialGradient
+            id="btnGlow"
+            cx={0}
+            cy={0}
+            rx={64}
```
swapnil-kr1: **Hard-coded radial gradient values:** `cx={0}`, `cy={0}`, `rx={64}`, `ry={64}` are magic numbers that won't scale with different button sizes. For small buttons (e.g., icon-only), a 64px radius glow may extend well beyond bounds, and for larger buttons it may appear as a small spotlight in the corner. Consider deriving these from the button's actual dimensions (via `onLayout`) or at minimum extracting them as named constants with a comment explaining the assumed button size.

packages/blade/src/components/Button/BaseButton/StyledBaseButton.native.tsx
```diff
@@ -25,6 +26,7 @@ const StyledPressable = styled(Animated.createAnimatedComponent(Pressable))<
     alignSelf: 'center',
     display: 'flex',
     flexDirection: 'row',
```
swapnil-kr1, line 28: **Adding `overflow: 'hidden'` globally to the pressable may clip other visual effects.** This is needed to contain the SVG overlay within bounds, but it will also clip any elevation/shadow applied to the button itself, tooltips or badges that overflow, and any future press ripple animations. Consider applying the overflow constraint only to a wrapper `View` around the overlay + children, rather than the root pressable.

swapnil-kr1 (dismissed):

swapnil-kr1 (dismissed):

swapnil-kr1 (approved):

PR #3547, opened 2026-06-17
feat(SideNav): add subtle press effect to SideNavLink

himanshusharma-rzp:
## Description

Adds a subtle, tactile **press effect** to `SideNavLink`. When an item is pressed, its inner content (icon, text, badge/`titleSuffix`, chevron, and the `trailing` button) gently scales down and springs back on release, while the row/background frame stays full-size. This gives clicks a responsive, physical feel consistent with the press feedback already used in Blade's `Button`.

## Changes

- **`SideNavLink.web.tsx`**: On pointer press, a `.nav-link-pressed` class is toggled on the container (`StyledNavLinkContainer`) via pointer event handlers. CSS then scales the anchor's content (`.styled-nav-link > *`) and the trailing button wrapper (`.show-on-link-hover`) to `0.98`, leaving the row background/padding untouched. The effect covers plain L1/L2/L3 links and the L3 accordion trigger.
- **`tokens.ts`**: New `getNavItemPressTransitions()` helper and `NAV_ITEM_PRESSED_SCALE` constant. Transitions reuse existing motion tokens — `2xquick` + `emphasized` easing for press-in, `moderate` + `overshoot` easing for the springy release. Added a `PRESSED` class name.
- **Snapshots**: Updated `SideNav.web` and `SideNav.ssr` snapshots for the new styled-components output.
- **`.gitignore`**: Ignore a local-only explanation doc.

### Implementation notes
- **Pure CSS, no re-renders** — the press state is a direct `classList` mutation on the container, not React state.
- **Content-only scale** — the background/hover-highlight pill stays full-size; only the content shrinks (per design intent).
- **Pointer-only** — keyboard users retain the existing focus ring (no behavioral change for them); `:active` isn't reliable for keyboards without extra JS, which was intentionally avoided.
- `:has()` was deliberately **not** used because JSDOM (test env) throws on it; the class-toggle approach works everywhere.

## Additional Information

- Verified live in Storybook: on real pointer press, content computes `scale(0.98)` while the container/anchor background remains `none` (full size); release returns to `scale(1)` with the overshoot spring.
- The press scale (`0.98`) is intentionally subtle since the stable frame reduces perceived motion — easy to tune via `NAV_ITEM_PRESSED_SCALE` in `tokens.ts` if a stronger effect is desired.

## Component Checklist

- [ ] Update Component Status Page
- [x] Perform Manual Testing in Other Browsers 
- [ ] Add KitchenSink Story
- [x] Add Interaction Tests (if applicable) 
[...]

PR #3548, opened 2026-06-17
feat(SideNav): add subtle press effect to SideNavLink

himanshusharma-rzp:
## Description

Adds a subtle, tactile **press effect** to `SideNavLink`. When an item is pressed, its inner content (icon, text, badge/`titleSuffix`, chevron, and the `trailing` button) gently scales down and springs back on release, while the row/background frame stays full-size. This gives clicks a responsive, physical feel consistent with the press feedback already used in Blade's `Button`.

> Re-raised from a canonical-repo branch (previously #3547, opened from a fork — the coverage-comment CI step can't authenticate on fork PRs because secrets aren't exposed).

## Changes

- **`SideNavLink.web.tsx`**: On pointer press, a `.nav-link-pressed` class is toggled on the container (`StyledNavLinkContainer`) via pointer event handlers. CSS then scales the anchor's content (`.styled-nav-link > *`) and the trailing button wrapper (`.show-on-link-hover`) to `0.98`, leaving the row background/padding untouched. The effect covers plain L1/L2/L3 links and the L3 accordion trigger.
- **`tokens.ts`**: New `getNavItemPressTransitions()` helper and `NAV_ITEM_PRESSED_SCALE` constant. Transitions reuse existing motion tokens — `2xquick` + `emphasized` easing for press-in, `moderate` + `overshoot` easing for the springy release. Added a `PRESSED` class name.
- **Snapshots**: Updated `SideNav.web` and `SideNav.ssr` snapshots for the new styled-components output.
- **`.gitignore`**: Ignore a local-only explanation doc.

### Implementation notes
- **Pure CSS, no re-renders** — the press state is a direct `classList` mutation on the container, not React state.
- **Content-only scale** — the background/hover-highlight pill stays full-size; only the content shrinks (per design intent).
- **Pointer-only** — keyboard users retain the existing focus ring (no behavioral change for them); `:active` isn't reliable for keyboards without extra JS, which was intentionally avoided.
- `:has()` was deliberately **not** used because JSDOM (test env) throws on it; the class-toggle approach works everywhere.

## Additional Information

- Verified live in Storybook: on real pointer press, content computes `scale(0.98)` while the container/anchor background remains `none` (full size); release returns to `scale(1)` with the overshoot spring.
- The press scale (`0.98`) is intentionally subtle since the stable frame reduces perceived motion — easy to tune via `NAV_ITEM_PRESSED_SCALE` in `tokens.ts` if a stronger effect is desired.

## Component Checklist

- [ ] Add KitchenSink Story
- [x] Add changeset


.gitignore
```diff
@@ -100,3 +100,6 @@ knowledgebase-ts-errors.json
 .cursor/worktrees/
 .cursor/rules/frontend-blade-rules.mdc
 .playwright-mcp/
+
+# local-only explanation docs
+packages/blade/src/components/SideNav/SIDENAV_EXPLAINED.md
```
saurabhdaware: @rzp-slash-public revert this gitignore change from this PR

packages/blade/src/components/SideNav/SideNavItems/SideNavLink.web.tsx
```diff
@@ -21,13 +27,39 @@ import { BaseText } from '~components/Typography/BaseText';
...
 import { getFocusRingStyles } from '~utils/getFocusRingStyles';
 
-const { SHOW_ON_LINK_HOVER, HIDE_WHEN_COLLAPSED, STYLED_NAV_LINK } = classes;
+const { SHOW_ON_LINK_HOVER, HIDE_WHEN_COLLAPSED, STYLED_NAV_LINK, PRESSED } = classes;
+
+// Direct DOM class mutation — no re-renders. Toggles the press class on the container so
+// the CSS below can shrink the inner content (icon, text, badge, chevron, trailing button)
+// while the row itself stays at full size.
+const navLinkPressHandlers = {
+  onPointerDown: (e: React.PointerEvent) => e.currentTarget.classList.add(PRESSED),
```
saurabhdaware: @rzp-slash-public can you explain me why javascript is needed here? can the same thing be done in CSS? any pros and cons of either approaches?

saurabhdaware: @himanshusharma-rzp can you check the CSS approach if it works for you? you can also just comment here like this to fix it if overall things look good to you https://github.com/razorpay/blade/pull/3548#discussion_r3426482295

saurav12 (dismissed):
Lgtm

saurav12 (approved):

saurabhdaware (changes requested):
https://github.com/razorpay/blade/pull/3548#discussion_r3426723652

@himanshusharma-rzp there's one comment that is yet to be resolved. can you check it?

PR #3552, opened 2026-06-17
fix(rn): vertically center placeholder/value text in Dropdown trigger

rohitr-raz:
## Summary

- Fixes placeholder and selected value text appearing at the top of the `SelectInput` / `AutoComplete` dropdown trigger on React Native instead of being vertically centered
- Root cause: `BaseInputTagSlot.native.tsx` renders a `BaseBox` with `flexDirection="row"` and `minHeight` for the dropdown trigger, but had no `alignItems` set — defaulting to `stretch`, which caused the inner `ScrollView` to not fill the container height correctly, making `contentContainerStyle.alignItems: 'center'` ineffective
- Fix: add `alignItems="center"` to the outer `BaseBox` so the ScrollView (and its text content) is vertically centered within the `minHeight` container

## Test plan

- [ ] Open Dropdown → SelectInput story in React Native Storybook
- [ ] Verify placeholder text is vertically centered in the input
- [ ] Select an option and verify the selected value text is vertically centered
- [ ] Verify multiselect with tags still renders correctly
- [ ] Verify AutoComplete trigger text is also centered


rohitr-raz:
😕

PR #3633, opened 2026-06-30
fix(blade-svelte): add Playground stories for working Storybook Controls

rohankokane-dev:
## Summary
- Adds an args-driven `Playground` story as the first story for 8 Svelte components whose Controls panel was non-functional
- Root cause: stories used `asChild` with hardcoded props and never spread `defineMeta` args into the component
- Static matrix/showcase stories are unchanged

## Components fixed
- Chip, ChipGroup
- Breadcrumb
- Card, Card Interactive
- Accordion
- AvatarGroup
- BottomSheet

## Test plan
- [ ] Open Storybook for each component above
- [ ] Select **Playground** (default autodocs story)
- [ ] Change Controls (`type`, `variant`, `size`, `color`, etc.) and confirm the rendered component updates
- [ ] Confirm static matrix stories still render as before

### DIFF
--- a/packages/blade-svelte/src/components/Accordion/Accordion.stories.svelte
+++ b/packages/blade-svelte/src/components/Accordion/Accordion.stories.svelte
@@ -6,7 +6,11 @@
     title: 'Components/Accordion',
     component: Accordion,
     tags: ['autodocs'],
-    args: {},
+    args: {
+      variant: 'transparent',
+      size: 'large',
+      showNumberPrefix: false,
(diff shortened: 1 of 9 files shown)


packages/blade-svelte/src/components/Accordion/Accordion.stories.svelte
saurabhdaware, line 1: @razorpay/slash-public push empty commit on this PR

PR #3686, opened 2026-07-08
feat(rn): add React Native support for StepGroup

rohitr-raz:
## Summary

Adds React Native support for the `StepGroup` component family (previously web-only).

- Adds native implementations for `StepGroup.native.tsx`, `StepItem.native.tsx`, and `StepLine.native.tsx` — vertical & horizontal orientations, nesting, markers, and step progress.
- **Layout parity fix:** `StepItem` children (e.g. a trailing `Button`) centered on native because the native Blade Button has `alignSelf: 'center'`, which overrides the parent's `alignItems`. The children container now renders as a `flex-start` row (with wrap), so children left-align like on web.
- Adds native unit tests + snapshots.

https://github.com/user-attachments/assets/97912ef0-06a2-4b41-82ef-558811f9603f

## Test plan

- [x] `yarn test:react-native StepGroup` passes
- [x] Visually verified on iOS simulator — `Submit Documents` and other children now left-aligned (matches web)
- [ ] CI green


packages/blade/src/components/StepGroup/StepItem.native.tsx
```diff
@@ -1,14 +1,219 @@
-import type { StepItemProps } from './types';
```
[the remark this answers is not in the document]
rohitr-raz: done

packages/blade/src/components/StepGroup/StepItem.native.tsx
```diff
@@ -1,14 +1,235 @@
...
+              ? 'surface.text.gray.disabled'
+              : isSelected
+              ? 'surface.text.primary.normal'
+              : titleColor ?? 'surface.text.gray.subtle'
+          }
+          weight={isNested ? 'regular' : 'medium'}
+        >
+          {title}
+        </Text>
+        {timestamp ? (
```
rohankokane-dev: Native conditionally renders `timestamp`/`description` (`{timestamp ? ... : null}`), collapsing the vertical space when omitted. Web (`StepItem.web.tsx`) renders these unconditionally, reserving blank space via `marginY` even when the prop is absent — so a list mixing items with/without timestamps looks less aligned on native than web. Since this PR is framed as a layout-parity fix, worth deciding intentionally: match web's unconditional render, or note this as a deliberate native improvement in the PR description.

packages/blade/src/components/StepGroup/tokens.ts
```diff
@@ -5,6 +5,11 @@ const markerLineDotWidth = sizeTokens['2'];
 const markerLineDotSpacing = sizeTokens['4'];
 const markerLineWidth = sizeTokens['2'];
 const itemTopMargin = sizeTokens['6'];
+// Fixed length of a single horizontal connector segment (each side of the marker).
+// Native cannot flex-grow this segment because the horizontal ScrollView gives an
```
rohankokane-dev, line 9: New `markerLineHorizontalLength` token has no isolated test — it's only indirectly exercised via the general horizontal-orientation snapshot test. Low risk given it's a single constant, but flagging for completeness.

packages/blade/src/components/StepGroup/StepItem.native.tsx
rohitr-raz: matched webs unconditional render

packages/blade/src/components/StepGroup/tokens.ts
rohitr-raz: Added an isolated test for the markerLineHorizontalLength token.

packages/blade/src/components/StepGroup/StepItem.native.tsx
```diff
@@ -1,14 +1,235 @@
...
+  );
+
+  const enhancedMarker = marker
+    ? React.cloneElement(marker, {
+        isDisabled: isDisabled ?? marker?.props?.isDisabled,
+      })
+    : undefined;
+
+  const headerContent = isInteractive ? (
+    <Pressable
```
rohankokane-dev, line 172: **Parity gap (a11y + interactive surface):** Web `StepItem.web.tsx` renders the interactive header as `as={href ? 'a' : 'button'}` (see `InteractiveItemHeaderBox`) and sets `width: '100%'`. Web tests assert `getByRole('link')` / `getByRole('button')`.

Native `Pressable` has neither:
1. `makeAccessible({ role: href ? 'link' : 'button' })` (or `accessibilityRole`) — VoiceOver/TalkBack won't match web link vs button semantics.
2. Full-width press target — wrap content `Box` should stretch (`width="100%"` / `alignSelf: 'stretch'`) so the tappable area matches web's full header column.

**Action:** Mirror BaseLink native — spread `makeAccessible({ role: href ? 'link' : 'button', disabled: isDisabled })` on `Pressable`, and make the pressable header fill the header column width. Add a native test parallel to web's href→link / onClick→button role checks.

packages/blade/src/components/StepGroup/StepItem.native.tsx
```diff
@@ -1,14 +1,235 @@
...
+      {stepItemHeaderJSX}
+    </Box>
+  );
+
+  return (
+    <BaseBox
+      display="flex"
+      flexDirection={isVertical ? 'row' : 'column'}
+      gap={itemLineGap[size]}
+      alignItems={isVertical ? undefined : 'center'}
```
rohankokane-dev, line 194: **Parity gap (horizontal typography):** Web `StepItem.web.tsx` sets `textAlign={isVertical ? 'left' : 'center'}` on the root item so horizontal steps center title/timestamp/description.

Native sets `alignItems="center"` for the column but never centers text. On RN, `textAlign` does not inherit from a parent `View` the way CSS does on web — titles stay start-aligned.

**Action:** When `!isVertical`, pass `textAlign="center"` to the title / timestamp / description `Text` nodes (and ensure the text column stretches to item width so centering is visible).

packages/blade/src/components/StepGroup/StepItem.native.tsx
rohankokane-dev: **Parity gap (layout when props omitted):** Still present on HEAD — native conditionally mounts timestamp/description (`{timestamp ? … : null}` / `{description ? … : null}`).

Web `StepItem.web.tsx` always renders both `Text` nodes (timestamp keeps `marginY="spacing.2"` even when empty), so vertical rhythm stays stable when only some items have timestamp/description.

**Action:** Match web — always render both `Text` components unconditionally so spacing/alignment stays consistent across steps in a group.

packages/blade/src/components/StepGroup/StepGroup.native.tsx
```diff
@@ -1,14 +1,175 @@
...
+      {...makeAnalyticsAttribute(rest)}
+    >
+      {childrenWithIndex}
+    </BaseBox>
+  );
+
+  if (isHorizontal && _nestingLevel === 0) {
+    return (
+      <StepGroupContext.Provider value={contextValue}>
+        <ScrollView horizontal showsHorizontalScrollIndicator={false}>
```
rohankokane-dev: **Parity gap (`ref` + styled props on horizontal root):** Web `StepGroup.web.tsx` puts `ref`, `getStyledProps(rest)`, `width`/`maxWidth`, and `overflowX: 'auto'` on the **same** root `BaseBox` (the scroll container).

Native wraps content in `ScrollView` but leaves `ref` + styled props on the **inner** `BaseBox`. Consequences vs web:
1. `ref` does not point at the scrollable host (can't measure/scroll the horizontal scroller).
2. Margin / other styled props apply inside the scroller, not around it — outer spacing diverges from web.

**Action:** Wrap `ScrollView` in a `BaseBox` that receives `ref` + `getStyledProps(rest)` (and keep inner content box for flex row), or forward `ref` to `ScrollView` and move margin/styled props to an outer wrapper so the public root behaves like web.

packages/blade/src/components/StepGroup/StepItem.native.tsx
```diff
@@ -1,14 +1,235 @@
...
+  titleColor,
+  timestamp,
+  description,
+  stepProgress = 'none',
+  marker,
+  trailing,
+  isSelected,
+  isDisabled,
+  href,
+  onClick,
```
rohankokane-dev, line 64: **Parity gap (`target` prop dropped):** Web `StepItem.web.tsx` destructures `target` and passes it to the interactive `<a>` (`href` + `target`).

Native omits `target` from the destructure, so it falls into `…rest` → `makeAnalyticsAttribute(rest)` and is never applied. `Linking.openURL` also has no `target` equivalent, but the prop is still part of the shared public `StepItemProps` API.

**Action:** Explicitly destructure `target` (even if unused on native) so it is not forwarded into analytics attrs, and document in a short comment that `target` is web-only / no-op on native (same pattern as other cross-platform link components). Optionally no-op with a `__DEV__` note if both `href` + `target` are set.

rohankokane-dev (commented):
Parity review: StepGroup — 5 inline comments

Compared native impl against web `StepGroup.web.tsx` / `StepItem.web.tsx` / `StepLine.web.tsx`. Comments are web↔RN public API / UX / a11y parity gaps only (intentional ScrollView/fixed-width horizontal layout choices left alone).

packages/blade/src/components/StepGroup/StepItem.native.tsx
rohitr-raz: The “falls into …rest → makeAnalyticsAttribute(rest)” part is incorrect: makeAnalyticsAttribute only keeps data-analytics* / elementtiming* keys, so target is dropped, not forwarded.

packages/blade/src/components/StepGroup/StepItem.native.tsx
rohitr-raz: done

packages/blade/src/components/StepGroup/StepItem.native.tsx
rohitr-raz: done

packages/blade/src/components/StepGroup/StepItem.native.tsx
rohitr-raz: done

packages/blade/src/components/StepGroup/StepGroup.native.tsx
rohitr-raz: done

packages/blade/src/components/StepGroup/StepGroup.native.tsx
```diff
@@ -1,14 +1,190 @@
...
+              };
+              return React.cloneElement(nestedChild as React.ReactElement, {
+                /*
+                  Collapsible aligns its children to the start (alignItems: 'flex-start'), so the
+                  body shrink-wraps to its content width by default. StepItem relies on width: '100%'
+                  to fill the row and let its header column flex-grow — without a definite full width
+                  the header collapses to 0 and only the markers stay visible. Forcing the body to
+                  full width gives StepItem a definite width to resolve against.
+                */
+                width: nestedChildProps.width ?? '100%',
```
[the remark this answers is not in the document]
rohitr-raz: Fixed in d1c6eb5c6 — `stepItemIndex` is now a mutable `{ current }` counter so Collapsible recursion propagates increments to siblings after it. Nested StepGroups still get a fresh counter.

rohankokane-dev (approved):

PR #3687, opened 2026-07-08
feat(rn): add React Native support for FilterChipSelectInput

rohitr-raz:
## Summary

Adds React Native support for the `FilterChipSelectInput` component (previously web-only, rendered a "not yet implemented for native" stub).

- Adds native `FilterChipSelectInput.native.tsx` that renders inside a `Dropdown` on React Native and registers the `FilterChipSelectInput` `componentId` so `Dropdown`'s child validation passes (this was the cause of the "invalid Dropdown child" render error on native).
- Adds native `FilterChipGroupContext.native.tsx` and `ListViewFiltersContext.native.tsx`.
- Extends `BaseFilterChip.native.tsx` for the select-input trigger behaviour.
- Uses the BottomSheet-based selection flow on native (native has no dropdown menu).
- Adds native unit tests + snapshot.

https://github.com/user-attachments/assets/9c6d97df-5998-4f38-852e-2909ad60ab47

## Known limitations / out of scope

- `FilterChipGroup` is **not** implemented on native — `FilterChipGroup.native.tsx` still throws `"FilterChipGroup is not yet implemented for native"`. This PR only adds native support for a standalone `FilterChipSelectInput` inside a `Dropdown`; the grouped `FilterChipGroup` wrapper usecase remains unsupported on native and should be tracked as a separate follow-up.

## Test plan

- [x] `FilterChipSelectInput.native.test.tsx` — 8 tests + 1 snapshot passing (`jest -c jest.native.config.js`). Covers default render, open-on-press, single-select, multi-select, clear button flow, controlled value, disabled, and testID.
- [x] Verified on iPhone 17 Pro Max simulator: "Controlled Filter Chip Select Input – Single" story renders correctly (trigger + chevron + clear button), no render error.
- [ ] Reviewer: verify selection flow via BottomSheet on device/simulator.

Part of the RN migration batch alongside ButtonGroup (#3685) and StepGroup (#3686).

### DIFF
--- a/packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
+++ b/packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
@@ -61,8 +61,10 @@ const _FilterChipSelectInput = (props: FilterChipSelectInputProps): React.ReactE
 
   const isUnControlled = options.length > 0 && props.value === undefined;
   const { listViewSelectedFilters, setListViewSelectedFilters } = useListViewFilterContext();
-  const { clearFilterCallbackTriggerer, setFilterChipGroupSelectedFilters } =
-    useFilterChipGroupContext();
+  const {
+    clearFilterCallbackTriggerer,
+    setFilterChipGroupSelectedFilters,
+  } = useFilterChipGroupContext();
 
   const getValuesArrayFromIndices = (): string[] => {
     let indices: number[] = [];
@@ -95,7 +97,7 @@ const _FilterChipSelectInput = (props: FilterChipSelectInputProps): React.ReactE
 
     if (isUnControlled) {
       if (listViewSelectedFilters[label]) {
-        const savedIndices = listViewSelectedFilters[label] as unknown as number[];
+        const savedIndices = (listViewSelectedFilters[label] as unknown) as number[];
         setSelectedIndices(savedIndices);
         const inputValue = savedIndices.map((selectionIndex) => options[selectionIndex].value);
         setUncontrolledInputValue(inputValue);
@@ -106,7 +108,7 @@ const _FilterChipSelectInput = (props: FilterChipSelectInputProps): React.ReactE
         setSelectedIndices([]);
       }
     } else if (listViewSelectedFilters[label]) {
-      const savedIndices = listViewSelectedFilters[label] as unknown as number[];
+      const savedIndices = (listViewSelectedFilters[label] as unknown) as number[];
       setSelectedIndices(savedIndices);
     } else if (valueNotEmpty && !isValueAndSelectedIndicesSynced && options.length > 0) {
       const newSelectedIndices =
--- a/packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
+++ b/packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
@@ -61,7 +61,9 @@ const _FilterChipSelectInput = (props: FilterChipSelectInputProps): React.ReactE
     changeCallbackTriggerer,
   } = useDropdown();
   const valueTitle =
-    typeof value === 'string' ? (options.find((option) => option.value === value)?.title ?? value) : undefined;
+    typeof value === 'string'
+      ? options.find((option) => option.value === value)?.title ?? value
+      : undefined;
 
   const isUnControlled = options.length > 0 && props.value === undefined;
   const { listViewSelectedFilters, setListViewSelectedFilters } = useListViewFilterContext();
--- a/packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
+++ b/packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
@@ -16,7 +16,7 @@ import { useFirstRender } from '~utils/useFirstRender';
 
 type FilterChipSelectInputProps = Pick<
   BaseFilterChipProps,
-  'value' | 'label' | 'testID' | 'onClick' | 'selectionType'
+  'onKeyDown' | 'value' | 'label' | 'testID' | 'onClick' | 'selectionType' | 'onBlur'
 > & {
   accessibilityLabel?: string;
   onChange?: (props: { name: string; values: string[] }) => void;
@@ -29,6 +29,8 @@ const _FilterChipSelectInput = (props: FilterChipSelectInputProps): React.ReactE
   const idBase = useId('filter-chip-select-input');
   const {
     onClick,
+    onBlur,
+    onKeyDown,
     accessibilityLabel,
     testID,
     value,
@@ -46,6 +48,7 @@ const _FilterChipSelectInput = (props: FilterChipSelectInputProps): React.ReactE
     options,
     selectedIndices,
     onTriggerClick,
+    onTriggerKeydown,
     dropdownBaseId,
     isOpen,
     activeIndex,
@@ -57,7 +60,8 @@ const _FilterChipSelectInput = (props: FilterChipSelectInputProps): React.ReactE
     controlledValueIndices,
     changeCallbackTriggerer,
   } = useDropdown();
-  const valueTitle = options.find((option) => option.value === value)?.title ?? value;
+  const valueTitle =
+    typeof value === 'string' ? (options.find((option) => option.value === value)?.title ?? value) : undefined;
 
   const isUnControlled = options.length > 0 && props.value === undefined;
   const { listViewSelectedFilters, setListViewSelectedFilters } = useListViewFilterContext();
@@ -74,7 +78,7 @@ const _FilterChipSelectInput = (props: FilterChipSelectInputProps): React.ReactE
       indices = selectedIndices;
     }
 
-    return indices.map((selectionIndex) => options[selectionIndex].value);
+    return indices.map((i) => options[i]?.value).filter((v): v is string => v !== undefined);
   };
 
   useEffect(() => {
@@ -99,7 +103,9 @@ const _FilterChipSelectInput = (props: FilterChipSelectInputProps): React.ReactE
       if (listViewSelectedFilters[label]) {
         const savedIndices = (listViewSelectedFilters[label] as unknown) as number[];
         setSelectedIndices(savedIndices);
-        const inputValue = savedIndices.map((selectionIndex) => options[selectionIndex].value);
+        const inputValue = savedIndices
+          .map((i) => options[i]?.value)
+          .filter((v): v is string => v !== undefined);
         setUncontrolledInputValue(inputValue);
         setFilterChipGroupSelectedFilters((prev) =>
           prev.includes(label) ? prev : [...prev, label],
@@ -120,7 +126,7 @@ const _FilterChipSelectInput = (props: FilterChipSelectInputProps): React.ReactE
       setSelectedIndices(newSelectedIndices);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isUnControlled, options]);
+  }, [isUnControlled, options, value]);
 
   const getTitleFromValue = (value: string): string => {
     const option = options.find((option) => option.value === value);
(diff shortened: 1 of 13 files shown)


rohankokane-dev (commented):
## Automated code review — major findings

- **[MAJOR][test-coverage][75%]** `packages/blade/src/components/Dropdown/__tests__/FilterChipSelectInput.native.test.tsx` — Only 5 tests (default render, open-on-press, single-select, disabled, testID). No test for `selectionType="multiple"`, no test for the clear button flow (`handleClearButtonClick`/`StyledFilterCloseButton`), no test for controlled `value` sync, no test for `ListViewFiltersContext`/`FilterChipGroupContext` persistence. `BaseFilterChip.native.tsx` (266 lines, largest new file) has no dedicated unit test at all. Fix: add coverage for multi-select, clear button, controlled value, and context persistence.

- **[MAJOR][sanity/process][70%]** `packages/blade/src/components/Dropdown/FilterChipGroup.native.tsx` (unchanged in this PR) — Still throws `"FilterChipGroup is not yet implemented for native"`. This is the common real-world wrapper for multiple `FilterChipSelectInput`s, so the primary grouped usecase remains broken on native despite the PR description implying native support is complete for this component. Fix: call out this gap explicitly in the PR description (or as a tracked follow-up) so it isn't assumed resolved.

_Full categorized review (incl. minor/nit items) available on request._

rohitr-raz:
done 😉

packages/blade/src/components/FilterChip/BaseFilterChip.native.tsx
```diff
@@ -1,13 +1,265 @@
...
+        </Box>
+        <Box flexDirection="row" alignItems="center" paddingRight="spacing.1">
+          <ChevronDownIcon size="small" color="interactive.icon.gray.muted" />
+        </Box>
+      </StyledFilterTrigger>
+      {isSelected ? (
+        <>
+          <Divider orientation="vertical" variant={isDisabled ? 'muted' : 'subtle'} />
+          <StyledFilterCloseButton
+            accessibilityLabel={`Clear ${label} value`}
```
rohankokane-dev: `accessibilityLabel` is set as a raw prop here, bypassing `makeAccessible()` (used two lines up for `StyledFilterTrigger`). This skips `accessibilityRole="button"` and the explicit `accessible: true` that `makeAccessible` sets, so the close button may not reliably announce as an interactive button to VoiceOver/TalkBack. Worth wrapping with `{...makeAccessible({ label: \`Clear ${label} value\`, role: 'button' })}` to match `StyledFilterTrigger`'s pattern.

packages/blade/src/components/FilterChip/BaseFilterChip.native.tsx
```diff
@@ -1,13 +1,265 @@
...
+const StyledFilterChip = styled(View)<{
+  theme: Theme;
+  $isSelected?: boolean;
+  $isDisabled?: boolean;
+}>(({ theme, $isDisabled, $isSelected }) => {
+  return {
+    // When unselected the dashed border is drawn via an SVG overlay (see DashedBorder) so we
+    // rely on it entirely: the native border has zero width, keeping the overlay's coordinate
+    // space aligned with the border-box and avoiding iOS's buggy dashed+radius render.
+    borderWidth: $isSelected ? ((makeBorderSize(theme.border.width.thin) as unknown) as number) : 0,
```
rohankokane-dev: `(makeBorderSize(theme.border.width.thin) as unknown) as number` — `makeBorderSize`'s native override already returns `number` directly (unlike web's `${T}px` string), so this double-cast looks like dead code, only needed because of the eslint-disable for `no-unnecessary-type-assertion`. The other call at line 253 doesn't need any cast. Worth removing the redundant cast here.

packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
```diff
@@ -1,13 +1,232 @@
...
+    setSelectedIndices,
+  ]);
+
+  useEffect(() => {
+    if (clearFilterCallbackTriggerer) {
+      handleClearButtonClick();
+    }
+  }, [clearFilterCallbackTriggerer, handleClearButtonClick]);
+
+  useEffect(() => {
```
rohankokane-dev, line 165: Native merges web's two separate `useEffect`s (both keyed on `changeCallbackTriggerer` in `FilterChipSelectInput.web.tsx`) into one. Functionally equivalent today since both fire on the same dependency in the same render pass, but it's an unnecessary structural divergence from the web source of truth — a future web-side edit to one effect won't have an obvious native counterpart to update. Worth mirroring web's two-effect split, or noting the intentional merge in a comment.

packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
```diff
@@ -1,13 +1,232 @@
...
+    if (selectionType === 'single') {
+      if (uncontrolledInputValue.length > 0) {
+        return getTitleFromValue(uncontrolledInputValue[0]);
+      }
+      return '';
+    }
+    return uncontrolledInputValue;
+  };
+
+  const handleClearButtonClick = useCallback((): void => {
```
rohankokane-dev: `handleClearButtonClick` is wrapped in `useCallback` here but is a plain function on web (`FilterChipSelectInput.web.tsx`). Harmless, but another small unforced divergence between the two "parity" implementations.

packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
rohitr-raz: mirrored web

packages/blade/src/components/FilterChip/BaseFilterChip.native.tsx
rohitr-raz: done

packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
rohitr-raz: done

packages/blade/src/components/FilterChip/BaseFilterChip.native.tsx
rohitr-raz: done

packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
```diff
@@ -1,13 +1,248 @@
...
+    selectionType,
+    isControlled,
+    setSelectedIndices,
+    controlledValueIndices,
+    changeCallbackTriggerer,
+  } = useDropdown();
+  const valueTitle =
+    typeof value === 'string'
+      ? options.find((option) => option.value === value)?.title ?? value
+      : undefined;
```
rohankokane-dev: **Parity gap — controlled multi-select chip value/display**

Web (`FilterChipSelectInput.web.tsx`) computes:
```ts
const valueTitle = options.find((option) => option.value === value)?.title ?? value;
```
When `value` is a `string[]` (controlled multi), the find fails and `?? value` still passes the **array** into `BaseFilterChip`, so `selectionType === 'multiple'` renders the Counter and clear button.

Native narrows to strings only and sets `valueTitle` to `undefined` for arrays, then falls back to `getUnControlledFilterChipValue()`. Controlled multi never syncs into `uncontrolledInputValue`, so the chip stays unselected (no counter / no clear) despite a non-empty controlled `value`.

**Fix:** Mirror web — when `Array.isArray(value)`, pass `value` through (e.g. `valueTitle = typeof value === 'string' ? (options.find(...)?.title ?? value) : value`), or otherwise derive the display value from the controlled `value` prop for multi. Add a controlled `selectionType="multiple"` test.

packages/blade/src/components/FilterChip/BaseFilterChip.native.tsx
```diff
@@ -1,13 +1,272 @@
...
+  };
+
+  return (
+    <StyledFilterChip
+      $isDisabled={isDisabled}
+      $isSelected={isSelected}
+      ref={ref}
+      onLayout={isSelected ? undefined : handleLayout}
+    >
+      <StyledFilterTrigger
```
rohankokane-dev, line 201: **Parity gap — `id` prop not forwarded to trigger**

Web `BaseFilterChip.web.tsx` passes `id={id}` onto `StyledFilterTrigger`. Native destructures `id` (L178) but never applies it to `StyledFilterTrigger`.

**Fix:** Forward `id` onto the trigger `Pressable` (and/or native accessibility id equivalent) so consumers relying on `id` for linking/tests match web.

packages/blade/src/components/FilterChip/BaseFilterChip.native.tsx
```diff
@@ -1,13 +1,272 @@
...
+        <Box flexDirection="row" alignItems="center" paddingRight="spacing.1">
+          <ChevronDownIcon size="small" color="interactive.icon.gray.muted" />
+        </Box>
+      </StyledFilterTrigger>
+      {isSelected ? (
+        <>
+          <Divider orientation="vertical" variant={isDisabled ? 'muted' : 'subtle'} />
+          <StyledFilterCloseButton
+            {...makeAccessible({ label: `Clear ${label} value`, role: 'button' })}
+            onPress={isDisabled ? undefined : () => onClearButtonClick?.()}
```
rohankokane-dev: **Parity gap — clear handler omits `{ value }` (and weakens shared types)**

Web clear button calls:
```ts
onClearButtonClick?.({ value: value ?? '' })
```
Native calls `onClearButtonClick?.()` with no payload. That forced a shared `types.ts` change from `({ value }: { value: string | string[] })` to `(props?: { value: ... })`, loosening the web API contract for all `BaseFilterChip` consumers.

`FilterChipSelectInput` replaces this handler, but direct `BaseFilterChip` usage on native still diverges.

**Fix:** Call `onClearButtonClick?.({ value: value ?? '' })` like web, and restore the required `{ value }` shape in `BaseFilterChipProps`.

packages/blade/src/components/FilterChip/BaseFilterChip.native.tsx
```diff
@@ -1,13 +1,272 @@
...
+    height: FILTER_CHIP_HEIGHT_NATIVE,
+    borderRadius: theme.border.radius.small,
+    borderStyle: 'solid',
+    backgroundColor: theme.colors.surface.background.gray.intense,
+    flexDirection: 'row',
+    alignSelf: 'flex-start',
+    // The SVG dashed border must not be clipped when unselected; the selected state clips its
+    // divider/close button to the rounded corners.
+    overflow: $isSelected ? 'hidden' : 'visible',
+    opacity: $isDisabled ? 0.5 : 1,
```
rohankokane-dev: **Parity gap — disabled visual tokens**

Web disabled state uses interactive color tokens only (border/text `interactive.*.gray.disabled`); the label stays `interactive.text.gray.subtle`, and the chip does **not** apply container `opacity`.

Native sets `opacity: $isDisabled ? 0.5 : 1` on the whole chip while still painting value text with the disabled token — double-fades the value and also fades label/icons/chevron unlike web.

**Fix:** Drop container opacity and match web’s disabled token mapping (border + value text disabled; label/icons muted/subtle as on web).

rohankokane-dev (commented):
Parity review: FilterChipSelectInput — 4 inline comments.

Compared native implementation against web FilterChipSelectInput.web.tsx / BaseFilterChip.web.tsx. Comments cover API/behavior/visual token gaps only (not RN framework nits).

packages/blade/src/components/Dropdown/FilterChipSelectInput.native.tsx
rohitr-raz: mirrored web

packages/blade/src/components/FilterChip/BaseFilterChip.native.tsx
rohitr-raz: Forwarded id onto the trigger Pressable

packages/blade/src/components/FilterChip/BaseFilterChip.native.tsx
rohitr-raz: done

packages/blade/src/components/FilterChip/BaseFilterChip.native.tsx
rohitr-raz: done

rohankokane-dev (approved):

PR #3708, opened 2026-07-10
feat(native): added React Native migration agent workflows and tooling

shivam-rzpay:
## Summary

- Adds the React Native migration agent pipeline: `plan-rn-migration`, `execute-rn-migration`, and `verify-rn-migration` agents (both `.claude` and `.agents` variants) plus the `migrate-to-rn` skill for orchestrating component migrations
- Adds RN migration rules (`.claude/rules/rn-migration.md`), report/plan/patch-request templates, and orchestrator guardrail updates
- Enables fully parallel verification with per-slot simulators, a Primary Interaction Gate to block unverified core behavior, and pre-flight/port-isolation checks
- Adds a PreToolUse hook to surface existing migration artifacts.
- Updates `patch-storybook-hermes.js` for the native Storybook workflow


.agents/skills/migrate-to-rn/SKILL.md
```diff
@@ -0,0 +1,425 @@
...
+---
+
+## Step 6: Create PR
+
+> ⛔ **Human approval required.** Do NOT run any command in this step (`git add`, `git commit`, `git push`, `gh pr create`) unless the user explicitly approved at the Final Gate in Step 5. If approval was not given in this conversation, stop and ask. This applies per component — approval for one component does not cover another.
+
+For each component that passed:
+
+```bash
+cd {Worktree} && git add -A
```
swapnil-kr1: `git add -A` here stages **everything** in the worktree, including the migration artifacts written to `{Worktree}/.claude/artifacts/{Name}/` — the discovery/plan/verification reports and all `screenshots/*.png`. Neither `.claude/artifacts/` nor `.claude/worktrees/` is in `.gitignore` (`git check-ignore` returns nothing for them on this branch), so every component PR would carry its internal reports and screenshots.

Suggest either adding these paths to `.gitignore`, or staging only the component source explicitly, e.g.:
```bash
cd {Worktree} && git add packages/blade/src/components/{Name}
```

.agents/skills/migrate-to-rn/SKILL.md
```diff
@@ -0,0 +1,425 @@
...
+```bash
+ls packages/blade/src/components/{Name}/ > /dev/null 2>&1 && echo "EXISTS" || echo "NOT FOUND"
+```
+
+If not found → report error, skip that component.
+
+### 0.3 Check Native Status
+
+```bash
+grep -rl "throwBladeError" packages/blade/src/components/{Name}/ --include="*.native.tsx" | head -5
```
swapnil-kr1: This coarse gate treats any `.native.tsx` that contains `throwBladeError` as a stub → "needs migration". But `plan-rn-migration.md` explicitly warns that already-implemented components (`Tabs.native.tsx`, `Avatar.native.tsx`, `AccordionButton.native.tsx`) contain `throwBladeError` purely for input validation and are **not** stubs.

So a fully-migrated component can be mis-classified here as needing migration, and the orchestrator would spin up a worktree + Plan agent for it unnecessarily. Consider reusing the plan agent's own "genuine-stub detection" (body-shape / line-count heuristic) instead of bare string presence, so the entry gate and the Plan phase agree.

.claude/settings.json
```diff
@@ -0,0 +1,17 @@
+{
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Agent",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "input=$(cat); sub=$(printf \"%s\" \"$input\" | jq -r \".tool_input.subagent_type // \\\"\\\"\"); printf \"%s\" \"$sub\" | grep -qE \"(plan|execute|verify)-rn-migration\" || exit 0; prompt=$(printf \"%s\" \"$input\" | jq -r \".tool_input.prompt // \\\"\\\"\"); listing=$(printf \"%s\" \"$prompt\" | grep -oE \"[^ \\\"]*/\\.claude/artifacts/[A-Za-z0-9_-]+\" | sort -u | while read -r d; do [ -d \"$d\" ] && { printf \"=== %s ===\\n\" \"$d\"; ls -1 \"$d\" 2>/dev/null; printf \"\\n\"; }; done); [ -n \"$listing\" ] || exit 0; printf \"%s\" \"$listing\" | jq -Rs '{hookSpecificOutput:{hookEventName:\"PreToolUse\",additionalContext:(\"Existing RN migration artifacts already on disk — READ these before spawning; decide the next phase from disk state, not from turn history:\\n\\n\" + .)}}'",
```
swapnil-kr1: This registers a repo-level `PreToolUse` hook that fires on **every** `Agent` tool call for **anyone** who opens blade in Claude Code — not only during an RN migration. It early-exits for non-`*-rn-migration` subagents, but it still shells out to `jq` + `grep` on each Agent spawn and assumes `jq` is installed on the contributor's machine (a missing `jq` makes the hook error rather than no-op).

Two things to confirm:
- Is an always-on, shared hook the intent, vs. shipping it alongside the `migrate-to-rn` skill/plugin so it's opt-in for people actually running the migration?
- If `jq` is a hard dependency, worth documenting it (and guarding with a `command -v jq` check so the hook degrades gracefully).

rohitr-raz: 1)kept it in the shared .claude/settings.json because Claude Code doesn't give a clean way to attach a hook to a skill/plugin lifecycle — a repo-level hook is the only reliable place for it to fire during orchestration.

2)I'll add a command -v jq >/dev/null 2>&1 || exit 0 guard at the very top so the hook degrades gracefully, and document jq as a prerequisite for the migration workflow.

.agents/skills/migrate-to-rn/SKILL.md
rohitr-raz: updated the skill to reuse plan agents own "genuine-stub detection"

.agents/skills/migrate-to-rn/SKILL.md
rohitr-raz: done

swapnil-kr1 (approved):

PR #3709, opened 2026-07-10
feat(DatePicker): add visibleMonth/defaultVisibleMonth to anchor calendar independent of value

rkdotxyz:
## Summary
- Adds `visibleMonth` (controlled), `defaultVisibleMonth` (uncontrolled), and `onVisibleMonthChange` to `DatePicker`/`DateRangePicker`, letting the calendar render a specific month without pre-selecting a date.
- Previously the rendered month always fell back to `selectedValue[0]`/`value` (or today) — there was no way to anchor the calendar independently, so a "comparison" range picker could not default to the period before a primary range picker's selection without also filling in a value.
- `currentDate` priority in `Calendar.web.tsx` is now: `visibleMonth`/`defaultVisibleMonth` → existing internal `date`/`defaultDate` → `selectedValue` → today. Navigation (next/prev month/year/decade) keeps both the legacy internal state and the new `visibleMonth` state in sync so behavior is unaffected when the new prop isn't used.
- Fully backwards compatible — no behavior change when the new props are omitted.

## Motivating use case
A "comparison" range picker that should default its visible calendar to the same-length period immediately preceding a primary range picker's selection, while leaving the comparison `value` empty for the user to pick:

```tsx
<DatePicker
  selectionType="range"
  label={{ start: 'Compare to' }}
  defaultVisibleMonth={dayjs(primaryStart)
    .subtract(rangeLengthInDays + 1, 'day')
    .toDate()}
/>
```

See the new **"Comparison Range (visibleMonth)"** story in `DatePicker.stories.tsx` for a live example with a primary + comparison picker pair.

## Test plan
- [x] New unit tests in `__tests__/DatePicker.visibleMonth.web.test.tsx`: `defaultVisibleMonth` anchors the calendar with no `value` set; `visibleMonth` takes priority over the first date of a controlled `value`.
- [x] Full `DatePicker` test suite passes (`yarn test:react` — 10/10 suites green), including existing interaction test stories, unaffected.
- [x] `tsc --noEmit` shows no new type errors introduced (pre-existing unrelated storybook/vite module-resolution errors are unchanged).
- [x] Lint clean on new code (pre-existing `import/extensions` warning on `renderWithTheme.web` import matches existing test file convention, not introduced by this change).
- [ ] Manual Storybook check of the new story (`Components/DatePicker` → "Comparison Range (visibleMonth)") — recommend a maintainer eyeball this visually before merge.

Changeset included (`minor`, new non-breaking prop).

### DIFF
--- a/packages/blade/src/components/DatePicker/types.ts
+++ b/packages/blade/src/components/DatePicker/types.ts
@@ -82,9 +82,10 @@ type CalendarProps<SelectionType extends DateSelectionType> = Pick<
    */
   defaultVisibleMonth?: Date;
   /**
-   * Callback which fires when the rendered month changes via calendar navigation
-   * (next/previous month, year, or decade). Does not fire when the `visibleMonth`
-   * prop is updated externally by the consumer.
+   * Callback which fires when the rendered month changes, either via calendar
+   * navigation (next/previous month, year, or decade) or when a date in a
+   * different month is selected. Does not fire when the `visibleMonth` prop
+   * is updated externally by the consumer.
    */
   onVisibleMonthChange?: (date: Date) => void;
 
--- a/packages/blade/src/components/DatePicker/types.ts
+++ b/packages/blade/src/components/DatePicker/types.ts
@@ -60,6 +60,31 @@ type CalendarProps<SelectionType extends DateSelectionType> = Pick<
   onPickerChange?: (picker: PickerType) => void;
 
   /**
+   * Controlled month that the calendar renders, independent of the selected `value`.
+   *
+   * Useful when you want the calendar to open on a specific month without pre-selecting
+   * a date — e.g. a "comparison" range picker that should default to the period
+   * immediately preceding a primary range picker's selection.
+   *
+   * Falls back to the first date of `value`/`defaultValue` (or today) when not set.
+   *
+   * @example
+   * // Anchor a comparison picker to the month before the primary range's start
+   * <DatePicker selectionType="range" visibleMonth={dayjs(primaryStart).subtract(1, 'month').toDate()} />
+   */
+  visibleMonth?: Date;
+  /**
+   * Uncontrolled variant of `visibleMonth`. Sets the initial month the calendar renders,
+   * without pre-selecting a date.
+   */
+  defaultVisibleMonth?: Date;
+  /**
+   * Callback which fires when the rendered month changes, whether via the calendar's
+   * next/previous navigation or programmatically through `visibleMonth`.
+   */
+  onVisibleMonthChange?: (date: Date) => void;
+
+  /**
    * Controlled isOpen state
    */
   isOpen?: boolean;
--- a/packages/blade/src/components/DatePicker/types.ts
+++ b/packages/blade/src/components/DatePicker/types.ts
@@ -68,9 +68,8 @@ type CalendarProps<SelectionType extends DateSelectionType> = Pick<
    *
    * Falls back to the first date of `value`/`defaultValue` (or today) when not set.
    *
-   * When `visibleMonth` is controlled, keep it in sync via the existing `onChange`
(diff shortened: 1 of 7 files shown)


packages/blade/src/components/DatePicker/types.ts
```diff
@@ -59,6 +59,36 @@ type CalendarProps<SelectionType extends DateSelectionType> = Pick<
...
+   * without pre-selecting a date.
+   */
+  defaultVisibleMonth?: Date;
+  /**
+   * Callback which fires when the rendered month changes, either via calendar
+   * navigation (next/previous month, year, or decade) or when a date in a
+   * different month is selected. Does not fire when the `visibleMonth` prop
+   * is updated externally by the consumer.
+   */
+  onVisibleMonthChange?: (date: Date) => void;
```
swapnil-kr1, line 90: Lets rely on: onChange

rkdotxyz: Done — removed the controlled `visibleMonth` prop and `onVisibleMonthChange` callback entirely (commit 0cd3418). Kept the uncontrolled `defaultVisibleMonth` anchor; consumers can track rendered-month changes via the existing `onChange`/`onNext`/`onPrevious` callbacks instead.

packages/blade/src/components/DatePicker/types.ts
```diff
@@ -59,6 +59,31 @@ type CalendarProps<SelectionType extends DateSelectionType> = Pick<
...
+  /**
+   * Uncontrolled variant of `visibleMonth`. Sets the initial month the calendar renders,
+   * without pre-selecting a date.
+   */
+  defaultVisibleMonth?: Date;
+  /**
+   * Callback which fires when the rendered month changes, whether via the calendar's
+   * next/previous navigation or programmatically through `visibleMonth`.
+   */
+  onVisibleMonthChange?: (date: Date) => void;
```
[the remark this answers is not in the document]
rkdotxyz: Follow-up: onVisibleMonthChange has since been removed entirely per review feedback (commit 0cd3418) — this external-sync question is now moot. Use onChange/onNext/onPrevious to track the anchor.

packages/blade/src/components/DatePicker/types.ts
```diff
@@ -59,6 +59,35 @@ type CalendarProps<SelectionType extends DateSelectionType> = Pick<
...
+   */
+  visibleMonth?: Date;
+  /**
+   * Uncontrolled variant of `visibleMonth`. Sets the initial month the calendar renders,
+   * without pre-selecting a date.
+   */
+  defaultVisibleMonth?: Date;
+  /**
+   * Callback which fires when the rendered month changes via calendar navigation
+   * (next/previous month, year, or decade). Does not fire when the `visibleMonth`
```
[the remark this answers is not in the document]
rkdotxyz: Follow-up: this JSDoc clause is moot now — onVisibleMonthChange was removed entirely in favor of the existing onChange/onNext/onPrevious callbacks (commit 0cd3418).

packages/blade/src/components/DatePicker/types.ts
rkdotxyz: Correction to my earlier reply: I've scoped this properly now (commit b93d488) — only `onVisibleMonthChange` is removed. Both `visibleMonth` (controlled) and `defaultVisibleMonth` (uncontrolled) are kept. Consumers keep a controlled `visibleMonth` in sync via the existing `onChange` (selection) and `onNext`/`onPrevious` (navigation) callbacks.

rkdotxyz: Update: reverted the earlier changes (commit 5fdee58) — `onVisibleMonthChange` and the full `visibleMonth` API are restored to the pre-review state. Reopening this thread so the suggestion remains open for discussion.

saurabhdaware (approved):

PR #3720, opened 2026-07-13
feat(SankeyChart): add React Native support

rohitr-raz:
## Summary

Adds React Native support for `SankeyChart` via a new `SankeyChart.native.tsx` implementation:

- SVG-based native rendering of nodes and ribbons from the computed Sankey layout
- Coordinate hit-testing for nodes/links
- Tap-to-stay tooltip overlay
- Isolates the recharts tooltip payload type behind `Platform.Select` so the native build never depends on recharts types at the field level (the shared `types.ts` retains a type-level import of `TooltipContentProps` from recharts, consistent with the pattern used in `DonutChart/types.ts` — the native `payload` variant resolves to `undefined`, so recharts types are never referenced at runtime on native)

Also includes native unit tests + snapshots.

## Test plan

- [x] Native unit tests pass (`SankeyChart.native.test.tsx`)
- [x] Typecheck clean
- [x] Verified on iOS simulator: node/ribbon rendering, coordinate hit-testing, and tap-to-stay tooltip


packages/blade/src/components/Charts/SankeyChart/SankeyChart.native.tsx
```diff
@@ -1,31 +1,1163 @@
...
+  j: 0.24,
+  k: 0.51,
+  l: 0.24,
+  m: 0.87,
+  n: 0.57,
+  o: 0.58,
+  p: 0.57,
+  q: 0.57,
+  r: 0.35,
+  s: 0.5,
```
swapnil-kr1: Are we following same in web

rohitr-raz: No — web is different here.

**Web** uses canvas `measureText` via `calculateTextWidth` (`CommonChartComponents/utils.ts`), with the theme’s actual `fontFamily` / weight / size — not a hardcoded Inter advance table.

**Native** can’t do that synchronously (no DOM/canvas 2D; `onLayout`/`onTextLayout` are post-render), and chip/label layout needs widths in the sync layout pass, so we approximate with the Inter advance-width table.

packages/blade/src/components/Charts/SankeyChart/SankeyChart.native.tsx
```diff
@@ -1,31 +1,1163 @@
...
 
-const ChartSankey = (_props: ChartSankeyProps): React.ReactElement => {
-  throwBladeError({
-    message: 'ChartSankey is not yet implemented for native',
-    moduleName: 'ChartSankey',
+type SankeyLayout = {
+  nodeLayouts: NodeLayout[];
+  linkLayouts: LinkLayout[];
+  countPerDepth: number[];
+};
```
swapnil-kr1, line 122: move to types.ts

rohitr-raz: Moved `RechartsLink`, `NodeLayout`, `LinkLayout`, and `SankeyLayout` into `types.ts`.

swapnil-kr1 (dismissed):

swapnil-kr1 (dismissed):

swapnil-kr1 (approved):

PR #3721, opened 2026-07-13
feat(blade-svelte): BladeProvider theme context

rohankokane-dev:
## Summary
- Add `BladeProvider` with theme token injection, color scheme resolution, and `useTheme` hook
- Wire Storybook globals through `BladeThemeDecorator`
- Add BladeProvider stories and migrate dev app / existing stories to provider-based theming

## Stack
Depends on #TBD — stacked on `feat/blade-svelte-theme-core`

## Test plan
- [ ] `yarn test` in `packages/blade-svelte`
- [ ] Storybook: BladeProvider stories (light/dark/system, custom brand)
- [ ] Storybook: Card + Installation stories render with theme globals


packages/blade-svelte/src/components/BladeProvider/CreateThemeFlowExplorer.svelte
saurabhdaware, line 1: if this is related to documentation, can you maybe move this + stories inside BladeProvider/docs/? too many files on the root of BladeProvider

packages/blade-svelte/src/utils/resolveComponentStyleOverride.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    expect(
+      resolveComponentStyleOverride<ButtonSlot>(
+        'Button',
+        { root: 'instance-cta', icon: 'instance-icon' },
+        getter,
+      ),
+    ).toEqual({
+      root: 'instance-cta',
```
saurabhdaware, line 45: provider-cta is the class that we have defined on our side? if that is the case, shouldn't the rendered classname be `provider-cta instance-cta` instead of replacing the `provider-cta` with `instance-cta`?

saurabhdaware: Like will this not remove our styles completely even if they only want to override one style prop?

PR #3746, opened 2026-07-15
feat(blade-mcp): add svelte framework support (P0)

rohankokane-dev:
## Summary
- Add optional `framework` param (`react` | `svelte`, default `react`) to component and general doc tools
- Introduce `knowledgebase/svelte/` with Installation, Usage, and Button pilot docs
- Return clear availability errors when svelte component docs are missing
- Patterns remain react-only until svelte patterns exist

## Test plan
- [ ] `yarn typecheck` in packages/blade-mcp
- [ ] `yarn test` in packages/blade-mcp
- [ ] `yarn tsc:knowledgebase` at repo root
- [ ] MCP tool `get_blade_component_docs` with `framework: svelte` returns Button docs
- [ ] MCP tool with missing svelte component returns availability message


packages/blade-mcp/knowledgebase/svelte/components/Button.md
```diff
@@ -0,0 +1,124 @@
+## Component Name
```
kamleshchandnani, line 1: the usage guidelines unless its very framework related thing shall be shared right?
eg:
```
## Important Constraints

- `variant="tertiary"` can only be used with `color="primary"` or `color="white"`
- `avatars` only render on `size="large"` buttons
- `accessibilityLabel` is required for icon-only buttons
```

packages/blade-mcp/src/tools/getBladeComponentDocs.ts
```diff
@@ -11,20 +11,28 @@ import {
...
-const bladeComponentsList = getBladeDocsList('components');
-const bladeComponentsListString = bladeComponentsList.join(', ');
+const reactComponentsList = getBladeDocsList('components', 'react');
+const svelteComponentsList = getBladeDocsList('components', 'svelte');
+const reactComponentsListString = reactComponentsList.join(', ');
+const svelteComponentsListString = svelteComponentsList.join(', ');
 
 const getBladeComponentDocsToolName = 'get_blade_component_docs';
-const getBladeComponentDocsToolDescription = `Fetch the Blade Design System docs for the given list of components. Use this to get information about the components and their props while adding or changing a component.`;
+const getBladeComponentDocsToolDescription = `Fetch the Blade Design System docs for the given list of components. Use this to get information about the components and their props while adding or changing a component. Pass framework="svelte" for Blade Svelte docs (limited catalog) or framework="react" for the full React catalog. When framework is omitted, it is auto-detected from the consumer project's package.json (@razorpay/blade-svelte → svelte, @razorpay/blade → react; defaults to react).`;
```
anuraghazra, line 24: > (limited catalog) 

Why do we have to mention limited catalog etc? 

> it is auto-detected from the consumer project's package.json (@razorpay/blade-svelte → svelte, @razorpay/blade → react; defaults to react)

If framework is omited that means we could not auto-infer from the consumer project right? because only way the framework can ever be set with confidence is if the agent first checked the package.json and then set it. 

So this statement is kind of reversed. We should ask it to infer framework from package.json and then fallback to something.

anuraghazra: Wait... checked the code below. 

So we are first relaying on agent's auto inference and then falling back to the detectFrameworkFromProject()??

Should we do reverse? 🤔 

const framework = detectFrameworkFromProject() ?? framework.

PR #3811, opened 2026-07-29
fix(PhoneNumberInput): guard Intl.DisplayNames to prevent crash on import

### LINKED ISSUE #2490: TypeError: Intl.DisplayNames is not a constructor
Slack: https://razorpay.slack.com/archives/CMQ3RBHEU/p1737430650258699

MDN: [Intl.DisplayNames](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames). It states that Intl.DisplayNames was added in Chrome for Android in version 81. However, according to Sentry, the browser version is 80.x.

Image

chinmayv095:
## Description

Fixes #2490.

`CountrySelector` builds the country-name formatter at **module scope**:

```tsx
// packages/blade/src/components/Input/PhoneNumberInput/CountrySelector.web.tsx#L17
const countryNameFormatter = new Intl.DisplayNames(['en'], { type: 'region' });
```

`Intl.DisplayNames` only landed in Chrome 81, and the Sentry report in #2490 is from Chrome 80. Because the constructor runs while the module is being *evaluated* rather than when the component renders, the `TypeError` isn't scoped to the country selector — it takes down **every import of `PhoneNumberInput`**, and anything that transitively imports it. It fires even when the selector is never rendered, including with `showCountrySelector={false}`.

I verified this rather than assuming it, by importing the current `master` file with the API removed:

```
✓ ORIGINAL code throws at import time when Intl.DisplayNames is missing
✓ FIXED code imports cleanly when Intl.DisplayNames is missing
```

## Changes

- **New `countryNameFormatter.ts`** — a shared, platform-neutral module that resolves `Intl.DisplayNames` **lazily** behind a feature check, caches the result, and falls back to the country code (`"IN"`) when the API is missing or its construction throws. The `try/catch` covers engines that expose the constructor but throw for `type: 'region'` when the backing ICU data isn't bundled.
- **`CountrySelector.web.tsx` / `.native.tsx`** — both had the identical module-scope line; they now import the shared formatter, removing the duplication.
- **`PhoneNumberInput.web.tsx` / `.native.tsx`** — import from the new module. `of()` now returns `string` rather than `string | undefined`, so the two `!` assertions on `countryNameFormatter.of(countryCode)` are gone (the file-level `no-unnecessary-type-assertion` disable is still needed for the unrelated assertions below them).
- **Tests** — 5 cases covering the supported path, the missing-API fallback, the throwing-constructor fallback, single construction, and a direct regression test that importing the module does not throw. 100% coverage of the new file.
- **Changeset** — `patch`.

`countryNameFormatter` was never exported from the package (`index.ts` only exports `PhoneNumberInput` and its props type), so dropping it from `CountrySelector`'s exports is internal-only — no breaking change.

### Why a guard instead of the polyfill?

The natural alternative is to import `@formatjs/intl-displaynames` on web the way `PhoneNumberInput.native.tsx` already does. I deliberately didn't, because that polyfill plus its `en` locale data would land in the web bundle for **every** consumer in order to serve Chrome 80 — a poor trade in a bundle-size-tracked design system.

The lazy guard costs nothing at runtime and is strictly better than the status quo for teams that *do* want full names on legacy browsers: because the formatter is now resolved on first use rather than at module evaluation, a consumer-supplied polyfill is picked up as long as it's installed before the first render. With the current eager construction, the module crashes before any polyfill could help.

Happy to switch to bundling the polyfill on web instead if you'd prefer that trade-off — it's a small change on top of this.

### Behaviour on affected runtimes

| | Before | After |
|---|---|---|
| Modern browsers | `"India"` | `"India"` (unchanged) |
| React Native | `"India"` (polyfilled) | `"India"` (unchanged) |
| Chrome 80 | 💥 crash on import | `"IN"` |

## Additional Information

Verified locally against `master` (`8831cbd`):

- `yarn test:react PhoneNumberInput` → 3 suites passed (13 passed, 2 pre-existing skips)
- `yarn test:react-native PhoneNumberInput` → 1 suite passed (5 tests)
- `yarn typecheck` → clean (web + native)
- `eslint` on all 6 touched files → clean

## Component Checklist

- [ ] Add KitchenSink Story 
- [x] Add changeset

### DIFF
--- a/.changeset/phone-number-input-intl-displaynames-guard.md
+++ b/.changeset/phone-number-input-intl-displaynames-guard.md
@@ -1,0 +1,9 @@
+---
+'@razorpay/blade': patch
+---
+
+fix(PhoneNumberInput): guard `Intl.DisplayNames` so unsupported runtimes stop crashing on import
+
+`CountrySelector` built `new Intl.DisplayNames(['en'], { type: 'region' })` at module scope. On runtimes without the API — Chrome < 81, and React Native engines shipped without full ICU data — that threw a `TypeError` while the module was still being evaluated, so the failure took down every import of `PhoneNumberInput` instead of just the country selector UI. It fired even when the selector was never rendered, including with `showCountrySelector={false}`.
+
+The formatter is now resolved lazily behind a feature check and falls back to the country code (`"IN"`) when the API is unavailable, mirroring what `Intl.DisplayNames` does for unknown codes with its default `fallback: 'code'`. Resolving it lazily also means a consumer-supplied polyfill is picked up, as long as it is installed before the first render.
--- a/packages/blade/src/components/Input/PhoneNumberInput/CountrySelector.native.tsx
+++ b/packages/blade/src/components/Input/PhoneNumberInput/CountrySelector.native.tsx
@@ -13,8 +13,7 @@ import {
 import { BottomSheet, BottomSheetBody, BottomSheetHeader } from '~components/BottomSheet';
 import { ChevronUpDownIcon } from '~components/Icons';
 import { BaseBox } from '~components/Box/BaseBox';
-
-const countryNameFormatter = new Intl.DisplayNames(['en'], { type: 'region' });
+import { countryNameFormatter } from './countryNameFormatter';
 
 type CountryData = {
   code: CountryCodeType;
@@ -131,4 +130,4 @@ const CountrySelector = ({
   );
 };
 
-export { CountrySelector, countryNameFormatter };
+export { CountrySelector };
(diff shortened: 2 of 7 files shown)


chinmayv095:
Hi — checking in on this one. It's still conflict-free against `master` and CI is green.

The bug it fixes (#2490) is a crash at **import** time, not render time: because `new Intl.DisplayNames(...)` runs at module scope, the `TypeError` takes down every import of `PhoneNumberInput` and anything that transitively imports it — including when `showCountrySelector={false}` and the selector never renders.

If the size is what's holding it up, I'm happy to cut it down. The 7 files are mostly the two platform variants plus a shared helper and its tests; the minimum viable version is the lazy feature-check in `CountrySelector.web.tsx` alone, and I can split the `.native.tsx` parity and the shared `countryNameFormatter.ts` module into a follow-up. Say the word and I'll push that.

Also happy to adjust the fallback behaviour — right now it falls back to the country code (`"IN"`) when the API is missing — if you'd rather it did something else.

PR #3812, opened 2026-07-30
feat(Pagination): hide controls that cannot do anything

tejanishiraz-star:
## Problem

Reported on [`Patterns/ListView --default`](https://blade.razorpay.com/?path=/story/patterns-listview--default), and true for any `Table` with `TablePagination`:

1. **Pagination renders even when everything fits on one page.** With fewer rows than the smallest page size (10), the `10 rows / page` picker and both arrows are pure visual noise — nothing to navigate, nothing useful to pick.
2. **Boundary arrows render disabled.** On page 1 the Previous arrow is a permanently dead control; same for Next on the last page. A dead control invites clicks and communicates nothing.

## Changes

- `Pagination` / `TablePagination` render nothing when every item already fits on a single page.
- The previous arrow is unmounted on the first page and the next arrow on the last page, instead of being rendered disabled. Setting `isDisabled` is unaffected — both arrows still render, disabled.
- New optional `totalItemCount` prop on `Pagination`. It only knew `totalPages` before, so "fewer than 10 items" was not derivable: 30 items at 50 / page is also `totalPages === 1`, but there the size picker still matters because switching to 10 / page yields 3 pages. `TablePagination` forwards the table's count automatically (`totalItemCount` for server pagination, the row count from `TableContext` for client pagination), so consumers get the fix without touching their code.
- New optional `showOnSinglePage` prop (default `false`) on both `Pagination` and `TablePagination`. It gates the auto-hide so consumers that need the footer/label to always render can opt back into the previous always-render behaviour. The default behaviour is unchanged — the footer still hides when everything fits on one page.

**Hide rule:** `totalItemCount <= 10` (the smallest page size). With no count provided, `totalPages <= 1 && pageSize === 10`. On mobile the label and size picker are already hidden, so `totalPages <= 1` hides the component there too rather than leaving an orphan "Showing 1 of 1 pages". Setting `showOnSinglePage` bypasses all of these hide conditions.

`TablePagination`'s no-op `BaseBox` wrapper was dropped so a hidden pagination leaves no empty `div` in the table footer.

## Verification

- `yarn typecheck` clean. `yarn test:react Pagination Table ListView` → 53 passed, 29 snapshots. Added 7 `Pagination` tests and 1 `Table` test (fewer than 10 rows → no pagination footer); 5 `Pagination` snapshots updated.
- Real Chrome against local Storybook:
  - `Components/Table/API → TablePagination`: page 1 → no left arrow, right arrow present. Last page → left arrow present, no right arrow.
  - `Components/Pagination → Hidden When All Items Fit On One Page` (new story): renders nothing.
  - `Components/Pagination → Opt Out Of Auto-Hide With showOnSinglePage` (new story): renders the footer even with 5 items.
  - `Components/Pagination → Disabled State`: arrows still rendered, disabled.
  - `Patterns/ListView → Default` (>10 rows): unchanged.

## Notes for reviewers

- This is a **behaviour change by default**, hence the minor bump — existing consumers of `Pagination` / `TablePagination` will see the footer disappear on small datasets and the boundary arrows come and go. Layout is allowed to reflow rather than reserving space for the hidden arrow; the group is right-aligned, so the shift is small.
- Consumers that need the old always-render behaviour can set `showOnSinglePage` to opt out of the auto-hide.
- `Table/Examples/Pagination` stories are CodeSandbox embeds against published blade, so they don't exercise this change locally — `components-table-api--table-pagination-story` was used instead.

### DIFF
--- a/.changeset/pagination-hide-redundant-controls.md
+++ b/.changeset/pagination-hide-redundant-controls.md
@@ -7,3 +7,4 @@ feat(Pagination): hide pagination controls that cannot do anything
 - `Pagination` and `TablePagination` no longer render at all when every item already fits on a single page, so a table with fewer rows than a page stops showing an unusable `10 rows / page` picker and dead navigation arrows.
 - The previous page button is no longer rendered on the first page, and the next page button is no longer rendered on the last page, instead of being rendered in a permanently disabled state. Setting `isDisabled` still renders both buttons, disabled.
 - `Pagination` accepts a new optional `totalItemCount` prop so it can tell "one page because there are barely any items" apart from "one page because the page size is large". `TablePagination` forwards the table's item count automatically.
+- A new optional `showOnSinglePage` prop (default `false`) lets consumers opt out of the auto-hide and keep the pagination footer always rendered, preserving prior behaviour when needed.
--- a/.changeset/pagination-hide-redundant-controls.md
+++ b/.changeset/pagination-hide-redundant-controls.md
@@ -1,0 +1,9 @@
+---
+'@razorpay/blade': minor
+---
+
+feat(Pagination): hide pagination controls that cannot do anything
+
+- `Pagination` and `TablePagination` no longer render at all when every item already fits on a single page, so a table with fewer rows than a page stops showing an unusable `10 rows / page` picker and dead navigation arrows.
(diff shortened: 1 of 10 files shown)


tejanishiraz-star:
Split into two independently reviewable PRs, since these are two unrelated behaviour changes:

- #3813 — hide the boundary navigation buttons instead of rendering them disabled
- #3814 — hide the pagination entirely when all items fit on a single page

Same code, no behaviour dropped. Closing this one in favour of those.

PR #3874, opened 2026-08-13
fix(SpotlightPopoverTour): prevent scroll lock when anchor is taller than viewport

saurabhdaware:
## Summary
- Fixes a bug reported in Subscriptions v2 where clicking "Take a tour" freezes the page: when a step's anchor (e.g. a large table) is taller than the viewport, it can never reach the hardcoded `0.5` intersection threshold, so the tour keeps calling `scrollIntoView({ block: 'center' })`. Combined with body scroll lock + full-screen mask, this pushes the popover off-screen with no way for the user to recover.
- `useIntersectionObserver` now also observes the `0` threshold as a fallback.
- `scrollToStep` detects anchors taller than the viewport and relaxes the visibility check (`intersectionRatio > 0` instead of `isIntersecting`), aligning them to the top (`block: 'start'`) instead of the center.
- Adds a `WithLargeAnchor` story (`SpotlightPopoverTour/docs/Tour.stories.tsx`) reproducing the bug with a 200vh-tall anchor, used for verification and as a regression guard.
- Adds a changeset.

## Root cause
`IntersectionObserver` only fires when the observed ratio crosses a specified `threshold`. With only `threshold: 0.5` set, an anchor that can never occupy 50% of the viewport (because it's taller than the viewport itself) never flips `isIntersecting` to `true`, so the tour's scroll-into-view effect kept re-firing with `block: 'center'`, moving the anchor's already-off-screen edges further out of view — while body scroll was locked, making the page feel frozen.

## Test plan
- [x] Added `WithLargeAnchor` story with a 200vh anchor to reproduce the bug
- [x] Verified via Storybook + agent-browser: before the fix, starting the tour scrolled the popover fully off-screen (footer/Next button unreachable) with scroll locked
- [x] After the fix: starting the tour keeps the anchor's top edge in view with the popover's footer/Next button visible and clickable; progressing to the second (normal-sized) step behaves normally
- [x] `yarn test:react SpotlightPopoverTour` — 2 suites / 6 tests / 3 snapshots, all passing, no regressions
- [x] `yarn typecheck` — passes (web + native)

### DIFF
--- a/packages/blade/src/components/SpotlightPopoverTour/Tour.web.tsx
+++ b/packages/blade/src/components/SpotlightPopoverTour/Tour.web.tsx
@@ -17,6 +17,7 @@ import type { SpotlightPopoverTourMaskRect, SpotlightPopoverTourProps } from './
 import { SpotlightPopoverTourMask } from './TourMask';
 import { transitionDelay } from './tourTokens';
 import { useTheme } from '~components/BladeProvider';
+import { isBrowser } from '~utils';
 import { useIsomorphicLayoutEffect } from '~utils/useIsomorphicLayoutEffect';
 
 const asHTMLElement = (el: TourElement | null): HTMLElement | null => el as HTMLElement | null;
@@ -130,6 +131,8 @@ const SpotlightPopoverTour = ({
   );
 
   const scrollToStep = useCallback(() => {
+    if (!isBrowser()) return;
+
     const ref = refIdMap.get(steps[delayedActiveStep]?.name);
     const el = asHTMLElement(ref?.current ?? null);
     if (!el) return;
--- a/packages/blade/src/components/SpotlightPopoverTour/Tour.web.tsx
+++ b/packages/blade/src/components/SpotlightPopoverTour/Tour.web.tsx
@@ -47,10 +47,12 @@ const SpotlightPopoverTour = ({
   const [isScrolling, setIsScrolling] = useState(false);
 
   const currentStepRef = refIdMap.get(steps[activeStep]?.name);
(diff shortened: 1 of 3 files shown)


packages/blade/src/components/SpotlightPopoverTour/Tour.web.tsx
```diff
@@ -132,13 +134,22 @@ const SpotlightPopoverTour = ({
     const el = asHTMLElement(ref?.current ?? null);
     if (!el) return;
 
-    // If the element is already in view, don't scroll
-    if (intersection?.isIntersecting) return;
+    const viewportHeight = window.innerHeight || document.documentElement.clientHeight;
```
anuraghazra, line 140: raw window calls will fail in SSR env.

anuraghazra (dismissed):

packages/blade/src/components/SpotlightPopoverTour/Tour.web.tsx
saurabhdaware: @razorpay/slash-public fix this ^

kamaleshs-bridge4: Fixed by adding an `isBrowser()` guard from `~utils` at the top of `scrollToStep` before any `window` access, which is the existing SSR-safe pattern used across the codebase (e.g. `useDropdown.ts`, `useBreakpoint.ts`). This prevents the raw `window.innerHeight` / `document.documentElement` calls from executing in an SSR (Node) environment. [resolved by agent]

swapnil-kr1 (approved):

PR #3895, opened 2026-08-18
fix(SpotlightPopoverTour, PopupArrow): match spotlight to highlighted component, remove arrow seam, round arrow tip

urvi-shah-09:
## Summary

Three visual fixes to `SpotlightPopoverTour` (and one shared `PopupArrow` change). All three were found while building an FTUX tour on the Razorpay merchant dashboard's Analytics drill-down view — none of them is specific to that tour; every one is a default that all `SpotlightPopoverTour` consumers inherit (63 files across 8 areas of the dashboard alone).

Full write-up with token traces, composited colour values and before/after visuals: **[AI doc](https://aidocs.razorpay.com/app/d/doc_7ftr7pnc6i66jdkv)**.

All three resolve in **light and dark**, and none introduces a colour-scheme branch.

### 1. The spotlight now takes the shape of what it highlights

`TourMask` hardcoded its corner radius to `theme.spacing[2]` — 4px, and a *spacing* token used as a radius. So the spotlight was drawn at 4px around a soft-cornered `Card`, a pill, or a square button alike, and nothing in the public API could change it (`borderRadius` was a local `const`).

The mask now derives its radius from the spotlit element:

- element **has** a corner radius → the spotlight takes **the same radius**;
- element **draws no corner at all** (the step highlights plain text) → falls back to `border.radius.large`, agreeing with the popover, instead of squaring off to a hard corner.

Because `SpotlightPopoverTourStep` clones its child to attach a ref, consumers commonly wrap their UI in a layout element just to forward one. Measuring that wrapper is wrong twice over: it has no radius to inherit, **and** a wrapper stretched by its parent's layout (`alignItems="stretch"` in a flex row) is taller than the component inside it — which left the spotlight's padding uneven, 6px on three sides and 18px at the bottom.

So the spotlight resolves its target once and uses it for **both geometry and radius**: when the step's element paints nothing of its own (no radius, border or background) and wraps a single child that fills it on at least one axis, it traces that child. The padding is now even on all four sides. The fill check stops it shrinking onto an inner element that merely happens to be first — a wrapper around narrower content is still measured as the wrapper.

> [!NOTE]
> This is the target's radius **verbatim**, not `radius + padding / 2`. There is a standard concentric-geometry argument for the latter (the spotlight sits 6px outside the target), but "the spotlight matches the component" is the intended rule and the small corner offset it leaves is deliberate. Please don't add the offset back.

### 2. The popover no longer shows a line where the arrow meets the card

`TourPopover` painted the arrow with `popup.background.gray.subtle` while the card body uses `popup.background.gray.moderate` — in light mode that's `#FFFFFF` against `#F6F6F6` once composited over the tour scrim, with a hard geometric boundary between the two shapes.

This is drift, not a design choice: `Popover.web.tsx` already passes the matching token, and `popup.background.subtle` is marked `@deprecated`. Comparing majors makes the origin unambiguous — in v11 both the arrow and card used `popup.background.subtle` and there was no seam; the v11 → v12 popup-token migration moved the card to `gray.moderate` and missed the tour's arrow.

The arrow abuts the card edge with **zero overlap** (Blade passes no `strokeWidth`, so `FloatingArrow` positions it at `[side]: '100%'`), so matching the fill token makes the two composite identically.

> [!WARNING]
> Please don't "improve" this by overlapping the arrow onto the card to hide the join. `gray.moderate` is translucent (0.94), so two stacked layers composite to `1 − 0.06² = 0.9964` and produce a **brighter** hairline — trading a dark seam for a light one.

### 3. The arrow tip now reads as rounded

`tipRadius` came from `border.radius['2xsmall']` (2). `tipRadius` is a **ratio**, not a px value — `FloatingArrow` scales it against the arrow's own dimensions — so on Blade's 22×12 arrow it blunted the tip by only 1.5px, which disappears into antialiasing at 1× and reads as a hard point against the popup's 16px corners.

| `tipRadius` | rendered path | curve width | tip blunted by |
|---|---|---|---|
| `2` (before) | `M0,0 H22 L13.75,9 Q11,12 8.25,9 Z` | 5.5px | 1.5px of 12 |
| **`4` (after)** | `M0,0 H22 L16.5,6 Q11,12 5.5,6 Z` | 11px | 3px of 12 |
| `8` | `M0,0 H22 L22,0 Q11,12 0,0 Z` | 22px | degenerate — becomes a lens |

`xsmall` (4) is the largest value that still leaves straight edges.

---

## Scope notes

- **`PopupArrow` is shared**, so change 3 also restyles the `Popover` and `Tooltip` arrows. That is arguably right — one arrow, one shape — but it is the widest-reaching change here and is your call. If you'd rather scope it, `tipRadius` can be lifted to a prop; happy to split it into its own PR.
- **Change 1 is web-only.** React Native measures via `measureInWindow`, which reports geometry but no style, so there is no computed radius to read there. `SpotlightPopoverTourMaskRect.borderRadius` is optional and left `undefined` on native, which keeps its existing radius. Matching it on native needs an explicit prop — left as a follow-up.
- **Change 2 is web-only too, deliberately.** I applied it to native first and then reverted: native's `PopupArrow` draws an opaque backing path (`surface.background.gray.intense`) beneath the fill and adds a stroke, so swapping the fill token alone does
Before after_2
ˀ
Before after_1
ˀˀn't make the arrow match the card — it just moves the mismatch (dark theme: from ~2/255 darker than the card to ~3/255 lighter). Fixing native means addressing that backing path.
- **Dark mode, honestly:** the seam's tokens diverge on more axes in dark, but composited over the 80%-black dark scrim the delta is only ~2/255 (vs ~9/255 in light) and it inverts. Dark is *less* visible, not worse — the fix is asked for on token-correctness grounds, not because dark looks broken today.
2_before
3_before
4_before
1_before

## Before / after
Before after_2
Before after_1

### Before — the tour in the product

The FTUX tour on the Razorpay dashboard's Analytics drill-down view (`apps/entp-analytics`), on a dev build with no local styling on the popover or the spotlight — so this is what Blade renders by default. In every step the spotlight outline is near-square regardless of what it highlights, its padding is uneven at the bottom, and the arrow meets the card body along a visible edge.

<details>
<summary>All four steps</summary>

<br>

**Step 1 of 4 — Choose a metric.** Clearest look at the arrow join.

Tour step 1 of 4, 'Choose a metric': the blue spotlight outline around the metric trigger has near-square corners, and the popover's arrow meets the card body along a visible edge.

**Step 2 of 4 — Filter your data.** Clearest look at the radius mismatch: the Filters row it highlights has soft corners, and so does the popover beneath it, but the outline between them is near-square.

Tour step 2 of 4, 'Filter your data': a near-square blue spotlight outline around the soft-cornered Filters row, with a rounder popover directly beneath it.

**Step 3 of 4 — Split by a dimension.** Same mismatch on the Split by row.

Tour step 3 of 4, 'Split by a dimension': the same near-square spotlight outline around the Split by row.

**Step 4 of 4 — Add more chart.** Placement flips to `top`, so the arrow points down; both issues follow the placement.

Tour step 4 of 4, 'Add more chart': the popover sits above its target with the arrow pointing down, and the spotlight outline around 'Add a new chart' is near-square.

</details>

### After — measured

Side-by-side visuals, both themes, and the full token trace: **[AI doc](https://aidocs.razorpay.com/app/d/doc_7ftr7pnc6i66jdkv)**. Chromatic on this PR carries the visual diffs, and the `SpotlightPopoverTour › Basic` story reproduces it live.

That story anchors its step to a `<Box>` wrapping a `<Card>` — the layout-wrapper case — so it exercises the fall-through as well as the radius match. Measured in both builds:

| | `master` | this PR |
|---|---|---|
| Spotlight cut-out `rx` | `4` (hardcoded) | `12` |
| Highlighted `Card` computed radius | `12px` | `12px` |
| Radius match? | ❌ near-square outline cuts across the card's corners | ✅ exact |
| Gap to the card, L/R/T/B | `6 / 6 / 6 / 18` | `6 / 6 / 6 / 6` |
| Padding even? | ❌ wrapper is stretched 12px taller than the card | ✅ equidistant |
| Luminance through the arrow/body join | `231` vs a `247` body — a dark hairline | constant — no dip, no peak |
| One continuous shape? | ❌ border drawn across the join | ✅ outline traces body + arrow |

## Testing

- `yarn typecheck` — passes (web + native)
- `yarn test:react` — 2269 passed, 6 snapshots updated
- `yarn test:react-native` — 1611 passed, no snapshot changes after the native revert
- `yarn lint:blade` — 0 errors
- New unit tests cover all three radius branches: the element's own radius, a radius resolved through a layout wrapper, and the no-corner fallback.

Snapshot diffs are exactly the three intended changes — the arrow `d` path, the mask `rx`/`ry`, and the arrow `fill`. Chromatic will show the visual deltas on `SpotlightPopoverTour`, `Popover` and `Tooltip`.

> [!NOTE]
> **Unrelated pre-existing issue spotted, not fixed here:** in `Tour.web.test.tsx` the `should not have a11y violations` test calls `assertAccessible(baseElement)` without `await`, so that assertion has never actually run. It does currently fail — `TourPopover` always sets `aria-labelledby={titleId}`, but `PopoverHeader` only renders that id when a `title` is given, so a title-less step leaves a dangling `aria-labelledby`. I've left both alone to keep this PR scoped; happy to raise it separately.


saurav12:
@urvi-shah-09 Posted comment on the [aidocs](https://aidocs.razorpay.com/app/d/doc_7ftr7pnc6i66jdkv).

ankitpunia:
Discussed with Rastogi. This is approved.

saurav12:
@urvi-shah-09 I've reverted the arrow radius change, it was looking too rounded. Although, the line between the popover and the main container was a right change so retaining that one.

Put it in a stacked PR on top of this one: #3899. It also fixes something I noticed while looking at the arrow — the gap between the arrow tip and the spotlight varies per step, because the mask resolves the step's ref to the painted component while `TourPopover` still anchors to the raw ref, so any wrapper overhang lands in that gap.

saurav12 (approved):
Looks good, i reverted the arrow radius as it was looking too rounded in a stacked PR. Everything else looks good to me.

PR #3899, opened 2026-08-18
fix(SpotlightPopoverTour, PopupArrow): revert the arrow tip radius and even out the arrow-to-spotlight gap

saurav12:
Stacked on #3895 — review that one first, this PR targets its branch.

## Summary

**1. Reverted the arrow tip radius back to `2xsmall`.** In practice `xsmall` read as too rounded. `tipRadius` is a ratio scaled against the arrow's own width and height, so the effect grows much faster than the token number suggests — at `xsmall` it bows the arrow's straight edges rather than just blunting the tip. Reverting also takes the `Popover` and `Tooltip` snapshots back to their pre-existing output, so this stack no longer touches those two components at all.

The seam fix from #3895 is untouched and retained — matching the arrow's fill token and continuing the outline across the join was the right call, and it is what makes the body and the arrow read as one container.

**2. Evened out the gap between the arrow tip and the spotlight.** Nothing sets that gap directly. `TourPopover` offsets the popup by `spacing[4] + ARROW_HEIGHT` (24px) so that, after the arrow's own 12px protrusion, the tip lands `GAP` (12px) from the anchor's edge, while `TourMask` draws the halo 6px past the traced element — an intended constant 6px.

That only holds if both measure the same element, and they did not. The mask resolves the step's ref through `resolveSpotlightTarget` to trace the painted component, but `TourPopover` set its position reference to the raw ref. Any overhang between the wrapper and the component inside it was added straight to the gap, so a step whose wrapper is stretched by a flex row (`alignItems="stretch"`) sat visibly further from its spotlight than a step whose wrapper happened to match its card. `resolveSpotlightTarget` moves into the shared web utils and both sides now resolve through it.

Only `setPositionReference` uses the resolved element; `setReference` stays on the original step element, since that one is the interaction and focus target rather than the geometry source.

## Test plan

- [x] `yarn test:react SpotlightPopoverTour Popover Tooltip` — 46 tests pass
- [x] `Popover` and `Tooltip` snapshots verified byte-identical to `master`
- [x] Typecheck clean for every touched file
- [x] Storybook: step through the `SpotlightPopoverTour` default story and confirm the arrow sits the same distance from the spotlight on the Refunds step (stretched wrapper) as on the Disputes step
- [x] Confirm the arrow tip rounding matches the pre-#3895 appearance

## Notes

- The changeset from #3895 is edited rather than added to, so the merged changelog describes what actually ships. Its third section no longer claims the tip was rounded, and now documents the gap fix instead.
- React Native is unchanged. Its tour has no equivalent target resolution, so this mismatch cannot arise there in the same form.
- `TourMask`'s `padding` prop still means "total added across both axes" rather than per-side, so `spacing[4]` reads as 12 but paints 6. Left alone deliberately, but it is the remaining reason the numbers do not visually cancel.

### DIFF
--- a/.changeset/spotlight-tour-radius-seam-tip.md
+++ b/.changeset/spotlight-tour-radius-seam-tip.md
@@ -2,7 +2,7 @@
 '@razorpay/blade': minor
 ---
 
-fix(SpotlightPopoverTour, PopupArrow): match the spotlight to the highlighted component, remove the popover's arrow seam, and round the arrow tip
+fix(SpotlightPopoverTour): match the spotlight to the highlighted component, remove the popover's arrow seam, and even out the arrow-to-spotlight gap
 
 Three visual fixes. All three apply to light and dark — none introduces a colour-scheme branch.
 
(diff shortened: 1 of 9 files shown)


kamaleshs-bridge4 (dismissed):
## ✨ Agentic PR Review ✨

**Status:** Approved ✅

### UI Review

✅ 5 passed

<details>
<summary>Passing checks (5)</summary>

| Check |
|-------|
| ✅ [SpotlightPopoverTour Basic](https://61c19ee8d3d282003ac1d81c-namjbgwlbi.chromatic.com/?path=/story/components-spotlightpopovertour--basic) |
| ✅ [SpotlightPopoverTour CustomPlacement](https://61c19ee8d3d282003ac1d81c-namjbgwlbi.chromatic.com/?path=/story/components-spotlightpopovertour--custom-placement) |
| ✅ [Popover Default](https://61c19ee8d3d282003ac1d81c-namjbgwlbi.chromatic.com/?path=/story/components-popover--default) |
| ✅ [Tooltip Default](https://61c19ee8d3d282003ac1d81c-namjbgwlbi.chromatic.com/?path=/story/components-tooltip--default) |
| ✅ [Tooltip WithTitle](https://61c19ee8d3d282003ac1d81c-namjbgwlbi.chromatic.com/?path=/story/components-tooltip--with-title) |

</details>

### Usage

```jsx
import { SpotlightPopoverTour, SpotlightPopoverTourStep } from '@razorpay/blade/components';

<SpotlightPopoverTour
  steps={[
    {
      name: 'intro',
      title: 'Welcome',
      placement: 'top',
      content: () => <Text>A quick tour of this feature</Text>,
    },
  ]}
  isOpen={isOpen}
  activeStep={0}
  onStepChange={(step) => setActiveStep(step)}
  onFinish={() => setIsOpen(false)}
>
  <SpotlightPopoverTourStep name="intro">
    <Button>Get started</Button>
  </SpotlightPopoverTourStep>
</SpotlightPopoverTour>
```

saurabhdaware (dismissed):

swapnil-kr1 (approved):

PR #3965, opened 2026-09-07
feat(BarChart): compare bars against a min–max reference range

premika-pixel:
## What

`ChartBarWrapper` can now show a shaded **min–max reference range** behind its bars — the bar counterpart to the LineChart reference band ([#3824](https://github.com/razorpay/blade/pull/3824)). Built for the Optimizer "Industry SR" surface, where a merchant's daily success rate is read against the industry's 25th–75th percentile band.

Screenshot 2026-09-07 at 6 28 42 PM
Screenshot 2026-09-08 at 10 47 00 AM

Two shapes:

**1. One range for the whole chart** — always visible, with a legend swatch.

```tsx
<ChartBarWrapper data={successRateData}>
  <ChartReferenceBand lowerDataKey="industryLow" upperDataKey="industryHigh" name="Industry range" />
  <ChartXAxis dataKey="period" />
  <ChartYAxis label="Success rate (%)" />
  <ChartTooltip />
  <ChartLegend />
  <ChartBar dataKey="successRate" name="Success rate" barSize={24} />
</ChartBarWrapper>
```

**2. A range per bar** — for grouped charts, revealed on hover, one at a time.

```tsx
<ChartBar
  dataKey="upi"
  name="UPI"
  rangeLowerDataKey="upiLow"
  rangeUpperDataKey="upiHigh"
  rangeName="UPI industry range"
/>
```

| Prop on `ChartBar` | Type | Default | Description |
|---|---|---|---|
| `rangeLowerDataKey` | `string` | – | Lower (min) bound key; band shows when this and the upper key are both set |
| `rangeUpperDataKey` | `string` | – | Upper (max) bound key |
| `rangeName` | `string` | `'Industry range'` | Legend + tooltip label |
| `rangeColor` | colour token | the bar's colour | Band fill |
| `showRangeLegend` | `boolean` | `false` | Legend swatch — off by default because the band is hover-only |

## Why

`ChartReferenceBand` already existed, but only `ChartLineWrapper` had a layer that paints between its bounds — dropping it into a bar chart was a **silent no-op**, and BarChart's documented accepted-children list omitted it entirely. `ChartReferenceLine` only draws one constant line, and Recharts' `ReferenceArea` only draws a fixed rectangle (already rejected in `_decisions/decisions.md` for exactly this reason).

The prototype had to build the band *outside* the chart — an absolutely-positioned SVG sibling scraping the rendered DOM for the plot box and bar positions, plus an app-level hover hook and manual colour swapping. Not portable, and hard-coded the y-axis to `0–100`.

## How

- **y** comes from two invisible bound series, so the range folds into the y-domain and a band taller than the bars is never clipped — no `domain={[0, 100]}` required.
- **x** is re-anchored to the owning series' **bar centres**, read from the rendered bar rects. A bound series sits at the *category* centre, which would stack every band on the same x in a grouped chart.
- Edges are **straight chords** flat-extended to both plot edges — a bar chart has no trend curve for the band to follow.
- Bound series are `<Line>`, not `<Bar>`, deliberately: Recharts allocates a slot per bar series inside each category group, so two extra invisible bars would shrink and shift the real ones.

### Hover

Hovering a bar **never recolours it**. The hovered bar keeps its own colour; emphasis comes from the other series fading, and the hovered category is shaded so the revealed band reads against its own bars.

### Tooltip

- A standalone `<ChartReferenceBand>` now contributes its bounds to the tooltip. Previously only a series' own `range*` props did, so a chart-level band drew the shaded area but showed no range row. A bar's own `range*` props still win.
- Both the legend swatch and the tooltip's range row take a band's default label from one shared constant. They read the name by different routes, and defaulting in only one of them left the same nameless band reading `Industry range` in the legend and `Industry` in the tooltip.

> **Previously in this PR, now removed:** an earlier revision also made `ChartTooltip` honour the `formatter` prop it accepted but ignored, rendering `62%` and a `48%–70%` range row. That is a behaviour change for **every** chart type, so per review it has been reverted out of here and follows as its own patch PR — carrying with it a fix for the arguments passed when formatting the range bounds. Values in the stories therefore render unformatted (`2000`, not `2000%`) until that lands.

## Scope

**Web only.** `BarChart.native.tsx` is untouched; React Native parity is a follow-up, and the geometry helper is written to be shared when it lands.

## Docs

- `_decisions/decisions.md` — §3.5 gains the per-bar `range*` props; §3.8.2 retitled for LineChart **and** BarChart, with the export-scope note corrected.
- `packages/blade-mcp/knowledgebase/components/BarChart.md` — its accepted-children constraint previously omitted `ChartReferenceBand`, so an AI agent asked "can a bar chart have a reference band?" would have said no. Now documents both shapes with runnable examples.

## Testing

- 10 new web tests covering: band layer for a standalone band, no layer when absent, straight-segment closed path, tooltip value + range row (pinned via `defaultIndex`), bound series per declaring bar, no band before hover, legend defaults, and the single-bound-key no-op.
- `yarn typecheck` (web + native) clean · **110** web chart tests pass · **83** native chart tests pass · lint at the pre-existing baseline (zero net new).
- Snapshots regenerated for the added container `<div>`.

## Reviewer notes

Two things I'd like a second opinion on:

1. **`showRangeLegend` default asymmetry.** It defaults to `false` here but follows the line's `showLegend` on `ChartLine`. Deliberate — line bands are always visible, bar bands are not — but worth a conscious call.
2. **Overlap with `origin/fix/chart-reference-band-review-comments`.** That branch scopes `ChartReferenceBand` out of the shared `CommonChartComponents` barrel, justified as *"using it inside a BarChart/AreaChart would silently render invisible bound lines with no visible band."* That reasoning no longer holds for BarChart. I've updated the note in `decisions.md`, but the two branches need reconciling before either merges — if that one lands first, `ChartReferenceBand` must be re-exported from **both** `LineChart` and `BarChart`.

Design spec and both states are built as a 12-variant Figma component set (`Bar Chart - Industry SR`), mirroring `Line Chart - Industry SR`.

Component AI doc - https://aidocs.razorpay.com/app/d/doc_iij4qqnsnythrhnn


packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
```diff
@@ -0,0 +1,417 @@
...
+ * Centre x of every bar rendered for one bar series, left→right.
+ *
+ * The series is found by its `barSeriesClass` rather than by position, so toggling another series
+ * off in the legend (which renders no group at all) can't shift the lookup onto the wrong bars.
+ *
+ * `ChartBar`'s custom shape draws two `<rect>`s per bar (the body fill plus a thin accent edge at
+ * the same x), so the raw list has duplicates — they're de-duplicated here, leaving one centre per
+ * category.
+ */
+const getBarCentres = (surface: Element, dataKey: string): number[] | null => {
```
kamaleshs-bridge4, line 71: 🔵 **[MINOR]** · _code-quality-critique_ · confidence: 7/10

**Problem:** Reference band geometry assumes horizontal layout — getBarCentres reads x/width from bar rects, getPlotBounds queries the x-axis line, getCategoryColumn builds a vertical column, and buildBarBandPath extends to plotLeft/plotRight. For layout='vertical' bar charts all produce incorrect positions with no guard or warning.

**Suggestion:** Add a layout guard that skips band rendering for vertical layout (and documents the limitation), or extend the geometry functions to handle vertical: read y/height, query the y-axis, build a horizontal row, extend to plotTop/plotBottom.

kamaleshs-bridge4 (approved):
## ✨ Agentic PR Review ✨

**Status:** Approved ✅

### UI Review

✅ 4 passed

<details>
<summary>Passing checks (4)</summary>

| Check | Screenshot |
|-------|------------|
| ✅ [BarChartWithReferenceBand — standalone band, legend swatch, tooltip formatter + range](https://61c19ee8d3d282003ac1d81c-zanqdosccd.chromatic.com/?path=/story/components-charts-barchart--bar-chart-with-reference-band) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/18-59-48/components-charts-barchart--bar-chart-with-reference-band-1.png"></a> |
| ✅ [GroupedBarChartWithMultipleReferenceBands — bands hidden until hover, no legend swatches](https://61c19ee8d3d282003ac1d81c-zanqdosccd.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/18-59-48/components-charts-barchart--grouped-bar-chart-4.png"></a> |
| ✅ [GroupedBarChartWithMultipleReferenceBands — hover reveals one band, shades category, fades other series](https://61c19ee8d3d282003ac1d81c-zanqdosccd.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/18-59-48/components-charts-barchart--grouped-bar-chart-4.png"></a> |
| ✅ [GroupedBarChartWithMultipleReferenceBands — showReferenceBand=false and existing GroupedBarChart regression check](https://61c19ee8d3d282003ac1d81c-zanqdosccd.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/18-59-48/components-charts-barchart--grouped-bar-chart-4.png"></a> |

</details>

### Usage

```jsx
import {
  ChartBar,
  ChartBarWrapper,
  ChartXAxis,
  ChartYAxis,
  ChartTooltip,
  ChartLegend,
  ChartReferenceBand,
} from '@razorpay/blade/components';

// Shape 1: one chart-wide band via ChartReferenceBand
<ChartBarWrapper data={data}>
  <ChartReferenceBand lowerDataKey="industryLow" upperDataKey="industryHigh" name="Industry range" />
  <ChartXAxis dataKey="period" />
  <ChartYAxis />
  <ChartTooltip formatter={(v) => `${v}%`} />
  <ChartBar dataKey="successRate" name="Success rate" />
</ChartBarWrapper>

// Shape 2: per-bar range via range* props on ChartBar
<ChartBarWrapper data={data}>
  <ChartXAxis dataKey="period" />
  <ChartYAxis />
  <ChartBar dataKey="card" rangeLowerDataKey="cardLow" rangeUpperDataKey="cardHigh" rangeName="Card range" />
  <ChartBar dataKey="upi" rangeLowerDataKey="upiLow" rangeUpperDataKey="upiHigh" rangeName="UPI range" />
</ChartBarWrapper>
```

kamaleshs-bridge4 (approved):
## ✨ Agentic PR Review ✨

**Status:** Approved ✅

### UI Review

✅ 4 passed

<details>
<summary>Passing checks (4)</summary>

| Check | Screenshot |
|-------|------------|
| ✅ [BarChart with Reference Band](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--bar-chart-with-reference-band) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/19-00-28/components-charts-barchart--bar-chart-with-reference-band-1.png"></a> |
| ✅ [Grouped BarChart with Multiple Reference Bands](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/19-00-28/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands-2.png"></a> |
| ✅ [Default BarChart (regression check)](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--default-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/19-00-28/components-charts-barchart--default-chart-3.png"></a> |
| ✅ [Grouped BarChart (regression check)](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/19-00-28/components-charts-barchart--grouped-bar-chart-4.png"></a> |

</details>

### Usage

```jsx
import { ChartBar, ChartBarWrapper, ChartReferenceBand, ChartXAxis, ChartYAxis, ChartTooltip, ChartLegend } from '@razorpay/blade/components';

// Shape 1: one range for the whole chart
<ChartBarWrapper data={successRateData}>
  <ChartReferenceBand lowerDataKey="industryLow" upperDataKey="industryHigh" name="Industry range" />
  <ChartXAxis dataKey="period" />
  <ChartYAxis label="Success rate (%)" />
  <ChartTooltip formatter={(value) => `${Number(value)}%`} />
  <ChartLegend />
  <ChartBar dataKey="successRate" name="Success rate" barSize={24} />
</ChartBarWrapper>

// Shape 2: a range per bar (grouped)
<ChartBarWrapper data={methodData}>
  <ChartXAxis dataKey="period" />
  <ChartYAxis label="Success rate (%)" />
  <ChartTooltip formatter={(value) => `${Number(value)}%`} />
  <ChartLegend />
  <ChartBar dataKey="upi" name="UPI" rangeLowerDataKey="upiLow" rangeUpperDataKey="upiHigh" rangeName="UPI industry range" />
</ChartBarWrapper>
```

packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
```diff
@@ -85,77 +109,146 @@ const _ChartBar: React.FC<ChartBarProps> = React.memo(
...
+          shape={(props: unknown) => {
+            const { fill, x, y, width, height, index: barIndex } = props as RechartsShapeProps;
+            // Two independent fades compose here: which category is hovered (activeIndex) and which
+            // series is hovered (hoveredDataKey).
+            const categoryOpacity = isNumber(activeIndex)
+              ? barIndex === activeIndex
+                ? 1
+                : NON_HOVERED_SERIES_OPACITY
+              : 1;
+            const fillOpacity =
```
kamaleshs-bridge4, line 202: 🔵 **[MINOR]** · _code-quality-critique_ · confidence: 7/10

**Problem:** The series-level hover fade (isOtherSeriesHovered) and onMouseEnter/onMouseLeave→setHoveredBar wiring are attached unconditionally to every ChartBar. This changes the hover behavior of existing grouped bar charts that don't use reference bands — previously the whole hovered category highlighted; now only the hovered series does — and triggers a wrapper re-render on every bar hover even when no band is present. The decisions doc describes this fade only under the per-bar reference-range section, so the universal application appears ungated.

**Suggestion:** Gate the hover tracking and isOtherSeriesHovered fade so they only activate when the chart declares a reference band (or the bar has range props), keeping plain bar charts on their original category-wide highlight behavior and avoiding needless re-renders.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
```diff
@@ -0,0 +1,417 @@
...
+  if (activeIndex < 0 || centres.length === 0) return null;
+  const bounds = getPlotBounds(surface);
+  const gridLines = Array.from(
+    surface.querySelectorAll<SVGLineElement>('.recharts-cartesian-grid-horizontal line'),
+  ).map((line) => Number(line.getAttribute('y1')));
+  const axisLine = surface.querySelector('.recharts-xAxis .recharts-cartesian-axis-line');
+  const baseline = Number(axisLine?.getAttribute('y1'));
+  if (!bounds || !Number.isFinite(baseline)) return null;
+
+  const top = gridLines.length > 0 ? Math.min(...gridLines) : 0;
```
kamaleshs-bridge4, line 145: 🔵 **[MINOR]** · _code-quality-critique_ · confidence: 6/10

**Problem:** Hovered-category highlight column uses grid-line y as the plot top, falling back to 0 when ChartCartesianGrid is absent — so the shaded rect extends from the SVG origin down to the x-axis instead of from the plot top. The PR's own grouped-band story omits the grid, so this overshoot is demonstrable in the flagship example.

**Suggestion:** Fall back to the y-axis line's y1 (`.recharts-yAxis .recharts-cartesian-axis-line`) instead of 0 when no horizontal grid lines are present, so the highlight starts at the actual plot top.

kamaleshs-bridge4 (dismissed):
## ✨ Agentic PR Review ✨

**Status:** Approved ✅

### UI Review

✅ 8 passed

<details>
<summary>Passing checks (8)</summary>

| Check | Screenshot |
|-------|------------|
| ✅ [BarChart with standalone reference band](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--bar-chart-with-reference-band) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/20-06-36/components-charts-barchart--bar-chart-with-reference-band-2.png"></a> |
| ✅ [BarChart with standalone reference band tooltip](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--bar-chart-with-reference-band) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/20-06-36/components-charts-barchart--bar-chart-with-reference-band-2.png"></a> |
| ✅ [Grouped BarChart with per-bar reference bands](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/20-06-36/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands-4.png"></a> |
| ✅ [Grouped BarChart per-bar band hover reveal and tooltip](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/20-06-36/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands-4.png"></a> |
| ✅ [Existing Default BarChart no regression](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--default-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/20-06-36/components-charts-barchart--default-chart-5.png"></a> |
| ✅ [Existing Grouped BarChart no regression](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/20-06-36/components-charts-barchart--grouped-bar-chart-6.png"></a> |
| ✅ [Existing Stacked BarChart no regression](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-barchart--stacked-bar-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/20-06-36/components-charts-barchart--stacked-bar-chart-7.png"></a> |
| ✅ [LineChart reference band no regression](https://61c19ee8d3d282003ac1d81c-qegqgktnoe.chromatic.com/?path=/story/components-charts-linechart--line-chart-with-reference-band) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/20-06-36/components-charts-linechart--line-chart-with-reference-band-8.png"></a> |

</details>

### Usage

```jsx
import { ChartBarWrapper, ChartBar, ChartReferenceBand, ChartXAxis, ChartYAxis, ChartTooltip, ChartLegend } from '@razorpay/blade/components';

// One range for the whole chart — always visible, legend swatch by default
<ChartBarWrapper data={data}>
  <ChartReferenceBand lowerDataKey="industryLow" upperDataKey="industryHigh" name="Industry range" />
  <ChartXAxis dataKey="period" />
  <ChartYAxis />
  <ChartTooltip formatter={(v) => `${Number(v)}%`} />
  <ChartLegend />
  <ChartBar dataKey="successRate" name="Success rate" />
</ChartBarWrapper>

// A range per bar — revealed on hover, no legend swatch by default
<ChartBarWrapper data={groupedData}>
  <ChartXAxis dataKey="period" />
  <ChartYAxis />
  <ChartTooltip />
  <ChartLegend />
  <ChartBar dataKey="card" name="Card" rangeLowerDataKey="cardLow" rangeUpperDataKey="cardHigh" rangeName="Card industry range" />
  <ChartBar dataKey="upi"  name="UPI"  rangeLowerDataKey="upiLow"  rangeUpperDataKey="upiHigh"  showRangeLegend />
</ChartBarWrapper>
```

packages/blade/src/components/Charts/BarChart/types.ts
```diff
@@ -44,6 +44,34 @@ type ChartBarProps = {
...
+   * Fill color of this bar's range band.
+   * @default the bar's own resolved color
+   */
+  rangeColor?: ChartsCategoricalColorToken | ChartSequentialColorToken;
+  /**
+   * Whether to show a legend swatch for this bar's range band. Off by default because the band is
+   * only on screen while its bar is hovered.
+   * @default false
+   */
+  showRangeLegend?: boolean;
```
kamaleshs-bridge4, line 74: 🔵 **[MINOR]** · _api-decision-critique_ · confidence: 7/10

**Problem:** showRangeLegend defaults to false on ChartBar but defaults to true (via showLegend) on ChartLine — same prop name, different default across chart types

**Suggestion:** Document the cross-chart default difference explicitly in the JSDoc, or align defaults if the hover-only rationale is not deemed sufficient

kamaleshs-bridge4 (commented):
## ✨ Agentic PR Review ✨

### UI Review

✅ 6 passed

<details>
<summary>Passing checks (6)</summary>

| Check | Screenshot |
|-------|------------|
| ✅ [BarChart with Reference Band](https://61c19ee8d3d282003ac1d81c-icadqvpzkl.chromatic.com/?path=/story/components-charts-barchart--bar-chart-with-reference-band) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/21-02-57/components-charts-barchart--bar-chart-with-reference-band-1.png"></a> |
| ✅ [BarChart with Reference Band — Tooltip](https://61c19ee8d3d282003ac1d81c-icadqvpzkl.chromatic.com/?path=/story/components-charts-barchart--bar-chart-with-reference-band) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/21-02-57/components-charts-barchart--bar-chart-with-reference-band-2.png"></a> |
| ✅ [Grouped BarChart with Multiple Reference Bands](https://61c19ee8d3d282003ac1d81c-icadqvpzkl.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/21-02-57/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands-3.png"></a> |
| ✅ [SimpleBarChart — Regression](https://61c19ee8d3d282003ac1d81c-icadqvpzkl.chromatic.com/?path=/story/components-charts-barchart--simple-bar-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/21-02-57/components-charts-barchart--simple-bar-chart-4.png"></a> |
| ✅ [GroupedBarChart — Regression](https://61c19ee8d3d282003ac1d81c-icadqvpzkl.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/21-02-57/components-charts-barchart--grouped-bar-chart-5.png"></a> |
| ✅ [StackedBarChart — Regression](https://61c19ee8d3d282003ac1d81c-icadqvpzkl.chromatic.com/?path=/story/components-charts-barchart--stacked-bar-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/21-02-57/components-charts-barchart--stacked-bar-chart-6.png"></a> |

</details>

### Usage

```jsx
import { ChartBar, ChartBarWrapper, ChartReferenceBand, ChartXAxis, ChartYAxis, ChartTooltip, ChartLegend } from '@razorpay/blade/components';

<ChartBarWrapper data={successRateData}>
  <ChartReferenceBand lowerDataKey="industryLow" upperDataKey="industryHigh" name="Industry range" />
  <ChartXAxis dataKey="period" />
  <ChartYAxis label="Success rate (%)" />
  <ChartTooltip formatter={(value) => `${Number(value)}%`} />
  <ChartLegend />
  <ChartBar dataKey="successRate" name="Success rate" barSize={24} />
</ChartBarWrapper>
```

urvi-shah-09:
Approved

saurav12 (commented):
Thorough review pass. Really nice PR description, the reasoning in the `How` section made this much faster to review, and the two things you flagged for a second opinion were the right things to flag.

**Sanity:** CI 21/21 green, changesets cover both touched packages. No human review yet.

I've left inline comments. Grouping them here by how much I think they should hold the merge:

### Would hold the merge

1. **Non-hovered series now fade on every bar chart, not just ones with a band.** The hover wiring and the opacity multiplier aren't gated on `hasReferenceBand`, so every multi-series `ChartBarWrapper` already in production changes behaviour on this minor. It also compounds with the existing category fade down to `0.04`.
2. **Crash for numeric or function `dataKey`.** `barSeriesClass(dataKey as string)` runs for every bar, and `DataKey<T>` is `string | number | ((obj) => any)`.

### Worth splitting out

3. The **`ChartTooltip` formatter fix** is a behaviour change across every chart type, bundled into a BarChart feature so it can't be reverted independently.
4. The **formatter is called with the wrong arguments for the range row**, so ordinary formatters that branch on `name` produce wrong bounds.

### Fixable in review

5. Silent geometry fallback in `reanchor` when `connectNulls` drops null points.
6. Both observers torn down and rebuilt on every bar hover.
7. MutationObserver watching `x`/`width` fires per animation frame.
8. ~70% of `useBarReferenceBand.tsx` duplicates `LineChart/useReferenceBand.tsx`, and the two have already diverged on the debounce strategy.
9. `sourceSignature` omits the bound classes, so changing a range key can leave a stale band.
10. New props land on the shared `types.ts`, so they surface on React Native and no-op.
11. The per-bar band is mouse-only: no keyboard path, nothing on touch, no text alternative.
12. Unconditional wrapper `<div>` with inline styles, which is what drives the snapshot churn.

Plus smaller notes inline on the sanitizer duplication and the `'Industry range'` default.

One process point: the export conflict with `origin/fix/chart-reference-band-review-comments` that you raised in the reviewer notes should be a hard merge blocker rather than a note. If that branch lands first, `ChartReferenceBand` isn't importable for bar chart users at all (this PR's `BarChart/index.ts` deliberately doesn't re-export it), and the changeset's "no import changes needed" line becomes untrue.

Last thing on testing: CI is green, but the untested paths are the risky ones. Nothing covers the length-mismatch fallback, resize recompute, non-string data keys, or the formatter argument shape.

packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
```diff
@@ -85,77 +109,146 @@ const _ChartBar: React.FC<ChartBarProps> = React.memo(
...
+              categoryOpacity * (isOtherSeriesHovered ? NON_HOVERED_SERIES_OPACITY : 1);
```
saurav12: **Blocking: this changes every existing bar chart, not just ones with a reference band.**

`updateHoveredBar` is wired to `onMouseEnter`/`onMouseLeave` on every `<RechartsBar>` below with no check for whether the chart has a band, so `hoveredDataKey` is always populated and `isOtherSeriesHovered` always applies. Every multi-series `ChartBarWrapper` already shipped in production will start fading its other series on hover after this minor release.

It also compounds with the pre-existing category fade. A bar that is both in a non-hovered series and a non-active category renders at `0.2 * 0.2 = 0.04`, which is effectively invisible:

```
fillOpacity = categoryOpacity * (isOtherSeriesHovered ? 0.2 : 1)
```

The changeset does mention the fade, but frames it as part of the band feature. It isn't scoped to the feature.

Suggest gating the hover wiring and this multiplier on `hasReferenceBand`, and getting design sign-off separately if the fade is genuinely wanted on all bar charts. The shaded category column is already correctly gated, so this looks like an oversight rather than intent.

packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
```diff
@@ -85,77 +109,146 @@ const _ChartBar: React.FC<ChartBarProps> = React.memo(
...
-          const fillOpacity = isNumber(activeIndex) ? (barIndex === activeIndex ? 1 : 0.2) : 1;
-          const gap = DISTANCE_BETWEEN_STACKED_BARS;
-          const isVertical = layout === 'vertical';
-
-          if (isVertical) {
+      <>
+        {boundSeries}
+        <RechartsBar
+          {...rest}
+          className={barSeriesClass(dataKey as string)}
```
saurav12: **Blocking: this throws for a numeric or function `dataKey`.**

This runs unconditionally for every bar, and `ChartBarProps['dataKey']` is `RechartsBarProps['dataKey']`, which recharts defines as:

```ts
export type DataKey<T> = string | number | ((obj: T) => any);
```

`barSeriesClass` calls `dataKey.replace(...)`, so a number or a function gives `TypeError: dataKey.replace is not a function`. Function data keys like `dataKey={(d) => d.metrics.sales}` are a normal recharts pattern, so this is reachable for existing consumers who don't use the new feature at all. The `as string` cast is what hides it from the type checker.

`perLineBandClass(dataKey as string, ...)` above has the same cast, though with narrower exposure since it's behind `hasRange`.

Suggest a `typeof dataKey === 'string'` guard, skipping the class (and the band) when it doesn't hold.

packages/blade/src/components/Charts/CommonChartComponents/CommonChartComponents.web.tsx
```diff
@@ -499,15 +521,17 @@ const CustomTooltip = ({
...
+            {formatValue(upperValue)}
           </Text>
         </Box>
       ) : null}
     </Box>
   );
 };
 
-const ChartTooltip: React.FC<ChartTooltipProps> = (props) => {
+const ChartTooltip: React.FC<ChartTooltipProps> = ({ formatter, ...props }) => {
```
saurav12: **This is a cross-cutting behaviour change riding along in a feature PR.**

`ChartTooltip` is shared by every chart type. `formatter` was type-accepted but ignored, so anyone passing one today sees no effect. After this it applies, and their tooltip output changes on upgrade with no warning.

You call this out honestly in the description, and the fix itself is right. The concern is packaging: bundled into a BarChart feature, it can't be reverted independently if it breaks a surface. Would you consider splitting it into its own patch PR?

packages/blade/src/components/Charts/CommonChartComponents/CommonChartComponents.web.tsx
```diff
@@ -450,12 +452,32 @@ const CustomTooltip = ({
...
 
   const toolTipColor = getChartColor(item.dataKey, item.name, dataColorMapping ?? {}, chartName);
 
+  /**
+   * Run a value through the consumer's `formatter` so units (`%`, currency, …) show up in the
+   * tooltip. Recharts formatters may return either the formatted value or a `[value, name]` tuple;
+   * only the value half is used here. The range bounds are formatted with the same function since
+   * they measure the same thing as the series value.
+   */
+  const formatValue = (value: unknown): React.ReactNode => {
```
saurav12: **The arguments are wrong when this formats the range bounds.**

For the range row, `formatValue(lowerValue)` passes `item.name` as the name, `item` as the entry and `[item]` as the payload array. But `item` is the *series* entry, so `entry.value` is the series value, not the bound.

A perfectly ordinary formatter breaks:

```tsx
formatter={(value, name) => (name === 'Success rate' ? `${value}%` : value)}
```

That formats both bounds as if they were the series value. Any formatter reading the third argument rather than the first has the same problem.

`index` is also the tooltip row index here, not the data index recharts would normally pass.

If the intent is just "apply the same unit formatting to the bounds", it may be worth documenting that only the `value` argument is meaningful for range rows, or passing a synthetic entry whose `value` is the bound.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
```diff
@@ -0,0 +1,417 @@
...
+
+        const upperAnchors = parsePathAnchors(upperCurve.getAttribute('d') ?? '');
+        const lowerAnchors = parsePathAnchors(lowerCurve.getAttribute('d') ?? '');
+        if (upperAnchors.length === 0 || lowerAnchors.length === 0) return;
+
+        // Re-anchor to this series' bar centres where they line up 1:1 with the bounds; otherwise
+        // keep the bound series' own x (the single centred bar case).
+        const centres = source.hoverOnly ? getBarCentres(surface, source.id) : null;
+        const reanchor = (anchors: PixelPoint[]): PixelPoint[] =>
+          centres && centres.length === anchors.length
```
saurav12: **Silent fallback to the wrong geometry on any dataset with gaps.**

The bound `<Line>`s use `connectNulls`, so null data points are dropped from the rendered path, while bar rects are still emitted for those categories. The two lengths then diverge, this guard fails, and the band falls back to the bound line's own x, which is the *category* centre. That's exactly the misalignment the re-anchoring exists to avoid, and it happens with no warning.

The same mismatch also makes `getCategoryColumn` wrong: `width = (bounds.right - bounds.left) / centres.length` assumes `centres.length` equals the category count, and `centres[index]` assumes the indices line up.

Worth either mapping centres to data indices explicitly rather than relying on array position, or at minimum warning in dev when the lengths disagree so it fails loudly instead of quietly.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
```diff
@@ -0,0 +1,417 @@
...
+      cleanups.push(() => resizeObserver.disconnect());
+    }
+    return () => cleanups.forEach((cleanup) => cleanup());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [
+    data,
+    hasReferenceBand,
+    sourceSignature,
+    hoveredDataKey,
+    hoveredBarIndex,
```
saurav12: **Both observers are torn down and recreated on every bar hover.**

`hoveredDataKey` and `hoveredBarIndex` are in the dependency array, so moving the mouse across bars disconnects and rebuilds the `MutationObserver` and `ResizeObserver` each time. Add a consumer passing an inline `data={[...]}` literal (very common) and it happens on every render.

Suggest splitting the effect: one that owns observer lifecycle keyed on the things that actually change the observation target, and a separate one that just recomputes on hover state.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
```diff
@@ -0,0 +1,417 @@
...
+        timerId = setTimeout(() => {
+          timerId = null;
+          computeBands();
+        }, 0);
+      });
+      mutationObserver.observe(container, {
+        childList: true,
+        subtree: true,
+        attributes: true,
+        attributeFilter: ['d', 'x', 'width'],
```
saurav12, line 492: LineChart's version watches only `d`. Bars animate `x`, `width` and `height`, so during the entry animation this fires every frame, and each one schedules a full `querySelectorAll` sweep of the surface plus a `setState`.

Worth pausing the recompute while `shouldAnimatedBar` is true, and doing one pass on animation end.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
```diff
@@ -0,0 +1,417 @@
...
+import type { ChartBarProps } from './types';
+
+type ChartData = { [key: string]: unknown };
+
+/**
+ * One band to draw. `lowerClass` / `upperClass` are the classNames of the invisible bound series
+ * whose rendered geometry supplies the y pixels. `hoverOnly` bands are revealed only while their
+ * bar is hovered — several grouped bars can each declare a range without the bands piling up.
+ */
+type BandSource = {
```
saurav12, line 32: **Roughly 70% of this file is a verbatim copy of `LineChart/useReferenceBand.tsx`.**

`BandSource`, `BandGeometry`, `geomsEqual`, the `bandSources` memo, the whole `useLayoutEffect` skeleton, `renderReferenceBands` and `referenceBandLegendInfos` are all duplicated. Two copies of brittle DOM scraping against recharts internal class names will drift as recharts is upgraded.

They've already diverged, in fact: the line hook debounces with `requestAnimationFrame`, this one with `setTimeout(0)`, for the same problem. The comment below argues rAF is wrong because it never fires while the document is hidden. If that's right, LineChart is now carrying that bug and should be fixed alongside. If it isn't, these should match.

The description says the geometry helper is written to be shared. I'd go further and extract the whole shell, leaving only `buildBarBandPath`, `getBarCentres` and the hover logic bar-specific.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
```diff
@@ -0,0 +1,417 @@
...
+  const [bandGeoms, setBandGeoms] = useState<BandGeometry[]>([]);
+  const [highlight, setHighlight] = useState<{
+    x: number;
+    width: number;
+    y: number;
+    height: number;
+  } | null>(null);
+
+  // Stable key of the sources' identity so the effect re-runs when bands are added/removed/recolored.
+  const sourceSignature = useMemo(
```
saurav12, line 321: The signature is only `id:fillColor`. It omits `lowerClass`/`upperClass` (derived from the range data keys), `hoverOnly` and `name`.

So changing `rangeLowerDataKey` while `dataKey`, colour and the `data` reference stay the same leaves a stale band on screen. Inherited from LineChart, but more reachable here now that per-bar ranges are configurable.

Including the bound classes in the signature would be a one-line fix.

packages/blade/src/components/Charts/BarChart/types.ts
```diff
@@ -44,6 +44,34 @@ type ChartBarProps = {
...
    */
   barSize?: RechartsBarProps['barSize'];
+  /**
+   * Data key for the lower (min) bound of this bar's reference range.
+   * A band is drawn for this bar only when both `rangeLowerDataKey` and `rangeUpperDataKey` are set.
+   *
+   * The band is revealed while this bar is hovered, so several bars can each declare their own
+   * range without the bands overlapping.
+   */
+  rangeLowerDataKey?: string;
```
saurav12, line 61: **These land on the cross-platform type surface but only work on web.**

`BarChart.native.tsx` imports `ChartBarProps` from this file (no platform suffix), so `rangeLowerDataKey`, `rangeUpperDataKey`, `rangeName`, `rangeColor` and `showRangeLegend` will autocomplete for React Native consumers and silently do nothing.

The PR scopes itself to web deliberately, which is fine, but the type surface doesn't reflect that. Either gate with `Platform.Select` / an explicit `@platform web` JSDoc note, or ship native parity in the same release.

packages/blade/src/components/Charts/BarChart/types.ts
```diff
@@ -44,6 +44,34 @@ type ChartBarProps = {
...
+  /**
+   * Data key for the upper (max) bound of this bar's reference range.
+   */
+  rangeUpperDataKey?: string;
+  /**
+   * Label shown for this bar's range in the legend and the tooltip.
+   * @default 'Industry range'
+   */
+  rangeName?: string;
```
saurav12, line 74: **Accessibility: the per-bar band is mouse-only.**

Hover is the only way to reveal it. There's no keyboard focus path, nothing on touch, and no text alternative. For a feature whose entire purpose is conveying comparative data, that leaves keyboard and mobile users with no access to it at all.

Since hover-reveal is the deliberate design for the grouped shape, this needs a decision rather than a code fix. Options might be revealing on focus as well as hover, or always rendering the range in the tooltip (which already happens) and treating the band as pure enhancement, documented as such.

Separately on this prop: `'Industry range'` as the default, and the tooltip's `'Industry'` fallback, bake Optimizer vocabulary into a generic design-system default. `'Reference range'` would match the component name. It's inherited from `ChartLine`, so it's a consistency-versus-correctness call, but worth making consciously.

packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
```diff
@@ -271,31 +444,29 @@ const ChartBarWrapper: React.FC<ChartBarWrapperProps & TestID & DataAnalyticsAtt
...
-                setActiveIndex(undefined);
-              }}
-              layout={layout}
-              data={data}
-            >
-              {barChartModifiedChildrens}
-            </RechartsBarChart>
-          </RechartsResponsiveContainer>
+        <BarChartContext.Provider value={barChartContextValue}>
+          <div ref={containerRef} style={{ width: '100%', height: '100%' }}>
```
saurav12: This wrapper is added for every bar chart regardless of whether it has a band, which is what drives the ~3900 lines of snapshot churn and changes the rendered DOM for every existing consumer.

Two things:

- Blade's convention is `BaseBox`, not a raw `div` with an inline `style` object.
- Attaching `containerRef` to the existing `BaseBox` above would avoid the extra node entirely, and with it the DOM change and most of the snapshot diff.

packages/blade/src/components/Charts/BarChart/tokens.ts
```diff
@@ -6,16 +6,35 @@ const DISTANCE_BETWEEN_BARS = 2;
...
 const componentIds = {
   chartBar: 'ChartBar',
 };
 
+/**
+ * Stable className on a bar series' rendered group, so the reference-band layer can find *that*
+ * series' bars by dataKey. Index-based lookup would break the moment a series is toggled off in the
+ * legend, since a hidden bar renders no group at all.
+ */
+const barSeriesClass = (dataKey: string): string =>
```
saurav12: This inlines the same `/[^a-zA-Z0-9_-]/g` regex that `sanitizeBandKey` already exports from `../utils/referenceBandUtils`. Two copies that have to stay in sync, and the band classes and the series class are derived from the same key, so drift between them would be a subtle bug.

Also worth noting the sanitizer collides: `a.b` and `a b` both become `a-b`, and `querySelector` takes the first match, so two such series would read each other's bars.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
saurav12: Following up on `f590e17a5`, since I think this one was missed rather than deliberately deferred. The commit covers the rest of the review thoroughly, but this thread isn't mentioned in the message and the code is unchanged, now at `useBarReferenceBand.tsx:332`:

```ts
const reanchor = (anchors: PixelPoint[]): PixelPoint[] =>
  centres && centres.length === anchors.length
    ? anchors.map((anchor, index) => ({ x: centres[index], y: anchor.y }))
    : anchors;
```

Restating the concern briefly. The bound `<Line>`s are rendered with `connectNulls`, so a null value in a range key drops that anchor from the rendered path, while the bar rect for that category is still emitted. `anchors.length` then comes back one short of `centres.length`, the guard fails, and the band falls back to the bound line's own x, which is the category centre. In a grouped chart that's the stacked-on-one-x failure the re-anchoring exists to prevent, and it happens with no warning.

`getCategoryColumn` carries the same assumption independently, so the shaded column drifts with it:

```ts
const width = (bounds.right - bounds.left) / centres.length;
// ...
x: bounds.left + activeIndex * width,
```

Both treat array position as equivalent to data index. That holds only while every category renders exactly one rect and every anchor survives.

One thing worth checking before deciding how far to take this: I reasoned about the `connectNulls` behaviour from the recharts docs rather than reproducing it, so it's worth a quick story with a `null` in one of the range keys to confirm the anchor actually drops. If it does, mapping centres to data indices explicitly rather than relying on position would be the robust fix. If that's more than this PR should carry, a dev-only warning when the two lengths disagree would at least make it fail loudly instead of quietly drawing in the wrong place, and the `@platform web` JSDoc could note that range keys must be dense.

Not blocking on the robust fix specifically. Blocking on it not failing silently.

For what it's worth, the rest of the commit checks out. I verified each claim against the code rather than the message: both blockers are properly gated, the `Math.min` change leaves the no-band path bit-identical, the formatter revert is clean (`CommonChartComponents.web.tsx` is no longer in the diff against master at all), and snapshot churn is down to 761 lines. The vertical-layout skip and the y-axis fallback for the column top are good catches that I'd missed.

kamaleshs-bridge4 (dismissed):
## ✨ Agentic PR Review ✨

**Status:** Approved ✅

### UI Review

✅ 6 passed

<details>
<summary>Passing checks (6)</summary>

| Check | Screenshot |
|-------|------------|
| ✅ [BarChart with Reference Band](https://61c19ee8d3d282003ac1d81c-pnwjloqkfl.chromatic.com/?path=/story/components-charts-barchart--bar-chart-with-reference-band) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/16-17-48/components-charts-barchart--bar-chart-with-reference-band-1.png"></a> |
| ✅ [Grouped BarChart with Multiple Reference Bands (initial render)](https://61c19ee8d3d282003ac1d81c-pnwjloqkfl.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/16-17-48/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands-2.png"></a> |
| ✅ [Grouped BarChart with Multiple Reference Bands (hover reveal)](https://61c19ee8d3d282003ac1d81c-pnwjloqkfl.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/16-17-48/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands-3.png"></a> |
| ✅ [Default BarChart (regression)](https://61c19ee8d3d282003ac1d81c-pnwjloqkfl.chromatic.com/?path=/story/components-charts-barchart--default-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/16-17-48/components-charts-barchart--default-chart-4.png"></a> |
| ✅ [Grouped BarChart (regression — hover behavior)](https://61c19ee8d3d282003ac1d81c-pnwjloqkfl.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/16-17-48/components-charts-barchart--grouped-bar-chart-5.png"></a> |
| ✅ [Vertical BarChart (regression — layout)](https://61c19ee8d3d282003ac1d81c-pnwjloqkfl.chromatic.com/?path=/story/components-charts-barchart--vertical-bar-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/16-17-48/components-charts-barchart--vertical-bar-chart-6.png"></a> |

</details>

### Usage

```jsx
import { ChartBar, ChartBarWrapper, ChartReferenceBand } from '@razorpay/blade/components';

<ChartBarWrapper data={data}>
  <ChartReferenceBand lowerDataKey="industryLow" upperDataKey="industryHigh" name="Industry range" />
  <ChartBar dataKey="successRate" rangeLowerDataKey="low" rangeUpperDataKey="high" rangeName="Range" />
</ChartBarWrapper>
```

saurav12:
Found while checking the tooltip in Storybook: **the range label has two different defaults depending on where it renders.**

| Where | Default | Source |
|---|---|---|
| Legend | `'Industry range'` | `useBarReferenceBand.tsx:241` — `props.rangeName ?? 'Industry range'` |
| Tooltip | `'Industry'` | `CommonChartComponents.web.tsx:502` — `range?.rangeName ?? 'Industry'` |

They diverge because `rangeMap` passes the prop straight through without applying a default:

```ts
// BarChart.web.tsx:419
rangeName: props.rangeName,
```

So a `<ChartBar>` that sets `rangeLowerDataKey` / `rangeUpperDataKey` but omits `rangeName`, which is the documented happy path since `rangeName` is optional, gets a legend swatch reading "Industry range" and a tooltip row reading "Industry" for the same band. With `showRangeLegend` on, both are on screen at once.

I suspect this is why the grouped story hardcoded `rangeName` per method rather than relying on the default.

Worth picking one string and sourcing both from it, probably a shared constant next to `REFERENCE_BAND_DEFAULT_COLOR` in `CommonChartComponents/tokens.ts`. One caveat on scope: the tooltip fallback is pre-existing master code shared with LineChart, so changing `'Industry'` shifts LineChart's tooltip too. Same cross-cutting-tooltip-change question you already navigated with the `formatter` revert, so your call whether it belongs here or in the follow-up patch PR.

---

Separately, I've pushed `765f574a3` to this branch with two small fixes — please review and revert if you'd rather they went elsewhere.

**1. The tooltip's range row was misaligned by 4px.** Its indent was a hardcoded `spacing.5` (16px) standing in for a sum: the colour swatch (`spacing.4`, 12px) plus the gap to the series name (`spacing.3`, 8px), which is 20px.

```diff
-          paddingLeft="spacing.5"
+          paddingLeft="spacing.6"
```

Measured in a running Storybook: `Card` sat at 271.7px against `Card industry range` at 267.7px. After the change every name and range label in the tooltip shares one left edge. Like the default mismatch above, this is in the shared `CustomTooltip` and predates this PR, so it fixes LineChart's row too — happy to pull it into the follow-up patch instead if you'd prefer this PR stay narrow.

**2. Dropped the series name from the story's `rangeName`.** `'Card industry range'` becomes `'Industry range'`. The row already sits directly under its own series name and colour swatch, so repeating the method name reads as noise. Story-only; component defaults untouched.

64 chart web tests and 41 snapshots pass unchanged. The tooltip renders on hover and isn't captured in any snapshot, so nothing needed regenerating. I added a line to the changeset since the alignment fix is user-visible in a shipped component.

One stale item while you're here: the PR description still says the tooltip "renders `62%` and a `48%–70%` range row", which was true before the `formatter` revert. The changeset was updated correctly; the description wasn't.

kamaleshs-bridge4 (commented):
## ✨ Agentic PR Review ✨

### UI Review

✅ 2 passed

<details>
<summary>Passing checks (2)</summary>

| Check | Screenshot |
|-------|------------|
| ✅ [BarChartWithReferenceBand](https://61c19ee8d3d282003ac1d81c-gbusmddckx.chromatic.com/?path=/story/components-charts-barchart--bar-chart-with-reference-band) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/16-50-26/components-charts-barchart--bar-chart-with-reference-band-1.png"></a> |
| ✅ [GroupedBarChartWithMultipleReferenceBands](https://61c19ee8d3d282003ac1d81c-gbusmddckx.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/16-50-26/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands-2.png"></a> |

</details>

### Usage

```jsx
import { ChartBarWrapper, ChartBar, ChartReferenceBand, ChartXAxis, ChartYAxis, ChartTooltip, ChartLegend } from '@razorpay/blade/components';

<ChartBarWrapper data={data}>
  <ChartReferenceBand lowerDataKey="low" upperDataKey="high" name="Industry range" />
  <ChartXAxis dataKey="period" />
  <ChartYAxis label="Success rate (%)" />
  <ChartTooltip />
  <ChartLegend />
  <ChartBar dataKey="value" name="Value" rangeLowerDataKey="low" rangeUpperDataKey="high" rangeName="Industry range" />
</ChartBarWrapper>
```

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
premika-pixel: You're right that it was missed rather than deferred — it was in my first triage, fell out when I rewrote the plan, and I didn't notice it was gone. Sorry for the round trip.

Fixed in `89e5028eb`.

**On the empirical check you asked for** — confirmed, though from the recharts source rather than a story, which I think settles it more firmly:

- `recharts/es6/shape/Curve.js:76` — `var formatPoints = connectNulls ? points.filter(defined) : points;`
- `recharts/es6/shape/Curve.js:38` — `var defined = p => isWellBehavedNumber(p.x) && isWellBehavedNumber(p.y);`
- `recharts/es6/cartesian/Bar.js:592` — `height = isNan(computedHeight) ? 0 : computedHeight;`

So the anchor really is dropped, and the bar rect really does survive as a zero-height rect. Your reasoning was right on both halves.

**The fix.** Anchors are mapped back to their data row rather than matched by array position, reusing `getDefinedNumericPoints` from `nullBridgeUtils` (the LineChart null-bridge already needed exactly this mapping). Anchor `k` belongs to row `indices[k]`, so it anchors to `centres[row]`. A gap now costs that one anchor instead of discarding the re-anchoring for the whole series.

Two things that fell out of it:

- `getBarCentres` was de-duplicating the flat rect list **by x value**, which made the array length depend on the values rather than on the number of bars. Since the whole point of that array is that position `i` is row `i`, it now reads one rect per `.recharts-bar-rectangle` group instead.
- `getCategoryColumn` now takes the row count explicitly instead of inferring it from `centres.length`. I deliberately did **not** re-centre it on the hovered bar: the column shades the whole category, and in a grouped chart the bar centre and the category centre differ by design — that difference is the thing the re-anchoring exists to honour, so centring on the bar would slide the shading off the period it marks.

**Failing loudly**: where the anchor count and the defined-row count disagree, the model of recharts' output is wrong — that now warns in dev and falls back. I used `logger({ type: 'warn' })` rather than `throwBladeError`, since `throwBladeError` actually throws in dev and a data shape shouldn't take the chart down.

**Test**: a grouped chart with a null bound, asserting the revealed band lands on its own series' bar centres. I checked it fails against the previous commit and passes against this one — a regression test that passes both ways isn't worth much. Existing snapshots are unchanged, which is the evidence that the `getBarCentres` rewrite didn't move any existing geometry.

**One thing I found while writing that test, worth a separate decision.** `ChartBarWrapper` types `data` as `{ [key: string]: string | number }`, so a null bound isn't expressible in BarChart's public type at all — while `ChartLineWrapper` uses `{ [key: string]: unknown }` and its own stories pass `null`. The runtime path is still reachable (a row that simply lacks the key is `undefined`, which recharts treats identically), so the fix stands either way, but the two sibling charts disagreeing on their data type looks like an oversight. I've recorded it in `decisions.md` as an open ask rather than widening a public type mid-review — happy to fold it in here if you'd rather.

**Also in this commit**: the collision half of the `sanitizeBandKey` thread, which I'd left unaddressed. `a.b` and `a b` both sanitize to `a-b` and `querySelector` takes the first match, so two such series would read each other's bars. That now warns in dev naming both keys. Making the sanitizer injective would change LineChart's class names too, which seemed disproportionate to the likelihood — flagging it as a follow-up instead, but say the word if you'd rather have it properly injective.

packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
premika-pixel: Fixed in `f590e17a5`. You were right that this was an oversight, not intent — the shaded column was gated and this wasn't.

`hasReferenceBand` now comes down through `BarChartContext`, and both the mouse handlers and the opacity multiplier are gated on it, so a chart with no band never attaches the handlers at all (no wrapper re-render on hover either). There's a regression test asserting no bar in a band-less grouped chart drops below full opacity.

On the `0.04`: the two fades are now combined with `Math.min` rather than multiplied. A bar is either de-emphasised or it isn't — being de-emphasised on two axes shouldn't compound. The no-band path is bit-identical to before, which I think you verified independently.

packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
premika-pixel: Fixed in `f590e17a5`. Good catch — this was reachable by consumers who never touch the feature.

I removed the `as string` casts rather than relocating them, since the cast was the actual defect: it's what stopped `tsc` from seeing `DataKey<T>`. A `seriesKey` local narrows to `string | null`, and a non-string key opts out of the band instead of throwing.

Worth reporting: the regression test I added for this **caught two more instances of the same cast** that neither of us had spotted — one in `useBarReferenceBand`'s `bandSources` memo and one in the `rangeMap` memo. Both fixed in the same commit.

packages/blade/src/components/Charts/CommonChartComponents/CommonChartComponents.web.tsx
premika-pixel: Agreed, and reverted out of this PR in `f590e17a5` — `CommonChartComponents.web.tsx` is no longer in the diff against master at all.

It follows as its own patch PR, carrying the argument-shape bug from the thread below. What stays here is the standalone-band → `rangeMap` wiring, which is genuinely part of this feature: without it a chart-level band drew the shaded area but showed no range row.

The visible cost is that the stories show `2000` rather than `2000%` until that PR lands. That's the right trade for being able to revert the two independently.

packages/blade/src/components/Charts/CommonChartComponents/CommonChartComponents.web.tsx
premika-pixel: You're right, and thank you for working through the argument shape — I'd missed that `item` is the series entry, so `entry.value` is the series value rather than the bound. A formatter branching on `name` would silently format both bounds as the series value.

Since the formatter work is moving to its own PR (thread above), this fix goes with it — passing a synthetic entry whose `value` is the bound, so the third argument means what a formatter would expect. Not left as a documentation caveat.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
premika-pixel: Fixed in `f590e17a5`. Split into two effects: one owning observer lifecycle, keyed only on the observation target (`containerRef`, `hasReferenceBand`), and one recomputing on the geometry inputs. The observers now call through a ref holding the latest `computeBands`, so hover state changing no longer rebuilds them.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
premika-pixel: I've left the code as-is here, and want to put the reasoning up for disagreement rather than change it quietly.

The per-frame cost isn't a `querySelectorAll` sweep plus a `setState` — every mutation cancels the pending callback before rescheduling, so the expensive sweep runs **once**, after the last frame of the animation, not once per frame. What runs per frame is a single `.some()` over the records, which short-circuits on the first band-layer hit.

`x`/`width` are in the filter deliberately: bar centres are read from the rects themselves, so unlike LineChart the bar geometry is an input, not just `d`.

`f590e17a5` did move this to `requestAnimationFrame` to match LineChart (see the divergence thread). If you still think an explicit animation pause earns its keep on top of the debounce, I'll add it — I just didn't want to add machinery for a cost I couldn't measure.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
premika-pixel: Agreed on the substance; deferring the extraction, but fixing the divergence now.

On the debounce: I checked my own justification and it doesn't hold. The stale band I was working around only reproduces in a permanently hidden headless pane — in a real tab a pending rAF resolves as soon as the tab is visible. So LineChart does **not** carry the bug my comment implied, and `f590e17a5` moves this hook back to `requestAnimationFrame` so the two match. My original comment overstated the case.

The extraction itself I'd rather do as its own PR — it touches a shipped component, and the pieces that would stay bar-specific (`buildBarBandPath`, `getBarCentres`, hover) have just changed shape again in `b90d6d715`, so extracting now would mean extracting a moving target. Happy to open it straight after this lands.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
premika-pixel: Fixed in `f590e17a5` — the signature now includes `lowerClass`, `upperClass` and `hoverOnly` alongside the id and colour, so changing a range key can't leave a stale band.

packages/blade/src/components/Charts/BarChart/types.ts
premika-pixel: Addressed in `f590e17a5`, though by documenting rather than gating.

Each of the five props now carries an explicit `@platform web` note saying it's accepted but inert on React Native. `Platform.Select` would make the type itself honest, but native parity is the real fix and I didn't want to ship a type shape that would need unwinding a release later. Native parity is on the follow-up list.

packages/blade/src/components/Charts/BarChart/types.ts
premika-pixel: Two separate points, handled differently.

**Accessibility** — you're right that this needs a decision, not a code fix. The decision: the band is a **visual enhancement**, not the carrier of the data. Every value it encodes is in the tooltip, which is reachable without a pointer, and the chart-wide band additionally has a legend swatch. That's now written down in `decisions.md` rather than left implicit, and revealing on keyboard focus is recorded as an open ask — it needs bars to become focusable, which is shared chart behaviour rather than a band concern.

**`'Industry range'`** — I'm not changing it, and want to be explicit about why rather than quietly leaving it. You're right on the substance: it's Optimizer vocabulary in a generic default. But it's inherited from `ChartLine` (`LineChart/types.ts:87`) and the `'Industry'` fallback lives in the shared tooltip (`CommonChartComponents.web.tsx:521`). Changing it on `ChartBar` alone would *create* an inconsistency where there currently is one shared wart, and changing it everywhere is a cross-chart rename affecting shipped LineChart consumers. That deserves its own decision rather than riding along here.

packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
premika-pixel: Fixed in `f590e17a5`. `BaseBox` is a `styled.div`, so it forwards `ref` to the DOM node — `containerRef` moved onto it and the wrapper is gone. Both points were right: the convention one and the churn one.

Snapshot diff went from **3,937 lines to 761**, and more tellingly the 1,587 lines of *rewritten existing* snapshots dropped to 11 — so existing charts' DOM is genuinely untouched now, rather than just less noisy.

packages/blade/src/components/Charts/BarChart/tokens.ts
premika-pixel: Both halves now addressed.

The duplication was fixed in `f590e17a5` — `barSeriesClass` imports `sanitizeBandKey` instead of re-inlining the regex.

The collision I initially missed; fixed in `b90d6d715`. Two sources resolving to the same sanitized class now warn in dev, naming both dataKeys. I stopped short of making the sanitizer injective because that changes LineChart's class names too and felt disproportionate to the likelihood — but it is a real latent bug rather than a theoretical one, so if you'd rather have it properly injective (data attributes instead of classNames would do it) I'm happy to take that on here.

kamaleshs-bridge4 (dismissed):
## ✨ Agentic PR Review ✨

**Status:** Approved ✅

### UI Review

✅ 4 passed

<details>
<summary>Passing checks (4)</summary>

| Check | Screenshot |
|-------|------------|
| ✅ [BarChart with standalone reference band](https://feat-bar-chart-reference-band--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/story/components-charts-barchart--bar-chart-with-reference-band) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/18-30-12/components-charts-barchart--bar-chart-with-reference-band-1.png"></a> |
| ✅ [Grouped BarChart with multiple per-bar reference bands](https://feat-bar-chart-reference-band--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/18-30-12/components-charts-barchart--grouped-bar-chart-with-multiple-reference-bands-2.png"></a> |
| ✅ [Default BarChart (regression check)](https://feat-bar-chart-reference-band--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/story/components-charts-barchart--default) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/18-30-12/components-charts-barchart--default-3.png"></a> |
| ✅ [Grouped BarChart (regression check)](https://feat-bar-chart-reference-band--61c19ee8d3d282003ac1d81c.chromatic.com/?path=/story/components-charts-barchart--grouped-bar-chart) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3965/ui-critique/18-30-12/components-charts-barchart--grouped-bar-chart-4.png"></a> |

</details>

### Usage

```jsx
import { ChartBarWrapper, ChartBar, ChartReferenceBand, ChartXAxis, ChartYAxis, ChartTooltip, ChartLegend } from '@razorpay/blade/components';

// Shape 1: chart-wide reference band (always visible, legend swatch by default)
<ChartBarWrapper data={successRateData}>
  <ChartReferenceBand lowerDataKey="industryLow" upperDataKey="industryHigh" name="Industry range" />
  <ChartXAxis dataKey="period" />
  <ChartYAxis label="Success rate (%)" />
  <ChartTooltip />
  <ChartLegend />
  <ChartBar dataKey="successRate" name="Success rate" barSize={24} />
</ChartBarWrapper>

// Shape 2: per-bar reference range (hover-revealed, color-matched to bar)
<ChartBarWrapper data={groupedData}>
  <ChartXAxis dataKey="period" />
  <ChartYAxis label="Success rate (%)" />
  <ChartTooltip />
  <ChartBar dataKey="upi" name="UPI" rangeLowerDataKey="upiLow" rangeUpperDataKey="upiHigh" rangeName="UPI industry range" />
  <ChartBar dataKey="card" name="Card" rangeLowerDataKey="cardLow" rangeUpperDataKey="cardHigh" rangeName="Card industry range" />
</ChartBarWrapper>
```

premika-pixel:
Three things here, all now addressed.

### 1. The two defaults — fixed in `264e3da12`

Confirmed, and it was worse than the table shows. The standalone band diverges too, and further:

| Band, no name given | Legend | Tooltip |
|---|---|---|
| Per-bar | `Industry range` | `Industry` |
| Standalone | `Reference band` | `Industry` |

So the chart-wide case showed two entirely unrelated strings. Your diagnosis of the cause was exactly right — `rangeMap` passing the prop through undefaulted, so it fell to the shared fallback.

On the scope caveat you raised: I've taken the option that **doesn't** touch the shared `'Industry'` fallback, so LineChart is unaffected. Both defaults are now single constants (`REFERENCE_BAND_DEFAULT_NAME` next to `REFERENCE_BAND_DEFAULT_COLOR` as you suggested, and `BAR_RANGE_DEFAULT_NAME` in `BarChart/tokens.ts`) applied at **both** the legend and `rangeMap`. That makes the shared fallback unreachable from BarChart rather than changing it — consistent with where we landed on `formatter`, and it leaves the `'Industry'` string for the follow-up patch PR to deal with across both charts if we want it gone.

Tests assert the same string appears in both surfaces for a nameless band, in both shapes. I checked both fail with the defaults removed.

And yes — that is why the grouped story hardcoded `rangeName` per method. It was papering over this.

### 2. Your `765f574a3` — reviewed, keeping both

**The 4px alignment fix**: correct, and the derivation is right — `spacing.4` + `spacing.3` = 20px, which is `spacing.6`, not the hardcoded `spacing.5`. Keeping it here rather than moving it: unlike the `formatter` change it's a pure visual correction with no API or behavioural surface, so there's nothing for a consumer to have depended on and nothing to revert independently. The argument for splitting `formatter` was that it silently changed output for anyone already passing one; that doesn't apply to a 4px indent.

**Dropping the series name from the story's `rangeName`**: agreed, and it reads better. It also stops the story masking the bug in (1), which is a second reason to keep it.

Thanks for measuring both in a running Storybook — that's more than I did.

### 3. Stale PR description — fixed

You're right, I updated the changeset and missed the description. Corrected: the example no longer passes `formatter`, and the tooltip section now says plainly that the `formatter` work was reverted out and why, including that values render unformatted until the follow-up lands. The `62%` / `48%–70%` claim is gone.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
premika-pixel: Addressed in `f590e17a5`. Correct on all four functions — the geometry is horizontal throughout.

Taken the first option: the band is now skipped entirely under `layout="vertical"` rather than drawn in the wrong place, with the limitation documented in `types.ts`, `decisions.md` and the blade-mcp knowledgebase. There's a test asserting no band layer renders for a vertical chart. Extending the geometry to mirror onto the other axis is a follow-up.

packages/blade/src/components/Charts/BarChart/BarChart.web.tsx
premika-pixel: Addressed in `f590e17a5` — this one was also raised in human review and was the main blocker on the PR.

Both the hover wiring and the `isOtherSeriesHovered` multiplier are now gated on `hasReferenceBand`, threaded through `BarChartContext`. A chart with no band never attaches the handlers, so there's no wrapper re-render on hover either. Regression test asserts no bar in a band-less grouped chart drops below full opacity.

One thing beyond what you flagged: the two fades were also *multiplying* (`0.2 × 0.2 = 0.04`), which made a bar that was both the wrong series and the wrong category effectively invisible. They're combined with `Math.min` now.

packages/blade/src/components/Charts/BarChart/useBarReferenceBand.tsx
premika-pixel: Fixed in `f590e17a5`. The fallback is now the y-axis line's `y1` rather than `0`, exactly as suggested.

Worth confirming your observation that it was demonstrable in the flagship example — the grouped band story does omit the grid, so this was visible there rather than hypothetical.

packages/blade/src/components/Charts/BarChart/types.ts
premika-pixel: Addressed in `f590e17a5` by documenting rather than aligning.

The JSDoc now spells out the divergence and the reason: a line's band is always on screen, so a permanent legend swatch describes something visible, whereas a bar's band appears only while its bar is hovered — a permanent swatch would advertise something that usually isn't there. The asymmetry is deliberate, but it was previously implicit, which is the fair part of the critique.

saurabhdaware (approved):

PR #3973, opened 2026-09-09
fix(blade-svelte): keep formatted TextInput in sync with controlled value

rohankokane-dev:
## Summary
- Controlled `format` mode on Svelte `TextInput` now reconciles the display against the parent `value` after every change, so consumer sanitisation (digit-only card fields) and programmatic prefill/reset actually show up on screen.
- Formatter no longer leaks a trailing delimiter when the value ends on a group boundary, and the caret stays after the typed character when a delimiter is inserted.
- Adds unit tests plus Storybook stories that reproduce the checkout-style controlled card form (NFC autofill / DCC clear).

## Test plan
- [ ] Type letters mixed with digits in a formatted + controlled card field — letters vanish from the input, stored value stays digits-only
- [ ] Type letters after a digit-group boundary (e.g. `1234` then `abc`) — display stays `1234`, no leaked letters
- [ ] Programmatic prefill and clear (Visa/Amex buttons, reset) update the visible field
- [ ] Type the 5th digit of a card number — caret lands after the digit, not before the inserted space
- [ ] 16-digit value against `#### #### #### #### ###` has no trailing space
- [ ] Storybook: `Text Input Formatted Controlled` and `Text Input Checkout Card (Controlled)` (NFC Autofill / Switch to GPay)

### DIFF
--- a/packages/blade-svelte/src/components/Input/TextInput/TextInput.svelte
+++ b/packages/blade-svelte/src/components/Input/TextInput/TextInput.svelte
@@ -74,13 +74,34 @@
     untrack(() => (formatter ? formatter.formatValue(value ?? defaultValue ?? '') : '')),
   );
 
-  // Re-format the current raw value whenever the pattern changes at runtime.
+  // In controlled `format` mode the parent's `value` is the source of truth, so
+  // reconcile the display against it after EVERY change — not just when the
+  // `value` reference changes. Tracking `formattedValue` (the optimistic value
+  // `handleChange` writes) is what makes consumer sanitisation stick: when the
+  // parent strips a just-typed character (e.g. a letter in a digit-only card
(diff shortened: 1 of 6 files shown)


packages/blade-svelte/src/components/Input/TextInput/__tests__/TextInput.test.ts
saurabhdaware, line 1: @razorpay/slash-public can you add tests for uncontrolled version of this as well if they don't already exist?

packages/blade-svelte/src/components/Input/TextInput/TextInput.svelte
saurabhdaware, line 1: @razorpay/slash-public add appropriate changeset

saurabhdaware (approved):

PR #3974, opened 2026-09-09
fix(blade): keep formatted TextInput in sync with controlled value

rohankokane-dev:
## Summary

Port of the Svelte fix ([#3973](https://github.com/razorpay/blade/pull/3973)) to React.

### Bugs fixed

**Trailing delimiter** — `format()` appended a delimiter even when the value ended exactly on a group boundary. Example: 16 digits into \`#### #### #### #### ###\` produced \`"4111 1111 1111 1111 "\` (trailing space). Fix: break before appending a delimiter when no more input chars remain.

**Caret positioning** — After typing a digit that lands on a group boundary (e.g. 5th digit of a card number), the formatter inserts a space, shifting the digit right. The old logic used raw \`selectionStart\` which now sat *before* the typed digit. Fix: anchor the caret by counting user characters (non-delimiters) so it stays glued to the typed char regardless of how many delimiters were inserted.

### Changes

| File | Change |
|---|---|
| \`useFormattedInput.ts\` | Trailing delimiter guard + user-character-anchored caret |
| \`TextInput.web.test.tsx\` | 5 new tests covering letter-stripping, prefill, reset, no trailing delimiter, caret position |
| \`TextInputFormat.stories.tsx\` | \`CheckoutCardControlled\` story — NFC autofill + DCC provider switch simulation |

## Test plan
- [ ] Type letters mixed with digits in a card field — letters vanish, stored value stays digits-only
- [ ] Type letters after "1234" (no digit change) — display stays "1234", no leaked chars
- [ ] Programmatic prefill (NFC Autofill button) and clear (Switch to GPay) update the visible field
- [ ] Type the 5th digit — caret lands after it, not before the inserted space
- [ ] 16-digit value shows no trailing space

### DIFF
--- a/packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
+++ b/packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
@@ -599,3 +599,87 @@ describe('<TextInput />', () => {
     expect(getByLabelText('Enter name')).toHaveAttribute('data-analytics-name', 'Enter name');
   });
 });
+
+// ─── Formatted TextInput — controlled value sync ───────────────────────────────
+// Reproduces the checkout AddNewCard scenario: parent strips non-digits in
+// onChange and feeds the cleaned value back via `value`.
+const ControlledCardInput = (): ReactElement => {
+  const onlyDigits = (v: string): string => v.replace(/\D/g, '');
+  const [value, setValue] = useState<string>('');
+  return (
+    <>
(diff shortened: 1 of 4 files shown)


packages/blade/src/components/Input/TextInput/__tests__/TextInput.web.test.tsx
saurabhdaware, line 1: @razorpay/slash-public add tests for uncontrolled version of this as well if it doesn't already exist

PR #3981, opened 2026-09-10
fix: bottomsheet scroll lock issue

rohankokane-dev:
## Description

`body-scroll-lock-upgrade` locks iOS by moving the page into `body { position: fixed; top: -scrollY }` and restoring it with `window.scrollTo()` on unlock, which loses the reading position; Android and desktop take its `overflow: hidden` branch and were unaffected. Both overlays now use a ref-counted lock that applies the same `overflow: hidden` bookkeeping on every platform and blocks background scrolling on iOS by intercepting `touchmove`, so nothing mutates page geometry. BottomSheet additionally acquires its lock at most once per open instead of on every content re-measurement.

## Component Checklist

- [ ] Add KitchenSink Story
- [ ] Add changeset

### DIFF
--- a/packages/blade-svelte/src/components/BottomSheet/BottomSheet.svelte
+++ b/packages/blade-svelte/src/components/BottomSheet/BottomSheet.svelte
@@ -1,11 +1,6 @@
 <script lang="ts">
-  import { onMount, untrack } from 'svelte';
+  import { onDestroy, onMount, untrack } from 'svelte';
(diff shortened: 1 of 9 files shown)


packages/blade-svelte/src/components/BottomSheet/BottomSheet.svelte
saurabhdaware, line 1: @razorpay/slash-public can you add a test for this in browserstack tests?

saurabhdaware (dismissed):

saurabhdaware (approved):

PR #3989, opened 2026-09-11
feat(SliderInput): add SliderInput component (web)

saurav12:
> **Stacked on #3988** (the neutral focus ring), which this needs. Review that one first; this PR targets it rather than `master`, so the diff here is only the slider.

Adds `SliderInput`, a web-only slider built from the Figma spec. React Native is a `throwBladeError` stub.

---

# API reference

Reviewers are seeing this component for the first time, so this is the whole public surface and the reasoning behind the shape of it.

## Minimal usage

```jsx
<SliderInput label="Volume" />
```

Uncontrolled, 0–100, step 1.

### Value

| Prop | Type | Default | Notes |
| --- | --- | --- | --- |
| `value` | `number` | — | Controlled value. Snapped to `step` before rendering, so an off-step value moves the thumb to the nearest step rather than sitting between two. |
| `defaultValue` | `number` | — | Initial value when uncontrolled. |
| `onChange` | `({ name, value }) => void` | — | Fires on **every** change, including each pointer move during a drag. |
| `onChangeEnd` | `({ name, value }) => void` | — | Fires **once** when an interaction commits: pointer release, or key up. |

**Why two callbacks.** A drag fires `onChange` on every pointer move, which is right for live display and wrong for anything expensive. `onChangeEnd` is the one to hang a network call on. It is deliberately in v1 rather than added later, because adding it afterwards tends to mean consumers have already debounced `onChange` by hand.

### Range

| Prop | Type | Default | Notes |
| --- | --- | --- | --- |
| `min` | `number` | `0` | |
| `max` | `number` | `100` | Always reachable, even when it is not a whole number of `step`s from `min`. |
| `step` | `number` | `1` | A step **size**, not a count: `step={10}` over 0–100 gives 11 values. |

**On `max` always being reachable.** With `min={0} max={100} step={33.33}` the last whole step lands at 99.99 and `max` is a sliver away. Both are real, selectable stops — dropping one would make a value the consumer asked for unreachable. Only the *label* is suppressed when two collide (see below).

### Display

| Prop | Type | Default | Notes |
| --- | --- | --- | --- |
| `showMarkers` | `boolean` | `false` | Tick markers on the track, one per step. |
| `showScale` | `boolean` | `false` | Row of value labels beneath the track. |
| `showScaleValues` | `boolean` | `true` | When `false`, the scale shows only `min` and `max`. |
| `showValueIndicator` | `boolean` | `true` | Readout above the thumb on hover, focus and drag. |
| `formatValue` | `(value: number) => string` | — | Applied everywhere the value is shown: scale labels, the indicator, **and `aria-valuetext`**. |

Markers and scale labels are dropped automatically when the track is too narrow to space them out, so a slider in a tight column degrades instead of turning into a smear. `min` and `max` are always kept.

### Inherited, not redeclared

These are picked directly off the shared form types, so they behave identically to `TextInput`, `PasswordInput` and the rest:

- From `FormInputLabelProps`: `label`, `labelPosition`, `necessityIndicator`
- From `FormInputValidationProps`: `helpText`, `errorText`, `successText`, `validationState`
- From `BaseInputProps`: `name`, `isDisabled`, `isRequired`, `onFocus`, `onBlur`, `accessibilityLabel`, `testID`
- Plus `StyledPropsBlade` and `data-analytics-*`

`validationTextPlacement` is `Omit`ted: there is no field box to render text inside of.

The label and hint render through the same `FormLabel` / `FormHint` components, and the hint precedence uses `BaseInput`'s exported `getHintType` rather than a reimplementation, so the rule for which hint wins cannot drift.

## Pairing with another input

There is no built-in pairing prop. The slider is controlled and pairing is composition:

```jsx
const [radius, setRadius] = React.useState(16);

<SliderInput label="Corner Radius" labelPosition="left" value={radius}
  onChange={({ value }) => setRadius(value)} />
<TextInput value={String(radius)} onChange={...} />
```

A dedicated prop would have to take a position on validation, formatting and commit timing for an input it does not own. `value` + `onChange` already express it. See the `PairedWithTextInput` story.

## Not in the API, deliberately

- **No vertical orientation.** Not in the design, and it would change the pointer maths, the indicator placement and the scale layout. `aria-orientation="horizontal"` is stated explicitly so adding it later is not a breaking change.
- **No range (two-thumb) mode.** A different component, not a prop.
- **No `size`.** Not in the Figma spec.

---

# Implementation notes

## Positioning is pure CSS `calc()`

Thumb, fill, markers and scale labels are positioned with `calc()` percentages against the track, so **nothing is measured to render**. The first server render is already correct and there is no hydration shift. `ResizeObserver` is used only to decide whether labels still *fit*, never to position anything.

## Accessibility

Full ARIA slider pattern: `role="slider"` with `aria-valuemin` / `aria-valuemax` / `aria-valuenow` / `aria-valuetext`, plus `aria-orientation`, `aria-labelledby`, `aria-describedby`, `aria-disabled`, and `tabIndex` dropping to `-1` when disabled.

Complete APG keyboard set — arrows, `PageUp` / `PageDown`, `Home` / `End` — all tested.

Pointer targets meet WCAG 2.5.8 through a 44px pseudo-element, so the visual thumb stays 12px.

`aria-valuetext` runs through `formatValue`, so a screen reader hears `₹3k` rather than `3000`.

**Known gap: `prefers-reduced-motion` is not handled.** It is handled nowhere in Blade — the setting appears in no source file — and the decision is to address it system-wide rather than per component. Flagged explicitly in `_decisions.md` because the usual justification is weaker here than where it was set: this component carries real motion, including a spring-driven odometer. Happy to revisit if reviewers disagree.

## Motion

Every duration and easing comes from `theme.motion.*`; there are no hardcoded milliseconds or bezier literals. The thumb, fill and indicator share one `sliderInputMotion` block so they cannot drift apart.

Movement easing is suppressed while **scrubbing**, not merely while the pointer is down. A click on a far part of the track should glide there; a drag must pin the thumb to the finger, where easing reads as lag. Those are the same `pointerdown`, so they are told apart afterwards by a 3px slop.

The indicator's number rolls as an odometer — digit columns driven continuously from the value — rather than swapping. A swap has to finish one value before starting the next, so a drag changing the value every ~20ms left 19 half-faded copies stacked up and never resolving. Columns driven by the value have no such limit. Measured at a realistic drag speed they trail by ~1.5 units and land exactly on release. Content with no digits in it still swaps as a whole.

## Worth a reviewer's eye

- **`mask-composite: intersect`** draws the marker cut-outs. Verified in Chrome; **not yet verified in Safari.** This is the main thing I would like a second pair of eyes on.
- The odometer's DOM text reads `01234567890` because each column carries every digit. The real value is stated once in a visually hidden node and the columns are `aria-hidden`, so screen readers, find-in-page and copy all get the value rather than the columns.
- RTL uses logical properties throughout and the pointer maths reads the computed direction, but Blade has no RTL infrastructure to test against, so **no RTL support is claimed**.

`_decisions.md` covers the reasoning in full, including the options not taken.

## Checks

- [x] Web and native typecheck
- [x] Lint
- [x] 67 unit tests, plus SSR and `assertAccessible`
- [x] Stories, including `PairedWithTextInput`, `Steps` and `FormattedValue`
- [x] Changeset

### DIFF
--- a/packages/blade/src/components/BaseAnimatedValue/__tests__/BaseAnimatedValue.web.test.tsx
+++ b/packages/blade/src/components/BaseAnimatedValue/__tests__/BaseAnimatedValue.web.test.tsx
@@ -61,6 +61,30 @@ describe('<BaseAnimatedValue />', () => {
     expect(getByText('published')).toBeInTheDocument();
   });
 
+  it('should write rapid changes in place rather than stacking animated copies', async () => {
+    const user = userEvents.setup();
+    const Rapid = (): React.ReactElement => {
+      const [value, setValue] = React.useState(0);
+      return (
+        <>
+          {/* One click, many changes, exactly as a drag produces. */}
+          <button type="button" onClick={() => [1, 2, 3, 4, 5].forEach((v) => setValue(v))}>
+            churn
+          </button>
+          <BaseAnimatedValue value={value} testID="rapid" />
+        </>
+      );
+    };
+
+    const { getByText, getByTestId } = renderWithTheme(<Rapid />);
+    await user.click(getByText('churn'));
+
+    // A roll only reads when there is time to read it, so changes arriving mid-swap replace
+    // the content instead of queueing another copy behind it.
+    expect(getByTestId('rapid').children).toHaveLength(1);
+    expect(getByText('5')).toBeInTheDocument();
+  });
+
   it('should match snapshot', () => {
     const { container } = renderWithTheme(<BaseAnimatedValue value={7} />);
     expect(container).toMatchSnapshot();
--- a/packages/blade/src/components/BaseAnimatedValue/__tests__/BaseAnimatedValue.web.test.tsx
+++ b/packages/blade/src/components/BaseAnimatedValue/__tests__/BaseAnimatedValue.web.test.tsx
@@ -1,0 +1,68 @@
+import React from 'react';
+import userEvents from '@testing-library/user-event';
+import { BaseAnimatedValue } from '../BaseAnimatedValue.web';
+import renderWithTheme from '~utils/testing/renderWithTheme.web';
+
+describe('<BaseAnimatedValue />', () => {
+  it('should render the value when given no children', () => {
+    const { getByText } = renderWithTheme(<BaseAnimatedValue value={42} />);
+    expect(getByText('42')).toBeInTheDocument();
+  });
+
+  it('should render children in place of the value', () => {
+    // The raw value drives the direction while the formatted string is what is shown, so a
+    // unit or a currency does not stop the swap from moving the right way.
+    const { getByText, queryByText } = renderWithTheme(
+      <BaseAnimatedValue value={1200}>₹1,200</BaseAnimatedValue>,
+    );
+    expect(getByText('₹1,200')).toBeInTheDocument();
+    expect(queryByText('1200')).not.toBeInTheDocument();
+  });
+
+  /**
+   * The change is driven through state rather than RTL's `rerender`, which re-renders without
+   * the theme provider that `renderWithTheme` wrapped the first render in.
+   */
+  const Swapper = ({
+    from,
+    to,
+  }: {
+    from: string | number;
+    to: string | number;
+  }): React.ReactElement => {
+    const [value, setValue] = React.useState<string | number>(from);
+    return (
+      <>
+        <button type="button" onClick={() => setValue(to)}>
+          swap
+        </button>
+        <BaseAnimatedValue value={value} />
+      </>
+    );
+  };
+
+  it('should show the new value after a change', async () => {
+    const user = userEvents.setup();
+    const { getByText } = renderWithTheme(<Swapper from={1} to={2} />);
+    expect(getByText('1')).toBeInTheDocument();
+
+    await user.click(getByText('swap'));
+    expect(getByText('2')).toBeInTheDocument();
+  });
+
+  it('should take non-numeric content without complaint', async () => {
+    // There is no meaningful "up" between two words, so this cross-fades rather than
+    // travelling. It still has to render.
+    const user = userEvents.setup();
+    const { getByText } = renderWithTheme(<Swapper from="draft" to="published" />);
+    expect(getByText('draft')).toBeInTheDocument();
+
+    await user.click(getByText('swap'));
+    expect(getByText('published')).toBeInTheDocument();
+  });
+
+  it('should match snapshot', () => {
+    const { container } = renderWithTheme(<BaseAnimatedValue value={7} />);
+    expect(container).toMatchSnapshot();
+  });
+});
--- a/packages/blade/src/components/BaseAnimatedValue/__tests__/BaseAnimatedValue.web.test.tsx
+++ b/packages/blade/src/components/BaseAnimatedValue/__tests__/BaseAnimatedValue.web.test.tsx
@@ -3,20 +3,27 @@ import userEvents from '@testing-library/user-event';
 import { BaseAnimatedValue } from '../BaseAnimatedValue.web';
 import renderWithTheme from '~utils/testing/renderWithTheme.web';
 
+/**
+ * A rolling number carries every digit in each column, so its text reads `01234567890`. The one
+ * node that states the actual value is the hidden one, which is what a screen reader gets and
+ * what these assertions go by.
+ */
+const readValue = (container: HTMLElement): string =>
+  container.querySelector('[data-blade-component="visually-hidden"]')?.textContent ?? '';
+
 describe('<BaseAnimatedValue />', () => {
   it('should render the value when given no children', () => {
-    const { getByText } = renderWithTheme(<BaseAnimatedValue value={42} />);
-    expect(getByText('42')).toBeInTheDocument();
+    const { container } = renderWithTheme(<BaseAnimatedValue value={42} />);
+    expect(readValue(container)).toBe('42');
   });
 
   it('should render children in place of the value', () => {
-    // The raw value drives the direction while the formatted string is what is shown, so a
-    // unit or a currency does not stop the swap from moving the right way.
-    const { getByText, queryByText } = renderWithTheme(
+    // The raw value drives the roll while the formatted string is what is shown, so a unit or
+    // a currency rides along as literals between the digit columns.
+    const { container } = renderWithTheme(
       <BaseAnimatedValue value={1200}>₹1,200</BaseAnimatedValue>,
     );
-    expect(getByText('₹1,200')).toBeInTheDocument();
-    expect(queryByText('1200')).not.toBeInTheDocument();
+    expect(readValue(container)).toBe('₹1,200');
   });
 
   /**
@@ -43,16 +50,15 @@ describe('<BaseAnimatedValue />', () => {
 
   it('should show the new value after a change', async () => {
     const user = userEvents.setup();
-    const { getByText } = renderWithTheme(<Swapper from={1} to={2} />);
-    expect(getByText('1')).toBeInTheDocument();
+    const { container, getByText } = renderWithTheme(<Swapper from={1} to={2} />);
+    expect(readValue(container)).toBe('1');
 
     await user.click(getByText('swap'));
-    expect(getByText('2')).toBeInTheDocument();
+    expect(readValue(container)).toBe('2');
   });
 
-  it('should take non-numeric content without complaint', async () => {
-    // There is no meaningful "up" between two words, so this cross-fades rather than
-    // travelling. It still has to render.
+  it('should swap the whole value when there are no digits to roll', async () => {
+    // Words have no places to drive columns from, so they cross-fade as a whole instead.
     const user = userEvents.setup();
     const { getByText } = renderWithTheme(<Swapper from="draft" to="published" />);
     expect(getByText('draft')).toBeInTheDocument();
@@ -61,14 +67,14 @@ describe('<BaseAnimatedValue />', () => {
     expect(getByText('published')).toBeInTheDocument();
   });
 
-  it('should write rapid changes in place rather than stacking animated copies', async () => {
+  it('should write rapid changes to a swapped value in place rather than stacking copies', async () => {
     const user = userEvents.setup();
     const Rapid = (): React.ReactElement => {
-      const [value, setValue] = React.useState(0);
+      const [value, setValue] = React.useState('a');
       return (
         <>
           {/* One click, many changes, exactly as a drag produces. */}
-          <button type="button" onClick={() => [1, 2, 3, 4, 5].forEach((v) => setValue(v))}>
+          <button type="button" onClick={() => ['b', 'c', 'd', 'e'].forEach((v) => setValue(v))}>
             churn
           </button>
           <BaseAnimatedValue value={value} testID="rapid" />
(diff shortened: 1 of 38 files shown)


packages/blade/src/components/BaseAnimatedValue/__tests__/BaseAnimatedValue.web.test.tsx
saurabhdaware, line 1: Isn't this too much of motion? doesn't look very subtle the way we animate in other components 

https://github.com/user-attachments/assets/e06c616f-5323-4d34-80c2-412b4e3e74bd

packages/blade/src/components/Input/SliderInput/SliderInput.web.tsx
saurabhdaware, line 1: @saurav12 Do we have any usecase of range selection sliderinput? something like this:

Image

@razorpay/slash-public If a usecase like this comes, will the current API be able to handle this? (just respond here. dont implement anything)

packages/blade/src/components/Input/SliderInput/types.ts
```diff
@@ -0,0 +1,112 @@
...
+     * The initial value when the slider is uncontrolled.
+     */
+    defaultValue?: number;
+
+    /**
+     * Called on every change, including each pointer move during a drag.
+     *
+     * Bind live display to this. For anything expensive, use `onChangeEnd` instead.
+     */
+    onChange?: (args: { value: number }) => void;
```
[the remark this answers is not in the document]
saurabhdaware: @razorpay/slash-public Implement this change ^

packages/blade/src/components/Input/SliderInput/SliderInput.web.tsx
saurav12: @saurabhdaware This is known. For now we don't need this as there is no use-case for now. But I've added this in the plan.

packages/blade/src/components/BaseAnimatedValue/__tests__/BaseAnimatedValue.web.test.tsx
saurav12: Looks good to me, but maybe it will be distracting 🙈

packages/blade/src/components/Input/SliderInput/types.ts
```diff
@@ -0,0 +1,112 @@
...
+  FormInputLabelProps,
+  'label' | 'labelPosition' | 'necessityIndicator'
+> &
+  // `validationTextPlacement: 'inside'` has no meaning here: the slider has no field box to
+  // render text inside of.
+  Omit<FormInputValidationProps, 'validationTextPlacement'> &
+  Pick<
+    BaseInputProps,
+    | 'name'
+    | 'onFocus'
```
kamaleshs-bridge4: 🔵 **[MINOR]** · _api-decision-critique_ · confidence: 8/10

**Problem:** onFocus/onBlur inherit FormInputOnEvent (value?: string) while onChange/onChangeEnd return value: number — the same component hands a consumer a number from one callback and a string from the other, forcing Number() parsing in focus/blur handlers (SliderInput.web.tsx:330 passes String(currentValue))

**Suggestion:** Override onFocus/onBlur with a slider-specific signature (e.g. ({ name?: string; value: number }) => void) so all four event callbacks agree on the value type

packages/blade/src/components/BaseAnimatedValue/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  /** Overrides the direction derived from `value`. */
+  direction?: AnimatedValueDirection;
+
+  /**
+   * Motion duration token for the swap.
+   *
+   * Worth passing when this sits inside something that animates on its own timing, so the two
+   * stay in step when either is retuned rather than agreeing by coincidence.
+   *
+   * @default 'xquick'
```
kamaleshs-bridge4, line 43: 🔵 **[MINOR]** · _code-quality-critique_ · confidence: 7/10

**Problem:** JSDoc @default says 'xquick' but the actual default in BaseAnimatedValue.web.tsx (line 188) is 'moderate'

**Suggestion:** Change @default 'xquick' to @default 'moderate' to match the implementation

kamaleshs-bridge4 (dismissed):
## ✨ Agentic PR Review ✨

**Status:** Approved ✅

### UI Review

✅ 11 passed

<details>
<summary>Passing checks (11)</summary>

| Check | Screenshot |
|-------|------------|
| ✅ [SliderInput Default](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--default) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--default-1.png"></a> |
| ✅ [SliderInput WithMarkers](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--with-markers) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--with-markers-2.png"></a> |
| ✅ [SliderInput WithScale](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--with-scale) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--with-scale-3.png"></a> |
| ✅ [SliderInput LabelPositionLeft](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--label-position-left) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--label-position-left-4.png"></a> |
| ✅ [SliderInput Disabled](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--disabled) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--disabled-5.png"></a> |
| ✅ [SliderInput UnevenSteps](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--uneven-steps) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--uneven-steps-6.png"></a> |
| ✅ [SliderInput Steps](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--steps) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--steps-7.png"></a> |
| ✅ [SliderInput ValidationStates](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--validation-states) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--validation-states-8.png"></a> |
| ✅ [SliderInput FormattedValue](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--formatted-value) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--formatted-value-9.png"></a> |
| ✅ [SliderInput PairedWithTextInput](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--paired-with-text-input) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--paired-with-text-input-10.png"></a> |
| ✅ [SliderInput NarrowWidths](https://61c19ee8d3d282003ac1d81c-gewlygbosa.chromatic.com/?path=/story/components-input-sliderinput--narrow-widths) | <a href="https://raw.githubusercontent.com/razorpay/blade/__ci_artifacts/artifacts/review/PR-3989/ui-critique/18-06-03/components-input-sliderinput--narrow-widths-11.png"></a> |

</details>

### Usage

```jsx
import { SliderInput } from '@razorpay/blade/components';

<SliderInput label="Volume" min={0} max={100} step={25} showMarkers showScale />
```

packages/blade/src/components/Input/SliderInput/utils.ts
```diff
@@ -0,0 +1,256 @@
...
+  if (snapped >= lastWholeStep && max - value < value - lastWholeStep) return max;
+
+  return clamp(roundToStep(snapped, step), min, max);
+};
+
+/**
+ * Every value that gets a marker: one per step, plus `max` when it is not already a stop.
+ */
+const getMarkerValues = ({ min, max, step }: ValueRange): number[] => {
+  if (step <= 0 || max <= min) return [min, max];
```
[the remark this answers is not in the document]
saurav12: Fixed in 5dd639271. `getMarkerValues` now returns `[min]` when `min === max`, so there are no duplicate keys. Added a unit test.

packages/blade/src/components/Input/SliderInput/useSliderInput.web.ts
```diff
@@ -0,0 +1,257 @@
...
+  const stopDragging = React.useCallback(
+    (event: React.PointerEvent<HTMLDivElement>) => {
+      if (!isDragging) return;
+      if (event.currentTarget.hasPointerCapture?.(event.pointerId)) {
+        event.currentTarget.releasePointerCapture(event.pointerId);
+      }
+      setIsDragging(false);
+      setIsScrubbing(false);
+      pointerDownX.current = null;
+      endInteraction(getValueAtPointer(event.clientX));
```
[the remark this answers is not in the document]
saurav12: Fixed in 5dd639271. `onPointerCancel` now uses a separate handler that commits the last value the slider reached, not `event.clientX`. Added a test that cancels with `clientX: 0`.

packages/blade/src/components/Input/SliderInput/types.ts
saurav12: Fixed in 5dd639271. `onFocus` and `onBlur` now have a slider-specific signature `({ name?: string; value: number }) => void`, so all four callbacks give a number. Added a test.

packages/blade/src/components/BaseAnimatedValue/__tests__/BaseAnimatedValue.web.test.tsx
saurav12: Checked again. We will keep the current motion.

packages/blade/src/components/BaseAnimatedValue/types.ts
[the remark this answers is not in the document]
saurav12: The `'moderate'` default is intended. The motion was reviewed and we keep it. The JSDoc `@default` still says `'xquick'` and needs a correction to `'moderate'`.

saurav12: Same as the thread above: `'moderate'` is the intended default. The JSDoc needs a correction to match.

saurabhdaware (approved):

PR #3991, opened 2026-09-14
feat(TreeView): add `size` prop with a new `small` size

saurav12:
## Description

Adds a `size` prop to `TreeView` with a new `small` size, for dense surfaces such as sidebars, file trees and long dropdown overlays. `size` is tree-wide and defaults to `medium`, so existing usage is unchanged.

> [!IMPORTANT]
> This PR contains one visual change to the **existing** `medium` size — see "Letter-spacing fix" below. Everything else is gated behind the new prop.

## Changes

**Sizing**

- New `TreeView/treeViewTokens.ts` holds the size-keyed token map; `size` is propagated through `TreeViewContext` and consumed by the chevron, checkbox, title, indentation and LoadMore row.
- Row height is derived from the title's line-height plus `BaseMenuItem`'s vertical padding, rather than rounded to a `size` token, so both sizes land exactly on the design spec.
- Indentation is one chevron slot plus its 4px gap, which is what makes a child's title sit directly under its parent's title at both sizes. It stays on the row's own `padding-left` (not the nested group) so hover / selected / focus backgrounds remain full-bleed at any depth.

| | `small` | `medium` (default) |
| --- | --- | --- |
| Row height | 33px | 36px |
| Title | Body Small | Body Medium |
| Chevron slot / icon | 16 / 12px | 20 / 16px |
| Checkbox | `small` | `medium` |
| Indentation per level | 20px | 24px |

**Letter-spacing fix (affects existing `medium` usage)**

`getTextProps` only maps `size` onto font tokens inside its `variant === 'body'` branch, and `TreeViewItem` called it without a `variant`. `medium` looked correct only because 14px happens to be the function's default, and the `size` argument was ignored entirely — which is why `small` rendered at 14px until this was fixed.

Passing `variant: 'body'` fixes `small` and restores the tracking `medium` should always have had: **`letter-spacing: 0px` → `-0.182px`** (Blade's `letterSpacings.50`, i.e. -1.3% of font size). The `TreeViewLoadMore` label already rendered with this tracking because it uses `<Text>`, so before this fix a single tree had tracked load-more labels sitting beneath untracked item titles.

**`leading` / `trailing` sizing**

These are consumer-provided `ReactNode`s, so `TreeView` does not rewrite their props. Two rules, both documented:

- **Icons follow the tree's `size`** — `leading`, and a trailing icon.
- **Trailing Counter / Badge / Text are always `size="small"`** at both sizes; they are secondary markers and shouldn't out-weigh the title. Counter and Badge default to `medium`, so this must be passed explicitly.

**Stories**

- New `Sizes` and `Leading & Trailing` stories. `Sizes` shows the full anatomy at both sizes, including a working "Show more" row.
- **Fixes the missing docs page** — `TreeView` never set `tags: ['autodocs']`, so `parameters.docs.page` never rendered. This also un-breaks the TreeView link in `DropdownDocs.tsx`, which points at `/?path=/docs/components-treeview--docs`.
- Shortened the story names (e.g. `Dropdown: FilterChip + Footer (Optimiser case)` → `In Dropdown with Filter Chip`).

**Docs**: `_decisions/decisions.md` and the blade-mcp knowledgebase both updated with the sizing contract and the trailing-item rules.

## Additional Information

**Review notes**

- Story renames change every story ID, so Chromatic will treat all of them as new stories rather than diffs. Worth knowing while reviewing the letter-spacing change, which touches every existing TreeView row.
- Pre-existing behaviour, unchanged by this PR but relevant to the new size: `BaseMenuItem` drops its row padding from 8px to 4px under a breakpoint, so on narrow viewports rows become 25px / 28px rather than 33 / 36. The design spec only covers desktop; worth deciding separately whether `small` should keep 8px throughout.

**Tests**: added a `size` suite covering indentation, chevron slot, title typography and the LoadMore offset at both sizes, plus snapshots. Note that neither `toHaveStyle` nor `toHaveStyleRule` can resolve Blade's specificity-hacked styled-components classes (`.c4.c4.c4.c4.c4`), nor the nested `& > [role="treeitem"]` indentation rule — so there's a small `expectCssRule` helper that matches the injected stylesheet against the element's generated class.

All 82 TreeView tests pass; typecheck and lint are clean.

## Component Checklist

- [x] Add KitchenSink Story
- [x] Add changeset


saurabhdaware (commented):
https://61c19ee8d3d282003ac1d81c-shrurnaqkl.chromatic.com/?path=/docs/components-treeview--docs

this page is not loading 🤔

saurav12:
> https://61c19ee8d3d282003ac1d81c-shrurnaqkl.chromatic.com/?path=/docs/components-treeview--docs
> 
> this page is not loading 🤔

Loading for me though.

anuraghazra (dismissed):

kamaleshs-bridge4 (approved):
## ✨ Agentic PR Review ✨

**Status:** Approved ✅

### UI Review

⏭️ 1 skipped

### Usage

```jsx
import { TreeView, TreeViewItem } from '@razorpay/blade/components';

<TreeView size="small">
  <TreeViewItem title="Item" value="item" />
</TreeView>
```

saurabhdaware (approved):

